PR #29, opened 2016-05-18
Fresh 318

metas-ts:


de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/modelvalidator/AnnotatedModelInterceptor.java
```diff
@@ -372,7 +372,12 @@ else if (parameterTypes.length == 2)
 			set = new TreeSet<Pointcut>();
 			mapPointcuts.put(key, set);
 		}
-		set.add(pointcut);
+
```
metas-ts: @teosarca f*** yeah. thx. but i think we should even fail in this case, at east in deveoper mode..i think i'll to that quickly..

PR #38, opened 2016-05-20
FRESH-311 Packvorschriften from different Org shown in Leergut

metas-lc:
new function to take available M_HU_PIs for given org


de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/impl/HandlingUnitsDAO.java
```diff
@@ -485,6 +488,23 @@ public I_M_HU_PI_Version retrievePICurrentVersionOrNull(final I_M_HU_PI pi)
...
 				.list(I_M_HU_PI.class);
 	}
+	
+	@Override
+	public List<I_M_HU_PI> retrieveAvailablePIsForOrg(final Properties ctx, final int adOrgId)
+	{
+		final ImmutableList.Builder<I_M_HU_PI> huPIsForOrg = ImmutableList.builder();
+		for (final I_M_HU_PI huPI : retrieveAvailablePIs(ctx))
+		{
+			if (huPI.getAD_Org_ID() != adOrgId)
```
metas-ts: @metas-lc what if the user chose to create the huPIs on client level, i.e. with AD_Org_ID=0?

PR #43, opened 2016-05-23
FRESH-334 Only allow partners from the same org in BPProduct

metas-rc:


de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5446010_sys_FRESH_334_Modify_BPProduct_Primary_Key.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE c_bpartner_product DROP CONSTRAINT c_bpartner_product_pkey;
```
metas-ts: * when changing the PK DDL, please also check, if there should be an accompanying change int the AD
* i think that
  - introducing c_bpartner_product_id as single-column key and 
  - adding a UC constraint that includes "where isactive='Y' would be better and cleaner
wdyt? if in doubt, i recomment to consult  @teosarca

de.metas.business/src/test/java/de/metas/purchasing/api/impl/BPartnerProductDAOTest.java
```diff
@@ -0,0 +1,230 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+public class BPartnerProductDAOTest
```
metas-ts: **great!** automation in testing is the future :-)

PR #68, opened 2016-06-06
Fresh 349

metas-rc:


de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5446810_mig_FRESH_349_AD_PInstance_in_AD_Archive_Migration.sql
```diff
@@ -0,0 +1,40 @@
+
+DROP VIEW IF EXISTS tmp_RV_IO_TO_Archive;
```
metas-ts: This migration might take too long. I analyzed the view's select on a large DB and it did a sequential scan on both AD_PInstance and M_InOut. 
If we run it during rollout, it is likely to blow our maintainance window to pieces
(i caused this to happen a few times in the past :-$, we can't afford repeating every mistake i made ;-) ).

So, pls decide with @metas-mk if we want to provide this partial migration at all.

If the answer is yes, then please provide a function that only deals with `n` records at a time and can be run with `"de.metas.async".executesqlasync()`

PR #78, opened 2016-06-13
FRESH-280 Period sorting in all Dropdowns where uses year-month numeric

metas-lc:
added C_Period Order reference and updated processes


de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/70-de.metas.fresh/5447100_sys_FRESH-280_C_Period_Order_Reference_and_update_processes.sql
```diff
@@ -0,0 +1,100 @@
+-- 13.06.2016 12:34
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+INSERT INTO AD_Reference (AD_Client_ID,AD_Org_ID,AD_Reference_ID,Created,CreatedBy,EntityType,IsActive,IsOrderByValue,Name,Updated,UpdatedBy,ValidationType) VALUES (0,0,540658,TO_TIMESTAMP('2016-06-13 12:34:50','YYYY-MM-DD HH24:MI:SS'),100,'U','Y','N','C_Period Ordered',TO_TIMESTAMP('2016-06-13 12:34:50','YYYY-MM-DD HH24:MI:SS'),100,'T')
```
metas-ts: imho this whole script belongs into /de.metas.adempiere.adempiere.migration-sql/src/main/sql/postgresql/system/10-de.metas.adempiere 

Also, please use the correct entity type (for the location above, it's 'D'). imho you can search&replace 'U' with the correct entity type,  rollout_mf15_test_migration_scripts  will check it it's valid.

metas-ts: Hm, wait..it's all about different columns which are all over the place, right?

PR #79, opened 2016-06-13
Fresh 309

metas-rc:


de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/70-de.metas.fresh/5447110_mig_FRESH-309_Insert_Missing_M_AttributeInstance_Entries_For_InvoiceLines.sql.sql
```diff
@@ -0,0 +1,106 @@
+
```
metas-ts: about all three files: 
* pls use 99 instead of 100 for the createdBy and updatedBy that way we can better identify records that were updated or inserted made via such a migration
* i tried the select parts on a big customer-DB and there were no columns selected for any of the three. and the selects returned after <100ms..i think there is a bug..
* pls remove the double ".sql" from the filename (ok, that's just me beeing picky)

PR #81, opened 2016-06-13
FRESH-392: BP_Group and BP_Group_Acct entries moved to org 0

metas-rc:


metas-ts:
Closing without integrating, as skyped. This change is not wanted by every user, but custom to a particular client.

PR #111, opened 2016-06-24
Fresh 378a

metas-rc:


de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoice/model/validator/C_Invoice.java
```diff
@@ -13,11 +13,11 @@
  *
```
metas-ts: we are not (yet?) totally consistent on this, but: i suggest as a best practice to have MI-methods in the "home" module of the records they influence. 
Here, it's a C_Invoice change that has an influences on C_Invoice_Candidate, so please move it to a C_Invoice-MI within de.metas.invoicecandiate

de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/modelvalidator/C_Invoice_Candidate.java
```diff
@@ -421,4 +407,31 @@ public void errorIfTaxNotFound(final I_C_Invoice_Candidate candidate)
...
 	}
+
+	/**
+	 * Mark the candidate as Processed_Override = Y because in case it has IsToClear on true. This kind of candidate is not relevant for the invoice
+	 * Could happen when the partner has a subscription for the kind of product in the invoice candidate
+	 * 
+	 * @param candidate
+	 */
+	@ModelChange(timings = { ModelValidator.TYPE_AFTER_NEW, ModelValidator.TYPE_AFTER_CHANGE }, ifColumnsChanged = { I_C_Invoice_Candidate.COLUMNNAME_IsToClear })
+	public void ProcessIfIsToClear(final I_C_Invoice_Candidate candidate)
```
metas-ts: pls make the method lowercase :-)

PR #179, opened 2016-07-07
Fresh 504

metas-rc:


metas-ts:
OK to integrate after the SQL proved OK

Issue #100, opened 2016-06-22
EDI wrong handover location in Picking Terminal


metas-mk:
EDI wrong handover location in Picking Terminal
* see c_olcand > OK
* see c_order > NOK
* see m_shipmentschedule > NOK
* see Picking Terminal > NOK

metas-mk:
Internal task key: FRESH-435

PR #189, opened 2016-07-11
Fresh-435 #100 EDI wrong handover location in Picking Terminal

metas-ts:


metas-ts:
ok for integrateion after all migration scripts proved OK

Issue #232, opened 2016-07-21
Separate c_flatrate_terms from the normal procurement and RfQ in procurement Excel


metas-dh:
case from IT for https://metasfresh.atlassian.net/browse/FRESH-402: 
* a bpartner has a "normal" procurement c_flatrate_term for product P000157, and also takes part in RfQ bidding for the same product
* if he wins the bidding, a special c_flatrate_term is created for the RfQ of P000157, as well as procurement candidates
* when you run the procurement Excel for this bpartner, there is only 1 line for P000157, for the  "normal" procurement c_flatrate_term, with the prices from this term for each month
* BUT: qty promised  includes the qties from procurement candidates that were created in the RfQ process that the bpartner one, which had different price! (qty delivered will prob. also be included, see https://metasfresh.atlassian.net/browse/FRESH-191)
* => the data from the 2 c_flatrate_terms  get mixed up in Excel, NOK
* => separate the data, maybe with an additional line for the c_flatrate_term from RfQ?

PR #238, opened 2016-07-25
FRESH-546 #232 Separate c_flatrate_terms from the normal procurement and RfQ in procurement Excel

metas-lc:
added isRfQConditions in C_Flatrate_Conditions, filtered out rfq data
from procurementx excel


de.metas.procurement.base/src/main/sql/postgresql/ddl/functions/docs_flatrate_term_all_procurements_conditions.sql
```diff
@@ -452,11 +452,13 @@ SELECT	C_Flatrate_Term_ID,
 
 FROM C_Flatrate_Term ft
 JOIN PMM_Product p ON p.PMM_Product_ID = ft.PMM_Product_ID
+JOIN C_Flatrate_Conditions fc ON ft.C_Flatrate_Conditions_ID = fc.C_Flatrate_Conditions_ID
 WHERE ft.type_conditions='Procuremnt' AND ft.isActive='Y'
 	AND ft.startdate::date<=now()::date AND ft.enddate::date>=now()::date
 	AND ft.PMM_Product_ID IS NOT NULL AND Bill_BPartner_ID IS NOT NULL
 	AND COALESCE(ft.Bill_BPartner_ID = $1, true)
 	AND COALESCE(p.M_Product_ID = $2, true)
+	AND NOT EXISTS (SELECT r.C_Flatrate_Conditions_ID FROM C_RfQ r WHERE fc.C_Flatrate_Conditions_ID = r.C_Flatrate_Conditions_ID)
```
metas-ts: i checked the datamodel a bit: there is a column `C_RfQResponseLine.C_Flatrate_Term_ID`; the term is created and referenced from there when the RfQ is closed (thx @metas-dh ).
So, what about "beeing referenced by a C_RfQResponseLine" to be the criterion to exclude a term?

Asking because I'm a bit uneasy about excluding a term because its conditions is referenced from some RfQ..you can spoil the excel just by creating a C_RfQ record and selecting a particular conditions.

Btw, please generally remember to ignore inactive records and document that are not completed or closed.

metas-ts:
made a minor change; waiting to see if "rollout_mf15_test_migration_scripts" passes

Issue #243, opened 2016-07-26
C_Invoice_Candidate - Processed not always updated if IsToClear


metas-ts:
`C_Invoice_Candidates` that are flagged with `IsToClear=Y` are not updated asnychronously.
they are updated by a model intersceptor, But if this fails for whatever reason,
they are stuck and a user would need to manually unset and re-set their `Processed_Override` value.

PR #244, opened 2016-07-26
FRESH-552 #243 making sure that Processed_Override is honored...

metas-ts:
..also if IsToClear=Y and if the IC is updated via async processor (not
model interceptor)


metas-ts:
our build system has problems with hashes in branch names. i created https://github.com/metasfresh/metasfresh/pull/245 instead

Issue #293, opened 2016-08-16
Create sortof ad-hoc dlm for the async-tables


metas-ts:
the case of async tables is rather simple,
and if they grow large, they have a high impoact on overal performance,
due to the way async is currently implemented

PR #294, opened 2016-08-16
FRESH-592 #293 changes & additions:

metas-ts:
* moving existing SQL to 42-de.metas.async folder
* adding 80-async-dlm folder
* adding tables and function to move old async workpackage records to
"archive" files in the dlm schema


metas-ts:
check if the build succeeds, and if the artifact contains the del migration scripts

PR #318, opened 2016-08-22
Fresh 579

metas-lc:


metas-ts:
@metas-lc for such changes, it's perfectly OK if you merge those pull requests yourself :-).
Just let me know if you need details on how to do it.

PR #325, opened 2016-08-24
FRESH-580 applying default filter to summary in invoice candidates

metas-rc:


de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandDAO.java
```diff
@@ -1330,11 +1330,33 @@ public final boolean isToRecompute(final I_C_Invoice_Candidate ic)
...
+	public String getSQLDefaultFilter(final Properties ctx)
+	{
+		// Only filter invoice candidates of the organizations this role has access to
+		final IUserRolePermissions userRolePermissions = Env.getUserRolePermissions(ctx);
+
+		final StringBuilder defaultFilter = new StringBuilder("");
+
+		final String orgIDsAsString = userRolePermissions.getAD_Org_IDs_AsString();
+
+		if (!orgIDsAsString.isEmpty())
```
metas-ts: * unless you are 110% sure that orgIDAsString is never null, please use `!Check.isEmpty(...)` instead.
* if you are sure, then please add something line"Never return <code>null</code>" to the javadoc of getAD_Org_IDs_AsString().

PR #381, opened 2016-09-13
Fresh 653

metas-lc:


de.metas.swat/de.metas.swat.base/src/main/java/de/metas/request/service/async/spi/impl/C_Request_CreateFromInout.java
```diff
@@ -0,0 +1,147 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
```
metas-ts: please add documentation about that the class does

de.metas.swat/de.metas.swat.base/src/main/java/de/metas/request/api/IRequestTypeDAO.java
```diff
@@ -0,0 +1,37 @@
...
+ * 
+ * You should have received a copy of the GNU General Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+public interface IRequestTypeDAO extends ISingletonService
+{
+
```
metas-ts: plese document which request type is the vendor one and which request type is the customer one. also, can it return null, or is there an exception in case the type is not unique?

de.metas.swat/de.metas.swat.base/src/main/java/de/metas/request/service/IRequestCreator.java
```diff
@@ -0,0 +1,41 @@
...
+{
+
+	/**
+	 * Create an R_Request based on a list of M_InOutLine IDs
+	 * 
+	 * @param ctx
+	 * @param inOutLineIds
+	 * @param trxName
+	 */
+	void createRequest(final Properties ctx, final List<Integer> docLineIds, final String trxName);
```
metas-ts: might be not needed to be *implemented* today, but please document that it only works with inoutlines and sketch out how to extend in fututre.

metas-ts: btw, what happens if someone calls this with a bunch of C_OrderLine_IDs? are there any checks down the road? did you consider adding I_M_InOutLines instead of the IDs?

PR #390, opened 2016-09-14
Fresh 661

metas-lc:


de.metas.endcustomer.mf15/de.metas.endcustomer.mf15.base/src/main/jasperreports/de/metas/docs/purchase/order/report.properties
```diff
@@ -47,6 +47,7 @@ totalsum=ENDSUMME
 paymentcond=Zahlungsbedingungen:
 
 page=Seite
+Page=Seite
```
metas-ts: why do we have both "Page" and "page"? this smells a bit like future trouble...

Issue #423, opened 2016-09-22
Dropdown List shows wrong records on rare cases


metasnw:
case 1 : 

1. open window price
2. go to price list and enter Base Pricelist
3. go to price list version and have a record where Base Pricelistversion is filled
4. grid view
5. duplicate record
6. NOK: field Base Pricelistversion shows entries not matching the base price list selected in 2.

impact:
occurs on single pc, not reproducable until now

next step:
detail logging enabled...

metas-ts:
Please note:

I cherry-picked @teosarca 's fix into the branch fresh-708-gh423. which was branched off of tag [v4.36.35](https://github.com/metasfresh/metasfresh/releases/tag/v4.36.35). I had too fix some conflicts there.
that fix was successfully tested on the customer site.

Now, I also cherry-picked the fix into master, without conflicts.

PR #434, opened 2016-09-24
Fresh 708 gh #423

metas-ts:


metas-ts:
not going to integrate this branch into master. the actual fix was areleady cherry-picked (without conflicts).
also see my recent comment on #423

PR #511, opened 2016-10-25
Fresh 783

metas-rc:


de.metas.business/src/main/java/org/adempiere/mm/attributes/api/IAttributeDAO.java
```diff
@@ -190,4 +189,14 @@
 	 * @return true if given attribute is expected to have a huge amount of {@link I_M_AttributeValue}s.
 	 */
 	boolean isHighVolumeValuesList(I_M_Attribute attribute);
+
+	/**
+	 * Retrieve an M_AttributeValue entry by it's attribute and Name.
+	 * Only use if you are sure there aren't several attribute values with the same name!
```
metas-ts: Since you use this method, how do you make sure there aren't several attribute values with the same name ;-) ?

AFAIU, since every admin-user might add attribute values, a dev calling this method can never be sure, unless he/she writes it's own list-retrieval method and checks there retunred lists's size, before calling your method ;-).
Checking a large client's DB, i found that there are in fact dublicate names.
Also note that the attribute values can have not-0 AD_Client_IDs and AD_Org_IDs.

So i would propose to 
* add a UC constraint on isactive='Y' that also includes attribute-ID, name, org-ID and client-ID
now, and we would clean up the doubles from their DB
* also provide the current ctx to the method, and return the first match (preferning AD_Org_ID>0, AD_Client_ID>0 etc)

de.metas.business/src/main/java/org/adempiere/mm/attributes/api/impl/AttributeDAO.java
```diff
@@ -126,6 +125,19 @@ public I_M_AttributeValue retrieveAttributeValueOrNull(final I_M_Attribute attri
 	}
 
 	@Override
+	public I_M_AttributeValue retrieveAttributeValueOrNull_ForName(final I_M_Attribute attribute, final String name)
+	{
+
+		return Services.get(IQueryBL.class)
+				.createQueryBuilder(I_M_AttributeValue.class, attribute)
+				.addEqualsFilter(I_M_AttributeValue.COLUMN_M_Attribute_ID, attribute.getM_Attribute_ID())
```
metas-ts: please *always* filter for active records, unless you have an explicit reason not to

de.metas.swat/de.metas.swat.base/src/main/sql/postgresql/system/40-sw01_swat/5452310_mig_FRESH-783_Migrate_Old_Inouts_async.sql
```diff
@@ -0,0 +1,167 @@
+
+
+
+CREATE OR REPLACE FUNCTION create_request_document_number ()
```
metas-ts: I suggest that you also prepend a `temp_` in front of this function, and add a [comment](https://www.postgresql.org/docs/9.1/static/sql-comment.html) to both the function or the table to indicate the github issue-URL and and that they are not required anymore after the migration finished.

de.metas.swat/de.metas.swat.base/src/main/sql/postgresql/system/40-sw01_swat/5452310_mig_FRESH-783_Migrate_Old_Inouts_async.sql
```diff
@@ -0,0 +1,167 @@
...
+
+$BODY$
+LANGUAGE sql volatile;
+
+
+
+
+DROP TABLE IF EXISTS temp_M_InOutLine_To_R_Request; 
+
+CREATE TABLE temp_M_InOutLine_To_R_Request
```
metas-ts: To me it looks as if this `CREATE TABLE` actually does the main work, iterating those 1.8m inout lines metioned by mark. Therefore i think this does not solve the problem of the migration taking too long for the rollout window.

**Update:** i just checked at our biggest client, with the from, join and where clause from this select:
```
SELECT count(iol.*)
FROM M_InOutLine iol
 	JOIN M_InOut io on iol.M_InOut_ID = io.M_InOut_ID
 WHERE iol.QualityDiscountPercent > 0
 		and iol.IsActive = 'Y' 
 		and io.IsActive = 'Y'
  		and io.MovementDate >= '2015-01-01'
```
**Returns: 5679 rows.**

@metas-mk @metas-rc : please clarify how make iol we really need to migrate.

de.metas.swat/de.metas.swat.base/src/main/sql/postgresql/system/40-sw01_swat/5452310_mig_FRESH-783_Migrate_Old_Inouts_async.sql
```diff
@@ -0,0 +1,167 @@
...
+
+)
+;
+
+
+ 
+
+
+-- execute async
+SELECT "de.metas.async".executesqlasync('
```
metas-ts: for a large `temp_M_InOutLine_To_R_Request` this single insert might still be problematic.
Before doing it, we would need to look into the documentation to see which locks such an INSERT would aquiere and which other operations it might block.

The whole idea of the `executesqlasync` function is to split a job into many chunk, each one running in a relatively short transaction.

de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/POInfoColumn.java
```diff
@@ -88,23 +90,39 @@ public POInfoColumn(int ad_Column_ID, String columnName, String columnSQL, int d
 		}
 
 		int displayTypeToSet = displayType;
-		if (columnName.equals("AD_Language") || columnName.equals("EntityType"))
+
+		ITableRefInfo tableRefInfo = null;
+
+		if (ad_Reference_Value_ID > 0 && (isTableDisplayType(displayType) || isSearchDisplayType(displayType)))
+		{
+			// permit warnings because we shouldn't have Table or Search types with no reference table IDs.
```
metas-rc: Docu added

de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/POInfoColumn.java
```diff
@@ -124,14 +142,36 @@ else if (Services.get(IColumnBL.class).isRecordColumnName(columnName))
...
+
+		toBigDecimalOrNull(ValueMin, "ValueMin");
 		ValueMax = valueMax;
 		ValueMax_BD = toBigDecimalOrNull(ValueMax, "ValueMax");
 		IsTranslated = isTranslated;
 		IsEncrypted = isEncrypted;
 		IsAllowLogging = isAllowLogging;
 	}   // Column
 
+	private boolean isSearchDisplayType(int displayType)
```
metas-rc: Done

de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/api/impl/QualityNoteDAO.java
```diff
@@ -0,0 +1,47 @@
...
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+public class QualityNoteDAO implements IQualityNoteDAO
+{
+
+	@Override
+	public I_M_QualityNote retrieveQualityNoteForName(final Properties ctx, final String name)
+	{
```
metas-ts: to check: do we have a UC on name?

de.metas.swat/de.metas.swat.base/src/main/sql/postgresql/system/40-sw01_swat/5452470_sys_FRESH-783_QualityNote_Table.sql
```diff
@@ -0,0 +1,239 @@
...
+;
+
+-- 26.10.2016 17:56
+-- URL zum Konzept
+INSERT INTO AD_Table_Trl (AD_Language,AD_Table_ID, Name, IsTranslated,AD_Client_ID,AD_Org_ID,Created,Createdby,Updated,UpdatedBy) SELECT l.AD_Language,t.AD_Table_ID, t.Name, 'N',t.AD_Client_ID,t.AD_Org_ID,t.Created,t.Createdby,t.Updated,t.UpdatedBy FROM AD_Language l, AD_Table t WHERE l.IsActive='Y' AND l.IsSystemLanguage='Y' AND l.IsBaseLanguage='N' AND t.AD_Table_ID=540794 AND NOT EXISTS (SELECT * FROM AD_Table_Trl tt WHERE tt.AD_Language=l.AD_Language AND tt.AD_Table_ID=t.AD_Table_ID)
+;
+
+-- 26.10.2016 17:56
+-- URL zum Konzept
+CREATE SEQUENCE M_QUALITYNOTE_SEQ INCREMENT 1 MINVALUE 0 MAXVALUE 2147483647 START 1000000
```
metas-ts: Please separate DDL and DML and add a commit, between the two, to avoid "pending trigger events" error

de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/api/impl/QualityNoteDAO.java
metas-ts: yes we do

PR #602, opened 2016-11-15
Fresh 802

metas-rc:


de.metas.procurement.base/src/main/sql/postgresql/system/50-de.metas.procurement/5453400_sys_FRESH-802_FlatrateTermCreateForPartners_ExcludeProcurement.sql
```diff
@@ -0,0 +1,5 @@
+-- 15.11.2016 15:42
+-- URL zum Konzept
+UPDATE AD_Val_Rule SET Code='C_Flatrate_Conditions.Type_Conditions NOT IN (''FlatFee'',''HoldingFee'', ''Subscr'', ''Procuremnt'')',Updated=TO_TIMESTAMP('2016-11-15 15:42:30','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Val_Rule_ID=540237
```
metas-ts: i think we should rather have "whitelist" at this point, but we can as well do that next time we touch this part

Issue #677, opened 2016-12-07
make customs report faster


metas-lc:

metas-dh:
Results of IT 1 

* App.Server: sp80cloudit
* Buildno: #1.156
* DB: declouddb001 : 5432 / sp80_cloud_it

https://github.com/metasfresh/metasfresh-documentation/blob/master/_tests_collection/Testcase_FRESH-978.md: OK

PR #678, opened 2016-12-07
FRESH-978 #677 make customs report faster

metas-lc:
update sql from purchase customs inout jasper
take out sqls from jaspers into functions


metas-ts:
as of now, the PR should rather aim at master, not stable (sry @metas-lc for the confusion)
creating a new PR

PR #671, opened 2016-12-06
FRESH-896 #626 Initial Shipment Window Layout for webui

metas-mk:
Initial Definition of Shipment Window in Webui. Including Tabs, Fields, Grid and extended View and also Sidelist.

PR #680, opened 2016-12-08
FRESH-990 #679 Bug in ClientUpdateValidator

metas-ts:
successfully tested with a local maven build


metas-ts:
i made a mistake vcreating the branch, now we have the problem of #671 in here.
waiting a bit until #671 is fixed

Issue #695, opened 2016-12-12
swingUI: provide lib-dirs to access user jars and dlls at runtime 


homebeaver:
see https://forum.metasfresh.org/t/schnittstellt-zu-ms-com-excel-outlook-via-groovy/98
for a documentation of this issue.
This will be the first part of the implementation - the jars needed.

The second will be the dlls

homebeaver:
I created a pull request #697 which changes the pom in de.metas.endcustomer.mf15.swingui

homebeaver:
I changed the title to match the gereral goal.

metas-ts:
Hi @homebeaver,
on the question of how to make external libs available for metasfresh, I found [this stackoverflow question](http://stackoverflow.com/questions/30745132/external-library-folder-for-spring-boot), and bascially, the answer seems to be that one can add the respective paths to the application's `application.property` file.
The documentation is [here](https://docs.spring.io/spring-boot/docs/current/reference/html/executable-jar.html#executable-jar-property-launcher-features).

Right now, the metasfresh-swingUI client does not come with such a file, but I'm pretty sure that one can either place it next to the "main" jar (in the `lib` folder) or one can provide the values as command line properties (like for windows `-Dloader.path=".\userlib,.\userlib-x64"`).

So, asuming that we can sort out the part of actual making available the external libaries quickly, the next question I see is *how* the metasfresh-client shall access them.

The simplest answer coming to my mind is to create a model interceptor (a.k.a. model validator), place it into an external jar and add a row which references that model interceptor to the `AD_ModelValidator` table.
That table is read on startup. The system finds the class name in the `ModelValidationClass` column, loads and instantiates it and that interceptor can then register the handlers for the model and document changes on which the system is supposed to interact with the COM bridge.

So, it actually would be like any other module with the only difference that it's placed in an external jar.

To see how such a model interceptor is done more concretely, I recomment to take a look at e.g. [metasfresh-dlm](https://github.com/metasfresh/metasfresh/tree/master/de.metas.dlm). It's started with a module interceptor named [Main](https://github.com/metasfresh/metasfresh/blob/master/de.metas.dlm/src/main/java/de/metas/dlm/model/interceptor/Main.java) which extends [AbstractModuleInterceptor](https://github.com/metasfresh/metasfresh/blob/master/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/modelvalidator/AbstractModuleInterceptor.java).
[...]

PR #697, opened 2016-12-12
runtime support for COM-api, issue #695

homebeaver:
find a documentation on https://forum.metasfresh.org/t/schnittstellt-zu-ms-com-excel-outlook-via-groovy/98

This will be the first part to support COM-api (microsoft outlook,excel,...) - the jars needed.
On runtime the swingUI needs also the jacob-1.18-x64.dll (or jacob-1.18-x86.dll respectively) in the PATH


homebeaver:
Thank you for review @metas-ts,
yes @metasnw told us, that he is not willing to include the jacob/scriptom the mf-client classpath.

As our customer needs this api to outlook and excel, we need either this or another implemented alternative to satisfy the requirement.

metas-ts:
Hi, @homebeaver @metasnw 
FYI, i'm closing this PR in the context of what was written further up. As i already wrote,

> please let me and/or @teosarca know if and when you need further advise, discussion or even the implementation of this approach.

of course the implementation would need further organisation and prioritization which you would need to do together with @metasnw and your customer..

PR #701, opened 2016-12-12
Fresh 934

metas-rc:


de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/util/TimeUtil.java
```diff
@@ -1107,4 +1106,46 @@ public static final Timestamp copyOf(final Timestamp timestamp)
...
+		final Calendar cal = asCalendar(date);
+		cal.set(Calendar.HOUR_OF_DAY, 0);
+		cal.set(Calendar.MINUTE, 0);
+		cal.set(Calendar.SECOND, 0);
+		cal.set(Calendar.MILLISECOND, 0);
+
+		cal.setFirstDayOfWeek(Calendar.MONDAY);
+
+		// FIXME: This shall be taken from Locale, but Locale it is not reliable (doesn't always work the same way)
+		cal.setMinimalDaysInFirstWeek(4);
```
metas-ts: Hi @metas-rc, 
it's good that you have a tests for this,
but please also describe why you chose `minimalDaysInFirstWeek=4` and not e.g. 3 or 5, and where the analog postgresql function is taking that number from.

The background to this request is that I feel a lot of FUD around the number-of-week, mostly due to ignorance. 
On jira the number is different than it is on my Outlook, and if we print this on a HU-label we need to know which number we print and why we print it (and ofc we need to make sure that the number returned by postgresql is the same returned by java, and why it is the same).
And I think if i'm ignorant and feel FUD, most others also will, unless we have it documented.

metas-ts:
@metas-rc 
I integrated the change, but didn't yet delete the branch.
Pls add the documentation whan you have time

Issue #710, opened 2016-12-14
MRP Product Info: Qtyies issued to a production shall be subtracted from onhand qty


metas-ts:
Orig: I as an user want the qtyies issued to a production be subtracted from storage on receipt.
! Case is about Action_Issue (not receipt). After performing action_issue the stock (qtyonhand) in MRP Product Info for the issued raw material shall be adjusted accordingly.

metas-ts:
Just merged a WI. Please move back to me for finalization once its ready for monday's demo

metas-dh:
Results of IT 1 

* App.Server: mf15cloudit
* Buildno: #1.319
* DB: declouddb001 : 5432 / mf15_cloud_it

* Qty onhand correctly reduced when qty issued in production process: OK
* Qty onhand correctly inceased again when issue was reversed (in Cost Collector): NOK, qty remains the same

PR #728, opened 2016-12-16
FRESH-905 #710 MRP Product Info: Qtyies issued to a production shall be subtracted from onhand qty

metas-ts:
Change to trigger mrp product info changes from M_Transaction records


metas-ts:
Just merged a WI. Please move back to me for finalization once its ready for monday's demo

Issue #712, opened 2016-12-15
Extending C_Flatrate_Term shall respect Flag shallComplete


metas-mk:
Extending C_Flatrate_Term shall respect Flag shallComplete

metas-ts:
could not reproduce any issue, neither could find any case on the customer's DB

PR #751, opened 2016-12-22
FRESH-1035 #712 Respect Complete flag in Extending terms

metas-rc:


de.metas.contracts/src/main/java/de/metas/flatrate/api/impl/FlatrateBL.java
```diff
@@ -431,20 +428,11 @@ private I_C_Invoice_Candidate createCand(
 		final boolean isSOTrx = true;
 
 		final int taxId = Services.get(ITaxBL.class).getTax(
-				ctx
-				, term
-				, taxCategoryId
-				, productId
-				, -1 // chargeId
+				ctx, term, taxCategoryId, productId, -1 // chargeId
```
metas-ts: hi, @metas-rc 
it's not a big problem, but did you import the formatter (and other) settings from https://github.com/metasfresh/metasfresh-eclipse-config after your recent upgrade to neon?

metas-ts:
Hi, as written in skpye, i don't think the change actually solves a problem.
@metas-mk is talking to the user once more to figure what the problem actually is.

PR #756, opened 2016-12-23
Fresh 890v3

metas-rc:


de.metas.procurement.base/src/main/java/de/metas/procurement/base/IPMMProductBL.java
```diff
@@ -45,4 +48,17 @@
...
+	/**
+	 * Get the PMM_Product for the date, product, partner and M_HU_PI_Item_Product that fit the given ASI the most.
+	 * This means the attribute instances from the ASI of the PMM Product must be a subset of the given ASI's attribute instances
+	 * 
+	 * @param date
+	 * @param productId
+	 * @param partnerId
+	 * @param huPIPId
+	 * @param asi
+	 * @return
```
metas-ts: Also, what shall happen if two pmmProducts are different but similar matches? e.g. the order line has 
Bio and ADR and there is one pmmProduct for Bio and another one for ADR? 
Solutions i see:
* throw an exception so the user needs to alter the masterdata before they can continue with the order
* introduce a SeqNo column to PMM_Product and pick the one with the lower SeqNo in this case. The UC of the PMM_Product table should then conntain the SqlNo as opposed to the ASI, the SeqNo should be
 * either always mandatory and updated by an SQL default rule or tab callout
 * or the UC needs to contain a "coalesce" SQL expression to make sure to handle NULL values

metas-rc: I chose to use the UC with COALESCE because in case of simple PMM_Product entries with no attributes the seqNo is not needed so it could stay null.
If we need the seqNo for other purposes in the future we will probably need to make it mandatory, but not at the time being.

Issue #766, opened 2017-01-04
fix for "DocumentPA will not be intercepted because final classes are not supported"


metas-ts:
```org.adempiere.util.exceptions.ServicesException: WARNING: Class class de.metas.document.impl.DocumentPA will not be intercepted because final classes are not supported
```
this plus a big stacktrace is shown on metasfresh (server?) startup. 
The issue was there all the time, but now surfaces in an annoying way

metas-dh:
Integrated, built and deployed for integration test

PR #767, opened 2017-01-04
FRESH-1101 #766: make the class not-final; assign cacheName to @Cached

metas-ts:
FRESH-1101 #766 fix for "DocumentPA will not be intercepted because
final classes are not supported"

### DIFF
--- a/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/document/impl/DocumentPA.java
+++ b/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/document/impl/DocumentPA.java
@@ -13,15 +13,14 @@ package de.metas.document.impl;
  * 
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


de.metas.swat/de.metas.swat.base/src/main/java/de/metas/document/impl/DocumentPA.java
```diff
@@ -202,7 +201,7 @@ private int performUpdate(final int bPartnerLocationId, final String sql,
 		}
 	}
 
-	@Cached
+	@Cached(cacheName = Table_Name + "#by#" + COLUMNNAME_AD_Org_ID + "#" + COLUMNNAME_DocBaseType + "#" + COLUMNNAME_DocSubType + "#" + COLUMNNAME_IsDefault)
```
metas-ts: The cache name is needed to make sure metasfresh know to reset this case, when the table name changes

de.metas.swat/de.metas.swat.base/src/main/java/de/metas/document/impl/DocumentPA.java
```diff
@@ -51,55 +50,55 @@
 import de.metas.inout.model.I_M_InOut;
 import de.metas.interfaces.I_C_DocType;
 
-public final class DocumentPA implements IDocumentPA
+public class DocumentPA implements IDocumentPA
```
metas-ts: this is the actual fix

PR #769, opened 2017-01-04
Fresh 890v3

metas-rc:


de.metas.procurement.base/src/main/java/de/metas/procurement/base/impl/PMMPricingBL.java
```diff
@@ -164,7 +164,7 @@ public boolean updatePriceFromContract(final IPMMPricingAware pricingAware)
 		}
 		
 		final BigDecimal flatrateAmtPerUOM = dataEntryForProduct.getFlatrateAmtPerUOM();
-		if (flatrateAmtPerUOM == null || flatrateAmtPerUOM.signum() <= 0)
+		if (flatrateAmtPerUOM == null || flatrateAmtPerUOM.signum() < 0)
```
metas-ts: yup, i think that's it

de.metas.procurement.base/src/main/sql/postgresql/system/50-de.metas.procurement/5455010_sys_FRESH-890_SeqNoInPMMProduct.sql
```diff
@@ -0,0 +1,93 @@
...
+
+
+
+
+UPDATE PMM_Product SET SeqNo = 10 where M_AttributeSetInstance_ID IS  NULL;
+
+UPDATE PMM_Product SET SeqNo = 20 where M_AttributeSetInstance_ID IS NOT NULL;
+
+
+--- !!!!!!!!!!!!!!!!!! IF THERE ARE MORE THAN 2 PMM_PRODUCT ENTRIES FOR THE SAME PRODUCT, M_HU_PI_ITEM_PRODUCT AND PARTNER, PLEASE UPDATE THE SEQNO MANUALLY HERE!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
metas-ts: Phew, that's a bit problematic...
we usually only have a 15 minutes rollout window. If this SQL fails because some user created some `PMM_Prodcut`s, we have a problem and might not be able to fix things in time..

It should be possible to solve this with an update that's getting the `SeqNo` valued from a select with a *window function*
..though i don't know how exactly, but am pretty sure it wou,ld work... maybe @metas-lc can help?

de.metas.procurement.base/src/test/java/de/metas/procurement/base/impl/PMMProductBLTest.java
```diff
@@ -0,0 +1,519 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+public class PMMProductBLTest
```
metas-ts: nice, thx :-) for the next test, i recomment to also add brief javadocs that outline what the test is doing, 
like e.g. "Verifies that if there is just a PMM_Product whith a non-matching ASI, then {@code null} is returned"

Issue #770, opened 2017-01-04
When extending a procurement contract, null becomes 0.00


metas-ts:
TODO: 
* fix it in the code
* provide an SQL to update 0,00 to `null` (we assume that the current users are not willingly giving stuff away for free)

metas-dh:
Integrated, built and deployed for integration test

PR #772, opened 2017-01-04
FRESH-1103 #770 When extending a procurement contract, null becomes 0.00

metas-ts:


metas-ts:
Note: only [this commit](https://github.com/metasfresh/metasfresh/pull/772/commits/a88d56a8eba550a256671d1a030586f963a65188) is directly about this issue. The others are about updating the jmockit library

Issue #759, opened 2017-01-03
Destroyed HU causes problem with shipment creation


metas-ts:
Discovered by @metas-dh in today's e2e

The problem is that when checking for not-yet-delivered `M_ShipmentSchedule_QtyPicked` records, the system does not check if they reference actually destroyed HUs. It would need to discount shuch records

metas-dh:
Integrated, built and deployed for integration test

metas-dh:
Results of IT 1 

https://github.com/metasfresh/metasfresh-documentation/blob/master/_tests_collection/Testcase_%23759.md: OK

PR #775, opened 2017-01-05
FRESH-1091 #759: take into account the HU status

metas-ts:
ignore `M_ShipmentSchedule_QtyPicked` if they reference
destroyed/inactive HUs

FRESH-1091 #759 Destroyed HU causes problem with shipment creation


metas-ts:
I already merged, still please review&comment when you have time

PR #876, opened 2017-01-27
Fresh 1176

metas-rc:


de.metas.handlingunits.client/src/test/java/de/metas/handlingunits/client/terminal/editor/view/HUKeysByBarcodeCollectorTest.java
```diff
@@ -146,4 +195,251 @@ public void matcher_NotMatched_NullHUBarcode()
 
 		assertBarcodeMatches(false, barcodeMatcher, hu);
 	}
+	
+	@Test
+	public void matcher_StringAttribute()
+	{
+		
+	    //barcode matches a string attribute
```
metas-ts: Please add such informations to the javadoc so that's it's more visible

Issue #863, opened 2017-01-24
Process confirmation window needs to be changed


metas-dh:
Follow up for #835 Switch off Process Confirmation Window

* switching off allowing the process rerun was not effective: process confirmation window is still displayed in e.g. creating shipments, orders and invoicing (without rerun option)
* changes in code necessary

metas-lc:
IT. OK: https://github.com/metasfresh/metasfresh-documentation/blob/master/_tests_collection/Testcase_%23863.md

PR #885, opened 2017-01-30
FRESH-1221 #863 No Result Window for Prosesses that don't allow rerun

metas-rc:


de.metas.adempiere.adempiere/client/src/main/java/de/metas/process/ui/ProcessPanel.java
```diff
@@ -605,8 +605,15 @@ public void unlockUI(final ProcessInfo pi)
 			{
 				closeWindow = true;
 			}
+			
+			// #863
+			// Do not show result window for any process that doesn't allow rerun
+			if(!_allowProcessReRun)
```
metas-ts: Hi @metas-rc , with this change, does the system still respect the `setShowProcessLogs()` setting we talked about?

metas-rc: Hi. Yes, it is respected. I tested for invoice candidates that cannot be invoiced ans shipmentschedules that are already shipped. The error messages are still shown in the popup, without the misleading question about restarting.

Issue #813, opened 2017-01-16
hide packing instruction and qty when null


metas-lc:

metas-dh:
Results of IT1:

https://github.com/metasfresh/metasfresh-documentation/blob/master/_tests_collection/Testcase_%23813.md: OK

* we need to decide what to do about "No Handling Unit"; atm this can be selected as PI in order, and gets displayed in the jasper, although it's actually no PI at all: => see https://github.com/metasfresh/metasfresh/issues/960 about No Handling Unit

Issue #903, opened 2017-02-01
Jenkins build error with slash in branch name


metas-ts:

metas-ts:
the fix for this issue is in PR #905

PR #905, opened 2017-02-01
#813 hide packing instruction and qty when null

metas-ts:
FRESH-1161 #813 hide packing instruction and qty when null


Jenkinsfile
```diff
@@ -274,14 +274,17 @@ echo "Setting BUILD_VERSION_PREFIX=$BUILD_VERSION_PREFIX"
 final BUILD_VERSION=BUILD_VERSION_PREFIX + "." + env.BUILD_NUMBER;
 echo "Setting BUILD_VERSION=$BUILD_VERSION"
 
-// metasfresh-task-repo is a constrant (does not depent or the task/branch name) so that maven can find the credentials in our provided settings.xml file
+// metasfresh-task-repo is a constant (does not depend or the task/branch name) so that maven can find the credentials in our provided settings.xml file
```
metas-ts: Note that these changes relate to issue #903. They seemed necessary in the former branch name `pr/811` or something..

Issue #914, opened 2017-02-02
Customs Report Weight adjustment


metas-mk:
Customs Report Weight adjustment
- shall not be gross nor net. shall be gross - weight of palette (25kg)

metas-ts:
cherry-picked into the `stable` branch

metas-ts:
integrated changeset 1b2e464 and also cherry-picked it to branch `release-2017-06`

metas-dh:
Results of IT 2

https://github.com/metasfresh/metasfresh-documentation/blob/master/_tests_collection/Testcase_%23914.md: OK

   * problems found earlier this week (qual. discount, >1 LU in inout) all fixed

PR #916, opened 2017-02-02
#914 adjust weight in Docs_Purchase_InOut_Customs_Details function

metas-lc:
FRESH-1289  #914 Customs Report Weight adjustment
update Docs_Purchase_InOut_Customs_Details function, changed qty


de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/Docs_Purchase_InOut_Customs_Details.sql
```diff
@@ -63,6 +64,17 @@ FROM
 			LEFT OUTER JOIN M_HU_Attribute thuwn ON thu.M_HU_ID = thuwn.M_HU_ID
 				AND thuwn.M_Attribute_ID = ((SELECT M_Attribute_ID FROM M_Attribute WHERE value = 'WeightGross'))
 				AND thuwn.isActive = 'Y'
+			--need LU to calculate the weight
```
metas-ts: As i said to @metas-mk yesterday, it should also be possible to get the LU palette's tara 
via selecting the TU level HUs and with their weightTare values..
now i see that it might also be easier (for the DB at least)

but that's just a side note

de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/Docs_Purchase_InOut_Customs_Details.sql
```diff
@@ -63,6 +64,17 @@ FROM
...
+			LEFT OUTER JOIN M_HU_Assignment lhuas ON iol.M_InOutLine_ID = lhuas.Record_ID AND lhuas.isActive = 'Y' 
+								AND lhuas.AD_Table_ID = ((SELECT Get_Table_ID('M_InOutLine'))) and lhuas.M_HU_Assignment_id != thuas.M_HU_Assignment_id
+			LEFT OUTER JOIN M_HU lhu ON lhuas.M_LU_HU_ID = lhu.M_HU_ID
+			LEFT OUTER JOIN M_HU_PI_Version luv ON lhu.M_HU_PI_Version_ID = luv.M_HU_PI_Version_ID
+			LEFT OUTER JOIN M_HU_PI_Item lhupi ON luv.M_HU_PI_Version_ID = lhupi.M_HU_PI_Version_ID and lhupi.itemtype='PM'
```
metas-ts: i think, here (not 100% sure about other places), we *need* to check for `isactive='Y'` because an M_HU_PI_Item coudl jsut be there, but inactive and then it's product's weight would be counted.

de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/Docs_Purchase_InOut_Customs_Details.sql
```diff
@@ -63,6 +64,17 @@ FROM
 			LEFT OUTER JOIN M_HU_Attribute thuwn ON thu.M_HU_ID = thuwn.M_HU_ID
 				AND thuwn.M_Attribute_ID = ((SELECT M_Attribute_ID FROM M_Attribute WHERE value = 'WeightGross'))
 				AND thuwn.isActive = 'Y'
+			--need LU to calculate the weight
+			LEFT OUTER JOIN M_HU_Assignment lhuas ON iol.M_InOutLine_ID = lhuas.Record_ID AND lhuas.isActive = 'Y' 
+								AND lhuas.AD_Table_ID = ((SELECT Get_Table_ID('M_InOutLine'))) and lhuas.M_HU_Assignment_id != thuas.M_HU_Assignment_id
+			LEFT OUTER JOIN M_HU lhu ON lhuas.M_LU_HU_ID = lhu.M_HU_ID
```
metas-ts: maybe rather than going with HU-PI, why not go 
M_HU => M_HU_Item => M_HU_PI_Item
but that's also just a brainfart tbh

Issue #913, opened 2017-02-02
Include branch name in build version string


metas-ts:
here the branch is missing: 

![image](https://cloud.githubusercontent.com/assets/1264789/22544530/69ad31d8-e935-11e6-9d2e-20a3fc32d288.png)

which can be a problem when rolling out from branches that only have a few builds each

metas-dh:
Results of IT 1 

* branch name included as well, e.g. 4.55.54-master-1.589: OK

PR #918, opened 2017-02-03
#913 include branch name in build version string

metas-ts:
also provide a fallback for local builds


metas-ts:
Works now: 
image

PR #922, opened 2017-02-03
Fresh 1234

metas-rc:


de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandBL.java
```diff
@@ -1292,6 +1299,17 @@ public void handleReversalForInvoice(final org.compiere.model.I_C_Invoice invoic
 				final I_C_Invoice_Candidate invoiceCandidate = ilaToReverse.getC_Invoice_Candidate();
 				invoiceCandidate.setProcessed_Override(null); // reset processed_override, because now that the invoice was reversed, the users might want to do something new with the IC.
 
+				// #870
+				// Make sure that, when an invoice is reversed, the QtyToInvoice_Override and PriceEntered_Override are set back in the invoice candidate based on the values in the allocations
+				{
+					final int invoiceCandidateId = invoiceCandidate.getC_Invoice_Candidate_ID();
+					final BigDecimal qtyToInvoice_Override = ilaToReverse.getQtyToInvoice_Override();
+					final BigDecimal priceEntered_Override = ilaToReverse.getPriceEntered_Override();
+					Services.get(ITrxManager.class)
```
metas-ts: hi, what made you use the trx manager to make the change after commit? wouldn't it be simple to just update and save the `invoiceCandidate` isntance you already have here?

metas-rc: Hi. 
The invoice candidate **only** updates during the update process.
Together with Teo we implemented this as a workaround until we have the invoice candidate updating process with an event table (we discussed about it, didn't create a task yet)

Issue #460, opened 2016-10-10
Reference to HU Attributes instead of identical copies


metas-mk:
Reference to HU Attributes instead of identical copies.
Idea here is even to abstain from HU Attributes on Virtual PI, TU Level already has pure content (same product, same attributes, same packing instructions).
This would direclt fit to the new representation logic for the user and will reduce the amount of data transactions and records up to hundreds per LU. This should be a great performance gain.

metas-ts:
Inserting the concept written by @teosarca here:

# Draft concept (version 2)

The idea to optimize the number of records created when we receive identical HUs is to keep the information as compressed as possible.
(i.e. don't create 10000TUs when they are not needed because under a certain point, user don't care about that information or is not relevant or is not needed).

To better understand the idea, let's take following example.
We are receving 2LUs of 40TUs each. An TU will contain 10Kg of Tomatoes. We consider having 12 attributes.

In the example below there are 2 parts: the "compresed part" and the "exploded items".
Initially we will create one LU with the "compressed part" and having the "exploded items" part missing because we are not exploding now.
We will exploded only when needed and if needed.

Now take a look at the "exploded part". That's basically how our LUs are looking now, before this change.
The "exploded part" can cohexist with the "compressed part".
If we will explode something from compressed part, we will subtract the Qty and the packing materials from our compressed special VHU and we will create new LU_Item node(s) with the exploded TUs.


```
LU
	LU_Attribute x 12
	
	//
	// Compressed part (the bag)
	//
	
	LU_Item
		VHU(Special VHU!)
			VHU_Attribute x 0 <-shall not be needed
			VHU_Item (Material)
				VHU_Item_Storage: Product=Tomatoes, Qty=2x40x10Kg
			VHU_Item(Packing): Qty=2, M_HU_PI_Item.M_HU_PackingMaterial_ID=Palet
			VHU_Item(Packing): Qty=2x40, M_HU_PI_Item.M_HU_PackingMaterial_ID=IFCO

	//
	// Exploded items: 40 TUs (NOTE: we can explode maximum, one LU, that's why we don't have here 80 TUs)
	//
	
	LU_Item <-included TUs
		TU x 40
			TU_Attribute x 12
			TU_Item
				VHU
					VHU_Attribute x 12
					VHU_Item
						VHU_Item_Storage: Product=Tomatoes, Qty=10Kg
[...]

PR #932, opened 2017-02-06
#460 Reference to HU Attributes instead of identical copies

metas-ts:
Bugfix: the system didn't create an receipts anymore


metas-ts:
also cherry-picked this fix to branch `release-2016-06`

Issue #797, opened 2017-01-11
Zoom does not open new document


metas-rc:
Steps to reproduce:
* Go to a completed sales invoice;
* Create a Credit memo or an adjustment charge (from the button)
* => NOK: An empty invoice window is displayed in the end, instead of the newly created document

Note: The document was correctly created. It should have been displayed as specified in org.adempiere.invoice.process.CreateCreditMemoFromInvoice.postProcess(boolean) and org.adempiere.invoice.process.CreateAdjustmentChargeFromInvoice.postProcess(boolean).

IMO debugging should be start around org.compiere.apps.AEnv.zoom(int, int).

metas-lc:
IT. OK: https://github.com/metasfresh/metasfresh-documentation/blob/master/_tests_collection/Testcase_%23797.md

PR #935, opened 2017-02-08
Gh797 metasfresh

metas-rc:
FRESH-1136 #797

### DIFF
--- a/de.metas.adempiere.adempiere/client/src/main/java-legacy/org/compiere/apps/AEnv.java
+++ b/de.metas.adempiere.adempiere/client/src/main/java-legacy/org/compiere/apps/AEnv.java
@@ -650,6 +650,7 @@ public final class AEnv
 
 		final MQuery query = zoomInfo.createZoomQuery();
 
+		// task #797 Make sure the windows are displayed in the same thread as the process
 		SwingUtilities.invokeLater(() -> {
 			final AWindow frame = new AWindow();
 			if (!frame.initWindow(windowIdToUse, query))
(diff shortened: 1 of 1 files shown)


de.metas.adempiere.adempiere/client/src/main/java-legacy/org/compiere/apps/AEnv.java
```diff
@@ -633,30 +633,33 @@ public static void zoom(
...
-		
-		final AWindow frame = new AWindow();
-		if (!frame.initWindow(windowIdToUse, query))
-		{
-			return;
-		}
-		addToWindowManager(frame);
-		showCenterScreenOrMaximized(frame);
+
+		// task #797 Make sure the windows are displayed in the same thread as the process
```
metas-ts: hi, i don't fully understand the comment..from the invokelater-javadoc i understand that that with this method, we actually make sure the windows are displayed in the *AWT event dispatching thread*, even although this zoom invokaction might or might be coming from another thread..
am i getting it right?

metas-rc: Hello.
Right. As I said yesterday, I don't understand what thread interferes with the process but we found out while debugging that a field had a certain value in an if body that was defined on that field to have a different value. So we realized it's from another thread. But I don't know why. Teo suggested to try this and see if it works. It did. 
Please, fix the comment to be more easy to understand. I didn't know how to write it right.

PR #940, opened 2017-02-08
Fresh 1234

metas-rc:


de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandBL.java
```diff
@@ -1368,10 +1368,27 @@ else if (creditMemo && !creditedInvoiceReinvoicable)
 	private static void setQtyAndPriceOverride(final int invoiceCandidateId, final BigDecimal qtyToInvoiceOverride, final BigDecimal priceEnteredOverride)
 	{
 		final I_C_Invoice_Candidate invoiceCandidate = InterfaceWrapperHelper.create(Env.getCtx(), invoiceCandidateId, I_C_Invoice_Candidate.class, ITrx.TRXNAME_ThreadInherited);
-		invoiceCandidate.setQtyToInvoice_Override(qtyToInvoiceOverride);
-		invoiceCandidate.setPriceEntered_Override(priceEnteredOverride);
+
```
metas-ts: hi  @metas-rc 
i'm wondering: instead of having those ifs&stuff here, why not making the `qtyToInvoiceOverride` and `priceEnteredOverride` params `null` (i mean, in line 1307f) if they respective column value is null?

that can be done using `InterfaceWrapperHelper.getValueOrNull(model, columnName)`

metas-ts:
..still trying to figure out why the jenkins builds failed..hope to fix it soon

Issue #489, opened 2016-10-12
Implement DLM within single logical tables


metas-ts:
This issue is about implementing issue metasfresh/metasfresh-documentation#35

metas-lc:
note for IT: see if something pops-up during e2e

metas-dh:
tested during e2e, swing sp80:

* problems related to DLM were solved by #1076
* we will create further new issues and follow ups for future DLM-related problems

PR #950, opened 2017-02-12
#489 bunch of fixes

metas-ts:


metas-ts:
Fixes include
* `MColumn`: never add `DEFAULT NULL` to a new column's DDL; it seems to make postgresql update each row (tested in pgAdmin, checking DDL processing times)
* `MColumn`: put create-column DDLs as parameters into `SELECT public.db_alter_table(...)` calls to make sure that also DLM-views get the new column.
* code that creates `M_Transactions`: don't store `M_AttributeSetInstance_ID`s in the trx, unless they are "real"; by not-real, I mean ASI-IDs that don't have real ASIs but were stored in `M_Storage`s only for the sake of FIFO/LIFO.
* `C_Queue_WorkPackage`: MI to allow deleting workpackages (deletes elements etc)
* DLM
 * register a "watchdoc" to and allows it to stop a partitioner service's run if e.g. sales and purchase records are added together
 * make DLMConnectionCustomizer simpler to use
 * a lot of SQL fixes
 * declaring a lot of reference-columns as "partition boundaries"
* javadoc and cleanup

PR #953, opened 2017-02-13
Release 2017 06+patch4

homebeaver:
Hi @metas-ts ,
this is an implementation for #695 

In my opinion there is no need for any interceptor, because the jars are not loaded at startup.
I tested it with the simple helloworld example from 
https://forum.metasfresh.org/t/schnittstelle-zu-ms-com-excel-outlook-via-groovy/98/10
and with an excel ActiveXObject : 

just copy the needed jar in userlib, the dll in userllib-x64


homebeaver:
> no solution for the RUN_metasfresh.sh

I cannot test a *nix client impl in my env.

> Question: why did you decide to put the additional lib folders into the swing-client's folder?

Each time I downloaded a new client I have to create the log folder. 

> provide the hello-world-example ... jar

Yes. But github does not accept jars! I'll pack the jar + src in a zip

> readme with reference to #695 

OK

metas-ts:
to be continued in pull request #962

PR #962, opened 2017-02-14
Master+patch5

homebeaver:
cherry-picked and with reference to #695

You'll find a hello-world-example (jar + src) in a zip in #695


metas-ts:
cleaning up this PR, it became obsolete with https://github.com/metasfresh/metasfresh-dist/pull/1

Issue #1206, opened 2017-03-27
Remove Gebinderückgabe from Shipment Note for mf15 endcustomer


metasnw:
### Is this a bug or feature request?
feature request

### What is the current behavior?
Gebinderückgabe is always displayed on shipment note 

#### Which are the steps to reproduce?
1. Create a shipment
2. open print preview

### What is the expected or desired behavior?
remove that text from jasper only on mf15 endcustomer

metas-dh:
Results of IT1
tested on mf15, Swing & WebUI:

* if no packing item is used in shipment (e.g. only VPI), no "Gebinde Retouren" displayed in the jasper anymore: OK
* if packing item was used, "Gebinde Retouren" section is still included: OK
* regression: only on mf, always displayed for other customers: OK

PR #1213, opened 2017-03-28
Gh1206

metas-lc:
https://github.com/metasfresh/metasfresh/issues/1206


de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/Docs_Sales_InOut_Root.sql
```diff
@@ -32,7 +33,20 @@ $$SELECT
...
 	CASE WHEN io.DocStatus IN ('DR', 'IP') THEN 'de_CH' ELSE $2 END AS AD_Language,
 	isHidePackingMaterialInShipmentPrint,
-	mb.email
+	mb.email,
+	CASE
+		WHEN
+		EXISTS(
+			SELECT 0
+			FROM M_InOutLine iol1
+			INNER JOIN M_Product p ON iol1.M_Product_ID = p.M_Product_ID AND p.isActive = 'Y'
```
metas-ts: why do you only join the product if it is active? what's the benenig of that?
unfortunately i already integrated the other repo, so I'm now removing it. please contact me asap, so we might add it once again

Issue #1362, opened 2017-04-22
Activate auto numbering for incoming and outgoing payments


metasnw:
### Is this a bug or feature request?
bug

### What is the current behavior?
payment document No's are not automatically numbered in web UI

metasnw:
CAUTION: this one is not an easy as it changes existing client doctype payment. We need to see what will be the impact on existing systems. May be we find a way to only apply to new systems?

metasnw:
@metas-ts
Were you able to create the dynamic script?

metas-ts:
@metasnw yes, it's here: https://github.com/metasfresh/metasfresh/issues/1362#ref-commit-0a9fb08 and integrated together with your changes

metas-rc:
Results of IT

* Tested on webui


* I manually created paymenst 9 incoming and outgoing)  and their document numbers were filled correctly, increasing with 1, different sequences for each payment doctype.

metas-rc:
The document number being readonly will be dealt with in https://github.com/metasfresh/metasfresh/issues/1860

PR #1363, opened 2017-04-22
#1362 Activate auto numbering for incoming and outgoing payments

metasnw:
created two new sequences for incoming and outgoing payment starting
with 1 and 2 Million. Changing EXISTING doctype incoming and outgoing
payment to be number controlled and use that two new sequences
https://github.com/metasfresh/metasfresh/issues/1362


metasnw:
@metas-ts yes, that is true.
Mark suggest to put counter scripts into the repo of projects were we do not want that.

metas-ts:
Hi @metasnw ,
> Mark suggested to put counter scripts into the repo of projects where we do not want that.

i'm not sure...what about existing users that do not have dedicated repos? re84 comes to mind, but maybe others too?

maybe we should better just update the master-DB with your script, and in addition provide the script to those clients that actually want or need it or where we made sure it's safe?

metasnw:
@metas-ts I think the risk is higher that people ran into problems because its not working in webUI than to change doc numbers as payments are not external documents.

Issue #1507, opened 2017-05-11
Add the possibility to close or not invoice candidates when canceling a contract


cristinamghita:
### Is this a bug or feature request?
Feature request
### What is the current behavior?
The invoice candidate will not be invoiced when contract is terminated.
### What is the expected or desired behavior?
We want to allow to cancel a contract, but to not close invoice candidate

cristinamghita:
Test:

* close a contract that has invoice candidates
*check that the candidates were closed be default

metas-rc:
Results of IT

Tested on swing

Note: The handler that closes invoice candidates in not active in mf15 to the candodates were never closed.
Furthermore, the shipment schedules for orderlines that belong to a running contract are not created ( this is an old feature)

Regression: Ending contracts works correctly.

PR #1530, opened 2017-05-15
#1507 Add the possibility to close or not invoice candidates 

cristinamghita:
#1507


metas-ts:
hi @cristinamghita 
it looks like the changes in here are a subset of the changes in PR #1536 ..right?

Issue #1506, opened 2017-05-11
Add the possibility to track easily contracts that ended naturally


cristinamghita:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
You can't see easilly when a 3 month contract ended
### What is the expected or desired behavior?
Add in FT, the option for contract status called EndingContract.

cristinamghita:
Test:
* end a contract
* all shall be as before

metas-rc:
Results of IT

Tested on swing
Regression: Created a contract and canceled it. ll worked fine

PR #1536, opened 2017-05-15
#1506 Add the possibility to track easily contracts that ended naturally

cristinamghita:
#1506


de.metas.contracts/src/main/java/de/metas/flatrate/api/IContractChangeBL.java
```diff
@@ -50,7 +50,15 @@
...
 	 *            for the cancel conditions.
+	 * @param isCloseInvoiceCandidate this value is forwarded to the given term's <code>IsCloseInvoiceCandidate</code> column and will determine what to do with invoice candidates for the term which were not
+	 *            yet (fully) invoiced. See {@link de.metas.flatrate.invoicecandidate.spi.impl.FlatrateTermHandler}
 	 * @throws SubscriptionChangeException if <code>changeDate</code> is before the term's EndDate and if there is no {@link de.metas.flatrate.model.I_C_Contract_Change} record for that date
 	 */
-	void cancelContract(I_C_Flatrate_Term term, Timestamp changeDate);
+	void cancelContract(I_C_Flatrate_Term term, Timestamp changeDate,  boolean isCloseInvoiceCandidate);
+	
+	/**
+	 * ending naturally a contract
```
metas-ts: what does "ending" a contract mean?

de.metas.contracts/src/main/java/de/metas/flatrate/api/IFlatrateHandler.java
```diff
@@ -49,4 +49,20 @@
 	 * @task https://github.com/metasfresh/metasfresh/issues/549
 	 */
 	void afterExtendFlatrateTermCreated(final I_C_Flatrate_Term oldTerm, final I_C_Flatrate_Term newTerm);
+	
+	
+	/**
+	 * Invoked by {@link IFlatrateBL#de.metas.flatrate.api.impl.FlatrateBL.extendContract0(I_C_Flatrate_Term, boolean, boolean, I_C_OrderLine, String)} after the given <code>term</code> was checked.
```
metas-ts: what does "checked" (from javadoc) and "ended" from the method name mean?

de.metas.contracts/src/main/java/de/metas/flatrate/api/impl/DefaultFlatrateHandler.java
```diff
@@ -98,4 +98,16 @@ public void afterExtendFlatrateTermCreated(I_C_Flatrate_Term oldTerm, I_C_Flatra
 	{
 		// nothing
 	}
+	
+	@Override
```
metas-ts: i suggest to add the "does nothing" javadoc where to (and to beforeFlatrateTermCreated())

de.metas.contracts/src/main/java/de/metas/flatrate/invoicecandidate/spi/impl/FlatrateTermHandler.java
```diff
@@ -309,7 +309,7 @@ public void setOrderedData(final I_C_Invoice_Candidate ic)
 
 		ic.setDateOrdered(getDateOrdered(term));
 
-		if (X_C_Flatrate_Term.CONTRACTSTATUS_Gekuendigt.equals(term.getContractStatus()))
+		if (X_C_Flatrate_Term.CONTRACTSTATUS_Gekuendigt.equals(term.getContractStatus()) && term.isCloseInvoiceCandidate()) // terminate invoice candidates only when the flag is set
```
metas-ts: i think in this case it would be much better to call `IInvoiceCandBL.closeInvoiceCandidate()`

de.metas.contracts/src/main/java/de/metas/flatrate/process/C_Flatrate_Term_Change.java
```diff
@@ -66,7 +66,7 @@ protected String doIt()
 		
 		final I_C_Flatrate_Term currentTerm = InterfaceWrapperHelper.create(getCtx(), getRecord_ID(), I_C_Flatrate_Term.class, get_TrxName());
 
-		Services.get(IContractChangeBL.class).cancelContract(currentTerm, changeDate);
+		Services.get(IContractChangeBL.class).cancelContract(currentTerm, changeDate, true);
```
metas-ts: maybe we should here have a parameter so the user can decide if remaining ICs shall be closed or not?

de.metas.contracts/src/main/java/de/metas/flatrate/api/IContractChangeBL.java
cristinamghita: When we have a limited contract, for example, a 3 months contract, when those 3 months expire, we want that the contract to have a status that is saying that: that the contract ended naturally.

de.metas.contracts/src/main/java/de/metas/flatrate/api/IFlatrateHandler.java
cristinamghita: Checked refers to the fact that the flatrate term is not autorenew. So was checked for that.
Ended form the method, is referring to the fact that the contract has expired, and if it does need to be extended, then the method is applying some code, after the contract has ended,

de.metas.contracts/src/main/java/de/metas/flatrate/api/impl/DefaultFlatrateHandler.java
cristinamghita: OK. Agree.

metas-ts:
hi @cristinamghita ,
i think i just understood that the chakges in here are a superset of PR #1530 ..is that correct?
i got a bit lost in the SQL..how is "tracking contracts that ended naturally" achieved? and as written in the review, i'm not sure what §ended" naturally actually means..

de.metas.contracts/src/main/java/de/metas/flatrate/invoicecandidate/spi/impl/FlatrateTermHandler.java
cristinamghita: Hm.. OK. I asked previously about this flag, if we still need it.
I will use that.

de.metas.contracts/src/main/java/de/metas/flatrate/process/C_Flatrate_Term_Change.java
cristinamghita: I just wanted to preserve the old behaviour. By default is to close invoice candidates.

metas-ts: ok, let's add the parameter when someone needs it

Issue #1636, opened 2017-05-24
Translate Actions in Production Issue/Receive Window


metasnw:
### Is this a bug or feature request?
feature

### What is the current behavior?
All actions are English 

#### Which are the steps to reproduce?
1. set language DE
1. open window production order
1. see actions

### What is the expected or desired behavior?
all german if german is set to DE

PR #1637, opened 2017-05-24
#1636 Translate Actions in Production Issue/Receive Window to de_DE

metasnw:
changing processes and a few messages
https://github.com/metasfresh/metasfresh/issues/1636


metasnw:
@metas-mk I am not sure if you like the german terms I gave so please review.
I tried to keep em drop dead simple

metas-mk:
@metasnw 
Thanks for the PR. I've done the review. I like the translations. One minor thing. "Issue/ Receipt" and "Issue/ Receipt (barcode)" is translated w/ "Produzieren ..." (1st char uppercase). "Issue HU" and "Receive HU" are translated w/ 1st chat lowercase translations (hinzufügen, empfangen). I would suggest to also have these uppercase.

Issue #1669, opened 2017-05-27
Translating Manufacturing Order Field and List Values of Planning Status to de_DE


metasnw:
### Is this a bug or feature request?
feature

### What is the current behavior?
field and value is all English
![image](https://cloud.githubusercontent.com/assets/16079655/26520765/01290e86-42d9-11e7-835f-8b85cd267e27.png)

#### Which are the steps to reproduce?
see above

### What is the expected or desired behavior?
translated

PR #1670, opened 2017-05-27
#1669 Translating Manufacturing Order Field and List Values of Planning Status to de_DE

metasnw:
https://github.com/metasfresh/metasfresh/issues/1669


metas-mk:
@metasnw 
The review is done. Nice Translations. Thanks.

PR #1686, opened 2017-05-29
#1306 Avoid returning the same Hu multiple times

metas-rc:
https://github.com/metasfresh/metasfresh/issues/1306
https://metasfresh.atlassian.net/browse/FRESH-1936


de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/inout/impl/CustomerReturnsInOutProducer.java
```diff
@@ -82,7 +83,7 @@ public static final CustomerReturnsInOutProducer newInstance()
 	/**
 	 * List of handling units that have to be returned to vendor
 	 */
-	private final List<HUToReturn> _husToReturn = new ArrayList<>();
+	private final Set<HUToReturn> _husToReturn = new HashSet<>();
```
metas-ts: i'm always a bit nervous about doing `hashCode()` and `equals()`with our model classes, such as  `I_M_HU´.
Are you sure it will work as expected. If in doubt, please also consult with @teosarca.
If you can't make sure that it works as expected, please work with e.g. IDs instead to rule out duplciated HUs

de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/inout/impl/CustomerReturnsInOutProducer.java
```diff
@@ -82,7 +84,8 @@ public static final CustomerReturnsInOutProducer newInstance()
 	/**
 	 * List of handling units that have to be returned to vendor
 	 */
-	private final List<HUToReturn> _husToReturn = new ArrayList<>();
+	private final Set<HUToReturn> _husToReturn = new TreeSet<>(Comparator.comparing(HUToReturn::getM_HU_ID)
```
metas-ts: very nice :-) :+1:

Issue #1712, opened 2017-06-01
Wrong Handling Unit Assignment in Shipment Schedule via aggregated Picking


metas-mk:
### Is this a bug or feature request?
Bug
### What is the current behavior?
2 Sales Orders with same Business Partner and Location and Produkt, lead to the wrong result in Handling Unit Assignments via Picking Terminal. They are picked together in one sum and seem to be wrongly distributed onto the shipment schedule lines.
#### Which are the steps to reproduce?
Try and see.
### What is the expected or desired behavior?
Distribute the Handling Units and Assignments onto the correct Shipment Schedules and Shipments.

metas-ts:
Cherry-Picked to the release branch, see https://jenkins.metasfresh.com/job/ops/job/cherry-pick-issue-script/155/

metas-lc:
IT
swing
1. org2 having a config to not aggregate products on picking. org1 default config
2. login with org2, create 2 orders, same bpartner, location, product. You can add different qtys
3. go to picking slot 
=> there are 2 rows for your product
4. select them and do picking using hu editor, with already existing TUs
=> 2 inouts created, hu assignments set ok, but i didn't get lines for TU in inout.. ok?

5. repeated the same testcase for org1. 
=> got 1 line in picking terminal, 2 inouts, both containing TUs and LUs, correct qty and assignments

PR #1736, opened 2017-06-06
#1712 Wrong Handling Unit Assignment in Shipment Schedule via aggregated Picking

metas-ts:
Wrong Handling Unit Assignment in Shipment Schedule via aggregated Picking #1712


metas-ts:
I think i reproduced the problem in a unit test as follows:
* create two TUs, one with 11 CUs in it then other one with 10 CUs in it
* create two shipment schedules, one with QtyToDeliver=11, one with QtyToDeliver=10
* give it all to the allocator in the order such that when iterating the schedules and HUs, the 10-schedule hits the 11-TU

Result:
* from the 11-TU, 10 are allocated to the 10-schedule.
* from the 11-TU, 1 is allocated to the 11-schedule.
* from the 10-TU, 10 are allocated to the 11-schedule.

Of course we could solve *this* problem by adding code to look for "exact" qty matches and allign the HUs' order accordingly...but that's not a general solution.

* We could also look into transforming the given HUs to match the shipment schedules
* Or we could stop allowing multiple scheds to be aggregated into one packing item

PR #1750, opened 2017-06-07
#1742 Translate to en_US

metasnw:
https://github.com/metasfresh/metasfresh/issues/1742


metasnw:
@metas-mk I translated also ad_menu for those two windows but they remain untranslated even after cache reset and server restart. Do you have an idea?
image

metas-mk:
@metasnw 
> I translated also ad_menu for those two windows but they remain untranslated even after cache reset and server restart. Do you have an idea?

please check if "browse name" is translated in menu already

PR #1826, opened 2017-06-15
#1825 Disable Quickstart, Admin and System Administrator Roles for WebUI

metasnw:
#1825


metasnw:
@metas-mk please review if that is ok with you.

PR #1922, opened 2017-06-29
Gh1876

metas-lc:
https://github.com/metasfresh/metasfresh/issues/1876


metas-ts:
Hi @metas-lc the jasper files can't be compiled on jenkins. some classes are missing. if you can locate the jar, i guess i can all it to the maven depdendencies..
or we switch back to the servlet..

teosarca:
@metas-ts @metas-lc  I think i've found it:
image

so it shall be:
```
			<groupId>net.sf.barcode4j</groupId>
			<artifactId>barcode4j</artifactId>
			<version>2.1</version>
```

Issue #2054, opened 2017-07-21
Window Design WebUI: Manufacturing Order


metas-rc:
* Belongs to https://github.com/metasfresh/metasfresh/issues/2045

### Is this a bug or feature request?
* FR
### What is the current behavior?
* The window Manufacturing Order does not respect our corner stones and layout conventions
#### Which are the steps to reproduce?
* Check and see
### What is the expected or desired behavior?
* Make the window Manufacturing Order respect our corner stones and layout conventions

metas-lc:
IT
Manufacturing order:
CORNER STONES FOR DOCUMENTS
TOP RIGHT:  Target document type, Document no, Doc date OK
written as Document Type, No. and Date OK
mandatory and important fields are on top left OK
org and client on bottom right OK

grid view: 
first 3: doc type, no date OK
last 2: docstatus, org NOK? it's only org

Included tabs: 
workflow, hu assignments, cost, auswertung and produkt attribute does not have white background NOK?

Filter:
docstatus is missing. NOK?

metas-lc:
created followup https://github.com/metasfresh/metasfresh/issues/2156
so i'll close this issue

PR #2072, opened 2017-07-26
Gh2054 Enhance Manufacturing Order webui window

metas-rc:
https://github.com/metasfresh/metasfresh/issues/2054


metas-ts:
i guess it doesn't make sense to try and integrate this PR after more than one month (might also well be that it's obsolete anyways)

Issue #2116, opened 2017-08-08
Use handler when matching partner in ESR import


cristinamghita:
### Is this a bug or feature request?
feature request
### What is the expected or desired behavior?
In method de.metas.payment.esr.dataimporter.ESRDataLoaderUtil.setValuesFromESRString(I_ESR_ImportLine, String), the matching of the partner should happen trough a handler, not directly, because in some cases, we might want different behaviour.
For the end user is the same thing.

metas-ts:
Hi @cristinamghita,
 why did you call it a bug?
"in some cases, we might want different behaviour" doesn't strike me as a malfunction ;-)

cristinamghita:
Yes, you are right! Feature request!

cristinamghita:
### Test

- Import an esr file with a invoice from a different org the then one that is in esr import => **Import fails. OK**

- Import an esr file with a bpartner from a different org the then one that is in esr import => **Import fails. OK**

cristinamghita:
#Results of IT 
## Tested on webui
### Test case: import an invoice form same org 
* create an invoice in org 001 using a partner from org 001
* Generate the file for that invoice using https://docs.google.com/spreadsheets/d/1W1fUBiDt5tm49AM3WNpmZIilOvi5LQY8f-tYkCAr84A/edit#gid=0
* Import the file
* Result **OK**
  * File imported 
  * line created
  * invoice paid
  * all processed

PR #2124, opened 2017-08-09
Improve handler

cristinamghita:
#2116 Use handler when matching partner in ESR import

### DIFF
--- a/de.metas.payment.esr/src/main/java/de/metas/payment/esr/dataimporter/ESRDataLoaderUtil.java
+++ b/de.metas.payment.esr/src/main/java/de/metas/payment/esr/dataimporter/ESRDataLoaderUtil.java
@@ -230,9 +230,6 @@ public class ESRDataLoaderUtil
 			}
 			else
 			{
-				addMatchErrorMsg(importLine,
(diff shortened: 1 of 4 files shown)


de.metas.payment.esr/src/main/java/de/metas/payment/esr/dataimporter/ESRDataLoaderUtil.java
```diff
@@ -218,23 +218,24 @@ private void setValuesFromESRString(
...
+			
+			// check the org: should not match with invoices from other orgs
+			if (match)
+			{
+				bPartnerId = bPartner.getC_BPartner_ID();
+				importLine.setC_BPartner(bPartner);
+				importLine.setESR_Document_Status(X_ESR_ImportLine.ESR_DOCUMENT_STATUS_PartiallyMatched);
+				
+			}
+			else
```
metas-ts: i saw that our DefaultESRLineHandler already adds its own error message, so i think, if the org's don't fit, the message will be default be added twice.
I *think* it make sense for the handler to add its message..so the question is why to keep the "else"..

cristinamghita: Hm.. you are right about the message, but in the message we need to make sure that the bp is not set.

Issue #1099, opened 2017-03-06
Fix Materialentnahme movement creation


metas-rc:
### Is this a bug or feature request?
* Bug
### What is the current behavior?
* Movement line for TU is not created

#### Which are the steps to reproduce?
* Create a purchase order for a product on a TU;
* Create a receipt for this (remember the HU value)
* Go to Hanling Unit POS, select the HU value 
* Click on Materialentnahme 
* Go to the created movement and then to its lines
* =>NOK: There are only lines for the LU and for the Product. No line for the TU 
### What is the expected or desired behavior?
* A line for TU should also be created.

metas-ts:
Note: the BL is in `InventoryHUEditorModel.doDirectMoveToWarehouse()`

metas-dh:
rolled out this morning

 Results of IT 1
tested on sp80 swing

https://github.com/metasfresh/metasfresh-documentation/blob/master/_tests_collection/Testcase_%231099.md: OK

PR #2125, opened 2017-08-10
#1099 adding message

metasnw:
for window.browseTree.caption


metasnw:
wrong repo, wrong sql

Issue #2112, opened 2017-08-07
Reversing a vendor-return-InOut needs to reverse the HU states


metas-ts:
### Is this a bug or feature request?

Bug

### What is the current behavior?

If you create a vendto return (`IHUInOutBL.createVendorReturnInOutForHUs()`) and then later reverse the `M_InOut`in question, the HUs of the reversed vendor return stay in the status "shipped".

Problems with that
* the HUs aren't there besides there is no completed document telling where they went
* the HUs cant be handeled (e.g. in the particular case, the customer wanted to do another vendor return)

#### Which are the steps to reproduce?

try it

### What is the expected or desired behavior?

If the `M_InOut` is reversed, the shipped HUs shall be "back"

metas-rc:
This task also covers reversing HU states on Customer return reverse.
Also the problem about NPE on completing shipment is solved.

metas-lc:
IT
I. sales (return):
1. create and complete a purchase order, warehouse X
=> HU is planning, wh X
2. make receipt to wh Y
=> hu is active, wh Y
3. create and complete a sales order, warehouse Y
4. go to picking terminal and pick the hu from step 2
=> hu is picked, wh Y
5. create a sales inout using the qty of the HU
=> hu is shipped, wh Y
6. go to shipment and create a customer return with that HU
=> HU is active, wh quality
7. go to customer return and cancel (stornieren) the inout
=> HU is shipped, wh Y
=> no movement was made. OK?
8. go to shipment again, and create another customer return with that HU
=>  HU is active, wh quality

9. repeat the same case with multiple HUs
=> works the same

NOK: you get too many product lines in customer return. Repeat the test above to see (first time i did it, happened on steps 6 and 8, and the second time, only on 8)

II. purchase (return):
1. create and complete a purchase order, wh X
=> HU is planning, wh X
2. make receipt to wh Y
=> hu is active, wh Y
3. create a vendor return for your receipt
=> hu is Shipped, wh Y
4. go to vendor return and cancel (stornieren) the inout
=> hu is Active, wh Y
5. create another vendor return
=> hu is Shipped, wh Y

III. sales (normal)
1. create and complete a purchase order, warehouse X
=> HU is planning, wh X
2. make receipt to wh Y
=> hu is active, wh Y
3. create and complete a sales order, warehouse Y
4. go to picking terminal and pick the hu from step 2
=> hu is picked, wh Y
5. create a sales inout using the qty of the HU
[...]

PR #2301, opened 2017-09-01
Gh2112 Enhancements in returns reversal and movements

metas-rc:
https://github.com/metasfresh/metasfresh/issues/2112

### DIFF
--- a/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/IHUPackingMaterialsCollector.java
+++ b/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/IHUPackingMaterialsCollector.java
@@ -155,4 +155,8 @@ public interface IHUPackingMaterialsCollector<T>
 	 */
 	IHUPackingMaterialsCollector<T> setProductIdSortComparator(Comparator<Integer> productIdsSortComparator);
 
+	void setisCollectTUNumberPerOrigin(boolean b);
+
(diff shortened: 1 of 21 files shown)


de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/IHUPackingMaterialsCollector.java
```diff
@@ -155,4 +155,8 @@ void addLU(I_M_HU luHU,
 	 */
 	IHUPackingMaterialsCollector<T> setProductIdSortComparator(Comparator<Integer> productIdsSortComparator);
 
+	void setisCollectTUNumberPerOrigin(boolean b);
```
metas-ts: what is this method about? also imho it shoul be named `setCollectTUNumberPerOrigin` (->javadoc)

de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/IHUPackingMaterialsCollector.java
```diff
@@ -155,4 +155,8 @@ void addLU(I_M_HU luHU,
...
+
+	void setisCollectAggregatedHUs(boolean b);
```
metas-ts: * same about method naming
* also about those two methods:
  * when does it make sense to call `setisCollectAggregatedHUs(false)` (->javadoc)
  * or maybe we don't need them at all?

Issue #2342, opened 2017-09-05
Make Material Disposal Lines and their Movements respect all the details from the HUs


metas-rc:
Followup for https://github.com/metasfresh/metasfresh/issues/2143

### Is this a bug or feature request?
* Bug
### What is the current behavior?
* Material disposals for split HUs are not correctly created. 
* The movements are only created for LUs, not the TUs
#### Which are the steps to reproduce?
* Go to Wareneingang POS, Select 3 lines from the same purchase order, split them to be on an LU each
* Go to Korrektur, select the HUs, do Material Disposal (Entsorgung)
* Go to the created inventory/inventories
* NOK: The M_HU_PI_Item_Product is not set
* NOK: Sometimes the qtyTU is 0
* Go to the created movements ( to Gebindelager)
* NOK: Lines only for LUs
### What is the expected or desired behavior?
* Movements shall have lines for LUs and TUs with the correct qtys
* Inventory lines must contain the correct qtyTU and M_HU_PI_Item_Product

metas-ts:
cherry-picked the 3 commits from PR https://github.com/metasfresh/metasfresh/pull/2382
into branch `2017-09-12`

metas-lc:
IT
1. swing:
bestellung with 3 products (qtys 50, 40, 50) -> wareneingang pos -> open all 3 products at the same time -> put each CU on TUs and LUs (5/1, 4/1, 5/1) -> select all, create receipt -> select the line again and press korrectur -> select all 3 LUs, select Dispose 
=> 
new Dispose docu created with correct qty TU and packvorschrift OK
1 movement created for LU and TUs with correct qtys OK

2. webui
bestellung with 3 products (qtys 50, 40, 50) -> receipt candidates -> open all 3 products at the same time (receive CUs) -> put each CU on TUs and LUs (5/1, 4/1, 5/1) -> select all, create receipt -> select the line again and press korrectur -> select all 3 LUs, select Dispose 
=> 
new Dispose docu created with correct qty TU and packvorschrift OK
1 movement created only for LU (TUs are missing) NOK
now you can't reverse the disposal NOK

metas-ts:
cherry-picked the commit from PR https://github.com/metasfresh/metasfresh/pull/2439 to `release` and `2017-09-12`

metas-lc:
IT
webui
[...]

PR #2382, opened 2017-09-07
Gh2342 1

metas-rc:
https://github.com/metasfresh/metasfresh/issues/2342

Make sure the qtyTU and M_HU_PI_Item_Product are correctly set in inventory lines


de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/allocation/impl/InventoryAllocationDestination.java
```diff
@@ -372,14 +384,55 @@ private I_M_InventoryLine getCreateInventoryLine(final I_M_InOutLine inOutLine,
 		return inventoryLine;
 	}
 
-	private BigDecimal countTUs(final IHUContext huContext, final I_M_HU hu, final I_M_InOutLine receiptLine)
+	private BigDecimal countTUs(final IHUContext huContext, final I_M_HU hu, final I_M_InventoryLine inventoryLine)
 	{
+		// Find out if the hu is VHU or Aggregate -> Move to TU level
+		// Add the TU recursively
```
metas-ts: i don't understand this comment...assume the `hu` is virtual: 
i understand that in this case its parent is the TU that you actually need. 
But what does "aggregate" mean in this context?

de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/allocation/impl/InventoryAllocationDestination.java
```diff
@@ -372,14 +384,55 @@ private I_M_InventoryLine getCreateInventoryLine(final I_M_InOutLine inOutLine,
...
+					.filter(i -> X_M_HU_PI_Item.ITEMTYPE_Material.equals(i.getItemType()))
+					.findFirst().orElse(null);
+
+			if (materialItem != null)
+			{
+				huPIP = huPiItemProductDAO.retrievePIMaterialItemProduct(materialItem, partner, product, date);
+			}
+		}
+
+		inventoryLine.setM_HU_PI_Item_Product(huPIP);
```
metas-ts: Also this method's name is `countTUs`, so why do we set the inventoryline's M_HU_PI_Item_Product ? seems to be a rather unexpected sideeffect..

Issue #529, opened 2016-10-26
Possibility to easily create more than 1 referenced Doc per Invoice


metas-mk:
Possibility to easily create more than 1 referenced Doc per Invoice
- Credit Memo
- Additional Invoice

metas-ts:
Idea: don't let the "source invoice" reference the addtitional invoices or credit memory anymore. zoom-into should suffice

metas-ts:
cherry-picked the fixes about manual credit memo creation to branch `2017-09-12`
* 5773667ee9aee52e0f43e4a176158891d3e1f313
* d52bbcdbbe955af385a2cf62101d32de68b91431

metas-lc:
IT
go to sales invoice and create a new one
go to actions and create some credit memos for this invoice
=> you can create multiple credit memos for one invoice OK
=> you can zoom in to all credit memos OK
=> ref_invoice_id is set only to credit memos and not to your invoice as well OK

PR #2384, opened 2017-09-07
#529 Support manual credit memo creation

metas-rc:
+ code cleanup
https://github.com/metasfresh/metasfresh/issues/529


de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoice/model/validator/C_Invoice.java
```diff
@@ -188,16 +188,18 @@ public void allocateInvoiceAgainstCreditMemo(final I_C_Invoice creditMemo)
 
 		final I_C_Invoice parentInvoice = InterfaceWrapperHelper.create(creditMemo.getRef_Invoice(), I_C_Invoice.class);
```
metas-ts: will `creditMemo.getRef_Invoice()` always return null if `creditMemo.getRef_Invoice_ID()` is null? i mean, with POWrapper, and POJOWrapper? Tbh I'm not 100% sure. If you aren't either, please check for `creditMemo.getRef_Invoice_ID() > 0` instead, before getting the parent invoice.

Otherwise, `parentInvoice` might not be `null` but an "empty" object without values inside.

metas-ts: done

PR #2388, opened 2017-09-07
Gh2112 edi problem for customer returns

metas-rc:
https://github.com/metasfresh/metasfresh/issues/2112


metas-ts:
i CPed the EDI changes into metasfresh/metasfresh#2389
the non-edi changes appear to be here by accident

Issue #2393, opened 2017-09-08
report property "summary" is defined twice in purchase order


metasnw:
### Is this a bug or feature request?
bug

### What is the current behavior?
files report_<lang>.properties contain 2x summary 
one time "summary" which is not used
one time "Summary" which is actually used

confusing anybody searching for summary and changing the value

#### Which are the steps to reproduce?
check here metasfresh\de.metas.fresh\de.metas.fresh.base\src\main\jasperreports\de\metas\docs\purchase\order\report_de_DE.properties

### What is the expected or desired behavior?

remove the wrong one

PR #2394, opened 2017-09-08
#2393 report property "summary" is defined twice in purchase order

metasnw:
removed the wrong one


metasnw:
@metas-lc can you take a look if that change is fine?

metas-lc:
@metasnw the change looks fine

PR #2464, opened 2017-09-13
#2457 Fix duplicating locations

cristinamghita:
#2457 Duplicate locations on partner import when multiple contacts


metas-ts:
Hi @cristinamghita ,
FYI, I did some refactorings with the goal of making this rather old-style code less "lenghty".

PR #2536, opened 2017-09-21
#2342 Create snapshots for HUs only when the inventory is completed.

metas-rc:
https://github.com/metasfresh/metasfresh/issues/2342


de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/allocation/impl/InventoryAllocationDestination.java
```diff
@@ -311,9 +311,15 @@ private I_M_InventoryLine getCreateInventoryLineInDispute(final I_M_InOutLine re
 
 	public List<I_M_Inventory> completeInventories()
```
metas-ts: I recomment you check out the clean code book's chapter on Functions..particularly page 35ff and see if you can do something about this method name and its content

de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/allocation/impl/InventoryAllocationDestination.java
```diff
@@ -417,8 +423,9 @@ private I_M_InventoryLine getCreateInventoryLine(
 		final ISnapshotProducer<I_M_HU> currentSnapshotProducer = huSnapshotProducerByInventoryId.get(inventoryLine.getM_Inventory_ID());
 		Check.errorIf(currentSnapshotProducer == null, "Missing SnapshotProducer for M_Inventory_ID={}", inventoryLine.getM_Inventory_ID());
 
+		// add each top level HU to be added to the snapshot of the inventory
 		currentSnapshotProducer.addModel(topLevelHU);
-		currentSnapshotProducer.createSnapshots();
+
```
metas-ts: hi, the currentSnapshotProducer.createSnapshots(); you moved was in the method naed `createHuSnapshot()`.
I arranged it that way in this commit: 5d8d73b4c1165b7d8f85b048641fd2dcc2988421
* please check out the problem i tried to address, and if it's still solve with this change (i guess "yes")
* please figure aout a more fitting name for this method (you can also or consider inlining it..whatever makes the code more readable)

Issue #2144, opened 2017-08-16
Action for Warehouse Movements


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, the user can only move HU via Action for Quality Issue Warehouses. It's not possible to move in other Warehouses.
#### Which are the steps to reproduce?
Check and see.
### What is the expected or desired behavior?
Create an identical action, but allow to move from and to all possible warehouses.

metas-lc:
IT
swing:
receipt pos -> korrectur -> select an HU and pick MoveHUToWarehouse
=> a list with whs appear and you can select one => movement created for HUs and product OK

handling units pos ->  select an HU and pick MoveHUToWarehouse
=> a list with whs appear and you can select one => movement created for HUs and product OK

probably NOK: when you do MoveHUToWarehouse you can also select the warehouse where you are currently in. And if you select that one, you get an error

webui:
hu editor

metas-ts:
Also cherry-picked this fix to `2017-09`

metas-lc:
IT
repeated the testcases above
=> all OK, no errors, the process was added in quickactions OK
but you can still choose the warehouse you are already in (Lagerort == Lagerort An and Merkmale == Attribute Set Instance To) NOK

metas-lc:
IT
I  can still choose the warehouse i am already in (Lagerort == Lagerort An and Merkmale == Attribute Set Instance To) NOK

metas-lc:
IT
repeated the testcase above -> works fine, the wh you are in doesn't appear anymore

metas-ts:
cherry-picked to `release`

PR #2537, opened 2017-09-21
Gh2144 Exclude HUs'warehouses when moving them to another warehouse

metas-rc:
https://github.com/metasfresh/metasfresh/issues/2144


de.metas.handlingunits.client/src/main/java/de/metas/handlingunits/client/terminal/editor/model/impl/MovementsAnyWarehouseModel.java
```diff
@@ -87,4 +85,40 @@ protected void createMovements()
 
 	}
 
+	/**
+	 * Get the warehouses existing for the hus' organization , excluding those which currently contain the HUs given as parameters
+	 * 
+	 * @param hus
+	 * @return
+	 */
+	public static List<org.compiere.model.I_M_Warehouse> getWarehousesUnlessOfHUs(final List<I_M_HU> hus)
```
metas-ts: teo and i had trouble understand this method from either javadoc or name. We suggest to rename it e.g. to 
`retrieveWarehousesWhichContainNoneOf`

de.metas.handlingunits.client/src/main/java/de/metas/handlingunits/client/terminal/editor/model/impl/MovementsAnyWarehouseModel.java
```diff
@@ -87,4 +85,40 @@ protected void createMovements()
...
+
+		// used for deciding the org and context
+		final I_M_HU firstHU = hus.get(0);
+
+		final int orgId = firstHU.getAD_Org_ID();
+		final Properties ctx = InterfaceWrapperHelper.getCtx(firstHU);
+
+		final List<org.compiere.model.I_M_Warehouse> warehouses = Services.get(IWarehouseDAO.class).retrieveForOrg(ctx, orgId);
+
+		final ImmutableList<org.compiere.model.I_M_Warehouse> huWarehouses = hus
```
metas-ts: we suggest to instead do
```java
final List<org.compiere.model.I_M_Warehouse> huWarehouses = hus
                 .stream()
                 .map(I_M_HU::getM_Locator_ID)
                 .distinct()
                 .map(id -> InterfaceWrapperHelper.load(id, I_M_Locator.class))
                 .map(org.compiere.model.I_M_Locator::getM_Warehouse_ID)
                 .distinct()
                 .map(id -> InterfaceWrapperHelper.load(id, I_M_Warehouse.class))
```

Btw, do you know if `distinct()` depends on the streamed items to be ordered?

de.metas.handlingunits.client/src/main/java/de/metas/handlingunits/client/terminal/editor/model/impl/MovementsAnyWarehouseModel.java
```diff
@@ -87,4 +85,40 @@ protected void createMovements()
...
+			// should never happen
+			return Collections.emptyList();
+		}
+
+		// used for deciding the org and context
+		final I_M_HU firstHU = hus.get(0);
+
+		final int orgId = firstHU.getAD_Org_ID();
+		final Properties ctx = InterfaceWrapperHelper.getCtx(firstHU);
+
```
metas-ts: What we think we would have done here is something like:
* extend IWarehouseDAO with a method such as `retrieve(WarehouseQuery)` and WarehouseQuery would have `getAD_Org_ID()` and `getExcludedWarehouseIds()` (maybe more)
* and then *first` then get my IDs from those streams and pass them to the warehouseDAO..instead of first getting all warehouses and then removing some...

But in order of getting the job done now, this is also OK

de.metas.handlingunits.client/src/test/java/de/metas/handlingunits/client/terminal/editor/model/impl/MovementsAnyWarehouseTests.java
```diff
@@ -0,0 +1,115 @@
...
+		final I_M_HU hu2 = createHU(locator2);
+		final I_M_HU hu3 = createHU(locator3);
+
+		List<I_M_HU> hus = new ArrayList<>();
+		hus.add(hu1);
+
+		final List<I_M_Warehouse> warehousesUnlessOfHUs1 = MovementsAnyWarehouseModel.getWarehousesUnlessOfHUs(hus);
+
+		Assert.assertFalse(warehousesUnlessOfHUs1.contains(warehouse1));
+		Assert.assertTrue(warehousesUnlessOfHUs1.contains(warehouse2));
```
metas-ts: Please take a look at assertj..
you and e.g. write this one as
```java
assertThat(warehousesUnlessOfHUs1).contains(warehouse2);
```

de.metas.handlingunits.client/src/main/java/de/metas/handlingunits/client/terminal/editor/model/impl/MovementsAnyWarehouseModel.java
metas-rc: From the examples I found online, (e.g. https://howtodoinjava.com/java-8/java-stream-distinct-examples/)  the stream doesn't have to be ordered.

de.metas.handlingunits.client/src/main/java/de/metas/handlingunits/client/terminal/editor/model/impl/MovementsAnyWarehouseModel.java
metas-rc: Thanks for the suggestion.

de.metas.handlingunits.client/src/test/java/de/metas/handlingunits/client/terminal/editor/model/impl/MovementsAnyWarehouseTests.java
metas-rc: Thanks! 
I notices the assertThat in other tests but I didn't know how to import it. I found the import online and just added it manually: import static org.assertj.core.api.Assertions.*;

Issue #2496, opened 2017-09-18
Different subscription Receiver for given timeframe


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, it is not possible to have a different subscription receiver for a given timeframe.
#### Which are the steps to reproduce?
Check and see.
### What is the expected or desired behavior?
It shall be possible to define a temporary different subscription receiver for a timeframe.

metas-rc:
Note: I tested this task on swing only, to see if it's fine functionally.
Testing it on webui is not doable at the time being because the frontend part for the date range parameter is not yet implemented.
This is the frontend task that will take care of it: metasfresh/metasfresh-webui-frontend#981

metas-rc:
Results of IT

* I created test data based on the documentation here: #2502

metas-rc:
Results of IT 2

* I tested again on swing because the frontend issue is not yet solved

* I created test data based on the documentation here: #2502
* I already had some shipment schedules created for my Contract.
* I selected my initial contract and ran the process " C_SubscriptionProgress_ChangeReceipients". I selected a time range that was overlapping with the timerange for which I had shipment schedules
* I selected another patner and address.
* I ran the process
* => OK: The process ran with no eror 
* => OK: DropShip_BPartner_ID and Address are modified in the Subscription Progress entries that are affected by the recepient change, the values are the ones I set.
* => OK: The Partner and Address in the shipment shcedules from the dates affected by the receipient change are also changed to the values I set.
* I ran the process again, now for a time range when I don't have shipment schedules.
* => OK : The process ran with no eror 
* => OK: The C_SubscriptionProgress entries for those dates when I changed the receipient have the
DropShip_BPartner_ID and Address that I chose in the process. 


* NOK: The Shipment Schedule is not displayed in the Subscription Progress entries even if they appear in the database.

Issue #2495, opened 2017-09-18
Pause subscription delivery for given timeframe


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, it is not possible to pause subscriptions for a given timeframe.
#### Which are the steps to reproduce?
Check and see.
### What is the expected or desired behavior?
It shall be possible to pause subscriptions.

metas-ts:
Hi eoss-sh @metasnw @metas-mk

i'm currently working in this implementation..i'm unclear about what to about **existing** delivery-pauses. My suggestions are:

* To remove pauses, add an action "remove pause" that is available if any pause-related Abo-Verlause (`C_SubscriptionProgress`) record is currently selected.
* to modify existing pauses, i suggest that whenever the "pause"-Process is called with dateFrom - dateTo values that overlap with an existing pause, then that existing pause is *removed* and the new one is created.

wdyt?

metas-ts:
Also, there is an important implementation "detail" to note:
We have those already mentioned Abo-Verlauf (`C_SubscriptionProgress`) records, where one can see past deliveries and influence planned (future) deliveries.

Now, each night, `C_SubscriptionProgress` records are turned into shipment schedule (Lieferdispo) records. Inserting a pause is only possible **before** this happened. 
If a shipment schedule already exists, but there shall be a pause, then this needs to be dealt with in the Lieferdispo window itself.
If you disaggree, we can discuss why this is IMHO a good approach or how to do it differently.

Also note that we can configure how many days in advance a shipment schedule for a given `C_SubscriptionProgress` record is created.

metas-rc:
Note: I tested this task on swing only, to see if it's fine functionally. 
Testing it on webui is not doable at the time being because the frontend part for the date range parameter is not yet implemented.
This is the frontend task that will take care of it: https://github.com/metasfresh/metasfresh-webui-frontend/issues/981

metas-rc:
Results of IT


* I followed the stept described in the IT of this task, to prepare entries in the testing databse: #2502
* Notice that the C_SubscriptionProgress entries that have shipment schedule entries counterparts have status = Shipment Open
* Go to the created Contract, run the process C_SubscriptionProgress_InsertPause
* Select a date range that covers a part of the days for which you have shipment schedules ( e.g. tomorrow, the day after nd the day after)
* Run the process
* Go to the shipment schedules for those days
* OK: They are closed ( Processed = true)
[...]

PR #2563, opened 2017-09-26
Pause subscription delivery for given timeframe #2495

metas-ts:
#2495
#2496


metas-ts:
@teosarca could you please take a look and comment? codebeat is not happy, but I still think the changes are reasonable..

Issue #2664, opened 2017-10-06
flatrate import: Make use of isdefaultcontact


metasnw:
### Is this a bug or feature request?
FR

### What is the current behavior?
new default flags implemented here: #2459 are ignored

#### Which are the steps to reproduce?
1. have a partner with issshiptocontact_default and isbilltocontact_default set to Y
1. import a flatrate for this partner
1. NOK: contact fields in flatrate term are empty

### What is the expected or desired behavior?
1. use isbilltocontact_default for billtouser
1. use issshiptocontact_default for dropshipuser

metas-ts:
cherry-picked into release

cristinamghita:
**To the IT team: have some regression in FT area. This task implied refactoring some contract validators.**

metasnw:
tested on 5.30 and did not work.
invoice partner has one billto_default_contact
dropship partner has a shipto_default_contact:

![image](https://user-images.githubusercontent.com/16079655/31584155-2707a484-b1a9-11e7-8c7c-610b6e08ba82.png)

metasnw:
Second problem:
Partner value of Invoice Partner is not resolved:
![image](https://user-images.githubusercontent.com/16079655/31584220-8bd4e9ca-b1aa-11e7-9eb8-603d8b66420d.png)

cristinamghita:
Tested with : 
```G0027,Abo,2017-01-03,2017-12-12,P002737,3.45,2,G0027,2016-12-27,2017-04-04```

metasnw:
problem No 1: tested on master and it worked fine
problem No 2: when using encoding UTF you have to use UTF with BOM or otherwise there will be a special character in front of the first value

PR #2682, opened 2017-10-09
#2664 flatrate import: Make use of isdefaultcontact 

cristinamghita:
#2664 flatrate import: Make use of isdefaultcontact


de.metas.contracts/src/main/java/de/metas/contracts/inoutcandidate/SubscriptionShipmentScheduleHandler.java
```diff
@@ -99,14 +100,31 @@
 		documentLocationBL.setBPartnerAddress(documentLocation);
 
 		final I_C_Order order = create(term.getC_Order_Term(), I_C_Order.class);
-		newSched.setDeliveryRule(order.getDeliveryRule());
-		newSched.setDeliveryViaRule(order.getDeliveryViaRule());
+		if (order == null)
```
metas-ts: i think it would be better to *alwasy* go with the term's values..the are set from the order when the term is created in `SubscriptionBL`, so they are there. and they might be changed in later terms..so the old order's rules might even be wrong...I'm not 100% sure if the old order is forwarded to reference, if a term is extended..even if it isn't we might change that in future

de.metas.contracts/src/main/java/de/metas/contracts/inoutcandidate/SubscriptionShipmentScheduleHandler.java
```diff
@@ -99,14 +100,31 @@
...
+		}
+		
 
 		newSched.setQtyOrdered(subscriptionLine.getQty());
 		newSched.setQtyOrdered_Calculated(subscriptionLine.getQty());
 		newSched.setQtyReserved(subscriptionLine.getQty());
 
-		newSched.setLineNetAmt(newSched.getQtyReserved().multiply(term.getC_OrderLine_Term().getPriceActual()));
+		final  I_C_OrderLine orerLine =  term.getC_OrderLine_Term();
+		if (orerLine == null)
```
metas-ts: same point I made about `C_Order_Term` note that this code dates back very far, so even if it's modernized here and there, it might contain all sorts of crap

de.metas.contracts/src/main/java/de/metas/contracts/interceptor/C_Order.java
```diff
@@ -48,83 +48,40 @@
...
-		return null;
-	}
-
-	@Override
-	public String modelChange(final PO po, final int type) throws Exception
+	@ModelChange(timings = { ModelValidator.TYPE_BEFORE_CHANGE })
+	public void updateDataEntry(final I_C_Order order)
 	{
-		if (type != TYPE_BEFORE_CHANGE)
+		if (InterfaceWrapperHelper.isValueChanged(order, I_C_Order.COLUMNNAME_DateOrdered))
```
metas-ts: i think you can use ignoreColumnsChanged within the ModelChange annotation for this

de.metas.contracts/src/main/java/de/metas/contracts/flatrate/impexp/FlatrateTermImporter.java
```diff
@@ -136,43 +143,26 @@ private void setDropShipBPartner(@NonNull final I_I_Flatrate_Term importRecord,
 		contract.setDropShip_BPartner_ID(dropShipBPartnerId);
 	}
 
-	private void setDropShipLocation(@NonNull final I_C_Flatrate_Term contract)
+	private void setDropShipUser(@NonNull final I_C_Flatrate_Term contract)
```
metas-ts: it might be a good idea to make methods static, if they don't use any of the classes members..but that's only a minor tip..not a must..

Issue #2727, opened 2017-10-13
Tax is not set properly when creating ICs from flatrate term


cristinamghita:
### Is this a bug or feature request?
bug
#### Which are the steps to reproduce?
Import FT and check the IC for that FT. It will have TaxNotFound
or
Extend a flatrate term and the IC of the new FT also will have TaxNotFound

### What is the expected or desired behavior?
The IC should have set the proper tax.

#### Test cases:
1. price system
   1. price list germany, tax excluded
      1. price list version valid from 01.10.2017
      1. product tax category 19%
      1. product price 90
   1. price list switzerland, tax included
      1. price list version valid from 01.11.2017
      1. product tax category 2,5%
1. FT 
   1. price 100 
   1. country of location: germany
   1. startdate 01.10.2017
   1. Expectation: tax 19%, net price 100, gross price 119
1. FT 2
   1. price 100 
   1. country of location: switzerland
   1. startdate 01.11.2017
   1. Expectation: tax 2,5%, net price 88,88  gross price 100

Nice to have:
1. FT 3 
   1. price NULL
   1. country of location: germany
   1. startdate 01.10.2017
   1. Expectation: tax 19%, net price 90, gross price 107,1

cristinamghita:
**Note to the dev:**
Check this ```de.metas.contracts.impl.FlatrateBL.createCand(Properties, I_C_Flatrate_Term, I_C_Flatrate_DataEntry, int, BigDecimal, String)```
The tax is retrieved with taxCategoryId = -1, which is not treated.

metasnw:
ToDos from my POV:
1. add c_taxcategory_id to c_flatrate_term
1. use that to compute c_tax_id on c_invoice_candidate
1. add c_taxcategory_id to flatrate import

metas-ts:
> add c_taxcategory_id to c_flatrate_term

i'm not sure..c_taxcategory_id is already in the pricing system..where also the price actual is coming from when a contract is extended

cristinamghita:
c_taxcategory_id  isn't in M_ProductPrice  ?

cristinamghita:
**Concept**
* Create pricing context based on flatrateTerm
* Create pricing result and computing price, but in the end, in the new IC, use price and uom from ft
* if the price is null in UOM. use the computed price(allow FT with price null??)
* Use the tax category computed like that

metas-ts:
Hi, here are my comments:

> Create pricing context based on flatrateTerm

[...]

PR #2748, opened 2017-10-16
#2727 Tax is not set properly when creating ICs from flatrate term 

cristinamghita:
#2727 Tax is not set properly when creating ICs from flatrate term


de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
```diff
@@ -1125,6 +1121,9 @@ private void extendContract0(
 			nextTerm.setDropShip_BPartner_ID(currentTerm.getDropShip_BPartner_ID());
 			nextTerm.setDropShip_Location_ID(currentTerm.getDropShip_Location_ID());
 
+			nextTerm.setC_TaxCategory_ID(currentTerm.getC_TaxCategory_ID());
```
metas-ts: hi, i'm not 100% sure about the context, i only took a quick look, but: do you set the priceactual in here? i could not fid it..if it's not set in there, then i think you also should *not* set the two tax values.
i think it's computed somewhere in a listener..look for code calling the setter.
more: imho it should **not** just be copied. instead, it needs to be computed, just like priceactuall itself

cristinamghita: Yes, you are right! Fixed in follow up task: #2752

Issue #1937, opened 2017-07-02
Remove Window and Tab Translations which are not System Language


metas-mk:
### Is this a bug or feature request?
Feature Request (Housekeeping)
### What is the current behavior?
currently, there are some Translation entries available for en_GB and fr_FR which are not marked as System language.
#### Which are the steps to reproduce?
Open different windows from WebUI menu and check.
### What is the expected or desired behavior?
Do some Housekeeping and provide a Script to tidy the Translations which are not System Language.

PR #2776, opened 2017-10-20
#1937 Maintaining System Languages and Translations

metas-mk:
Maintaining System Languages and Translations

FRESH-2791 https://github.com/metasfresh/metasfresh/issues/1937


metas-ts:
Hi @metas-mk i guess it makes more sense to repeat the procedure now..
if you disagree, please reopen and fix the SQL

Issue #2781, opened 2017-10-20
Change Default Address Layout for B2C Partners for country CH


metasnw:
### Is this a bug or feature request?
bug

### What is the current behavior?
if partner is company = N and has got a contact
the address in order, invoice, etc. is rendered without name for country CH

### What is the expected or desired behavior?
shall be the same as for country DE

cristinamghita:
## Result of IT
### Tested on webui
* Created a new bpartner which is not company
* Add 2 locations:
  * one for DE
  * one for CH
* Created order and then invoice, using DE location
  * Address generated does not show the country since is local. The client is in DE. **OK**
* Created order and then invoice, using CH location
  * Address generated shows the country **OK**
    ![image](https://user-images.githubusercontent.com/15378036/35104490-b0a0297e-fc71-11e7-88ba-408b011287da.png)

PR #2782, opened 2017-10-20
#2781 Change Default Address Layout for B2C Partners for country CH

metasnw:
#2781 adding contact to displaysequence


metasnw:
@metas-mk impacted by this change are only partners that are not companies in CH

Issue #2800, opened 2017-10-23
Make de.metas.document.archive.process.ExportArchivePDF work in webui


cristinamghita:
### Is this a bug or feature request?
FR
### What is the current behavior?
pdf is not open
#### Which are the steps to reproduce?
Run process Open PDF from Printing queue window
### What is the expected or desired behavior?
Process should work and the pdf file should be open in browser.

cristinamghita:
Follow up task for a proper name:
[#2807 Fix archive name when pdf printing](https://github.com/metasfresh/metasfresh/issues/2807)

metasnw:
IT was successful

PR #2801, opened 2017-10-23
#2800  Make de.metas.document.archive.process.ExportArchivePDF work in webui 

cristinamghita:
#2800  Make de.metas.document.archive.process.ExportArchivePDF work in webui

### DIFF
--- a/de.metas.document.archive/de.metas.document.archive.base/src/main/java/de/metas/document/archive/process/ExportArchivePDF.java
+++ b/de.metas.document.archive/de.metas.document.archive.base/src/main/java/de/metas/document/archive/process/ExportArchivePDF.java
@@ -64,9 +64,9 @@ public class ExportArchivePDF extends JavaProcess implements IProcessPreconditio
 	
 	private void openPdfFile(@NonNull final byte[] data, @NonNull final String contentType, @NonNull final String filename)
 	{
-		final boolean backEndorSwingMode = Ini.getRunMode() == RunMode.BACKEND || Ini.isClient();
+		final boolean backEndOrSwing = Ini.getRunMode() == RunMode.BACKEND || Ini.isClient();
 		
-		if (backEndorSwingMode)
+		if (backEndOrSwing)
 		{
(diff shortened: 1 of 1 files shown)


de.metas.document.archive/de.metas.document.archive.base/src/main/java/de/metas/document/archive/process/ExportArchivePDF.java
```diff
@@ -51,9 +55,24 @@ protected String doIt()
 		final IArchiveBL archiveBL = Services.get(IArchiveBL.class);
 		final byte[] data = archiveBL.getBinaryData(archive);
 		final String contentType = archiveBL.getContentType(archive);
-		final String filename = null;
+		final String filename = String.valueOf(archive.getRecord_ID());
```
teosarca: hmm, is this OK? Is it opened nicely on webui?

cristinamghita: Yes. I mean is just used as the pdf name the id. (the easiest way to have a name)
image
)

But, in the end I think we should use the archive name. 
But that is broken now.
I will do a follow up task. Is related to pdf printing, I think.

teosarca: thanks, also pls check what happens when u try to save it (ctrl+s). What's the suggested filename?
I guess it would be your number (without pdf extension)...

cristinamghita: Yes. Is the number and no extension. ('All files' where is the extension)
And after you save it and you try to open, windows is suggesting to open it with adobe acrobat reader.

cristinamghita:
Created this [#2807 Fix archive name when pdf printing](https://github.com/metasfresh/metasfresh/issues/2807)

Issue #2894, opened 2017-11-02
Sys Config to set Waiting Time for Async


metas-rc:
### Is this a bug or feature request?
* FR
### What is the current behavior?
* There are 3 minutes which the backend takes until it fires up the async processor.

#### Which are the steps to reproduce?
* Check and see
### What is the expected or desired behavior?
* Please, introduce a sys config to set this time and initialize it with 60 seconds

metas-rc:
Note: The setting is made here: de.metas.async.model.validator.Main.getInitDelayMillis()

metas-rc:
THe following update set the sys config to 10 seconds, the value that we use in developer mode. I added it in the after import script, too.

`UPDATE AD_SysConfig SET Value = 10000 WHERE AD_SysConfig_ID=541168 ;`

metas-lc:
IT
checked the sysconfig, is set to 60000 OK

PR #2896, opened 2017-11-02
#2894 sys config to set initDelayMills for Async

metas-rc:
https://github.com/metasfresh/metasfresh/issues/2894

### DIFF
--- a/de.metas.async/src/main/java/de/metas/async/model/validator/Main.java
+++ b/de.metas.async/src/main/java/de/metas/async/model/validator/Main.java
@@ -25,11 +25,11 @@ package de.metas.async.model.validator;
 import org.adempiere.ad.migration.logger.IMigrationLogger;
 import org.adempiere.ad.modelvalidator.AbstractModuleInterceptor;
 import org.adempiere.ad.modelvalidator.IModelValidationEngine;
-import org.adempiere.ad.service.IDeveloperModeBL;
 import org.adempiere.ad.session.MFSession;
 import org.adempiere.impexp.IImportProcessFactory;
 import org.adempiere.impexp.spi.impl.AsyncImportProcessBuilder;
 import org.adempiere.impexp.spi.impl.AsyncImportWorkpackageProcessor;
+import org.adempiere.service.ISysConfigBL;
 import org.adempiere.util.Check;
 import org.adempiere.util.Services;
 import org.compiere.Adempiere.RunMode;
@@ -57,6 +57,10 @@ import de.metas.event.IEventBusFactory;
  */
 public class Main extends AbstractModuleInterceptor
 {
+	public static transient String SYSCONFIG_Async_InitDelayMillis = "Async_InitDelayMillis";
+
+	public static transient int THREE_MINUTES = 3 * 60 * 1000;
+
 	@Override
 	public void onInit(final IModelValidationEngine engine, final I_AD_Client client)
 	{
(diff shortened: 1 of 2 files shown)


de.metas.async/src/main/java/de/metas/async/model/validator/Main.java
```diff
@@ -57,6 +57,10 @@
  */
 public class Main extends AbstractModuleInterceptor
 {
+	public static transient String SYSCONFIG_Async_InitDelayMillis = "Async_InitDelayMillis";
```
metas-ts: * Please prefix this with `de.metas.async`.
* It's important to have the UOM (millis) like you have them here
* in this case, seconds might also be sufficuently granular, but millis is also ok

de.metas.async/src/main/sql/postgresql/system/42-de.metas.async/5476160_sys_gh2894_SysConfig_Async_InitDelayMillis.sql
```diff
@@ -0,0 +1,10 @@
+-- 2017-11-02T15:32:55.344
+-- URL zum Konzept
+INSERT INTO AD_SysConfig (AD_Client_ID,AD_Org_ID,AD_SysConfig_ID,ConfigurationLevel,Created,CreatedBy,EntityType,IsActive,Name,Updated,UpdatedBy,Value) VALUES (0,0,541168,'S',TO_TIMESTAMP('2017-11-02 15:32:55','YYYY-MM-DD HH24:MI:SS'),100,'de.metas.async','Y','Async_InitDelayMillis',TO_TIMESTAMP('2017-11-02 15:32:55','YYYY-MM-DD HH24:MI:SS'),100,'60000')
```
metas-ts: please add some documentation like "When the async component is loaded, this delay makes sure that it was fully started by the time the async component starts to process packages" ...or something more sensible..

metas-rc: Thanks. Documentation added in both the description of the sys config and in the code where it is used.

de.metas.async/src/main/java/de/metas/async/model/validator/Main.java
metas-rc: Done

de.metas.async/src/main/java/de/metas/async/model/validator/Main.java
```diff
@@ -57,7 +57,11 @@
  */
 public class Main extends AbstractModuleInterceptor
 {
-	public static transient String SYSCONFIG_Async_InitDelayMillis = "Async_InitDelayMillis";
+	/**
```
metas-ts: * i think for a dev looking at the code, it's clear. also because that constant is (afaiu) used only on this class.
  * to that end, please make the two constants `private` if they aren't intended to be used anywhere else. If they are referenced in a test case, plase annotate them with `@VisibleForTesting` from the guava library
* but i'm fine with the comment if you want to keep it
* a good constant should be not only upper-case, fut also `final` (not `transient `) :-P

metas-rc: Done:
* deleted the javadoc. It's undertstandable from the code what that sys config is supposed to be, and in the database the description is kept. So if somebody wants to update it will know what it means from there.
* made the constants private. They are not used in any test.
* made the constant uppercase and final.

Thanks for the review! :)

Issue #2917, opened 2017-11-06
Allow reactivating procurement contracts


metas-ts:
### Is this a bug or feature request?
Bug
### What is the current behavior?

* Currently, `C_Flatrate_Term`s can't be reactivated in general

#### Which are the steps to reproduce?
Try it out

### What is the expected or desired behavior?

For contracts that have the type *procurement* ("Liefervereinbahrung") we still need a way to change their prices. Ideally this would be done with a process and a nice modal dialog,
but for non, we just add a way of allowing an admin to enable reactivating for specific contract types

metas-dh:
Results of IT1
tested in spv e2e

* contracts with type procurement ("Liefervereinbahrung"): can be reactivated, to change the price: OK
* contracts with e.g. type Leergut: cannot be reactivated, you need to change the SysConfig for that: OK

PR #2918, opened 2017-11-06
Allow reactivating procurement contracts #2917

metas-ts:
https://github.com/metasfresh/metasfresh/issues/2917


metas-ts:
this branch shall *not* be deleted for the time being!

Issue #2920, opened 2017-11-06
Create automated test for extending contracts


cristinamghita:
### Is this a bug or feature request?
FR
### What is the current behavior?
We don;t have any automated test regarding extending a contract
### What is the expected or desired behavior?
Have unit tests

metas-rc:
Nothing to IT, it's only about unit tests

PR #2925, opened 2017-11-07
#2920 Create automated test for extending contracts 

cristinamghita:
#2920 Create automated test for extending contracts


cristinamghita:
Done

Issue #2905, opened 2017-11-03
referenced documents: sales order to invoice candidates


metas-lc:
### Is this a bug or feature request?
bug
### What is the current behavior?
ICs that are made after inout (like HUs) are not linked with the sales order in referenced documents
#### Which are the steps to reproduce?
1. create&complete a sales order with a product + HU
2. create inout
3. zoom in from sales order to invoice candidates 
=> there is only 1 NOK (the HU is missing)
4. zoom in from inout to ICs 
=> there are 2 ICs OK 
### What is the expected or desired behavior?
step 3: shall see both ICs (also the ones from inout)

metas-ts:
..the reason is *probably* related to

> the zoom from sale order might just zoom via C_OrderLine_ID, and the iol-IC only references the C_Order_ID

in that case the fix would be to make sure that it zooms by C_Order_ID and not orderline....probably if the problem is in that field, the C_ORdeRLine-base zoom should be removed altogether

metas-dh:
Results of IT1
tested in mf15

* zoom in from sales order to invoice candidates: IC for product, and ICs for HUs from inout(s): OK
* reversing / voiding shipment: all ICs still displayed in referenced documents, qty just 0: OK
* working the same for purchase order / material receipt: OK

PR #2927, opened 2017-11-07
Gh2905 Make the relation between order and invoice candidate not depend on orderline

metas-rc:
https://github.com/metasfresh/metasfresh/issues/2905


metas-rc:
Done.
Thx for feedback!

metasnw:
expectation of users that just created an order with the intention to invoice it expect a direct way to jump

Issue #2956, opened 2017-11-11
drop allowconsolidateinvoice from database


metasnw:
### Is this a bug or feature request?
FR

### What is the current behavior?
db column is there although its not used anymore as it was replaced with so_invoice_aggregation_id

#### Which are the steps to reproduce?
check table and column

### What is the expected or desired behavior?
1. migrate existing records to std aggregation
1. drop column from db

metasnw:
convert existing records:
```
update c_bpartner
set so_invoice_aggregation_id= 1000000
where so_invoice_aggregation_id is null and allowconsolidateinvoice ='Y';
```

metas-lc:
IT
allowconsolidateinvoice is not in c_bpartner anymore 
there is no element for it 
order to invoice creation works fine, no errors
OK

PR #2957, opened 2017-11-11
#2956 drop allowconsolidateinvoice from database

metasnw:
#2956 setting default aggregation on partners that have
isconsolidateinvoice = Y, dropping column


teosarca:
@metasnw agree. I think
* we shall backup that table column, just in case somebody needs it (e.g. create table backup.C_BPartner_before_2956 as select * from C_Bpartner)
* simply drop that column

@metas-ts  wdyt?

de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5476950_sys_gh2956_drop_allowconsolidateinvoice.sql
```diff
@@ -0,0 +1,7 @@
+--set default aggregation for partner with activated consolidation
+update c_bpartner
```
metas-ts: hi @metasnw i think I aggree with what you wrote..
so imho we should drop the column (and create a backup, as teo suggested),
but not add an so_invoice_aggregation_id to partners that have none right now.

also with this search:
image

I get a lot of commented-out code..so while were are here, i propose to remove that commented-out code too..
maybe we can ask @metas-rc to finish this task according to the proposal?

metas-ts: update @metasnw :
i yesterday asked @metas-rc to add it to her todo-list

Issue #2911, opened 2017-11-03
Create Action for changing price and qty for a contract


metasnw:
### Is this a bug or feature request?
FR

### What is the current behavior?
you can change price or qty after completing the contract only in invoice candidate

#### Which are the steps to reproduce?
1. complete a contract and see

### What is the expected or desired behavior?
have an action allowing to change price and qty 

results

- [ ] price is written to priceactual
- [ ] qty is written to 
   - [ ] flatrate term plannedqtyperunit
   - [ ] qty in subscription progress for future records that have not been delivered
- [ ] if invoice candidate for the term has not been invoiced its recalculated with new price and qty. if it was invoiced leave it.

cristinamghita:
### Concept
* Create a new process that will allow changing price and Qty

cristinamghita:
We have 2 new processes. See image.
![image](https://user-images.githubusercontent.com/15378036/32906297-2ac40ac0-cb05-11e7-8927-04ac2c49ddce.png)

metas-dh:
Results of IT1
tested in mf15

* new processes for changing price and qty for a contract: OK (need translation: #3200)
* price is written to priceactual after changing price process: OK
* qty is written to
   * flatrate term plannedqtyperunit: only in grid view, single view still shows old plannedqtyperunit: NOK
   * qty in subscription progress for future records that have not been delivered: OK
* if invoice candidate for the term has not been invoiced its recalculated with new price and qty: new price is OK, new qty is OK in grid view only, single view still shows old plannedqtyperunit: NOK
   * if it was invoiced leave it: OK

PR #3000, opened 2017-11-16
#2911 Create Action for changing price and qty for a contract 

cristinamghita:
#2911 Create Action for changing price and qty for a contract


de.metas.contracts/src/main/java/de/metas/contracts/flatrate/process/C_Flatrate_Term_ChangePriceQty.java
```diff
@@ -0,0 +1,35 @@
...
+import de.metas.process.Param;
+
+public class C_Flatrate_Term_ChangePriceQty extends JavaProcess
+{
+
+	public static final String PARAM_PRICEACTUAL = I_C_Flatrate_Term.COLUMNNAME_PriceActual;
+	public static final String PARAM_PLANNED_QTY_PER_UNIT = I_C_Flatrate_Term.COLUMNNAME_PlannedQtyPerUnit;
+
+
+	@Param(parameterName = PARAM_PRICEACTUAL, mandatory = false)
```
metas-ts: i would suggest to just inline the column names here, and not have the two constants..as far as i ca see therey aren't used anywhere outside this process class.
....or ...teo sais, it's ok to have them, but make them `private`

de.metas.contracts/src/main/java/de/metas/contracts/impl/ContractChangePriceQtyRepository.java
```diff
@@ -0,0 +1,106 @@
...
+ */
+
+/**
+ * @author metas-dev <dev@metasfresh.com>
+ *
+ */
+@Component
+public class ContractChangePriceQtyRepository
+{
+	public void changePriceIfNeeded(@NonNull final I_C_Flatrate_Term contract, final BigDecimal priceActual)
```
metas-ts: i think this is not a repository..
a repositor would just get ythe entities from our DB for you..
we think this is a service
..aand, please use the `@Service` annotation..to make it more transparent..

de.metas.contracts/src/test/java/de/metas/contracts/impl/AbstractFlatrateTermTest.java
```diff
@@ -325,8 +338,8 @@ protected I_C_Flatrate_Term createFlatrateTerm(@NonNull final I_C_Flatrate_Condi
 		contract.setDropShip_BPartner(getBpartner());
 		contract.setDropShip_Location(bpLocation);
 		contract.setDropShip_User(user);
-		contract.setPriceActual(BigDecimal.valueOf(2));
-		contract.setPlannedQtyPerUnit(BigDecimal.ONE);
+		contract.setPriceActual(PRICE);
```
metas-ts: please rename to `QTY_TEN` and `PRICE_ONE`..it makes it more readable when you use them somewhere else

de.metas.contracts/src/main/java/de/metas/contracts/impl/ContractChangePriceQtyRepository.java
cristinamghita: Yes. After reading, I also had that feeling. But wasn't sure. Is the first attempt.
Thx!

de.metas.contracts/src/main/java/de/metas/contracts/flatrate/process/C_Flatrate_Term_ChangePriceQty.java
cristinamghita: agree

de.metas.contracts/src/test/java/de/metas/contracts/impl/AbstractFlatrateTermTest.java
cristinamghita: agree

cristinamghita:
Done! Thx for feedback!

PR #3019, opened 2017-11-20
Gh699 webui DefaultValue for M_HU_PI_Item_product.C_UOM_ID

metas-rc:
https://github.com/metasfresh/metasfresh-webui-api/issues/669


de.metas.handlingunits.base/src/main/sql/postgresql/system/5477570_sys_gh699-webui_DefaultValue_For_M_HU_PI_Item_Product_UOM.sql
```diff
@@ -0,0 +1,4 @@
+-- 2017-11-20T12:00:14.710
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+UPDATE AD_Column SET DefaultValue='@SQL=select c_uom_id from m_product where M_product_ID=@M_Product_ID@',Updated=TO_TIMESTAMP('2017-11-20 12:00:14','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Column_ID=549286
```
teosarca: i find this fix too fragile... it's only by coincidence that the C_UOM_ID default value is evaluated before M_Product_ID.

Looking a bit more into de.metas.handlingunits.model.validator.M_HU_PI_Item_Product i've observed that setNameAndDescription is called BEFORE setUOMItemProduct. That's why we get the case:
image

teosarca: i suggest to have only one beforeSave method there and call the other methods in right order.
Keep in mind that JVM is not guaranteeing that the order of  the methods as they are defined in java file will be preserved.

PR #3199, opened 2017-12-12
port current MRP Product Info Window to the webui #706 

metas-ts:
port current MRP Product Info Window to the webui https://github.com/metasfresh/metasfresh-webui-api/issues/706


metas-ts:
..instead we are going to directly integrate https://github.com/metasfresh/metasfresh/pull/3217

Issue #3280, opened 2018-01-05
Contracts: Make sure that if extending process crashes while extending, doesn't rollback all


cristinamghita:
### Is this a bug or feature request?
Bug

PR #3283, opened 2018-01-05
#3280 Contracts: Make sure that if extending process crashes while extending, doesn't rollback all 

cristinamghita:
#3280 Contracts: Make sure that if extending process crashes while extending, doesn't rollback all


de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
```diff
@@ -1006,46 +1007,25 @@ else if (diffPercent.signum() < 0 && diffPercent.abs().compareTo(conditions.getM
...
 	@Override
-	public void extendContract(
-			final I_C_Flatrate_Term currentTerm,
-			final boolean forceExtend,
-			final Boolean forceComplete,
-			final Timestamp nextTermStartDate,
-			final I_C_OrderLine ol)
+	public void extendContract(final @NonNull ContractExtendingRequest context)
 	{
+		Services.get(ITrxManager.class).run(ITrx.TRXNAME_ThreadInherited, (TrxRunnable)localTrxName -> {
```
metas-ts: i think we don't need the cast to `TrxRunnable`

de.metas.contracts/src/main/java/de/metas/contracts/IFlatrateBL.java
```diff
@@ -89,17 +93,30 @@
 	void updateEntry(I_C_Flatrate_DataEntry dataEntry);
 
 	/**
-	 * Create a new flatrate term using the given term as template. The new term's C_Year will be the year after the given term's C_Year.
-	 *
-	 * <b>IMPORTANT:</b> the method might set the given term's C_FlatrateTerm_Next_ID, but won't save it!
+	 * term to extend
+	 * forceExtend - will create a new term, even if the given <code>term</code> has <code>IsAutoRenew='N'</code>
+	 * forceComplete - will complete a new term (if one has been created), even if it has <code>IsAutoComplete='N'</code>
```
metas-ts: here is an outdated (a.k.a. lying) comment

cristinamghita: yes, right!
Thx!

de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
cristinamghita: Yes, is added by default by eclipse, due to a setting

Issue #3070, opened 2017-11-24
bold missing in pdfs generated with jasper


metas-lc:
### Is this a bug or feature request?
bug
### What is the current behavior?
before, we used to solve it by adding this " pdfFontName="Helvetica-Bold"" on the bold fields.
But now this is not working well anymore, and if you do that you get a warning about the use of PDF font is deprecated and replaced by the Font Extension. 
#### Which are the steps to reproduce?
1. run a print preview of a jasper you know it has bold fields, without  " pdfFontName="Helvetica-Bold""
=> the fields are bold OK
2. export as pdf
=> fields are not bold NOK
### What is the expected or desired behavior?
find how Font Extension works and make exported PDFs bold, without using the deprecated solution

metas-lc:
will postpone this task for now, because where we have helvetica bold, the text is double bolded. Will have to take out all helvetica bolds in all jaspers, with this task

PR #3334, opened 2018-01-16
#3070 add dependency for fonts

metas-lc:
https://github.com/metasfresh/metasfresh/issues/3070


metas-ts:
clöosing, as talked with @metas-lc

Issue #3354, opened 2018-01-19
Different subscription Receiver for ever


metasnw:
### Is this a bug or feature request?
FR

### What is the current behavior?
behaves like described in #2496

### What is the expected or desired behavior?
the user needs the option to permanently change the delivery partner

cristinamghita:
## Rough concept
* Modify process de.metas.contracts.subscription.process.C_SubscriptionProgress_ChangeRecipient to allow DateTo empty
* Modify de.metas.contracts.subscription.impl.subscriptioncommands.ChangeRecipient.createRecipienRecordsQuery(ChangeRecipientsRequest) so that filter all subscriptions, starting with date From and allowing to chande the receiver to all

cristinamghita:
CPed to Release branch

metas-rc:
Results of IT

* Have a contract
* Open the process for chaning receipient
* OK: The date parameter is hidden after I click on the permanent flag 
* NOK: I can't run the process because the date parameter is still mandatory 
```
Error: Mandatory field DateGeneral not...
Server error
Mandatory field DateGeneral not filled
```

![image](https://user-images.githubusercontent.com/16001442/37906136-cafb48cc-3109-11e8-83ad-2a8c135ab933.png)

metas-rc:
Results of IT


Tested on webui

* prepared testing data using IT here: https://github.com/metasfresh/metasfresh/issues/2502

* OK: the process is runnable now with the flag IsPermanentReceipient on true
* OK: The date parameter dissapears after I set the flag on true
* OK: After I run the process, all the exisitng subscritpions from now on are changed to have the selected patner and location


Regression: 
* Use he process with a date range
* => OK: The subscription progress entries in the selected range are changed according to the parameters
* OK: The rest of the subsctiprions are left untouched.

PR #3390, opened 2018-01-26
#3354 Different subscription Receiver for ever 

cristinamghita:
#3354 Different subscription Receiver for ever


de.metas.contracts/src/main/java/de/metas/contracts/subscription/process/C_SubscriptionProgress_ChangeRecipient.java
```diff
@@ -50,6 +50,9 @@
 	@Param(parameterName = "DropShip_User_ID", mandatory = false)
 	private int DropShip_User_ID;
 
+	@Param(parameterName = "IsPermanentRecipient", mandatory = true)
```
metas-ts: watching the full file, i saw that dateTo is aparently still mandatory..
how does it work? maybe you should make it non-mandatory, or set a default and set the param's display logic (if possible)?

de.metas.contracts/src/main/java/de/metas/contracts/subscription/impl/SubscriptionService.java
```diff
@@ -84,14 +84,17 @@ public static int changeRecipient(@NonNull final ChangeRecipientsRequest changeR
...
 		@Builder
 		private ChangeRecipientsRequest(
 				@NonNull final I_C_Flatrate_Term term,
 				@NonNull final Timestamp dateFrom,
 				@NonNull final Timestamp dateTo,
 				final int DropShip_BPartner_ID,
 				final int DropShip_Location_ID,
-				final int DropShip_User_ID)
+				final int DropShip_User_ID,
+				boolean IsPermanentRecipient)
```
metas-ts: afaiu, here you could assert that IsPermanentRecipient needs to be true, or dateTo needs to be != null

de.metas.contracts/src/main/java/de/metas/contracts/subscription/process/C_SubscriptionProgress_ChangeRecipient.java
cristinamghita: @metas-ts if the IsPermanentRecipient is set on yes, that entire field is ignored.
DateFrom and DateTo come from a date field range parameter, so are all the time field up.
But, what I could do(not sure if is working in webui this) is to add a display logic for the dates field, which will be hidden if the user ticks IsPermanentRecipient flag.
WDYT?

Issue #3355, opened 2018-01-19
Add option to automatically credit open invoice for terminated contract / subscription


metasnw:
### What is the current behavior?
http://docs.metasfresh.org/webui_collection/EN/Contract_termination.html

### What is the expected or desired behavior?

- [ ] new checkbox set to Y telling "automatically credit any existing invoice for this contract"
- [ ] if an invoice with an open amount (complete or partially) exists create a credit memo for the invoice and allocate so the invoice is paid

cristinamghita:
## Dev test
### Simple contract
* Create a sales order with flatrate conditions
* Invoice the IC for that
* Go to contract window and cancel the contract, choosing to credit open invoices
     ![image](https://user-images.githubusercontent.com/15378036/35554451-6179893a-05a4-11e8-9c52-ab24c120df42.png)
* Result **OK**
    *  The contract was cancelled **OK**
    * Invoice was credited **OK**
### Contract extended
* Create a sales order with flatrate conditions
* Extend that contract
* Invoice the IC for both contracts
* Go to contract window and cancel the initial contract, choosing to credit open invoices
     ![image](https://user-images.githubusercontent.com/15378036/35554451-6179893a-05a4-11e8-9c52-ab24c120df42.png)
* Result **OK**
    *  Both contracts were cancelled **OK**
    * Both invoices were credited **OK**

metas-ts:
CPed into `release`

metas-lc:
IT

see steps: 9, 22

. invoice - cancel contract - credit memo OK
1. create new sales order, add product and a flatrate condition for that bpartner and product
2. complete the order
=> new contract was created OK
=> status running (laufend) OK
3. create invoice
4. go to contract and cancel it (kündigen) + credit open invoices
=> contract status is now Quit (Gekündigt) OK
=> new credit memo was created for my invoice. Invoice is paid OK

. extend contract - invoices - cancel contracts - credit memos NOK?
5. create a new sales order,  add product and same flatrate condition from step 1
6. complete the order
=> new contract was created OK
=> status running (laufend) OK
7. extended contract 
=> contract extended with a month in my case
=> new contract created, new IC created
8. go to IC window and invoice both ICs (one by one)
=> 2 new invoices created
9. go to contract window and cancel the initial contract, choosing to credit open invoices
=> Error:
[...]

PR #3401, opened 2018-01-30
#3355 Add option to automatically credit open invoice for terminated contract / subscription 

cristinamghita:
#3355 Add option to automatically credit open invoice for terminated contract / subscription

### DIFF
--- a/de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateDAO.java
+++ b/de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateDAO.java
@@ -73,6 +73,7 @@ import de.metas.document.engine.IDocument;
 import de.metas.invoicecandidate.api.IInvoiceCandDAO;
 import de.metas.invoicecandidate.model.I_C_Invoice_Candidate;
 import de.metas.logging.LogManager;
+import de.metas.util.stream.StreamUtils;
 import lombok.NonNull;
 
 public class FlatrateDAO implements IFlatrateDAO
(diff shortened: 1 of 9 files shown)


de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateDAO.java
```diff
@@ -843,4 +841,21 @@ public I_C_Flatrate_Term retrieveAncestorFlatrateTerm(@NonNull final I_C_Flatrat
...
+	public List<I_C_Invoice> retrieveInvoicesForCurrentContract(@NonNull final I_C_Flatrate_Term contract)
+	{
+		final Properties ctx = InterfaceWrapperHelper.getCtx(contract);
+		final String trxName = InterfaceWrapperHelper.getTrxName(contract);
+		final List<I_C_Invoice> currentFlatRateTermInvoices = new ArrayList<>();
+		final List<I_C_Invoice_Candidate> icsForCurrentTerm = Services.get(IInvoiceCandDAO.class).fetchInvoiceCandidates(ctx, I_C_Flatrate_Term.Table_Name, contract.getC_Flatrate_Term_ID(), trxName);
+		icsForCurrentTerm.forEach(ic -> {
+			final List<I_C_InvoiceLine> iclList = Services.get(IInvoiceCandDAO.class).retrieveIlForIc(ic);
+			iclList.stream()
+					.filter(il -> !currentFlatRateTermInvoices.contains(il.getC_Invoice()))
```
metas-ts: hi, i just researched a bit and added (to the `master` branch)
`StreamUtils.distinctByKey()` which returns a filter predicate..with this, you could write something like
```java
iclList.stream()
  .filter(StreamUtils.distinctByKey(I_C_InvoiceLine::getC_Invoice_ID))
  .map(il -> il.getC_Invoice())
  .collect(Collectors.asList());
```
i think it's more readable, and we don't have to assume that I_C_Invoice.equals works our way

metas-ts: Other thing: i think codacy has a hint for a quick-fix for you:
"Unnecessary use of fully qualified name 'InterfaceWrapperHelper.save' due to existing static import 'org.adempiere.model.InterfaceWrapperHelper.save'"

Issue #3413, opened 2018-01-30
Credit Limit data structure as subtab in Businesspartner


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently the Credit Limit functionality only allows 1 setting for credit limit of a business partner.
#### Which are the steps to reproduce?
Open and see,
### What is the expected or desired behavior?
Create a new Data Structure that has the following content (besides the default columns):
a) Reference List entry "credit limit type" e.g. managing director, insurance
b) "date"
c) "createdBy"
d) "credit limit amount"

Create a Credit Limit indicator in BusinessPartner Header (virtual column) that says something like "95%" (which means the business partner credit limit has reached 95%).

cristinamghita:
##Rough concept
* Create a new table called C_BPartner_CreditLimit with next columns:
  * Type (ref list e.g. managing director, insurance)
  * Amount
  * Date
  * ApprovedBy
  * C_BPartner_ID
  * and the usual mandatory fields
* Create a virtual column in C_BPartner - CreditLimitUsage - that will show the percent obtained from total opened balanced and credit limit
* for computing total opened balance, use the logic from org.adempiere.bpartner.service.impl.BPartnerStatsDAO.retrieveTotalOpenBalance(IBPartnerStats)

cristinamghita:
![image](https://user-images.githubusercontent.com/15378036/35631344-5288af24-06ac-11e8-88fe-a318ea1e2939.png)

metas-lc:
IT 
Please check the following NOKs: steps 5, 6 and probably 10

1. open bpartner, select one and go to last subtab: Business Partner Credit Limit
2. add a new line (insurance, amt 100, date from and date to)
3. add a new line, exactly the same as the line above
=> you can still see both lines and there is no red line indicating the Bpartner is not saved (frontend issue) NOK
4. remove the second line
5. set the amount to 0 
=> you get errors (division by 0) and you can't open bparther anymore NOK
6. created some invoices 
=> Credit limit Usage is still 0. NOK the credit limit usage doesn't update if you don't recreate the limit line
7. add a new line with Managing Director, amt 50
=> Credit limit Usage updated OK
8. if you have insurance and managing, always managing has priority OK
9. no matter what currency you have in invoice, the Credit limit Usage updates OK
10. i reverted an invoice and Credit limit Usage didn't update NOK i think
--
[...]

PR #3424, opened 2018-01-31
#3413 Credit Limit data structure as subtab in Businesspartner 

cristinamghita:
#3413 Credit Limit data structure as subtab in Businesspartner


de.metas.adempiere.adempiere/base/src/main/java-gen/org/compiere/model/X_C_BPartner_CreditLimit.java
```diff
@@ -0,0 +1,207 @@
...
+      org.compiere.model.POInfo poi = org.compiere.model.POInfo.getPOInfo (ctx, Table_Name, get_TrxName());
+      return poi;
+    }
+
+	/** Set Betrag.
+		@param Amount 
+		Betrag in einer definierten Währung
+	  */
+	@Override
+	public void setAmount (java.math.BigDecimal Amount)
```
metas-ts: i think if we have an amount, we also need a currency here..wdyt?

de.metas.adempiere.adempiere/base/src/main/java-gen/org/compiere/model/X_C_BPartner_CreditLimit.java
```diff
@@ -0,0 +1,207 @@
...
+		Integer ii = (Integer)get_Value(COLUMNNAME_C_BPartner_ID);
+		if (ii == null)
+			 return 0;
+		return ii.intValue();
+	}
+
+	/** Set Datum.
+		@param DateGeneral Datum	  */
+	@Override
+	public void setDateGeneral (java.sql.Timestamp DateGeneral)
```
metas-ts: from the column name, i don't get what DateGeneral is about..
if i think it should rather be `validFrom`or `validTo`..if i correctly guessed that the date is about

de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5484320_sys_gh3413_CreateTableAndWindo_C_BPartner_CreditLimit.sql
```diff
@@ -0,0 +1,1184 @@
+-- 2018-01-31T14:17:00.493
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+INSERT INTO AD_Table (AccessLevel,ACTriggerLength,AD_Client_ID,AD_Org_ID,AD_Table_ID,CopyColumnsFromTable,Created,CreatedBy,EntityType,ImportTable,IsActive,IsAutocomplete,IsChangeLog,IsDeleteable,IsDLM,IsHighVolume,IsSecurityEnabled,IsView,LoadSeq,Name,ReplicationType,TableName,Updated,UpdatedBy) VALUES ('3',0,0,0,540929,'N',TO_TIMESTAMP('2018-01-31 14:17:00','YYYY-MM-DD HH24:MI:SS'),100,'D','N','Y','N','N','Y','N','N','N','N',0,'Bussines Partner Credit Limit','L','C_BPartner_CreditLimit',TO_TIMESTAMP('2018-01-31 14:17:00','YYYY-MM-DD HH24:MI:SS'),100)
+;
```
metas-ts: if i see this correctly, you have `IsChangeLog=N` ..i recommend to have it =Y..this table is about a sensitive topic and even if this table is "append-only", it can't hurt

de.metas.adempiere.adempiere/base/src/main/java-gen/org/compiere/model/X_C_BPartner_CreditLimit.java
cristinamghita: Hm.. I don't think so, because the amount will be converted all the time, I think in the currency of the client. But need to check with Mark.

de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5484320_sys_gh3413_CreateTableAndWindo_C_BPartner_CreditLimit.sql
cristinamghita: Agree!

de.metas.adempiere.adempiere/base/src/main/java-gen/org/compiere/model/X_C_BPartner_CreditLimit.java
cristinamghita: Checking with Mark

de.metas.adempiere.adempiere/base/src/main/java-gen/org/compiere/model/X_C_BPartner_CreditLimit.java
cristinamghita: I added as a virtual column.

Issue #3426, opened 2018-01-31
Dunning Level is not set in invoice after generating dunning doc


metasnw:
### Is this a bug or feature request?
bug

### What is the current behavior?
if you create a dunning doc for an invoice the field c_dunning_level_id remains null

#### Which are the steps to reproduce?
1. follow http://docs.metasfresh.org/webui_collection/EN/Dunning_Run.html
1. open the invoice and check dunning level
1. NOK: its null

### What is the expected or desired behavior?
the field should be filled so people can see the dunning level right on the invoice
if not filled the sql to find is quite heavy:
http://docs.metasfresh.org/sql_collection/c_dunning.html

cristinamghita:
Fixed. See :
![image](https://user-images.githubusercontent.com/15378036/35635837-087c957e-06b8-11e8-9799-fb79435942de.png)

metas-ts:
cherry-picked to `release`

metas-dh:
Results of IT1
tested in mf15

* dunning level set in invoice after generating dunning doc (WebUI: in Advanced Edit): OK
* relation dunning - dunning candidate not working in webui (in Swing there is an extra tab for that)

cristinamghita:
SQL to update old data:
```
update C_Invoice
set C_DunningLevel_ID=cand.C_DunningLevel_ID
FROM (SELECT C_DunningLevel_ID, record_id, AD_Table_ID FROM C_Dunning_Candidate ) AS cand
WHERE C_Invoice.C_Invoice_ID = cand.record_id and cand.AD_Table_ID=318
AND C_Invoice.C_DunningLevel_ID is null;```

PR #3427, opened 2018-01-31
#3426 Set dunning level in invoice

cristinamghita:
#3426 Dunning Level is not set in invoice after generating dunning doc

### DIFF
--- a/de.metas.dunning/src/main/java/de/metas/dunning/invoice/model/validator/C_Dunning_Candidate.java
+++ b/de.metas.dunning/src/main/java/de/metas/dunning/invoice/model/validator/C_Dunning_Candidate.java
@@ -10,12 +10,12 @@ package de.metas.dunning.invoice.model.validator;
  * it under the terms of the GNU General Public License as
  * published by the Free Software Foundation, either version 2 of the
  * License, or (at your option) any later version.
- * 
+ *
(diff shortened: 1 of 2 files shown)


de.metas.dunning/src/main/java/de/metas/dunning/invoice/model/validator/C_Dunning_Candidate.java
```diff
@@ -43,18 +42,35 @@
...
+		if (candidate.getAD_Table_ID() != InterfaceWrapperHelper.getTableId(I_C_Invoice.class)
+				|| !candidate.isDunningDocProcessed())
+		{
+			return;
+		}
+
+		final Properties ctx = InterfaceWrapperHelper.getCtx(candidate);
+		final String trxName = InterfaceWrapperHelper.getTrxName(candidate);
+
+		final I_C_Invoice invoice = InterfaceWrapperHelper.create(ctx, candidate.getRecord_ID(), I_C_Invoice.class, trxName);
```
metas-ts: i recomment to instead use `InterfaceWrapperHelper.load() next time

cristinamghita: Thx!

PR #3605, opened 2018-03-03
add mvnw so that users don't need to install mvn to build from cmdline

metas-ts:


metas-rc:
Nothing to IT. Moving to done.

Issue #3624, opened 2018-03-06
Pharma: When Operation Code is 2, deactivate the product


cristinamghita:
### Is this a bug or feature request?
FR

metas-lc:
IT
imported a file with 3 lines: to add a product, update a product and deactivate a product
(for the update and deactivate i had to create new products with that value)
checked m_product widow => import worked fine OK

PR #3629, opened 2018-03-06
#3624 Pharma: When Operation Code is 2, deactivate the product 

cristinamghita:
#3624 Pharma: When Operation Code is 2, deactivate the product


de.metas.vertical.pharma/src/main/java/de/metas/impexp/product/PharmaProductImportProcess.java
```diff
@@ -94,26 +94,44 @@ protected void updateAndValidateImportRecords()
...
 		}
-		else
+		else if (!DEACTIVATE_OPERATION_CODE.equals(operationCode))
 		{
-			newProduct = false;
-			product = updateProduct(importRecord, existentProduct);
-		}
+			final I_M_Product product;
+			final boolean newProduct;
+			if (existentProduct == null || (existentProduct == null && importRecord.getM_Product_ID() <= 0))
```
metas-ts: it's not clear to me what this expression is about..
if i get it right, it can be simplified to 
`(existentProduct == null || importRecord.getM_Product_ID() <= 0)`
or is there a `!` missing somewhere?

also, why not assign the result of this boolean expression directly to `final boolean newProduct;`?

cristinamghita: @metas-ts no, is not the same;
I had ```(existentProduct == null || importRecord.getM_Product_ID() <= 0)```
And now I modified to ```(existentProduct == null || (existentProduct == null && importRecord.getM_Product_ID() <= 0))```.
The first condition does not verify correctly this case:
* The product exists in the db , but wasn't match yet in I_Pharma_Product
Bur thinking now, actually this should not be a problem. Because should have been already solved by this ```org.adempiere.impexp.product.MProductImportTableSqlUpdater.dbUpdateProductsByValue(String)```
So I'll check why that is not doing what suppose to do!

Issue #3634, opened 2018-03-07
Pick in manufacturing Order: Picking Slot Filter on selected BPartner from Shipment Schedule


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
In Pick Action in Manufacturing Order (on received HU), the possible Picking Slots are not filtered although a shipment schedule line is (pre-)selected.
#### Which are the steps to reproduce?
Open, try and see.
### What is the expected or desired behavior?
Pre-Filter the available Picking Slots based on Businesspartner from selected Shipment Schedule. If that is not filled yet, then show all available picking slots of disable recording of Picking Slot before selecting a Shipment Schedule.

metas-ts:
`AD_Process.Value='WEBUI_M_HU_Pick'`

metas-ts:
implemented in `2018-12`

metas-dh:
Results of IT1
tested in mf15

tested with a manual PP_Order, with the bpartner from sales order set:
* when no picking slot was selected for the shipment schedule already, then all free slots were selectable in manufacturing order, pick action: OK
* when a picking slot was already selected for the shipment schedule, only the free slots were selectable, but not the one for this shipment schedule already selected: NOK

PR #3646, opened 2018-03-07
#3634-app - Pick in manufacturing Order: Picking Slot Filter on selected BPartner from Shipment Schedule

metas-ts:
* add display logic for picking slot parameter
* also minor code changes (unrelated)

Pick in manufacturing Order: Picking Slot Filter on selected BPartner from Shipment Schedule #3634


metas-ts:
pls don't delete this branch!

PR #3681, opened 2018-03-13
#3414 Credit Limit check in Sales Order 

cristinamghita:
#3414 Credit Limit check in Sales Order


de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/bpartner/service/IBPartnerStatsDAO.java
```diff
@@ -83,21 +66,7 @@
 	 */
 	I_C_BPartner retrieveC_BPartner(BPartnerStats bpStats);
```
metas-ts: pls get rid on this I_C_Bpartner returning method..same as the other week:
* https://github.com/metasfresh/metasfresh/pull/3597#discussion_r171797165
* https://github.com/metasfresh/metasfresh/pull/3597#pullrequestreview-100712414

Issue #3674, opened 2018-03-12
Support Country in local address sequence


metasnw:
### Is this a bug or feature request?
FR

### What is the current behavior?
this local address sequence:
`@BP@ @CON@ @A1@ @A2@ @A3@ @A4@ @P@ @C@ @CO@`

does not show the local country in the rendered address although `@CO@` is defined

#### Which are the steps to reproduce?
1. create a sales order for a partner with the same country as the org you are creating the order
1. print preview
1. NOK: country i missing although @CO@is part of the local sequence

### What is the expected or desired behavior?
address renderer shall use explicit configuration of @CO@

cristinamghita:
### Dev test
* Create new sales order
* Check the address rendered in the new order
```
Test Kunde 1
teststraße 996
45678 teststadt 3
Deutschland
```

metas-ts:
:cherries: picked to `release`

metasnw:
IT was successful

PR #3685, opened 2018-03-13
#3674 Support Country in local address sequence 

cristinamghita:
#3674 Support Country in local address sequence


de.metas.adempiere.adempiere/base/src/main/java/de/metas/adempiere/service/impl/AddressBuilder.java
```diff
@@ -247,29 +249,37 @@ else if (outStr.lastIndexOf("\n") == outStr.length() - 1 && text.trim().length()
...
 				final String add = location.getPostal_Add();
 				if (add != null && add.length() > 0)
+				{
 					outStr.append("-").append(add);
+				}
 			}
 			else if (token.equals("CO"))
 			{
 				final String countryName;
 				if (isLocalAddress)
```
metas-ts: why don't we translate the country name if `isLocalAddress==true`?

cristinamghita: @metas-ts Thinking now, I think is a bit more complicated.
Because we have client language and the system base language.
System base language is German, so actually in C_Country, the names are in German for all countries.
So you need to take the name from C_Country if local is de_DE, and from translation, if otherwise.
Correct?

Issue #3669, opened 2018-03-12
Create all periods of a contract on creation


metasnw:
### Is this a bug or feature request?
FR

### What is the current behavior?
when creating a new contract via sales order only the first period is created
when you have multiple periods chained you can not
1. invoice all periods at once (e.g. 3 years in advance)
1. see the masterenddate 

#### Which are the steps to reproduce?

|transition| automatic extension
|---|---|
|1 year then 2nd year | yes
|2nd year then end| no

1. create a sales order using that condition
1. NOK: only the first period is created
1. NOK: masterenddate is null although it should be the enddate of the second transition

### What is the expected or desired behavior?
1. have a new flag in condition "create all periods"
  1. if yes, create the whole chain right when sales order is completed
1. make sure all periods inherit the masterenddate from the last period

metas-ts:
hi @metasnw 
as a remark, we should try to develop a common language.
i'm currently standing here:
* period = `C_Period`
* contract = depends on context, but usually it's `C_Flatrate_Term`
you stand (afaiu) here
* period = `C_Flatrate_Term`
* contract = ?

Also, given our skype-talk about this matter, I'm not sure i understand why you expect this behavior you outlined?

In particular afaiu, the flag's correct label would be "Create all periods, if the configured chain of flatrate transitions indicates a finite number of periods"..i see a few problems with that, both in terms of (stable) implementation and also ease of use.

Why not implement it so that we can have one 3-year C_Flatrate_Term with a yearly invoicing-calendar?

metasnw:
> Why not implement it so that we can have one 3-year C_Flatrate_Term with a yearly invoicing-calendar?

because we might have price and qty changes for each year and need to document them for invoicing

metasnw:
tested on branch system:
chained 3x conditions and transitions
set autoextension on year 1 and year 2

creating new contract:
1. OK: all 3x periods where created right away
1. OK: masterstart and enddate is set on all periods

terminating contract:
1. terminated contract with tomorrows date
1. OK: all periods are terminated

PR #3710, opened 2018-03-16
#3669 Create all periods of a contract on creation 

cristinamghita:
#3669 Create all periods of a contract on creation


cristinamghita:
Will not be an infinite number of C_Flatrate_Terms, because, if the transition does not have next ft and extension is not forced, nothing happens.
The difference between auto renew and auto extension is:
* auto renew - will renew automatically the contract when the notice date is due
* auto extension means that will extend the contract from the begining, if the transition is set like that

Issue #3718, opened 2018-03-19
Contracts: Transform IsAutoRenew and IsAutoExtension flags in a list


cristinamghita:
### Is this a bug or feature request?
FR

cristinamghita:
Follow up [#3669 Create all periods of a contract on creation](https://github.com/metasfresh/metasfresh/issues/3669)

metas-ts:
:cherries: picked to `release`

metas-rc:
Results of IT

Tested on webui


* OK: In the window Contract Periods, the field Extension Type is now list.

1) If the Extention type is for all periods

* I created 3 Contract Periods pointing to 3 contract terms, 1 year each, first 2 periods have next contract perms.
* I set their extention type to all periods
* I made a sales order and I selected the first, completed
* => OK : I got 3 contracts created for the next 3 years

2) If the Extention type is for Autorenewal

* I created a contract period pointing to a contract term, with Extention type for AutoRenew, 1 month
* I made a sales order and I selected the term, completed
* => OK: I got 1 contract
* => OK: The contract was flagged as IsAutoRenew
* => OK: The master end date was not set
* I extended the contract
* => OK: Got 1 new 1 month contract
* => OK: Both have master end date null
* Canceled the contract with a date (May 1 2018)
* => OK; Master end date mecame May 1 for both of them
* => OK: IsAutoRenew flag turned to false

3) Extention type not set
 * I created a contract period pointing to a contract term, with Extention type null, 1 month.
* I made a sales order and I selected the term, completed
* => OK: I got 1 contract
* => OK: Contract has IsAutoRenew = false
* => OK: Master end date 1 month since now
* I manually extended the contract 
* => OK: I got a new contract
* => OK: Masterenddate for both turned to 2 months from now

PR #3731, opened 2018-03-20
#3718 Contracts: Transform IsAutoRenew and IsAutoExtension flags in a list 

cristinamghita:
#3718 Contracts: Transform IsAutoRenew and IsAutoExtension flags in a list


de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
```diff
@@ -1012,31 +1013,64 @@ else if (diffPercent.signum() < 0 && diffPercent.abs().compareTo(conditions.getM
...
+				final I_C_Flatrate_Term nextTerm = currentTerm.getC_FlatrateTerm_Next();
+				final I_C_Flatrate_Conditions nextConditions = nextTerm.getC_Flatrate_Conditions();
+				Check.assumeNotNull(nextConditions, "C_Flatrate_Conditions shall not be null!");
+
+				nextTransition = nextConditions.getC_Flatrate_Transition();
+				Check.assumeNotNull(nextTransition, "C_Flatrate_Transition shall not be null!");
+
+				contextUsed = ContractExtendingRequest.builder()
 						.AD_PInstance_ID(context.getAD_PInstance_ID())
 						.contract(nextTerm)
```
metas-ts: if you annotate ContractExtendingRequest with `@Builder(toBuilder=true), you should be able to do 
```java
contextUsed = contextUsed.toBuilder()
  .AD_PInstance_ID(context.getAD_PInstance_ID())
  .contract(nextTerm).build();
```
and avoid the properties that don't change

de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
```diff
@@ -1012,31 +1013,64 @@ else if (diffPercent.signum() < 0 && diffPercent.abs().compareTo(conditions.getM
...
 						.forceComplete(true)
 						.nextTermStartDate(null)
 						.build();
-				extendContract(nextContext);
+				contracts.add(nextTerm);
 			}
+			while (X_C_Flatrate_Transition.EXTENSIONTYPE_ExtendAll.equals(nextTransition.getExtensionType()) && nextTransition.getC_Flatrate_Conditions_Next_ID() > 0);
 
+			if (X_C_Flatrate_Transition.EXTENSIONTYPE_ExtendAll.equals(nextTransition.getExtensionType()))
+			{
```
metas-ts: why do we need this if?

de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
```diff
@@ -1012,31 +1013,64 @@ else if (diffPercent.signum() < 0 && diffPercent.abs().compareTo(conditions.getM
...
+	private void updateMasterEndDate(final List<I_C_Flatrate_Term> contracts)
+	{
+		// update master end date
+		final Timestamp endDate = contracts.stream()
+				.sorted(Comparator.comparing(I_C_Flatrate_Term::getC_Flatrate_Term_ID).reversed())
+				.findFirst()
+				.map(I_C_Flatrate_Term::getEndDate)
+				.orElse(null);
+
+		if (endDate != null)
```
metas-ts: i suggest to reduce the nesting depth by returning if `endDate==null`

de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
```diff
@@ -1012,31 +1013,64 @@ else if (diffPercent.signum() < 0 && diffPercent.abs().compareTo(conditions.getM
...
+		{
+			contracts
+					.stream()
+					.filter(contract -> contract.getMasterEndDate() == null)
```
metas-ts: why discard contracts which have a MasterEndDate? couldn't it be an outdated one? what do we loose if we remove this filter?

de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
cristinamghita: Nice!
I did not know

de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
cristinamghita: Because masterendate shall not be filled if is annual renew.
In that case, we don't know when the contract is ending.

de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
cristinamghita: thinking now, I think we actually don;t need at all

de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
cristinamghita: hm.. good point!
all should have same date anyway

de.metas.contracts/src/main/sql/postgresql/system/50-de.metas.contracts/5488820_sys_gh3718_AddExtensonTypeList.sql
```diff
@@ -0,0 +1,144 @@
...
+;
+
+-- 2018-03-19T17:36:43.637
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+INSERT INTO AD_Ref_List_Trl (AD_Language,AD_Ref_List_ID, Description,Name, IsTranslated,AD_Client_ID,AD_Org_ID,Created,Createdby,Updated,UpdatedBy) SELECT l.AD_Language,t.AD_Ref_List_ID, t.Description,t.Name, 'N',t.AD_Client_ID,t.AD_Org_ID,t.Created,t.Createdby,t.Updated,t.UpdatedBy FROM AD_Language l, AD_Ref_List t WHERE l.IsActive='Y' AND l.IsSystemLanguage='Y' AND l.IsBaseLanguage='N' AND t.AD_Ref_List_ID=541602 AND NOT EXISTS (SELECT 1 FROM AD_Ref_List_Trl tt WHERE tt.AD_Language=l.AD_Language AND tt.AD_Ref_List_ID=t.AD_Ref_List_ID)
+;
+
+-- 2018-03-19T17:37:27.314
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+INSERT INTO AD_Ref_List (AD_Client_ID,AD_Org_ID,AD_Ref_List_ID,AD_Reference_ID,Created,CreatedBy,EntityType,IsActive,Name,Updated,UpdatedBy,Value,ValueName) VALUES (0,0,541603,540843,TO_TIMESTAMP('2018-03-19 17:37:27','YYYY-MM-DD HH24:MI:SS'),100,'de.metas.contracts','Y','Extend contract for first period',TO_TIMESTAMP('2018-03-19 17:37:27','YYYY-MM-DD HH24:MI:SS'),100,'EO','ExtendOne')
```
metas-ts: pls change the description to something like 'Extend the contract when the notice date is reached'
..because afaiu that's what it does, right?

de.metas.contracts/src/main/sql/postgresql/system/50-de.metas.contracts/5488820_sys_gh3718_AddExtensonTypeList.sql
```diff
@@ -0,0 +1,144 @@
...
+
+-- 2018-03-19T17:27:44.493
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+INSERT INTO AD_Reference_Trl (AD_Language,AD_Reference_ID, Description,Help,Name, IsTranslated,AD_Client_ID,AD_Org_ID,Created,Createdby,Updated,UpdatedBy) SELECT l.AD_Language,t.AD_Reference_ID, t.Description,t.Help,t.Name, 'N',t.AD_Client_ID,t.AD_Org_ID,t.Created,t.Createdby,t.Updated,t.UpdatedBy FROM AD_Language l, AD_Reference t WHERE l.IsActive='Y' AND l.IsSystemLanguage='Y' AND l.IsBaseLanguage='N' AND t.AD_Reference_ID=540843 AND NOT EXISTS (SELECT 1 FROM AD_Reference_Trl tt WHERE tt.AD_Language=l.AD_Language AND tt.AD_Reference_ID=t.AD_Reference_ID)
+;
+
+-- 2018-03-19T17:36:43.631
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+INSERT INTO AD_Ref_List (AD_Client_ID,AD_Org_ID,AD_Ref_List_ID,AD_Reference_ID,Created,CreatedBy,EntityType,IsActive,Name,Updated,UpdatedBy,Value,ValueName) VALUES (0,0,541602,540843,TO_TIMESTAMP('2018-03-19 17:36:43','YYYY-MM-DD HH24:MI:SS'),100,'de.metas.contracts','Y','Extend contract for all periods',TO_TIMESTAMP('2018-03-19 17:36:43','YYYY-MM-DD HH24:MI:SS'),100,'EA','ExtendAll')
+;
```
metas-ts: pls describe better what it does
* that it's done on complete (right?)
* that there is a chain of "extends", until a transition record without this falg and a next conditions reference is reached

PR #3779, opened 2018-03-28
#3354 Different subscription Receiver for ever 

cristinamghita:
#3354 Different subscription Receiver for ever


de.metas.contracts/src/main/java/de/metas/contracts/subscription/impl/SubscriptionService.java
```diff
@@ -89,16 +86,23 @@ public static int changeRecipient(@NonNull final ChangeRecipientsRequest changeR
...
 		{
 			this.term = term;
 
-			Preconditions.checkArgument(!dateTo.before(dateFrom), "The given dateTo may not be before the given dateFrom; dateFrom=%s; dateTo=%s", dateFrom, dateTo);
+			this.IsPermanentRecipient = IsPermanentRecipient;
+
+			if (!IsPermanentRecipient)
+			{
+				Preconditions.checkArgument(!dateTo.before(dateFrom), "The given dateTo may not be before the given dateFrom; dateFrom=%s; dateTo=%s", dateFrom, dateTo);
+				Preconditions.checkNotNull(dateFrom, "The dateForm shall not be null");
```
metas-ts: better do the null-check before the before-check :-)

cristinamghita: pfu.. right! :)))

PR #3782, opened 2018-03-28
#3580 picking label process report

metas-rc:
https://github.com/metasfresh/metasfresh/issues/3580


de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/report/HUReportService.java
```diff
@@ -274,4 +274,53 @@ else if (X_M_HU_PI_Version.HU_UNITTYPE_LoadLogistiqueUnit.equals(huUnitType))
 			}
 		}
 	}
+
+	public void printPickingLabel(final HUToReportWrapper huToReport, final boolean isAutoPrintRequired)
+	{
+		if (huToReport == null)
+		{
+			logger.error("Param 'hu'==null; nothing to do");
```
metas-ts: i'm not sure if this should be logged as `error`
...afaiu, if this method is called from `WEBUI_Picking_PickQtyToNewHU` and  `!isPickingLabelAutoPrintEnabled()`, then there shouldn't be and error message in our log...that's just one example (one of theses `if`s), but i suspect it's similar with the others..

On a related note, if i look at `WEBUI_Picking_TU_Label.checkPreconditionsApplicable()`, i see a few similarities..
maybe you can extract into a method within *this* class. that method might return a `ProcessPreconditionsResolution` or maybe an `IPair<Boolean, ITranslatable>` with the result and an optional `rejection-AD_Message`.

goal: make sure tht the logic about whether a TU-label can/shall be printed is centralized as much as possible.

metas-rc: I will make the log entries WARN. SO it would also be easy for to debug if there is something poorly configured .

Issue #3748, opened 2018-03-22
Pharma:  Regular import does not fail if prices aren't imported


cristinamghita:
### Is this a bug or feature request?
bug

metas-lc:
imported file before setting pricelists
=> file not imported:  ProductPriceImporter.InvalidProductPriceList,  OK
imported file after setting pricelists
=> file was imported OK

PR #3785, opened 2018-03-28
#3748 Pharma: Regular import does not fail if prices aren't imported 

cristinamghita:
#3748 Pharma: Regular import does not fail if prices aren't imported

### DIFF
--- a/de.metas.business/src/main/java/org/adempiere/impexp/product/ProductPriceImporter.java
+++ b/de.metas.business/src/main/java/org/adempiere/impexp/product/ProductPriceImporter.java
@@ -9,6 +9,7 @@ import static org.adempiere.model.InterfaceWrapperHelper.save;
 import java.math.BigDecimal;
 import java.time.LocalDate;
 
+import org.adempiere.exceptions.AdempiereException;
 import org.adempiere.model.InterfaceWrapperHelper;
 import org.adempiere.pricing.api.IPriceListDAO;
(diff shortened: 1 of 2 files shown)


de.metas.business/src/main/java/org/adempiere/impexp/product/ProductPriceImporter.java
```diff
@@ -61,17 +62,19 @@ public void createProductPrice_And_PriceListVersionIfNeeded()
 	{
 		if (!isValidPriceRecord())
 		{
-			return;
+			throw new AdempiereException("ProductPriceImporter.InvalidProductPriceList");
 		}
 
-		final I_M_PriceList_Version plv = getCreatePriceListVersion(request.getPriceListId(), request.getValidDate());
-		createProductPriceOrUpdateExistentOne(plv);
+		if (request.getPriceListId() > 0)
```
metas-ts: i think that if we arrived here, if means that 
* `isValidPriceRecord()`  returned true which means that
* `request.getPriceListId() > 0`

so in summary, i don't understand what you are doing here :blush: ..or maybe you removed the wrong line in `isValidPriceRecord()` ?

cristinamghita: Yes, I removed the wrong line and then reconsidered..
This should be : request.getPrice().signum() > 0
Fixed!

Issue #3892, opened 2018-04-19
Contracts: Allow possibility to copy the prices when extending


cristinamghita:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?

case:

test setup:
-priceliste versions
-01.01.2016 = Price 1
-01.01.2017 = Price 2

compensation group adding discount
flatrate condition has autoextend immediately
execution:

Order
-datepromised: 01.06.2016
product A with Price 1 and 3 years contract condition
create compensation group adding discount
line #2: product B with Price -1 (discount)
flatrate term
NOK: product A has on second term price=2 (should have price=1)
NOK: product B has Price 0 but should remain -1

cristinamghita:
## Dev tests
### Test - Don't compute prices
* Create a new chain of flat rate condition with flag IsCalculatePrice on N.
   * this means that the prices for the next flat rate term will be copied from the previous one
    ![image](https://user-images.githubusercontent.com/15378036/39004828-21d5246e-4407-11e8-9612-9b330251ae07.png)
* Create a sales order using first condition (and having a compensation group)
   ![image](https://user-images.githubusercontent.com/15378036/39005001-8eb880f8-4407-11e8-946f-7454715a5d3c.png)
* Complete the order and check created contracts
  ![image](https://user-images.githubusercontent.com/15378036/39004554-498771ac-4406-11e8-9863-2b53ff5fc334.png)

metasnw:
IT of the requirement was successfull:
1. OK: price is recalculated like before when option is off
1. OK: price is remaining the same on both discount and product  when the option is on

@cristinamghita any regression tests you recommend?

cristinamghita:
@metasnw yes.
Usual case. I described in my tests .

metas-dh:
Regression testing: all worked fine, closing this issue.

PR #3893, opened 2018-04-19
#3892 Contracts: Allow possibility to copy the prices when extending 

cristinamghita:
#3892 Contracts: Allow possibility to copy the prices when extending


de.metas.contracts/src/main/sql/postgresql/system/50-de.metas.contracts/5491360_sys_gh3892_AddIsCalculatePrice.sql
```diff
@@ -0,0 +1,42 @@
+-- 2018-04-19T18:01:52.240
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+INSERT INTO AD_Element (AD_Client_ID,AD_Element_ID,AD_Org_ID,ColumnName,Created,CreatedBy,Description,EntityType,Help,IsActive,Name,PrintName,Updated,UpdatedBy) VALUES (0,543964,0,'IsCalculatePrice',TO_TIMESTAMP('2018-04-19 18:01:51','YYYY-MM-DD HH24:MI:SS'),100,'','de.metas.contracts','','Y','Calculate Price','Calculate Price',TO_TIMESTAMP('2018-04-19 18:01:51','YYYY-MM-DD HH24:MI:SS'),100)
```
metas-ts: a Y/N column `IsCalculatePrice` and the text "CAlculate Price" doesn't give the user enough info.
E.g. what should a user expect if  they set IsCalculatePrice=N ? a prompt to set the price manually? everything fille with zeros?
Adding more documentation would be one solution, but i'm not sure how the user can access it.
so omho the best solution is to have a list with two options:
* column/field onFlatrateTermExtend (or similar, please alligne with existing column names)
* values: "copy prices from predecessor" and "calculate from pricing data"

metas-ts: btw, please chag you the test failure

metas-ts: Another thing: do we *really* want to have this option in C_Flatrate_Condition?
The issue sais "Contracts: Allow possibility to copy the prices when extending"
* At which stage shall this be allowed (C_Flatrate_Condition is a rather early stage), and 
* in which cases shall the pricese be calculated and in which cases shall they be copied (with this implementation, they are *always* calculated or *always* copied)

metasnw: results from our talk:
1. Checbox => List box with explicit names
1. Fix tests
3. we keep the option in flatrate conditions

Issue #3889, opened 2018-04-19
Report for Pharma Permission/ Authorisation control


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, there is no report available for Pharma Permission/ Authorisation control.
#### Which are the steps to reproduce?
Open, try and see.
### What is the expected or desired behavior?
Create a Jasper Report for Pharma Permission/ Authorisation control.
Parms (optional):
- Business Partner
- Manufacturing Permission
- Wholesale Permission
- Agent
- Pharmacy Permission
- Veterinary Permission

Columns:
- Business Partner Name
- Business Partner Value
- Manufacturing Permission
- Wholesale Permission
- Agent
- Pharmacy Permission
- Veterinary Permission
- Authorisation Date for "Type A"

- 5 Fields in customer w/ Shipment Auth + Date
- 3 Fields in Vendor w/ Receipt Auth + Date

metas-lc:
Liefer-/Bezugsberechtigung Kontrolle
Pharma Permission/ Authorisation control 
Report contains:
title and parameters,
columns: name, name2, bp value, customer/vendor, manufacturer perm, wholesale perm, agent perm, pharmacy perm, veterinary perm, shipment/receipt perm, change date

metas-lc:
taking it back to fix sql

metas-lc:
see how those columns are set: https://github.com/metasfresh/metasfresh/issues/3890

metas-dh:
Results of IT1
tested in mf15

* de_DE version: Arzneiverm. / Arzneim.vermittler: take out .../, not needed
* date: shown in 2 lines, show only in 1 line: NOK
* sorting?

![2018-04-25 13_46_38-liefer-bezugsberechtigung kontrolle pdf](https://user-images.githubusercontent.com/18696715/39245014-d4276b10-4892-11e8-8d41-e5ef95b2aee9.png)

* parameters in process: 
     * de_DE: Pharmacie Permission, NOK
     * en_US: same as de_DE, NOK

![2018-04-25 14_07_51-metasfresh](https://user-images.githubusercontent.com/18696715/39245027-ddf3ab2c-4892-11e8-8e62-917e8cb26a63.png)

metas-lc:
fixed: 
date on 2 lines
sorting (date desc - nulls last, name asc)
rename Arzneiverm
added missing trls

metas-dh:
Results of IT2
tested in mf15

[...]

PR #3924, opened 2018-04-24
#3889 Report for Pharma Permission/ Authorisation control 

metas-lc:
#3889


de.metas.vertical.pharma/src/main/sql/postgresql/ddl/pharma_permission_control_report.sql
```diff
@@ -50,7 +52,8 @@ SELECT
 	isCustomer, isVendor, unionType,
 	--parameters
 	(SELECT name FROM C_BPartner WHERE C_BPartner_ID = $1) AS param_bp_name,
-	$2 AS param_manufacturer, $3 AS param_wholesale, $4 AS param_agent, $5 AS param_pharmacie, $6 AS param_veterinary
+	$2 AS param_customer,
```
metas-ts: when touching code, please look for ways to improve it..according to teo, we can use parameter names instead of `$n` as of postgresql 9.2 or something..

Issue #4018, opened 2018-05-07
QtyCU not respected in Quarantine DD_Orderline


metas-rc:
Followup for https://github.com/metasfresh/metasfresh/issues/4012
### Is this a bug or feature request?
* Bug
### What is the current behavior?
* When a ddOrderLine is created for a quarantine HU that only consists of a CU, the qty CU is not respected.
#### Which are the steps to reproduce?
* Purchase order with a quarantine product
* Go to material receipt candidates, run Receive CU with Qty, put there the qty 1,
* Set the quarantine lot number
*  create material receipt.
* go to the created distribution order
* => NOK: The qty TU is 1 and the qty CU is 10.
### What is the expected or desired behavior?
* The qty CU shall only pe 1.

metas-rc:
Update: we have a model interceptor that is triggered after the dd_orderline is created and if it lacks an M_HU_PI_Item_Product, it searches the database and if it finds a TU for the product, it sets the packing info in the line for that TU.
I will add a sys config to allow this behaviour in the dd_orderline only if that sys config is on Y.

metas-rc:
I added a new sys config, named de.metas.handlingunits.DD_OrderLine_Enforce_M_HU_PI_Item_Product.

If this sys config is set on 'Y',  the DD_OrderLine will always have an M_HU_PI_Item_Product if fitting Packing Instructions are found in the db, even if the DD_Orderline was initially for CUs only. 
This M_HU_PI_Item_Product setting leads to also setting a qtyTU and the qtyCU will change to the qty that fitts those instructions (example: if a box fits 10 CUs, the qtyCU of the DD_OrderLine will be 10, no matter what it was before).

Sys config =  'N'  means that this functionality is not needed, and the DD_OrderLine will remain as it was created ( e.g. based on an inoutLine for CUs only)

metas-lc:
IT
Sys config = 'N'
1. make sure you have a lot lock, create/complete a purchase order 
2. create receipt, with lotnumber like you have in lotno, only CUs (e.g. 3)
=> dd order was created correctly, with CU only, correct qty (e.g. 3) OK
3. create another receipt, add the product on TU (e.g. 3 TUs)
=> dd order was created correctly, with product on TU, correct qty (e.g. 30 cu 3 TU) OK
4. create another receipt, add the product on TU and LU (e.g. 3 LUs)
=> dd order was created correctly, with product on TU, correct qty (e.g. 30 cu 3 TU) OK
.
[...]

PR #4051, opened 2018-05-14
Gh4018 option to not enforce M_HU_PI_Item_Product in DD_OrderLine

metas-rc:
https://github.com/metasfresh/metasfresh/issues/4018

### DIFF
--- a/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/ddorder/spi/impl/DDOrderLineHUDocumentHandler.java
+++ b/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/ddorder/spi/impl/DDOrderLineHUDocumentHandler.java
@@ -24,6 +24,7 @@ package de.metas.handlingunits.ddorder.spi.impl;
 
 
 import org.adempiere.model.InterfaceWrapperHelper;
+import org.adempiere.service.ISysConfigBL;
 import org.adempiere.util.Check;
 import org.adempiere.util.Services;
 import org.compiere.model.I_M_Product;
@@ -42,6 +43,7 @@ import de.metas.handlingunits.model.X_M_HU_PI_Version;
  */
 public class DDOrderLineHUDocumentHandler implements IHUDocumentHandler
 {
+	private final static String SYSCONFIG_DD_OrderLine_Enforce_M_HU_PI_Item_Product = "DD_OrderLine_Enforce_M_HU_PI_Item_Product";
 	/**
 	 * This method assumes that the given document is <code>DD_OrderLine</code> and returns
 	 * <ul>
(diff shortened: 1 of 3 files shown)


de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/ddorder/spi/impl/DDOrderLineHUDocumentHandler.java
```diff
@@ -42,6 +43,7 @@
  */
 public class DDOrderLineHUDocumentHandler implements IHUDocumentHandler
 {
+	private final static String SYSCONFIG_DD_OrderLine_Enforce_M_HU_PI_Item_Product = "DD_OrderLine_Enforce_M_HU_PI_Item_Product";
```
metas-ts: please prepend de.metas.handlingunits

de.metas.handlingunits.base/src/main/sql/postgresql/system/5493530_sys_gh4018_SYSConfig_Enforce_DDOrderline_M_HU_PI_Item_Product.sql
```diff
@@ -0,0 +1,5 @@
+-- 2018-05-14T16:22:03.594
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+INSERT INTO AD_SysConfig (AD_Client_ID,AD_Org_ID,AD_SysConfig_ID,ConfigurationLevel,Created,CreatedBy,EntityType,IsActive,Name,Updated,UpdatedBy,Value) VALUES (0,0,541210,'S',TO_TIMESTAMP('2018-05-14 16:22:03','YYYY-MM-DD HH24:MI:SS'),100,'de.metas.handlingunits','Y','DD_OrderLine_Enforce_M_HU_PI_Item_Product',TO_TIMESTAMP('2018-05-14 16:22:03','YYYY-MM-DD HH24:MI:SS'),100,'-')
```
metas-ts: * why did you add the `-` instead of Y or N? 
* please add some documentation about what the sysconfig's implications are. Both to this `AD_SysConfig` record and the issue. then flag the issue as `type:documentation`. In particular, please document how *this* sysconfig affects the actual issue around the CuQty

de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/ddorder/spi/impl/DDOrderLineHUDocumentHandler.java
metas-rc: done

metas-rc:
I added a new sys config, named de.metas.handlingunits.DD_OrderLine_Enforce_M_HU_PI_Item_Product.

If this sys config is set on 'Y',  the DD_OrderLine will always have an M_HU_PI_Item_Product if fitting Packing Instructions are found in the db, even if the DD_Orderline was initially for CUs only. 
This M_HU_PI_Item_Product setting leads to also setting a qtyTU and the qtyCU will change to the qty that fitts those instructions (example: if a box fits 10 CUs, the qtyCU of the DD_OrderLine will be 10, no matter what it was before).

Sys config =  'N'  means that this functionality is not needed, and the DD_OrderLine will remain as it was created ( e.g. based on an inoutLine for CUs only)

de.metas.handlingunits.base/src/main/sql/postgresql/system/5493530_sys_gh4018_SYSConfig_Enforce_DDOrderline_M_HU_PI_Item_Product.sql
metas-rc: Done

Issue #4058, opened 2018-05-16
Add delivery times to Der Kurier


metas-ts:
### Is this a bug or feature request?
FR

* Please also add DK_DesiredDeliveryTime_From and DK_DesiredDeliveryTime_To to the derKurier-konfig;
similar to issue Add CustomerCode and CollectorCode to Der Kurier https://github.com/metasfresh/metasfresh/issues/4034
* please update existing configs (that don't yet have this value set) to 
  * from: 09:00
  * to: 17:00

metas-lc:
IT
1. checked shipper window, der kurier konfig. tab
=> Desired Delivery Time From: 9:00 AM
Desired Delivery Time To: 5:00 PM
OK
2. created new der kurier order, check orderline
=> Consignee Delivery Time From 9:00 AM
Consignee Delivery Time To 5:00 PM
OK

PR #4069, opened 2018-05-17
Gh4058 Der Kurier Shipper Config: Desired Time From and TO

metas-rc:
https://github.com/metasfresh/metasfresh/issues/4058


de.metas.shipper.gateway.derkurier/src/main/java/de/metas/shipper/gateway/derkurier/DerKurierDeliveryData.java
```diff
@@ -43,8 +45,12 @@ public static DerKurierDeliveryData ofDeliveryPosition(@NonNull final DeliveryPo
 	String customerNumber;
 
 	String parcelNumber;
-	
+
 	String customerCode;
-	
+
 	String collectorCode;
+
```
metas-ts: * please prefer the new java time API to timestamp (LocalDateTime etc)
* we need so specify times without date. E.g. whenever something is delivered, it needs to be between 9:00 and 17:00
  * suggest to take a look at `LocalTime`

metas-rc: Done.
Note that I had to also refactor some tests and I added some code to the TimeUtil.

Issue #4035, opened 2018-05-09
Proforma Invoice via Sales Order workflow


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, the Prepayment Functionality in Sales Order is deactivated.
#### Which are the steps to reproduce?
Open, try and see.
### What is the expected or desired behavior?
Reactivate and adapt the prepayment functionality via sales order.
- User creates a sales order with doctype prepayment order.
- The sales order is completed. No Shipment schedules nor Invoice candidates are created. The sales order document (Proforma Invoice) is sent to the customer for prepayment.
- The customer pays the proforma invoice. The Payment is recorded manually in Payment window.
- The shipment candidates are created. The invoice candidates are created.
- As soon as the invoice is generated, the invoice is marked as paid (because of prepayment allocation).

cristinamghita:
## Dev test
* Create a prepay order
   * Complete the order -> Status is _Waiting for payment_
    ![image](https://user-images.githubusercontent.com/15378036/40305230-4ed82b96-5d02-11e8-9d0c-8ca87c40d7dd.png)
   * No Shipment schedules nor Invoice candidates are created. 
* Now user creates a Payment is recorded manually in Payment window and completes the payment.
   *  No allocations is created, but the shipment candidates are created. The invoice candidates are created.
   ![image](https://user-images.githubusercontent.com/15378036/40305276-7b756d08-5d02-11e8-9d59-363d79fc78f5.png)
* User goes to the shipment schedule and creates the shipment
   ![image](https://user-images.githubusercontent.com/15378036/40305344-d24d2558-5d02-11e8-9d6f-d453ad613d7b.png)
* Then goes in invoice candidates window and generate the invoice.
   *  When the invoice is completed, the payment si allocated against this invoice and the invoice is marked as paid.
   ![image](https://user-images.githubusercontent.com/15378036/40305394-155dac14-5d03-11e8-971e-e7c47289ce54.png)


![image](https://user-images.githubusercontent.com/15378036/40305390-0fec618a-5d03-11e8-9b5b-51458e146c40.png)

metas-dh:
**Update:**

**questions:**
1. => new doctype is created, "Proforma Invoice": **OK**
[...]

PR #4077, opened 2018-05-21
#4035 Proforma Invoice via Sales Order workflow 

cristinamghita:
#4035 Proforma Invoice via Sales Order workflow


de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoice/model/validator/C_Invoice.java
```diff
@@ -254,4 +257,29 @@ private void deleteInvoiceLines(final I_C_Invoice invoice)
 			InterfaceWrapperHelper.delete(line);
 		}
 	}
+
+	@DocValidate(timings = { ModelValidator.TIMING_BEFORE_COMPLETE })
+	public void linkInvoiceToPaymentIfNeeded(final I_C_Invoice invoice)
+	{
+		final I_C_Order order = invoice.getC_Order();
```
metas-ts: * what if there are >orders and therefore `invoice.getC_Order_ID() <= 0`? 
* what if there are >1 invoices for one order and therefore only the order's paid amount needs to be distributed among them?

Talked with @metas-mk :
> what if there are >orders and therefore `invoice.getC_Order_ID() <= 0`?
* for a prepay order there may be just one invoice. so we need to make sure that even if the customer has "Sammelrechnung erlaubt", the system still needs to use the "no-aggregation" `C_Aggegration` for the prepay order's ICs

> what if there are >1 invoices for one order and therefore only the order's paid amount needs to be distributed among them?

* AFAIR long time ago we created `C_AllocationLine.C_Order_ID` for the prepay order use case.. i think it wasn't a great idea; can you please take a look and remove `C_AllocationLine.C_Order_ID` if it's still there?

de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoice/model/validator/C_Invoice.java
```diff
@@ -254,4 +257,29 @@ private void deleteInvoiceLines(final I_C_Invoice invoice)
...
+			InterfaceWrapperHelper.save(payment);
+
+			Services.get(IAllocationBL.class).autoAllocateSpecificPayment(invoice, payment, true);
+		}
+	}
+
+	@DocValidate(timings = { ModelValidator.TIMING_BEFORE_COMPLETE })
+	public void allocateInvoiceAgainstPaymentIfNeeded(final I_C_Invoice invoice)
+	{
+		final I_C_Order order = invoice.getC_Order();
```
metas-ts: note sure I get it right, but this method seems to be the same as `linkInvoiceToPaymentIfNeeded()` minus the `payment.setC_Invoice_ID(...);` part..why do we need it?

cristinamghita: Yes, right! wrong moment; the second should be TIMING_AFTER_COMPLETE

PR #4090, opened 2018-05-23
Gh3463 #3463 issue CUs

metas-rc:
https://github.com/metasfresh/metasfresh/issues/3463


de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/allocation/transfer/HUTransformService.java
```diff
@@ -986,4 +989,133 @@ public IMutableAllocationResult process(final IHUContext localHuContext)
...
+
+	private List<I_M_HU> luExtractCUs(@NonNull final I_M_HU sourceLU, final Quantity qtyCU)
+	{
+		Preconditions.checkArgument(qtyCU.signum() > 0, "qtyCU > 0");
+
+		final ImmutableList.Builder<I_M_HU> extractedCUs = new ImmutableList.Builder<>();
+
+		Quantity qtyCUsRemaining = qtyCU; // how many CUs we still have to extract
+
+		for (int i = 0; i < handlingUnitsDAO.retrieveIncludedHUs(sourceLU).size(); i++)
```
metas-ts: the `handlingUnitsDAO.retrieveIncludedHUs(sourceLU).size()` is eavluated on each iteration of the for loop
* performance
* i think due to the `luExtractTUs` that we have within the loop, it will be decreased by one each time.

de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/allocation/transfer/HUTransformService.java
```diff
@@ -986,4 +989,133 @@ public IMutableAllocationResult process(final IHUContext localHuContext)
...
+		{
+			if (qtyCUsRemaining.signum() <= 0)
+			{
+				break;
+			}
+
+			final List<I_M_HU> extractedTUs = luExtractTUs(sourceLU, 1);
```
metas-ts: By @teosarca :
>  i would expect that
`final List<I_M_HU> extractedTUs = luExtractTUs(sourceLU, 1);`
> to return ONLY one element
in that case i would use something like ListUtils.singleElement
to enforce that
and i would avoid the 
`for (final I_M_HU tu : extractedTUs)`

I wonder`when you might get >1 TUs

de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/allocation/transfer/HUTransformService.java
metas-rc: You're right. Fixed by extracting in in a variable beforehand.

Issue #4123, opened 2018-05-28
Dunning Jasper does not work


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Dunning Jasper does not work, calling the action throws an error.
#### Which are the steps to reproduce?
1. Create a dunning document
2. Run the action for the Jasper (Mahnbrief):

<details>

```
Error: org.adempiere.exceptions.AdempiereException: Caught class java.io.IOException...
Server error
org.adempiere.exceptions.AdempiereException: Caught class java.io.IOException trying to invoke URL http://mf15releasetest:8080/adempiereJasper/ReportServlet?AD_Process_ID=1000000&AD_PInstance_ID=1000766&AD_Language=de_DE&output=PDF; message: Server returned HTTP response code: 500 for URL: http://mf15releasetest:8080/adempiereJasper/ReportServlet?AD_Process_ID=1000000&AD_PInstance_ID=1000766&AD_Language=de_DE&output=PDF

{timestamp: 1527514451779, status: 500, error: "Internal Server Error",…}
error
:
"Internal Server Error"
exception
:
"org.adempiere.exceptions.AdempiereException"
message
:
"org.adempiere.exceptions.AdempiereException: Caught class java.io.IOException trying to invoke URL http://mf15releasetest:8080/adempiereJasper/ReportServlet?AD_Process_ID=1000000&AD_PInstance_ID=1000764&AD_Language=de_DE&output=PDF; message: Server returned HTTP response code: 500 for URL: http://mf15releasetest:8080/adempiereJasper/ReportServlet?AD_Process_ID=1000000&AD_PInstance_ID=1000764&AD_Language=de_DE&output=PDF"
path
:
"/rest/api/process/ADP_1000000/1000764/start"
status
:
500
timestamp
:
1527514451779
trace
:
[,…]
[0 … 99]
0
:
"org.adempiere.exceptions.AdempiereException: org.adempiere.exceptions.AdempiereException: Caught class java.io.IOException trying to invoke URL http://mf15releasetest:8080/adempiereJasper/ReportServlet?AD_Process_ID=1000000&AD_PInstance_ID=1000764&AD_Language=de_DE&output=PDF; message: Server returned HTTP response code: 500 for URL: http://mf15releasetest:8080/adempiereJasper/ReportServlet?AD_Process_ID=1000000&AD_PInstance_ID=1000764&AD_Language=de_DE&output=PDF"
1
:
[...]

PR #4126, opened 2018-05-29
 #4123 Fix documentPaid function

cristinamghita:
#4123 Dunning Jasper does not work


metas-ts:
on a search for documentpaid among *.java, *.sql files, it seems that only these four actually use it:
image

cristinamghita:
@metas-ts Tobi, I am not sure about checking the order.
Because, now, as we refactor code and reimplemented prepay order, a prepay order is paid only when its invoice is paid.
And that is covered by these  2 lines:
```
LEFT OUTER JOIN C_Invoice i ON (al.C_Invoice_ID = i.C_Invoice_ID)
```
and
```
OR (i.C_Order_ID = p_Record_ID AND p_AD_Table_ID = (SELECT AD_Table_ID FROM AD_Table WHERE TableName = 'C_Order'))
```

metas-ts:
so i suggest to just kick "C_Order" out as the quickest solution (2nd quickest: loos the whole funktion and make sure that `invoicepaid` works)

metas-ts:
:cherries: picked to `release`

PR #4171, opened 2018-06-04
integrate recent posting and flatrateterm fixes

metas-ts:


metas-ts:
please don't delete the branch!

PR #4279, opened 2018-06-26
gh3463 improvements in manufacturing issues

metas-rc:
https://github.com/metasfresh/metasfresh/issues/3210
https://github.com/metasfresh/metasfresh/issues/3463


metas-ts:
hi @metas-rc what's the difference between cuToNewCU and and cuToNewCU_AllocateMaximumPossibleQty in HUTransformService

metas-ts:
Hi @metas-rc  
Update II
..at least in the current version of cuToNewCU, it seems to me that if `qtyCU >= maximumQtyCU`, the splitting is not even done. Was this different in former versions?
If it was, then i wonder why not fix the existing method?

Plese verify the case you try to fix with a test case, if none exists yet.
There are a lot of existing tests in the `HUTransformService*Test` classes.

Note: some tests use xpath to verify the results, the most resent ones like use `HUTransformServiceReservationTests.husToNewCUs_mixed_source_HU()` HU-specific asserj assertions. If the problem can be fixes or is fixed in `cuToNewCU()`, please kick out `cuToNewCU_AllocateMaximumPossibleQty()`. If not, please ping me

Issue #4219, opened 2018-06-13
Import Datev Primanota for Payments, Allocations


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, it is not possible to import Payments from Datev prima nota.
#### Which are the steps to reproduce?
Open, try and see.
### What is the expected or desired behavior?
Enable the Import of Datev Primanota and allocate the Payments with invoices found.

metas-lc:
IT
new window: Import Datev Payment
1. create a new Data Import, with format: Datev Payment
2. attach a file, import it
=> note that you must write 0,00 in discount to be able to import the file. If you leave it empty, it won't import it OK?
3. check the new Payments
=> Partner Bank Account is null (and mandatory) OK?
=> they are completed, allocated OK
=> linked to the correct invoice OK

PR #4300, opened 2018-06-29
#4219 Import Datev Primanota for Payments, Allocations 

cristinamghita:
#4219 Import Datev Primanota for Payments, Allocations


de.metas.banking/de.metas.banking.base/src/main/java/de/metas/banking/payment/impexp/CPaymentImportTableSqlUpdater.java
```diff
@@ -0,0 +1,117 @@
...
+
+	public void updatePaymentImportTable(@NonNull final String whereClause)
+	{
+		dbUpdateBPartners(whereClause);
+		dbUpdateInvoices(whereClause);
+		dbUpdateIsReceipt(whereClause);
+		dbUpdateErrorMessages(whereClause);
+	}
+
+	private void dbUpdateBPartners(@NonNull final String whereClause)
```
metas-ts: imho we need to include AD_Org_ID in the whereclauses `C_BPartner.Value`, `C_Invoice.DocumentNo` etc can be ququal among different orgs

Issue #4270, opened 2018-06-22
Purchase Order Document with up to 4 digits


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, the Purchase order has a fixed amount of digits shown.
#### Which are the steps to reproduce?
Open, try and see.
### What is the expected or desired behavior?
The digits shown shall be depending on digits recorded in underlying Pricelist.

metas-lc:
devtest: purchase order, pricelist has price precision 1 (test with 2,3,4..)
add a line or update a price... e.g. 4.1234 => will be transformed to 4.10 and in jasper you'll see 4.1
(note that it's taken from orderline, so if you change a precision you should also change the price to match)

metas-rc:
Results of IT

Tested on webui

* OK : The report for purcahse order respects the price with the number of decimals from the orderline. 
* OK: If the orderline price was not changed ( it's the one taken from the priccelist)  the number of decimals is the one from the pricelist (e.g. in my orderline I had the price 1.00 and in the report I have it 1.000)  but I think this was wanted. 


* NOK: The price';s decimals are marked with "." while the qty's  ones are marked with ",". Please, make them look the same.
![image](https://user-images.githubusercontent.com/16001442/42519098-f4116b78-846b-11e8-908a-fff3609f0066.png)

metas-lc:
added a pattern that works up to 4 digits. For DE will have "," and for EN will have "."

metas-lc:
changed again: the pattern is dynamic, depending on the price entered. 
if you want to test with more than 4 digits, the number might not fit, but you can check in the database 
select * from de_metas_endcustomer_fresh_reports.Docs_Purchase_Order_Details(order_id, 'de_CH')
.
also added document note from product in this sql, to be used later in other reports. Not used in purchase order atm (should not affect)

metas-dh:
Results of IT2
tested in release

* setting a product price as 0,50: price is 0,50 in product prices and orderline, but displayed as 0,5 in Jasper for the purchase order: NOK

metas-dh:
Results of IT3
tested in mf15

product price: 1,1515

1. precision 1: Jasper en: 1.2, de: 1,2: OK
2. precision 2: Jasper en: 1.15, de: 1,15: OK
3. precision 3: Jasper en: 1.152, de: 1,152: OK
4. precision 4: Jasper en: 1.1515, de: 1,1515: OK
[...]

PR #4358, opened 2018-07-10
#4270 added pattern to match currency

metas-lc:
#4270


metas-ts:
hi @metas-lc the issue asked for the number of digits to be determined by the underlying price list:

> The digits shown shall be depending on digits recorded in underlying Pricelist.

Maybe you can compute the pattern (i.e. the number of `#`s) dynamically?

metas-lc:
there will be prices with 2,3 or 4 digits, but as you said, would be better to make it dynamic.. checking

metas-lc:
please check again. I used sql to make the pattern

Issue #4216, opened 2018-06-13
Follow up : #4153 Serial Letter Feature 


cristinamghita:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
Follow up [#4153 Serial Letter Feature ](https://github.com/metasfresh/metasfresh/issues/4153)
* Process of creating marketing contacts should have a parameter list, from where the user to choose which address will be used (billtodefault, shiptodefault).
* The list will be displayed only if MKTG_Platform has flag IsRequiredLocation on Y

metas-lc:
IT
Business Partner window: there are 2 actions: 
old one: Export in E-Mail Campaign with 1 parameter
new one: Export in Serienbrief Kampagne with 2 parameters

1. have a bpartner with multiple locations. Make one billtodefault and one shiptodefault
2. make sure you have a user on your bpartner

=> could not create user NOK

<details/>
2018-07-25 15:00:31.844 ERROR 12232 --- [http-nio-8080-exec-10] o.a.a.m.AnnotatedModelInterceptor        : Model interceptor method de.metas.marketing.base.interceptor.MKTG_ContactPerson#onChangeEmail threw an exception.
You can disable this method with SysConfig InterceptorEnabled_de.metas.marketing.base.interceptor.MKTG_ContactPerson#onChangeEmail='N' (with AD_Client_ID and AD_Org_ID=0!)
2018-07-25 15:00:31.863  WARN 12232 --- [http-nio-8080-exec-10] o.adempiere.ad.trx.api.impl.TrxManager   : AbstractTrxManager.call0 - caught class org.adempiere.exceptions.AdempiereException with message=NullPointerException: userRecord
2018-07-25 15:00:31.863  WARN 12232 --- [http-nio-8080-exec-10] o.adempiere.ad.trx.api.impl.TrxManager   : AbstractTrxManager.call0 - caught class org.adempiere.exceptions.AdempiereException with message=NullPointerException: userRecord
2018-07-25 15:00:31.866  WARN 12232 --- [http-nio-8080-exec-10] o.adempiere.ad.trx.api.impl.TrxManager   : AbstractTrxManager.call0 - caught class org.adempiere.exceptions.AdempiereException with message=NullPointerException: userRecord
[...]

PR #4398, opened 2018-07-20
Gh4216 create contact persons with the right address

metas-rc:
https://github.com/metasfresh/metasfresh/issues/4216


metas-ts:
Hi @metas-rc 
there is an SQL error: https://jenkins.metasfresh.com/job/metasfresh-dist/job/master/11500/console

metas-rc:
It looks like I somehow lost the part of the script where I created the reference. 
I created a new simila reference and I linked it woth the process para.
Changes were made in the same old script, which should be fine since the first execution failed. 
Changes puched.

Issue #4392, opened 2018-07-19
ESR RefNo too short after scan in Purchase Invoice action


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, the Process "Read Payment" (Zahlung einlesen) takes the number sequence as delivered by reader and saves as is.
#### Which are the steps to reproduce?
Open, try and see.
### What is the expected or desired behavior?
Sometimes it happens that the ESR Documents have ESR RefeNo that are too short. In such cases we shall fill the RefNo with leading zeroes up to the default length for the RefNo.

metas-lc:
window zahlung-zuordnung, process Zahlung einlesen
config: 
Payment String: 0100X XXXXXX X8>YYYYYYY00100000642001431904+ ZZZZZZZZZ> 
X XXXXXX X is the amount of the invoice you pay: for 11,90 it will be 0 000119 0
YYYYYYY is the account number 
ZZZZZZZZZ is the ESR account
.
in this case we have to test different length for ESR account (ZZZZZZZZZ)

PR #4410, opened 2018-07-25
#4392 Add leading zeroes to match length

cristinamghita:
#4392 ESR RefNo too short after scan in Purchase Invoice action


de.metas.payment.esr/src/main/java/de/metas/payment/spi/impl/ESRCreaLogixStringParser.java
```diff
@@ -171,7 +171,11 @@ public IPaymentString parse(final Properties ctx, final String paymentTextOrigin
 		else
 		{
 			esrReferenceNoToMatch = referenceNumber.substring(7, 15);
```
metas-ts: why do we have `esrReferenceNoToMatch = referenceNumber.substring(7, 15);` 
if `referenceNumberLength != 27`?

to me it seems that if `referenceNumberLength != 27` we do assume that `referenceNumberLength == 15`.
however, the case this very issue is about seems to a third case (length neither 27 nor 15).
if we have a "sortof" 27-digit string that somehow has just 23 digits, then `referenceNumber.substring(7, 15);` is probably not a great idea.

Please check out whats the issue, here and also write a testcase; extract the code under test into a dedicated method if required to have the test.

Issue #4416, opened 2018-07-26
Split Product Nutrition/ Allergene in 2 Tabs


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, the Product Nutrition/ Allergens can be recorded in Product Subtab.
#### Which are the steps to reproduce?
Open, try and see.
### What is the expected or desired behavior?
Split Product Nutrition/ Allergens in 2 Tabs. One for Nutrition, one for Allergens.

metas-dh:
Results of IT1
tested in mf15 & customer db

* Nutritions & Allergens in 2 tabs, in product window: OK
* de_DE: tab is called M_Product_Allergen, shall be Produkt Allergen: NOK
* de_DE: field in tab is called M_Allergen, shall be just Allergen: NOK
* en_US: tab is called Produkt Nährwerte, shall be Product Nutrition: NOK
* en_US: field in tab is called Nutrition/Allergen, shall be just Nutrition: NOK
* also the order of fields in the allergen tab is NOK imho: client_ID and org_ID are the first 2 fields, shall be at the end and allergen shall be the first field

=> need to check again when latest commit is rolled out

metas-rc:
on it

metas-dh:
Results of IT2
tested in mf15

* en_US: tab is called "Produkt Nutrition" (WebUI & Swing), shall be "Product Nutrition"
* en_US: in Product Nutrition: WebUI & Swing: "Quanitity", shall be "Quantity"
* order of fields: pls make sure the fields for AD_Org_ID and AD_Client_ID are shown at the end, both in Swing & WebUI!

metas-rc:
Fixed in branch gh4417

metas-dh:
Results of IT3
tested in mf15

* en_US: tab is called "Product Nutrition" (WebUI & Swing): OK
* en_US: in Product Nutrition: WebUI & Swing: "Quantity": OK
* order of fields in the tabs:
    * AD_Org_ID and AD_Client_ID at the end, in WebUI: OK
    * AD_Org_ID and AD_Client_ID at the beginning, in Swing: OK in customer db, mf15: Nutrition tab: "Nutrition" first, "Product" at the end, AD_Org_ID in the middle: NOK, pls make the order like in Allergen tab

metas-dh:
Results of IT4
tested in mf15

* order of fields in the tabs:
[...]

PR #4420, opened 2018-07-26
Gh4416 Nutrition facts and allergens split and displayed in webui

metas-rc:
https://github.com/metasfresh/metasfresh/issues/4416


metas-rc:
Fixed

Issue #4423, opened 2018-07-27
Add vattaxid in Docs_Purchase_Order_Details_Sum function


cristinamghita:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
Function Docs_Purchase_Order_Details_Sum should return also vattaxid

metas-dh:
Results of IT1
tested in mf15

normal PO:
* purchase order sums ok, working as before, taxes calculated correctly, Jasper fine as well: OK

need to check on customer db prob. as well

metas-dh:
checked with metas-mk: OK, rolled out at customer, working fine.

PR #4425, opened 2018-07-30
#4423 add vattaxid

cristinamghita:
#4423 Add vattaxid in Docs_Purchase_Order_Details_Sum function


cristinamghita:
@metas-ts , yes I you are right!
So I modified the function.
Thx!

Issue #4426, opened 2018-07-30
Error on new purchase order


metas-ts:
### Is this a bug or feature request?
Bug
### What is the current behavior?

<details>

```

2018-07-30 12:22:29.480  WARN 132 --- [https-jsse-nio-8443-exec-3] de.metas.ui.web.window.model.Document    : Failed initializing DocumentField{fieldName=C_Order.IsEdiEnabled, value=null, initalValue=null}, mode=NewDocument using [FAILED toString()]. Keeping current initial value.

org.adempiere.ad.expression.exceptions.ExpressionEvaluationException: * Nicht gefunden *: IsEdiRecipient
        at org.adempiere.ad.expression.api.impl.StringExpressionSupportTemplate$SingleParameterExpressionTemplate.evaluate(StringExpressionSupportTemplate.java:372)
        at de.metas.ui.web.window.model.Document$InitialFieldValueSupplier.getValue(Document.java:686)
        at de.metas.ui.web.window.model.Document.initializeField(Document.java:491)
        at de.metas.ui.web.window.model.Document.initializeFields(Document.java:409)
        at de.metas.ui.web.window.model.Document.access$2000(Document.java:92)
        at de.metas.ui.web.window.model.Document$Builder.build(Document.java:2101)
        at de.metas.ui.web.window.model.Document$Builder.initializeAsNewDocument(Document.java:2161)
        at de.metas.ui.web.window.model.sql.SqlDocumentsRepository.createNewDocument(SqlDocumentsRepository.java:314)
        at de.metas.ui.web.window.model.DocumentCollection.createRootDocument(DocumentCollection.java:325)
        at de.metas.ui.web.window.model.DocumentCollection.forRootDocumentWritable(DocumentCollection.java:258)
        at de.metas.ui.web.window.model.DocumentCollection.forDocumentWritable(DocumentCollection.java:228)
        at de.metas.ui.web.window.controller.WindowRestController.patchDocument0(WindowRestController.java:318)
        at de.metas.ui.web.window.controller.WindowRestController.lambda$patchDocument$5(WindowRestController.java:312)
        at org.adempiere.ad.trx.api.impl.TrxCallableWrappers$4.call(TrxCallableWrappers.java:193)
        at org.adempiere.ad.trx.api.impl.TrxCallableWrappers$5.call(TrxCallableWrappers.java:222)
        at org.adempiere.ad.trx.api.impl.AbstractTrxManager.call0(AbstractTrxManager.java:775)
[...]

PR #4427, opened 2018-07-30
#4426-app - Error on new purchase order

metas-ts:
don't load GridFieldVOs if there is no AD_Field,
Idk why it was done to start with, and it causes problems down the line, if those GridFieldVOs have default value expressions that point to other columns' context expressions.
Error on new purchase order #4426


metas-ts:
hi @teosarca 
note that with this PR, `AD_Field`s are now inner-joined. 
I.e. no AD_Field => no FieldVO.

The trigger was that i stumbled over another case (this task) where something failed (in this case creating a new record, on a particular system, hell to even try to reproduce), because one FieldVO was somehow there and the other somehow wasn't. and both did not have AD_Fields in the particular tab.
(purchase order lines).

This means that now we need the AD_Fields, if we want to work with them, e.g. in callouts. I think that's fair enough, and we are doing it anyways.

Issue #4467, opened 2018-08-07
Get rid of the legacy IProcess interface


metas-ts:
### Is this a bug or feature request?
FR
### What is the current behavior?

https://github.com/metasfresh/metasfresh-webui-api/issues/1010 introduced a problem with (jasper-)reports started from the menu tree.

#### Which are the steps to reproduce?

try to run e.g. `Umsatz Geschäftspartner`:

```
[...]
"Caused by: java.lang.ClassCastException: org.compiere.report.ReportStarter cannot be cast to de.metas.process.JavaProcess"
91:"   at de.metas.process.JavaProcess.currentInstance(JavaProcess.java:135)"
92:"   at de.metas.ui.web.process.adprocess.ADProcessInstanceController.updateParametersDocumentFromJavaProcessAnnotatedFields(ADProcessInstanceController.java:229)"
93:"   at de.metas.ui.web.process.adprocess.ADProcessInstanceController.processParameterValueChanges(ADProcessInstanceController.java:224)"
94:"   at de.metas.ui.web.process.ProcessRestController.lambda$null$6(ProcessRestController.java:237)"
[...]
```

### What is the expected or desired behavior?

instead of bothering with IProcess, remove it

metas-dh:
Results of IT1
tested in mf15

running reports in WebUI works fine again (e.g. Revenue report, Articel statistics), no errors: OK

metas-ts:
FYI, :cherries: picked https://github.com/metasfresh/metasfresh/issues/4467 to the `release` branch

PR #4469, opened 2018-08-07
#4476-app - Get rid of the legacy IProcess interface

metas-ts:
* remove the interface mostly change IProcess to JavaProcess
* clean up MPayment and "online processing"

Get rid of the legacy IProcess interface https://github.com/metasfresh/metasfresh/issues/4467


metas-ts:
wrong branch name

Issue #4507, opened 2018-08-22
Picked and Non-Picked Quantities Shipment generation


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, shipments can be created via Shipment Schedule action. The user has to decide
a) with picked Quantities
b) or without
If a Sales Order has picked and non-picked Quantities, then the shipment can only be done via 2 shipment notes/ documents.
#### Which are the steps to reproduce?
Open, try and see.
### What is the expected or desired behavior?
A new action for shipment creation shall recognize that picked Quantities are available and use those. If non-picked Quantities exist then it shall take those. Alle Lines shall be added to one Shipment (depending on Aggregation settings).

metas-dh:
Results of IT1
tested in release

1. 1 sales order with 2 orderlines, same product
2. picked partial qty for both orderlines:
* LU with 2 TUs for one
* standalone CU for the other
3. tried to create shipment from shipment schedule, with QtyPicked=Y and Complete=Y: no shipment was created, but no error msg: NOK

4. checked async:

<details>

```
18-09-17 09:57:17.735  WARN 18876 --- [WT-EventQueue-0] d.m.a.form.swing.SwingClientUIInstance   : java.lang.NullPointerException

java.lang.NullPointerException: null
	at org.compiere.grid.ed.VDocAction.dynInit(VDocAction.java:251)
	at org.compiere.grid.ed.VDocAction.<init>(VDocAction.java:96)
	at org.compiere.apps.APanel.actionButton0(APanel.java:2873)
	at org.compiere.apps.APanel.actionButton(APanel.java:2797)
	at org.compiere.apps.APanel.actionPerformed(APanel.java:1944)
	at org.compiere.grid.GridController$1.actionPerformed(GridController.java:335)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2348)
	at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402)
	at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259)
	at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:252)
	at java.awt.AWTEventMulticaster.mouseReleased(AWTEventMulticaster.java:289)
	at java.awt.Component.processMouseEvent(Component.java:6533)
[...]

PR #4564, opened 2018-09-10
Gh4507 Shipment creation with qty options

metas-rc:
https://github.com/metasfresh/metasfresh/issues/4507


de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/ShipmentScheduleWithHU.java
```diff
@@ -129,9 +129,14 @@ public static final ShipmentScheduleWithHU ofShipmentScheduleWithoutHu(
 
 	private I_M_InOutLine shipmentLine = null;
 
+
+	private M_ShipmentSchedule_QuantityTypeToUse qtyTypeToUse = M_ShipmentSchedule_QuantityTypeToUse.TYPE_D; // keep old functionality as default
```
metas-ts: qtyTypeToUse and isForPickedQty  seem to be sortof redundant;
doesn't TYPE_D and TYPE_BD imply that isForPickedQty = true ?

In that case, you could kick out `isForPickedQty ` and write a method `isForPickedQty()` that uses qtyTypeToUse

de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/ShipmentScheduleWithHU.java
```diff
@@ -144,6 +149,10 @@ private ShipmentScheduleWithHU(
 		this.vhu = shipmentScheduleAlloc.getVHU_ID() > 0 ? shipmentScheduleAlloc.getVHU() : null;
 		this.tuHU = shipmentScheduleAlloc.getM_TU_HU_ID() > 0 ? shipmentScheduleAlloc.getM_TU_HU() : null;
 		this.luHU = shipmentScheduleAlloc.getM_LU_HU_ID() > 0 ? shipmentScheduleAlloc.getM_LU_HU() : null;
+
+		this.setForPickedQty(isForPicked);
+		this.setQtyTypeToUse(qtyTypeToUse);
```
metas-ts: `this.setQtyTypeToUse(qtyTypeToUse);` doesn't have any effect right now because the `qtyTypeToUse` is `this.qtyTypeToUse`

de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/ShipmentScheduleWithHU.java
metas-ts: when looking at the places where `setQtyTypeToUse()` is called, i think it makes sense to 
extend on `ShipmentScheduleWithHU.ofShipmentSchedule...()` instead of having the public setter

de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/process/M_ShipmentSchedule_EnqueueSelection.java
```diff
@@ -54,8 +54,8 @@
 		implements IProcessPrecondition
 {
 
-	@Param(parameterName = "IsUseQtyPicked", mandatory = true)
-	private boolean isUseQtyPicked;
+	@Param(parameterName = "Quantity", mandatory = true)
```
metas-ts: with Quantity as param name, i'd expect a quantity..please call it quantitytype etc

de.metas.swat/de.metas.swat.base/src/main/sql/postgresql/system/48-de.metas.inoutcandidate/5500120_sys_gh4507_M_ShipmentSchedule_GenerateShipment_QtyOptions_Process.sql
```diff
@@ -0,0 +1,105 @@
...
+;
+
+-- 2018-08-29T15:05:32.570
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+INSERT INTO AD_Reference_Trl (AD_Language,AD_Reference_ID, Description,Help,Name, IsTranslated,AD_Client_ID,AD_Org_ID,Created,Createdby,Updated,UpdatedBy) SELECT l.AD_Language,t.AD_Reference_ID, t.Description,t.Help,t.Name, 'N',t.AD_Client_ID,t.AD_Org_ID,t.Created,t.Createdby,t.Updated,t.UpdatedBy FROM AD_Language l, AD_Reference t WHERE l.IsActive='Y' AND l.IsSystemLanguage='Y' AND l.IsBaseLanguage='N' AND t.AD_Reference_ID=540907 AND NOT EXISTS (SELECT 1 FROM AD_Reference_Trl tt WHERE tt.AD_Language=l.AD_Language AND tt.AD_Reference_ID=t.AD_Reference_ID)
+;
+
+-- 2018-08-29T15:06:10.817
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+INSERT INTO AD_Ref_List (AD_Client_ID,AD_Org_ID,AD_Ref_List_ID,AD_Reference_ID,Created,CreatedBy,EntityType,IsActive,Name,Updated,UpdatedBy,Value,ValueName) VALUES (0,0,541681,540907,TO_TIMESTAMP('2018-08-29 15:06:10','YYYY-MM-DD HH24:MI:SS'),100,'de.metas.inoutcandidate','Y','Picked',TO_TIMESTAMP('2018-08-29 15:06:10','YYYY-MM-DD HH24:MI:SS'),100,'P','Picked')
```
metas-ts: not sure if there is already some other agreement in place,
but i would like to have the german value in here. otherwise there will be another issue with more overhead later.
please get them from a german collegue, and also take care of the de_CH and en_US trls

de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/ShipmentScheduleWithHU.java
metas-rc: Done.
I had to keep also the other version, without QtyTypeToUse, because these methods are not only used by the shipment creator. In some places we don't have this info, e.g. when we link QtyPicked with shipment schedules.

Issue #4519, opened 2018-08-27
Data Migration Allergene/ Nutritional Values


metas-rc:
* Support data migration for allergens and nutrition facts.

metas-ts:
thix PR is obsolete; the task was solved in a different way

PR #4651, opened 2018-10-04
#4519-app - Data Migration Allergene/ Nutritional Values

metas-ts:
#4519


metas-ts:
thix PR is obsolete; the task was solved in a different way

PR #4731, opened 2018-11-02
Gh4663 safe Add base language to element translation

metas-rc:
https://github.com/metasfresh/metasfresh/issues/4663


metas-ts:
Hi @metas-rc i assigned to you now, pls unassign when you think it's ready for review
(incl. when it builds successfully).

Issue #4717, opened 2018-10-30
Set Order-Checkup printout receiver via AD_Workflow


metas-ts:
### Is this a bug or feature request?
FR

metas-ts:
Done: 
* add new AD_Workflow.AD_USer_InCharge_ID column
* use it instead of M_Warehouse.AD_User_InCharge_ID when creating production-order checkup reports

cristinamghita:
##Dev test
*Prerequisites*
* Make sure that sys config 'de.metas.fresh.ordercheckup.CreateAndRouteJasperReports.OnSalesOrderComplete' is set on Y
* restart the server after setting that on Y
* create a product planning for the product that you will use to create the sales order
  *  make sure that the product planning has  set a warehouse, a resource, and a workflow set
  *  make sure that the workflow has a user in charge set
  * make sure that the warehouse that you will use to create the sales order has a PP_Plant set and that plant has a resource set
* create a sales order using the warehouse and product  used above
* complete the order
   * check the newly created C_Order_MFGWarehouse_Report and see that the AD_User_Responsible is the one from workflow

metas-dh:
Results of IT1
tested in mf15

SetUp:
=> as described above, for AD_User: Admin, Automatik-Benutzer (could not select all existing users, just the Automatik-Benutzer ones: is that OK?)

Test:
1. Created Sales Order for product with acc. plan data
2. Checked the resp. reports:
* Bestellkontrolle: AD_User_Responsible is Admin, Automatik-Benutzer, OK
* Bestellkontrolle spedition: AD_User_Responsible is metasfresh: NOK
=> both are C_Order_MFGWarehouse_Report, so shouldn't both have the same AD_User_Responsible set?

PR #4748, opened 2018-11-08
#4717 Set Order-Checkup printout receiver via AD_Workflow

cristinamghita:
#4717 Set Order-Checkup printout receiver via AD_Workflow


de.metas.adempiere.adempiere/client/src/main/java-legacy/org/compiere/apps/wf/WFPanel.java
```diff
@@ -97,7 +100,7 @@ public void setSimpleWorkflowWindow(boolean isSimpleWorkflowWindow)
 
 	/** Workflow WhereClause : General, Document Process, Document Value */
 	private static final String WORKFLOW_WhereClause = "WorkflowType IN ("
-			+ DB.TO_STRING(MWorkflow.WORKFLOWTYPE_General)
+			+ DB.TO_STRING(X_AD_Workflow.WORKFLOWTYPE_Allgemein)
```
metas-ts: Please update the workflowType AD_Ref_List to make it (I think the column is `ValueName`) "General", nocht "Allgemein" etc..i.e. the english version. The german one crept in years ago and keeps popping up here and there..

Issue #4764, opened 2018-11-19
Translation of Greetings in WebUI


metas-rc:
### Is this a bug or feature request?
* FR
### What is the current behavior?
* There is not webui window for Greetings
#### Which are the steps to reproduce?
* Open and see
### What is the expected or desired behavior?
* Create new Window in WebUI
* Add Menu Entry
* Reference from Greeting window to Greeting Translation window

metas-dh:
Results of IT1
tested in mf15

* new window "Greeting" / "Anreden": OK
* new window "Greetings Translation" / "Anreden Übersetzung": OK
* both included in menu: OK
* reference from Greeting window to Greeting Translation window: no referenced docs, NOK:

![2018-11-22 16_39_24-metasfresh](https://user-images.githubusercontent.com/18696715/48912186-3e621600-ee75-11e8-90c5-1ecb8c7824f1.png)

PR #4765, opened 2018-11-19
#4764 greetings trl window

metas-rc:
https://github.com/metasfresh/metasfresh/issues/4764


metas-ts:
@metas-rc please unassign from yourself when ready

Issue #4777, opened 2018-11-20
Import Address GLN via I_BPartner


metas-rc:
### Is this a bug or feature request?
* FR
### What is the current behavior?
* Currently the GLN cannot be set with the BPartner Import process
#### Which are the steps to reproduce?
* Check and see
### What is the expected or desired behavior?
* Add the GLN column in I_BPartner table
* Add logic to make it propagate to the C_BPartner_Location.

metas-ts:
hi @metas-rc 

afaiu some I-table columns are also lookup columns where the system decides whether it shall create or update an existing record.
I think GLN shall be such a column

-------------------------

As talked with @metas-mk :
while we are at it, please also import 
* C_BPartner_Location.ExternalId
* AD_User.ExternalId
* C_BPartner.ExternalId

note: as "lookup-columns" (like GLN), they shall have the highest prio..i.e. if the I_BPartner record has an BPartnerLocation_ExternalId and GL, then the system shall first check if there is a C_BPartner_Location with the external ID and then update even that record's GLN.

PR #4780, opened 2018-11-21
#4777 GLN in BPartner Import

metas-rc:
https://github.com/metasfresh/metasfresh/issues/4777

### DIFF
--- a/de.metas.business/src/main/java/de/metas/bpartner/impexp/BPartnerImportTableSqlUpdater.java
+++ b/de.metas.business/src/main/java/de/metas/bpartner/impexp/BPartnerImportTableSqlUpdater.java
@@ -4,10 +4,13 @@ import static org.adempiere.impexp.AbstractImportProcess.COLUMNNAME_I_ErrorMsg;
 import static org.adempiere.impexp.AbstractImportProcess.COLUMNNAME_I_IsImported;
 
 import org.adempiere.ad.trx.api.ITrx;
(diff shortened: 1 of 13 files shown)


metas-ts:
hi @metas-rc 
it seems that this PR is still WIP because the ExternalId columns are not yet handled.
Pls unassign it from yourself when done.

de.metas.business/src/main/java/de/metas/bpartner/impexp/BPartnerImportTableSqlUpdater.java
```diff
@@ -64,10 +66,19 @@ public void updateBPartnerImtortTable(@NonNull final String whereClause)
 
 		dbUpdateAdUserIdsFromExisting(whereClause);
 
-		dbUpdateCbPartnerIds(whereClause);
+		dbUpdateCbPartnerIdsFromValue(whereClause);
```
metas-ts: Imho `ExternalId` (if set) needs to have higher prio than `Value`,
so 
`dbUpdateCbPartnerIdsFromValue(whereClause);`
and 
`dbUpdateCbPartnerIdsFromC_BPartner_ExternalId(whereClause);`

need to be the other way aground.
Agree?

metas-rc: :+1:

Issue #4787, opened 2018-11-26
Product Document Note Translation


metas-rc:
### Is this a bug or feature request?
* FR
### What is the current behavior?
* Currently, the Product Document Note cannot be translated.
#### Which are the steps to reproduce?
* try
### What is the expected or desired behavior?
* Add the Document/ Note into Product Translation Window

metas-dh:
Results of IT1
tested in mf15

* Document Note in Product Translation Window: OK
* text in Document Note from Product Translation Window is set for the resp. language set for bpartner: NOK => bpartner had en_US set (that was also set as menu language), but document note text in tooltip and order confirmation was the de_DE one => #4804

metas-dh:
after checking the resp. task in me03, i think making sure the trl is correctly used was part of the original requirements => giving it back to dev

metas-dh:
Results of IT2
tested in mf15

* text in Document Note from Product Translation Window is set for the resp. language set for bpartner: 
=> de_DE text / en_US for the bpartners with the resp. language: 
* displayed in tooltip: OK
* displayed in order confirmation: de_DE document note displayed in en_US order: checked with metas-cg: should show no document note at all: NOK

metas-dh:
Results of IT3
tested in mf15

* displayed in order confirmation: de_DE document note displayed in de_DE order jasper, en_US : document note displayed in en_US order jasper (acc. to bpartner language): OK
* orderline is displayed too often (doubled): in case of 1 orderline, there are 2 displayed in the jasper, in case of 2 orderlines there are 4 in jasper: NOK

PR #4807, opened 2018-11-29
Gh4787 Document note in order line

metas-rc:
https://github.com/metasfresh/metasfresh/issues/4787


metas-rc:
Hi. I tested locally that the tooltip shows the doc note in the language set in C_BPartner. But the build was failing last time I was on this and didn't get to create an instance ad do the screen shots .

Issue #4659, opened 2018-10-10
Configurable Mailtext for Documents in DocOutbound


metas-ts:
### Is this a bug or feature request?
FR
### What is the current behavior?
There are two AD_Messages:
* `MailWorkpackageProcessor.EmailSubject`
* `MailWorkpackageProcessor.EmailMessage`

whatever the contained is used for every outgoing mail

#### Doc-OutBoundLog

* If the `C_Doc_OutBoundLog` has a `C_DocType` and if that `C_Doctype` has an `AD_BoilerPlate`,
then use that boilerplate's subject and textsnippet
  * Note: variables and sub-boilerplates are not in this issue's scope, they won't work in here
* otherwise, fall back to the two AD_Messages from above
* for both AD_Messages and Boilerplate, if there are any translations for the required language, they shall be used.
  * the required language is determined by checking the following places and using the first one that has a language
    * If `C_Doc_OutBoundLog.CurrentEMailRecipient_ID > 0`, then look for the language of that user
    * If `C_Doc_OutBoundLog.C_BPArtner_ID > 0`  then look for the language of that bpartner
    * If `C_Doc_OutBoundLog.CurrentEMailRecipient_ID > 0`, but the respective user has no language set, then look for that user's bPartner's language
    * Look for the current context-language (e.g. the language of the metasfresh user that created the WP)
    * If all fails, return the base language

#### Marketing

Marketing uses boilerplate, so since we make that boilerplate language-aware, I thought that marketing-contacts also need to be language aware now (to reduce inconsistencies)

*  `MKTG_ContactPerson` now has a non-manadatory language field
* when a new `MKTG_ContactPerson` is created from `AD_User` it gets the user's language.
   * if the user has no language, but the user's bpartner has, it gets the bpartner's language
* if the contact's language is saved, then the for that language, the same rules apply that are described here: https://github.com/metasfresh/metasfresh/issues/4092

metas-ts:
Notes to IT: 
* we have some unit test coverage when it comes to synching language and email between AD_USer and MKTG_ContactPerson, so imho that doesn need (vigorous) testing
* please test the marketing-letter feature with a translated boilerplate
* also, please test the mail-from-docoutbound-feature, with a translated boilerplate to be set in the respective doctype
* i did very little dev-testing. my excluse: i wanted to use cypress, but then thought that it's more important to create this issue to enable future testing: https://github.com/metasfresh/metasfresh-webui-frontend/issues/2002

[...]

PR #4809, opened 2018-11-30
#4659-app - Configurable Mailtext for Documents in DocOutbound

metas-ts:
* MKTG_ContactPerson's C_Location is now editable *if* there is no C_BPartner set
* MKTG_ContactPerson.AD_Language is editable via WebUI
* AD_User.AD_Language is also editable in the bpartner window (adv edit)

https://github.com/metasfresh/metasfresh/issues/4659#issuecomment-442050655


metas-dh:
IT was done in another task, see above.

Issue #4795, opened 2018-11-27
Product No. of customer also with ASI


metas-rc:
### Is this a bug or feature request?
* FR
### What is the current behavior?
* There is no ASI field in C_BPartner_Product
#### Which are the steps to reproduce?
* Open and see
### What is the expected or desired behavior?

* C_BPartner_Product must have an ASI column
* There should be no duplicate asis per product and partner

Example:
* Product P001
* Partner G001
* Attribute A1

Entry 1: C_BPartner_Product for P001 and G001 , ASI : A1 = 20
Entry 2: C_BPartner_Product for P001 and G001, ASI: null
Entry 3: C_BPartner_Product for P001 and G001, ASI: A1 = 30

* All of the above are correct

Entry 4: C_BPartner_Product for P001 and G001, ASI: A1 = 30 will not be correct because the ASI would be similar with an existing one.

* Create a validator that does now allow a new entry to be added with the same attribute value.

metas-dh:
Results of IT2
tested in customer db

4. product no. is correctly displayed in customer reports:
a. if there is a product no. in C_BPartner_Product matching the attribute in orderline, this is displayed: OK
b. if the attribute in orderline has no specific product no. set in C_BPartner_Product, the product no. with no attribute set is displayed: OK
c. if there is no product no. set in C_BPartner_Product, then no product. no is displayed: OK
d. correctly displayed in all reports:
* order confirmation: OK
* shipment: OK
* invoice: OK

PR #4810, opened 2018-12-03
#4795 Add function getC_BPartner_Product_Details

cristinamghita:
#4795 Product No. of customer also with ASI


cristinamghita:
hi!
@metas-ts is not working ok.
We get all the results, not just one, the perfect match.
But I will try to understand what you did and fix it for our cases.

Issue #4924, opened 2019-02-01
Problem with manual invoice and non-default product-UOM


metas-ts:
### Is this a bug or feature request?
Bug
### What is the current behavior?

* make sure you know which UOM is flagged as default (most of the time that's "EA")
* make sure you have a product that has a different UOM, with no conversion rule between the two UOMs
* create a sales invoice
* add a line with the product, i.e. press "Add new" to open the dialog
  * there is a readonly field with the default UOM, read-only
  * you you get error messages and can'T select you product and save the new order line

PR #4925, opened 2019-02-01
#4924 Problem with manual invoice and non-default product-UOM

metas-ts:
#4924


metas-ts:
finishing this means more effort than redoing (if it it's not yet done..parts are for sure)

Issue #4944, opened 2019-02-08
Improve metasfresh reports


cristinamghita:
### Is this a bug or feature request?
FR

metas-dh:
Results of IT1
tested in mf15

* order confirmation Jasper: dates are alligned with confirmation no.: OK
* footer: line is slightly bigger, more central: OK

PR #4948, opened 2019-02-11
 #4944 Improve metasfresh reports

cristinamghita:
#4944 Improve metasfresh reports


cristinamghita:
> I'm not sure what the general consequences of these changes might be; please consult with @metas-dh, Mark or Norbert.

It is confirmed with @metas-mk

PR #4957, opened 2019-02-12
Hide parameter when multi select

cristinamghita:
Make Issue CUs from Source HU work for Multi Select
https://github.com/metasfresh/metasfresh-webui-api/issues/1142


cristinamghita:
> from what i can see, the new process parameter is not translated..

Is never shown that parameter. I added only in order be able to hide the Qty  one.

Issue #4920, opened 2019-01-31
Improve creating product prices when importing IFA products


cristinamghita:
### What is the expected or desired behavior?
* the product prices shall be created in order
* user buffer when loading imported records

metas-dh:
see screenshots in https://github.com/metasfresh/me03/issues/997 for IT.

PR #4958, opened 2019-02-12
#4920 Improve creating product prices when importing IFA products 

cristinamghita:
#4920 Improve creating product prices when importing IFA products


de.metas.vertical.pharma/src/main/java/de/metas/impexp/product/PharmaProductImportProcess.java
```diff
@@ -220,7 +225,8 @@ protected void afterImport()
 				.addOnlyActiveRecordsFilter()
 				.filter(sqlFilter)
 				.create()
```
metas-ts: Some weeks ago we talked about a problem related to records not beeing processed in the order of their PLV?
Was that addressed?

At any rate, I suggest to order by ID here, in order for the code to be more predictable

de.metas.business/src/main/java/org/adempiere/impexp/product/MProductImportTableSqlUpdater.java
```diff
@@ -534,4 +534,15 @@ private void dbUpdateErrorMessages(@NonNull final String whereClause)
 		no = DB.executeUpdateEx(sql.toString(), ITrx.TRXNAME_ThreadInherited);
 		logger.warn("No Mandatory Pharma product category Name ={}", no);
 	}
+	
+	public void dbUpdateIsPriceCreated(@NonNull final String whereClause)
+	{
+		StringBuilder sql;
+		sql = new StringBuilder("UPDATE ")
+				.append(targetTableName + " i ")
+				.append(" SET IsPriceCreated = 'Y' ")
```
metas-ts: please use the constant instead of the hardcoded string. otherwise it's hard to find the use of this physical column in our code

de.metas.vertical.pharma/src/main/sql/postgresql/system/5511790_sys_gh4920_AddIsPriceCreated.sql
```diff
@@ -0,0 +1,55 @@
...
+;
+
+-- 2019-02-04T17:10:17.782
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+UPDATE AD_Table_Process SET WEBUI_DocumentAction='Y',Updated=TO_TIMESTAMP('2019-02-04 17:10:17','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Table_Process_ID=540603
+;
+
+-- 2019-02-04T17:10:20.410
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+UPDATE AD_Table_Process SET WEBUI_DocumentAction='Y',Updated=TO_TIMESTAMP('2019-02-04 17:10:20','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Table_Process_ID=540671
```
metas-ts: AD_Element_Trl for "Price created" is missing 
..but then, the  new column is not displayed, right? So i guess it's OK

de.metas.vertical.pharma/src/main/java/de/metas/impexp/product/PharmaProductImportProcess.java
cristinamghita: Yes, it was; here, we just retrieve records in order to get the matched price lists

de.metas.vertical.pharma/src/main/sql/postgresql/system/5511790_sys_gh4920_AddIsPriceCreated.sql
cristinamghita: yes, is not displayed; Is just for filtering old from new.

Issue #4968, opened 2019-02-18
Support attribute value type Number in view Report.Fresh_Attributes


cristinamghita:
### Is this a bug or feature request?
Bug

metas-dh:
Results of IT1
tested in live2439

quickly checked:

* Cable attribute is type number and is shown in the report: OK
* the product no. based on asi, is shown in the report: OK

PR #4969, opened 2019-02-18
#4968 Support Number for view Report.fresh_Attributes

cristinamghita:
#4968 Support attribute value type Number in view Report.Fresh_Attributes


de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/views/fresh_Attributes.sql
```diff
@@ -12,7 +12,7 @@ SELECT * FROM
 			WHEN a.Value = '1000001' AND (av.value IS NOT NULL AND av.value != '') THEN av.value -- Herkunft
 			WHEN a.Value = '1000021' AND (ai.value IS NOT NULL AND ai.value != '') THEN ai.Value -- MHD
 			WHEN a.Value = 'HU_BestBeforeDate' AND (ai.valuedate IS NOT NULL) THEN 'MHD: '|| to_char(ai.valuedate, 'DD.MM.YYYY') --Best Before Date
-			WHEN a.AttributeValueType = 'S' AND (ai.value IS NOT NULL AND ai.value != '') THEN ai.Value
+			 WHEN ((a.attributevaluetype = 'S'::bpchar OR a.attributevaluetype = 'N'::bpchar ) AND ((ai.value IS NOT NULL) AND ((ai.value)::text <> ''::text))) THEN ai.value
```
metas-ts: * do we need the explicit casts to `bpchar` and `text`? 
* why not use `a.attributevaluetype IN ('S', 'N')` instead? imho it's clearer..
* do we really need all those brackets?
* spend a bit of the saved space on also checking for e.g. `'   '`:

```suggestion
			 WHEN a.attributevaluetype IN ('S', 'N') AND COALESCE(TRIM(ai.value),'') != '' THEN ai.value
```

cristinamghita: Yes, you are right; did the change. Is fine.
We will have a new task for this view were we need to make it better and test all cases because now, is easy to break.

Issue #5051, opened 2019-03-18
Allow the formatting of IBAN in BPartner Bank 


cristinamghita:
### Is this a bug or feature request?
FR

metas-dh:
IT documented in me03 task.

PR #5052, opened 2019-03-18
#5051 Do not remove spaces

cristinamghita:
#5051 Allow the formatting of IBAN in BPartner Bank


cristinamghita:
> please check:
> 
> * where is the IBAN used?
> * might a space-less IBAN required at those places?

Mainly is displayed.
And everywhere, before to be used, the spaces are removed anyway.
The only place where the IBAN is not checked for spaces is `de.metas.payment.sepa.api.impl.SEPADocumentBL.getCreateHeader(Properties, int, List<I_SEPA_Export>, String)`
And I do not know if we need to do it or not.
But for safety, we can remove space first, there, I think.

Issue #5071, opened 2019-03-25
Don't allow narcotics in Trade (Sales & Purchase)


TheBestPessimist:
### Is this a bug or feature request?
FR

### What is the current behavior?
Currently the narcotics permissions are not handled.

#### Which are the steps to reproduce?
-

###### Abstract

Atm, BPartner has ShipmentPermissionPharma and ReceiptPermissionPharma dropdowns which can take values like A (Type A), B (Type B).
ShipmentPermissionPharma is for when we are selling to that customer.
ReceiptPermissionPharma is for when we are buying from that vendor.

Type A means that BPartner is allowed to sell/buy RX(prescription) products.
Type B means that BPartner is NOT allowed to sell/buy RX products.

In M_Product we have a flag for "RX/prescription product" and also we have a flag for "is narcotic product" (Betäubungsmittel).

TODO: Introduce Type C:
Introduce Type C which allows BPartner to buy/sell narcotic products.
Type C includes Type A, i.e. RX products.

IMPORTANT: Make sure

if a product is flagged as RX product and also as Narcotic
BPartner ShipmentPermissionPharma/ReceiptPermissionPharma is Type A
then that partner is not allowed to buy/sell that product.
Only with Type C it would be allowed.
TODO: Make sure importing the Product's Narcotic flag (Betäubungsmittel) works
TODO: Make sure nobody can set the Type C
To dev test it, just update it in database.

metas-dh:
IT documented in Don't allow narcotics in Trade (Sales & Purchase).

PR #5078, opened 2019-03-26
Gh5071: Don't allow narcotics in Trade Sales

TheBestPessimist:
Implements:  https://github.com/metasfresh/metasfresh/issues/5071


de.metas.vertical.pharma/src/test/java/de/metas/vertical/pharma/PharmaOrderLineInputValidatorTest.java
```diff
@@ -0,0 +1,265 @@
...
+		PharmaProductRepository.class,
+		C_OrderLine.class,
+})
+public class PharmaOrderLineInputValidatorTest
+{
+	@Rule public ExpectedException thrown = ExpectedException.none();
+
+	private C_OrderLine orderLineInterceptor;
+
+	@BeforeClass
```
metas-ts: note: for new unit tests it makes sense to use jupiter. but no need to change this one now

Issue #5058, opened 2019-03-20
Manufacturer Import IFA


cristinamghita:
### Is this a bug or feature request?
FR
### What is the current behavior?
None
### What is the expected or desired behavior?
We shall be able to import IFA Manufacturers

metas-dh:
Results of IT1
tested in customer IT

1. Import IFA Hersteller:
* bpartners from import are set as Manufacturer:`Y`: OK
* bpartners from import have location data set (address, phone, email etc.): OK
* bpartners already existing were updated, no doubles: OK

2. Import product IFA - IA68:
* products have correct manufacturer info set: OK
(this did not seem to work at first, but the import was not yet processed => after processing the products had the correct manufacturer info set, OK)

(checked quickly in mediosprod: bpartners & products were set correctly with manufacturer info as well)

PR #5087, opened 2019-03-27
#5058 Manufacturer Import IFA 

cristinamghita:
#5058 Manufacturer Import IFA


de.metas.vertical.pharma/src/main/java/de/metas/impexp/bpartner/IFABPartnerLocationImportHelper.java
```diff
@@ -92,47 +96,50 @@ private I_C_BPartner_Location fetchAndUpdateExistingBPLocation(@NonNull final I_
...
+				.collect(ImmutableList.toImmutableList());
+
+		final List<Address> filtered = matchedAddreses.stream()
+				.filter(address -> isMatched(importAddress, address))
+				.collect(ImmutableList.toImmutableList());
+
+		if (filtered.size() > 0)
+		{
+			final I_C_BPartner_Location bpartnerLocation = Services.get(IBPartnerDAO.class).getBPartnerLocationById(BPartnerLocationId.ofRepoId(importRecord.getC_BPartner_ID(), matchedAddreses.get(0).getBpLocationId()));
+			updateExistingBPartnerLocation(importRecord, bpartnerLocation);
```
metas-ts: don't we want to return the updated `bpartnerLocation` here?

cristinamghita: Yes! You are right!

Issue #5095, opened 2019-04-01
Enable set/unset of BPartner narcotics pharma permission


TheBestPessimist:
### Is this a bug or feature request?
FR

### What is the current behavior?
Pharma Bpartner narcotics permission cannot be seen or selected anywhere.

#### Which are the steps to reproduce?
N/A

### What is the expected or desired behavior?

The pharma narcotics permission flag should be visible read-only in 
Business Partner Pharma -> Customer/Vendor
Business Partner Pharma -> Customer/Vendor -> Advanced edit

The narcotics flag should be editable in
Business Partner Pharma -> Actions -> Update Vendor Pharma Permissions
Business Partner Pharma -> Actions -> Update Customer Pharma Permissions

Setting or unsetting the flag should also set/unset the Shipment/Receipt Type C permission.

TheBestPessimist:
dev-instance: https://dev382.metasfresh.com
version: 5.99.2_14888_gh5095

# Customer 

1. customer with no narcotics permission
![1 customer no narcotics permission](https://user-images.githubusercontent.com/4482210/55331988-1b87e400-549d-11e9-84df-10d7df8c9c79.png)

2. customer no narcotics permission -- advanced edit
![2  customer no narcotics permission -- advanced edit](https://user-images.githubusercontent.com/4482210/55331990-1c207a80-549d-11e9-983e-5705f9f299ce.png)

3. customer -- update customer pharma permissions -- enable narcotics
![3 customer -- update customer pharma permissions -- enable narcotics](https://user-images.githubusercontent.com/4482210/55331991-1c207a80-549d-11e9-8820-3b6ce86bef2b.png)

4. customer with narcotics permission
![4 customer with narcotics permission](https://user-images.githubusercontent.com/4482210/55331993-1c207a80-549d-11e9-84b2-2a10f3bf273d.png)

5. customer with narcotics permission -- advanced edit
![5 customer with narcotics permission -- advanced edit](https://user-images.githubusercontent.com/4482210/55331994-1c207a80-549d-11e9-9336-3962a358008f.png)



# Vendor




1. Vendor no narcotics permission
![1 vendor no narcotics permission](https://user-images.githubusercontent.com/4482210/55332187-918c4b00-549d-11e9-8a35-1ecde31ee39f.png)

2. Vendor no narcotics permission -- advanced edit
[...]

PR #5096, opened 2019-04-01
Enable set/unset of BPartner narcotics pharma permission

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/5095

### DIFF
--- a/de.metas.vertical.pharma/src/main/java/de/metas/vertical/pharma/PharmaVendorPermissions.java
+++ b/de.metas.vertical.pharma/src/main/java/de/metas/vertical/pharma/PharmaVendorPermissions.java
@@ -54,7 +54,8 @@ public final class PharmaVendorPermissions
 		{
 			permissionsBuilder.add(PharmaVendorPermission.PHARMA_WHOLESALE);
 		}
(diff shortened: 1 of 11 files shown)


de.metas.vertical.pharma/src/main/java/de/metas/vertical/pharma/PharmaVendorPermissions.java
```diff
@@ -54,6 +54,9 @@ public static PharmaVendorPermissions of(final I_C_BPartner bpartner)
 		{
 			permissionsBuilder.add(PharmaVendorPermission.PHARMA_WHOLESALE);
 		}
+		if (bpartner.isPharmaVendorNarcoticsPermission()) {
```
metas-ts:
```suggestion
		if (bpartner.isPharmaVendorNarcoticsPermission()) 
		{
```
please take a look and see if you have the metasfresh (eclipse-) formatting rules from metasfresh-dev.
The current formatting rules are not written in stone, but everyone should use the same

TheBestPessimist: My bad, i forgot to indent there properly. I do have the formatting rules selected.

Issue #5148, opened 2019-04-17
Process of importing products is not working anymore


cristinamghita:
### Is this a bug or feature request?
Bug

PR #5149, opened 2019-04-17
#5148 Use Optional

cristinamghita:
#5148 Process of importing products is not working anymore

### DIFF
--- a/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/impexp/AbstractImportProcess.java
+++ b/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/impexp/AbstractImportProcess.java
@@ -433,7 +433,7 @@ public abstract class AbstractImportProcess<ImportRecordType> implements IImport
 	protected final void runSQLAfterRowImport(@NonNull final ImportRecordType importRecord)
 	{
 		final List<DBFunction> functions = getDbFunctions().getAvailableAfterRowFunctions();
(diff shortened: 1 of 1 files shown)


de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/impexp/AbstractImportProcess.java
```diff
@@ -433,7 +433,7 @@ protected void afterImport()
 	protected final void runSQLAfterRowImport(@NonNull final ImportRecordType importRecord)
 	{
 		final List<DBFunction> functions = getDbFunctions().getAvailableAfterRowFunctions();
-		final Optional<Integer> dataImportId = InterfaceWrapperHelper.getValueOrNull(importRecord, COLUMNNAME_C_DataImport_ID);
+		final Optional<Integer> dataImportId =  Optional.ofNullable(InterfaceWrapperHelper.getValueOrNull(importRecord, COLUMNNAME_C_DataImport_ID));
```
metas-ts:
```suggestion
		final Optional<Integer> dataImportId =  InterfaceWrapperHelper.getValue(importRecord, COLUMNNAME_C_DataImport_ID);
```

cristinamghita: @metas-ts , no is not ok; InterfaceWrapperHelper.getValue will throw an exception when the column is not found. And I want to return null if the column is not found.

Issue #5145, opened 2019-04-17
Fix custom data entry section window -- line tab


TheBestPessimist:
### Is this a bug or feature request?
bug

### What is the current behavior?
Lines are shown for different sections

#### Which are the steps to reproduce?
open a section and see

metas-ts:
appears to be a local problem only; other appliation dictionary are OK

PR #5152, opened 2019-04-22
Fix custom data entry section window -- line tab

metas-ts:
https://github.com/metasfresh/metasfresh/issues/5145


metas-ts:
problem seemed to be only local

Issue #5177, opened 2019-05-07
Checking BtM and PZN with Modulo 11


metas-rc:
### Is this a bug or feature request?
* FR
### What is the current behavior?
* 
#### Which are the steps to reproduce?
*
### BTM number ###

* This will be a new column in the C_Bpartner table.
* It will be a string of 7 chars, the last of them being a modulo11 check digit
* Each time the BTM is set to a partner, we must make sure the checkdigit is valid.
* Each time a partner receives  Narcotic Permissions ( type c)  via the permission processes, we must verify that the partner has a (valid)  BTM set.

### PZN number ###

* This will be taken from the Product Value
*  It will be a string of 8 chars, the last of them being a modulo11 check digit
* It will be a string of 7 chars, the last of them being a modulo11 check digit ( see http://www.mobiliodevelopment.com/pzn-german-pharmacy-barcode/#gref ) 
* Each time a pharma product is inserted we must make sure it has a valid PZN as Value. 
* A product is Pharma Product if it belongs to a PharmaProductCategory

metas-rc:
Tested on live4661

* Pharma Product flag displayed in Pharma Product window

![image](https://user-images.githubusercontent.com/16001442/57519658-7d512e80-7324-11e9-87a7-f56a33ce2fe6.png)

* The product above already has a valid PZN (Value)
* Set the flag Pharma Product

![image](https://user-images.githubusercontent.com/16001442/57519752-bbe6e900-7324-11e9-8518-7543a9a7aafe.png)


* => OK The product is correctly saved


* Create a new product  -  it will receive a value that is not PZN  

![image](https://user-images.githubusercontent.com/16001442/57519845-e9cc2d80-7324-11e9-8f63-b8a86c814926.png)


* Set the flag Pharma Product: 

![image](https://user-images.githubusercontent.com/16001442/57519881-010b1b00-7325-11e9-81b7-03a3553d2056.png)


* => OK: There is an error message about the invalid pzn

* Go to window Pharma Business Partner ( on this system the window is called simply Business Partner because this is a pharma system) 

* Create a new entry

![image](https://user-images.githubusercontent.com/16001442/57520009-50514b80-7325-11e9-95c0-1c573ed51e6b.png)


* => OK: BTM displayed in the window


* Try to give this partner Narcotic permissions

![image](https://user-images.githubusercontent.com/16001442/57520079-7840af00-7325-11e9-9d78-9619ba12ada9.png)

[...]

PR #5187, opened 2019-05-09
Gh5177 Checking BtM and PZN with Modulo 11

metas-rc:
https://github.com/metasfresh/metasfresh/issues/5177


de.metas.vertical.pharma/src/main/java/de/metas/vertical/pharma/PharmaModulo11Validator.java
```diff
@@ -0,0 +1,99 @@
...
+{
+	private static final int PZN_MODULO11 = 11;
+
+	private static final int PZN7_Length = 7;
+	private static final int PZN7_InitialMultiplier = 2;
+
+	private static final int PZN8_Length = 8;
+	private static final int PZN8_InitialMultiplier = 1;
+
+	public static boolean isValid(final String pzn)
```
metas-ts: I suggest to have `@NonNull` (lombok) here

de.metas.vertical.pharma/src/main/java/de/metas/vertical/pharma/PharmaModulo11Validator.java
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+/**
+ * I took the information about PZN code from here
+ * http://www.pruefziffernberechnung.de/P/PZN.shtml
```
metas-ts: good to document where the info is coming from 👍

Issue #5199, opened 2019-05-14
Avoid updating AD_EventLog records


metas-ts:
### Is this a bug or feature request?
FR

PR #5200, opened 2019-05-14
5199-app - Avoid updating AD_EventLog records

metas-ts:
* eventlog records are created before the event is posted
* event handlers just add their log records, but *only* update the event log record to set its error=Y if they run into an error

Goal: eliminate almost all updates to AD_EventLog and therefore avoid a large number of dead rows in the DB
https://github.com/metasfresh/metasfresh/issues/5199


de.metas.adempiere.adempiere/base/src/main/java/de/metas/event/impl/EventBus.java
```diff
@@ -190,20 +201,16 @@ public void postEvent(@NonNull final Event event)
 			return;
 		}
 
+		if (event.isStoreEvent() && event.isLocalEvent())
```
metas-ts:
```suggestion
		// as long as we have just one common event-log-DB, we store events only on the machine they were created on, in order to avoid duplicates.
		if (event.isStoreEvent() && event.isLocalEvent())
```

de.metas.adempiere.adempiere/base/src/main/java/de/metas/event/log/EventLogService.java
```diff
@@ -132,54 +117,32 @@ public void storeEventLogEntry(@NonNull final EventLogEntry eventLogEntry)
...
+	private int retrieveEventLogIdUsingCacheOutOfTrx(@NonNull final UUID uuid)
 	{
-		final Supplier<Integer> valueInitializer = () -> retrieveOrCreateRecordIdOutOfTrx(uuid);
-		final int eventStoreRecordId = uuid2eventLogId.get(uuid, valueInitializer);
+		final int eventStoreRecordId = uuid2eventLogId.getOrLoad(uuid, this::retrieveRecordIdOutOfTrx);
 		return eventStoreRecordId;
 	}
 
-	private int retrieveOrCreateRecordIdOutOfTrx(final UUID uuid)
+	private Integer retrieveRecordIdOutOfTrx(final UUID uuid)
```
metas-ts:
```suggestion
	private Integer retrieveRecordIdOutOfTrx(@NonNull final UUID uuid)
```

Issue #5210, opened 2019-05-17
Customs Document


metas-rc:
### Is this a bug or feature request?
* FR
### What is the current behavior?
* We don't have a Customs Document/
#### Which are the steps to reproduce?
-
### Concept ###

* Add 2 new tables: 

#### C_Customs_Invoice ####
* C_Bpartner, C_BPartner_Location, AD_User
* Date
* DocumentNo
* Currency
* Basic Document columns

#### C_Customs_Invoice_line ####
* Product
* LineNetAmt
* Qty
* UOM
* Default columns

----------------------------------------------------
* Add a window for the tables above
* Add a new doctype _Custom Invoice_ ,  docBaseType _CUI - Custom Invoice_,  I won't add a docSubType yet. To be added when we have more options for this document.  Update the DocBaseType list.

* Add new flag in M_InOut _IsExportedToCustomsInvoice_ which turns true when an inout was processed into a Custom Invoice
* Add a new column in M_InOut called _Custom_Invoice_ . This will be filled when a Custom Invoice is created based on the M_InOut.  In the end, all the Shipments used in the Custom Invoice will be available in its Referenced Documents.

* Add the flag in the Shipment window and filter

* Create a process _Export Shipments to Custom Invoice_,
* Add a process parameter _Logistic Company_  ( this will be the C_BPartner of the Customs Invoice)
* Link the process with the table M_InOut, window Shipment.

#### Logic ####

* Create a process  M_Customs_Invoice_CreateFromInouts
* Parameters: Selected M_Inouts, a C_BPartner_ID called "Logistic Company" together with its Location and user.

* Collect all the inout lines of the selected inouts, map them by product id.
* The process creates a single Customs Invoice for all the selected inouts.
* We will need to decide upon which currecny to use, in case the selected inouts have different currencies. Suggestion: Take the currency from the Country of the C_BPartner_Location. @metas-ts  Is this ok? 
* Each entry of this map will turn into a Custom_Invoice_Line
* The qty on the custom invoice line will be the sum of all the qtys from the lines for the product. IMO I should convert them all to the product's basic C_UOM and this will also be the UOM of the Custom Invoice Line.
* The lineNetAmt will be the sum of prices from the inout lines of the product.  The prices will be converted in the Currency set in the header.
* Shall the document be completed automatically or left in Draft state? (@metas-mk ?) , or maybe it should be a parameter in the process?

* Add unit tests for the aggregation part.

[...]

PR #5223, opened 2019-05-21
Gh5210 Customs Invoice Creation

metas-rc:
https://github.com/metasfresh/metasfresh/issues/5210

### DIFF
--- a/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_InOut.java
+++ b/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_InOut.java
@@ -16,6 +16,8 @@
  *****************************************************************************/
 package org.compiere.acct;
 
+import static org.adempiere.model.InterfaceWrapperHelper.getTableId;
+
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.util.ArrayList;
@@ -24,7 +26,6 @@ import java.util.List;
 import java.util.Set;
 
 import org.adempiere.ad.dao.IQueryBL;
-import org.adempiere.model.InterfaceWrapperHelper;
 import org.adempiere.service.ISysConfigBL;
 import org.compiere.model.I_M_InOut;
 import org.compiere.model.I_M_InOutLine;
@@ -366,7 +367,7 @@ public class Doc_InOut extends Doc<DocLine_InOut>
 		if (MInOut.DOCSTATUS_Reversed.equals(m_DocStatus) && m_Reversal_ID > 0 && line.getReversalLine_ID() > 0)
 		{
 			// Set AmtAcctDr from Original Shipment/Receipt
-			if (!dr.updateReverseLine(InterfaceWrapperHelper.getTableId(I_M_InOut.class), m_Reversal_ID, line.getReversalLine_ID(), BigDecimal.ONE))
+			if (!dr.updateReverseLine(getTableId(I_M_InOut.class), m_Reversal_ID, line.getReversalLine_ID(), BigDecimal.ONE))
 			{
 				throw newPostingException().setDetailMessage("Original Receipt not posted yet");
 			}
@@ -388,7 +389,7 @@ public class Doc_InOut extends Doc<DocLine_InOut>
 		if (MInOut.DOCSTATUS_Reversed.equals(m_DocStatus) && m_Reversal_ID > 0 && line.getReversalLine_ID() > 0)
 		{
 			// Set AmtAcctCr from Original Shipment/Receipt
-			if (!cr.updateReverseLine(InterfaceWrapperHelper.getTableId(I_M_InOut.class), m_Reversal_ID, line.getReversalLine_ID(), BigDecimal.ONE))
+			if (!cr.updateReverseLine(getTableId(I_M_InOut.class), m_Reversal_ID, line.getReversalLine_ID(), BigDecimal.ONE))
 			{
 				throw newPostingException().setDetailMessage("Original Receipt not posted yet");
 			}
--- a/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_InOut.java
+++ b/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_InOut.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Set;
 
 import org.adempiere.ad.dao.IQueryBL;
+import org.adempiere.model.InterfaceWrapperHelper;
 import org.adempiere.service.ISysConfigBL;
 import org.compiere.model.I_M_InOut;
 import org.compiere.model.I_M_InOutLine;
@@ -45,21 +46,21 @@ import de.metas.util.Services;
 
 /**
  * Post Shipment/Receipt Documents.
- * 
+ *
  * <pre>
  *  Table:              M_InOut (319)
  *  Document Types:     MMS, MMR
  * </pre>
- * 
+ *
  * metas:
  * <li>copied from // metas: from https://adempiere.svn.sourceforge.net/svnroot/adempiere/branches/metas/mvcForms/base/src/org/compiere/acct/Doc_InOut.java, Rev 10177
  * <li>Changed for "US330: Geschaeftsvorfall (G113d): Summen-/ Saldenliste (2010070510000637)"
- * 
+ *
  * @author Jorg Janke
  * @author Armen Rizal, Goodwill Consulting
  *         <li>BF [ 1745154 ] Cost in Reversing Material Related Docs
  *         <li>BF [ 2858043 ] Correct Included Tax in Average Costing
- * 
+ *
  */
 public class Doc_InOut extends Doc<DocLine_InOut>
 {
@@ -128,7 +129,7 @@ public class Doc_InOut extends Doc<DocLine_InOut>
 
 	/**
 	 * Create Facts (the accounting logic) for MMS, MMR.
-	 * 
+	 *
 	 * <pre>
 	 *  Shipment
 	 *      CoGS (RevOrg)   DR
@@ -140,7 +141,7 @@ public class Doc_InOut extends Doc<DocLine_InOut>
 	 *      Inventory       DR
 	 *      NotInvoicedReceipt      CR
 	 * </pre>
-	 * 
+	 *
 	 * @param as accounting schema
 	 * @return Fact
 	 */
@@ -365,7 +366,7 @@ public class Doc_InOut extends Doc<DocLine_InOut>
 		if (MInOut.DOCSTATUS_Reversed.equals(m_DocStatus) && m_Reversal_ID > 0 && line.getReversalLine_ID() > 0)
 		{
 			// Set AmtAcctDr from Original Shipment/Receipt
-			if (!dr.updateReverseLine(MInOut.Table_ID, m_Reversal_ID, line.getReversalLine_ID(), BigDecimal.ONE))
+			if (!dr.updateReverseLine(InterfaceWrapperHelper.getTableId(I_M_InOut.class), m_Reversal_ID, line.getReversalLine_ID(), BigDecimal.ONE))
 			{
 				throw newPostingException().setDetailMessage("Original Receipt not posted yet");
 			}
@@ -387,7 +388,7 @@ public class Doc_InOut extends Doc<DocLine_InOut>
 		if (MInOut.DOCSTATUS_Reversed.equals(m_DocStatus) && m_Reversal_ID > 0 && line.getReversalLine_ID() > 0)
 		{
 			// Set AmtAcctCr from Original Shipment/Receipt
-			if (!cr.updateReverseLine(MInOut.Table_ID, m_Reversal_ID, line.getReversalLine_ID(), BigDecimal.ONE))
+			if (!cr.updateReverseLine(InterfaceWrapperHelper.getTableId(I_M_InOut.class), m_Reversal_ID, line.getReversalLine_ID(), BigDecimal.ONE))
 			{
 				throw newPostingException().setDetailMessage("Original Receipt not posted yet");
 			}
@@ -435,7 +436,7 @@ public class Doc_InOut extends Doc<DocLine_InOut>
 	/**
 	 * Creating a not null cost value whose precision matches the given <code>C_Currency_ID</code>'s.
 	 * Goal of this method is to avoid the warnings in FactLine.setAmtSource().
-	 * 
+	 *
 	 * @param currencyId
 	 * @param costs
 	 * @return costs rounded to currency precision
--- a/de.metas.business/src/main/java/de/metas/customs/CustomsInvoice.java
+++ b/de.metas.business/src/main/java/de/metas/customs/CustomsInvoice.java
@@ -10,8 +10,7 @@ import com.google.common.collect.ImmutableList;
 
 import de.metas.bpartner.BPartnerLocationId;
 import de.metas.document.DocTypeId;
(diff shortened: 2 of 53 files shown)


de.metas.business/src/main/java/de/metas/customs/CustomsInvoice.java
```diff
@@ -0,0 +1,90 @@
...
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+@Data
+@Builder(toBuilder = true)
+@FieldDefaults(level = AccessLevel.PRIVATE, makeFinal = true)
+public class CustomsInvoice
+{
+	@NonFinal
+	CustomsInvoiceId id;
```
metas-ts: So far i annotated my beans with `@Value` and therefore had everything `final`. And instead of updating some field, i created a new instance.
e.g. when i save a CustomsInvoice, instead of setting the id, i did something like this at the end of the `save` method:
```java
return myCustomsInvoiceWithoutId.toBuilder().id(myNewIdIJustGot).build();
```

@teosarca wdyt these two approaches?

de.metas.business/src/main/java/de/metas/customs/CustomsInvoiceService.java
```diff
@@ -0,0 +1,258 @@
...
+	public CustomsInvoiceService(
+			@NonNull final CustomsInvoiceRepository customsInvoiceRepo,
+			@NonNull final OrderLineRepository orderLineRepo)
+	{
+		this.customsInvoiceRepo = customsInvoiceRepo;
+		this.orderLineRepo = orderLineRepo;
+
+	}
+
+	public CustomsInvoice generateCustomsInvoice(final BPartnerLocationId bpartnerAndLocationId,
```
metas-ts: i believe you should create an immutable (=> `@Value`)request class, instead of having these 6 params

de.metas.business/src/main/java/de/metas/customs/CustomsInvoiceService.java
```diff
@@ -0,0 +1,258 @@
...
+
+		if (isComplete)
+		{
+			documentBL.processEx(customsInvoiceRecord, IDocument.ACTION_Complete, IDocument.STATUS_Completed);
+		}
+
+		return customsInvoice;
+	}
+
+	private CustomsInvoice createCustomsInvoice(final BPartnerLocationId bpartnerAndLocationId,
```
metas-ts: same advise (=> request class)

de.metas.business/src/main/java/de/metas/customs/CustomsInvoiceService.java
```diff
@@ -0,0 +1,258 @@
...
+		final IInOutDAO inoutDAO = Services.get(IInOutDAO.class);
+		final ICurrencyBL currencyBL = Services.get(ICurrencyBL.class);
+
+		final InOutLineId inOutLineId = inoutAndLineId.getInOutLineId();
+
+		final I_M_InOutLine inoutLineRecord = inoutDAO.getLineById(inOutLineId);
+
+		final OrderLineId orderLineId = OrderLineId.ofRepoIdOrNull(inoutLineRecord.getC_OrderLine_ID());
+
+		if (Check.isEmpty(orderLineId))
```
metas-ts: Did you check with mark if this is OK? Otherwise, you might need to fall back to the invoice candidate, or maybe using the pricing engine..
After all, they might want to export something that is just on the shipment line and has no order line. btw, what about packaging inout lines? They don't have an order lines

EDIT: OK, so it seems as if packaging lines are both going to end up in a customs invoice. still, pls clarify with mark..

de.metas.business/src/main/java/de/metas/customs/process/M_InOut_Create_CustomsInvoice.java
```diff
@@ -0,0 +1,224 @@
...
+		{
+			return false;
+		}
+
+		if (inOutBL.isReversal(shipment))
+		{
+			return false;
+		}
+
+		if (!documentBL.isStatusCompletedOrClosedOrReversed(shipment.getDocStatus()))
```
metas-ts: is it really a requirement that we can't create a customs invoice for a completed shippment??

de.metas.business/src/main/java/de/metas/customs/process/M_InOut_Create_CustomsInvoice.java
```diff
@@ -0,0 +1,224 @@
...
+		return shipmentLinesToExport;
+
+	}
+
+	private List<InOutAndLineId> retrieveValidLinesToExport(final InOutId shipmentId)
+	{
+		final IInOutDAO inOutDAO = Services.get(IInOutDAO.class);
+		List<InOutAndLineId> shipmentLines = new ArrayList<>();
+
+		final ImmutableList<InOutAndLineId> shipmentLinesToExport = inOutDAO.retrieveLinesForInOutId(shipmentId)
```
metas-ts: * if the service class only supports shipmentline with an orderline, i think you should filter for that in here
* maybe instead of filtering with a predicate, you should create some service/DAO/repository to get those lines, and in there, use IQuery to just get from the DB the stuff you really need

de.metas.business/src/main/java/de/metas/customs/CustomsInvoiceService.java
metas-rc: He said it's ok for the first round. He will ask the customer tomorrow

de.metas.business/src/main/java/de/metas/customs/process/M_InOut_Create_CustomsInvoice.java
metas-rc: Yes. Confirmed with Mark that only completed and closed inouts are to be exported.

de.metas.business/src/main/java/de/metas/customs/process/M_InOut_Create_CustomsInvoice.java
metas-rc: done

de.metas.business/src/main/java/de/metas/customs/CustomsInvoiceService.java
metas-rc: done

de.metas.business/src/main/java/de/metas/customs/CustomsInvoiceService.java
metas-rc: done

de.metas.business/src/main/java/de/metas/customs/process/M_InOut_Create_CustomsInvoice.java
metas-ts: sry, i had in mind, to filter for it when making the IQuery. 
but i guess we won't have 1000s of shipmentlines without orderlines, so this is also OK

de.metas.business/src/main/java/de/metas/letters/model/MADBoilerPlate.java
```diff
@@ -349,7 +349,7 @@ else if (parent_table_id == I_C_Payment.Table_ID)
 		{
 			rq.setC_Payment_ID(parent_record_id);
 		}
-		else if (parent_table_id == I_M_InOut.Table_ID)
+		else if (parent_table_id == InterfaceWrapperHelper.getTableId(I_M_InOut.class))
```
metas-ts:
```suggestion
		else if (parent_table_id == getTableId(I_M_InOut.class))
```

de.metas.business/src/main/java/de/metas/customs/process/M_InOut_Create_CustomsInvoice.java
metas-rc: Right. I didn't notice I used the wrong method :blush: 
I changed it to
```
if (documentBL.isStatusStrOneOf(shipment.getDocStatus(), DocStatus.Completed.getCode(), DocStatus.Closed.getCode()))
```

de.metas.business/src/main/java/de/metas/customs/process/M_InOut_Create_CustomsInvoice.java
metas-rc: if all goes well, all the lines with no C_OrderLine_ID will already be dropped at this point: 
```
	if (shipmentLineRecord.isPackagingMaterial())
		{
			return false;
		}

```

de.metas.business/src/main/java/de/metas/customs/CustomsInvoiceService.java
```diff
@@ -0,0 +1,271 @@
...
+	public CustomsInvoiceService(
+			@NonNull final CustomsInvoiceRepository customsInvoiceRepo,
+			@NonNull final OrderLineRepository orderLineRepo)
+	{
+		this.customsInvoiceRepo = customsInvoiceRepo;
+		this.orderLineRepo = orderLineRepo;
+
+	}
+
+	public CustomsInvoice generateCustomsInvoice(final CustomsInvoiceRequest customsInvoiceRequest)
```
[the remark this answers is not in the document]
metas-rc: done

de.metas.business/src/main/java/de/metas/customs/CustomsInvoiceService.java
```diff
@@ -0,0 +1,271 @@
...
+		{
+			documentBL.processEx(customsInvoiceRecord, IDocument.ACTION_Complete, IDocument.STATUS_Completed);
+
+			customsInvoice = customsInvoiceRepo.updateDocActionAndStatus(customsInvoice);
+		}
+
+		return customsInvoice;
+	}
+
+	private CustomsInvoice createAndProcessCustomsInvoice(final CustomsInvoiceRequest customsInvoiceRequest)
```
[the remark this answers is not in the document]
metas-rc: done

de.metas.business/src/main/java/de/metas/customs/CustomsInvoiceService.java
```diff
@@ -0,0 +1,271 @@
...
+				.lines(customsInvoiceLines)
+				.build();
+
+		customsInvoice.updateLineNos();
+
+		return customsInvoice;
+	}
+
+	private CustomsInvoiceLine createCustomsInvoiceLine(
+			final ProductId productId,
```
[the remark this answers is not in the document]
metas-rc: done

de.metas.business/src/main/java/de/metas/customs/CustomsInvoiceService.java
```diff
@@ -0,0 +1,271 @@
...
+			final Collection<InOutAndLineId> shipmentLinesForProducts,
```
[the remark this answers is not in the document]
metas-rc: done

de.metas.business/src/main/java/de/metas/customs/CustomsInvoiceService.java
```diff
@@ -0,0 +1,271 @@
...
+			final CurrencyId currencyId)
```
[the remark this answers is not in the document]
metas-rc: done

de.metas.business/src/main/java/de/metas/customs/event/CustomsInvoiceUserNotificationsProducer.java
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	/** Topic used to send notifications about shipments/receipts that were generated/reversed asynchronously */
+	public static final Topic EVENTBUS_TOPIC = Topic.builder()
+			.name("de.metas.customs.UserNotifications")
+			.type(Type.REMOTE)
+			.build();
+
+	// services
+	private final transient IDocumentBL docActionBL = Services.get(IDocumentBL.class);
```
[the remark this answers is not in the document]
metas-ts: done

de.metas.business/src/main/java/de/metas/customs/process/ShipmentLinesForCustomsInvoiceRepo.java
```diff
@@ -0,0 +1,142 @@
...
+		{
+			return false;
+		}
+
+		if (!documentBL.isStatusStrOneOf(shipment.getDocStatus(), DocStatus.Completed.getCode(), DocStatus.Closed.getCode()))
+		{
+			return false;
+		}
+
+		if (shipment.isExportedToCustomsInvoice())
```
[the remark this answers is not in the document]
metas-ts: i like it the way it is

de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_InOut.java
```diff
@@ -365,7 +367,7 @@ private void createFacts_PurchasingReturnLine(final Fact fact, final DocLine_InO
 		if (MInOut.DOCSTATUS_Reversed.equals(m_DocStatus) && m_Reversal_ID > 0 && line.getReversalLine_ID() > 0)
 		{
 			// Set AmtAcctDr from Original Shipment/Receipt
-			if (!dr.updateReverseLine(MInOut.Table_ID, m_Reversal_ID, line.getReversalLine_ID(), BigDecimal.ONE))
+			if (!dr.updateReverseLine(getTableId(I_M_InOut.class), m_Reversal_ID, line.getReversalLine_ID(), BigDecimal.ONE))
```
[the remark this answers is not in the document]
metas-ts: i like it the way it is

de.metas.business/src/test/java/de/metas/customs/CustomsInvoiceServiceTest.java
```diff
@@ -0,0 +1,644 @@
...
+	private I_C_UOM uom1;
+
+	private I_C_UOM uom2;
+
+	private BigDecimal convertionMultiplier;
+	private BigDecimal conversionDivisor;
+
+	private ProductId product1;
+
+	private PlainCurrencyDAO currencyDAO;
```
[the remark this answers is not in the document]
metas-ts: done

de.metas.business/src/main/java/de/metas/letters/model/MADBoilerPlate.java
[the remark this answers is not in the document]
metas-ts: done

de.metas.business/src/main/java/de/metas/customs/CustomsInvoiceRepository.java
```diff
@@ -0,0 +1,229 @@
...
+
+		saveRecord(record);
+
+		line.setId(CustomsInvoiceLineId.ofRepoId(record.getC_Customs_Invoice_ID(), record.getC_Customs_Invoice_Line_ID()));
+
+	}
+
+	public CustomsInvoice updateDocActionAndStatus(@NonNull final CustomsInvoice customsInvoice)
+	{
+		final I_C_Customs_Invoice customsInvoiceRecord = load(customsInvoice.getId(), I_C_Customs_Invoice.class);
```
metas-ts: This occurs to me only now (a bit late) and i think we should go with it rather than doing another iteration, but:

As far as i understand DDD, in this case we should have one `CustomsInvoice` entity, all with lines and external references.
This repo should then provide one `save()` method and maybe one or more `getBy..(...)`. 
The other public methods like `setCustomsInvoiceToShipment`, `updateDocActionAndStatus` etc don't belong in there.
Responsibilities should be less distributed over `CustomsInvoiceRepository` and `CustomsInvoiceService`.

So, let's try to keep it in the back of our heads to next time, when writing something new or getting to touch this code again

Issue #5198, opened 2019-05-14
HU Mass Disposal process


metas-rc:
### Is this a bug or feature request?
* FR
### What is the current behavior?
* Currently we have the process "Move to garbage" which is buggy and doesn't set C_Activity_ID in the inventory
#### Which are the steps to reproduce?
* Check and see
### What is the expected or desired behavior?
* Add a new process "Mass disposal"

Concept: 

Concept

Step 1)
* Create a process similar with M_HU_MoveToGarbage ( de.metas.ui.web.handlingunits.process.WEBUI_M_HU_MoveToGarbage ) 
* The process will be named M_HU_Mass_Disposal

* the process will have 3 parameters:
   - C_Activity_ID (Kostenstelle) 
   - Description
   - IsCompleteInventory ( default true, hidden unless needed by a customer)

* The Physical Inventory that will be created will receive the C_Activity_ID and Description that were provided as parameters.
* The inventory will be completed or not, based on the isCompleteInventory parameter..

* The inventory will have HU Assignments 
   - Add the tab in the inventory window  ( window id = 341)  and take care of the logic behind it.

* The process must work for any HU type ( CU, TU or LU) as long as the HU is top-level. 
* If any selected HU is not top-level, the process will not be available

 8If the selected HUs belong to multiple warehouses, there will be one inventory / warehouse.

Step 2)  Support disposal Reactivation, Voiding. 
 *  Make sure the hu snapshots are correctly created when the inventory is created and the HUs are correctly recreated on reactivation/ voiding.

Dev hints: 

 * The old process throws a NPE when called for a CU. The reason is that the VHU case here de.metas.handlingunits.inventory.impl.InventoryAllocationDestination.retrieveTU(I_M_HU) doesn't get a parent. The hotfix would be to just return the vhu in that case but I am not sure this is right. I will have to debug this part and see.

PR #5283, opened 2019-06-06
#5198 Set the top level HU in the inventory lines

metas-rc:
https://github.com/metasfresh/metasfresh/issues/5198


de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/inventory/impl/InventoryAllocationDestination.java
```diff
@@ -185,7 +189,7 @@ public IAllocationResult load(final IAllocationRequest request)
 
 			//
 			// Get/create the inventory line based on the info from material receipt and request
-			final I_M_InventoryLine inventoryLine = getCreateInventoryLine(receiptLine, hu.getM_HU_ID(), movementDate);
+			final I_M_InventoryLine inventoryLine = getCreateInventoryLine(receiptLine, topLevelHU.getM_HU_ID(), movementDate);
```
metas-ts: Couldn't it happen that this method is called twice with two different `request`s whose respective hus have the same top level HU?
In this case we would still run into the UC, right?

metas-rc: I tested this case and it's working correctly. If a child of the same toplevel HU comes, the qty of the xisting line is modified. No new line is created

Issue #5306, opened 2019-06-12
Fallback for missing AD_WF_Node.S_Resource_ID


metas-ts:
### Is this a bug or feature request?
FR

metas-ts:
see PR-closing-comment

PR #5307, opened 2019-06-12
5306-app - Fallback for missing AD_WF_Node.S_Resource_ID

metas-ts:
If some of the `PP_Order`'s `AD_WorkFlow`'s activities don't have a resource, then attempt to fall back to the `PP_Order`'s resource or to the `PP_POrder`'s `PP_Product_Planning`'s resource for those activities.
#5306


metas-ts:
the S_Resource_ID from PP_Product_Planning is the plant; it's part of the "selector".
The S_Resource_ID from the PP_Order is laso the plant. neither of them are the actual "resource" in the sense that we have in AD_WF_Nodes

Issue #5358, opened 2019-07-03
By-Products not being shown in the HU-bin


metasnw:
### Is this a bug or feature request?
bug

mykloudz:
**Issues:**
**1.** In /Logistics/Handling Unit Editor if we are trying to Pick a HU line item and in the Pick POP-UP when we input a keyword in "Delivery Scheduling" field, the results should not include SO details for which HU line items are already picked.
**2.** Also the result should only contain those suggestions for which the order item attributes in SO are matching with HU line item we are trying to pick.
**3.** If we produce a By-Product while processing a Manufacturing Order, the same By-Product was not getting displayed in the HU bin when we want to use it as component for some other Manufacturing Order.

**Solution for Issue No. 1:** We altered **getShipmentScheduleValues** method in WEBUI_M_HU_Pick.java class and also added these methods: **getPickingCandidatesForShipmentSchedules** method in PickingCandidateService.java class, **getByShipmentScheduleIds** in PickingCandidateRepository.java class, **isShipmentSchedulePicked** method in WEBUI_M_HU_Pick class.
**Solution for Issue No. 2:** We altered **getShipmentScheduleValues** method in WEBUI_M_HU_Pick.java class and also added these methods: **getSingleHUEditorRow** in WEBUI_M_HU_Pick.java class, **doesHUAttribsMatchOrderLineAttribs** method in WEBUI_M_HU_Pick class, **doesHUAttribsMatch** method in WEBUI_M_HU_Pick class.
**Solution for Issue No. 3:** We altered **isHuIdIssued** method in HUPPOrderQtyDAO.java class.

PR #5362, opened 2019-07-04
Fix for By-Products not being shown in the HU-bin #5358

mykloudz:
#5358
This PR contains fix for By-Products not being shown in the HU-bin when we open it for issuing them as components during Manufacturing Order Processing


mykloudz:
Hi @metas-ts, we have checked the build and the fixes are working fine. Could you please merge the PRs?

metasnw:
@metas-ts a description has been added to the issue here: #5381

mykloudz:
@metas-ts and @metasnw, request you to check the PR and do the needful.

de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/pporder/api/impl/HUPPOrderQtyDAO.java
```diff
@@ -104,6 +104,7 @@ public boolean isHuIdIssued(@NonNull final HuId huId)
 		return Services.get(IQueryBL.class).createQueryBuilder(I_PP_Order_Qty.class)
 				.addOnlyActiveRecordsFilter()
 				.addEqualsFilter(I_PP_Order_Qty.COLUMN_M_HU_ID, huId)
+				.addEqualsFilter(I_PP_Order_Qty.COLUMN_PP_Order_BOMLine_ID, null)
```
metas-ts: This filter contradicts the next line, AFAIU therefor this method will never return true.
Can you please outline why the filter is needed?

mykloudz: In the next line you are applying a filter (i.e., **.addNotNull(I_PP_Order_Qty.COLUMN_PP_Order_BOMLine_ID)**) which picks records with only null values in "PP_ORDER_BOMLINE_ID" column of "PP_ORDER_QTY" table(we actually didn't understood why records corresponding to null values are only being picked here). Because of this filter By-Products produced during the Manufacturing process were not getting displayed in the HU-Bin. So to overcome that we applied this filter.

mykloudz:
Hi @metas-ts any updates on merging the PR to master? or anything left-out from our end? Please update.

Issue #5386, opened 2019-07-17
Create a fallback for auth token when using Jasper Json reports 


cristinamghita:
### Is this a bug or feature request?
FR

PR #5391, opened 2019-07-19
#5386 Create a fallback for auth token when using Jasper Json reports 

cristinamghita:
#5386 Create a fallback for auth token when using Jasper Json reports


de.metas.report/metasfresh-report-via-ssh-tunnel.launch
```diff
@@ -7,12 +7,10 @@
...
 <booleanAttribute key="org.eclipse.jdt.launching.ALLOW_TERMINATE" value="false"/>
 <mapAttribute key="org.eclipse.jdt.launching.CONNECT_MAP">
 <mapEntry key="hostname" value="localhost"/>
-<mapEntry key="port" value="18791"/>
+<mapEntry key="port" value="8791"/>
```
metas-ts: pls stick with port `18791` because it's a common practice to mal `localhost:1xxxx` to `remote:xxxx`

de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5527630_sys_gh5386_CreateJsonReportsUserAndHisAuthToken.sql
```diff
@@ -0,0 +1,26 @@
...
+INSERT INTO AD_User_Record_Access (Access,AD_Client_ID,AD_Org_ID,AD_Table_ID,AD_User_ID,AD_User_Record_Access_ID,Created,CreatedBy,IsActive,Record_ID,Updated,UpdatedBy) VALUES ('W',1000000,1000000,114,100,540004,TO_TIMESTAMP('2019-07-19 12:29:12','YYYY-MM-DD HH24:MI:SS'),100,'Y',540057,TO_TIMESTAMP('2019-07-19 12:29:12','YYYY-MM-DD HH24:MI:SS'),100)
+;
+
+update ad_user set value ='JsonReports', name ='Json Reports' where ad_user_id=540057;
+
+
+
+-- Create a default auth token for user Json Reports
+
+INSERT INTO public.ad_user_authtoken (ad_client_id, ad_org_id, ad_role_id, ad_user_authtoken_id, ad_user_id, authtoken, created, createdby, description, isactive, updated, updatedby) VALUES (1000000, 1000000, 1000000, 540001, 540057, '0d46ecc8fc8148049ddc7eb19433d3b6', '2019-07-19 09:32:54.000000', 100, null, 'Y', '2019-07-19 09:32:59.000000', 100);
```
metas-ts: how will we make sure that each instance we maintain gets their own default token?

de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5527630_sys_gh5386_CreateJsonReportsUserAndHisAuthToken.sql
```diff
@@ -0,0 +1,26 @@
+-- create user Json Reports
+
+-- 2019-07-19T09:29:11.356Z
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+INSERT INTO AD_User (AD_Client_ID,AD_Language,AD_Org_ID,AD_User_ID,Comments,Created,CreatedBy,Description,Fresh_Gift,IsAccountLocked,IsActive,IsBillToContact_Default,IsDecider,IsDefaultContact,IsFullBPAccess,IsInPayroll,IsLoginAsHostKey,IsManagement,IsMFProcurementUser,IsMultiplier,IsSystemUser,Login,LoginFailureCount,Name,NotificationType,Processing,UnlockAccount,Updated,UpdatedBy,Value) VALUES (1000000,'en_US',1000000,540057,'',TO_TIMESTAMP('2019-07-19 12:29:11','YYYY-MM-DD HH24:MI:SS'),100,'User used for authentication for Json Reports','N','N','Y','N','N','N','Y','N','N','N','N','N','Y','Json Reports',0,'Json Reports','N','N','N',TO_TIMESTAMP('2019-07-19 12:29:11','YYYY-MM-DD HH24:MI:SS'),100,'jreports')
```
metas-ts: does it need to be a "SystemUser=Y" to work? Asking because i think we should make sure that one can't actually log into the webUI with this user

de.metas.adempiere.adempiere/base/src/main/java/de/metas/security/UserAuthTokenRepository.java
```diff
@@ -96,26 +96,22 @@ private static String generateAuthTokenString()
...
-	public UserAuthToken retrieveByUserId( @NonNull final UserId userId,  @NonNull final RoleId roleId)
+	public UserAuthToken retrieveByUserIdOrNull( @NonNull final UserId userId,  @NonNull final RoleId roleId)
 	{
 		final List<I_AD_User_AuthToken> userAuthTokens = Services.get(IQueryBL.class)
 				.createQueryBuilder(I_AD_User_AuthToken.class)
 				.addOnlyActiveRecordsFilter()
 				.addEqualsFilter(I_AD_User_AuthToken.COLUMN_AD_User_ID, userId)
-				.addEqualsFilter(I_AD_User_AuthToken.COLUMN_AD_Role_ID, roleId)
-				.setLimit(2)
+				.addInArrayFilter(I_AD_User_AuthToken.COLUMN_AD_Role_ID, roleId, RoleId.ADMIN)
```
metas-ts:
```suggestion
				.addEqualsFilter(I_AD_User_AuthToken.COLUMN_AD_Role_ID, roleId)
```

I think it's dangerous to fall back to one of the most powerfull roles we have in metasfresh.
If we can't find a better solution then imho we need so create a dedicated role for the JsonReports user and specify that, same way we do specify the jasper-json-user itself

de.metas.report/report-service/src/main/java/de/metas/adempiere/report/jasper/JsonDataSourceService.java
```diff
@@ -94,6 +95,29 @@ public InputStream getInputStream(@NonNull final ReportContext reportContext) th
...
+		final RoleId roleId = Env.getLoggedRoleId();
+
+		UserAuthToken token = userAuthTokenRepo.retrieveByUserIdOrNull(userId, roleId);
+
+		//
+		// fallback to default Jasper Json user
+		if (token == null)
+		{
+			final ClientId clientId = Env.getClientId();
+			Services.get(IUserDAO.class).retrieveUserIdByValue(JSONREPORTS_USER_VALUE, clientId);
```
metas-ts: As written further up, if there is no better solution (that is not falling back to the "can-do-anything-rolle"), pls specify a dedicated role.
Also, I'm not sure if we want a fallback after all. There might be a token for the currently logged in user and role, but it migh be "too weak" for reporting. So maybe better don't fall back, but use the dedicated user&role in the first place?

de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5527630_sys_gh5386_CreateJsonReportsUserAndHisAuthToken.sql
cristinamghita: With a random generator

de.metas.adempiere.adempiere/base/src/main/java/de/metas/security/UserAuthTokenRepository.java
```diff
@@ -96,16 +96,18 @@ private static String generateAuthTokenString()
...
 
-	public UserAuthToken retrieveByUserId( @NonNull final UserId userId,  @NonNull final RoleId roleId)
+	public UserAuthToken retrieveByUserId(@NonNull final UserId userId, @NonNull final RoleId roleId)
 	{
 		final List<I_AD_User_AuthToken> userAuthTokens = Services.get(IQueryBL.class)
 				.createQueryBuilder(I_AD_User_AuthToken.class)
 				.addOnlyActiveRecordsFilter()
 				.addEqualsFilter(I_AD_User_AuthToken.COLUMN_AD_User_ID, userId)
 				.addEqualsFilter(I_AD_User_AuthToken.COLUMN_AD_Role_ID, roleId)
+		//		.setLimit(1) // we allow more than one valid token at a time
```
metas-ts:
```suggestion
```

de.metas.adempiere.adempiere/base/src/main/java/de/metas/security/RoleId.java
```diff
@@ -45,6 +46,10 @@ public static RoleId ofRepoId(final int repoId)
 		{
```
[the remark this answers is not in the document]
metas-ts: done

de.metas.adempiere.adempiere/base/src/main/java/de/metas/security/UserAuthTokenRepository.java
```diff
@@ -103,19 +103,15 @@ public UserAuthToken retrieveByUserId( @NonNull final UserId userId,  @NonNull f
 				.addOnlyActiveRecordsFilter()
```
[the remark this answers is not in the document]
metas-ts: obsolete now

PR #5403, opened 2019-07-24
5346-app - Pricing Condition fixes

metas-ts:
- GridTabVO - the actual fix: avoid exception by calling the user/role dependending UIDisplayedEntityTypes.isEntityTypeDisplayedInUIOrTrueIfNull() only if rolled shall be applied
- UserRolePermissionsKey - thorw more informative error
https://github.com/metasfresh/metasfresh/issues/5346

### DIFF
--- a/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/GridTabVO.java
+++ b/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/GridTabVO.java
@@ -210,7 +210,7 @@ public class GridTabVO implements Evaluatee, Serializable
 			// Apply role permissions
 			if(vo.applyRolePermissions)
 			{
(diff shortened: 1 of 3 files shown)


de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/GridTabVO.java
```diff
@@ -218,6 +210,14 @@ private static boolean loadTabDetails (final GridTabVO vo, final ResultSet rs)
 			// Apply role permissions
 			if(vo.applyRolePermissions)
 			{
+				// If EntityType is not displayed, hide this tab; note that this decision role-specific
```
metas-ts:
```suggestion
				// If EntityType is not displayed, hide this tab; note that this decision is role-specific
```

Issue #5444, opened 2019-08-21
Have dedicated window for organisational master data business partner


metasnw:
### Is this a bug or feature request?
FR

### What is the current behavior?
if you want to change the master data like address of your organisation to change you are updating it in window business partner on a partner that is linked to your organisation

### What is the expected or desired behavior?
have a separate window allowing to change this org business partner and do not show it in the usual business partner windows

metasnw:
![image](https://user-images.githubusercontent.com/16079655/63467828-952d6800-c466-11e9-80e3-da525220d53f.png)

PR #5446, opened 2019-08-22
#5444 Have dedicated window for organisational master data business partner 

metasnw:
#5444


metas-ts:
going to integrate after successfol jenkins build

Issue #5481, opened 2019-09-04
Some checkPreconditionsApplicable implementations mix up "all" selected and "none" selected


metas-ts:
..because in both cases, there are no *concretely* selected record IDs

### Is this a bug or feature request?
Bug

PR #5482, opened 2019-09-04
5481-app - Some checkPreconditionsApplicable implementations mix up...

metas-ts:
..."all" selected and "none" selected

introduce SelectionSize to be returned by IProcessPrecondition instead of the primitive int
fix all the code that calls IProcessPrecondition.getSize() to also take into account that possibly **all** records are selected

https://github.com/metasfresh/metasfresh/issues/5481


metas-ts:
**PLEASE DON'T DELETE THE BRANCH.** It's a release branch.

PR #5547, opened 2019-09-18
default ordering added in accounting transactions window

cristinastefan90:
Windows default order by in grid view


metas-ts:
OK to integrate after sugccessful build

Issue #5536, opened 2019-09-17
Allow possibility to run TU Label from BPartner


cristinamghita:
### Is this a bug or feature request?
FR

PR #5561, opened 2019-09-23
#5536 Allow possibility to run TU Label from BPartner 

cristinamghita:
#5536 Allow possibility to run TU Label from BPartner

### DIFF
--- a/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/C_BPartner_Location_Label.sql
+++ b/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/C_BPartner_Location_Label.sql
@@ -18,7 +18,7 @@ FROM generate_series(1, copies) i
              (
              SELECT
                org.Name                                                                                as org_name,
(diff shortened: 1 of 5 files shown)


de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/C_BPartner_Location_Label.sql
```diff
@@ -0,0 +1,41 @@
...
+  )
+AS
+$$
+SELECT t.*
+FROM generate_series(1, copies) i
+  CROSS JOIN LATERAL
+             (
+             SELECT
+               org.Name                                                                                as org_name,
+               COALESCE(NULLIF(bp_loc.gln, ''),
```
metas-ts:
```suggestion
               COALESCE(NULLIF(TRIM(bp_loc.gln), ''),
```
will also catch the case where a user entered some spaces but *think* they left the GLN empty

cristinamghita: @metas-ts Agree!

Issue #5566, opened 2019-09-24
Add integration with Google Maps


TheBestPessimist:
### Is this a bug or feature request?
FR

TheBestPessimist:
# Documentation

```
dev-instance: https://dev758.metasfresh.com
login: dev
version: 5.123.2_17431_gh5566
```


## 10. Geocoding Config dialog
![10  Geocoding Config dialog](https://user-images.githubusercontent.com/4482210/66029985-0d2c8a80-e509-11e9-83b3-346e290eb1bd.png)



## 20. Add new Config (only 1 should be added) gives 2 options-- Google Maps, or Open Street Maps
![20  Add new Config (only 1 should be added) gives 2 options-- Google Maps, or Open Street Maps](https://user-images.githubusercontent.com/4482210/66029992-1158a800-e509-11e9-9f44-696f8e38404c.png)



## 30. Google Maps default settings -- the user has to add the API key
![30  Google Maps default settings -- the user has to add the API key](https://user-images.githubusercontent.com/4482210/66029999-161d5c00-e509-11e9-9792-50989fbf1120.png)



## 40. Open Street Maps default settings
![40  Open Street Maps default settings](https://user-images.githubusercontent.com/4482210/66030004-19184c80-e509-11e9-82c4-727716cc0c08.png)


## 50. Select Google Maps or Open Street Maps as provider



## 51. Test Geocoding Update on the hidden Location window with id 121
![50  Test Geocoding Update on the hidden Location window with id 121](https://user-images.githubusercontent.com/4482210/66030036-2f260d00-e509-11e9-9d59-87f361c51c6a.png)


## 60. Run action Geocoding Update
![60  Run action Geocoding Update](https://user-images.githubusercontent.com/4482210/66030042-33eac100-e509-11e9-99ae-6e314deff071.png)



## 65. You are informed that some locations are enqueued
![65  You are informed that some locations are enqueued](https://user-images.githubusercontent.com/4482210/66030080-42d17380-e509-11e9-931f-6b9415d6cd7c.png)



## 70. The location is (eventually) updated with Geographical Coordinates
![70  The location is (eventually) updated with Geographical Coordinates](https://user-images.githubusercontent.com/4482210/66030085-449b3700-e509-11e9-9cf4-d5b69c8edd17.png)

TheBestPessimist:
# Endpoint for retrieving the geolocation config

```
GET /rest/api/geolocation/config
```
[...]

PR #5597, opened 2019-10-01
 Add integration with Google Maps

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/5566

### DIFF
--- a/de.metas.adempiere.adempiere/base/src/main/java/de/metas/location/geocoding/GeocodingConfig.java
+++ b/de.metas.adempiere.adempiere/base/src/main/java/de/metas/location/geocoding/GeocodingConfig.java
@@ -1,10 +1,3 @@
-package de.metas.location.geocoding;
-
-import de.metas.location.geocoding.provider.GeocodingProviderName;
-import lombok.Builder;
-import lombok.NonNull;
-import lombok.Value;
-
 /*
  * #%L
  * de.metas.adempiere.adempiere.base
@@ -15,17 +8,24 @@ import lombok.Value;
  * it under the terms of the GNU General Public License as
  * published by the Free Software Foundation, either version 2 of the
  * License, or (at your option) any later version.
- * 
+ *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
  * GNU General Public License for more details.
- * 
+ *
  * You should have received a copy of the GNU General Public
  * License along with this program. If not, see
  * <http://www.gnu.org/licenses/gpl-2.0.html>.
  * #L%
  */
+
+package de.metas.location.geocoding;
+
+import de.metas.location.geocoding.provider.GeocodingProviderName;
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
 
 @Value
 @Builder
@@ -41,8 +41,7 @@ public class GeocodingConfig
 	@Builder
 	public static class OpenStreetMapsConfig
 	{
-		@NonNull
-		final String baseURL;
+		@NonNull final String baseURL;
 		int cacheCapacity;
 		long millisBetweenRequests;
 	}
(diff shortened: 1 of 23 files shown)


de.metas.adempiere.adempiere/base/src/main/java/de/metas/location/geocoding/provider/GeocodingProviderFactory.java
```diff
@@ -0,0 +1,107 @@
...
+			return createGoogleMapsProvider(config.getGoogleMapsConfig());
+		}
+		else if (GeocodingProviderName.OPEN_STREET_MAPS.equals(providerName))
+		{
+			return createOSMProvider(config.getOpenStreetMapsConfig());
+		}
+		else
+		{
+			throw new AdempiereException("Unknown provider: " + providerName);
+		}
```
metas-ts: you could have done this in a switch..might be just a matter of tase though

de.metas.adempiere.adempiere/base/src/main/java/de/metas/location/geocoding/GeocodingConfig.java
```diff
@@ -0,0 +1,58 @@
...
+package de.metas.location.geocoding;
+
+import de.metas.location.geocoding.provider.GeocodingProviderName;
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
+
+@Value
+@Builder
+public class GeocodingConfig
```
metas-ts: I * think* I would have made GeocodingConfig an interface with just the getProviderName method.
But this way it's also OK for me..but might be something to consider when/if we get the 3rd provider

de.metas.adempiere.adempiere/base/src/test/java/de/metas/location/geocoding/asynchandler/LocationGeocodeEventHandlerTest.java
```diff
@@ -0,0 +1,162 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.math.BigDecimal;
+import java.util.Optional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Must use jUnit 4 because our version of junit5 and jmockit are incompatible and an error is thrown.
```
metas-ts: just as a sidenote: when i mock nowadaws, i tend to use mockito; it's better integrates with spring (and afair junit5).

de.metas.adempiere.adempiere/base/src/test/java/de/metas/location/geocoding/provider/openstreetmap/NominatimOSMGeocodingProviderImplTest.java
```diff
@@ -114,7 +110,6 @@ void bestCoordinatesFindsMetasDEByFullAddress()
 				.contains(expectedCoordinates);
 	}
 
-
 	@Test
 	@DisplayName("bestCoordinates cannot find metasRO by postal and wrong country")
```
metas-ts: nice, i wasn't aware of that annotation

de.metas.adempiere.adempiere/base/src/main/java/de/metas/location/geocoding/provider/GeocodingProviderFactory.java
TheBestPessimist: very good point!

de.metas.adempiere.adempiere/base/src/main/java/de/metas/location/geocoding/GeocodingConfig.java
TheBestPessimist: cannot guarantee i'll remember this advice until next time 😊

de.metas.adempiere.adempiere/base/src/test/java/de/metas/location/geocoding/asynchandler/LocationGeocodeEventHandlerTest.java
TheBestPessimist: will update the test when i get back to this, during the integration with frontend.

de.metas.adempiere.adempiere/base/src/test/java/de/metas/location/geocoding/provider/openstreetmap/NominatimOSMGeocodingProviderImplTest.java
TheBestPessimist: only available in junit5, but it's nice coz the IDE shows the "display name" instead of the test name.

Issue #5606, opened 2019-10-02
Payment window calculation of skonto yes/ no when adjusting date


cristinastefan90:
### Is this a bug or feature request?
Bug

PR #5607, opened 2019-10-02
#5606 Payment window calculation of skonto yes/ no when adjusting date

cristinastefan90:
https://github.com/metasfresh/metasfresh/issues/5606


de.metas.business/src/main/java/de/metas/payment/api/IPaymentDAO.java
```diff
@@ -100,4 +101,13 @@
...
 	Stream<PaymentId> streamPaymentIdsByBPartnerId(BPartnerId bpartnerId);
+
+	/**
+	 * Updates the discount and the payment based on DateTrx and the payment term policy.
+	 * 
+	 * @param payment
+	 * @param c_Invoice_ID
+	 * @param c_DocType
+	 */
+	void updateDiscountAndPayment(I_C_Payment payment, int c_Invoice_ID, I_C_DocType c_DocType);
```
metas-ts: pls remove those superflous "@param ..." java doc lines
(they would be OK if they added important information)

de.metas.business/src/main/java/de/metas/payment/api/impl/AbstractPaymentDAO.java
```diff
@@ -181,4 +188,68 @@ public BigDecimal getInvoiceOpenAmount(I_C_Payment payment, final boolean credit
 				.listIds(PaymentId::ofRepoId)
 				.stream();
 	}
+	@Override
+	public void updateDiscountAndPayment(I_C_Payment payment,int c_Invoice_ID, I_C_DocType c_DocType)
```
metas-ts: please add an abstract method `updateDiscountAndPayment` 6to this class, and 
* move the implementation to `PaymentDAO`
* also add an implementation to `PlainPaymentDAO` (which dan for now just throw an apache-commons `NotImplementedException`)

Background: when running in unit test mode, Services.get() will by default return `PlainPaymentDAO`, which can contain some mockup code that doesn't connect to DB.

de.metas.business/src/main/java/de/metas/payment/api/impl/PaymentDAO.java
```diff
@@ -23,15 +23,23 @@
  */
 
 import java.math.BigDecimal;
+import java.sql.PreparedStatement;
```
metas-ts: i would ask you to clean up this imports, but i guess they will be needed when the updateDiscountAndPayment implementation is back :-)

de.metas.business/src/main/java-legacy/org/compiere/model/MPayment.java
```diff
@@ -737,7 +737,7 @@ public void setWriteOffAmt(final BigDecimal WriteOffAmt)
 	public void setOverUnderAmt(final BigDecimal OverUnderAmt)
 	{
 		super.setOverUnderAmt(OverUnderAmt == null ? BigDecimal.ZERO : OverUnderAmt);
-		setIsOverUnderPayment(getOverUnderAmt().compareTo(BigDecimal.ZERO) != 0);
+//		setIsOverUnderPayment(getOverUnderAmt().compareTo(BigDecimal.ZERO) != 0);
```
metas-ts:
```suggestion
```

de.metas.business/src/main/java/de/metas/payment/api/impl/AbstractPaymentDAO.java
```diff
@@ -181,4 +188,68 @@ public BigDecimal getInvoiceOpenAmount(I_C_Payment payment, final boolean credit
 				.listIds(PaymentId::ofRepoId)
 				.stream();
 	}
+	@Override
```
metas-ts:
```suggestion
/*
TODO please consider the following improvement
- create an AD_Table like `C_InvoiceOpenAmounts` that has the required values (`C_BPartner_ID`, `C_Currency_ID`, `InvoiceOpen`...) as columns
- put this select-stuff into a DB function. that function shall return the `C_InvoiceOpenAmounts` as result (i.e. the metasfresh `C_InvoiceOpenAmounts` table is not a physical table in the DB; it's just what the DB-function returns)
- create a model class `I_C_InvoiceOpenAmounts` for your AD_Table.
- have the `PaymentDAO` implementation invoke the DB-function to get it's `I_C_InvoiceOpenAmounts` (=> you can do this with IQueryBL)
- have the `PlainPaymentDAO ` implementation return "plain" instances of `I_C_InvoiceOpenAmounts`
  - that way, one can write a unit test where they first create one or two plain `I_C_InvoiceOpenAmounts`s and then query them in their test
*/
	@Override
```

Issue #5645, opened 2019-10-09
BOM Components followup - Comment, Change notice and Expected result to be added


cristinastefan90:
### Is this a bug or feature request?
FR

PR #5646, opened 2019-10-09
#5645 BOM Components followup - Comment, Change notice and Expected result to be added 

cristinastefan90:
https://github.com/metasfresh/metasfresh/issues/5645


metas-ts:
can be integrated when build by jenkins

Issue #5659, opened 2019-10-14
Picklist in Picking Terminal v2 automatically picks, but shall not 


cristinamghita:
### Is this a bug or feature request?
Bug

PR #5660, opened 2019-10-14
#5659 Implement the possibility to just create picking candidates without really pick them

cristinamghita:
#5659 Picklist in Picking Terminal v2 automatically picks, but shall not

### DIFF
--- a/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/candidate/commands/CreatePickingCandidatesCommand.java
+++ b/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/candidate/commands/CreatePickingCandidatesCommand.java
@@ -1,0 +1,100 @@
+package de.metas.handlingunits.picking.candidate.commands;
+
+import org.adempiere.ad.trx.api.ITrxManager;
+import org.compiere.model.I_C_UOM;
+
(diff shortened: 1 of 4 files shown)


de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/candidate/commands/CreatePickingCandidatesCommand.java
```diff
@@ -0,0 +1,100 @@
...
+		this.pickingCandidateRepository = pickingCandidateRepository;
+
+		this.shipmentScheduleId = request.getShipmentScheduleId();
+		this.pickFromHuId = request.getPickFromHuId();
+		this.pickingSlotId = request.getPickingSlotId();
+	}
+
+	public PickHUResult perform()
+	{
+		return trxManager.callInThreadInheritedTrx(this::performInTrx);
```
metas-ts: Do we need `callInThreadInheritedTrx` here?
If we *really* need it, please add a little comment or javadoc about "why"

cristinamghita: Good piint! We do not need it!

Issue #5740, opened 2019-10-31
Improve Customs Tariff


cristinamghita:
### Is this a bug or feature request?
FR

PR #5745, opened 2019-10-31
#5740 Improve Customs Tariff 

cristinamghita:
#5740 Improve Customs Tariff


cristinamghita:
> are we sure that "Allow creating multiple customs invoice for the same shipment" ( [2fd9d07](https://github.com/metasfresh/metasfresh/commit/2fd9d07ac92186bd91913bc3b20b79b752f575c1) ) is OK?
> Otherwise OK for me, feel free to integrate when build

@metas-ts yes; Was an explicit request from @metas-mk

Issue #5522, opened 2019-09-13
Commission calculation


metas-ts:
### Is this a bug or feature request?
FR

metas-ts:
# Dev-Test WIP

## new product:

there is a "system-product" for commission settlement.

https://yourhost.metasfresh.com/window/140/540420

<details>

![image](https://user-images.githubusercontent.com/1264789/67418969-166ebb80-f5cc-11e9-9efd-1fc4243cdb12.png)

</details>

## new in bpartner

PR #5751, opened 2019-11-01
5522-app - Commission calculation

metas-ts:
- do precise serializing/deserializing of Instants (toString / parse was useless)
- in commission facts, display "Updated" instead of the now useful but not anymore userfriendly timestamp-string
#5522

### DIFF
--- a/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/util/TimeUtil.java
+++ b/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/util/TimeUtil.java
@@ -1839,4 +1839,50 @@ public class TimeUtil
 			return date.toInstant().atZone(zoneId);
 		}
 	}
+
+	/**
+	 * Counterpart of {@link #deserializeInstant(String)}.
+	 */
+	public static String serializeInstant(@NonNull final Instant instant)
+	{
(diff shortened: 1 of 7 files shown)


de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/util/TimeUtil.java
```diff
@@ -1839,4 +1839,50 @@ public static ZonedDateTime convertToTimeZone(@NonNull final ZonedDateTime date,
...
+	public static String serializeInstant(@NonNull final Instant instant)
+	{
+		return Long.toString(instant.getEpochSecond()) + "." + Long.toString(instant.getNano());
+	}
+
+	/**
+	 * Deserializes a string such as {@code "12345.234567"} into an {@link Instant}.
+	 * <p>
+	 * Notes:
+	 * <li>I didn't want to use jackson because this is not inteded to serialize/deserialize a whole object-tree, but that to play a role in very concrete business logic, so i want it to be more transparent how this is done.
```
metas-ts:
```suggestion
	 * <li>I didn't want to use jackson because this is not inteded to serialize/deserialize a whole object-tree, but be called from very concrete business logic, so i want it to be more transparent how this is done.
```

Issue #5637, opened 2019-10-07
Add DPD Shipper and DPD integration


metas-rc:
### Is this a bug or feature request?
* FR
### What is the current behavior?
*
#### Which are the steps to reproduce?
*
### What is the expected or desired behavior?
* ToDo:

 * API DPD https://www.dpd.com/nl_en/business_customers/help_en_tools/labelprint_tools/dpd_integration_services in metasfresh
 * logistics order, label generation, status information workflow
 * receive status information for the package, URL for package tracking
 * quick reference from order to package(s)

PR #5791, opened 2019-11-15
 Add DPD Shipper and DPD integration

TheBestPessimist:
Dpd Products: Use E12 for same-country and CLASSIC for international shipping 

https://github.com/metasfresh/metasfresh/issues/5637


TheBestPessimist:
>  i saw a bunch of commented code and things that say WIP, but i assume it's just like that and you will in fact continue to work at those places, so it's fine with me

Yes, it's related to DE->US shipping which depends on https://github.com/metasfresh/me03/issues/3157

PR #5826, opened 2019-11-26
Delivery contract with same product but different attributes

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/5823

### DIFF
--- a/de.metas.procurement.base/src/main/java-gen/de/metas/procurement/base/model/I_PMM_Balance.java
+++ b/de.metas.procurement.base/src/main/java-gen/de/metas/procurement/base/model/I_PMM_Balance.java
@@ -32,10 +32,6 @@ public interface I_PMM_Balance
 	 */
 	public int getAD_Client_ID();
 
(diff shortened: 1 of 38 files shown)


metas-ts:
closing the PR without integration:
* https://github.com/metasfresh/metasfresh/pull/5826/commits/1e38fc037ec00a2e61532c185f78b86c47dc5911: there won't be a (e2e-)test for the changes in some time, and we can't rule out that the code change e.g. introduces NPEs or other glitches

* https://github.com/metasfresh/metasfresh/pull/5826/commits/3cefa39c0b4d1ab6b7de52a8e8293252569a2437: as i wrote
> Can the customer freely pick the  PMM_Product.ProductName ?
>* if yes => remove ASI_ID from Identifiercolumns
>* if no => remove ASI_ID from Identifiercolumns and add SeqNo isntead

..productname is read-only, and set in `de.metas.procurement.base.impl.PMMProductNameBuilder`), so i don'T see a benefit in adding SeqNo

Issue #5834, opened 2019-11-27
Sales Order recording without Billto Partner


TheBestPessimist:
### Is this a bug or feature request?
FR

### What is the expected or desired behavior?

ToDo:

- Don't initialize the data when BillTo is missing
- Show a translated (german) Error message
- After creating the missing Business Partner data allow recording the data further without moving.

When entering an order for a business partner without an associated invoice partner, an English error message is displayed and the entered data is deleted again.

PR #5843, opened 2019-11-27
 Dont throw an error notification in case no Bpartner bill-to address exists, but only show an error message over the field

TheBestPessimist:
In this way the bpartner field is no longer cleared (when the transaction is rolled back by the thrown exception)

https://github.com/metasfresh/metasfresh/issues/5834

### DIFF
--- a/de.metas.business/src/main/java-legacy/org/compiere/model/CalloutOrder.java
+++ b/de.metas.business/src/main/java-legacy/org/compiere/model/CalloutOrder.java
@@ -314,7 +314,11 @@ public class CalloutOrder extends CalloutEngine
 
 		if (!orderBL.setBillLocation(order))
 		{
-			throw new BPartnerNoBillToAddressException(bpartner);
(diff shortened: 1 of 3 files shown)


de.metas.business/src/main/java-legacy/org/compiere/model/CalloutOrder.java
```diff
@@ -314,7 +314,11 @@ public String bPartnerLocation(final ICalloutField calloutField)
 
 		if (!orderBL.setBillLocation(order))
 		{
-			throw new BPartnerNoBillToAddressException(bpartner);
+			final String localizedMessage = new BPartnerNoBillToAddressException(bpartner).getLocalizedMessage();
+			calloutField.fireDataStatusEEvent(
```
metas-ts: wtf, i had no clue we can do this :-)

TheBestPessimist: yeah i blame teo as well, he's the one who showed this trickery :D

PR #5859, opened 2019-12-02
3365 Translated the Order/Location and Product/Partner tabs.

dragospodariu96:
https://github.com/metasfresh/me03/issues/3365


metas-ts:
the build didn't succeed; see https://jenkins.metasfresh.com/job/metasfresh-dist/job/master/18518/console
Hint: you probably need to convert the SQL top UTF-8

PR #5875, opened 2019-12-06
Reactivate Shipment, update Shipmentline delivery catch weight => does not update invoice candidate

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/5874

### DIFF
--- a/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandInvalidUpdater.java
+++ b/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandInvalidUpdater.java
@@ -163,7 +163,7 @@ import static org.adempiere.model.InterfaceWrapperHelper.newInstance;
 		final ICUpdateResult result = new ICUpdateResult();
 		try (final IAutoCloseable updateInProgressCloseable = invoiceCandBL.setUpdateProcessInProgress())
 		{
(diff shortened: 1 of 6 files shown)


de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandInvalidUpdater.java
```diff
@@ -356,23 +354,23 @@ private void populateC_InvoiceCandidate_InOutLine(
 		final List<I_M_InOutLine> inoutLines = inOutDAO.retrieveLinesForOrderLine(orderLine, I_M_InOutLine.class);
 		for (final I_M_InOutLine inOutLine : inoutLines)
 		{
-			if (invoiceCandDAO.existsInvoiceCandidateInOutLinesForInvoiceCandidate(ic, inOutLine))
+			// create a new PO or update the unique existing one
+			I_C_InvoiceCandidate_InOutLine iciol = invoiceCandDAO.retrieveInvoiceCandidateInOutLine(ic, inOutLine);
+			if (iciol == null)
```
metas-ts: here is the fix, right?

TheBestPessimist: Yes: the problem was that existing `I_C_InvoiceCandidate_InOutLine` were not updated with new/updated catch weight data.

PR #5939, opened 2019-12-18
Gh5933 Translate "Posting Status"

metas-rc:
https://github.com/metasfresh/metasfresh/issues/5933


metas-ts:
not yet build, but the SQL looks legit and i hope it won't fail

Issue #160, opened 2016-07-06
work around the classcastexception in java2d


metas-ts:
It's about http://stackoverflow.com/questions/34188495/how-can-i-work-around-the-classcastexception-in-java2d-bug-id-7172749

metas-ts:
Internal: FRESH-498

PR #6010, opened 2020-01-08
#160 Updated Shipment Disposition Export Function.

dragospodariu96:
https://github.com/metasfresh/ic114/issues/160


metas-ts:
Note: i checked https://github.com/metasfresh/metasfresh/blob/gh160_4/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/ddl/functions/shipmentDispositionExcelDownload.sql
and it seems that  it's the same version as `5540900_sys_gh160_updatedPublicShipmentDispositionExportFunction.sql `from this PR.

@dragospodariu96 is that the case?
If not, you can push the correct version of  `.../ddl/functions/shipmentDispositionExcelDownload.sql` directly to `master` jsut pls let me know when you pushed.

Issue #5987, opened 2020-01-06
Multiple Customs Invoices for the same Shipment


metas-rc:
### Is this a bug or feature request?
* FR

PR #6020, opened 2020-01-10
Gh5987 4 

metas-rc:
* set the exported flag in shipment after customs invoice created
* cleanup in db and interfaces ( delete the old columns that are no longer used)

https://github.com/metasfresh/metasfresh/issues/5987

### DIFF
--- a/de.metas.business/src/main/java/de/metas/customs/CustomsInvoiceRepository.java
+++ b/de.metas.business/src/main/java/de/metas/customs/CustomsInvoiceRepository.java
@@ -13,7 +13,6 @@ import java.util.Set;
 import org.adempiere.ad.dao.IQueryBL;
 import org.compiere.model.I_C_Customs_Invoice;
 import org.compiere.model.I_C_Customs_Invoice_Line;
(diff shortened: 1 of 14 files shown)


de.metas.business/src/main/java/de/metas/customs/CustomsInvoiceRepository.java
```diff
@@ -382,4 +385,18 @@ private static CustomsInvoiceLineId extractCustomsInvoiceLineId(final I_M_InOutL
 		return CustomsInvoiceLineId.ofRepoId(record.getC_Customs_Invoice_ID(), record.getC_Customs_Invoice_Line_ID());
 	}
 
+	public void setExportedInCustomsInvoice(final InOutId shipmentId)
```
metas-ts: A repository should be about loading and storing entities. Thi that regard, this repo is already pretty much f'ed up, but we should not make things worse. Therefore, i would say please move this method to `IInoutDAO`

metas-rc: done

PR #6037, opened 2020-01-14
#6031 Handle Schema Modification in Version

metas-rc:
https://github.com/metasfresh/metasfresh/issues/6031

### DIFF
--- a/de.metas.business/src/main/java/de/metas/phonecall/service/PhonecallSchemaRepository.java
+++ b/de.metas.business/src/main/java/de/metas/phonecall/service/PhonecallSchemaRepository.java
@@ -1,6 +1,8 @@
 package de.metas.phonecall.service;
 
+import static org.adempiere.model.InterfaceWrapperHelper.getTableId;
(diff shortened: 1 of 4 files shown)


de.metas.business/src/main/java/de/metas/phonecall/service/PhonecallSchemaRepository.java
```diff
@@ -290,4 +293,38 @@ else if (schemaVersionRecord.isMovePhonecallDay())
...
+		}
+
+		schemasToInvalidate.stream()
+				.forEach(schemaId -> schemas.resetForRecordId(TableRecordReference.of(phonecallSchemaTableId, schemaId)));
+
+		schemas.resetForRecordId(TableRecordReference.of(phonecallSchemaTableId, phonecallSchemaId.getRepoId()));
+
+	}
+
+	private List<I_C_Phonecall_Schema_Version_Line> retrieveLinesWithDifferentSchemas(PhonecallSchemaVersionId phonecallSchemaVersionId)
```
metas-ts: doesn't this method return all lines?

PR #6080, opened 2020-01-23
6079-app - OLCand fix handover and dropship processing

metas-ts:
WIP:
* don't return bpartner, locaiton and contact one by one, but return a bpartnerInfo instance
* return handover resp. dropship info a Optional, so the caller itself can decide whether to fallback in case there is no explicitly specified handover/dropship

https://github.com/metasfresh/metasfresh/issues/6079


metas-ts:
`master` is the wrong target branch..

Issue #6125, opened 2020-02-04
Docaction Permission not taken properly for Bank Statement


dragospodariu96:
### Is this a bug or feature request?
BUG
### What is the current behavior?
Currently, when adding, changing or removing document action permissions for a role, the changes are not seen when logging in under that role, unless a cache reset is done.
#### Which are the steps to reproduce?
Login with a user with access to the Role window, go to the role window, and add, modify or remove an available document action permission for another role. Then login to that role and the changes are not seen until a cache reset is done.
### What is the expected or desired behavior?
The changes should be seen without manually reseting the cache.

PR #6189, opened 2020-02-14
#6125 updated bank statement doc type.

dragospodariu96:
https://github.com/metasfresh/metasfresh/issues/6125

### DIFF
--- a/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5552160_sys_gh6125UpdatedBankStatementDocType.sql
+++ b/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5552160_sys_gh6125UpdatedBankStatementDocType.sql
@@ -1,10 +1,10 @@
 update c_bankstatement
-set c_doctype_id = (select dt.c_docType_id 
-					from c_doctype dt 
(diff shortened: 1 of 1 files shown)


de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5552160_sys_gh6125UpdatedBankStatementDocType.sql
```diff
@@ -0,0 +1,10 @@
+update c_bankstatement
+set c_doctype_id = (select dt.c_docType_id 
+					from c_doctype dt 
+					where (dt.ad_org_id=c_bankstatement.ad_org_id or dt.ad_org_id = 0)
+					and dt.DocbaseType = 'CMB' 
+					and dt.DocSubType= 'BS' 
+					limit 1),
+	updatedBy = 99,
+	Updated=(select now())
```
metas-ts:
```suggestion
	Updated=('2020-02-14 09:15:32.812324+00') /*literal rsult of select now()*/
```

with
 
> the timestamp i just got from select now() the important thing is that you have one timestamp in the SQL

i meant to have the *literal* timestamp in the SQL. it's very crude, but that way we might have a chance of identifying the file via fulltext-search

Issue #6201, opened 2020-02-17
Support packing instructions for invoice line 'Batch entry'


pvpurcarcosmin:
### Is this a bug or feature request?
FR

PR #6202, opened 2020-02-17
Support packing instructions for invoice line 'Batch entry'

pvpurcarcosmin:
https://github.com/metasfresh/metasfresh/issues/6201

### DIFF
--- a/de.metas.swat/de.metas.swat.base/src/main/sql/postgresql/system/48-de.metas.order/5552410_sys_gh6201_add_packing_item_validation_rule_for_invoiceLine.sql
+++ b/de.metas.swat/de.metas.swat.base/src/main/sql/postgresql/system/48-de.metas.order/5552410_sys_gh6201_add_packing_item_validation_rule_for_invoiceLine.sql
@@ -1,0 +1,54 @@
+-- 2020-02-17T13:14:26.869Z
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+INSERT INTO AD_Val_Rule (AD_Client_ID,AD_Org_ID,AD_Val_Rule_ID,Code,Created,CreatedBy,EntityType,IsActive,Name,Type,Updated,UpdatedBy) VALUES (0,0,540480,'-- GENERAL
+M_HU_PI_Item_Product.AD_Org_ID IN (0, @AD_Org_ID@)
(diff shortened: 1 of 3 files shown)


de.metas.swat/de.metas.swat.base/src/main/sql/postgresql/system/48-de.metas.order/5552410_sys_gh6201_add_packing_item_validation_rule_for_invoiceLine.sql
```diff
@@ -0,0 +1,54 @@
...
+(
+	-- AUFTRAG
+	(
+		''@IsSOTrx@'' = ''Y''
+	)
+
+	-- BESTELLUNG
+	OR
+	(
+		''@IsSOTrx@'' = ''N''
```
metas-ts: Why do we need all this pricing stuff this only for purchase invoices and not for sales invoices?
Spoiler: we do need it for both :-) !

Background:
in sales orders, the packaging is decided on shipment-time. So the sales order does not have to worry about what packaging will be shippend and invoiced later. Only for purchase orders we need the packaging products' prices on onrder time.

However AFAIU
the invoice line *does* need to worry if the invoiced packaing product has a price!

de.metas.swat/de.metas.swat.base/src/main/sql/postgresql/system/48-de.metas.order/5552410_sys_gh6201_add_packing_item_validation_rule_for_invoiceLine.sql
```diff
@@ -0,0 +1,54 @@
+-- 2020-02-17T13:14:26.869Z
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+INSERT INTO AD_Val_Rule (AD_Client_ID,AD_Org_ID,AD_Val_Rule_ID,Code,Created,CreatedBy,EntityType,IsActive,Name,Type,Updated,UpdatedBy) VALUES (0,0,540480,'-- GENERAL
+M_HU_PI_Item_Product.AD_Org_ID IN (0, @AD_Org_ID@)
```
metas-ts: replace `@M_Product_ID@` with `@M_Product_ID/0@` to avoid errors in the log if this val-rule is run before a product was selected.
Make sure with @dragospodariu96 you both know wtf the `/0` is about

Issue #6214, opened 2020-02-19
Jasper Report: Customer/ Vendor Kontenblatt (Account details)


TheBestPessimist:
### Is this a bug or feature request?
FR

PR #6249, opened 2020-02-24
Jasper Report: Customer/ Vendor Kontenblatt (Account details)

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/6214

### DIFF
--- a/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/BusinessPartnerAccountSheetReport.sql
+++ b/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/BusinessPartnerAccountSheetReport.sql
@@ -1,11 +1,12 @@
-DROP FUNCTION IF EXISTS BusinessPartnerAccountSheetReport(p_c_bpartner_id numeric, p_dateFrom date, p_dateTo date, p_ad_client_id numeric, p_ad_org_id numeric, p_isSoTrx TEXT);
+DROP FUNCTION IF EXISTS BusinessPartnerAccountSheetReport(p_c_bpartner_id numeric, p_dateFrom date, p_dateTo date, p_ad_client_id numeric, p_ad_org_id numeric, p_isSoTrx TEXT, p_ad_language text);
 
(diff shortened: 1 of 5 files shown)


de.metas.acct.base/src/main/sql/postgresql/ddl/functions/BusinessPartnerAccountSheetReport.sql
```diff
@@ -0,0 +1,250 @@
...
+                 FROM temp_BusinessPartnerAccountSheetReport t
+                 ORDER BY t.dateacct, t.created, t.documentno;
+END;
+$BODY$
+    LANGUAGE plpgsql
+    VOLATILE;
+
+
+/*
+How to run:
```
metas-ts: please add this as a comment to the function; it's more accessible; feel free to add more infos. you can also add the public issue's URL there

de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/ddl/functions/getBPOpenAmtToDate.sql
```diff
@@ -22,7 +22,7 @@ BEGIN
                                        p_DateType := (
```
metas-ts: does `invoiceOpenToDate()` properly deal with credit memos?

de.metas.acct.base/src/main/sql/postgresql/ddl/functions/BusinessPartnerAccountSheetReport.sql
TheBestPessimist: done.

de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/ddl/functions/getBPOpenAmtToDate.sql
TheBestPessimist: If you refer to negating the amount in case the docbasetype is of credit memo, then no, it does not.

I don't think we should fix that function in this issue, i think we should make a new issue for the fix. We should also create a real example and set what we expect and what should and shouldn't happen.

PR #6349, opened 2020-03-13
#6348 avoid null qty in pricing context

metas-rc:
https://github.com/metasfresh/metasfresh/issues/6348

### DIFF
--- a/de.metas.business/src/main/java/de/metas/pricing/service/impl/PricingBL.java
+++ b/de.metas.business/src/main/java/de/metas/pricing/service/impl/PricingBL.java
@@ -113,18 +113,22 @@ public class PricingBL implements IPricingBL
 		pricingCtx.setBPartnerId(bPartnerId);
 		pricingCtx.setConvertPriceToContextUOM(true); // backward compatibility
 
+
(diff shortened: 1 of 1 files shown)


de.metas.business/src/main/java/de/metas/pricing/service/impl/PricingBL.java
```diff
@@ -113,18 +113,22 @@ public IEditablePricingContext createInitialContext(
...
+
 		if (quantity != null)
 		{
 			if (quantity.signum() != 0)
 			{
 				pricingCtx.setQty(quantity.toBigDecimal());
 			}
+
 			pricingCtx.setUomId(quantity.getUomId());
 		}
```
metas-ts:
```suggestion
		if (quantity == null)
		{			
			pricingCtx.setQty(ONE);
		} 
		else 
		{
			pricingCtx.setQty(quantity.isZero() ? ONE: quantity.toBigDecimal());
			pricingCtx.setUomId(quantity.getUomId());
		}
```

if think this should be equivalent, pls check. at any rate, i think it's clearer

Issue #6121, opened 2020-02-04
Jasper Report: `Account Sheet`


TheBestPessimist:
### Is this a bug or feature request?
FR

TheBestPessimist:
# Trying to figure out why `Account Sheet` returns different balance compared to `Saldobilanz`



`Saldobilanz` uses the sql function `acctBalanceToDate` to compute the balance, so all results of `Account Sheet` should match those.

### Saldobilanz report with date 2020-02-01

Amount is `-3'344'306.88`.

Same result as
```sql
SELECT *
FROM
    de_metas_acct.acctBalanceToDate(1002995,
                                    1000000,
                                    '2020-02-01'::date,
                                    1000000,
                                    'N',
                                    'N');
```




### Saldobilanz report with date 2020-02-02

Amount is `-3'353'615.50`

Same result as
```sql
SELECT *
FROM
    de_metas_acct.acctBalanceToDate(1002995,
                                    1000000,
                                    '2020-02-02'::date,
                                    1000000,
                                    'N',
                                    'N');
```


### Account Sheet for period 2020-02-01 - 2020-02-02


![image](https://user-images.githubusercontent.com/4482210/76495610-e53e3d00-643f-11ea-9bf4-c3d79753e011.png)


It seems that `Account sheet` is adding that credit (haben) to the beginning value. If it would subtract the credit, then the ending balances would match between the 2 reports.

-3342603.64 - 283.05 - 1308.39 - 111.8 = -3,353,615.5


The problem seems to be in the following sql in the `Account Sheet` report:
```sql
    --
    -- Update the current balance for each row.
    -- This implementation uses a rolling sum over the previous rows
    WITH beginningBalance_fa AS
             (
                 SELECT tmp_fa.fact_acct_id,
                        (
                                tmp_fa.endingBalance
                                + sum(acctbalance(tmp_fa.account_id, tmp_fa.amtacctdr, tmp_fa.amtacctcr)) -- !!!! acctbalance should return negative here
                                  OVER
                                      (
                                      PARTITION BY tmp_fa.account_id
                                      ORDER BY tmp_fa.dateacct, tmp_fa.fact_acct_id ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW
                                      )
                            ) AS                                                           beginningBalance,
                        acctbalance(tmp_fa.account_id, tmp_fa.amtacctdr, tmp_fa.amtacctcr) transactionBalance
[...]

PR #6367, opened 2020-03-19
Hide DocType with name  `** Neu **`

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/6121


TheBestPessimist:
I forgot to add migration script, sorry 'bout that.

Added the translations.

Issue #6386, opened 2020-03-23
Optimization of Procurement window


dragospodariu96:
### Is this a bug or feature request?
FR
### What is the current behavior?
The 'Create Purchase Order' action only takes the selected procurement material documents.
#### Which are the steps to reproduce?
Go to the 'procurement material' window and use the 'Create purchase order' action
### What is the expected or desired behavior?
When clicking on the 'Create Purchase Order' action, all entries should with a qty entered shou8ld be taken into consideration and purchase orders created from them

PR #6397, opened 2020-03-25
#6386 modified create purchase orders from procurement window process.

dragospodariu96:
https://github.com/metasfresh/metasfresh/issues/6386


de.metas.procurement.base/src/main/java/de/metas/procurement/base/order/async/PMM_GenerateOrders.java
```diff
@@ -66,7 +64,20 @@ public Result processWorkPackage(final I_C_Queue_WorkPackage workPackage, final
...
-		return Services.get(IQueueDAO.class).retrieveItemsSkipMissing(workpackage, I_PMM_PurchaseCandidate.class, ITrx.TRXNAME_ThreadInherited);
+		return Services.get(IQueueDAO.class).retrieveAllItems(workpackage, I_PMM_PurchaseCandidate.class);
+	}
+
+	private List<I_PMM_PurchaseCandidate> getItemsWithQty(List<I_PMM_PurchaseCandidate> candidates)
+	{
+		List<I_PMM_PurchaseCandidate> filteredList = new ArrayList<I_PMM_PurchaseCandidate>();
+		for (I_PMM_PurchaseCandidate candidate : candidates)
+		{
+			if (candidate.getQtyOrdered().compareTo(BigDecimal.ZERO) > 0)
```
metas-ts: * use `BigDecimal.signum()` instead

de.metas.procurement.base/src/main/java/de/metas/procurement/base/order/async/PMM_GenerateOrders.java
```diff
@@ -66,7 +64,20 @@ public Result processWorkPackage(final I_C_Queue_WorkPackage workPackage, final
...
+			{
+				filteredList.add(candidate);
+			}
```
metas-ts: * explicitly log the IDs of the `PMM_PurchaseCandidate` which you skip here
  * use Loggables.withLogger(logger, Level.DEBUG) for that, to make sure it's in the async-workpackage-log as well as in the actually logback-log

de.metas.procurement.base/src/main/java/de/metas/procurement/base/order/async/PMM_GenerateOrders.java
```diff
@@ -66,7 +64,20 @@ public Result processWorkPackage(final I_C_Queue_WorkPackage workPackage, final
 	private List<I_PMM_PurchaseCandidate> retrieveItems()
 	{
 		final I_C_Queue_WorkPackage workpackage = getC_Queue_WorkPackage();
-		return Services.get(IQueueDAO.class).retrieveItemsSkipMissing(workpackage, I_PMM_PurchaseCandidate.class, ITrx.TRXNAME_ThreadInherited);
+		return Services.get(IQueueDAO.class).retrieveAllItems(workpackage, I_PMM_PurchaseCandidate.class);
```
metas-ts: Already when enqeuting this, make sure to only add `PMM_PurchaseCandidate`s that have a quantity.
it's still possible theat they are zero-ed menwhile, but for that, there will be logging, see below

de.metas.procurement.base/src/main/java/de/metas/procurement/base/order/async/PMM_GenerateOrders.java
```diff
@@ -68,19 +68,6 @@ public Result processWorkPackage(final I_C_Queue_WorkPackage workPackage, final
 		return Services.get(IQueueDAO.class).retrieveItemsSkipMissing(workpackage, I_PMM_PurchaseCandidate.class, ITrx.TRXNAME_ThreadInherited);
 	}
 
-	private List<I_PMM_PurchaseCandidate> getItemsWithQty(List<I_PMM_PurchaseCandidate> candidates)
```
metas-ts: alright, good you removed it again; `OrderLineAggregation` already does that job (i saw it only now 😊 ).

de.metas.procurement.base/src/main/java/de/metas/procurement/base/order/async/PMM_GenerateOrders.java
```diff
@@ -1,16 +1,17 @@
...
-import org.adempiere.ad.trx.api.ITrx;
-import org.adempiere.exceptions.AdempiereException;
-
 import de.metas.async.api.IQueueDAO;
 import de.metas.async.model.I_C_Queue_WorkPackage;
 import de.metas.async.spi.WorkpackageProcessorAdapter;
 import de.metas.procurement.base.model.I_PMM_PurchaseCandidate;
 import de.metas.procurement.base.order.impl.OrdersGenerator;
 import de.metas.util.Services;
+import org.adempiere.ad.trx.api.ITrx;
```
metas-ts, line 9: please quickly clean up the imports; i think at least BigDecimal is not needed anymore

metas-mk:
> one tiny request regarding imports..
> 
> PS: one thing i had forgotten: i think we need to put special care to the process name: almost every process besides this one works only on the selected items.
> So maybe we need to name it something like "Gefilterte zu Bestellungen" / "Filtered to pruchase orders"
> Wdyt?

Does this process only work on filtered records with order qty > 0 or for all records with order qty > 0 (no matter what is filtered).

I would prefer the process name "Bestellungen generieren für Mengen > 0" or "Create Purchase for Quantities > 0"

dragospodariu96:
@metas-mk works just like in the swing client, taking all records currentyl in the view that have qtytoorder and qtyordered > 0. It takes into consideration all filters (like date, bpartner, organisation, etc) but regardless of how many records are manually highlighted, it filters all that are in the view at the moment the action is pressed.

Issue #6421, opened 2020-04-01
Flatrate term for commission can sometimes not be created bc mandatory field disappears


metas-dh:
### Is this a bug or feature request?
Bug afais
### What is the current behavior?
When creating a flatrate term (Vertragsbedingung) for commission, a mandatory field disappears once u select "Commission" as the contract type; but the document can not be completed w/o this (so the user has to set this first).
#### Which are the steps to reproduce?
See vid in me03 issue.

Issue #6299, opened 2020-03-03
Commission invoice candidates created too early, directly after sales order


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
The commission ICs are created directly after the sales order is completed, not after the delivery.
#### Which are the steps to reproduce?
1. Set up commission, make sure the flatrate term has invoice: after delivery set
2. Create a sales order, with a sales rep set
3. Complete the order: IC for the sales rep is already created: **NOK**
### What is the expected or desired behavior?
The ICs should only be created after the delivery for the sales order.

(found in e2e, see links for example in me03 issue).

PR #6424, opened 2020-04-01
solve commission related problems

metas-ts:
* CommissionShareHandler: solve metasfresh/metasfresh#6299 for non-item products
  * note that by default the commission product is in fact such a non-item (i.e. service) product
*  5556000_sys_gh6421_fix_c_flatrate_condition_mandatory_problem.sql: solve metasfresh/metasfresh#6421 by updating the AD_Column's mandatory logic
* OrderCheckupBuilder: minor unrelated


metas-ts:
planning to directly CP to `release` for IT after successful build, because it's part of the e2e anyways

Issue #6394, opened 2020-03-25
(Sales) Pricelist comparison new-last before new from menu for all Partners


TheBestPessimist:
### Is this a bug or feature request?
FR

PR #6431, opened 2020-04-02
Fine tuning after review

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/6394


de.metas.fresh/de.metas.fresh.base/src/main/jasperreports/de/metas/reports/pricelist/report_bpblock.jrxml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Created with Jaspersoft Studio version 6.5.1.final using JasperReports Library version 6.5.1  -->
```
cristinamghita, line 2: Is still  need this report?
Can't you use the general one?

dragospodariu96: Using the general one which has a much higher height, a blank space appears on new pages for new bpartners. As the general one is only used in the page title where there is enough blank space until the report starts, that is not a problem there, however in this case it is.

PR #6445, opened 2020-04-06
Fix using system AD_User_ID for contact in some places instead of BPartnerContactId

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/6302

### DIFF
--- a/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleBL.java
+++ b/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleBL.java
@@ -29,6 +29,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Properties;
 
+import de.metas.bpartner.BPartnerContactId;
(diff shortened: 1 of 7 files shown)


de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleBL.java
```diff
@@ -205,16 +209,14 @@ public I_C_BPartner getC_BPartner_Effective(final I_M_ReceiptSchedule rs)
...
 	@Override
-	public int getAD_User_Effective_ID(final I_M_ReceiptSchedule rs)
+	public BPartnerContactId getBPartnerContactID(@NonNull final I_M_ReceiptSchedule rs)
 	{
-		final int adUserOverrideID = rs.getAD_User_Override_ID();
-		if (adUserOverrideID > 0)
-		{
-			return adUserOverrideID;
-		}
+		final int cBPartnerIdRepo = rs.getC_BPartner_Override_ID() > 0 ? rs.getC_BPartner_Override_ID() : rs.getC_BPartner_ID();
```
metas-ts, line 216: that's OK, but next time, please use `CoalesceUtil.firstGreaterThanZero(int...)`

PR #6455, opened 2020-04-07
Gh6432

metas-rc:
Accounting Transactions view and open linked transactions process


de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/ProcessExecutionResult.java
```diff
@@ -827,58 +857,15 @@ public String toString()
...
-
-		public String getWindowIdString()
-		{
-			return adWindowIdStr;
-		}
-
-		public OpenTarget getTarget()
-		{
-			return target;
+			return getRecords().get(0);
```
metas-ts, line 868: given the method name, i think we should use 
`de.metas.util.collections.CollectionUtils.singleElement(Collection<T>)`
or, we should call the method getFirstRecord, if there can be >1

metas-ts: as talked: there is always at least one record

PR #6469, opened 2020-04-10
Gh6459 app

metas-ts:
https://github.com/metasfresh/metasfresh/issues/6459


metas-ts:
code looks (still) OK
next: deploy to dev-isntance and run e2e

Issue #6476, opened 2020-04-13
User Query filters errors


TheBestPessimist:
### Is this a bug or feature request?
BUG

PR #6478, opened 2020-04-13
User Query filters errors

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/6476

### DIFF
--- a/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/ColumnDisplayTypeProvider.java
+++ b/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/ColumnDisplayTypeProvider.java
@@ -1,0 +1,28 @@
+/*
+ * #%L
+ * de.metas.adempiere.adempiere.base
+ * %%
+ * Copyright (C) 2020 metas GmbH
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/ColumnDisplayTypeProvider.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+package org.compiere.model;
+
+public interface ColumnDisplayTypeProvider
```
metas-ts, line 25: i suggest you flag this as `@FunctionalInterface`, so we can inject lambdas

de.metas.adempiere.adempiere/base/src/main/java/org/compiere/apps/search/UserQueryFieldHelper.java
```diff
@@ -0,0 +1,103 @@
...
+			{
+				return DisplayType.toBooleanString((Boolean)valueObj);
+			}
+		}
+		catch (final Exception ex)
+		{
+			String error = ex.getLocalizedMessage();
+			if (error == null || error.isEmpty())
+			{
+				error = ex.toString();
```
metas-ts, line 96: i think here you lost the part from the legacy code where the error is logger. pls add some logging; (Level.WARN)

TheBestPessimist: Added error logs

de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/ColumnDisplayTypeProvider.java
TheBestPessimist: done.

PR #6479, opened 2020-04-13
timebookings initial commit

pvpurcarcosmin:
https://github.com/metasfresh/metasfresh/issues/5816


de.metas.serviceprovider/src/main/java/de/metas/serviceprovider/everhour/EverhourImporterService.java
```diff
@@ -0,0 +1,284 @@
...
+					.stream()
+					.map(interval -> buildGetTeamTimeRecordsRequest(request.getAuthToken(), interval))
+					.map(everhourClient::getTeamTimeRecords)
+					.flatMap(List::stream)
+					.filter(timeRecord -> isGithubID(timeRecord.getTask().getId()))
+					.forEach(this::importTimeBooking);
+		}
+		catch (final Exception e)
+		{
+			Loggables.withLogger(log, Level.ERROR).addLog(IMPORT_TIME_BOOKINGS_LOG_MESSAGE_PREFIX + e.getMessage());
```
metas-ts, line 113: i think in such a case it's good log log, but AFAIU (didn't look deep) you should rethrow while adding currently available info
e.g. 
```java
throw AdempiereException.wrapIfNeeded(e).appendParametersToMessage().setParameter("request", request);
```

de.metas.serviceprovider/src/main/java/de/metas/serviceprovider/issue/S_Issue.java
```diff
@@ -0,0 +1,50 @@
...
+
+	public S_Issue(final ExternalReferenceRepository externalReferenceRepository)
+	{
+		this.externalReferenceRepository = externalReferenceRepository;
+	}
+
+	@ModelChange(timings = ModelValidator.TYPE_AFTER_DELETE)
+	public void afterDelete(@NonNull final I_S_Issue record)
+	{
+		externalReferenceRepository.deleteByRecordIdAndType(record.getS_Issue_ID(), ExternalReferenceType.ISSUE_ID);
```
metas-ts, line 48: not 100% sure..did you check if `record.getS_Issue_ID()` is still set to be `>0` at this point? might be just out of habit, but i think i would do this on `TYPE_BEFORE_DELETE`. Anyways if the delete fails elsewehere, the transaction will be rolled back, so the external reference won't be gone

de.metas.serviceprovider/src/main/java/de/metas/serviceprovider/milestone/S_Milestone.java
```diff
@@ -0,0 +1,49 @@
...
+	private final ExternalReferenceRepository externalReferenceRepository;
+
+	public S_Milestone(final ExternalReferenceRepository externalReferenceRepository)
+	{
+		this.externalReferenceRepository = externalReferenceRepository;
+	}
+
+	@ModelChange(timings = ModelValidator.TYPE_AFTER_DELETE)
+	public void afterDelete(@NonNull final I_S_Milestone record){
+		externalReferenceRepository.deleteByRecordIdAndType(record.getS_Milestone_ID(), ExternalReferenceType.MILESTONE_ID);
```
metas-ts, line 47: same remark wrt after-delete vs before-delete

de.metas.serviceprovider/src/main/java/de/metas/serviceprovider/timebooking/S_TimeBooking.java
```diff
@@ -0,0 +1,99 @@
...
+		if (Check.isBlank(hoursAndMinutes))
+		{
+			return;
+		}
+
+		final Matcher matcher = hmmPattern.matcher(hoursAndMinutes);
+
+		if (!matcher.matches())
+		{
+			throw new AdempiereException(" Incorrect format! Please enter a value in H:mm format."
```
metas-ts, line 92: Please create an `AD_Message` and request a german trl. then you can call `de.metas.i18n.IMsgBL.getTranslatableMsgText(AdMessageKey, Object...)` to get an ITranslatableString and add that as constructor parameter to the AdempiereException.
Also invoke `markAsUserValidationError()` on the adempeireException before throwing it

de.metas.serviceprovider/src/main/java/de/metas/serviceprovider/user/AD_User.java
```diff
@@ -0,0 +1,50 @@
...
+
+	public AD_User(final ExternalReferenceRepository externalReferenceRepository)
+	{
+		this.externalReferenceRepository = externalReferenceRepository;
+	}
+
+	@ModelChange(timings = ModelValidator.TYPE_AFTER_DELETE)
+	public void afterDelete(@NonNull final I_AD_User record)
+	{
+		externalReferenceRepository.deleteByRecordIdAndType(record.getAD_User_ID(), ExternalReferenceType.USER_ID);
```
metas-ts, line 48: again, same remark wrt before-delete vs after-delete

de.metas.serviceprovider/src/main/java/de/metas/serviceprovider/issue/S_Issue.java
pvpurcarcosmin: yep, the ID is still there, but now there I think about it, it's better to have the removal before delete as we could add some triggers on the referenced tables in order to prevent inconsistent
delete operations done directly in DB.

Issue #6485, opened 2020-04-15
PLV Name adjustment


TheBestPessimist:
### Is this a bug or feature request?
FR

PR #6502, opened 2020-04-21
PLV name adjustment

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/6485


de.metas.business/src/main/java/de/metas/pricing/interceptor/M_PriceList_Version.java
```diff
@@ -0,0 +1,74 @@
...
+@Component
+public class M_PriceList_Version
+{
+	@Init
+	public void registerCallouts()
+	{
+		Services.get(IProgramaticCalloutProvider.class).registerAnnotatedCallout(this);
+	}
+
+	@CalloutMethod(columnNames = I_M_PriceList_Version.COLUMNNAME_ValidFrom)
```
metas-ts, line 55: Actually invoking it as callout method is AFAIU not needed. We would only need that if we had to updated it (on ValidFrom-Change) which the PLV was noch yet saved.
Although i guess it also doesn't hurt..

PR #6503, opened 2020-04-21
Gh6495 throw error on missing account

metas-rc:
https://github.com/metasfresh/metasfresh/issues/6495


de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc.java
```diff
@@ -1328,15 +1330,18 @@ else if (acctType == ACCTTYPE_PPVOffset)
...
 				}
 
 				return accountId;
 			}
 			else
 			{
-				log.error("No record found for: account Type=" + acctType + ", Record=" + get_ID() + ", SQL=" + sql + ", sqlParams=" + sqlParams);
-				return null;
+				throw new AdempiereException(MSG_NoAccountFound, get_ID(), acctType)
+						.setParameter("sql", sql)
```
metas-ts, line 1341:
```suggestion
						.markAsUserValidationError()
						.setParameter("sql", sql)
```

PR #6541, opened 2020-04-28
#6540 reverse credit memos that are not based on an invoice

metas-rc:
https://github.com/metasfresh/metasfresh/issues/6540


de.metas.swat/de.metas.swat.base/src/test/java/de/metas/invoicecandidate/api/impl/InvoiceCandBLHandlCompleteAndReversalTest.java
```diff
@@ -313,6 +322,79 @@ private void doTestCreditMemoReinvoicableAndReversal(final boolean reverseInvoic
 		assertThat(invoiceCandBL.sumupQtyInvoicedAndNetAmtInvoiced(ic).get().getLeft().getStockQty().toBigDecimal(), comparesEqualTo(expectedQtyInvoicedSumAfterReversal));
 	}
 
+	private void doTestCreditMemoReversal( final BigDecimal reversalIlaExpectedQtyInvoiced)
```
metas-ts, line 325: pls add a javadoc with a brief description of what the new test does. not required before integrating, but it would be good to add it soon

Issue #6543, opened 2020-04-28
Make currencyrate editable in Bankstatementline, GL Journal


dragospodariu96:
### Is this a bug or feature request?
FR
### What is the current behavior?
Currency rate is not editable
#### Which are the steps to reproduce?
Go to a gl_Journal line or bankstatement line
### What is the expected or desired behavior?
That the currencyrate will be editable and the fact lines created will also have the line currencyrate

PR #6544, opened 2020-04-28
#6543 Currency Rate now editable in Bank statement and gl journal lines. Account Transactions created from them will also use the currency rate set in the line.

dragospodariu96:
#6543


de.metas.acct.base/src/main/java-legacy/org/compiere/acct/FactLine.java
```diff
@@ -1070,9 +1085,20 @@ public void convert()
 
 			setAmtAcctDr(amtAcctDr);
 			setAmtAcctCr(amtAcctCr);
-			setCurrencyRate(currencyRate.getConversionRate());
+
+			final String docType = getDoc().getDocumentType();
+
+			if (docType.equals(DOCTYPE_GL_Journal) || docType.equals(DOCTYPE_Bank_Statement))
+			{
+				final BigDecimal lineCurrencyRate = getDocLine().getValueAsBD("currencyrate", null);
```
metas-ts, line 1093: pls consult @teosarca on this

de.metas.acct.base/src/main/java-legacy/org/compiere/acct/FactLine.java
```diff
@@ -1070,9 +1085,20 @@ public void convert()
...
+				setCurrencyRate(lineCurrencyRate != null ? lineCurrencyRate : currencyRate.getConversionRate());
```
metas-ts, line 1094: here you can use `CoalesceUtil.coalesce(....)` instead

Issue #6521, opened 2020-04-23
Vendor Return takes the wrong price


dragospodariu96:
### Is this a bug or feature request?
FR
### What is the current behavior?
In vendor return, the m_inout_line entries don't have the c_orderline_id set, where applicable.
#### Which are the steps to reproduce?
Go to a vendor return entry.
### What is the expected or desired behavior?
The c_orderline_id should be set.

PR #6553, opened 2020-04-30
#6521 Showing price from purchase order in billing candidate and generated purchase invoice.

dragospodariu96:
https://github.com/metasfresh/metasfresh/issues/6521


metas-ts:
going to create th3e PR against master

PR #6556, opened 2020-05-04
Gh6521 2

metas-ts:
#6521


backend/de.metas.business/src/main/java/de/metas/order/impl/OrderLineBL.java
```diff
@@ -773,6 +773,12 @@ public ProductPrice getCostPrice(final org.compiere.model.I_C_OrderLine orderLin
 				.build();
 	}
 
+	@Override
+	public BigDecimal getPriceActual(final org.compiere.model.I_C_OrderLine orderline)
```
metas-ts, line 777: as said earlier: returning a BigDecimal as a price without any context is dangerous. instead, return `ProductPrice`

backend/de.metas.device.scales/src/test/java/de/metas/device/scales/endpoint/TcpConnectionEndPointTest.java
```diff
@@ -129,8 +129,6 @@ public void run()
 								final String anotherWrongServerReturnString = MockedEndpoint.createWeightString(new BigDecimal(weight + 10)) + ISiscCmd.SICS_CMD_TERMINATOR;
 								out.write(anotherWrongServerReturnString.getBytes(ICmd.SICS_CMD_CHARSET));
 								System.out.println("TcpConnectionEndPointTest" + ": server-socked-thread replied with anotherWrongServerReturnString=" + anotherWrongServerReturnString);
-
-								out.flush();
```
metas-ts, line 133: unrelated change that got in here in the context of #6205 .. need to check this myself

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
```diff
@@ -823,7 +822,7 @@ public static PriceAndTax calculatePriceAndTax(
 				.taxCategoryId(pricingResult.getTaxCategoryId())
 				//
 				.priceEntered(pricingResult.getPriceStd())
-				.priceActual(pricingResult.getPriceStd())
+				.priceActual(inoutLine.getC_OrderLine_ID() != 0 ? Services.get(IOrderLineBL.class).getPriceActual(inoutLine.getC_OrderLine()) : pricingResult.getPriceStd())
```
metas-ts, line 825: we can'T combine a number from a C_OrderLine with an UOM from the prcing engine. there is no guarantee that they fit together

dragospodariu96: removed old logic, using the existing getcostprice() method from orderlineBL

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderLineBL.java
dragospodariu96: removed that completely, using getcostprice() which returns a Productprice with price actual and uomID

Issue #6548, opened 2020-04-29
Lookup Process parameters shall allow inactive records


cristinamghita:
### Is this a bug or feature request?
FR

PR #6560, opened 2020-05-04
Gh6548

metas-ts:
#6548


metas-ts:
(waiting for build to succeed)

PR #6562, opened 2020-05-04
#6520 processes to recompute selected candidates

metas-rc:
https://github.com/metasfresh/metasfresh/issues/6520


backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/inoutcandidate/process/M_ShipmentSchedule_Recompute.java
```diff
@@ -0,0 +1,74 @@
...
+
+	@Override
+	protected void postProcess(final boolean success)
+	{
+		if (!success)
+		{
+			return;
+		}
+		
+		invalidateView();
```
metas-ts, line 72: do we need the invalidateView(); invocation?
AFAIU the invalidation service's implementation should fire a cache-invalidation event, so this extra invalidate shouldn't be required..

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/invoicecandidate/process/C_Invoice_Candidate_Recompute.java
```diff
@@ -0,0 +1,52 @@
...
+
+	@Override
+	protected void postProcess(final boolean success)
+	{
+		if (!success)
+		{
+			return;
+		}
+
+		invalidateView();
```
metas-ts, line 50: same here....should AFAIU work without this

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/invalidation/impl/ShipmentScheduleInvalidateRepository.java
```diff
@@ -609,5 +611,25 @@ public void releaseRecomputeMarkerOutOfTrx(final PInstanceId pinstanceId)
...
+	public final void invalidateShipmentSchedulesFor(@NonNull final IQuery<I_M_ShipmentSchedule> shipmentScheduleQuery)
+	{
+		final int count = shipmentScheduleQuery.insertDirectlyInto(I_M_ShipmentSchedule_Recompute.class)
+				.mapColumn(I_M_ShipmentSchedule_Recompute.COLUMNNAME_M_ShipmentSchedule_ID, 
+						I_M_ShipmentSchedule.COLUMNNAME_M_ShipmentSchedule_ID)
+				
+				.execute()
+				.getRowsInserted();
+
+		logger.debug("Invalidated {} shipment schedules for {}", new Object[] { count, shipmentScheduleQuery });
```
metas-ts, line 625: for some reason i started the review at the bottom..getting to this point from there, i think you need to fire a cache invalidation event here. that way 
* every client will be notified and can update their view (not just the user who ran the process)
* we (probably) don't need to create FUD by having the invalidateView call in the process

metas-rc: done directly in de.metas.inoutcandidate.invalidation.impl.ShipmentScheduleInvalidateRepository.invalidateShipmentSchedulesFor(IQuery<I_M_ShipmentSchedule>)

PR #6593, opened 2020-05-06
Gh6566 2 customer retention: repair logic, refactoring, test

metas-rc:


backend/de.metas.contracts/src/main/java/de/metas/contracts/impl/CustomerRetentionRepository.java
```diff
@@ -64,13 +68,18 @@
 	private final IInvoiceDAO invoiceDAO = Services.get(IInvoiceDAO.class);
 	private final IContractsDAO contractsDAO = Services.get(IContractsDAO.class);
 	private final ContractInvoiceService contractInvoiceService;
-	
+
+	public I_C_Customer_Retention getById(final CustomerRetentionId customerRetentionId)
```
metas-ts, line 72: pls add `@NonNull` at least to public methods like this one

backend/de.metas.contracts/src/main/java/de/metas/contracts/impl/CustomerRetentionRepository.java
```diff
@@ -80,134 +89,111 @@ public I_C_Customer_Retention createNewCustomerRetention(@NonNull final BPartner
...
+	@VisibleForTesting
+	void setNewCustomer(@NonNull final CustomerRetentionId customerRetentionId)
 	{
-		I_C_Customer_Retention customerRetention = retrieveCustomerRetention(bpartnerId);
-
-		if (customerRetention == null)
-		{
-			customerRetention = createNewCustomerRetention(bpartnerId);
-		}
+		final I_C_Customer_Retention customerRetention = getById(customerRetentionId);
```
metas-ts, line 99: please add a comment or better an Check.assumeNotNull here to make it explicit the you made sure elsewhere this won't be null

PS: if you need to rely on something outside this class, such as the model interceptors, i suggest to invoke that getCreate method instead..

backend/de.metas.contracts/src/main/java/de/metas/contracts/impl/CustomerRetentionRepository.java
metas-rc: Done.

backend/de.metas.contracts/src/main/java/de/metas/contracts/impl/CustomerRetentionRepository.java
metas-rc: Check.assumeNotNull done.

Issue #6610, opened 2020-05-08
Material Withdrawal cannot run because wrong locator is detected


TheBestPessimist:
### Is this a bug or feature request?
BUG

PR #6612, opened 2020-05-08
Material Withdrawal cannot run because wrong locator is detected

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/6610


backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/handlingunits/process/HUMoveToDirectWarehouseService.java
```diff
@@ -136,7 +131,7 @@ private final void generateMovement(@NonNull final I_M_HU hu)
 			// Move the HU
 			final I_M_Movement movement = new HUMovementBuilder()
 					.setContextInitial(PlainContextAware.newWithThreadInheritedTrx())
-					.setWarehouseFrom(IHandlingUnitsBL.extractWarehouse(hu))
+					.setLocatorFrom(IHandlingUnitsBL.extractLocator(hu))
```
metas-ts, line 134: 👍

PR #6621, opened 2020-05-10
Websockets Refactoring

petrican:


petrican:
This became complex to refactor. Even if this achieves the initial purpose (having all subscription in one connection) it becomes a burden to integrate it and make sure every functionality works as it used to. Due to the lack of time to spend on this I decided that is best to keep existing functionality which does work properly and come back to it after the big refactoring which is more important.

PR #6631, opened 2020-05-11
Blue moon evt tracing

metas-ts:
* Delete ActiveMQ event endpoint
* Improve logging around events
* Add distributed event tracing


metas-ts:
image

metas-ts:
One APM transaction sample (btw, it's a `C_Order - complete`):
image

--------------------

Scrolling down, we find a span where the web-api fired a cache-invalidation event:
image

-------------------

the event was instrumented such that we can click on it...:

PR #6637, opened 2020-05-12
Gh6566 4 refactor and improve

metas-rc:


metas-rc:
Note: wrong master end date did not prove to be a problem for this issue, in the end. If it ever becomes a problem we will create a followup.

PR #6639, opened 2020-05-12
#6566 refactoring and improvements

metas-rc:


backend/de.metas.contracts/src/main/java/de/metas/contracts/bpartner/ContractPartnerRepository.java
```diff
@@ -0,0 +1,34 @@
...
+
+@Repository
+public class ContractPartnerRepository
+{
+	public ImmutableSet<BPartnerId> retrieveAllPartnersWithContracts()
+	{
+		IQueryBL queryBL = Services.get(IQueryBL.class);
+
+		final IQuery<I_C_Flatrate_Term> contractsFilter = queryBL.createQueryBuilder(I_C_Flatrate_Term.class)
+				.addOnlyActiveRecordsFilter()
```
metas-ts, line 22: now i'm wondering, what about e.g. in-progress terms. i think you need to filter only for completed and closed ones

backend/de.metas.contracts/src/main/java/de/metas/contracts/bpartner/ContractPartnerRepository.java
```diff
@@ -0,0 +1,34 @@
...
+import com.google.common.collect.ImmutableSet;
+
+import de.metas.bpartner.BPartnerId;
+import de.metas.contracts.model.I_C_Flatrate_Term;
+import de.metas.util.Services;
+
+@Repository
+public class ContractPartnerRepository
+{
+	public ImmutableSet<BPartnerId> retrieveAllPartnersWithContracts()
```
metas-ts, line 17:
```suggestion
	public ImmutableSet<BPartnerId> retrieveIdsOfPartnerWithContracts()
```

* include the "Ids" in the name
* in other mf places, "All" implies also inactive, which is not the case here

metas-ts: There might be 500k bpartners; please return an iterator
(not sure if i approved the set in a previous PR, but i think it's wrong and we have `IQuery.iterateIds(...)`)

metas-rc:
> Pls check my comments.
> We were talking about fixing incorrect masterenddates: shouldn't there be a script for that?

~~As per my comment that I forgot to write~~ Those cases are not affecting the reports after my other changes. I talked with Mark about them and he suggested to only fix them in a followup if they become problematic in the future and if this is visible to the customer so we can budget the fix in another task.
After my last remote test, there are only 13 lines in the report without customer retention for customers that are really no (longer)  customers.

image

backend/de.metas.contracts/src/main/java/de/metas/contracts/bpartner/ContractPartnerRepository.java
metas-rc: Right. Done.

backend/de.metas.contracts/src/main/java/de/metas/contracts/bpartner/ContractPartnerRepository.java
metas-rc: Agree. Making it an iterator.

PR #6644, opened 2020-05-13
replace the executor for the blocking threadpool with a simpler version

metas-ts:
..that is also very probably more stable. thx to https://stackoverflow.com/a/50361304/1012103
Also improve logging


metas-ts:
(update: i'm still not where i want to be in terms of this PR..still working on it now and then)

metas-ts:
this PR is currently not ready. options:
* instead, port PR https://github.com/metasfresh/metasfresh/pull/6720 to `blue_moon`
* bare minimum: manually sync from master (PR https://github.com/metasfresh/metasfresh/pull/6720) 
  * `backend/de.metas.util/src/main/java/org/adempiere/util/concurrent/BlockingExecutorWrapper.java` 
  * `backend/de.metas.async/src/main/java/de/metas/async/processor/impl/ThreadPoolQueueProcessor.java`

Issue #6687, opened 2020-05-19
Optimize Screen behavior for "Rüstliste erfassen" in shipment schedule


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
Issue occurs when opening the modal in windows/500221 (Shipment Disposition)
#### Which are the steps to reproduce?
Trying to edit the attributes on the last field row open the popup outside the visible area
### What is the expected or desired behavior?
Popup should be visible

PR #6688, opened 2020-05-19
Add dropup for the last items such that they are visible #6687

petrican:


frontend/src/components/widget/Attributes/AttributesDropdown.js
```diff
@@ -125,8 +130,17 @@ class AttributesDropdown extends Component {
...
+    const { isVisible, dropdownCoords } = this.state;
+
     return (
-      <div className="attributes-dropdown panel-shadowed panel-primary panel-bordered panel-spaced">
+      <div
+        ref={this.selector}
+        className={classnames(
+          'attributes-dropdown panel-shadowed panel-primary panel-bordered panel-spaced',
+          { 'hidden ': !isVisible },
+          { 'attributes-dropup': dropdownCoords && dropdownCoords.y > 680 }
```
teosarca, line 141: pls extract 680 as `const` at top of this file.
In this way it would be easier to track/tune hardcoded constants.

petrican:
> pls check my comments.
> 
> Btw, I am wondering if we couldn't find some solution that is not so fragile and directly linked to the Y coordinate of where we clicked,
> because in some "corner" cases would fail.
> e.g.
> image

Need to think. Takes more time to find a solution that would work in all case scenarios.

teosarca:
@petrican  then pls leave it as it is. Just extract the constant to not have that 680 hardcoded deep in the code.

PR #6711, opened 2020-05-22
Disable `shouldComponentUpdate` optimization

siemiatj:
It should be less aggressive as it breaks modals. 

Related to #6666

### DIFF
--- a/frontend/src/containers/DocList.js
+++ b/frontend/src/containers/DocList.js
@@ -2,7 +2,6 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import classnames from 'classnames';
-import _ from 'lodash';
 
 import { updateUri } from '../actions/AppActions';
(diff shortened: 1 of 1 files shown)


frontend/src/containers/DocList.js
```diff
@@ -23,13 +23,6 @@ const EMPTY_OBJECT = {};
  * @extends Component
  */
 class DocList extends Component {
-  shouldComponentUpdate(nextProps) {
```
petrican, line 26: If we remove the checks from the shouldComponentUpdate wouldn't this impact the re-renders ? Could be this tuned up to allow the re-renderings in sync with the quickActions store updates ?

petrican, line 26: If we remove the checks from the shouldComponentUpdate wouldn't this impact the re-renders ? Could be this tuned up to allow the re-renderings in sync with the quickActions store updates ?

siemiatj: I will tune it but it's a hotfix to unblock the customer.

PR #6715, opened 2020-05-22
Frontend 6712 fix dates

petrican:
Moment should use the locale set


petrican:
If there is a simpler way of getting the locale and fix I am open to suggestions. The fieldValueToString is ....a pain. That should reside in a util file and be separated from the TableCell logic ..

PR #6716, opened 2020-05-22
Proper fix over the temporary hotfix for DocList/quickactions

siemiatj:
Turns out using `PureComponent` works even better than trying to manually compare props before the update. Especially when opening modals, as it decreases the number of refreshes 3x

Related to #6666


siemiatj:
They also tend to vary between each consecutive run.

PR #6720, opened 2020-05-23
Master async processor

metas-ts:
* replace the executor for our blocking threadpool with a version that is simpler to understand, more recent also very probably more stable. thx to stackoverflow.com/a/50361304/1012103
* Also improve logging
* also fix a  big performance problem with respect to saving event logging


metas-ts:
# QA 
* note that i didn't really change any processor business logic; only how things are done under the hood

# Tests done:
* create sales order, shipment & invoice
* mass-created 1k C_OlCands and turned their orders' shipment-schedules into shipments
  * (note that this failed in the beginning)

PR #6721, opened 2020-05-25
Cypress no longer needs an env variable

TheBestPessimist:
Make cypress take all its config from `/cypress/config.js`. In this way only the config can be modified whenever cypress has to be run

### DIFF
--- a/e2e/.gitignore
+++ b/e2e/.gitignore
@@ -11,3 +11,6 @@ cypress/config.js
 /cypress/config.js.remote
 /cypress/config.js.local
 /metasfresh-e2e.iml
+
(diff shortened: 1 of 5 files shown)


TheBestPessimist:
Question: what is the use of `PLUGIN_API_URL: 'bla'`? I see it's not used in any .js file. Can we remove it? 

If there are no arguments for its use, i will remove it.

TheBestPessimist:
I am not sure if Jenkins is affected by the change. Does jenkins use the `cypress/config.js` file or env variables?

metas-ts:
i don'T know by heart. you'll have to try it out

PR #6730, opened 2020-05-26
fix NPE in CustomerTradeMarginPricingRule

metas-ts:


backend/de.metas.contracts/src/main/java/de/metas/contracts/pricing/trade_margin/CustomerTradeMarginPricingRule.java
```diff
@@ -71,20 +72,25 @@
...
 	{
+		final ILoggable loggable = Loggables.withLogger(logger, Level.DEBUG);
 		if (!result.isCalculated())
 		{
-			Loggables.withLogger(logger, Level.DEBUG).addLog("Not applying due to missing calculated price!");
+			loggable.addLog("applies - pricingResult.isCalculated=false -> return false");
 			return false;
 		}
 		final BPartnerId customerId = pricingCtx.getBPartnerId();
+		if (customerId == null)
```
metas-ts, line 84: this is the actual NPE-fix

PR #6734, opened 2020-05-27
when converting quantities, round up

metas-ts:


metas-ts:
wrong target branch. has to be  `invisible_croissant_hotfix`

PR #6737, opened 2020-05-27
when converting quantities, round up

metas-ts:


metas-ts:
was done, see https://github.com/metasfresh/metasfresh/pull/6744

Issue #6738, opened 2020-05-27
New Action and reduced window for Request


dragospodariu96:
### Is this a bug or feature request?
FR
### What is the current behavior?
No customized request window or action to create request from the 'kontakt mobil' window
#### Which are the steps to reproduce?
Go to the kontakt mobil window
### What is the expected or desired behavior?
Have a custom request window for bodymed & action to create request in the 'Kontakt Mobil' window.

PR #6739, opened 2020-05-27
#6738 Added the case when we want to create a request from the user window.

dragospodariu96:
#6738


metas-ts:
going to create a new PR to `blonde_monkey`

Issue #6745, opened 2020-05-27
Fix completing inventory with lines that have an HU already assigned flow.


pvpurcarcosmin:
### Is this a bug or feature request?
BUG

PR #6748, opened 2020-05-28
moved to callout

pvpurcarcosmin:
https://github.com/metasfresh/metasfresh/issues/6745

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/inventory/InventoryLineRecordService.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/inventory/InventoryLineRecordService.java
@@ -39,6 +39,7 @@ import de.metas.inventory.HUAggregationType;
 import de.metas.inventory.IInventoryBL;
 import de.metas.inventory.InventoryId;
 import de.metas.product.ProductId;
+import de.metas.quantity.QuantitiesUOMNotMatchingExpection;
 import de.metas.quantity.Quantity;
 import de.metas.uom.IUOMConversionBL;
 import de.metas.uom.IUOMDAO;
@@ -106,6 +107,9 @@ public class InventoryLineRecordService
 
 	public void syncToHUs(@NonNull final I_M_Inventory inventoryRecord)
 	{
+		//make sure inventoryLine.qtyBooked is up to date before syncing to HU
+		recomputeQtyBookedFromStorageForInventory(InventoryId.ofRepoId(inventoryRecord.getM_Inventory_ID()));
+
 		final Inventory inventory = inventoryLineRepository.toInventory(inventoryRecord);
 		syncToHUs(inventory);
 	}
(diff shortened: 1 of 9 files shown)


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/inventory/InventoryLineRecordService.java
```diff
@@ -405,6 +423,64 @@ public HUAggregationType computeHUAggregationType(
...
+		}
+		else
+		{
+			inventoryLine.setQtyBook(BigDecimal.ZERO);
+		}
+	}
+
+
+	@NonNull
+	private Optional<Quantity> getBookedQtyFromStorage(@NonNull final I_M_InventoryLine inventoryLineRecord)
```
metas-ts, line 451: AFAIS  it doesn't work like this; the method doesn't cover the case of multiple HUs for one line.
Rough sketch of how to do it instead:
* change the method to work with `de.metas.handlingunits.inventory.InventoryLine` instead of `I_M_InventoryLine`
  * maybe when it's not a dirty record we work with, the whole code can be moved out of this dirty service
* use `inventoryLine.getInventoryLineHUs()` to get the `n` HUs a line can have
* you probably need IHandlingUnitDAO to load the respective I_M_HUs; (maybe there is an easy way to load them all at once)
* use `IHandlingUnitsBL.getStorageFactory()` and IHUStorageFactory.getHUProductStorages() to get the quantities.
* maybe you don't need `uomConversionBL` to convert them to the inventory line's UOM..might be that's done by the repository (need to check)
* :exclamation: By the way: check out how the processes from today's book-club do it. maybe you can simply extract and/or reuse that code.
* fun fact: your change of `InventoryRepository.retrieveLineRecords` to `public` triggered all these thoughts..opening a nice API like that to server raw database records was kind of a smell :-D
* :exclamation: 
  * AFAIU InventoryLine is not really the aggregate root; Inventory is..
  * in the `M_InventoryLine` MI you need to update a single line though..not sure that to do
  * ..maybe let InventoryRepository load an inventory with just the line in question?
  * the good news is that AFAIK @teosarca will be around tomorrow for consultation
  * ...did i mention that there *might* be some code already doing all this?

pvpurcarcosmin: 🤦 yep, you're right, I didn't put much thought into it as it was more like a quick fix 🙃, thx for raising the issue I'll check that process and will try to find a way to reuse the code.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/inventory/InventoryRepository.java
```diff
@@ -84,6 +87,8 @@
 	private final IUOMConversionBL convBL = Services.get(IUOMConversionBL.class);
 	private final IWarehouseDAO warehousesRepo = Services.get(IWarehouseDAO.class);
 	private final IQueryBL queryBL = Services.get(IQueryBL.class);
+	private final IHUStorageDAO huStorageDAO = Services.get(IHandlingUnitsBL.class).getStorageFactory().getHUStorageDAO();
```
metas-ts, line 90: why use IHUStorageDAO?

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/inventory/InventoryRepository.java
```diff
@@ -465,6 +498,30 @@ public void saveInventoryLineHURecords(@NonNull final InventoryLine inventoryLin
 		}
 	}
 
+	public Optional<Quantity> getFreshBookedQtyFromStorage(@NonNull ProductId productId, @NonNull UomId inventoryLineUOMId, @NonNull HuId huId)
+	{
+		Optional<Quantity> bookedQty = Optional.empty();
+
+		I_M_HU hu = huDAO.getById(huId);
+
+		final I_M_HU_Storage huStorage = huStorageDAO.retrieveStorage(hu, productId);
```
metas-ts, line 507: i think at this point we would be be better off using *not* the DB records, but hustorage, as sketched out in my last review

PR #6756, opened 2020-05-29
#6738 Refactored CreateRequest process and added case for kontakt mobil.

dragospodariu96:
#6738

### DIFF
--- a/backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/request/process/WEBUI_CreateRequest.java
+++ b/backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/request/process/WEBUI_CreateRequest.java
@@ -1,9 +1,11 @@
 package de.metas.ui.web.request.process;
 
+import ch.qos.logback.classic.Level;
 import de.metas.bpartner.service.IBPartnerDAO;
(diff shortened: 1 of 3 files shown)


backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/request/process/WEBUI_CreateRequest.java
```diff
@@ -63,76 +63,76 @@ public WEBUI_CreateRequest()
...
+		{
+			final I_AD_User user = userDAO.getById(UserId.ofRepoId(getProcessInfo().getRecord_ID()));
+			request = createRequestFromUser(user);
 		}
 		else
 		{
 			throw new IllegalStateException("Not supported: " + tableName);
 		}
+		save(request);
+		getResult().setRecordToOpen(TableRecordReference.of(request), requestWindowId.get().getRepoId(), OpenTarget.SingleDocumentModal);
```
metas-ts, line 87: don't blindly assume that `requestWindowId` is present. 
Suggestion: implement `IProcessPrecondition`, then check that in `checkPreconditionsApplicable` if the window is not available, then log a message and and use `rocessPreconditionsResolution.rejectWithInternalReason(..)`

PR #6772, opened 2020-06-01
Rewrite Table component

siemiatj:
And a lot more

Related to #6706


petrican:
Build fails. Pls fix the linter error.

siemiatj:
> Build fails. Pls fix the linter error.

Yeah I'm aware. There's still some commented out code but it can already be used as a base for adding missing tests.

frontend/src/components/table/Table.js
```diff
@@ -57,500 +34,141 @@ class Table extends Component {
...
-            }
-          });
+  setContextMenu = (
+    clientX,
+    clientY,
+    fieldName,
+    supportZoomInto,
+    supportFieldEdit
+  ) => {
+    this.setState({
```
petrican, line 73: I dont understand why are we still using local component state for the table

petrican:
Can't all these be moved to the table state under the reducer ? https://github.com/metasfresh/metasfresh/blob/master/frontend/src/utils/tableHelpers.js#L23

frontend/src/components/app/DocumentList.js
```diff
@@ -236,20 +249,21 @@ export default class DocumentList extends Component {
...
                   resetInitialValues={onResetInitialFilters}
                 />
               )}
 
-              {rowData && staticFilters && (
+              {staticFilters && (
                 <FiltersStatic
-                  {...{ windowType, viewId }}
+                  {...{ viewId }}
+                  windowType={windowId}
```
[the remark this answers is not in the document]
siemiatj: It's not something I'm focusing on in this effort, but you're right.

siemiatj:
As discussed - when using `redux` it's still fine to use local state for local things that don't necessarily have to be shared anywhere. Thanks to that the reducers are smaller and we don't have a bunch of action creators to just toggle flags. Moreover this PR splits Table into presentation and data layer. And local state is only used for visual things. Also having them in the redux state won't make any impact on rendeing as changing values in the store will still force a re-render of the component. So I think it's still totally fine to do it like that and it's in line with best practices. 

https://redux.js.org/faq/organizing-state

petrican:
> https://redux.js.org/faq/organizing-state

I agree with that with one remark. There are situations like in our case when state is altered at component level and is causing un-necessary re-renders. Pls see documented issue => https://github.com/metasfresh/me03/issues/4555#issuecomment-638141807 . Since we know we could lead into a similar situation in the future do you think is advisable to keep also state at component level ?

siemiatj:
> > https://redux.js.org/faq/organizing-state
> 
> "Find a balance that works for you, and go with it"
> 
> I agree with that with one remark. There are situations like in our case when state is altered at component level and is causing un-necessary re-renders. Pls see documented issue => [metasfresh/me03#4555 (comment)](https://github.com/metasfresh/me03/issues/4555#issuecomment-638141807) . Since we know we could lead into a similar situation in the future do you think is advisable to keep also state at component level ? This is the reason why I wanted this to be discussed within the frontend meeting.

Not sure I follow. You are aware that if we move anything to redux and will toggle it there instead of the local state it will still cause a rerender, because the prop value will change ?

petrican:
> > > https://redux.js.org/faq/organizing-state
> > 
> > 
> > "Find a balance that works for you, and go with it"
> > I agree with that with one remark. There are situations like in our case when state is altered at component level and is causing un-necessary re-renders. Pls see documented issue => [metasfresh/me03#4555 (comment)](https://github.com/metasfresh/me03/issues/4555#issuecomment-638141807) . Since we know we could lead into a similar situation in the future do you think is advisable to keep also state at component level ? This is the reason why I wanted this to be discussed within the frontend meeting.
> 
> Not sure I follow. You are aware that if we move anything to redux and will toggle it there instead of the local state it will still cause a rerender, because the prop value will change ?

The issue with that 'select' being in the state is that it re-renders with every component state alter causing side effect. What we want is to achieve predictability and only re-render the child components when we really need that.

siemiatj:
> > > > https://redux.js.org/faq/organizing-state
> > > 
> > > 
> > > "Find a balance that works for you, and go with it"
> > > I agree with that with one remark. There are situations like in our case when state is altered at component level and is causing un-necessary re-renders. Pls see documented issue => [metasfresh/me03#4555 (comment)](https://github.com/metasfresh/me03/issues/4555#issuecomment-638141807) . Since we know we could lead into a similar situation in the future do you think is advisable to keep also state at component level ? This is the reason why I wanted this to be discussed within the frontend meeting.
> > 
> > 
> > Not sure I follow. You are aware that if we move anything to redux and will toggle it there instead of the local state it will still cause a rerender, because the prop value will change ?
> 
> The issue with that 'select' being in the state is that it re-renders with every component state alter causing side effect. What we want is to achieve predictability and only re-render the child components when we really need that.

But select has not been in the state since the very beginning of this effort. Moving select and  rows was one of the premises to do this whole refactor :)

petrican:
> > > > > https://redux.js.org/faq/organizing-state
> > > > 
> > > > 
> > > > "Find a balance that works for you, and go with it"
> > > > I agree with that with one remark. There are situations like in our case when state is altered at component level and is causing un-necessary re-renders. Pls see documented issue => [metasfresh/me03#4555 (comment)](https://github.com/metasfresh/me03/issues/4555#issuecomment-638141807) . Since we know we could lead into a similar situation in the future do you think is advisable to keep also state at component level ? This is the reason why I wanted this to be discussed within the frontend meeting.
> > > 
> > > 
> > > Not sure I follow. You are aware that if we move anything to redux and will toggle it there instead of the local state it will still cause a rerender, because the prop value will change ?
> > 
> > 
> > The issue with that 'select' being in the state is that it re-renders with every component state alter causing side effect. What we want is to achieve predictability and only re-render the child components when we really need that.
> 
> But select has not been in the state since the very beginning of this effort. Moving select and rows was one of the premises to do this whole refactor :)

Agree. I was just emphasising the origin of the unnecessary re-renderings

petrican:
Please fix the failing tests. I haven't had the chance to look over the code. Hope I'll be able to do this during the day.

frontend/src/actions/TableActions.js
```diff
@@ -38,26 +39,81 @@ export function deleteTable(id) {
...
+ */
+export function updateTableSelection(id, selection, keyProperty = 'id') {
+  return {
+    type: types.UPDATE_TABLE_SELECTION,
+    payload: { id, selection, keyProperty },
+  };
+}
+
+/**
+ * Update table selection - deselect items or deselect all if an empty `ids`
```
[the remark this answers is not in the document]
petrican: Yes. I should have document that. My bad.

frontend/src/containers/Table.js
```diff
@@ -0,0 +1,268 @@
...
+  handleSelect = (ids, cb) => {
+    const {
+      updateTableSelection,
+      windowId,
+      viewId,
+      docId,
+      tabId,
+      keyProperty,
+    } = this.props;
+    let newSelected = [];
```
petrican, line 70: can we move all this logic from L:70-78 to a separate function that can live outside the Table component  in tableHelpers ?  let newSelection = getSelectionForIds(ids). Goal is to have the Table as thin as possible

frontend/src/containers/Table.js
```diff
@@ -0,0 +1,268 @@
...
+      if (!docId) {
+        updateDocList();
+      }
+    });
+  };
+
+  handleZoomInto = (fieldName) => {
+    const { entity, windowId, docId, tabId, viewId, selected } = this.props;
+
+    getZoomIntoWindow(
```
petrican, line 171: I would move this also outside the component.

frontend/src/containers/Table.js
```diff
@@ -0,0 +1,268 @@
...
+    emptyHint: table.emptyHint,
+    indentSupported: table.indentSupported,
+    collapsible: table.collapsible,
+    keyProperty: table.keyProperty,
+    size: table.size,
+    allowShortcut: handleShortcuts,
+    allowOutsideClick: state.windowHandler.allowOutsideClick,
+    modalVisible,
+    isGerman:
+      state.appHandler.me.language && state.appHandler.me.language.key
```
petrican, line 250: This is not quite reliable to get the language. I found out that the most reliable way is from the localStore where it is set as a result of the call to getSession.  See ex: https://github.com/metasfresh/metasfresh/pull/6824/files#diff-858b40143f05e165f6bc793cd0a6fa32R303

frontend/src/reducers/viewHandler.js
```diff
@@ -163,7 +160,7 @@ export default function viewHandler(state = initialState, action) {
           firstRow,
           headerProperties,
           pageLength,
-          result,
+          // result,
```
petrican, line 163: Not sure I understand why this is being removed. Pls explain

frontend/src/reducers/viewHandler.js
```diff
@@ -399,6 +377,23 @@ export default function viewHandler(state = initialState, action) {
...
+      const { id, showIncludedView } = action.payload;
+      const view = getLocalView(state, id);
+
+      return {
+        ...state,
+        views: {
+          ...state.views,
+          [`${id}`]: {
+            ...view,
+            isShowIncluded: !!showIncludedView,
```
petrican, line 390: why do we have isShowIncluded and also hasShowIncluded with the same value. Couldn't we just use one of them ?

frontend/src/__tests__/reducers/WindowHandler.test.js
```diff
@@ -67,63 +67,5 @@ describe('WindowHandler reducer', () => {
       expect(state).toEqual(expect.objectContaining({ master: expect.objectContaining({ data: { ...data.data } }) }));
     });
```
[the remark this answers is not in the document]
siemiatj: Probably, but for sure not as part of this effort. And Window/MasterWindow/Tabs could need their own refactorings too.

frontend/src/reducers/viewHandler.js
siemiatj: because we have rows in the tables reducer.

frontend/src/reducers/viewHandler.js
siemiatj: I have no idea. Looks like some hack from the old days. I did not wanted to touch it, since it's not in the Table component but looks to be safe.

frontend/src/containers/Table.js
siemiatj: But this is in the container, which is ~270 LOC's so I don't really see much sense. Also we'd then have to pass like 6 or 7 parameters.

frontend/src/containers/Table.js
siemiatj: Same as above.

frontend/src/containers/Table.js
siemiatj: This is not in `BIG_BANG` yet, so I can update that after I'll rebase it with master.

PR #6774, opened 2020-06-02
Add the position in the table to be accessible in the AttributesDropdown

petrican:
Fix popup for the case when we have a solo record.


petrican:
Not OK. Closing as it seems there is a case I wasn't aware when there are more attributes listed in the dropdown

PR #6775, opened 2020-06-02
Show scrollbar when exceeds the visible area #6687

petrican:


petrican:
As agreed we are going to use this solution with the automatic scrolling on vertical when the dropdown exceeds the visible area.  In the future follows to find out a more elegant solution that would not render the scrollbar at all.

Issue #6777, opened 2020-06-02
Fix refresh after running a process


cristinamghita:
### Is this a bug or feature request?
Bug

PR #6778, opened 2020-06-02
#6777 Shall refresh even if is single document

cristinamghita:
#6777 Shall refresh even if is single document

### DIFF
--- a/backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/process/adprocess/ADProcessPostProcessService.java
+++ b/backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/process/adprocess/ADProcessPostProcessService.java
@@ -96,9 +96,10 @@ public class ADProcessPostProcessService
 	public ProcessInstanceResult postProcess(@NonNull final ADProcessPostProcessRequest request)
 	{
 		final ProcessInfo processInfo = request.getProcessInfo();
+		final TableRecordReference currentSingleSelectedDocumentRef = processInfo.getRecordRefOrNull();
 		final ProcessExecutionResult processExecutionResult = request.getProcessExecutionResult();
 
-		invalidateDocumentsAndViews(request.getViewId(), processExecutionResult);
+		invalidateDocumentsAndViews(request.getViewId(), currentSingleSelectedDocumentRef, processExecutionResult);
 
 		return ProcessInstanceResult.builder(extractInstanceId(request))
 				.summary(extractSummary(processExecutionResult))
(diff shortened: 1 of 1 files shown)


backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/process/adprocess/ADProcessPostProcessService.java
```diff
@@ -126,10 +130,19 @@ private void invalidateDocumentsAndViews(final ViewId viewId, final ProcessExecu
...
-			viewSupplier.get().invalidateAll();
-			viewInvalidateAllCalled = true;
+			if (viewSupplier.get() != null)
+			{
+				viewSupplier.get().invalidateAll();
+				viewInvalidateAllCalled = true;
+			}
+			else if (currentSingleSelectedDocumentRef != null)
+			{
+				documentsCollection.invalidateDocumentByRecordId(
```
metas-ts, line 142: wanted to call, but i think it's better to list my questions..
* since we are in the `isRefreshAllAfterExecution()` branch, don't we have to find a way to invalidate all records?
* in case of `isRefreshAllAfterExecution() == false`, shouldn't we *still* make sure to invalidate the selected record?
* in the dev-test you wrote "select a product and run the process". So, i can imagine the test succeeds with a *single* record. but what if more than one record is selected?
* maybe consult with Teo?

cristinamghita: One short, the case of more than one record selected is already covered by previous lines:
image

>in case of isRefreshAllAfterExecution() == false, shouldn't we still make sure to invalidate the selected record?

Hm.. it is wanted? Not sure.

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/process/adprocess/ADProcessPostProcessService.java
```diff
@@ -126,10 +130,19 @@ private void invalidateDocumentsAndViews(final ViewId viewId, final ProcessExecu
 		//
 		// Refresh all
 		boolean viewInvalidateAllCalled = false;
-		if (processExecutionResult.isRefreshAllAfterExecution() && viewSupplier.get() != null)
+		if (processExecutionResult.isRefreshAllAfterExecution())
 		{
-			viewSupplier.get().invalidateAll();
-			viewInvalidateAllCalled = true;
+			if (viewSupplier.get() != null)
+			{
```
metas-ts, line 136:
```suggestion
			{ // multible rows selected
```

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/process/adprocess/ADProcessPostProcessService.java
```diff
@@ -126,10 +130,19 @@ private void invalidateDocumentsAndViews(final ViewId viewId, final ProcessExecu
...
+				viewSupplier.get().invalidateAll();
+				viewInvalidateAllCalled = true;
+			}
+			else if (currentSingleSelectedDocumentRef != null)
+			{
```
metas-ts, line 141:
```suggestion
			{ // single row selected
```

Issue #6782, opened 2020-06-03
Wrong Organisation set to HUs when doing Inventory or Material Receipt


TheBestPessimist:
### Is this a bug or feature request?
BUG

PR #6783, opened 2020-06-03
Wrong Organisation set to HUs when doing Inventory or Material Receipt

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/6782


TheBestPessimist:
Thanks @metas-ts it's a super nice implementation (i _should_ have thought of that) as it uses memoization, so perf should be ok.

Issue #6801, opened 2020-06-05
EDI - Allow getting excel-files from remote sftp


metas-ts:
### Is this a bug or feature request?
FR

PR #6802, opened 2020-06-05
EDI - Allow getting excel-files from remote sftp

metas-ts:
- application.properties: change xls.order.* to excel.order.* to make it clearer
- also rename class files from XLS* to Excel*
- ExcelOLCandRoute: introduce optional remote-to-local-route

https://github.com/metasfresh/metasfresh/issues/6801


metas-ts:
wrong target branch

Issue #6779, opened 2020-06-02
To be invoiced override 0 vs. null


dragospodariu96:
### Is this a bug or feature request?
FR
### What is the current behavior?
Can't set 0 to billing candidates qtytoinvoiceoverride field
#### Which are the steps to reproduce?
Go to the billing candidates window
### What is the expected or desired behavior?
To be able to set 0 to the qty to invoice override field

PR #6811, opened 2020-06-09
#6779 Now able to set 0 to invoice override field on catch weight billing candidates.

dragospodariu96:
#6779


metas-ts:
sorry, wanted to do a tiny comment fix and accidentally merged `master`
will fix it in a new PR

PR #6828, opened 2020-06-11
Extend Invoice reversal API to return affected ICs

pvpurcarcosmin:
https://github.com/metasfresh/metasfresh/issues/6827

### DIFF
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandDAO.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandDAO.java
@@ -3,7 +3,6 @@ package de.metas.invoicecandidate.api.impl;
 import ch.qos.logback.classic.Level;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
-import de.metas.adempiere.model.I_C_Invoice;
 import de.metas.aggregation.model.I_C_Aggregation;
 import de.metas.bpartner.BPartnerId;
 import de.metas.bpartner.service.IBPartnerDAO;
@@ -102,7 +101,6 @@ import java.util.Properties;
 import java.util.Set;
 
 import static org.adempiere.model.InterfaceWrapperHelper.delete;
-import static org.adempiere.model.InterfaceWrapperHelper.load;
 
 /*
  * #%L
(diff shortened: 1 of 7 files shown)


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandDAO.java
```diff
@@ -573,6 +572,30 @@ public boolean isAvoidRecreate(final I_C_Invoice_Candidate ic)
...
+	}
+
+	@NonNull
+	public List<I_C_Invoice_Candidate> retrieveInvoiceCandidates(@NonNull final InvoiceId invoiceId)
+	{
+		final I_C_Invoice invoice = load(invoiceId, I_C_Invoice.class);
+
+		final IQuery<I_C_Invoice_Line_Alloc> invoiceLineAllocQuery =
+				queryBL.createQueryBuilder(I_C_InvoiceLine.class, invoice)
+						.addEqualsFilter(I_C_InvoiceLine.COLUMNNAME_C_Invoice_ID, invoice.getC_Invoice_ID())
```
metas-ts, line 586:
```suggestion
		final IQuery<I_C_Invoice_Line_Alloc> invoiceLineAllocQuery =
				queryBL.createQueryBuilder(I_C_InvoiceLine.class, invoice)
						.addEqualsFilter(I_C_InvoiceLine.COLUMNNAME_C_Invoice_ID, invoiceId)
```

* the equals filter can handle `RepoIdAware`s
* we don't need to pass the invoice a context provider; the query-builder can get the thread-local current transaction or create a "local" one on the fly

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandDAO.java
```diff
@@ -573,6 +572,30 @@ public boolean isAvoidRecreate(final I_C_Invoice_Candidate ic)
...
+						//collect invoice line alloc
+						.andCollectChildren(I_C_Invoice_Line_Alloc.COLUMN_C_InvoiceLine_ID)
+						.addOnlyActiveRecordsFilter()
+						.create();
```
metas-ts, line 590: it's OK like this but you *might* (didn't check) also use `andCollect` to get from `I_C_Invoice_Line_Alloc` to `I_C_Invoice_Candidate`
for next time maybe..

pvpurcarcosmin: Oh, just checked the method, that's pretty cool, I'll give it a try. thx!

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandDAO.java
pvpurcarcosmin: I'll drop the loading of the invoice then, thx!

Issue #6848, opened 2020-06-15
Attribute ID shall not be shown in Attribute Editor/ Attribute Button


TheBestPessimist:
### Is this a bug or feature request?
BUG

PR #6849, opened 2020-06-15
Attribute ID shall not be shown in Attribute Editor/ Attribute Button

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/6848


backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5561410_gh6848_RemoveAll_ASI.Description_WhichAreAn_ASI_ID.sql
```diff
@@ -0,0 +1,42 @@
+-- where clause is the same for both queries
```
metas-ts, line 1: It's ok, but as a hint/suggestion for future: create the backup as you do, and then in the UPDATE part, update those ASI records that are in the backup table
That way I as reviewer would have to diff the two selects to make sure that there are no copy&paste difference-errors. Also i think it's cool

PR #6852, opened 2020-06-15
Fixing tests for tables & cleanup after merge with master

siemiatj:
Some more refactoring too.


frontend/src/components/table/TableWrapper.js
```diff
@@ -0,0 +1,452 @@
...
+import TableContextMenu from './TableContextMenu';
+import TableFilter from './TableFilter';
+import Table from './Table';
+import TablePagination from './TablePagination';
+
+const MOBILE_TABLE_SIZE_LIMIT = 30; // subjective number, based on empiric testing
+const isMobileOrTablet =
+  currentDevice.type === 'mobile' || currentDevice.type === 'tablet';
+
+class Wrapper extends PureComponent {
```
petrican, line 21: Can we name it TableWrapper to be in sync with the filename ?

Issue #6868, opened 2020-06-17
Packing Instruction selectable that does not fit to Businesspartner


TheBestPessimist:
### Is this a bug or feature request?
BUG

PR #6870, opened 2020-06-17
Packing Instruction selectable that does not fit to Businesspartner

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/6868


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/IHUPIItemProductQuery.java
```diff
@@ -32,17 +56,23 @@ default void setProductId(final ProductId productId)
...
-	void setDate(ZonedDateTime date);
+	/**
+	 * Match only those {@link I_M_HU_PI_Item_Product}s which belong to a {@link I_M_ProductPrice} of this PLV.
+	 */
+	void setPriceListVersionId(@Nullable final PriceListVersionId priceListVersionId);
+
+	@Nullable
+	PriceListVersionId getPriceListVersionId();
+
+	void setDate(@Nullable ZonedDateTime date);
```
metas-ts, line 75: note to self: rename to `setValidAtDate` .. i think it's clearer that way

Issue #6873, opened 2020-06-17
Refactoring grids/details view


siemiatj:
This is a ticket for work related to https://github.com/metasfresh/me03/issues/4362, but outside of the scope of the current `metasfresh` issues (like `MasterWindow` and related components, etc).

PR #6877, opened 2020-06-17
 Properly render dataEntries in tabs

siemiatj:
Had to be changed due to the changes in how we handle data in tables now (and removing some unnecessary structures). Now it just reads data from redux, like any other tables.

Related to #6873 & #6706 

Screenshot 2020-06-17 at 16 57 03


siemiatj:
@petrican this is what is causing blank pages in business partners on live7147

PR #6891, opened 2020-06-23
Fix switch pages after edit

petrican:


frontend/src/components/table/TableItem.js
```diff
@@ -255,14 +255,13 @@ class TableItem extends PureComponent {
           for (let [k, v] of Object.entries(fields)) {
             editedCells[k] = v;
           }
```
petrican, line 257: I put the setState within the if (resp[0] && resp[0].fieldsByName) { } as we want to update state only when editedCells was altered.

petrican:
https://www.loom.com/share/f74e2b0435ae4c76a35af17dae9474a4

petrican:
As agreed we leave this as it is and won't merge it until we test with your following PR that everything is going to work as expected. Thanks

petrican:
Merged as we needed a fix for that issue.

PR #6916, opened 2020-06-26
Send selection on process creation

petrican:


petrican:
https://www.loom.com/share/fcc27c9541b5487ea1ee5e3f990bc8ed

siemiatj:
I think it works fine in the Order Line. No idea how it's supposed to work with Order Tax for instance, as we're getting `pricing conditions` in the actions request. But maybe it's how it's implemented.

https://recordit.co/CJBTx6TgUu

PR #6929, opened 2020-07-01
Gh6867 2

metas-ts:
See https://github.com/metasfresh/metasfresh/pull/6906#issuecomment-652176241


backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/DocLine_InOut.java
```diff
@@ -159,7 +159,9 @@ public CostAmount getCreateShipmentCosts(final AcctSchema as)
 					.initialDocumentRef(CostingDocumentRef.ofShipmentLineId(getReversalLine_ID()))
 					.date(getDateAcct())
 					.build())
-					.getTotalAmountToPost(as);
+					.getTotalAmountToPost(as)
+					// Negate the amount coming form the costs because it must be negative in the accounting.
```
metas-ts, line 163:
```suggestion
					// Negate the amount coming from the costs because it must be negative in the accounting.
```

PR #6952, opened 2020-07-02
Try to fix the remaining failing e2e tests

siemiatj:
Looks like there were some changes in the API.


siemiatj:
@petrican do you know if we can run e2e from a branch or is it always master ?

petrican:
> @petrican do you know if we can run e2e from a branch or is it always master ?

@siemiatj  lets you jump to a job that will perform an e2e-test using this branch's latest e2e-docker image.

PR #6960, opened 2020-07-06
add PostgREST process type

pvpurcarcosmin:
https://github.com/metasfresh/metasfresh/issues/6957

### DIFF
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/postgrest/config/PostgRESTConfigRepository.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/postgrest/config/PostgRESTConfigRepository.java
@@ -1,0 +1,82 @@
+/*
+ * #%L
+ * de.metas.adempiere.adempiere.base
+ * %%
+ * Copyright (C) 2020 metas GmbH
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+package de.metas.postgrest.config;
+
+import de.metas.cache.CCache;
+import de.metas.organization.OrgId;
+import de.metas.util.Services;
+import lombok.NonNull;
+import org.adempiere.ad.dao.IQueryBL;
+import org.adempiere.exceptions.AdempiereException;
+import org.compiere.model.I_S_PostgREST_Config;
+import org.springframework.stereotype.Repository;
+
+import java.util.Optional;
+
+@Repository
+public class PostgRESTConfigRepository
+{
+	private final IQueryBL queryBL = Services.get(IQueryBL.class);
+
+	private final CCache<OrgId, Optional<I_S_PostgREST_Config>> cache = CCache.<OrgId, Optional<I_S_PostgREST_Config>> builder()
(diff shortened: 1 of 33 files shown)


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/postgrest/config/PostgRESTConfigRepository.java
```diff
@@ -0,0 +1,82 @@
...
+	{
+		return cache.getOrLoad(orgId, this::retrieveConfigFor);
+	}
+
+	private Optional<I_S_PostgREST_Config> retrieveConfigFor(@NonNull final OrgId orgId)
+	{
+		return queryBL
+				.createQueryBuilder(I_S_PostgREST_Config.class)
+				.addOnlyActiveRecordsFilter()
+				.addEqualsFilter(I_S_PostgREST_Config.COLUMNNAME_AD_Org_ID, orgId)
```
metas-ts, line 78: AFAIU the postgrest-config is not very org-specific, although it might be good to have the ability for org-specific tweaks..
However, i juggest to rather filter for `inArray(I_S_PostgREST_Config.COLUMNNAME_AD_Org_ID, orgId, OrgID.ANY)` (the filter method's real name is different) and then sort by AD_Org_ID descending

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/process/impl/ProcessRestController.java
```diff
@@ -0,0 +1,232 @@
...
+				.ok()
+				.contentType(MediaType.APPLICATION_JSON_UTF8)
+				.body(response);
+	}
+
+	private ResponseEntity<?> getResponse(@NonNull final ProcessExecutionResult processExecutionResult)
+	{
+		if (processExecutionResult.isError())
+		{
+			return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(Message.of("Ups! Looks like something went wrong on our end..."));
```
metas-ts, line 148: i'm not sure its great to send the whole stacktrace or if that's too much..
but on the other hand, here i think at least the exception class and message should be shown..and maybe, also the whole stacktrace after all (check e.g. the olcand-record).

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/process/impl/ProcessRestController.java
```diff
@@ -0,0 +1,232 @@
...
+
+			return ResponseEntity.ok()
+					.contentType(MediaType.parseMediaType(contentType))
+					.header(HttpHeaders.CONTENT_DISPOSITION,"attachment; filename=\"" + processExecutionResult.getReportFilename() + "\"")
+					.body(processExecutionResult.getReportData());
+		}
+		else
+		{
+			//this should never happen
+			return ResponseEntity.noContent().build();
```
metas-ts, line 171: if it should never happen, i suggest you throw an exception or do `Check.fail(...)`

backend/de.metas.business.rest-api/src/main/java/de/metas/rest_api/process/response/JSONProcessParamBasicInfo.java
```diff
@@ -0,0 +1,42 @@
...
+@Builder
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonDeserialize(builder = JSONProcessParamBasicInfo.JSONProcessParamBasicInfoBuilder.class)
+public class JSONProcessParamBasicInfo
+{
+	@JsonProperty("name")
+	String name;
+
+	@JsonProperty("description")
+	String description;
```
metas-ts, line 41: might turn out that we also need the type (integer, date, date+time, string) amd `ColumnName` in this class

backend/de.metas.report/de.metas.report.jasper.client/src/main/java/de/metas/report/jasper/client/interceptor/AD_Process.java
```diff
@@ -1,5 +1,11 @@
 package de.metas.report.jasper.client.interceptor;
```
metas-ts, line 1: pls move this class to a more "generic" module&package..or if that's not feasilbe, another possibility might be to create one for jasper and one for postgrest and write comments to mae sure future devs don't change them to interfere with each other

PR #6969, opened 2020-07-07
Downgrade js-base64

petrican:
We have similar issue as in here https://github.com/meinaart/cypress-plugin-snapshots/issues/126 and we need to downgrade

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -44,6 +44,7 @@
     "axios": "^0.18.1",
     "cypress-plugin-snapshots": "^1.2.6",
     "cypress-skip-and-only-ui": "^1.0.12",
+    "js-base64": "2.5.2",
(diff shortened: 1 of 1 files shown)


TheBestPessimist:
I took a look at the build you gave me: https://jenkins.metasfresh.com/job/ops/job/run_e2e_tests/927/consoleFull

image

There are 80 hits for `js-base64`, and all of them are for `cypress-plugin-snapshots`. 

Wouldn't it make more sense to pin the old version of `cypress-plugin-snapshots` instead of `js-base64`?

TheBestPessimist:
talked with petrica on skype. it's easier to pin the version of `js-base64` instead of trying to figure out which `cypress-plugin-snapshots` uses `js-base64<2.5.1`.

PR #6982, opened 2020-07-08
Fix picking shifting edge cases

petrican:
There is a function where we omitted to replace the logic with indentSupported.

https://www.loom.com/share/ec047bc109624f12bb34ba51cbd9e84f


petrican:
The initial logic for the check was bad as we didn't used `indentSupported`. Frankly I wasn't aware of all the cases where it should happen.

PR #6990, opened 2020-07-09
Fix catch weight issue

petrican:


petrican:
https://www.loom.com/share/5c787dd5c86d4fefa2836e9f9fdde493

petrican:
> Looks good but please add a test for this `validatePrecision` helper.

Tests for that function were added along with the refactoring of MasterWidget component in here https://github.com/metasfresh/metasfresh/blob/master/frontend/src/__tests__/utils/widgetHelper.test.js#L43

PR #6994, opened 2020-07-09
[Frontend] Fix selections when whole table is updated

siemiatj:
Fixes resetting selection when table data is reloaded and an annoying bug when setting/resetting widget value wouldn't sometimes work.


siemiatj:
Adding tests and fixing another issue with widgets in the same effort.

petrican:
e2e suite is passing fine - https://dashboard.cypress.io/projects/5yp4q1/runs/800/specs  . Need to do some double checks tomorrow morning and will merge this. Thanks

PR #6911, opened 2020-06-25
#1253 Work In Progress - Clone Lines.

dragospodariu96:
https://github.com/metasfresh/metasfresh-webui-api-legacy/issues/1253

PR #6996, opened 2020-07-10
#1253 No longer showing 'Clone Line' action when subtab's table is the same as the document's.

dragospodariu96:
#6911


backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/process/WEBUI_CloneLine.java
```diff
@@ -73,6 +73,11 @@ public ProcessPreconditionsResolution checkPreconditionsApplicable(final @NonNul
 		{
 			return ProcessPreconditionsResolution.rejectWithInternalReason("CopyRecordFactory not enabled for the table the row belongs to.");
 		}
+
+		if (ref.getTableName().equals(context.getTableName()))
+		{
+			return ProcessPreconditionsResolution.rejectWithInternalReason("Same Table as in document in sub-tab, can only have one line.");
```
metas-ts, line 79:
```suggestion
			// we have e.g. in the C_BPartner-window two subtabs ("Customer" and "Vendor") that show a different view on the same C_BPartner record. 
			// There, cloning the subtab-line makes no sense
			return ProcessPreconditionsResolution.rejectWithInternalReason("The main-window has the same Record as the sub-tab, there can only be one line.");
```

PR #7000, opened 2020-07-10
[Frontend] Fix filtering with pagination

siemiatj:
This is a contained change, so feel free to merge. I've checked all the filtering cases I could think of and everything seems to be running just fine.

https://recordit.co/QOEKJKT9mo


siemiatj:
e2e tests all good: https://jenkins.metasfresh.com/job/ops/job/run_e2e_tests/936/

petrican:
Thank you very much @siemiatj !

Issue #6967, opened 2020-07-06
Refreshing page looses pagination


siemiatj:
### Is this a bug or feature request?

Bug

### What is the current behavior?

https://recordit.co/VELjrQsCtV

TheBestPessimist:
# 🔍 IT

## Status: :white_check_mark: Pass

IT done on instance: localhost

1. :white_check_mark: Refresh on first page with no filters applied: view does not change
2. :white_check_mark: Refresh on different pages, including the last page: view does not change, page is not reset to 1
3. :white_check_mark: Add default filter and test steps 1,2: view does not change, page is not reset to 1
4. :white_check_mark: Add frequent filter and default filter and test steps 1,2: view does not change, page is not reset to 1

![image](https://user-images.githubusercontent.com/4482210/87653158-d1904380-c75d-11ea-9eb8-6d61b157ab8b.png)


Thank you for the fix, @PureLandFlying !

PR #7001, opened 2020-07-12
Refreshing page looses pagination #6967

PureLandFlying:
Related to #6967


petrican:
Thanks I've checked and it fixes the issue.

PR #7002, opened 2020-07-13
Fix selection for parent as a prerequisite for starting process

petrican:
sends correct parentViewId and corresponding selection

### DIFF
--- a/frontend/src/components/app/Modal.js
+++ b/frontend/src/components/app/Modal.js
@@ -154,6 +154,7 @@ class Modal extends Component {
       activeTabId,
       childViewId,
       childViewSelectedIds,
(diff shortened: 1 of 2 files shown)


petrican:
https://www.loom.com/share/2cf26314e27942428f579b3cd296e184

petrican:
After running the e2e suite I discovered that this is quite sensitive and is breaking some tests. Need to find a fix that would not break those.

petrican:
Closing this one @siemiatj - I have https://github.com/metasfresh/metasfresh/pull/7006 which does set correctly the windowId for the getTableId. Also for that one all the e2e tests are passing

PR #7035, opened 2020-07-20
Properly refresh viewId for included view

siemiatj:
I still have to figure out a way to test it but other than that - good to go.

### DIFF
--- a/frontend/src/actions/ViewActions.js
+++ b/frontend/src/actions/ViewActions.js
@@ -282,7 +282,8 @@ export function fetchDocument({
 
         dispatch(updateGridTable(tableId, tableData));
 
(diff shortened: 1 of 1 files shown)


siemiatj:
There's something weird happening on the backend: https://recordit.co/7EGGKnfpvi . Sometimes it sends the same `includedView` for other rows even though it shouldn't. Sometimes it even does that after I unpick HU. But it's completely random.

petrican:
Tried to deploy a dev instance with it but it failed https://chat.metasfresh.org/metasfresh/pl/tjbgm8htj3badyje8n79wb48ia. Will check in the morning. Thanks @siemiatj

PR #7038, opened 2020-07-21
Active field colour after patch only

petrican:
https://www.loom.com/share/c89bad4fd26c46d9a9dfaacfa574a989

As agreed not doing the highlighting would at least not raise the expectation of it was saved


siemiatj:
Prettier error, other than that good to merge

PR #7040, opened 2020-07-22
Customer return not working bc HUs cannot be selected

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/6420


TheBestPessimist:
approved by teo on MM

Issue #6997, opened 2020-07-10
Picking Terminal Allow partial delivery


TheBestPessimist:
### Is this a bug or feature request?
FR

PR #7055, opened 2020-07-24
Do not show action Picking Terminal V2 -> Picking -> Process if there are no rows available for processing

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/6997

### DIFF
--- a/backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/pickingV2/productsToPick/process/ProductsToPick_4EyesReview_ProcessAll.java
+++ b/backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/pickingV2/productsToPick/process/ProductsToPick_4EyesReview_ProcessAll.java
@@ -76,6 +76,12 @@ public class ProductsToPick_4EyesReview_ProcessAll extends ProductsToPickViewBas
 			return ProcessPreconditionsResolution.rejectWithInternalReason("only picker shall be allowed to process");
 		}
 
+		final List<ProductsToPickRow> rows = getRowsNotAlreadyProcessed();
+		if (rows.isEmpty())
+		{
+			return ProcessPreconditionsResolution.rejectWithInternalReason("no unprocessed rows");
+		}
+
 		if (partialDeliveryNotAllowed())
 		{
 			if (!getView().isApproved())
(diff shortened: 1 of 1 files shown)


backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/pickingV2/productsToPick/process/ProductsToPick_4EyesReview_ProcessAll.java
```diff
@@ -76,23 +76,31 @@ protected ProcessPreconditionsResolution checkPreconditionsApplicable()
 			return ProcessPreconditionsResolution.rejectWithInternalReason("only picker shall be allowed to process");
 		}
 
+		final List<ProductsToPickRow> rows = getRowsNotAlreadyProcessed();
```
metas-rc, line 79: I think you should check if there are **unpacked** rows as well, not only the ones that are not already processed. If the rows were picked but not yet packed, the "Process"  action is available and one can set the shipper transportation and believe the stuff really got in there, when in fact they didn't.

TheBestPessimist: If there are Picked rows but not Packed, the Process action does not appear.

image

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/pickingV2/productsToPick/process/ProductsToPick_4EyesReview_ProcessAll.java
```diff
@@ -205,8 +213,8 @@ private void deliverAndInvoice(final List<PickingCandidate> pickingCandidates)
 				.collect(ImmutableList.toImmutableList());
 	}
 
-	private boolean partialDeliveryNotAllowed()
+	private boolean partialDeliveryAllowed()
```
teosarca, line 216: why not prefixing with `is`?

TheBestPessimist: done

PR #7122, opened 2020-08-11
Gh7117

pvpurcarcosmin:


backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/issues/WEBUI_ServicesCompanies_Toggle_Processed.java
```diff
@@ -0,0 +1,93 @@
...
+		{
+			throw new AdempiereException("@NoSelection@");
+		}
+
+		selectedIssueIds
+				.stream()
+				.map(IssueId::ofRepoId)
+				.map(issueRepository::getById)
+				.peek(issueEntity -> issueEntity.setProcessed(processed))
+				.forEach(issueRepository::save);
```
metas-ts, line 78: There might be a lot of records to process, so i think it's better to go with 
* `IQueryBL#createCompositeQueryUpdater` (=> see e.g. `BPBankAccountDAO#deactivateIBANAccountsByBPartnerExcept`)
* then do `IQuery#updateDirectly` (=> see e.g. `AbstractSubscriptionDAO#increaseSeqNosOfSuccessors`)
* then do `IModelCacheInvalidationService#invalidate`  (=> imho should be OK to invalidate all, with `CacheInvalidateMultiRequest#allRecordsForTable`)

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/issues/WEBUI_ServicesCompanies_Toggle_Processed.java
```diff
@@ -0,0 +1,93 @@
...
+	protected String doIt() throws Exception
+	{
+		final IQueryBuilder<I_S_Issue> selectionQuery = queryBL.createQueryBuilder(I_S_Issue.class);
+
+		final IQueryFilter<I_S_Issue> selectionFilter = getProcessInfo().getQueryFilterOrElse(null);
+		if (selectionFilter == null)
+		{
+			throw new AdempiereException("@NoSelection@");
+		}
+		selectionQuery.filter(selectionFilter);
```
metas-ts, line 64: also filter out thise records that already have the target `processed` value set

PR #7124, opened 2020-08-12
Table edit shortcut

petrican:


petrican:
Demo on local box https://www.loom.com/share/f0d15fddd3be4d7fafb9ef49d32ac4b1

petrican:
Demo on  https://dev1253.metasfresh.com

https://www.loom.com/share/76e07a9bb04d4aafadad7f0cfb60c1c6

frontend/src/components/table/TableItem.js
```diff
@@ -75,7 +75,7 @@ class TableItem extends PureComponent {
     }
   }
 
-  initPropertyEditor = (fieldName) => {
+  initPropertyEditor = ({ fieldName, mark }) => {
```
siemiatj, line 78: what `mark` stands for ?

frontend/src/components/table/TableItem.js
```diff
@@ -195,9 +203,9 @@ class TableItem extends PureComponent {
    * @param {object} [item] - widget data object
    * @param {boolean} [select] - flag if selected cell should be cleared
    */
-  handleEditProperty = (e, property, focus, item, select) => {
+  handleEditProperty = (e, property, focus, item, select, mark) => {
```
siemiatj, line 206: Same -> add a comment what `mark` property is.

frontend/src/components/table/TableItem.js
```diff
@@ -212,9 +220,9 @@ class TableItem extends PureComponent {
    * @param {object} [item] - widget data object
    * @param {boolean} [select] - flag if selected cell should be cleared
    */
-  _editProperty = (e, property, focus, item, select) => {
+  _editProperty = ({ e, property, focus, item, select, mark }) => {
     if (item ? !item.readonly : true) {
-      if (this.state.edited === property && e) e.stopPropagation();
+      if (this.state.edited === property && e) e.persist(); // e.stopPropagation();
```
siemiatj, line 225: Why do we need to persist the event ? Are we using it in any way and need to access it past this point ?

frontend/src/components/table/TableWrapper.js
```diff
@@ -156,6 +156,32 @@ class TableWrapper extends PureComponent {
...
+      },
+      supportFieldEdit,
+    });
+  };
+
+  handleFastInlineEdit = () => {
+    const { selected } = this.props;
+    const { contextMenu, supportFieldEdit } = this.state;
+    // a routine needs to be added in here - to check if field is allowed to be edited
+    // if (contextMenu.supportFieldEdit && selected.length === 1) {
```
siemiatj, line 173: What's this comment for ? Is this `routine` something we need now ? In the future ?

petrican: Is just from an initial attempt to restrict using the shortcut. Commented out and forgot to remove the comment. Removed now

frontend/src/components/table/TableItem.js
petrican: Done

frontend/src/components/table/TableItem.js
petrican: Done

frontend/src/components/table/TableItem.js
petrican: Screenshot 2020-08-17 at 16 58 36

It is being use when you get the focus on child components. If you are to use e.stopPropagation while editing you will get a bunch of warnings as in the screenshot. This is the reason I removed it and persisted the event.

frontend/src/components/table/TableItem.js
```diff
@@ -228,7 +228,7 @@ class TableItem extends PureComponent {
    */
   _editProperty = ({ e, property, focus, item, select, mark }) => {
     if (item ? !item.readonly : true) {
-      if (this.state.edited === property && e) e.persist(); // e.stopPropagation();
+      if (this.state.edited === property && e) e.persist();
```
siemiatj, line 231: My comment was about `persist`, not `stopPropagation`

petrican: "If you want to access the event properties in an asynchronous way, you should call event.persist() on the event, which will remove the synthetic event from the pool and allow references to the event to be retained by user code."  - With other words if we do a setState down below using the event and we need data from it we use that
Screenshot 2020-08-18 at 07 54 24

If you do otherwise you'll get nullified event props.

Screenshot 2020-08-18 at 07 57 27

PR #7130, opened 2020-08-13
collecting classes for same application in same package

metas-ts:
..before, it was all routes in one package, all pojos in one package etc
Now it's e.g. all the stepcom.code in one package


metas-ts:
wrong traget branch :man_facepalming:

PR #7132, opened 2020-08-13
Restore the initial value on Escape press

petrican:


petrican:
https://www.loom.com/share/de3e68132e1146dcbb7a236791b28b3f

PR #7138, opened 2020-08-14
Fix case that failed on it

petrican:


petrican:
https://www.loom.com/share/2ad137056b53478b9fd12201a872a470

TheBestPessimist:
Functionally it works ok now.

PR #7141, opened 2020-08-14
implement two minor changes

metas-ts:
Product-Prefix
when convertings shipmentCandidates and receiptCandidtes to XML, we use a mapping org-value => prefix from shipping.properties and  set _artikel_nummer to ""<prefix>-<productNo>"

DocumentNote
* REST-API: in shipmentCandidates and receiptCandidtes, we now also export M_Product.DocumentNote
* when creating XML for shipmentCandidates, we set  _artikel_hinweistext from DocumentNote

https://github.com/metasfresh/metasfresh/issues/7140


metas-ts (commented):
lgtm

PR #7161, opened 2020-08-19
Fix NPE if an invoice candidate is in catch weight but none is given

metas-ts:
in this case, (like elsewhere too) we fall back to the nominal weight


metas-ts (commented):
lgtm

Issue #3856, opened 2018-04-14
WebUI Forecast Window remoce Calendar, Period, Year


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, in Window Forecast there are Fields Calendar (mandatory), Year (mandatory) and Period.
#### Which are the steps to reproduce?
Open, try and see.
### What is the expected or desired behavior?
The Window shall be adjusted. The Fields shall be hidden from Window (Header and Rows). Instead, the Promised Date shall be used only (mandatory).

PR #7169, opened 2020-08-20
Forecast: set datepromised if not exists

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/3856

### DIFF
--- a/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5565380_sys_gh3856_setM_Forecast.DatePromissed_IfNotExists.sql
+++ b/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5565380_sys_gh3856_setM_Forecast.DatePromissed_IfNotExists.sql
@@ -1,0 +1,18 @@
+UPDATE m_forecast
+SET datepromised = (
+    SELECT startdate::Timestamp
+    FROM c_period p
+    WHERE p.c_period_id = m_forecast.c_period_id
+
+    UNION
+
+    SELECT to_timestamp(fiscalyear, 'YYYY') /* eg. 2020.01.01 */
+    FROM c_year y
+    WHERE y.c_year_id = m_forecast.m_forecast_id
+
+    LIMIT 1
+),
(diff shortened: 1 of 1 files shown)


backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5565380_sys_gh3856_setM_Forecast.DatePromissed_IfNotExists.sql
```diff
@@ -0,0 +1,18 @@
...
+    FROM c_period p
+    WHERE p.c_period_id = m_forecast.c_period_id
+
+    UNION
+
+    SELECT to_timestamp(fiscalyear, 'YYYY') /* eg. 2020.01.01 */
+    FROM c_year y
+    WHERE y.c_year_id = m_forecast.m_forecast_id
+
+    LIMIT 1
```
metas-ts, line 13: u think you need to do sth like
```suggestion
    SELECT 1 as seq, startdate::Timestamp
    FROM c_period p
    WHERE p.c_period_id = m_forecast.c_period_id

    UNION

    SELECT 2, to_timestamp(fiscalyear, 'YYYY') /* eg. 2020.01.01 */
    FROM c_year y
    WHERE y.c_year_id = m_forecast.m_forecast_id <-- this seems wrong
    
    UNION 
    
    SELECT 3, '2010-01-01'
    
    ORDER BY seq
    LIMIT 1
```

* to make sure your subselect prefers the ones with a period_id
* m_forecast.c_year_id is not mandatory, so AFAIU we need a 3rd fallback

metas-ts (changes requested):

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5565380_sys_gh3856_setM_Forecast.DatePromissed_IfNotExists.sql
TheBestPessimist: Fixed

Issue #7171, opened 2020-08-20
Order candidates processing: Paypal/ creditcard payments w/ reference and prepayment


TheBestPessimist:
### Is this a bug or feature request?
bug

PR #7172, opened 2020-08-20
Allow allocation of Prepay Sales Orders 

TheBestPessimist:
We already allow payment allocation of Sales Order with DocType Prepay. 

We want to allow allocation for Sales Orders which are of DocType Sales Order and have a payment with matching external ID

https://github.com/metasfresh/metasfresh/issues/7171


backend/de.metas.business/src/main/java/de/metas/invoice/interceptor/C_Invoice.java
```diff
@@ -294,12 +297,34 @@ private void linkInvoiceToPaymentIfNeeded(final I_C_Invoice invoice)
...
+		if (order == null)
+		{
+			return false;
+		}
+		if (order.getC_Payment_ID() <= 0)
+		{
+			return false;
+		}
+
+		final boolean isPrepayOrder = Services.get(IDocTypeBL.class).isPrepay(DocTypeId.ofRepoId(order.getC_DocType_ID()));
```
metas-ts, line 312: not a critical request, but please extract `Services.get(...)` to a field, so that the services are all visible at the class beginning and not stashed away in the code

backend/de.metas.business/src/main/java/de/metas/invoice/interceptor/C_Invoice.java
```diff
@@ -294,12 +297,34 @@ private void linkInvoiceToPaymentIfNeeded(final I_C_Invoice invoice)
...
+		if (isPrepayOrder)
+		{
+			return true;
+		}
+
+		final ExternalId paymentExternalId = Services.get(IPaymentDAO.class).getExternalId(PaymentId.ofRepoId(order.getC_Payment_ID()));
```
metas-ts, line 318: ❗ why do the external IDs have to match? Mark doesn't know, maybe you discussed this with @metasnw ?

To put it differently:
* you have a sales order that reverences a payment.
* you invoice this sales order
* => why not allocate the sales order's payment to the invoice?

If you need to keep it like this, please add it to the documentation of `C_Order.C_Payment_ID`, 'C_Order.ExternalId` and `C_Payment.ExternalOrderId`, because i think it's intuitive to "forward" the referenced payment from the order..but if it's only forwarded if the external-ids match, then this needs to be be documented.

metas-ts (changes requested):

metas-ts (approved):

backend/de.metas.business/src/main/java/de/metas/invoice/interceptor/C_Invoice.java
TheBestPessimist: done

backend/de.metas.business/src/main/java/de/metas/invoice/interceptor/C_Invoice.java
TheBestPessimist: confirmed with torby: we allow any payment linked to the Sales Order, not just those with matching external ID.

PR #7182, opened 2020-08-24
Improve sales invoice status endpoint

metas-ts:
* Allow {invoiceDocumentNo} to be a prefix
* Return the 2-letter document status, so the caller
* isPaid true/false
* metasfreshId (invoice-ID)
* also improve javadocs a bit here and there


metas-ts (commented):
lgtm

PR #9841, opened 2020-08-25
Fix Jenkins build

siemiatj:
Related to #9840

PR #9842, opened 2020-08-25
Update dev packages and  remove the additional jest install step

siemiatj:
Related to #9841 . We've had all the dependencies frozen in `packages.json` but during the build process in CI there was a separate installation of `jest` test runner, which automatically installed the latest version. Since we're already using it, there's not really a need to do that (and also having a fixed version defined in the `packages` we know that it'll work :)). Let me know if my thinking is not correct.

Also updated a bunch of packages.


petrican:
> Related to #9841 . We've had all the dependencies frozen in `packages.json` but during the build process in CI there was a separate installation of `jest` test runner, which automatically installed the latest version. Since we're already using it, there's not really a need to do that (and also having a fixed version defined in the `packages` we know that it'll work :)). Let me know if my thinking is not correct.
> 
> Also updated a bunch of packages.

Your thinking is correct. Need to do some testing with this and will merge after. Thanks

petrican (approved):
lgtm. Thanks!

petrican:
Screenshot 2020-08-26 at 09 46 03
There's seem to be an issue that I spot after merging this. There's this warning that we need to fix also @siemiatj

frontend/buildfile.groovy
```diff
@@ -55,7 +55,6 @@ Map build(final MvnConf mvnConf, final Map scmVars, final boolean forceBuild=fal
 		sh 'yarn install'
 		sh 'yarn lint --quiet'
 							
-		sh "yarn add jest jest-junit --dev"
```
metas-ts, line 58: @siemiatj fine with me! :+1:

PR #9843, opened 2020-08-25
7048 fix cypress

petrican:
- Fix for failing cypress tests
- also upgraded Cypress to v5.0.0


petrican:
All tests are passing https://dashboard.cypress.io/projects/5yp4q1/runs/836/specs Tested against https://dev1267.metasfresh.com

e2e/Dockerfile
```diff
@@ -3,7 +3,7 @@
 #FROM cypress/base:10
 
 # See the available Docker images here: https://github.com/cypress-io/cypress-docker-images/tree/master/browsers
-FROM cypress/browsers:node13.6.0-chrome80-ff72
+FROM cypress/browsers:node14.7.0-chrome84
```
siemiatj, line 6: Should we also consider updating `node` to 14 elsewhere ? I think now Jenkins is running 13.6

siemiatj (approved):

e2e/Dockerfile
petrican: Yeah, probably in the near future. In this case I had to use the recommended image from cypress as it contains also the dependencies needed for Cypress 5.0.0 to run

Issue #7107, opened 2020-08-06
Date in modal overlay is not set correctly when double-clicking in calendar widget


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
When selecting a date with double click in a calendar widget, in a modal overlay, the date is not always set correctly.

#### Which are the steps to reproduce?
1. open bpartner window
2. call "create request" action
3. in reminder date, open the calendar widget, and select a date from **the following month** with double click: a different date is set, not the one u selected
note: this works when selecting the date with single click, and also with double click in main window
### What is the expected or desired behavior?
Correct date is set.

siemiatj:
@metas-dh Can't reproduce this neither on releasetestit nor releasetestuat:
https://recordit.co/lhV1aI7eqE

metas-dh:
@siemiatj did u actually use double click for selection? from the vid it looks like u single click on the date and then click somewhere else in the window, which works fine; the double clicking creates the problem.

PR #9859, opened 2020-08-28
Fix handling events in the datepicker

siemiatj:
Turns out I had some local changes that somehow didn't make it to the previous PR.

Related to #7107


petrican (approved):
lgtm. Thanks!

Issue #9862, opened 2020-08-31
Show serial number in production HU selection window


TheBestPessimist:
### Is this a bug or feature request?
FR

PR #9866, opened 2020-08-31
Show serial number in production HU selection window

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/9862

### DIFF
--- a/backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/handlingunits/HUEditorRow.java
+++ b/backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/handlingunits/HUEditorRow.java
@@ -1,27 +1,9 @@
 package de.metas.ui.web.handlingunits;
 
-import java.math.BigDecimal;
-import java.time.LocalDate;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Optional;
-import java.util.function.Function;
-
-import javax.annotation.Nullable;
-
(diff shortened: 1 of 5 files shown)


backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/handlingunits/HUEditorRow.java
```diff
@@ -264,6 +271,21 @@ private HUEditorRow(@NonNull final Builder builder)
...
 			attributesSupplier = Optional.empty();
 		}
+
+		if (hasAttributes() && attributesSupplier.get().get().hasAttribute(AttributeConstants.ATTR_SerialNo))
+		{
+			serialNo = attributesSupplier.get().get().getValueAsString(AttributeConstants.ATTR_SerialNo);
+		}
+		else
+		{
+			serialNo = "A";
```
metas-ts, line 281: does that mean we print an `A` if an HU has no serial number?

TheBestPessimist: on a first look, yes it should.
however, in my local testing i never got the "A" displayed. The column was always empty or with a valid string. Now i wonder why.

TheBestPessimist: Fixed

metas-ts (approved):

Issue #9853, opened 2020-08-27
Refactoring filters


petrican:
### Is this a bug or feature request?
F
### What is the current behavior?
Filters don't have their own branch in the redux store. We need to have that to separate the logic
#### Which are the steps to reproduce?
Monitor the redux store
### What is the expected or desired behavior?
Look in the redux store for the filter branch. Any action done to filters should reflect in the content under that branch.

PR #9870, opened 2020-08-31
9853 filters refactoring

petrican:
Related to #9853


frontend/src/components/filters/Filters.js
```diff
@@ -2,604 +2,131 @@ import counterpart from 'counterpart';
...
-    combinedFilters.forEach((item) => {
-      mappedFiltersData.set(item.filterId, item);
-    });
-    return mappedFiltersData;
-  };
-
-  // PARSING FILTERS ---------------------------------------------------------
-
-  /*
-   * parseActiveFilters - this function does three things:
```
siemiatj, line 87: As far as I understand this was divided into smaller functions. Comments were also carried over, right ?

frontend/src/components/filters/Filters.js
```diff
@@ -2,604 +2,131 @@ import counterpart from 'counterpart';
...
-      }
-
-      return acc;
-    }, []);
-  };
-
-  // SETTING FILTERS  --------------------------------------------------------
   /**
    * @method applyFilters
    * @summary This method should update docList
```
siemiatj, line 29: Well it doesn't really update the docList, since filters are in the store and DL should react automatically.

frontend/src/components/filters/Filters.js
```diff
@@ -2,604 +2,131 @@ import counterpart from 'counterpart';
...
-            }
-          });
+    const { updateDocList, filterId, updateActiveFilter } = this.props;
+    const { filtersActive: storeActiveFilters } = this.props.filters;
 
-        return filter;
-      });
-    }
+    // updating the active filters from the redux store with the filter passed as param
+    const newFiltersActive = setNewFiltersActive({
```
siemiatj, line 63: is there a case, where we set new active filter without updating active filter ? Maybe `setNewFiltersActive` should call `updateActiveFilter` ?

frontend/src/components/filters/Filters.js
```diff
@@ -678,37 +206,43 @@ class Filters extends PureComponent {
   }
 }
 
-/**
```
siemiatj, line 681: Why remove the proptypes comments ?

frontend/src/components/filters/FiltersFrequent.js
```diff
@@ -57,9 +50,7 @@ class FiltersFrequent extends PureComponent {
    * @summary ToDo: Describe the method
    * @todo Write the documentation
    */
-  handleClickOutside = () => {
-    this.outsideClick();
-  };
+  handleClickOutside = () => this.outsideClick();
```
siemiatj, line 53: Why not move logic from `outsideClick` here and remove it completely ?

frontend/src/containers/DocumentList.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable prettier/prettier */
```
siemiatj, line 1: ?

siemiatj (changes requested):
Looks good overall. Would love to see some tests, especially that reducer/actions unit tests are fairly straightforward.

frontend/src/actions/FiltersActions.js
```diff
@@ -1,8 +1,357 @@
...
+  };
+}
+
+/**
+ * @method updateWidgetShown
+ * @summary Updates the widgetShown in the store for the corresponding entity id with a boolean value
+ */
+export function updateWidgetShown({ id, data }) {
+  return {
+    type: types.UPDATE_WIDGET_SHOWN,
```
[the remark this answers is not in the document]
petrican: Done

frontend/src/actions/FiltersActions.js
```diff
@@ -1,8 +1,357 @@
...
+  };
+}
+
+/**
+ * @method updateNotValidFields
+ * @summary updates in the store the notValidFields flag
+ */
+export function updateNotValidFields({ filterId, data }) {
+  return {
+    type: types.UPDATE_FLAG_NOTVALIDFIELDS,
```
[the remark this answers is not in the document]
petrican: because filters were designed with this logic and was not in the plan to have it for all widgets. We can add such logic with the widget refactoring.

frontend/src/components/filters/Filters.js
petrican: I believe those are kind of redundant since Proptypes itself is pretty much self-explanatory for every field since you have the type. If you have 20 params there will be extra 20 lines. If for some special param like an object you want to add details you can put comment but if you just put @param {object} someObject then I believe is redundant.. since someObject: Proptype.object, is obvious what it is.

frontend/src/containers/DocumentList.js
petrican: Pfff. This must have been added somehow automatically by the editor. Thanks. Removed

frontend/src/utils/documentListHelper.js
```diff
@@ -142,7 +141,8 @@ const DLmapStateToProps = (state, props) => {
     parentSelected: parentSelector(state, parentTableId),
     modal: state.windowHandler.modal,
     rawModalVisible: state.windowHandler.rawModal.visible,
-    filters: state.filters,
+    filters: windowId && viewId && state.filters ? state.filters[filterId] : {},
```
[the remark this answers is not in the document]
petrican: Hmm. Not sure. This is faster

frontend/src/actions/FiltersActions.js
```diff
@@ -1,8 +1,357 @@
...
+export function clearStaticFilters({ filterId, data }) {
+  return {
+    type: types.CLEAR_STATIC_FILTERS,
+    payload: { filterId, data },
+  };
+}
+
+/**
+ * @method getParentFilter
+ * @summary as the name suggests the function is retrieving the filter data by key from the filterData
```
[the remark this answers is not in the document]
petrican: Updated

frontend/src/components/filters/Filters.js
petrican: we already have updateActiveFilter (main action that updates the active filter in the redux store with the newFiltersActive that we have set with the newFiltesActive value)

frontend/src/components/filters/FiltersFrequent.js
petrican: Good point. I'll consider it. Thanks!

frontend/src/actions/FiltersActions.js
```diff
@@ -1,8 +1,357 @@
...
+    });
+  }
+  return storeActiveFilters;
+}
+
+/**
+ * @method foundAmongActiveFilters
+ * @summary checks that the filterToAdd is found among the storeActiveFilters
+ */
+function foundAmongActiveFilters({ storeActiveFilters, filterToAdd }) {
```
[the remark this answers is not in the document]
petrican: updated the doc to avoid confusion

petrican:
@siemiatj pls re-check and see if everything is in place. I also added unit tests for the actions. 

I ran the automated e2e tests 
Screenshot 2020-10-06 at 16 20 42

and there's no diff between the master and this branch. I will merge this only after will go through all the testcases from the manual e2e list that we have.

frontend/src/__tests__/components/filters/Filters.test.js
```diff
@@ -269,6 +266,7 @@ describe('Filters tests', () => {
             visible: false,
           },
         },
+        filters: filtersStoreThree,
```
siemiatj, line 269: tree ? :)

petrican: No no. It's the no of the filtersStore snapshot. Used diff snapshots and named them One, Two, Three

frontend/src/components/filters/Filters.js
```diff
@@ -2,604 +2,133 @@ import counterpart from 'counterpart';
...
-      activeFilter,
-      activeFiltersCaptions,
-    } = this.state;
+    const widgetShown = filters ? filters.widgetShown : false;
+    const notValidFields = filters ? filters.notValidFields : false;
+
+    if (!filters || !viewId || !filters.filterData) return false;
+    const { filtersActive, filtersCaptions: activeFiltersCaptions } = filters;
+
+    const allFilters = annotateFilters({
```
siemiatj, line 125: Why can't this be handled by redux but instead we'll be calling this on each render and create a new js object ? Seems like we have everything we need in the `filters` prop.

frontend/src/components/filters/Filters.js
```diff
@@ -2,604 +2,133 @@ import counterpart from 'counterpart';
...
+      unannotatedFilters: filters.filterData,
+      filtersActive,
+    });
 
-    const allFilters = this.annotateFilters(
-      filterData.toIndexedSeq().toArray()
-    );
     const flatActiveFilterIds =
-      activeFilter !== null ? activeFilter.map((item) => item.filterId) : [];
+      filtersActive !== null ? filtersActive.map((item) => item.filterId) : [];
```
siemiatj, line 131: Same here - we'll be mapping over active filters on each render.

frontend/src/actions/FiltersActions.js
```diff
@@ -1,8 +1,355 @@
...
+    type: types.FILTER_UPDATE_WIDGET_SHOWN,
+    payload: { id, data },
+  };
+}
+
+/**
+ * @method clearStaticFilters
+ * @summary Clears the existing static filters for a filter branch in the redux store
+ */
+export function clearStaticFilters({ filterId, data }) {
```
siemiatj, line 83: Any particular reason why it's sometimes `id`, and other times `filterId` ? It's a minor thing but if we're already using named params let's stick to one convention.

frontend/src/components/filters/FiltersIncluded.js
```diff
@@ -202,26 +188,7 @@ class FiltersNotFrequent extends PureComponent {
   }
 }
 
-/**
- * @typedef {object} Props Component props
```
siemiatj, line 206: I think I've asked in another file, but why are we removing those ?

frontend/src/utils/documentListHelper.js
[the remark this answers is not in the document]
siemiatj: Only in case `filters[filterId]` exists. Otherwise we'll get undefined. I think it'd be best to have a cached selector here as filters might be a complicated object and hit us with unnecessary re-renders.

siemiatj (changes requested):

frontend/src/components/filters/FiltersIncluded.js
petrican: It's just because the format of the Proptypes is self-explanatory of what the param is and what type it has. I am for having these comments but only when is needed. In this case this would just add extra lines to the component that deliver the same information that exist in the Proptypes and that I found to be redundant.

frontend/src/actions/FiltersActions.js
petrican: Agree. Updated to filterId all over. Thanks

frontend/src/components/filters/Filters.js
petrican: Screenshot 2020-10-07 at 10 03 08
I have them mapped in the mapStateToProps but once I do that I get a DocList unit test failing :(

petrican: Thanks for the help in fixing this @siemiatj

siemiatj (approved):

PR #9884, opened 2020-09-03
Update canBeExportedFrom also on status-update via process

metas-ts:


metas-ts (commented):
code lgtm, but not yet tested

Issue #9904, opened 2020-09-06
Base Language Translation Missing


dragospodariu96:
### Is this a bug or feature request?
FR
### What is the current behavior?
When saving a product, an entry in the 'M_Product_Trl' table for the base language doesn't get created.
#### Which are the steps to reproduce?
Create a new product. Check the translations, the base language trl is missing
### What is the expected or desired behavior?
Also have an entry for the base language

PR #9905, opened 2020-09-06
#9904 Also saving a product trl in the base language.

dragospodariu96:
#9904

### DIFF
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/i18n/po/POTrlRepository.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/i18n/po/POTrlRepository.java
@@ -156,7 +156,7 @@ public class POTrlRepository
 		}
 		else
 		{
-			sql.append("AND (l." + I_AD_Language.COLUMNNAME_IsSystemLanguage + "='Y' AND l." + I_AD_Language.COLUMNNAME_IsBaseLanguage + "='N')");
+			sql.append("AND (l." + I_AD_Language.COLUMNNAME_IsSystemLanguage + "='Y' OR l." + I_AD_Language.COLUMNNAME_IsBaseLanguage + "='Y')");
 		}
 
 		sql.append(" AND t.")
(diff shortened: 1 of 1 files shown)


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/i18n/po/POTrlRepository.java
```diff
@@ -156,14 +153,14 @@ public final boolean insertTranslations(@NonNull final POTrlInfo trlInfo, final
 		}
```
metas-ts, line 153: Please check out the two differnt where-clauses: 
as talked yesterday, this whole if doesn't make any sense anymore.

metas-ts (approved):
as talked im MM yesterdady (you approved)
> ..and extend the help text for IsSystemLanguage to say that systemlanguage always means "trl be created"

metas-ts (changes requested):
see dismissed review https://github.com/metasfresh/metasfresh/pull/9905#pullrequestreview-483794522

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/i18n/po/POTrlRepository.java
dragospodariu96: There might be languages that are marked as isbaselanguage = 'y' and issystemlanguage='n' that we need trls for for the ad_element table. That's the scenario I left the if for. WDYT?

metas-ts: aaah, ok, i think now i got it.
then i think we should change this
```java
		if (I_AD_Element.Table_Name.equals(tableName))
		{
			sql.append("AND (l." + I_AD_Language.COLUMNNAME_IsSystemLanguage + "='Y' OR l." + I_AD_Language.COLUMNNAME_IsBaseLanguage + "='Y')");
		}
		else
		{
			sql.append("AND (l." + I_AD_Language.COLUMNNAME_IsSystemLanguage + "='Y')");
		}
```

to
```java
		sql.append("AND (l." + I_AD_Language.COLUMNNAME_IsSystemLanguage + "='Y'");
		if (I_AD_Element.Table_Name.equals(tableName))
		{ // for AD_Element we need to make sure to have the BaseLanguage, no matter if that's also a system-language
			sql.append(" OR l." + I_AD_Language.COLUMNNAME_IsBaseLanguage + "='Y'");
		}
		sql.append(")");
```

i think that way it's clearer

Issue #9908, opened 2020-09-07
OLCand - allow approval per selected OLCands


metas-ts:
### Is this a bug or feature request?
FR

PR #9909, opened 2020-09-07
split old table in base and concrete impl; add another concrete impl

metas-ts:
https://github.com/metasfresh/metasfresh/issues/9908


metas-ts (commented):
lgtm

Issue #7185, opened 2020-08-24
GL_Journal with PostingTypes "S" Static and "Y" Year End not working


KommunikativCh:
### Is this a bug or feature request?
BUG
### What is the current behavior?
Creating GL_Journal with PostingTypes "S" Static and "Y" Year End won't be creating bookings

#### Which are the steps to reproduce?
Create GL_Journal with PostingTypes "S" Static or "Y" Year End with Lines and change Status to Complete

### What is the expected or desired behavior?
This GL_Journal should create the bookings

PR #9911, opened 2020-09-08
7185 fix(gl_journal): PostingType S/Y

KommunikativCh:
#7185


metas-ts (approved):
@KommunikativCh sorry that it took so long

PR #9912, opened 2020-09-09
Add safety check for the case when view is undefined to prevent loop

petrican:


petrican:
https://www.loom.com/share/1b1ae552c69b42ad8d868ba90ca248b1

Issue #9865, opened 2020-08-31
Rewrite included views


siemiatj:

siemiatj:
Regression bug when opening an included view from an action in modal:

![Image Pasted at 2020-9-17 12-47](https://user-images.githubusercontent.com/513460/93474164-4830f500-f8f7-11ea-89af-f7fc6b136355.png)

PR #9915, opened 2020-09-09
Rewrite included views logic

siemiatj:
Related to #9865


frontend/src/actions/ViewActions.js
```diff
@@ -222,6 +223,28 @@ export function addViewLocationData(id, locationData, isModal) {
...
+ */
+export function updateViewSuccess(id, data, isModal) {
+  return {
+    type: UPDATE_VIEW_DATA_SUCCESS,
+    payload: { id, data, isModal },
+  };
+}
+
+/**
+ * @method toggleIncludedView
```
petrican, line 238: updateViewError

frontend/src/actions/ViewActions.js
```diff
@@ -222,6 +223,28 @@ export function addViewLocationData(id, locationData, isModal) {
   };
 }
 
+/**
+ * @method toggleIncludedView
+ * @summary success when updating view's properties
+ */
+export function updateViewSuccess(id, data, isModal) {
```
petrican, line 230: Can you please put params in an object updateViewSuccess({ id, data, isModal }) {

frontend/src/components/header/Actions.js
```diff
@@ -365,14 +365,14 @@ Actions.propTypes = {
 };
 
 const mapStateToProps = (state) => {
-  const includedView = state.listHandler.includedView;
+  const includedView = state.viewHandler.includedView;
```
petrican, line 368: const { includedView } = state.viewHandler;

frontend/test_setup/jestSetup.js
```diff
@@ -47,6 +47,9 @@ window.HTMLDivElement.prototype.getBoundingClientRect = function() {
 // process.on('unhandledRejection', (reason) => {
 //   console.log('REJECTION', reason)
 // })
+// process.on('uncaughtException', function(err) {
+//   console.log('EXCEPTION', err);
+// });
```
petrican, line 53: Is there a reason why you commented this ? This might ring a bell in case of valid uncaughtException situation within the components unit/functional testing.

petrican (changes requested):
Have some questions and small requests

frontend/test_setup/jestSetup.js
siemiatj: If there's an uncaught exception in the app, it will be thrown in the console. But in case of tests - not always. Also right now there are some minor problems (promises not finished before the test is done) that are polluting the log, but I don't have time to look into.

petrican: Opened https://github.com/metasfresh/mf15/issues/250 for that. Thanks

petrican:
Other than those small changes requested lgtm. Thank you!

Issue #7085, opened 2020-07-30
Extend DESADV for DELV vs ULCO


metas-ts:
### Is this a bug or feature request?
FRL

PR #9928, opened 2020-09-14
add InvoicableQtyBasedOn info to INVOIC and DESADV-(hdr); also:

metas-ts:
* CanonicalXSDGenerator does not output minOccurs and maxOccurs=1 because it creates a bunch of warnings

https://github.com/metasfresh/metasfresh/issues/7085


metas-ts (commented):
checked on dev-instance

Issue #9935, opened 2020-09-14
Assign deliveries and invoices to the project, automatically and manually


dragospodariu96:
### Is this a bug or feature request?
FR
### What is the current behavior?
After completing a sales order and generating shipments and invoices, the order's project doesn't get assigned to them
#### Which are the steps to reproduce?
Complete a sales order. Generate shipment(s). Generate Invoice(s). They don't have the project set..
### What is the expected or desired behavior?
That the generated shipments and invoices have the order's project set.

PR #9936, opened 2020-09-14
#9935

dragospodariu96:
#9935


metas-ts (commented):
ltgm, i'm just not sure if we should add project to the "general" shipment and invoice windows

dragospodariu96:
> ltgm, i'm just not sure if we should add project to the "general" shipment and invoice windows

@metas-ts I'll make custom windows for them then and move the script with the new window and then window adjustments to their repo. WDYT?

PR #9942, opened 2020-09-15
Fix qty input batch entry

petrican:


petrican:
https://www.loom.com/share/2608ad93c439478f88fe3190ca0c8c98

PR #9962, opened 2020-09-17
#9935 C_project_id propagation and window adjustments

dragospodariu96:
#9935

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_OrderLine.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_OrderLine.java
@@ -127,12 +127,12 @@ public class C_OrderLine
 	}
 
 	@ModelChange(timings = { ModelValidator.TYPE_BEFORE_NEW })
-	public void setProjectToOrderline(final I_C_OrderLine orderline)
+	public void setProjectToOrderline(final I_C_OrderLine orderLine)
 	{
-		final I_C_Order order = orderline.getC_Order();
-		if (order != null)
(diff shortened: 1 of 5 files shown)


backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_OrderLine.java
```diff
@@ -126,6 +126,16 @@ public void setQtyEnteredInPriceUOM(final I_C_OrderLine orderLine)
...
 	}
 
+	@ModelChange(timings = { ModelValidator.TYPE_BEFORE_NEW })
+	public void setProjectToOrderline(final I_C_OrderLine orderline)
+	{
+		final I_C_Order order = orderline.getC_Order();
+		if (order != null)
+		{
+			orderline.setC_Project_ID(order.getC_Project_ID());
+		}
```
metas-ts, line 136:
```suggestion
		if (orderline.getC_Order_ID() > 0 && orderLine.getC_Project_ID() <= 0)
		{
			final I_C_Order order = orderline.getC_Order();
			orderline.setC_Project_ID(order.getC_Project_ID());
		}
```

improvements:
* don't overwrite a rpoject that a used just set in the add-new modal
* don't load the order if you don't have to

metas-ts (changes requested):

backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_OrderLine.java
dragospodariu96: applied changes.

metas-ts (approved):

PR #9965, opened 2020-09-17
Fix the issues found on QA testing

petrican:
https://www.loom.com/share/e6e1a78abeed4dafa7e0c1f20ee0a849

Target branch master

### DIFF
--- a/frontend/src/components/table/TableContextMenu.js
+++ b/frontend/src/components/table/TableContextMenu.js
@@ -120,10 +120,19 @@ class TableContextMenu extends Component {
       },
 
       onComplete: () => {
-        const { references } = this.state;
+        let offset;
+        const { references, contextMenu } = this.state;
+        const { y: initialY } = contextMenu;
+        const mainPanel = document.querySelector('.panel-vertical-scroll');
+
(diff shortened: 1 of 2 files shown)


frontend/src/components/table/TableContextMenu.js
```diff
@@ -120,22 +124,45 @@ class TableContextMenu extends Component {
       },
 
       onComplete: () => {
-        const { references } = this.state;
+        let offset;
+        const { references, contextMenu } = this.state;
+        const { y: initialY } = contextMenu;
+        const mainPanel = document.querySelector('.panel-vertical-scroll');
```
siemiatj, line 130: I'm not sure how I feel about mixing in vanilla js here instead of using ReactDOM. Isn't it against some react best practices ?

siemiatj (approved):

frontend/src/components/table/TableContextMenu.js
petrican: I don't think that using the ref would help in here as you don't have any fnct exposed that would allow you.to close the calendars.

Issue #9969, opened 2020-09-18
Order from offer (construction) reference to offer


dragospodariu96:
### Is this a bug or feature request?
FR
### What is the current behavior?
Quotation doesn't have sidelist references to generated sales order
Generated sales orders from quotations don't have sidelist references to the quotation they came from
#### Which are the steps to reproduce?
Generate a sales order from a quotation
### What is the expected or desired behavior?
That sidelist references between them work

PR #9970, opened 2020-09-18
Order from offer (construction) reference to offer

dragospodariu96:
#9969


metas-ts (approved):
i'm not entirely sure metasfresh will figure out which window to open, but let's see

dragospodariu96:
> i'm not entirely sure metasfresh will figure out which window to open, but let's see

@metas-ts I also thought it wouldn't know which window to open but whlie testing it it worked. However I can use RecordZoomWindowFinder.findAdWindowId(I_C_Order.Table_Name); to make sure it's always the right window. (the c_order table has the sales order's window set).

PR #9984, opened 2020-09-22
Store the data in the redux store instead of local state

petrican:
-creates a separate branch in the redux store within the tables one for the headers. This is needed because if I were to use the corresponding table branch ID that would have been overwritten on view reload. This way I make sure it remains and is accessible across browsing(back, fw)

https://www.loom.com/share/a4e27010c4cf46e49d5ff6b5585b3795


frontend/src/components/table/TableHeader.js
```diff
@@ -145,4 +145,23 @@ TableHeader.propTypes = {
...
+      state.tables && state.tables.headers ? state.tables.headers[tableId] : {},
+  };
+};
+
+export default connect(
+  mapStateToProps,
+  {
+    setActiveSortFields,
+  }
+)(TableHeader);
```
siemiatj, line 167: Why not add a named export of a component disconnected from the store ? This way you won't have to do the whole shenanigans in the tests and use what we've had instead. Then the reducers/actions can be tested separately. We know, that `mapStateToProps` works as redux has a pretty good test coverage.

frontend/src/components/table/TableHeader.js
```diff
@@ -1,27 +1,27 @@
...
 import PropTypes from 'prop-types';
-
+import { setActiveSortFields } from '../../actions/TableActions';
 import { shouldRenderColumn, getSizeClass } from '../../utils/tableHelpers';
 import { getTableId } from '../../reducers/tables';
 
-export default class TableHeader extends PureComponent {
+class TableHeader extends PureComponent {
   UNSAFE_componentWillMount() {
     this.setInitialState();
```
siemiatj, line 11: I'm strongly against this. Since we have everything we need in the redux store already, why not use it when instantiating the table instead ? On `FETCH_DOCUMENT_SUCCESS` we get the `orderBy` value, and immediately update the table. So we can move this logic there and this whole lifecycle handler won't be needed anymore.

frontend/src/reducers/tables.js
```diff
@@ -326,6 +326,13 @@ const reducer = produce((draftState, action) => {
       const { id, active } = action.payload;
 
       draftState[id].activeSort = active;
+      return;
+    }
+    case types.SET_SORT_FIELDS: {
+      const { id, fields } = action.payload;
+      draftState.headers = {};
```
siemiatj, line 333: Shouldn't this be in the `tables[tableId].headers` ? Or even with a more meaningful name like `sortOptions` or something ?

siemiatj (changes requested):

frontend/src/components/table/TableHeader.js
petrican: Not sure what you mean in here. Don't we want the components to be independent of each other ?

petrican: I believe you were thinking on importing directly the component disconnected from the file. Done that and replaced the test with old one. With some small adapts

siemiatj (approved):

PR #9995, opened 2020-09-23
Automatically mark preselection for first entry in list

petrican:
https://www.loom.com/share/c5ce719dab3545a0973d358661a02edd

Notes: The functionality within the MenuOverlay component is done with DOM selectors. As much as I wanted to change that I cannot as this is beyond the scope or the budget allocated for the task. Also used the DOM selectors to comply with that code where is the case. @siemiatj

### DIFF
--- a/frontend/src/components/header/MenuOverlay.js
+++ b/frontend/src/components/header/MenuOverlay.js
@@ -343,6 +343,32 @@ class MenuOverlay extends Component {
   };
 
   /**
+   * @method rafAsync
+   * @summary - uses requestAnimationFrame method which tells the browser that you wish to perform an animation and requests
+   *            that the browser calls a specified function to update an animation before the next repaint
+   */
+  rafAsync = () => {
+    return new Promise((resolve) => {
+      requestAnimationFrame(resolve);
+    });
+  };
+
+  /**
+   * @method checkElement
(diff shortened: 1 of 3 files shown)


frontend/src/components/header/MenuOverlay.js
```diff
@@ -548,19 +582,24 @@ class MenuOverlay extends Component {
                 <div className="input-flex input-primary">
                   <i className="input-icon meta-icon-preview" />
 
-                  <DebounceInput
-                    debounceTimeout={250}
+                  <input
                     type="text"
-                    inputRef={(ref) => {
+                    ref={(ref) => {
```
siemiatj, line 587: Can we move this to a `setRef` function, so that we won't have an anonymous function here for no reason ?

siemiatj (approved):
Check my comment after dev test here https://github.com/metasfresh/me03/issues/5644#issuecomment-697328130

frontend/src/components/header/MenuOverlay.js
petrican: Done

petrican:
# 🚀🛠 Deployed to https://dev1308.metasfresh.com

- [x] checked that everything is up
- (version: 5.154.2_8_9981autoMarkPreselection)

PR #10007, opened 2020-09-24
Fixes for Dashboard breadcrumbs path

petrican:
https://www.loom.com/share/baa4cf8b760141feb8f224f6e98d4fe5


petrican:
I believe this a more appropriate fix that also fixes that issue when you come back from a page via sitemap

siemiatj (approved):

Issue #9980, opened 2020-09-21
Order indicator vanished after switching detail view back to grid view


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
Sorting indicator lost when come back from inner doc.
#### Which are the steps to reproduce?
If you have an active sorting and you go to one of the items and come back using back button you don't see the chevron showing sorting direction
### What is the expected or desired behavior?
Chevron icon should be present when using back button if you have an active sort

siemiatj:
@petrican I think I've found a much simpler solution : https://recordit.co/v1RgnYq2lK . Check my PR

PR #10008, opened 2020-09-24
Load orderBy on details view

petrican:
Related to #9980


petrican:
closing this one as I merged that solution with the local state

PR #10014, opened 2020-09-24
Alternative fix for ordering tables

siemiatj:
Related to #9980

### DIFF
--- a/frontend/src/__tests__/components/table/Table.test.js
+++ b/frontend/src/__tests__/components/table/Table.test.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { shallow, mount, render } from 'enzyme';
+import { shallow, mount } from 'enzyme';
 import { Provider } from 'react-redux';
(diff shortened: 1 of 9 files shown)


petrican:
So you are mapping the orderBy from the props and get the fields for the local state. That is indeed simple and it works. I still believe that the redux solution is more appropriate on a long term but implies having that extra data from the BE response whenever order changes. If user is going to hit back and then fw button it will not have the previous order on FW button click. If that was taken from the redux store it will always remain as it was set to. Going to merge this as it is works and doesn't break any existing functionality and will come back to make it with redux once we have the BE response as we need. Thanks for your time and effort to find solution for this @siemiatj !

siemiatj:
Agree. It'd be best to just have `orderBy` in the tables reducer, and remove it elsewhere (as well as the `defaultOrderBys` that's only causing confusion, `SORT_TAB`, and all the current code we use for sorting ).

Issue #10041, opened 2020-09-30
Don't crash the app when widget exists in layout but not data


siemiatj:
### Is this a bug or feature request?

Bug

PR #10042, opened 2020-09-30
Stop crashing when field data is missing

siemiatj:
Related to #10041


petrican (approved):
Works as expected. Thanks!

PR #10055, opened 2020-10-01
make sure that just one thread at a time can create new bpartner-stats

metas-ts:
..and only on app.
to avoid the error "ERROR: duplicate key value violates unique constraint "c_bpartner_product_stats_uq""


metas-ts (commented):
tsted locally

Issue #7126, opened 2020-08-13
Refactor widget components 


petrican:
### Is this a bug or feature request?
F
### What is the current behavior?

Widget components code is messy and causes unnecessary re-renderings

#### Which are the steps to reproduce?

Load the app and monitor re-renderings with he profiler

### What is the expected or desired behaviour?

There should be way less re-renderings and code should be written in a nicely structured way

PR #10056, opened 2020-10-01
Refactor Raw/MasterWidget

siemiatj:
Related to #7126


frontend/src/components/widget/RawWidget.js
```diff
@@ -352,14 +360,27 @@ export class RawWidget extends Component {
...
+    this.rawWidget.type = type;
+  };
+
   /**
-   * @method handleErrorPopup
-   * @summary ToDo: Describe the method.
-   * @param {*} value
+   * @method showErrorPopup
+   * @summary shows error message on mouse over
+   */
```
petrican, line 370: can't this be retained in the redux store ?

frontend/src/components/widget/WidgetRenderer.js
```diff
@@ -0,0 +1,629 @@
...
+      handleBackdropLock,
+      inputProps: {
+        placeholder: fields[0].emptyText,
+        disabled: readonly,
+        tabIndex: tabIndex,
+      },
+      onChange: (date) => handleChange(widgetField, date),
+    };
+    const dateRangeProps = {
+      mandatory: widgetData[0].mandatory,
```
petrican, line 149: would get those values before {mandatory,  validStatus } = widgetData[0];  also would add some safety checks

frontend/src/components/widget/WidgetRenderer.js
```diff
@@ -0,0 +1,629 @@
...
+                {...{
+                  isOpenDatePicker,
+                  timeZone,
+                }}
+              />
+            </div>
+          );
+        }
+      case 'ZonedDateTime':
+        return (
```
petrican, line 220: Would be nice to separate here to have <ZoneDateTime {...props} /> and have ZoneDateTime defined in a separate component

frontend/src/components/widget/WidgetRenderer.js
```diff
@@ -0,0 +1,629 @@
...
+                  null,
+                  true
+                )
+              }
+              tabIndex={tabIndex}
+            />
+          </div>
+        );
+      case 'Timestamp':
+        return (
```
petrican, line 266: Here also and for others like as well. Would be nice to separate here to have <Timestamp {...props} /> and have TimeStamp defined in a separate component

frontend/src/components/widget/WidgetRenderer.js
```diff
@@ -0,0 +1,629 @@
...
+              id,
+              filterWidget,
+            }}
+            handlePatch={onPatch}
+          />
+        );
+      case 'Switch':
+        return (
+          <label
+            className={classnames('input-switch', {
```
petrican, line 467: Basically if we reduce the size of the file and have the logic in separate granular components like in here <Switch {...props} /> I believe logic would be easy to follow and any further code changes would be smoother.

frontend/src/utils/widgetHelpers.js
```diff
@@ -0,0 +1,144 @@
...
+  }
+}
+
+/**
+ * @method formatValueByWidgetType
+ * @summary Performs patching at MasterWidget level, shaping in the same time the `value` for various cases
+ * @param {string} widgetType
+ * @param {string|undefined} value
+ */
+export function formatValueByWidgetType({ widgetType, value }) {
```
petrican, line 103: I believe all these should have unit tests. Some might already have.

petrican (changes requested):
It looks fine. Pls see my comments. I haven't done a rough analysis on performance (before/after). Pls also if you can do some comparisons to be able to also see the gain in rendering performance. Thanks

frontend/src/components/widget/Lookup/Lookup.js
```diff
@@ -300,7 +302,7 @@ class Lookup extends Component {
       rowId,
       tabIndex,
       validStatus,
-      recent,
+      // recent,
```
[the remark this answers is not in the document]
siemiatj: Because it was only used to pass an empty array. Don't ask me why: https://github.com/metasfresh/metasfresh/pull/10056/files#diff-92a02f4e1c97e8c29d0c3efc69b4917fL695

frontend/src/components/widget/RawWidget.js
siemiatj: of what ? We decided not to have widgets in the store as separate instances so we don't have a place to keep this.

frontend/src/components/widget/WidgetRenderer.js
siemiatj: `WidgetWrapper` returns `widgetData` as `[{}]` so this will just be falsy. One one hand having checks makes sense, but in our case we'd have to add them everywhere, where we're checking directly a property in widgetData object.

frontend/src/components/widget/WidgetRenderer.js
siemiatj: Well yes and no. It would be fairly straightforward if all the components would take the same props. But they don't. So you either have to pass all the props, that `WidgetRenderer` is getting, or build props object for each `widgetType`, or just define props in the element definition (so pretty much the same thing we have right now). So only in first case this would make the code shorter, but then every widget would get bazillion unnecessary properties that it won't use in any way. In the other two cases, we would only replace `<label>whatever</label>` with a component name.

siemiatj:
For some reason this sucked in changes outside of the frontend branch while rebasing. Closing in favor of https://github.com/metasfresh/metasfresh/pull/10134

Issue #10063, opened 2020-10-06
No longer show discontinued products in batch entry


dragospodariu96:
### Is this a bug or feature request?
FR
### What is the current behavior?
Discontinued products are shown in batch entry in orderlines
#### Which are the steps to reproduce?
Try to add a product via batch entry
### What is the expected or desired behavior?
Discontinued products to no longer be shown in batch entry

PR #10064, opened 2020-10-06
#10063

dragospodariu96:
#10063


backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/window/descriptor/sql/ProductLookupDescriptor.java
```diff
@@ -357,6 +356,11 @@ private static StringBuilder appendFilterByIsActive(final StringBuilder sqlWhere
 		return sqlWhereClause.append("\n p.").append(I_M_Product_Lookup_V.COLUMNNAME_IsActive).append("=").append(sqlWhereClauseParams.placeholder(true));
 	}
 
+	private static StringBuilder appendFilterByDiscontinued(final StringBuilder sqlWhereClause, final SqlParamsCollector sqlWhereClauseParams)
```
metas-ts, line 359: From looking at the code, i'm not sure that we can still enter discontinued products at not-order-places.
From a quick check, those include invoice lines and forecast lines; maybe more.

Please make sure that this still works

metas-ts (changes requested):

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/window/descriptor/sql/ProductLookupDescriptor.java
dragospodariu96: Modified ProductLookupDescriptor so to only show discontinued products when the new boolean param is set to false.

metas-ts (approved):

PR #10069, opened 2020-10-06
attribute matching fix

pvpurcarcosmin:
refs: https://github.com/metasfresh/metasfresh/issues/10068

### DIFF
--- a/backend/de.metas.material/dispo-commons/src/main/java/de/metas/material/dispo/commons/repository/atp/AvailableToPromiseResultBuilder.java
+++ b/backend/de.metas.material/dispo-commons/src/main/java/de/metas/material/dispo/commons/repository/atp/AvailableToPromiseResultBuilder.java
@@ -1,19 +1,17 @@
 package de.metas.material.dispo.commons.repository.atp;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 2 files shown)


backend/de.metas.material/dispo-commons/src/main/java/de/metas/material/dispo/commons/repository/atp/AvailableToPromiseResultBuilder.java
```diff
@@ -105,24 +103,7 @@ public AvailableToPromiseResult build()
...
-		if (!addedToAtLeastOneGroup)
-		{
-			throw new AdempiereException("No matching group found for AddToResultGroupRequest")
-					.appendParametersToMessage()
-					.setParameter("request", request)
-					.setParameter("buckets", buckets)
-					.setParameter("this", this);
-		}
-
+		buckets.forEach(bucket -> bucket.addQtyToAllMatchingGroups(request));
```
metas-ts, line 106:
```suggestion
		// note that we might select more quantities than we actually wanted (bc of the way we match attributes in the query using LIKE)
		// for that reason, we need to be lenient in case not all quantities can be added to a bucked
		buckets.forEach(bucket -> bucket.addQtyToAllMatchingGroups(request));
```

metas-ts (approved):

Issue #10079, opened 2020-10-07
Add window 'Counter Document' and 'Missing/Incomplete Counter Document' to WebUI role


dragospodariu96:
### Is this a bug or feature request?
FR
### What is the current behavior?
The 2 windows are added to the webui role, however they are not added to webui menu.
#### Which are the steps to reproduce?
Look for them in the webui menu, can't be found only if hardcoding window id
### What is the expected or desired behavior?
That they appear when searching for a window under the 'system' category

PR #10080, opened 2020-10-07
#10079

dragospodariu96:
#10079


backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5569920_sys_gh10079ModifyRV_Missing_Counter_Documents.sql
```diff
@@ -0,0 +1,133 @@
...
+-- This view is made for C_Order entries and their counter-documents
+-- If we start supporting other kind of counter documents in the fututre, this view will have to be updated with the logic on those tables, using unions
+
+CREATE VIEW RV_Missing_Counter_Documents
+AS
+
+
+SELECT
+    -- In the future, if we will add support for other documents (i.e: invoices), we will have to update the id column's logic using a case when logic.
+    -- https://github.com/metasfresh/mf15/issues/318
```
metas-ts, line 15: no urls of private repos pls

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5569920_sys_gh10079ModifyRV_Missing_Counter_Documents.sql
```diff
@@ -0,0 +1,133 @@
...
+    CASE
+        WHEN doc.Record_ID IS NOT NULL THEN (doc.Record_ID * 10) + 1
```
metas-ts, line 17:
```suggestion
        WHEN doc.C_OrderLine_ID IS NOT NULL THEN (doc.Record_ID * 10) + 1
```
i wrote
> explicitly return the C_OrderLine_ID which has the missing counter doc from the doc subselect, together with the Record_ID

exactly for the purpose of using it in `RV_Missing_Counter_Document_ID`

if we later add e.g. C_Invoice_Lines with missing-counter-docs then it will also be aliased to `Record_ID`, so using `doc.Record_ID` does not make sense

metas-ts (changes requested):

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5569920_sys_gh10079ModifyRV_Missing_Counter_Documents.sql
dragospodariu96: I thought the invoicelines would be handled in a different left join after looking at the sql and the doc one would only remain for orderlines., but we could also expand the view by joining with the invoice and invoiceline tables on the 'doc' select. Implemented the changes! 😓

metas-ts (approved):

PR #10090, opened 2020-10-09
Have a swing at fixing failing e2e tests

siemiatj:
Related to #10082.
- One of the problems here is that when we clear the filters, page is refreshed but cypress is so fast, it starts clicking elements before the reload finishes. Because of that it can't find some elements.
- A separate problem is that the UOM field is filled when the product is selected. So with the current architecture we won't do the request when nothing changed in the widget. And in this case prefilled value and the one we were typing are the same. I've tried adding a check for the current input value in the lookup helpers but looks like it's not gonna be that easy. 
- Also looks like picking from the picking terminal now doesn't open a modal, but a new window instead. So a lot of actions were tripping over that.


petrican (approved):
Lgtm! Thanks

PR #10107, opened 2020-10-12
#285

dragospodariu96:
https://github.com/metasfresh/mf15/issues/285

### DIFF
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/ProcessExecutionResult.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/ProcessExecutionResult.java
@@ -429,6 +429,7 @@ public class ProcessExecutionResult
 					.records(records)
 					.adWindowId(adWindowId)
 					.target(OpenTarget.GridView)
+					.targetTab(RecordsToOpen.TargetTab.SAME_TAB_OVERLAY)
 					.automaticallySetReferencingDocumentPaths(true)
 					.build());
 		}
@@ -449,6 +450,7 @@ public class ProcessExecutionResult
 					.records(records)
 					.adWindowId(adWindowId)
 					.target(OpenTarget.GridView)
+					.targetTab(RecordsToOpen.TargetTab.SAME_TAB_OVERLAY)
 					.automaticallySetReferencingDocumentPaths(true)
 					.build());
 		}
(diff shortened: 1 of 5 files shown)


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/ProcessExecutionResult.java
```diff
@@ -885,7 +884,7 @@ private RecordsToOpen(
 			this.records = ImmutableList.copyOf(records);
 			this.windowIdString = StringUtils.trimBlankToNull(adWindowId);
 			this.target = target != null ? target : OpenTarget.GridView;
-			this.targetTab = targetTab != null ? targetTab : TargetTab.SAME_TAB;
+			this.targetTab = targetTab != null ? targetTab : target == OpenTarget.GridView ? TargetTab.SAME_TAB_OVERLAY : TargetTab.SAME_TAB;
```
teosarca, line 887: avoid having logic here.
pls decide what kind of targetTab we need on the caller method,.

TheBestPessimist:
@dragospodariu96 please 🍒 this PR to `intensive_care_uat` as well

TheBestPessimist:
Another fix needed here:

In `de.metas.ui.web.process.adprocess.ADProcessPostProcessService#createResultAction` records to open might be null `final RecordsToOpen recordsToOpen = processExecutionResult.getRecordsToOpen();`.

This happens for example when running process `AllocatePayments`

image

TheBestPessimist (changes requested):
please ignore this.

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/ProcessExecutionResult.java
dragospodariu96: @teosarca then I would leave it as default in the builder to SAME_TAB unless set different and make sure whenever opentarget.gridview is used the targetTab gets passed as targettab.same_tab_overlay. WDYT?

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/ProcessExecutionResult.java
```diff
@@ -493,7 +496,6 @@ public void setRecordToOpen(@Nullable final TableRecordReference record, final @
 					.record(record)
 					.adWindowId(adWindowId)
 					.target(target)
-					.targetTab(RecordsToOpen.TargetTab.SAME_TAB)
```
teosarca, line 496: pls set the target tab too

PR #10112, opened 2020-10-13
ShipmentScheduleWithHUService: make sure that new Hus have our trxName

metas-ts:
HUAndItemsDAO don't assume it might have to be run in its own trx; caller needs to make sure the HU has the right trx


metas-ts (commented):
lgtm

PR #10117, opened 2020-10-15
10116 two entries dbl clk 

petrican:
https://www.loom.com/share/134c07dc3c954d948973bbd52a666c04


siemiatj (approved):
Please add a test in `WindowActions.test.js`

Issue #10125, opened 2020-10-16
Fix editing fields in data entry tabs


siemiatj:
### Is this a bug or feature request?

Bug

### What is the current behavior?

After editing the field, only the edited field is updated so other fields that might've updated in the process are skipped. Handling websocket events is also borked.

PR #10126, opened 2020-10-16
Fix handing/reacting to changes in data entry tabs

siemiatj:
Related to #10125


frontend/src/actions/WindowActions.js
```diff
@@ -229,6 +232,17 @@ export function unselectTab(scope) {
   };
 }
 
+/**
+ * @method setUpdatedTabLayout
+ * @summary Action creator to update tab's layout data in the store
+ */
+function setUpdatedTabLayout(tabId, layoutData) {
```
petrican, line 239: I would put the params in an object ({ tabId, layoutData })

frontend/src/actions/WindowActions.js
```diff
@@ -431,6 +443,24 @@ export function fetchTab({ tabId, windowId, docId, query }) {
   };
 }
 
+/*
+ * @method updateTabLayout
+ * @summary Action creator for fetching and updating single tab's layout
+ */
```
petrican, line 449: Here too doc param and pass them as object params

frontend/src/containers/MasterWindow.js
```diff
@@ -177,20 +181,20 @@ class MasterWindowContainer extends PureComponent {
       tabId: activeTabId,
     });
 
-    const tabLayout =
-      master.layout.tabs &&
-      master.layout.tabs.filter((tab) => tab.tabId === activeTabId)[0];
-
-    const orderBy = tabLayout.orderBy;
+    const tabLayout = getTab(master.layout, activeTabId);
+    const orderBy = tabLayout ? tabLayout.orderBy : null;
```
petrican, line 185: what happens if orderBy is undefined ?

petrican (changes requested):
Some small changes requested. Other than that it is fine. Thanks!

frontend/src/containers/MasterWindow.js
siemiatj: Nothing, we won't set the sorting order when requesting data.

PR #10138, opened 2020-10-21
#10103 refresh data entries

metas-rc:
Related to https://github.com/metasfresh/metasfresh/issues/10103

### DIFF
--- a/backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/dataentry/window/descriptor/factory/DataEntrySubTabBindingRepository.java
+++ b/backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/dataentry/window/descriptor/factory/DataEntrySubTabBindingRepository.java
@@ -228,7 +228,7 @@ public class DataEntrySubTabBindingRepository implements DocumentsRepository
 		final DocumentValuesSupplier documentValuesSupplier = new DataEntryDocumentValuesSupplier(
 				dataEntryRecord.get(),
 				dataEntryWebuiTools);
-		document.refreshFromSupplier(documentValuesSupplier);
+		document.refreshDataEntryFromSupplier(documentValuesSupplier);
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/window/model/Document.java
```diff
@@ -557,6 +558,10 @@ else if (FieldInitializationMode.Refresh == mode)
 				changesCollector.collectValueIfChanged(documentField, valueOld, REASON_Value_Refreshing);
 			}
 		}
+		else if (FieldInitializationMode.RefreshDataEntries == mode)
```
teosarca, line 561: what's the difference between Refresh and RefreshDataEntries here?

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/dataentry/window/descriptor/factory/DataEntrySubTabBindingRepository.java
```diff
@@ -228,7 +228,7 @@ private void refreshFromDataEntryRecord(
 		final DocumentValuesSupplier documentValuesSupplier = new DataEntryDocumentValuesSupplier(
 				dataEntryRecord.get(),
 				dataEntryWebuiTools);
-		document.refreshFromSupplier(documentValuesSupplier);
+		document.refreshDataEntryFromSupplier(documentValuesSupplier);
```
teosarca, line 231: why simply calling the refresh is not ok?
what's the difference?

metas-rc: * The refreshFromSupplier is calling ```de.metas.ui.web.window.model.DocumentChangesCollector.collectValueIfChanged(IDocumentFieldView, Object, ReasonSupplier)``` which only collects values if values were changed.
* I introduced the new option which instead calls ```de.metas.ui.web.window.model.DocumentChangesCollector.collectValueChanged(IDocumentFieldView, ReasonSupplier) ```. This one collects the values no matter what.

image

PR #10141, opened 2020-10-22
DraftInventoryLinesCreator: don't fail if inventory already contains HUs

metas-ts:
..just log; and don't attempt to add them again

also
* AdempiereTestHelper: add method to create org with specific time zone
* InventoryImportProcess: minor improvements


metas-ts (commented):
lgtm

PR #10148, opened 2020-10-22
Gh10127 export all shipment schedules for order

metas-rc:
https://github.com/metasfresh/metasfresh/issues/10127

### DIFF
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/ShipmentScheduleRepository.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/ShipmentScheduleRepository.java
@@ -93,9 +93,23 @@ public class ShipmentScheduleRepository
 		{
 			queryBuilder.addEqualsFilter(COLUMNNAME_ExportStatus, query.getExportStatus().getCode());
 		}
-		if (query.getCanBeExportedFrom() != null)
+
+		final Instant canBeExportedFrom = query.getCanBeExportedFrom();
+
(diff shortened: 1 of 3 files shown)


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/ShipmentScheduleRepository.java
```diff
@@ -227,5 +273,14 @@ private void save(@NonNull final ShipmentSchedule shipmentSchedule)
...
 		@Builder.Default
 		boolean includeProcessed = false;
+
+		@Builder.Default
+		boolean fromCompleteOrderOrNullOrder = false;
+
+		@Builder.Default
+		boolean orderByOrderId = false;
+		@Builder.Default
+		boolean noFutureExportOfOrder = false;
```
metas-ts, line 283: i could understand `noFutureExportOfOrder` only after looking at the code in `ShipmentScheduleRepository`..
pls add a javadoc-comment to it, oooor think of a better name...like..maybe sth like `onlyIfAllFromOrderExportable`?

metas-ts (changes requested):
just minor requests 😊

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/ShipmentScheduleRepository.java
metas-rc: Why not both? :)
Done

metas-ts (approved):

Issue #10143, opened 2020-10-22
Status changes for DESADV


TheBestPessimist:
### Is this a bug or feature request?
FR

PR #10152, opened 2020-10-23
Status changes for DESADV

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/10143

### DIFF
--- a/backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/edi_desadv/ChangeEDI_ExportStatusHelper.java
+++ b/backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/edi_desadv/ChangeEDI_ExportStatusHelper.java
@@ -27,6 +27,7 @@ import de.metas.edi.api.EDIDesadvId;
 import de.metas.edi.api.EDIDocOutBoundLogService;
 import de.metas.edi.api.EDIExportStatus;
 import de.metas.edi.api.IDesadvDAO;
+import de.metas.edi.model.DocOutboundLogId;
 import de.metas.edi.model.I_C_Doc_Outbound_Log;
 import de.metas.edi.model.I_C_Invoice;
 import de.metas.esb.edi.model.I_EDI_Desadv;
@@ -40,7 +41,6 @@ import de.metas.util.Services;
 import lombok.NonNull;
 import lombok.experimental.UtilityClass;
 import org.adempiere.ad.service.IADReferenceDAO;
-import org.adempiere.archive.ArchiveId;
 import org.adempiere.exceptions.AdempiereException;
 import org.adempiere.util.lang.impl.TableRecordReference;
 import org.compiere.SpringContextHolder;
@@ -93,7 +93,7 @@ public class ChangeEDI_ExportStatusHelper
 		desadvDAO.save(edi);
 	}
 
-	public void C_DocOutbound_LogDoIt(final EDIExportStatus targetExportStatus, final ArchiveId logId)
+	public void C_DocOutbound_LogDoIt(final EDIExportStatus targetExportStatus, final DocOutboundLogId logId)
 	{
 		// technical detail: the I_C_Doc_Outbound_Log is updated when we update the C_Invoice, via interceptor: de.metas.edi.model.validator.C_Invoice.updateDocOutBoundLog
 		final de.metas.edi.model.I_C_Doc_Outbound_Log docOutboundLog = ediDocOutBoundLogService.retreiveById(logId);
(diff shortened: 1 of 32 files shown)


backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/edi_desadv/ChangeEDI_ExportStatus_C_Doc_Outbound_Log_GridView.java
```diff
@@ -0,0 +1,140 @@
...
+
+		return ChangeEDI_ExportStatusHelper.computeParameterDefaultValue(fromExportStatus);
+	}
+
+	@Override
+	protected String doIt() throws Exception
+	{
+		final EDIExportStatus targetExportStatus = EDIExportStatus.ofCode(p_TargetExportStatus);
+
+		for (final ArchiveId logId : getSelectedDocOutboundLogIds())
```
metas-ts, line 127: ArchiveId shall not be used for C_Doc_OutBound_Log_ID.
(note that we have a dedicated `AD_Archive` table that is also *referenced* by `C_Doc_Outbound_Log_Line`)

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/edi_desadv/ChangeEDI_ExportStatusHelper.java
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	/**
+	 * TODO tbp: how to check if this docOutboundLog is for edi or not?
+	 * docOutboundLog.isEdiEnabled() is a virtual column!
+	 * in swing this is checked by """(select bp.IsEdiInvoicRecipient from C_BPartner bp where bp.C_BPartner_ID=C_Doc_Outbound_Log.C_BPartner_ID)"""
+	 * do i need to check for if the AD_Table is C_Invoice as well?
+	 * help maybe?
+	 */
+	public boolean checkIsInvoiceAndEDI(final I_C_Doc_Outbound_Log docOutboundLog)
```
metas-ts, line 146: in theory you could check for C_Invoice.IsEdiEnabled..but that's an extra load..
or you could check for edi-export-status to be not null..but you already do that in the process anyways..

metas-ts (changes requested):

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/edi_desadv/ChangeEDI_ExportStatus_C_Doc_Outbound_Log_GridView.java
TheBestPessimist: > ArchiveId shall not be used for C_Doc_OutBound_Log_ID.

So should i add the ID class `DocOutboundLogId`?

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/edi_desadv/ChangeEDI_ExportStatusHelper.java
TheBestPessimist: Checking for edi-export-status to be not null is not enough, because the `C_Doc_Outbound_Log` might also reference `Sales Order`s or `Delivery Note`s, and we don't want to change their IDs afaiu.

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/edi_desadv/ChangeEDI_ExportStatus_C_Doc_Outbound_Log_GridView.java
metas-ts: pls add `DocOutboundLogId`

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/edi_desadv/ChangeEDI_ExportStatusHelper.java
metas-ts: > or you could check for edi-export-status to be not null

...in addition to the check for `C_Invoice` which you do already

TheBestPessimist: done

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/edi_desadv/ChangeEDI_ExportStatus_C_Doc_Outbound_Log_GridView.java
TheBestPessimist: done

metas-ts (approved):

Issue #10201, opened 2020-11-10
Error when setting attributes in orderline


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
FE doesn't call correctly the BE endpoints during attribute set
#### Which are the steps to reproduce?
New SO try to set attributes
### What is the expected or desired behavior?
FE doesn't call correctly the BE endpoints during attribute set

PR #10202, opened 2020-11-10
Prevent dup PATCH request within attributes

petrican:
https://www.loom.com/share/1b71ee0380264bf5b23d911091db30b3

https://github.com/metasfresh/metasfresh/issues/10201

### DIFF
--- a/frontend/src/components/widget/Attributes/Attributes.js
+++ b/frontend/src/components/widget/Attributes/Attributes.js
@@ -230,7 +230,8 @@ export default class Attributes extends Component {
               }),
               () => {
                 cb && cb();
-                onBlur && onBlur(prop, value, id);
+                // onBlur && onBlur(prop, value, id);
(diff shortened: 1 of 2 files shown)


frontend/src/components/widget/Attributes/Attributes.js
```diff
@@ -203,7 +203,7 @@ export default class Attributes extends Component {
    * @todo Write the documentation
    */
   handlePatch = (prop, value, id, cb) => {
-    const { attributeType, onBlur } = this.props;
+    const { attributeType } = this.props;
```
siemiatj, line 206: You can also remove it from `WidgetRenderer`

siemiatj (approved):

frontend/src/components/widget/Attributes/Attributes.js
petrican: Done. Thanks for the suggestion

PR #10218, opened 2020-11-12
Browser history on applying filters

petrican:
https://www.loom.com/share/0ce8d2b00c75468ca38f0cad54b8d919

### DIFF
--- a/frontend/src/containers/DocumentList.js
+++ b/frontend/src/containers/DocumentList.js
@@ -154,7 +154,7 @@ class DocumentListContainer extends Component {
      * middleware reacting to route changes and reducers
      */
     if (
-      nextProps.viewId !== nextProps.queryViewId ||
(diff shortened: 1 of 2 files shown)


frontend/src/containers/DocumentList.js
```diff
@@ -155,6 +154,7 @@ class DocumentListContainer extends Component {
      * middleware reacting to route changes and reducers
      */
     if (
+      nextProps.viewId !== nextProps.queryViewId ||
```
siemiatj, line 157: please add a comment when this is the case so we won't have to figure it out in the future

petrican: Done. Thanks

siemiatj (approved):
Haven't tested it. If works ok with included views/modals with included views it's good to go.

PR #10277, opened 2020-11-23
Gh10226 plv validation rule limi changed

BodorAnnamaria:


metas-ts (approved):
sorry, i also hadn't seen the bug in my previous review

(~~`M_PriceList_Version.ValidFrom`~~ instead of `plv.ValidFrom` right?)

now it looks ok

BodorAnnamaria:
> sorry, i also hadn't seen the bug in my previous review
> 
> (~`M_PriceList_Version.ValidFrom`~ instead of `plv.ValidFrom` right?)
> 
> now it looks ok

yes, that caused the problem

PR #10278, opened 2020-11-24
Prevent multiple re-renderings

petrican:
Possible fix for fixing the Picking tray

https://www.loom.com/share/50bc628cc8194a3ca198d50193b65471

### DIFF
--- a/frontend/src/containers/DocumentList.js
+++ b/frontend/src/containers/DocumentList.js
@@ -134,8 +134,11 @@ class DocumentListContainer extends Component {
      * TODO: This could probably be handled by a combination of
      * middleware reacting to route changes and reducers
      */
+
     if (
-      (nextProps.viewId !== nextProps.queryViewId && nextProps.queryViewId) || // for the case when you applied a filter and come back via browser back button
+      (nextProps.viewId !== nextProps.queryViewId &&
+        nextProps.queryViewId &&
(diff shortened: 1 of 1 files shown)


frontend/src/containers/DocumentList.js
```diff
@@ -134,8 +134,11 @@ class DocumentListContainer extends Component {
      * TODO: This could probably be handled by a combination of
      * middleware reacting to route changes and reducers
      */
+
     if (
-      (nextProps.viewId !== nextProps.queryViewId && nextProps.queryViewId) || // for the case when you applied a filter and come back via browser back button
+      (nextProps.viewId !== nextProps.queryViewId &&
+        nextProps.queryViewId &&
+        !this.renderedSuccessfuly) || // for the case when you applied a filter and come back via browser back button
```
siemiatj, line 141: Just set this to false when initializing the component and move the comment.

siemiatj (approved):

frontend/src/containers/DocumentList.js
petrican: It will resolve to false since is not defined. Do you still want to add it ? Or just formatting the code would be sufficient ?

siemiatj: It's just for the sake of readability.

petrican: ok

PR #10283, opened 2020-11-24
make sure that we specify if date-without-time shall be 00:00 or 23:59

metas-ts:
also
* add unit test
* unrelated minor improvements in ICounterDocBL, ICounterDocHandler


metas-ts (commented):
lgtm

metas-ts:

Issue #10282, opened 2020-11-24
Manufacturing extension: MAC Address Generator


TheBestPessimist:
### Is this a bug or feature request?
FR

PR #10286, opened 2020-11-24
Manufacturing extension: MAC Address Generator

TheBestPessimist:
https://github.com/metasfresh/metasfresh/issues/10282
Signed-off-by: TheBestPessimist <cristian@tbp.land>


metas-ts (approved):
Generally LGTM; i'm wondering if we shouldn't even hardcode the AD_Sequence_ID of `555224` in the MacAddressGenerator.
And then have both methods `generateNextMacAddress()` and `generateNextMacAddress(DocSequenceId)`. wdyt?

TheBestPessimist:
I think we should not hardcode because we don't know for sure how that will be used.

The parent task has 150 hours budget, so probs a lot will change from now until actual usage.

PR #10288, opened 2020-11-24
Invoice is not ispaid = Y although fully allocated

BodorAnnamaria:
https://github.com/metasfresh/bc107/issues/360

### DIFF
--- a/backend/de.metas.business/src/main/java-legacy/org/compiere/model/MPayment.java
+++ b/backend/de.metas.business/src/main/java-legacy/org/compiere/model/MPayment.java
@@ -21,29 +21,6 @@
  */
 package org.compiere.model;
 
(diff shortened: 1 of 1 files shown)


metas-ts (commented):
i'm sorry, you did exactly what you concepted and i approved, but: 
now it occurs to me: it there other code that does the unlinking (set `C_Payment_ID 
 = null`) for the `C_Order` and the `C_Invoice`? If not, we might still need those bits.

BodorAnnamaria:
>  it there other code that does the unlinking (set `C_Payment_ID = null`) for the `C_Order` and the `C_Invoice`? 

We do not have. I updated the concept and PR also

metas-ts (approved):

PR #10298, opened 2020-11-24
redistribute syncadvise when creating bpartner from olcand-bpartnerinfo

metas-ts:
the olcand's bpartners' "inner" syncadvices are now applied on the bpartner-request's middle level.
That way a user can specify a bpartner to be not updated, but new locators or contacts to be inserted if they don't yet exist


metas-ts (commented):
lgtm

metas-ts:

Issue #10310, opened 2020-11-26
Manufacturing extension: material dispo shall work with PP_Product_Planning.MaxManufacturedQtyPerOrder


metas-ts:
### Is this a bug or feature request?
FR

PR #10311, opened 2020-11-26
add columns and fields

metas-ts:
#10310


metas-ts (commented):
lgtm

metas-ts (commented):
lgtm and now also works

Issue #10279, opened 2020-11-24
Add Marketing Channel Attribute and Window to User


dragospodariu96:
### Is this a bug or feature request?
FR
### What is the current behavior?
No 'Attribute Channel' attributes.
#### Which are the steps to reproduce?
Go to the user window
### What is the expected or desired behavior?
Have a Marketing Channels window and a field for adding them to a user.

PR #10315, opened 2020-11-26
#10279

dragospodariu96:
#10279


backend/de.metas.marketing/base/src/main/java/de/metas/marketing/base/callout/AD_User_TabCallout.java
```diff
@@ -0,0 +1,77 @@
...
+	private final IUserDAO userDAO = Services.get(IUserDAO.class);
+	private final ISysConfigBL sysConfigBL = Services.get(ISysConfigBL.class);
+
+	private static final AdMessageKey MSG_CAN_NOT_REMOVE_CHANNEL = AdMessageKey.of("de.metas.marketing.base.userMarketingChannelRemovalError");
+	private static final String SYS_CONFIG_MARKETING_CHANNELS_ENFORCED = "de.metas.marketing.EnforceUserHasMarketingChannels";
+
+	@Override
+	public void onSave(final ICalloutRecord calloutRecord)
+	{
+		I_AD_User user = calloutRecord.getModel(I_AD_User.class);
```
metas-ts, line 49:
```suggestion
		final I_AD_User user = calloutRecord.getModel(I_AD_User.class);
```

backend/de.metas.marketing/base/src/main/java/de/metas/marketing/base/callout/AD_User_TabCallout.java
```diff
@@ -0,0 +1,77 @@
...
+
+		if (!canBeSaved)
+		{
+			throw new AdempiereException(MSG_CAN_NOT_REMOVE_CHANNEL).markAsUserValidationError();
+		}
+	}
+
+	private boolean isMarketingChannelsUseEnforced()
+	{
+		return sysConfigBL.getBooleanValue(SYS_CONFIG_MARKETING_CHANNELS_ENFORCED, false, Env.getAD_Client_ID(Env.getCtx()), Env.getAD_Org_ID(Env.getCtx()));
```
metas-ts, line 75: please check this with the client-id and org-id of the `I_AD_User` in question.

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5573480_sys_gh10279AddEnforceMarketingChannelsSysConfig.sql
```diff
@@ -0,0 +1,5 @@
+-- 2020-11-27T06:12:31.788Z
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+INSERT INTO AD_SysConfig (AD_Client_ID,AD_Org_ID,AD_SysConfig_ID,ConfigurationLevel,Created,CreatedBy,EntityType,IsActive,Name,Updated,UpdatedBy,Value) VALUES (0,0,541346,'O',TO_TIMESTAMP('2020-11-27 08:12:31','YYYY-MM-DD HH24:MI:SS'),100,'de.metas.marketing.base','Y','de.metas.marketing.EnforceUserHasMarketingChannels',TO_TIMESTAMP('2020-11-27 08:12:31','YYYY-MM-DD HH24:MI:SS'),100,'N')
```
metas-ts, line 3: please add a description that explains that this applies 
* only to not-system users
* only whn added in the user-window

backend/de.metas.marketing/base/src/main/java/de/metas/marketing/base/interceptor/AD_User_MKTG_Channels.java
```diff
@@ -26,36 +26,55 @@
...
+		final int count = mktgChannelDao.retrieveMarketingChannelsCountForUser(UserId.ofRepoId(userMktgChannels.getAD_User_ID()));
 		if (count == 1)
 		{
 			throw new AdempiereException(MSG_CAN_NOT_REMOVE_CHANNEL).markAsUserValidationError();
 		}
 	}
+
+	private boolean isMarketingChannelsUseEnforced()
+	{
+		return sysConfigBL.getBooleanValue(SYS_CONFIG_MARKETING_CHANNELS_ENFORCED, false, Env.getAD_Client_ID(Env.getCtx()), Env.getAD_Org_ID(Env.getCtx()));
```
metas-ts, line 78: please check this with the client-id and org-id of the I_AD_User in question. (same as in the tab-callout)

backend/de.metas.marketing/base/src/main/java/de/metas/marketing/base/interceptor/MKTG_Channel.java
```diff
@@ -54,20 +58,30 @@ private MKTG_Channel()
...
 	public void checkIfCanBeDeleted(@NonNull final I_MKTG_Channel mktgChannel)
 	{
+		if (!isMarketingChannelsUseEnforced())
+		{
+			return;
+		}
+
 		boolean canBeDeleted = true;
 		Set<UserId> usersSet = mktgChannelDao.retrieveUsersHavingChannel(mktgChannel.getMKTG_Channel_ID());
 		for (UserId userId : usersSet)
```
metas-ts, line 68:
```suggestion
		for (final UserId userId : usersSet)
```

backend/de.metas.marketing/base/src/main/java/de/metas/marketing/base/interceptor/MKTG_Channel.java
```diff
@@ -54,20 +58,30 @@ private MKTG_Channel()
 			timings = { ModelValidator.TYPE_BEFORE_DELETE })
 	public void checkIfCanBeDeleted(@NonNull final I_MKTG_Channel mktgChannel)
 	{
+		if (!isMarketingChannelsUseEnforced())
```
metas-ts, line 61: But can we generally delete a marketingchannel just like that? wont there be a DB exception in case the channel is still referenced by `AD_User_MKTG_Channel` records?
So i think you probably don't need to change this method at all..

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/window/model/Document.java
```diff
@@ -2164,7 +2164,7 @@ public Document build()
 
 			final DocumentEntityDescriptor entityDescriptor = getEntityDescriptor();
 			final ITabCallout documentCallout = entityDescriptor.createAndInitializeDocumentCallout(document.asCalloutRecord());
-			document.documentCallout = ExceptionHandledTabCallout.wrapIfNeeded(documentCallout);
```
metas-ts, line 2167: what is the background of this change?

metas-ts (changes requested):

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/window/model/Document.java
dragospodariu96: The user was still saved because the exception wasn't handled properly

backend/de.metas.marketing/base/src/main/java/de/metas/marketing/base/interceptor/MKTG_Channel.java
dragospodariu96: @metas-ts clarified with @metas-mk It's ok to not be able to delete a marketing channel if a there is still a reference on the mktg_channel_id from the AD_User_MKTG_Channels table. This means we could completely delete this interceptor and the de.metas.marketing.base.api.IMKTGChannelDao#retrieveUsersHavingChannel method. However, AFAIU the possibility to only not be able to remove a channel if a user would remain with 0 will be implemented in another task, so I propose to comment out the line (+a comment with this task and explanation) that registers this interceptor and expand this in the future (A IMktgChannelBL class with a method that does necessary operation on the ad_user_mktg_channels table. WDYT?

metas-ts: @dragospodariu96 

> It's ok to not be able to delete a marketing channel if a there is still a reference on the mktg_channel_id from the AD_User_MKTG_Channels table. 

OK

>  This means we could completely delete this interceptor 

no, because we still want a user-friendly error message when they try to delete a `MKTG_Channel` that is still linked to users.
that's why i proposed to "revert" it.

> will be implemented in another task[...](+a comment with this task and explanation)

if you think it will help in future, feel free to add a comment or something, so a future dev doesn't have to start at zero

dragospodariu96: @metas-ts 

 Updated the interceptor's method. User friendly error message now shown when trying to delete a marketing channel that is still assigned to at least one user, regardless of the value from the 'enforce marketing channels use' sys config. 

OBS: Because regardless of that config, if we tried to delete a channel still assigned to a user, we would get an error due to the foreign key constraint on the ad_user_mktg_channels table

backend/de.metas.marketing/base/src/main/java/de/metas/marketing/base/interceptor/MKTG_Channel.java
```diff
@@ -56,18 +56,18 @@ public void checkIfCanBeDeleted(@NonNull final I_MKTG_Channel mktgChannel)
 	{
 		boolean canBeDeleted = true;
 		Set<UserId> usersSet = mktgChannelDao.retrieveUsersHavingChannel(mktgChannel.getMKTG_Channel_ID());
-		for (UserId userId : usersSet)
+		for (final UserId userId : usersSet)
 		{
-			if (mktgChannelDao.retrieveMarketingChannelsCountForUser(userId) == 1)
+
+			if (mktgChannelDao.retrieveMarketingChannelsCountForUser(userId) > 0 && !userDAO.isSystemUser(userId))
```
metas-ts, line 62: Even for system-users, you won't be able to delete the channel, simply because of the foreign-key-constraint we talked about.
Remember, you tried it when we had the screen-session..

metas-ts (changes requested):

backend/de.metas.marketing/base/src/main/java/de/metas/marketing/base/interceptor/MKTG_Channel.java
dragospodariu96: Yeah, forgot about that 😓 Removed the condition.

metas-ts (approved):

PR #10331, opened 2020-11-30
Propagate the salespartner-ID to the C_Order

metas-ts:
#10330

### DIFF
--- a/backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/salesrep/interceptor/C_Order.java
+++ b/backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/salesrep/interceptor/C_Order.java
@@ -68,8 +68,9 @@ public class C_Order
 			ifColumnsChanged = { I_C_Order.COLUMNNAME_C_BPartner_ID, I_C_Order.COLUMNNAME_Bill_BPartner_ID })
 	public void updateSalesPartnerFromCustomer(@NonNull final I_C_Order orderRecord, @NonNull final ModelChangeType type)
 	{
-		// if on a new C_Order record both SalesPartnerCode and the BPartner-IDs were set at the same time,
-		// then don't override the sales partner from the BPartners' mater data, but assume that the sales partner shall remain the way it is.
+		// If on a new C_Order record both SalesPartnerCode and the BPartner-IDs were set at the same time,
+		// then don't override the sales-partner-id from the BPartners' mater data, but assume that the sales-partner-id shall remain the way it was set on the new record.
+		// This implies that the master data might be updated on thid C_Order's after-new modelinterceptor method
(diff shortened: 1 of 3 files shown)


metas-ts (commented):
lgtm

backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/salesrep/interceptor/C_Order.java
```diff
@@ -62,8 +63,21 @@ public C_Order(
 		Services.get(IProgramaticCalloutProvider.class).registerAnnotatedCallout(this);
 	}
 
+	@ModelChange(
+			timings = { ModelValidator.TYPE_BEFORE_NEW, ModelValidator.TYPE_BEFORE_CHANGE },
+			ifColumnsChanged = { I_C_Order.COLUMNNAME_C_BPartner_ID, I_C_Order.COLUMNNAME_Bill_BPartner_ID })
+	public void updateSalesPartnerFromCustomer(@NonNull final I_C_Order orderRecord, @NonNull final ModelChangeType type)
+	{
+		// if on a new C_Order record both SalesPartnerCode and the BPartner-IDs were set at the same time,
+		// then don't override the sales partner from the BPartners' mater data, but assume that the sales partner shall remain the way it is.
```
metas-ts, line 72:
```suggestion
		// then don't override the sales-partner-id from the BPartners' mater data, but assume that the sales-partner-id shall remain the way it was set on the new record.
		// This implies that the master data might be updated on thid C_Order's after-new modelinterceptor method
```

backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/salesrep/interceptor/C_Order.java
```diff
@@ -62,8 +63,21 @@ public C_Order(
 		Services.get(IProgramaticCalloutProvider.class).registerAnnotatedCallout(this);
 	}
 
+	@ModelChange(
+			timings = { ModelValidator.TYPE_BEFORE_NEW, ModelValidator.TYPE_BEFORE_CHANGE },
+			ifColumnsChanged = { I_C_Order.COLUMNNAME_C_BPartner_ID, I_C_Order.COLUMNNAME_Bill_BPartner_ID })
+	public void updateSalesPartnerFromCustomer(@NonNull final I_C_Order orderRecord, @NonNull final ModelChangeType type)
+	{
+		// if on a new C_Order record both SalesPartnerCode and the BPartner-IDs were set at the same time,
```
metas-ts, line 71:
```suggestion
		// If on a new C_Order record both SalesPartnerCode and the BPartner-IDs were set at the same time,
```

PR #10346, opened 2020-12-03
Gh6310 new service handling units window

BodorAnnamaria:


backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/handlingunits/HUEditorViewFactoryTemplate.java
```diff
@@ -492,6 +492,10 @@ public String getSql(
 
 	private boolean isAlwaysUseSameLayout()
 	{
+		//When the Factory is ServiceHUEditorViewFactory always need to execute the customizeViewLayout method, is not configuration dependent
```
metas-ts, line 495: * please make the method `protected and override it in `ServiceHUEditorViewFactory`
* add the comment "When the Factory is ServiceHUEditorViewFactory always need to execute the customizeViewLayout method, is not configuration dependent" 
to the overriding method as a javadoc
* (it's cleaner)

metas-ts (changes requested):

metas-ts (approved):

cristinamghita:
Linked to https://github.com/metasfresh/metasfresh/issues/10414

PR #10353, opened 2020-12-03
Unlinking Quotation/Order before deleting

arthurlmf:
https://github.com/metasfresh/me03/issues/6304
https://github.com/metasfresh/metasfresh/issues/10340


metas-ts (approved):
i guess it's OK like this..it's a bit of a gray area imho..might also have created a "general" unlink-method in IOrderBL

Issue #10276, opened 2020-11-23
Some windows not usable sometimes in case of mass shipping / mass invoicing 


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
Unusable window during mass invoice / mass shipping
#### Which are the steps to reproduce?
Unusable window during mass invoice / mass shipping
### What is the expected or desired behavior?
Shoud work without issues.

PR #10356, opened 2020-12-04
Batch together ws events to refresh the data grid

siemiatj:
Related to #10276 

https://recordit.co/lGHmwiW19G


petrican (commented):
lgtm. I would add some small comment explaining why we did that debouncing. thanks

PR #10362, opened 2020-12-07
Change how full grid refreshes from websockets work

siemiatj:
Related to #10276 . Impact is limited to grid views only and only in reaction to a websocket event, so it's safe to merge anytime.

https://recordit.co/zfdKI8KD3p

### DIFF
--- a/frontend/src/actions/ViewActions.js
+++ b/frontend/src/actions/ViewActions.js
@@ -64,12 +64,17 @@ export function deleteView(id, isModal) {
 
 /**
  * @method fetchDocumentPending
- * @summary
+ * @summary request data for the document and set the pending flag to true
+ *
+ * @param {string} id - viewId
+ * @param {boolean} isModal
+ * @param {boolean} websocketRefresh - in case of data fetches caused by
+ * ws we won't set the pending flag to true
  */
-function fetchDocumentPending(id, isModal) {
+function fetchDocumentPending(id, isModal, websocketRefresh) {
   return {
     type: FETCH_DOCUMENT_PENDING,
-    payload: { id, isModal },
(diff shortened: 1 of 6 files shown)


frontend/src/containers/DocumentList.js
```diff
@@ -459,7 +471,7 @@ class DocumentListContainer extends Component {
    * @method getData
    * @summary Loads view/included tab data from REST endpoint
    */
-  getData = (id, page, sortingQuery, locationAreaSearch) => {
+  getData = (id, page, sortingQuery, locationAreaSearch, websocketRefresh) => {
```
petrican, line 474: too many params. Can we use object param in here ?

frontend/src/actions/ViewActions.js
```diff
@@ -416,7 +422,7 @@ export function fetchDocument({
         }
 
         // get quickactions
-        if (viewId && shouldFetchQuickActions) {
+        if (viewId && shouldFetchQuickActions && !websocketRefresh) {
```
petrican, line 425: As I understand from Teo this is NOK as quickactions shall be refreshed

petrican (changes requested):

siemiatj:
Ok, so now we're also refreshing quick actions and not showing the spinner overlay at all https://recordit.co/nZ6Gl8yDX3

frontend/src/containers/DocumentList.js
siemiatj: Dude this is a friggin' hack I want to remove as soon as possible and not spend a second more on it than needed.

petrican (approved):
lgtm

Issue #10363, opened 2020-12-07
Don't flash rows when there's no update


siemiatj:
### Is this a bug or feature request?

Bug

### What is the current behavior?

![broken](https://user-images.githubusercontent.com/513460/101362823-eaca9580-38a0-11eb-92a7-c68105aed236.gif)

#### Which are the steps to reproduce?

Open grid with editable fields and navigate around it

### What is the expected or desired behavior?

![fixed](https://user-images.githubusercontent.com/513460/101362870-f8801b00-38a0-11eb-845b-6240936e052e.gif)

PR #10367, opened 2020-12-07
Don't flash the row as if it updated when navigating around the table

siemiatj:
Related to #10363 

Before: https://recordit.co/OBUIdLzpnF

After: https://recordit.co/Dpn9l0JoQS


petrican (approved):
lgtm. Thanks

PR #10392, opened 2020-12-09
Fix setting parent for the included view

siemiatj:
This was breaking the quickactions fetching. Also refactored some functions with multiple params to accept named properties.
Tests will follow soon but preparing the fixtures for the failing cases is a tedious work and we need to start testing this asap.

Related to #10195

### DIFF
--- a/frontend/src/actions/WindowActions.js
+++ b/frontend/src/actions/WindowActions.js
@@ -1211,7 +1211,6 @@ export function handleProcessResponse(response, type, id, parentId) {
 
               await dispatch(
                 openModal({
-                  title: '',
                   windowId: action.windowId,
                   modalType: 'window',
                   isAdvanced: action.advanced ? action.advanced : false,
--- a/frontend/src/actions/WindowActions.js
+++ b/frontend/src/actions/WindowActions.js
@@ -1059,7 +1059,8 @@ export function createProcess({
   let pid = null;
 
   return async (dispatch, getState) => {
-    // creation of processes can be done only if there isn't any pending process https://github.com/metasfresh/metasfresh/issues/10116
+    // creation of processes can be done only if there isn't any pending process
+    // https://github.com/metasfresh/metasfresh/issues/10116
     const { processStatus } = getState().appHandler;
     if (processStatus === 'pending') {
       return false;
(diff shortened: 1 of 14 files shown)


frontend/src/actions/WindowActions.js
```diff
@@ -1201,17 +1210,13 @@ export function handleProcessResponse(response, type, id, parentId) {
...
-                  'window',
-                  null,
-                  null,
-                  action.advanced ? action.advanced : false,
-                  '',
-                  '',
-                  action.documentId
-                )
+                openModal({
+                  title: '',
```
petrican, line 1214: Should this be passed as param in here if it already has the same default value ?

frontend/src/components/app/Modal.js
```diff
@@ -214,6 +215,7 @@ class Modal extends Component {
             processType: windowId,
             viewId,
             parentId: parentWindowId,
```
petrican, line 217: I'm a bit confused. We are expecting in here parentWindowId and this is not passed anymore. This may lead to future logic bugs.

petrican (changes requested):
The changes are sensitive to the app functionality. I know that this fix is required asap but I believe it should be tested more before merging. Pls deploy a dev instance for some smoke testing. Done just a quick check and the BP does not work (Go to Business Partner and open one of the items)

petrican:
> The changes are sensitive to the app functionality. I know that this fix is required asap but I believe it should be tested more before merging. Pls deploy a dev instance for some smoke testing. Done just a quick check and the BP does not work (Go to Business Partner and open one of the items)

Update: My bad I was consuming the APi from another instance and spot this. It works fine against releasetestit
Screenshot 2020-12-09 at 13 54 28

frontend/src/actions/WindowActions.js
siemiatj: Good point

frontend/src/components/app/Modal.js
siemiatj: Yeah, a leftover since we don't use this anywhere

petrican:
On a quick smoke tests it seems to be working fine. Note that I haven't roughly tested

petrican (approved):

Issue #10397, opened 2020-12-09
Reference to commission Invoice candidate is missing in commission deed


adi-stefan:
### Is this a bug or feature request?
bug
### What is the current behavior?
In commission deed, there's no reference to the commission invoice candidate.
#### Which are the steps to reproduce?
On a commission deed, open the referenced documents. No Purchase Invoice Candidate.
### What is the expected or desired behavior?
When opening the referenced documents, there should be a Purchase Invoice Candidate.

l4b4r4b4b4:
I still face this issue. Does the commit @adi-stefan fix this?

PR #10399, opened 2020-12-09
#gh10397 added reference to sales order invoice & purchase order invoice, and logic to switch between displaying both  https://github.com/metasfresh/metasfresh/issues/10397

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/10397


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/document/references/RelationTypeZoomProvider.java
```diff
@@ -309,6 +310,11 @@ private String createZoomOriginQueryWhereClause(final IZoomSource zoomOrigin)
 						.append(originTableName)
 						.append(".")
 						.append(originRecordIdName);
+				if (whereClause != null)
```
metas-ts, line 313:
```suggestion
				if (Check.isNotBlank(whereClause))
```

...with `de.metas.util.Check`...this also takes care of e.g. `"    "`

metas-ts (approved):
please check https://github.com/metasfresh/metasfresh/pull/10399#pullrequestreview-548379034 before integrating

PR #10407, opened 2020-12-10
#10377 Also setting request date

dragospodariu96:
#10377

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/request/api/impl/RequestDAO.java
+++ b/backend/de.metas.business/src/main/java/de/metas/request/api/impl/RequestDAO.java
@@ -1,6 +1,7 @@
 package de.metas.request.api.impl;
 
 import de.metas.bpartner.BPartnerId;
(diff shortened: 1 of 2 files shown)


backend/de.metas.business/src/main/java/de/metas/request/api/impl/RequestDAO.java
```diff
@@ -62,6 +64,8 @@ public I_R_Request createRequest(@NonNull final RequestCandidate candidate)
 		request.setM_InOut_ID(candidate.getRecordRef() != null ?
 				candidate.getRecordRef().getTableName().equals(I_M_InOut.Table_Name) ? candidate.getRecordRef().getRecord_ID() : -1
 				: -1);
+
+		request.setDateTrx(TimeUtil.asTimestamp(Date.from(Instant.now())));
```
metas-ts, line 68: Please use `SystemTime` instead

metas-ts (approved):
please make that minor change. with that, lgtm to integrate

PR #10418, opened 2020-12-11
Gh10393

dragospodariu96:
https://github.com/metasfresh/mf15/issues/487

### DIFF
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/interfaces/I_C_OrderLine.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/interfaces/I_C_OrderLine.java
@@ -43,4 +43,8 @@ public interface I_C_OrderLine extends org.compiere.model.I_C_OrderLine, IDocume
 	void setHandOver_Location_ID(int handoverLocationId);
 	int getHandOver_Location_ID();
 	// @formatter:on
+
+	String COLUMNNAME_M_HU_PI_Item_Product_ID = "M_HU_PI_Item_Product_ID";
+	void setM_HU_PI_Item_Product_ID(int M_HU_PI_Item_Product_ID);
+	int getM_HU_PI_Item_Product_ID();
 }
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/interfaces/I_C_OrderLine.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/interfaces/I_C_OrderLine.java
@@ -44,7 +44,6 @@ public interface I_C_OrderLine extends org.compiere.model.I_C_OrderLine, IDocume
 	int getHandOver_Location_ID();
 	// @formatter:on
 
-	String COLUMNNAME_M_HU_PI_Item_Product_ID = "M_HU_PI_Item_Product_ID";
 	void setM_HU_PI_Item_Product_ID(int M_HU_PI_Item_Product_ID);
 	int getM_HU_PI_Item_Product_ID();
 }
--- a/backend/de.metas.business/src/main/java/de/metas/order/impl/OrderLineBL.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/impl/OrderLineBL.java
@@ -329,6 +329,12 @@ public class OrderLineBL implements IOrderLineBL
 	}
 
 	@Override
+	public void updateMHuPiItemProduct(final I_C_OrderLine orderLine, final int huPiItemProductId)
+	{
+		orderLine.setM_HU_PI_Item_Product_ID(huPiItemProductId);
+	}
+
+	@Override
 	public void updateLineNetAmtFromQty(
 			@NonNull final Quantity qtyOverride,
 			@NonNull final org.compiere.model.I_C_OrderLine orderLine)
--- a/backend/de.metas.business/src/main/java/de/metas/order/impl/OrderLineBL.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/impl/OrderLineBL.java
@@ -329,12 +329,6 @@ public class OrderLineBL implements IOrderLineBL
 	}
 
 	@Override
-	public void updateMHuPiItemProduct(final I_C_OrderLine orderLine, final int huPiItemProductId)
-	{
-		orderLine.setM_HU_PI_Item_Product_ID(huPiItemProductId);
-	}
-
-	@Override
 	public void updateLineNetAmtFromQty(
 			@NonNull final Quantity qtyOverride,
 			@NonNull final org.compiere.model.I_C_OrderLine orderLine)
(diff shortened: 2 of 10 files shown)


backend/de.metas.business/src/main/java/de/metas/order/impl/OrderLineBL.java
```diff
@@ -328,6 +328,12 @@ public void updateLineNetAmtFromQtyEntered(@NonNull final org.compiere.model.I_C
 		updateLineNetAmtFromQtyInPriceUOM(orderLine, qtyInPriceUOM);
 	}
 
+	@Override
+	public void updateMHuPiItemProduct(final I_C_OrderLine orderLine, final int huPiItemProductId)
```
metas-ts, line 332: not really sure what the point of this method is..

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/interfaces/I_C_OrderLine.java
```diff
@@ -43,4 +43,8 @@
 	void setHandOver_Location_ID(int handoverLocationId);
 	int getHandOver_Location_ID();
 	// @formatter:on
+
+	String COLUMNNAME_M_HU_PI_Item_Product_ID = "M_HU_PI_Item_Product_ID";
```
metas-ts, line 47: we already have this, in `de.metas.handlingunits.model.I_C_Order`

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/callout/C_OrderLine.java
```diff
@@ -47,6 +55,17 @@ public void updateQtyCU(final I_C_OrderLine orderLine, final ICalloutField field
 		Services.get(IOrderLineBL.class).updateLineNetAmtFromQtyEntered(orderLine);
 	}
 
+	@CalloutMethod(columnNames = de.metas.interfaces.I_C_OrderLine.COLUMNNAME_M_Product_ID)
+	public void onProductSetOrChanged(final de.metas.interfaces.I_C_OrderLine orderLine)
+	{
+		final Optional<I_M_HU_PI_Item_Product> huPiItemProduct = hupiItemProductDAO.retrieveDefaultForProduct(ProductId.ofRepoId(orderLine.getM_Product_ID()),
```
metas-ts, line 61: * 1 i think it would be better to do this in a ModelInterceptor..i believe it make sense to do this also if e.g. the order line is created via scheduler from an C_OLCand
* 2 if the new product has no M_HU_PI_Item_Product_ID, then i think C_OrderLine.M_HU_PI_Item_Product_ID should be set to -1
* 3 please add a method `Optional<HUPIItemProductId> retrieveDefaultIdForProduct` to IHUPIItemProductDAO`
  * rationale: try to avoid throwing around data records. we like to grow in the general direction of DDD.
  * also, we don't need (to load) the record for this

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/quickinput/orderline/OrderLineQuickInputDescriptorFactory.java
```diff
@@ -175,7 +175,7 @@ private ProductLookupDescriptor createProductLookupDescriptor(@NonNull final Opt
 						.setCtxTableName(null) // ctxTableName
 						.setCtxColumnName(IOrderLineQuickInput.COLUMNNAME_M_HU_PI_Item_Product_ID)
 						.setDisplayType(DisplayType.TableDir)
-						.setAD_Val_Rule_ID(540365) // FIXME: hardcoded "M_HU_PI_Item_Product_For_Org_and_Product_and_DatePromised"
+						.setAD_Val_Rule_ID(540365)// FIXME: hardcoded "M_HU_PI_Item_Product_For_Org_and_Product_and_DatePromised"
```
metas-ts, line 178: i didn't understand in the concept, why this was supposed to be changed..so i guess it's fine that it effectively was *not* changed at all

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderLineBL.java
adi-stefan: Me neither, removed.

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/interfaces/I_C_OrderLine.java
adi-stefan: Removed.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/callout/C_OrderLine.java
adi-stefan: 1. Moved to ModelInterceptor
2. HUPIItemProductId.toRepoId(null) will return -1, used that.
3. Added, though I've only used it for our particular case. There are other places which still call `retrieveDefaultForProduct`, but we'd need to test those changes....

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/impl/HUPIItemProductDAO.java
```diff
@@ -487,6 +488,7 @@ public I_M_HU_PI_Item_Product retrieveMaterialItemProduct(
 				// Get specific Product first
 				.addColumn(I_M_HU_PI_Item_Product.COLUMNNAME_IsAllowAnyProduct, Direction.Descending, Nulls.Last) // Y first, N second
 				.addColumn(I_M_HU_PI_Item_Product.COLUMNNAME_M_Product_ID, Direction.Descending, Nulls.Last)
+				.addColumn(I_M_HU_PI_Item_Product.COLUMNNAME_IsDefaultForProduct, Direction.Descending, Nulls.Last)
```
adi-stefan, line 491: This was missing from Dragos's changes, but was necessary (at least for the batch add)

backend/metasfresh-webui-api/src/main/java/de/metas/ui/web/quickinput/ddorderline/DDOrderLineQuickInputCallout.java
```diff
@@ -72,7 +73,7 @@ public void onProductChange(@NonNull final ICalloutField calloutField)
 		}
 		final I_DD_Order ddOrder = quickInput.getRootDocumentAs(I_DD_Order.class);
 
-		ddOrderLineQuickInput.setM_HU_PI_Item_Product(getDefaultPI(ddOrder, productId));
+		ddOrderLineQuickInput.setM_HU_PI_Item_Product_ID(getDefaultPI(ddOrder, productId).getM_HU_PI_Item_Product_ID());
```
metas-ts, line 76: please make sure we don't run into an NPE here

metas-ts (changes requested):
please check my request. when that's sorted, you can proceed without asking from another review

metas-ts (approved):

PR #10420, opened 2020-12-11
Remove unnecessary redux actions

siemiatj:
Fix flaky e2e picking tests and remove some redundant redux actions.

Related to #10177 & #10195


siemiatj:
Screenshot 2020-12-17 at 15 14 45

petrican (changes requested):
The PR is fine. Pls extract that id in a variable. Thinking that if we release today we should merge this on Monday to make sure any changes brought would be also tested for the next release.

PR #10433, opened 2020-12-15
Add property skipIfIndirectlyCalled to annotation CalloutMethod

metas-ts:
...and use it to set "manualPrice" to Y *only* if user manually edits priceEntered


backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/callout/annotations/CalloutMethod.java
```diff
@@ -59,4 +59,10 @@
 	 * Skip calling this callout if we are copying (with details).
 	 */
 	boolean skipIfCopying() default false;
+
+	/**
+	 * Skip calling this callout if it is called via another callout.
+	 * Use case: you want a callout to do stuff only if invoked by a user (or business logic).
```
metas-ts, line 65:
```suggestion
	 * Use case: you want a callout to do stuff only if invoked *directly* by a user (or business logic).
```

metas-ts (commented):
lgtm

PR #10479, opened 2020-12-31
add cucumber and remove jms

metas-ts:
* change procurement webui communication from rest-over-jms-jax-rs to rabbitmq
* allow workspace-migration-client, app and webapi to rund without local config files
* add cucumber testing with two examples for app
* remove jms and jax.rs legacy stuff


backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/de/metas/util/ConnectionUtil.java
```diff
@@ -0,0 +1,102 @@
...
+import lombok.experimental.UtilityClass;
+import org.compiere.db.CConnection;
+import org.compiere.db.CConnectionAttributes;
+import org.slf4j.Logger;
+
+import javax.annotation.Nullable;
+
+import static de.metas.common.util.CoalesceUtil.coalesce;
+
+@UtilityClass
```
metas-ts, line 37:
```suggestion

@UtilityClass
```

metas-ts (commented):
went over all files one more; should be good to integrate after successful build

PR #10481, opened 2021-01-04
StepComXMLDesadvBean and CompuDataDesadvBean: don't remove leading 0s

metas-ts:
also div minor fixes in metasfresh


metas-ts (commented):
lgtm

PR #10488, opened 2021-01-04
rearrange procurement-webui folders

metas-ts:


metas-ts (commented):
talked with kuba&petrica..ok to integrate after sucessful build

Issue #10497, opened 2021-01-07
I_BankStatement.DebitOrCreditIndicator interchange


metas-rc:
### Is this a bug or feature request?
* Bug
### What is the current behavior?
* In the Bank Statement Import, the Debit has the value S and the Credit has the value H. But they should be used the other way around.
### What is the expected or desired behavior?
Credit value must be S and Debit value must be H.

PR #10498, opened 2021-01-07
Interchange the DebitOrCreditIdentifier values in the Bank Statement Import

metas-rc:
https://github.com/metasfresh/metasfresh/issues/10497


metas-ts (approved):
despite that like i wrote i have some doubts that H=Debit and S=Credit is right, it looks good. please clarify this with mark

Issue #10501, opened 2021-01-08
Extended Payment REST Endpoint


adi-stefan:
### Is this a bug or feature request?
FR

### What is the current behavior?
Currently we're only able to create a C_Payment from the `/api/payment/inbound` endpoint.

#### Which are the steps to reproduce?
See above

### What is the expected or desired behavior?
Additional C_AllocationHdr and C_AllocationLine records should be created from the same entry point.

PR #10517, opened 2021-01-14
#gh10501 bug fixes

adi-stefan:
Allow a blank docBaseType if not using doc- as IdentifierString
If OrderIdentifier is of type external, persist to C_Payment.ExternalId;
if OrderIdentifier is not of type external retrieve and persist C_Payment.ExternalId and C_Payment.C_Order_ID
https://github.com/metasfresh/metasfresh/issues/10501

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/order/IOrderDAO.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/IOrderDAO.java
@@ -115,7 +115,7 @@ public interface IOrderDAO extends ISingletonService
 
 	/**
 	 * @param order
-	 * @param line
+	 * @param lineNo
 	 * @param clazz
 	 */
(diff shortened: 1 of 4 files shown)


backend/de.metas.business/src/main/java/de/metas/order/IOrderDAO.java
```diff
@@ -115,7 +115,7 @@ default I_C_OrderLine getOrderLineById(@NonNull final OrderAndLineId orderAndLin
 
 	/**
 	 * @param order
-	 * @param line
+	 * @param lineNo
 	 * @param clazz
 	 */
 	<T extends org.compiere.model.I_C_OrderLine> T retrieveOrderLine(I_C_Order order, int lineNo, Class<T> clazz);
```
metas-ts, line 121: nothing to do know, but if you come across useless javadoc link this, you can in future remove it altogether

metas-ts (approved):

PR #10544, opened 2021-01-21
refs: https://github.com/metasfresh/metasfresh/issues/10541

cp-ps:


backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/salesrep/process/CreateMissingCommissionEntriesForSalesRep.java
```diff
@@ -0,0 +1,421 @@
...
+			}
+		}
+	}
+
+	public void recalculateCommissionForCriteria(@NonNull final RecalculateCommissionCriteria commissionCriteria)
+	{
+		bPartnerBL.setBPartnerSalesRepId(commissionCriteria.getBPartnerId(), commissionCriteria.getTopLevelSalesRepId());
+
+		final InvoiceCandidateQuery invoiceCandidateQuery = InvoiceCandidateQuery.builder()
+				.dateToInvoiceInterval(commissionCriteria.getTargetInterval())
```
metas-ts, line 315: ..as written elsewhere: for invoice candidates, the DateOrdered date is the relevant one..

backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/salesrep/process/CreateMissingCommissionEntriesForSalesRep.java
```diff
@@ -0,0 +1,421 @@
...
+				.salesRepBPartnerId(commissionCriteria.getBPartnerId())
+				.build();
+
+		final InvoiceCandidateMultiQuery multiQuery = InvoiceCandidateMultiQuery.builder().query(invoiceCandidateQuery).build();
+
+		final List<I_C_Invoice_Candidate> invoiceCandidates = invoiceCandDAO.getByQuery(multiQuery);
```
metas-ts, line 321: we need to do the same for C_Invoice..we have aprox *200* commission instances which were triggered by an invoice or a credit memo

metas-ts (changes requested):
note, i'm not yet fully though..

backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/salesrep/process/CreateMissingCommissionEntriesForSalesRep.java
```diff
@@ -0,0 +1,428 @@
...
+	private final IBPartnerDAO bPartnerDAO = Services.get(IBPartnerDAO.class);
+	private final IUserDAO userDAO = Services.get(IUserDAO.class);
+	private final IBPartnerBL bPartnerBL = Services.get(IBPartnerBL.class);
+	private final IInvoiceCandDAO invoiceCandDAO = Services.get(IInvoiceCandDAO.class);
+
+	private static final String FROM_DATE = "2020-01-01";
+	private static final String TO_DATE = "2020-12-31";
+
+	@Override
+	protected String doIt() throws Exception
```
metas-ts, line 93: If think it makes sense to annotate the `doIt()` method with `@RunOutOfTrx`. that way the doIt() method is not run in a trx that might eventually time out and be killed

backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/salesrep/process/CreateMissingCommissionEntriesForSalesRep.java
```diff
@@ -0,0 +1,432 @@
...
+			else
+			{
+				return Optional.of(BPartnerId.ofRepoId(superSalesRepBPartner.getC_BPartner_ID()));
+			}
+		}
+	}
+
+	public void recalculateCommissionForCriteria(@NonNull final RecalculateCommissionCriteria commissionCriteria)
+	{
+		bPartnerBL.setBPartnerSalesRepId(commissionCriteria.getBPartnerId(), commissionCriteria.getTopLevelSalesRepId()); // set within the current transaction
```
metas-ts, line 323: problem here: it's sett in the current trx, but when we read this value in `CommissionHierarchyFactory#createFor()`, 
we read ot from a C_BPartner that was loaded with InterfaceWrapperHelper.loadOutOfTrx()
So there in CommissionHierarchyFactory, the C_BPartner does not have the TopLevelSalesRepId we are setting here and therefor we get no hierarchy

Suggestion 1: change CommissionHierarchyFactory to load it with `InterfaceWrapperHelper.load` (within-trx), and remove the new trxManager.runInNewTrx from line 339 (otherwise "within-trx" is within *another* trx)
Suggestion 2: change IBPartnerBL.setBPartnerSalesRepId to same the changed record out of trx as well

backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/salesrep/process/CreateMissingCommissionEntriesForSalesRep.java
```diff
@@ -0,0 +1,432 @@
...
+		final InvoiceCandidateMultiQuery multiQuery = InvoiceCandidateMultiQuery.builder().query(invoiceCandidateQuery).build();
+
+		final List<I_C_Invoice_Candidate> invoiceCandidates = invoiceCandDAO.getByQuery(multiQuery);
+
+		Loggables.withLogger(logger, Level.DEBUG).addLog("*** DEBUG: recalculating commission for criteria: {}, found {} ICs",
+				commissionCriteria, invoiceCandidates.size());
+		invoiceCandidates.forEach(invoiceCandidate -> {
+			try
+			{
+				trxManager.runInNewTrx(() -> invoiceCandFacadeService.syncICToCommissionInstance(invoiceCandidate, false));
```
metas-ts, line 339: I think we should remove this new trx. it's nice to fail just one IC, but on the other hand, this also adds complexity and is tbh over my head.
So, I propose to have each `RecalculateCommissionCriteria` in its own trx and that's it.
If invoiceCandFacadeService throws an exception, then catch it log the error, and do something like
```java
throw AdempiereException.wrapIfNeeded(e)
	.appendParametersToMessage()
	.setParameter("C_Invoice_Candidate_ID", invoiceCandidate.getC_Invoice_Candidate_ID);
```
that way the whole RecalculateCommissionCriteria is skipped and we can investigate in the log what went wrong

backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/salesrep/process/CreateMissingCommissionEntriesForSalesRep.java
```diff
@@ -0,0 +1,432 @@
...
+		try
+		{
+			recalculateCommissionCriteriaList.forEach(criteria -> {
+				try
+				{
+					trxManager.runInNewTrx(() -> recalculateCommissionForCriteria(criteria));
+				}
+				catch (final RuntimeException e)
+				{
+					Loggables.withLogger(logger, Level.ERROR)
```
metas-ts, line 168: Use IErrorManager#createIssue(java.lang.Throwable) to create an AD_Issue and log its AD_Issue_ID as well

metas-ts (changes requested):

backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/commissioninstance/services/CommissionInstanceService.java
```diff
@@ -83,12 +82,29 @@ public void createAndAddMissingShares(
 			return;
 		}
 
+		final Optional<HierarchyLevel> existingSharesHierarchyTopLevel = instance
+				.getShares()
+				.stream()
+				.filter(share -> share.getLevel() != null)
+				.map(SalesCommissionShare::getLevel)
+				.min(HierarchyLevel::compareTo);
```
metas-ts, line 90: shouldn't this be `max()` here?

metas-ts (approved):

Issue #10538, opened 2021-01-21
Move Haddex from Product to Partner


metas-rc:
### Is this a bug or feature request?
* FR
### What is the current behavior?
* Currently, the Haddex fields are on the Product level
#### Which are the steps to reproduce?
* See https://github.com/metasfresh/metasfresh/issues/10412
### What is the expected or desired behavior?
ToDo:
- Adjustment in Product
  - Keep the field Y/N Haddex Check `Haddex Prüfung erforderlich`.
  - drop the fields `Datum Haddex Prüfung` & `Haddex Kontroll Nr.` 

- Adjustment in Partner
  - Add the field Y/N Haddex Check `Haddex Prüfung erforderlich`. If this is set to Y then show
    - new field `Datum Haddex Prüfung` & `Haddex Kontroll Nr.`

- Adjustment in Sales Orderline
-   - in sales order don't allow the recording of products with haddex check = Y for a Partner (shipto) with and today - haddex date > sysconfig months haddex age (Base data now taken from Partner)

PR #10548, opened 2021-01-22
refactoring Haddex

metas-rc:
add haddex validations to:
* partner change
* promised date changed
* order completion

https://github.com/metasfresh/metasfresh/issues/10538

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
@@ -37,6 +37,7 @@ import de.metas.document.DocTypeId;
 import de.metas.document.DocTypeQuery;
 import de.metas.document.IDocTypeBL;
 import de.metas.document.IDocTypeDAO;
+import de.metas.i18n.AdMessageKey;
 import de.metas.i18n.IModelTranslationMap;
 import de.metas.i18n.ITranslatableString;
 import de.metas.interfaces.I_C_BPartner;
@@ -59,6 +60,7 @@ import de.metas.pricing.PricingSystemId;
 import de.metas.pricing.exceptions.PriceListNotFoundException;
 import de.metas.pricing.service.IPriceListBL;
 import de.metas.pricing.service.IPriceListDAO;
+import de.metas.product.IProductBL;
 import de.metas.product.ProductId;
 import de.metas.project.ProjectId;
 import de.metas.quantity.Quantity;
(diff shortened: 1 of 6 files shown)


backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
```diff
@@ -1051,4 +1062,77 @@ public ZoneId getTimeZone(@NonNull final I_C_Order order)
...
+		}
+
+		final org.compiere.model.I_C_BPartner partner = partnerDAO.getById(order.getC_BPartner_ID());
+
+		if (!partner.isHaddexCheck())
+		{
+			return false;
+		}
+
+		if (partner.getDateHaddexCheck() == null)
```
metas-ts, line 1121: are we sure this is what we want? what not insisten on a date if isHaddexCheck=true?

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
```diff
@@ -1051,4 +1062,77 @@ public ZoneId getTimeZone(@NonNull final I_C_Order order)
...
+
+		validateHaddexDate(order);
+	}
+
+	@Override
+	public void validateHaddexDate(final I_C_Order order)
+	{
+		final IBPartnerDAO partnerDAO = Services.get(IBPartnerDAO.class);
+		final org.compiere.model.I_C_BPartner partner = partnerDAO.getById(order.getC_BPartner_ID());
+		final long differenceBetweenHaddexCheckDateAndPromisedDateInMonths = Math.abs(
```
metas-ts, line 1094: why the `abs`? 
not 100% sure what to do if datePromised is before the haddex, check..but should it really depend on how many moths before it is?
maybe we should simply throw the error also if getDatePromised < getDateHaddexCheck ? 
(i guess it's a rare case, so maybe it doesn't make sense to engneer a dedicated error message..)

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
metas-rc: to be handled in a followup

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
metas-rc: to be handled in a followup

Issue #10560, opened 2021-01-27
Prevent Reverting Inventories if HU was already transformed


BodorAnnamaria:
### Is this a bug or feature request?
FR
### What is the current behavior?
Create a Physical Inventory with items.
Go to Handling Unit editor and select the Item from the inventory and select the Transform action.
After `Transform` action the Physical Inventory can be reversed and not all qty will be reversed.
### What is the expected or desired behavior?
Create a Physical Inventory with items.
Go to Handling Unit editor and select the Item from the inventory and select the Transform action.
After `Transform` action the Physical Inventory should not be reversed any more.

PR #10562, opened 2021-01-28
#6650 Prevent Reverting Inventories if HU was already transformed

BodorAnnamaria:
https://github.com/metasfresh/metasfresh/issues/10560

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/inventory/interceptor/M_Inventory.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/inventory/interceptor/M_Inventory.java
@@ -1,7 +1,7 @@
 package de.metas.handlingunits.inventory.interceptor;
 
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableMap;
 import de.metas.document.engine.IDocumentBL;
 import de.metas.handlingunits.HuId;
 import de.metas.handlingunits.exceptions.HUException;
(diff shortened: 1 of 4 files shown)


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/inventory/interceptor/M_Inventory.java
```diff
@@ -83,11 +88,26 @@ public void beforeComplete(final I_M_Inventory inventoryRecord)
 		inventoryLineRecordService.syncToHUs(inventoryRecord);
 	}
 
+	@DocValidate(timings = ModelValidator.TIMING_BEFORE_REVERSECORRECT)
+	public void checkHUTransformationBeforeReverseCorrect(final I_M_Inventory inventory)
+	{
+		final InventoryId inventoryId = InventoryId.ofRepoId(inventory.getM_Inventory_ID());
+		final ImmutableMap<InventoryLineId, HuId> inventoryHuIds =
+				inventoryDAO.retrieveLinesForInventoryId(inventoryId, I_M_InventoryLine.class)
```
metas-ts, line 96: M_InventoryLines lines might *or* might not directly reference an M_HU.
One line might have many HUs (via `M_InventoryLine_HU`).

As written today:
> that depends. but i think the repository method will give you the inventory's HUs, no matter if they are linked via M_InventoryLine_HU or M_InventoryLine

please don't ignore such remarks, even if i might be wrong and the repository method might not work as i wrote.
(i think that the code i had in mind is here: `de.metas.handlingunits.inventory.InventoryRepository#toInventoryLine(de.metas.handlingunits.model.I_M_InventoryLine, java.util.Collection<de.metas.handlingunits.model.I_M_InventoryLine_HU>)`)

metas-ts (changes requested):

metas-ts (approved):

metas-ts (approved):

Issue #10571, opened 2021-02-01
 Pricesystem of contract term not considered correctly


BodorAnnamaria:
### Is this a bug or feature request?
BUG
### What is the current behavior?
Create a Contact Term with a price1 for the product A.
The same product A has price2 for Pricing System.
Create a Sales Order with a pricing system.
Add a product A in the order line should take the price2.
Set contract term for order line and price should change to price1 but is not changing.
### What is the expected or desired behavior?
When setting the contact term on order line the price defined for the contract term should be displayed.

PR #10572, opened 2021-02-01
#6513 Pricesystem of contract term not considered correctly

BodorAnnamaria:
https://github.com/metasfresh/metasfresh/issues/10571


metas-ts (approved):

metas-ts (approved):
i think with this it should be OK now

PR #10616, opened 2021-02-09
#6513 Pricesystem of contract term not considered correctly

BodorAnnamaria:
https://github.com/metasfresh/metasfresh/issues/10571

fix Menge in Preiseinheit field calculation

### DIFF
--- a/backend/de.metas.contracts/src/main/java/de/metas/contracts/subscription/callout/C_OrderLine.java
+++ b/backend/de.metas.contracts/src/main/java/de/metas/contracts/subscription/callout/C_OrderLine.java
@@ -79,6 +79,7 @@ public class C_OrderLine
 		{
 			final BigDecimal qtyOrdered = orderLineBL.convertQtyEnteredToStockUOM(ol).toBigDecimal();
 			ol.setQtyOrdered(qtyOrdered);
+			ol.setQtyEnteredInPriceUOM(qtyOrdered);
 
 			orderLineBL.updatePrices(OrderLinePriceUpdateRequest.builder()
 					.orderLine(ol)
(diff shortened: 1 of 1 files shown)


backend/de.metas.contracts/src/main/java/de/metas/contracts/subscription/callout/C_OrderLine.java
```diff
@@ -79,6 +79,7 @@ public void onFlatrateConditions(final I_C_OrderLine ol, final ICalloutField fie
 		{
 			final BigDecimal qtyOrdered = orderLineBL.convertQtyEnteredToStockUOM(ol).toBigDecimal();
 			ol.setQtyOrdered(qtyOrdered);
+			ol.setQtyEnteredInPriceUOM(qtyOrdered);
```
metas-ts, line 82:
```suggestion
			final BigDecimal qtyEnteredInPriceUOM = orderLineBL.convertQtyEnteredToPriceUOM(ol).toBigDecimal();
			ol.setQtyEnteredInPriceUOM(qtyEnteredInPriceUOM);
```

background
* qtyOrdered is in the product's stock unit-of-measurement ("UOM", column `M_Product.C_UOM_ID`)
* QtyEnteredInPriceUOM needs to be in the price UOM (`C_OrderLine.Price_UOM_ID`)

metas-ts (changes requested):

backend/de.metas.contracts/src/main/java/de/metas/contracts/subscription/callout/C_OrderLine.java
metas-ts: PS: please make sure to look over your PR before devtesting etc, as suggested by the R&I template. 
In that case you might have noticed the discrepancy in column-names between `qtyOrdered` / `convertQtyEnteredToStockUOM` on one side and `qtyEnteredInPriceUOM` on the oter side.

metas-ts (approved):

PR #10621, opened 2021-02-09
Gh10599 allow zero in allocation posting

metas-rc:
https://github.com/metasfresh/metasfresh/issues/10599


metas-ts (approved):
great, looks like the fix you already talked me through last time. so i don't need to come back with question marks

PR #10625, opened 2021-02-10
Gh10624 doc outbound log selection printing

metas-rc:
https://github.com/metasfresh/metasfresh/issues/10624
(please, don't integrate yet)


backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/archive/api/impl/ArchiveDAO.java
```diff
@@ -35,15 +35,21 @@
 import org.adempiere.ad.trx.api.ITrx;
 import org.adempiere.archive.ArchiveId;
 import org.adempiere.archive.api.IArchiveDAO;
+import org.adempiere.exceptions.AdempiereException;
 import org.adempiere.model.InterfaceWrapperHelper;
 import org.adempiere.util.lang.ITableRecordReference;
+import org.apache.poi.ss.formula.functions.T;
```
metas-ts, line 41: this was *most* probably not intended :-)

backend/de.metas.printing/de.metas.printing.base/src/main/java/de/metas/printing/api/impl/PrintArchiveParameters.java
```diff
@@ -0,0 +1,50 @@
...
+import org.adempiere.archive.ArchiveId;
+
+import javax.annotation.Nullable;
+
+@Value
+@Builder
+
+public class PrintArchiveParameters
+{
+	@NonNull
```
metas-ts, line 41: not critical, but with `@Value` you don't need private and final

metas-ts (approved):
please fix the 1-2 minor things (at any rate, the `T`). then it's good to integrate

PR #10651, opened 2021-02-16
Don't show doc actions when they shouldn't be visible

siemiatj:
Looks like this was disabled when I did some quick changes to unblock Petrica on his inline tab widget effort. So this looks like a temporary thing that should be reverted. 

https://github.com/metasfresh/metasfresh/pull/10301/commits/d97d319b84ba30256c7f73d94f2371114d7d6664#diff-1cfcb26643fbe64684980f1488db853ee0b5abe32b52125ee175f6e9054bd87aL55

@petrican where can the inline tab be tested to be sure it's not affected ?


petrican:
> Looks like this was disabled when I did some quick changes to unblock Petrica on his inline tab widget effort. So this looks like a temporary thing that should be reverted.
> 
> [d97d319#diff-1cfcb26643fbe64684980f1488db853ee0b5abe32b52125ee175f6e9054bd87aL55](https://github.com/metasfresh/metasfresh/commit/d97d319b84ba30256c7f73d94f2371114d7d6664#diff-1cfcb26643fbe64684980f1488db853ee0b5abe32b52125ee175f6e9054bd87aL55)
> 
> @petrican where can the inline tab be tested to be sure it's not affected ?

Will check. I see the inlinetab in action on hevuat box but I am having some cors trouble so need to overcome that in order to test.

petrican:
I see some issues with the build. Will merge the master into it and see if the issue goes away. As I understood from Tobi that should fix it

petrican (approved):
Looks ok. I tested also the inlinetab and I did not spot any issues. Thanks

petrican:
If the build is successful it can be merged. Thanks

PR #10679, opened 2021-02-19
Update babel packages

siemiatj:
Move us to the latest version of babel transpiler and eslint. Also update `axios` because of a critical security fix.


petrican (approved):
lgtm. Thanks

Issue #10667, opened 2021-02-18
Create AD_Zebra_Config Table


BodorAnnamaria:
### Is this a bug or feature request?
FR
### What is the current behavior?
SSCC-Labels print can not be customized for a Customer
### What is the expected or desired behavior?
Create AD_Zebra_Config which will allow to customize SSCC-Labels print for Customer

Issue #6785, opened 2020-06-03
`Zoom into` not shown in table's context menu


siemiatj:
### Is this a bug or feature request?

Bug

### What is the current behavior?

There's no `zoom into` option in the table context menu

#### Which are the steps to reproduce?

Right click table row in sales order

### What is the expected or desired behavior?

Bring it back

PR #10680, opened 2021-02-19
#6785 SSCC-Labels für Pistor

BodorAnnamaria:
https://github.com/metasfresh/metasfresh/issues/10667


backend/de.metas.edi/src/main/java/de/metas/edi/api/ZebraConfigRepository.java
```diff
@@ -0,0 +1,90 @@
...
+
+	@Nullable
+	public I_AD_Zebra_Config getZebraConfigByIdOrDefault(@Nullable final ZebraConfigId zebraConfigId)
+	{
+		I_AD_Zebra_Config item = null;
+		if (zebraConfigId == null)
+		{
+			item = queryBL
+					.createQueryBuilderOutOfTrx(I_AD_Zebra_Config.class)
+					.addEqualsFilter(I_AD_Zebra_Config.COLUMNNAME_IsDefault, true)
```
metas-ts, line 54: in case you don't have yet, please add a unique constraint (in the "table and column" window) to make sure there there can be just one default config among the confils with IsActive='Y'

backend/de.metas.edi/src/main/java/de/metas/edi/process/EDI_GenerateCSV_FileForSSCC_Labels.java
```diff
@@ -31,30 +35,69 @@
...
+			final List<Integer> zebraConfigIds = queryBL
+					.createQueryBuilder(I_EDI_Desadv.class)
+					.filter(selectedRecordsFilter)
+					.andCollect(I_EDI_Desadv.COLUMNNAME_C_BPartner_ID, I_C_BPartner.class)
+					.andCollectChildren(I_C_BP_PrintFormat.COLUMNNAME_C_BPartner_ID, I_C_BP_PrintFormat.class)
+					.andCollect(I_C_BP_PrintFormat.COLUMNNAME_AD_Zebra_Config_ID, I_AD_Zebra_Config.class)
+					.create()
+					.listDistinct(I_AD_Zebra_Config.COLUMNNAME_AD_Zebra_Config_ID, Integer.class);
+
+			if (formatLinesWithNoZebraConfig.size() > 0 && zebraConfigIds.size() > 0)
```
metas-ts, line 80: Not sure I understand the purpose of `formatLinesWithNoZebraConfig`..i might guess the idea..but then, what about bpartners that have not even a `C_BP_PrintFormat`

backend/de.metas.edi/src/main/java/de/metas/edi/process/EDI_GenerateCSV_FileForSSCC_Labels.java
```diff
@@ -31,30 +35,69 @@
...
+			}
+		}
 		return ProcessPreconditionsResolution.accept();
 	}
 
 	void generateCSV_FileForSSCC_Labels(final Collection<Integer> desadvLinePackIDsToPrint)
 	{
 		final ZebraPrinterService zebraPrinterService = SpringContextHolder.instance.getBean(ZebraPrinterService.class);
-
+		final ZebraConfigRepository zebraConfigRepository = SpringContextHolder.instance.getBean(ZebraConfigRepository.class);
```
metas-ts, line 96: please have these as fields (member variables) in this class

metas-ts (changes requested):

backend/de.metas.edi/src/main/java/de/metas/edi/process/EDI_GenerateCSV_FileForSSCC_Labels.java
metas-rc: Idea: 
* Add a new method ``retrieveZebraConfigId`` in IBPartnerDAO. Will return the ID of the zebra config from the bp print format if it exists and default otherwise.
* modify the precondition like this: 

```java
	queryBL
				.createQueryBuilder(I_EDI_Desadv.class)
				.filter(queryFilter)
				.create()
				.list()
				.stream()
				.map(ediDesadv -> BPartnerId.ofRepoId(ediDesadv.getC_BPartner_ID()))
				.map(bpartnerId -> bpartnerDAO.retrieveZebraConfigId(bpartnerId))
				.collect(Collectors.toSet())
				.size();
```

=> size must be 1 for acceptance.

BodorAnnamaria: I added the method `retrieveZebraConfigId  `in `ZebraConfigRepository` I think the logic belongs to the zebraConfig and should be there

metas-rc: You are right.

backend/de.metas.edi/src/main/java/de/metas/edi/api/ZebraConfigRepository.java
```diff
@@ -0,0 +1,101 @@
...
+
+		if (defaultZebraConfigId == null)
+		{
+			throw new AdempiereException(msgBL.getTranslatableMsgText(MSG_NO_ZEBRA_CONFIG));
+		}
+
+		return defaultZebraConfigId;
+	}
+
+	public ZebraConfigId getZebraConfigForDesadvLinePackID(@NonNull final Integer desadvLinePackID)
```
metas-ts, line 72: AFAIU this method is not used anymore, right? therefore please remove it, as it's alo not working properly

metas-ts (approved):

backend/de.metas.edi/src/main/java/de/metas/edi/api/ZebraConfigRepository.java
BodorAnnamaria: the method was use. I refactored the code to be more clear

metas-ts (approved):

PR #10692, opened 2021-02-23
Fix builds

petrican:
Fix for the builds due to deprecated package https://github.com/metasfresh/me03/issues/7165


metas-ts (approved):
@siemiatj feel free to weigh in, but i would say, if it builds, it's fine :-)

note that also the "yarn lint" call is active in this branch

PR #10695, opened 2021-02-23
Fixes for material cockpit

metas-ts:
Fixes
* when extracting a date from a material event for the cockpit, then *do not* assume timezone=UTC. It leads to your time being converted to <the-previous-day> 23:00 and from there to <the-previous-day> 00:00
* be lenient if user switched off all Attributes in material cockpit and there is also no no, reservations etc. instead, just go with an "empty" main row that has no subrows


metas-ts (commented):
lgtm. but very hard to test locally; will integrated and deploy&check directly on customer UAT machine

Issue #10691, opened 2021-02-22
Create Window External system log


BodorAnnamaria:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
Create a new Window `External system log` based on a view
```SQL
CREATE VIEW ExternalSystem_Config_PInstance_Log_v AS
SELECT pil.ad_pinstance_id AS ExternalSystem_Config_PInstance_Log_v_ID,
       pip_request.info    AS External_Request, /*please change name to be like the process param*/
       pil.p_msg,
       pil.created,
       pi.ad_client_id,
       pi.record_id as ExternalSystem_Config_ID,
       pi.ad_org_id,
       pi.ad_pinstance_id,
       pi.isprocessing,
       pi.errormsg
FROM ad_pinstance pi
         JOIN ad_pinstance_para pip_request ON pi.ad_pinstance_id = pip_request.ad_pinstance_id AND pip_request.parametername = 'External_Request'
         JOIN ad_pinstance_log pil ON pil.ad_pinstance_id = pi.ad_pinstance_id
WHERE pi.AD_Table_ID = get_table_id('ExternalSystem_Config');
```

In Menu will be displayed near the `external system `window

PR #10706, opened 2021-02-24
Fenster für externe System-Meldungen

BodorAnnamaria:
https://github.com/metasfresh/metasfresh/issues/10691

- create index ad_pinstance_log_ad_pinstance_id_created
- create index ad_pinstance_para_perf
- create db function cast_to_numeric_or_null
- create view ExternalSystem_Config_PInstance_Log_v
- create Window External System Log


metas-ts (approved):
i moved files and updated then EntitiyTypes. with that - if it still builds - lgtm

PR #10716, opened 2021-02-26
Fix duplicated nested rows after updating a value

siemiatj:
Related to #10715


petrican (approved):
Looks lgtm. This should be tested roughly still to make sure it does not introduce other issues. Thanks

petrican:
Pls add this to the https://github.com/metasfresh/me03/issues/7209

Issue #10718, opened 2021-02-26
Add Frame Agreement and Order Call SO doc types


adi-stefan:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
`Frame Agreement` and `Order Call` can be selected as doc types for a sales order

PR #10719, opened 2021-02-26
#gh10718 added Frame Agreement and Order Call doctypes for SO

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/10718


backend/de.metas.acct.base/src/main/sql/postgresql/ddl/views/c_orderline_id_with_missing_shipmentschedule_v.sql
```diff
@@ -0,0 +1,23 @@
+CREATE OR REPLACE VIEW c_orderline_id_with_missing_shipmentschedule_v AS
```
metas-ts, line 1: please update the one we have (`backend/de.metas.swat/de.metas.swat.base/src/main/sql/postgresql/ddl/views/C_OrderLine_ID_With_Missing_ShipmentSchedule_v.sql`)

backend/de.metas.business/src/main/java-legacy/org/compiere/model/MOrder.java
```diff
@@ -1600,7 +1600,8 @@ private DocStatus completeIt0()
 		//
 		// Offers
 		if (X_C_DocType.DOCSUBTYPE_Proposal.equals(docSubType)
-				|| X_C_DocType.DOCSUBTYPE_Quotation.equals(docSubType))
+				|| X_C_DocType.DOCSUBTYPE_Quotation.equals(docSubType)
+				|| X_C_DocType.DOCSUBTYPE_Rahmenvertrag.equals(docSubType))
```
metas-ts, line 1604: please chreate the new doctypes such that we have here there EN names, (like "Quotation")

metas-ts (changes requested):

backend/de.metas.business/src/main/java-legacy/org/compiere/model/MOrder.java
metas-ts: So, do you have in here the SQL to deactivate the doc-types? I didn't see it, but in this large SQL file i might have overlooked it

adi-stefan: Yep, I've changed the script to create the docTypes with IsActive='N'

backend/de.metas.acct.base/src/main/sql/postgresql/ddl/views/c_orderline_id_with_missing_shipmentschedule_v.sql
adi-stefan: Missed that. Fixed.

backend/de.metas.business/src/main/java-legacy/org/compiere/model/MOrder.java
adi-stefan: >please chreate the new doctypes such that we have here there EN names, (like "Quotation")

Done.

metas-ts (approved):

Issue #10648, opened 2021-02-16
Add guarantee time in months


BodorAnnamaria:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
Create a new column `GuaranteeMonths ` in `m_product ` table which hold the product guarantee in months. The field should be displayed in Product window and hide the field `min guarantee days`  from the window.

PR #10720, opened 2021-02-26
#6977 Garantie bei Produkten 

BodorAnnamaria:
https://github.com/metasfresh/metasfresh/issues/10648

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/product/IProductDAO.java
+++ b/backend/de.metas.business/src/main/java/de/metas/product/IProductDAO.java
@@ -173,5 +173,5 @@ public interface IProductDAO extends ISingletonService
 
 	int getProductGuaranteeDaysMinFallbackProductCategory(@NonNull final ProductId productId);
 
-	int getGuaranteeMonthsInDays(I_M_Product product);
+	int getGuaranteeMonthsInDays(ProductId productId);
 }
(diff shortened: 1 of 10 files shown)


backend/de.metas.business/src/main/java/de/metas/product/IProductDAO.java
```diff
@@ -173,4 +173,5 @@ private ProductQuery(
 
 	int getProductGuaranteeDaysMinFallbackProductCategory(@NonNull final ProductId productId);
 
+	int getGuaranteeMonthsInDays(I_M_Product product);
```
metas-ts, line 176: avoid using data records like I_M_Product, I_C_Order etc outside DAOs and Repositories. Instead, please use (in this case) ProductId as parameter

backend/de.metas.business/src/main/java/de/metas/product/impl/ProductDAO.java
```diff
@@ -509,11 +515,28 @@ public int getProductGuaranteeDaysMinFallbackProductCategory(final @NonNull Prod
 		{
 			return productRecord.getGuaranteeDaysMin();
 		}
+		else if (productRecord.getGuaranteeMonths() != null && !productRecord.getGuaranteeMonths().isEmpty()) {
```
metas-ts, line 518: pls use de.metas.util.Check#isNotBlank instead

metas-ts (changes requested):
lgtm, besides two small (i think) changes

metas-ts (approved):

PR #10735, opened 2021-03-02
Fix auto-collapsing of updated rows

siemiatj:
Related to #10715 

Also fixes eslint.


petrican (approved):
Lgtm. Regarding the linter: There are some warnings that we will need to fix from the linter (these popup after the last update of the eslint parser). Mentioned this in here such that we would consider it while we fix other issues.
Screenshot 2021-03-02 at 08 11 12

PR #10742, opened 2021-03-02
Gh10734 Manufacturing:  Issue/Receipt  improvements

metas-rc:
https://github.com/metasfresh/metasfresh/issues/10734

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/allocation/transfer/HUTransformService.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/allocation/transfer/HUTransformService.java
@@ -70,6 +70,7 @@ import de.metas.quantity.Capacity;
 import de.metas.quantity.Quantity;
 import de.metas.quantity.StockQtyAndUOMQty;
 import de.metas.quantity.StockQtyAndUOMQtys;
+import de.metas.uom.UomId;
 import de.metas.util.Check;
 import de.metas.util.GuavaCollectors;
 import de.metas.util.Services;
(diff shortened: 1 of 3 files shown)


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/allocation/transfer/HUTransformService.java
```diff
@@ -240,7 +241,14 @@ public Quantity getMaximumQtyCU(@NonNull final I_M_HU cu, @NonNull final I_C_UOM
...
 
-		if (qtyCuExceedsCuHU && huIsCU)
+		final IHUStorageFactory storageFactory = handlingUnitsBL.getStorageFactory();
+		final IHUStorage storage = storageFactory.getStorage(cuOrAggregateHU);
+
+		final I_C_UOM huUOM = storage.getC_UOMOrNull();
+		final UomId huUOMId = UomId.ofRepoId(huUOM == null ? -1 : huUOM.getC_UOM_ID());
+		final boolean isSameUOM = qtyCU.getUomId().equals(huUOMId);
+
+		if (qtyCuExceedsCuHU && huIsCU && isSameUOM)
```
metas-ts, line 251: shouldn't we instead see if we can convert?

metas-rc: If this condition is fulfilled, the source HU will be used as it is. Otherwise the HU is transformed and will fit the UOM of the qtyCU (parameter)  which is exactly what we want: to have all the HUs under the bom line in the same UOM as the bom line.

Issue #10746, opened 2021-03-02
Fix reloading list view on browser `back`


siemiatj:
### Is this a bug or feature request?

Bug

### What is the current behavior?

View doesn't refresh when `back` is clicked.

#### Which are the steps to reproduce?

Apply a filter and then use `back`

PR #10747, opened 2021-03-02
Fix back button

siemiatj:
Related to #10746


petrican (changes requested):
The code is indeed fixing the issue but there is a problem with the build. You have a failing test that is braking it. Pls fix this. Thanks

petrican (approved):
lgtm. Thanks

PR #10768, opened 2021-03-05
10757 token auth

petrican:
https://www.loom.com/share/0638e151837e41efa181a72dc84979e1

Adapt the MF app Frontend to allow also authentication with token

### DIFF
--- a/frontend/src/api/index.js
+++ b/frontend/src/api/index.js
@@ -30,6 +30,7 @@ import {
 import {
   loginRequest,
   localLoginRequest,
+  loginWithToken,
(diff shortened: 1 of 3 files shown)


siemiatj (approved):
Maybe we could use that in cypress tests ? Just generate a token before running the tests and then use that for the duration of the test session

petrican:
> Maybe we could use that in cypress tests ? Just generate a token before running the tests and then use that for the duration of the test session

Brilliant idea! But how can we programatically generate that token ? I think I will open a mf15 issue for that

PR #10772, opened 2021-03-05
improvements while checking code to prepare a concept:

metas-ts:
- ITaxBL now returns TaxId instead of int
- avoid NPE in MOrderLine


metas-ts (commented):
lgtm

Issue #10748, opened 2021-03-03
Improve and extend copy discount lines on product


cristinamghita:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
* Copy only the discounts that do not exist. For the other update the values.
* Create a new process that will copy the discount using selected lines as target and source given as params

PR #10773, opened 2021-03-08
#10748 Improve and extend copy discount lines on product 

cristinamghita:
#10748 Improve and extend copy discount lines on product


metas-ts (approved):
I'm not really happy with adding even more BL to PricingConditionsRepository but i think at this point in this issue it's too late to use the issue for an overhaul.
So, fine with me. Let's try to think of it next time, when we make&review the concept for such an issue.

PR #10781, opened 2021-03-08
introduce app-Server API v1 and v2-pre

metas-ts:
Note: 
* v1 shall remain as it is
* v2-pre shall be renamed to v2 after having finished the changes that are breaking from v1's POV
* after that, v2 shall only be extended with non-breaking changes (i.e. not-mandatory extensions)


metas-ts (commented):
lgtm

Issue #10765, opened 2021-03-05
Invalid viewId: null on payment allocations


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
Invalid viewId: null on payment allocations requests
#### Which are the steps to reproduce?
Payment allocations
### What is the expected or desired behavior?
xhr should not fail and should have correct view in the path

PR #10787, opened 2021-03-09
Handle quickactions requests from websockets with modal correctly

siemiatj:
Related to #10765


petrican (commented):
Looks lgtm. Only one single issue: The target branch should be `scary_plum_uat`.

Issue #10801, opened 2021-03-12
Inline filter is broken


siemiatj:
### Is this a bug or feature request?

Bug

### What is the current behavior?

![Screenshot 2021-03-12 at 01 18 06](https://user-images.githubusercontent.com/513460/110881373-8d4f4600-82e0-11eb-8c92-ab9c218e8170.png)

PR #10802, opened 2021-03-12
Fix inline filters

siemiatj:
This is a fix for the release but I want to look deeper into that and check if we shouldn't add some more guards, as most of the time we assume everywhere there's only one active filter, or a filter has single parameter.

Related to #10801


petrican (approved):
lgtm. Thanks

Issue #10832, opened 2021-03-17
Possibel "Duplicate Key Error" when loading records with data entry tabs


metas-ts:
### Is this a bug or feature request?
B

PR #10833, opened 2021-03-17
Avoid assumption on whether data-entry data exists

metas-ts:
https://github.com/metasfresh/metasfresh/issues/10832


metas-ts (commented):
lgtm - going to integrate after successful build

PR #10848, opened 2021-03-18
Return even inactive C_BPartner_Locations at many places

metas-ts:
Because e.g. if you have a completed invoice and someone deactivated the bpartner-location that the invoice refers to, then the invoice should still be able to e.g. get the location's country.

=> In the end, after checking the call hierachy of `getBPartnerLocationById` i think that actually in every case, there should *not* be a `null` returned, even if the C_BPartner_Location was meanwhile deactivated.

### DIFF
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/IBPartnerDAO.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/IBPartnerDAO.java
@@ -115,7 +115,7 @@ public interface IBPartnerDAO extends ISingletonService
 
 	/**
 	 * @deprecated in all cases i can imagine, if the caller has a {@code bpartnerLocationId}, they need the actual record, even if it is inactive.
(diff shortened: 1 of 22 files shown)


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/IBPartnerDAO.java
```diff
@@ -113,13 +113,18 @@
 
 	Optional<BPartnerLocationId> getBPartnerLocationIdByGln(BPartnerId bpartnerId, GLN gln);
 
+	/**
+	 * @deprecated in all cases i can imagine, if the caller has a {@code bpartnerLocationId}, they need the actual record, even if it is inactive.
+	 * Think e.g. of a completed shipment. Therefore, please consder using {@link #getBPartnerLocationByIdEvenInactive(BPartnerLocationId)} instead.
```
metas-ts, line 118:
```suggestion
	 * Think e.g. of a completed shipment. Therefore, please consider using {@link #getBPartnerLocationByIdEvenInactive(BPartnerLocationId)} instead.
```

PR #10850, opened 2021-03-19
When creating orders from OLcands and when settings sales rep in order

metas-ts:
creating sales orders: don't set salesrep if equal to buyer
setting salesrep in orders: don't update C_BPArtner masterdata if salesrep equal to buyer


metas-ts (commented):
lgtm

Issue #10864, opened 2021-03-22
Clear filter after searching an item (not in db) seems it cannot be done


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
Search for something, clear, search again, clear -> clear freezes
#### Which are the steps to reproduce?
Search for something, clear, search again, clear -> clear freezes
### What is the expected or desired behavior?
Clear should work smoothly

PR #10865, opened 2021-03-22
Fix subsequent filters clearing

petrican:
https://github.com/metasfresh/metasfresh/issues/10864

https://www.loom.com/share/12198f3642984c06b4218d1c42f6d3fa

https://www.loom.com/share/09f5a162464c4b7ea70baa4f65479f51

### DIFF
--- a/frontend/src/containers/DocumentList.js
+++ b/frontend/src/containers/DocumentList.js
@@ -638,10 +638,13 @@ class DocumentListContainer extends Component {
 
   /**
    * @method resetInitialFilters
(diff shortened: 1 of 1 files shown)


siemiatj (changes requested):
I think we should actually do something like this there:

```
    if (!parameterName) {
      initialValuesNulled.delete(filterId);
    } else {
      initialValuesNulled.set(filterId, filterParams);
    }
```

as assigning the result of map action to a value makes no sense

siemiatj (approved):

Issue #10871, opened 2021-03-23
webui frontend: view not refreshed when filtered with an inline filter 


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
webui frontend: view not refreshed when filtered with an inline filter 
#### Which are the steps to reproduce?
window/541014/1000015

PR #10877, opened 2021-03-24
Fix duplicated requests on changing filters

siemiatj:
Also pulled some updated code that fixes errors in the console.

Related to #10871


petrican (approved):
lgtm

PR #10884, opened 2021-03-24
Master api bpartner v2 tmp

cp-ps:


backend/de.metas.externalreference/src/main/java/de/metas/externalreference/bpartner/BPartnerValueExternalReferenceType.java
```diff
@@ -0,0 +1,50 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import org.adempiere.exceptions.AdempiereException;
+import org.compiere.model.I_C_BPartner;
+
+import static de.metas.externalreference.model.X_S_ExternalReference.TYPE_BPartnerValue;
+
+@AllArgsConstructor
+@Getter
+public enum BPartnerValueExternalReferenceType implements IExternalReferenceType
```
metas-ts, line 35: why not have this enum-value in BPartnerExternalReferenceType?

backend/de.metas.externalreference/src/main/java/de/metas/externalreference/ExternalIdentifier.java
```diff
@@ -35,6 +36,8 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import static de.metas.util.Check.isEmpty;
+
 @Value
 public class ExternalIdentifier
```
metas-ts, line 42: please clarify the relation between this class and IdentifierString. If this class for v2 and identifierstring for v1?
but i found (just) one usage of IdentifierString in `de.metas.rest_api.v2.bpartner.BpartnerRestControllerTest#createOrUpdateBankAccount_create`

metas-ts (changes requested):

backend/de.metas.externalreference/src/main/java/de/metas/externalreference/bpartner/BPartnerValueExternalReferenceType.java
cif-ps: I will do the changes.

backend/de.metas.externalreference/src/main/java/de/metas/externalreference/ExternalIdentifier.java
cif-ps: I missed that, will fix it.

metas-ts (approved):

Issue #10890, opened 2021-03-25
Adapt FE formatting of the highlighted string with the order provided by the BE


petrican:
### Is this a bug or feature request?
F
### What is the current behavior?
Inlinetab headline order is hardcoded
#### Which are the steps to reproduce?
check inline tab
### What is the expected or desired behavior?
Order should be dynamic in sync with what the BE provides

PR #10894, opened 2021-03-26
InlineTab Header caption dynamic

petrican:
https://www.loom.com/share/5b15a549f21948a29d66e2d85f8ad1fa

https://github.com/metasfresh/metasfresh/issues/10890


frontend/src/utils/locale.js
```diff
@@ -24,3 +24,13 @@ export function setCurrentActiveLocale(lang) {
...
 }
+
+/**
+ * @method isGermanLanguage
+ * @summary Returns boolean value if the language is german or not
+ */
+export function isGermanLanguage(languageObj) {
+  return languageObj && languageObj.key
+    ? languageObj.key.includes('de')
+    : false;
```
teosarca, line 35: why not using the `isGermanLanguage` into Table.js since u took the code from there?

image

teosarca (changes requested):
check my comments

frontend/src/utils/locale.js
petrican: Screenshot 2021-03-27 at 07 06 18
But that is exactly what I did. You are looking somewhere in your branch and not in the actual PR

Issue #10897, opened 2021-03-26
Prevent setting Sales rep with the same value as bpartner


adi-stefan:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
Prevent setting C_BPartner_SalesRep_ID with the same value as C_BPartner_ID on `C_Invoice`, `C_OLCand`,`C_Order` and `C_BPartner`.

PR #10901, opened 2021-03-29
gh10897

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/10897

### DIFF
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/impl/BPartnerDAO.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/impl/BPartnerDAO.java
@@ -123,6 +123,7 @@ import static org.adempiere.model.InterfaceWrapperHelper.loadOutOfTrx;
 
 public class BPartnerDAO implements IBPartnerDAO
 {
+	private static final AdMessageKey MSG_SALES_REP_EQUALS_BPARTNER = AdMessageKey.of("SALES_REP_EQUALS_BPARTNER");
 	private static final Logger logger = LogManager.getLogger(BPartnerDAO.class);
 	private final IQueryBL queryBL = Services.get(IQueryBL.class);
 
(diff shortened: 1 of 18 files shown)


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/impl/BPartnerDAO.java
```diff
@@ -1711,4 +1712,13 @@ private static BPartnerPrintFormat toBPartnerPrintFormat(final I_C_BP_PrintForma
 				.printFormatId(PrintFormatId.ofRepoId(record.getAD_PrintFormat_ID()))
 				.build();
 	}
+
+	@Override
+	public void validateSalesRep(@NonNull final BPartnerId bPartnerId, @Nullable final BPartnerId salesRepId)
```
metas-ts, line 1717: this method doesn't access any data, so i think it's better up in BPartnerBL

backend/de.metas.contracts/src/test/java/de/metas/contracts/commission/salesrep/InvoiceTest.java
```diff
@@ -0,0 +1,71 @@
...
+		Assertions.assertNotEquals(0, invoice.getC_BPartner_SalesRep_ID());
+		save(invoice);
+
+		Assertions.assertEquals(0, invoice.getC_BPartner_SalesRep_ID());
+
+	}
+
+	private I_C_BPartner createPartner()
+	{
+		final I_C_BPartner partner = newInstance(I_C_BPartner.class);
```
metas-ts, line 64: FYI, you can also use `BusinessTestHelper.createBPartner();` (the dependency is already in the contracts-pom.xml)

metas-ts (changes requested):

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/impl/BPartnerDAO.java
adi-stefan: I wrote it like that, but BPartnerBL does not have a no-args constructor for it. Had a lot of problems writing unit-tests for it and sort of gave up. since it's not that big of a logic, it can also be in a Utility class. But could not find a good match for it.

metas-ts: but you have `BPartnerSalesRepTest` in this very PR, where you actually work with BPartnerBL. 
So it looks like you mastered this art after all.
A weel thought-through utility class would be still better, but for starts i think you can just move the code..

backend/de.metas.contracts/src/test/java/de/metas/contracts/commission/salesrep/InvoiceTest.java
adi-stefan: Changed to use `BusinessTestHelper.createBPartner();`

metas-ts (approved):

Issue #10841, opened 2021-03-18
add PurchaseOrderCandidate REST controller


adi-stefan:
### Is this a bug or feature request?
FR

PR #10907, opened 2021-03-30
gh10841_2

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/10841


backend/de.metas.business/src/main/java/de/metas/order/impl/OrderLinePriceCalculator.java
```diff
@@ -504,16 +502,15 @@ private Percent extractDiscount(final IPricingResult pricingResult, final SOTrx
...
-		{
-			return true;
-		}
-
 		final I_C_OrderLine orderLine = request.getOrderLine();
 		if (orderLine.isManualDiscount())
 		{
 			return false;
 		}
+		if (soTrx.isPurchase())
```
metas-ts, line 510: i wonder if this shouldn't be after `request.isUpdatePriceEnteredAndDiscountOnlyIfNotAlreadySet()` :thinking:

metas-ts (approved):

Issue #10925, opened 2021-04-02
If url contains a viewId, refreshing after login in another tab causes white screen (after timeout logout)


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
If url contains a viewId, refreshing after login in another tab causes white screen (after timeout logout)
#### Which are the steps to reproduce?
two tabs, logout in one
refresh the tab with the viewId: white page, NOK: i would expect the page to be refreshed or, if not possible, to get back to dashboard
### What is the expected or desired behavior?
Should redirect to the dashboard/login screen

PR #10926, opened 2021-04-02
Add safety checks for immer original

petrican:
In the latest immer version the original method doesn't allow passing undefined.

How I tested this on my local to make sure it works fine

https://www.loom.com/share/66dfe2381849489ca085b9cc34ac2713


metas-Kay:
related to: https://github.com/metasfresh/metasfresh/issues/10925

PR #10938, opened 2021-04-06
#10933 Extend process de.metas.ui.web.pricing.process.M_DiscountSchemaBreak_CopyToSelectedSchema_Product 

cristinamghita:


backend/metasfresh-dist/metasfresh-webui-api-remote-via-ssh-tunnel.launch
```diff
@@ -1,19 +1,19 @@
...
+<booleanAttribute key="org.eclipse.jdt.launching.ALLOW_TERMINATE" value="false"/>
+<mapAttribute key="org.eclipse.jdt.launching.CONNECT_MAP">
+<mapEntry key="hostname" value="localhost"/>
+<mapEntry key="port" value="8789"/>
```
metas-ts, line 15: our current convention is 18789, so i think you should rather update your ssh-config than the launch config

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/pricing/process/M_DiscountSchemaBreak_CopyToSelectedSchema_Product.java
```diff
@@ -111,13 +100,25 @@ protected String doIt()
 			pricingConditions.add(PricingConditionsId.ofRepoId(record.getM_DiscountSchema_ID()));
 
 		});
-		
+
+		// run copy for each product separate
+		products.forEach(product -> {
```
metas-ts, line 105: please use a for-loop. it's much easier to debug

metas-ts (approved):
AFAIU this issue shall be integrated today and i don't see why it couldn't but please, very soon do the following:
* check my comments
* find a better name for the process if you can (The 2 of "Rabatte kopieren mit Produktänderung 2" is not very informative)
* add a process description that a user can understand 
* add descriptions for process params (particularly explain wtf "Make target as source" means)

Issue #10860, opened 2021-03-22
Refactor the rest of the direct usages of RawWidget


siemiatj:
### Is this a bug or feature request?

Feature

### What is the current behavior?

In a few places we have direct usage of the `RawWidget` component.

### What is the expected or desired behavior?

The `WidgetWrapper` should be used in all cases.

PR #10960, opened 2021-04-09
Use WidgetWrapper everywhere

siemiatj:
Related to #10860


siemiatj:
Some tests need fixing

frontend/src/actions/IndependentWidgetsActions.js
```diff
@@ -0,0 +1,259 @@
...
+ * @param {number} windowId
+ * @param {string} docId
+ * @param {string} tabId
+ * @param {string} prop
+ * @param {any} value
+ */
+export function patchQuickInput({ windowId, docId, tabId, prop, value }) {
+  return (dispatch, getState) => {
+    const state = getState();
+    const id = state.widgetHandler.quickInput.id;
```
petrican, line 186: Probable would be nice to have some destruct in here and some safety checks

frontend/src/actions/IndependentWidgetsActions.js
```diff
@@ -0,0 +1,259 @@
...
+      subentityId: id,
+    }).then(({ data }) => {
+      const fields = data[0] && data[0].fieldsByName;
+
+      if (fields) {
+        dispatch(setQuickinputData(fields));
+      }
+
+      dispatch(setQuickinputDone());
+    });
```
petrican, line 207: .catch(...)   {  }
We need a way to say to the user that something went wrong. Use the message returned by the BE and render it in the err notification box we are using.

frontend/src/actions/IndependentWidgetsActions.js
```diff
@@ -0,0 +1,259 @@
...
+      'quickInput'
+    ).then(({ data }) => {
+      dispatch({
+        type: FETCH_QUICKINPUT_DATA,
+        payload: {
+          data: parseToDisplay(data.fieldsByName),
+          id: data.id,
+        },
+      });
+    });
```
petrican, line 234: also some sort of err handling. Push the message to the store to be rendered by the notif box

frontend/src/actions/IndependentWidgetsActions.js
```diff
@@ -0,0 +1,259 @@
...
+    return getLayout('window', windowId, tabId, 'quickInput', docId).then(
+      ({ data }) => {
+        dispatch({
+          type: FETCH_QUICKINPUT_LAYOUT,
+          payload: {
+            layout: data.elements,
+          },
+        });
+      }
+    );
```
petrican, line 257: catch err also

frontend/src/actions/TableActions.js
```diff
@@ -190,6 +184,68 @@ export function createTableData(rawData) {
...
+ *
+ * @param {string} tableId - table id
+ * @param {object} tableResponse - response data for the table
+ */
+export function fetchAttributes(tableId, tableResponse) {
+  return (dispatch, getState) => {
+    const state = getState();
+
+    // this check is only for unit tests purposes
+    if (state.tables) {
```
petrican, line 199: Since there are tests that need to be adapted probably we can ditch this part and we can write the fe code without adding specific code to handle tests

frontend/src/actions/TableActions.js
```diff
@@ -190,6 +184,68 @@ export function createTableData(rawData) {
...
+        dispatch(deleteViewAttributes());
+      }
+    }
+  };
+}
+
+/**
+ * @method deleteTable
+ * @summary Remove the table with specified `id` from the store
+ */
```
petrican, line 237: @param {string} id

frontend/src/actions/TableActions.js
```diff
@@ -190,6 +184,68 @@ export function createTableData(rawData) {
...
+export function deleteTable(id) {
+  return (dispatch) => {
```
petrican, line 239: safety check test for `undefined` ?

frontend/src/api/window.js
```diff
@@ -225,3 +225,48 @@ export function getPrintingOptions({ entity, windowId, docId, tabId, rowId }) {
...
 }
+
+/**
+ * @method initQuickActions
+ * @summary Fetch data for table quick input
+ * @param {string} entity - for example 'window'
+ * @param {string} windowId
+ * @param {string} docId
+ * @param {string} tabId
+ * @param {string} subentity
```
petrican, line 236: worth putting an example as you have it above for the entity on L:232

frontend/src/api/window.js
```diff
@@ -225,3 +225,48 @@ export function getPrintingOptions({ entity, windowId, docId, tabId, rowId }) {
...
+}
+
+/**
+ * @method completeRequest
+ * @summary Save changes in attributes/quick input
+ * @param {string} entity - for example 'window'
+ * @param {string} windowId
+ * @param {string} docId
+ * @param {string} tabId
+ * @param {string} subentity
```
petrican, line 253: example pls as on L: 249. Thanks

frontend/src/api/window.js
```diff
@@ -225,3 +225,48 @@ export function getPrintingOptions({ entity, windowId, docId, tabId, rowId }) {
...
+  entity,
+  docType,
+  docId,
+  tabId,
+  rowId,
+  subentity,
+  subentityId
+) {
+  return post(
+    `${config.API_URL}/${entity}${docType ? `/${docType}` : ''}${
```
petrican, line 266: Maybe we could make a bit of cleaning in here and have the values to be sent put in constants before forming the post.

`const docTypeParam = docType ? docType : '';`
...
and have in the end a cleaner thing

return post(`${config.API_URL}/${docTypeParam}/${docIdParam}/${tabIdParam}/${subEntity}/${entityId});

frontend/src/components/table/TableQuickInput.js
```diff
@@ -333,17 +284,25 @@ TableQuickInput.propTypes = {
   docType: PropTypes.any,
   docId: PropTypes.string,
   tabId: PropTypes.string,
-  allowShortcut: PropTypes.func.isRequired,
-  disableShortcut: PropTypes.func.isRequired,
-  modalVisible: PropTypes.bool.isRequired,
-  timeZone: PropTypes.string.isRequired,
+  layout: PropTypes.array,
+  data: PropTypes.object,
+  id: PropTypes.any,
```
petrican, line 289: string ?

petrican (changes requested):
Looks nice. Please try to fix the things pointed and the adapt the tests. Since this would  impact the widgets functionality I believe is wise to have it  roughly checked on a dev instance and also run the e2e automated tests and the manual tests against it. Thanks

frontend/src/actions/IndependentWidgetsActions.js
siemiatj: We already do, because this Promise is handled by `completeRequest`:

Screenshot 2021-04-19 at 16 10 14

siemiatj: We already do, because this Promise is handled by `completeRequest`:

Screenshot 2021-04-19 at 16 10 14

frontend/src/actions/IndependentWidgetsActions.js
siemiatj: It's already handled https://recordit.co/Iq2Zyl4D0y

frontend/src/actions/IndependentWidgetsActions.js
siemiatj: It's handled in the component.

frontend/src/actions/TableActions.js
siemiatj: Reducer takes care of that

Issue #10969, opened 2021-04-13
map actuator endpoints back to /


metas-ts:
### Is this a bug or feature request?
B
### What is the current behavior?

With spring-boot-2 the actuator-endpoints are by default under `/actuator`
Because of infrastructure-reasons it's currently not easy to make the actuator available on `/actuator/info`

### What is the expected or desired behavior?

The `info` endpoint shall once again be available under `/info` and not `/actuator/info`

PR #10970, opened 2021-04-13
change default actuator EP mapping back to how it was in spring-boot-2

metas-ts:
https://github.com/metasfresh/metasfresh/issues/10969


metas-ts (commented):
lgtm

Issue #10931, opened 2021-04-05
Filter default values are ignored 


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
Filter default values are ignored 
#### Which are the steps to reproduce?
Even if BE is sending correct default values those seem to be somehow ignored by the FE
### What is the expected or desired behavior?
FE should use that defaultValues

PR #10977, opened 2021-04-14
Fix useEffect and pass empty array

petrican:


metas-Kay:
related to: https://github.com/metasfresh/metasfresh/issues/10931

PR #10992, opened 2021-04-16
Use active filter as indicator

petrican:
https://www.loom.com/share/36e27c1b0f40400399fb3804f89c6538

https://github.com/metasfresh/metasfresh/issues/10931


siemiatj:
I've noticed only one thing off - filter value stays active even after clearing the single filter https://recordit.co/VSWfyV5uIa

Maybe this could also fix showing active filter name, after deselecting checkbox with only two values: https://recordit.co/yqj9kz0VBs ? But I'll leave this up to IT to decide

petrican:
I've added the code that deals with that situation.
Also It fixes another bug that I found while working on this.

Pls see the video https://www.loom.com/share/3afe124be41a4393b6915f23cc05c8d3 @siemiatj 

Please re-check this and let me know if it can be integrated. Thanks

siemiatj (approved):

PR #11009, opened 2021-04-21
Add catch for the updateTabLayout

petrican:
https://www.loom.com/share/d87541175f34405ca49e2c56272da638

### DIFF
--- a/frontend/src/containers/MasterWindow.js
+++ b/frontend/src/containers/MasterWindow.js
@@ -188,11 +188,13 @@ class MasterWindowContainer extends PureComponent {
       sortingOrder = (ordering.ascending ? '+' : '-') + ordering.fieldName;
     }
 
-    updateTabLayout(windowType, activeTabId).then(() => {
-      getTabRequest(activeTabId, windowType, docId, sortingOrder).then((rows) =>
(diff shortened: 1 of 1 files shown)


frontend/src/containers/MasterWindow.js
```diff
@@ -188,11 +188,13 @@ class MasterWindowContainer extends PureComponent {
...
-        updateTabTableData(tableId, rows)
-      );
-    });
+    updateTabLayout(windowType, activeTabId)
+      .then(() => {
+        getTabRequest(activeTabId, windowType, docId, sortingOrder).then(
+          (rows) => updateTabTableData(tableId, rows)
+        );
+      })
+      .catch((error) => error);
```
siemiatj, line 197: So we're catching this error and then what ? Was it breaking the app or was reported by QA ? Maybe at least console.error it so that it will be of any use to us ?

siemiatj (changes requested):

frontend/src/containers/MasterWindow.js
petrican: This was not breaking anything it was just outputing the err to console. This should be recorded in logstash normally and we should not pollute the console. Will add a todo for that

PR #11013, opened 2021-04-22
Fix patch for cases when the response comes as array

petrican:
https://www.loom.com/share/7e7710ea40e04877a2c961f0c208d84c

### DIFF
--- a/frontend/src/api/view.js
+++ b/frontend/src/api/view.js
@@ -110,12 +110,10 @@ export function patchRequest({
   ).then((rawResponse) => {
     // this is fixed on the FE because the BE is not consistent in sending the `documents` key with every PATCH request
     // this differs when patch is done within processes for example
-    if (rawResponse.data.documents) {
-      return Promise.resolve(rawResponse);
-    } else {
+    if (!rawResponse.data.documents) {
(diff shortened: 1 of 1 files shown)


frontend/src/api/view.js
```diff
@@ -107,7 +107,16 @@ export function patchRequest({
       (isAdvanced ? '?advanced=true' : '') +
       (isEdit ? '/edit' : ''),
     payload
-  );
+  ).then((rawResponse) => {
+    // this is fixed on the FE because the BE is not consistent in sending the `documents` key with every PATCH request
+    // this differs when patch is done within processes for example
+    if (rawResponse.data.documents) {
```
siemiatj, line 113: Why not:

```
    if (!rawResponse.data.documents) {
      rawResponse.data.documents = rawResponse.data;
    }
    return Promise.resolve(rawResponse);
```

siemiatj (approved):
I don't like this, but if it helps then I'm ok with this change. We can simplify it a bit though.

frontend/src/api/view.js
petrican: Yep that would work. Thanks

PR #11044, opened 2021-05-03
Invoice-Line: Update PriceActual as users edits PriceEntered and Discount

metas-ts:
* de.metas.invoice.callout.C_InvoiceLine: update priceActual on priceEntered or discount change
  * Encaspulate the clacualtion code in InvoiceLinePriceAndDiscount which is now also used from InvoiceLineBL
* CalloutInvoice: remove the code that sets priceActual:=priceEntered; it ignored the discount.
* minor improvements
  * remove obsolete FtpImporter and ftp4che dependency
  * Rename existing class PriceAndDiscount to OrderLinePriceAndDiscount
  * bump up commons-collections4


metas-ts (commented):
lgtm

Issue #10655, opened 2021-02-17
Pricesystem of contract term not considered correctly


BodorAnnamaria:
### Is this a bug or feature request?
BG

### What is the current behavior?
Create a Sales Order Line with the `contract term` set
change the qty => contract term will be ignored on calculation of qty and line amount(quantity in price is not taken correctly)

#### Which are the steps to reproduce?
Create a Sales Order
add on order line with qty
add the contract term
change the qty

### What is the expected or desired behavior?

Changing the qty the `Menge in Preiseinheit` and `Line Amount` should be calculated based on the term defined on the line

PR #11056, opened 2021-05-05
#gh10655 use contract term pricesystem if available when calculating Quantity in Price Unit and Line Amount

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/10655

### DIFF
--- a/backend/de.metas.contracts/src/main/java/de/metas/contracts/subscription/impl/SubscriptionBL.java
+++ b/backend/de.metas.contracts/src/main/java/de/metas/contracts/subscription/impl/SubscriptionBL.java
@@ -52,7 +52,6 @@ import de.metas.product.IProductDAO;
 import de.metas.product.IProductPA;
 import de.metas.product.ProductAndCategoryId;
 import de.metas.product.ProductId;
-import de.metas.quantity.QuantitiesUOMNotMatchingExpection;
 import de.metas.quantity.Quantity;
 import de.metas.tax.api.TaxCategoryId;
 import de.metas.uom.IUOMConversionBL;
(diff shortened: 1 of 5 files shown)


backend/de.metas.contracts/src/main/java/de/metas/contracts/subscription/impl/SubscriptionBL.java
```diff
@@ -1018,4 +997,79 @@ public boolean isActiveTerm(@NonNull final I_C_Flatrate_Term term)
...
+		}
+
+		// priceQty is the qty do be delivered during one complete subscription term
+		final Quantity priceQty = qtyOrderedPerRun.multiply(numberOfRuns);
+
+		// qty ordered needs to be set because it will be used to compute the
+		// line's NetLineAmount in MOrderLine.beforeSave()
+		ol.setQtyOrdered(priceQty.toBigDecimal());
+
+		ol.setQtyEnteredInPriceUOM(priceQty.toBigDecimal());
```
metas-ts, line 1063: i think priceQty needs to be converted to the `ol`s C_UOM (with uomConversionBL)

backend/de.metas.contracts/src/main/java/de/metas/contracts/subscription/impl/SubscriptionBL.java
```diff
@@ -1018,4 +997,79 @@ public boolean isActiveTerm(@NonNull final I_C_Flatrate_Term term)
 
 		return !isCancelledOrVoided;
 	}
+
+	@Override
+	public void updatePrices(
```
metas-ts, line 1002: plese call it e.g. updateQtysAndPrices

metas-ts (changes requested):

backend/de.metas.contracts/src/main/java/de/metas/contracts/subscription/impl/SubscriptionBL.java
adi-stefan: Done

backend/de.metas.contracts/src/main/java/de/metas/contracts/subscription/impl/SubscriptionBL.java
adi-stefan: Was unsure if I should throw an error if the conversion fails, or just use the `priceQty` result. Error made more sense. Feel free to disagree.

metas-ts (approved):

PR #11085, opened 2021-05-11
BPartner Rest V2: support external reference version

cif-ps:
https://github.com/metasfresh/metasfresh/issues/11084


misc/de-metas-common/de-metas-common-bpartner/src/main/java/de/metas/common/bpartner/v2/request/JsonRequestBPartnerUpsertItem.java
```diff
@@ -42,17 +44,23 @@
 	@NonNull
 	String bpartnerIdentifier;
 
-	@ApiModelProperty(position = 20,
+	@ApiModelProperty(position = 20, //
+			value = "Used only when the BPartner identifier is an external reference.")
+	String externalVersion;
```
metas-ts, line 49: place indicate that it's optional and what its purpose is if set

metas-ts (changes requested):
..minor request regarding API-doc for the version property

metas-ts (approved):

PR #11086, opened 2021-05-11
Shopware BPartner-SyncAdvise Konfig

fp-ps:


metas-ts (changes requested):
The new sync-advise fields need to be in `ExternalSystem_Config_Shopware6Mapping`

(also note that i fixed an unrelated problem in cucumber, so all the tests are once again working)

metas-ts (approved):

metas-ts:
lgtm, going to integrate after successfull build

PR #11087, opened 2021-05-11
Avoid assumption error "inOutLine shall not be about packing materials" if the inout is an empties return

metas-ts:


metas-ts (commented):
lgtm

PR #11115, opened 2021-05-19
Other external system product import

cp-ps:


misc/services/camel/de-metas-camel-externalsystems/core/src/main/java/de/metas/camel/externalsystems/core/to_mf/v2/ProductRouteBuilder.java
```diff
@@ -70,6 +75,29 @@ public void configure() throws Exception
...
+					final var jsonRequestProductUpsert = ((ProductUpsertCamelRequest)lookupRequest).getJsonRequestProductUpsert();
+
+					log.info("Product upsert route invoked with " + jsonRequestProductUpsert.getRequestItems().size() + " requestItems");
+					exchange.getIn().setBody(jsonRequestProductUpsert);
+				})
+				.marshal(CamelRouteHelper.setupJacksonDataFormatFor(getContext(), JsonRequestProductUpsert.class))
+				.removeHeaders("CamelHttp*")
+				.setHeader(CoreConstants.AUTHORIZATION, simple(CoreConstants.AUTHORIZATION_TOKEN))
+				.setHeader(Exchange.HTTP_METHOD, constant(HttpEndpointBuilderFactory.HttpMethods.PUT))
+				.toD("http://{{metasfresh.upsert-product-v2.api.uri}}/${header." + HEADER_ORG_CODE + "}");
```
metas-ts, line 100: I just thought, why do we have the hardcoded `http://` ? wouldn't it be better to include the protocol in the respective URI as well?
Because in future we might want to run camel a bit more remote and then we'd might need to use `https://` or whatever..
Not urgent for now, but I think it will be part of one of the next issues..

metas-ts (approved):

backend/de.metas.externalreference/pom.xml
```diff
@@ -31,6 +31,13 @@
             <version>${metasfresh-common.version}</version>
         </dependency>
 
+<!--        <dependency>-->
+<!--            <groupId>de.metas.purchasecandidate</groupId>-->
+<!--            <artifactId>de.metas.purchasecandidate.base</artifactId>-->
+<!--            <version>${metasfresh.version}</version>-->
+<!--            <scope>compile</scope>-->
+<!--        </dependency>-->
+
```
metas-ts, line 40:
```suggestion
```

metas-ts (approved):

Issue #11119, opened 2021-05-20
Create Sales/Purchase Price List action


BodorAnnamaria:
### Is this a bug or feature request?
FR
### What is the current behavior?
Currently exist the Create Price List action. This action in Price List Version dropdown is displaying the Price List Version for Sales and Purchase also.
### What is the expected or desired behavior?
Create two new action. One to create price list for sales  and one to create price list for purchase

PR #11121, opened 2021-05-20
#11119 Create Sales/Purchase Price List action

BodorAnnamaria:
https://github.com/metasfresh/metasfresh/issues/11119


metas-ts (changes requested):
generally lgtm, but please remove the old "Preisliste" process

metas-ts (approved):

PR #11138, opened 2021-05-21
UOM in inventory line aggregation

metas-rc:
https://github.com/metasfresh/metasfresh/issues/11137


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/inventory/draftlinescreator/InventoryLineAggregatorFactory.java
```diff
@@ -147,6 +161,9 @@ public AggregationType getAggregationType()
 			@NonNull
 			ProductId productId;
 
+			@Nullable
```
metas-ts, line 164:
```suggestion
			// needed for the case that stocking UOM has changed and there are still HUs with an old UOM
			@Nullable
```

metas-ts (approved):

PR #11142, opened 2021-05-24
11139 f default lookup

petrican:
https://www.loom.com/share/b92b964e361a4d69ab48ad1237d03459

Consider default values for Lookup, Datetimerange and Amount. If others will be needed they will be added in a sep issue.


petrican:
closing. Bad target

PR #11143, opened 2021-05-24
11139 f default lookup

petrican:
https://www.loom.com/share/b92b964e361a4d69ab48ad1237d03459

Consider default values for Lookup, Datetimerange and Amount. If others will be needed they will be added in a sep issue.

### DIFF
--- a/frontend/src/components/widget/DatetimeRange.js
+++ b/frontend/src/components/widget/DatetimeRange.js
@@ -21,16 +21,36 @@ class DatetimeRange extends Component {
 
   /**
    * @method componentDidMount
-   * @summary ToDo: Describe the method
-   * @todo Write the documentation
+   * @summary lifecycle hook - We set in the local state the `startDate` and `endDate` after the component is mounted
    */
   componentDidMount() {
-    const { value, valueTo } = this.props;
-    if (value && valueTo) {
+    const {
+      isFilterActive,
+      value,
+      valueTo,
+      defaultValue,
+      defaultValueTo,
+      field,
+      updateItems,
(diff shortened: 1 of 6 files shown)


frontend/src/components/widget/DatetimeRange.js
```diff
@@ -205,6 +225,11 @@ DatetimeRange.propTypes = {
   mandatory: PropTypes.any,
   validStatus: PropTypes.any,
   timePicker: PropTypes.any,
+  isFilterActive: PropTypes.bool,
+  defaultValue: PropTypes.string,
```
siemiatj, line 229: You can remove `value/valueTo` from props

frontend/src/components/widget/WidgetRenderer.js
```diff
@@ -172,7 +178,13 @@ class WidgetRenderer extends PureComponent {
       tabIndex,
       onShow,
       onHide,
+      isFilterActive: filterActiveState,
+      defaultValue: widgetData[0].defaultValue,
```
siemiatj, line 182: I assume we don't want to use `value` as a backup here, right ?

siemiatj (approved):
Nothing major, good to merge

frontend/src/components/widget/DatetimeRange.js
petrican: not quite we use defaultValue, defaultVauleTo as default ones and value,/valueTo as the normal ones

petrican:
Will do some more tests on the dev instance and then I will merge. Thanks @kuba

Issue #11118, opened 2021-05-20
"Qty shall be greather than zero" error when creating orderline with batch entry


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
batch entry - Qty shall be greather than zero pops up
#### Which are the steps to reproduce?
batch entry - randomply
### What is the expected or desired behavior?
no error should be shown

PR #11163, opened 2021-05-25
Change how refs are passed to widgets

siemiatj:
Related to #11118 #10860

### DIFF
--- a/frontend/src/components/table/TableQuickInput.js
+++ b/frontend/src/components/table/TableQuickInput.js
@@ -24,6 +24,7 @@ class TableQuickInput extends Component {
     this.state = {
       editedField: 0,
     };
+    this.rawWidgets = [];
   }
 
   componentDidMount() {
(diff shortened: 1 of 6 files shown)


frontend/src/components/table/TableQuickInput.js
```diff
@@ -24,43 +24,43 @@ class TableQuickInput extends Component {
...
-          break;
+      if (!item[0].value) {
+        if (this.rawWidgets) {
+          this.focusWidgetField();
         }
       }
     }
   }
 
+  focusWidgetField = (fieldNo) => {
```
petrican, line 50: Doc pls

frontend/src/components/table/TableQuickInput.js
```diff
@@ -133,6 +185,17 @@ class TableQuickInput extends Component {
     }
   }
 
+  setRef = (ref) => {
+    this.form = ref;
+  };
+
+  setWidgetsRef = (c) => {
```
petrican, line 192: pls use a diff name with more meaning for the name. Something like `componentNode` or `nodeRef` instead of just `c`

petrican (changes requested):
Minor things. Looks fine, I checked and it does focus as expected. There is one issue with the build that is failing. Pls fix that and then we can merge. Thanks

PR #11167, opened 2021-05-26
Introduce customer retention initial threshold

metas-rc:
https://github.com/metasfresh/metasfresh/issues/11164

### DIFF
--- a/backend/de.metas.contracts/src/main/java/de/metas/contracts/impl/ContractsDAO.java
+++ b/backend/de.metas.contracts/src/main/java/de/metas/contracts/impl/ContractsDAO.java
@@ -205,7 +205,7 @@ public class ContractsDAO implements IContractsDAO
 		return Services.get(IQueryBL.class).createQueryBuilder(I_C_Flatrate_Term.class)
 				.addOnlyActiveRecordsFilter()
 				.addOnlyContextClient()
-				.addEqualsFilter(I_C_Flatrate_Term.COLUMNNAME_Bill_BPartner_ID, bpartnerId.getRepoId())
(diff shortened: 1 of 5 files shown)


backend/de.metas.contracts/src/main/java/de/metas/contracts/impl/ContractsDAO.java
```diff
@@ -197,4 +197,20 @@ public I_C_Flatrate_Term retrieveLatestFlatrateTermForBPartnerId(@NonNull final
...
 	}
+
+	@Cached(cacheName = I_C_Flatrate_Term.Table_Name + "#by#BPartnerId")
+	@Override
+	public I_C_Flatrate_Term retrieveFirstFlatrateTermForBPartnerId(@NonNull final BPartnerId bpartnerId)
+	{
+		return Services.get(IQueryBL.class).createQueryBuilder(I_C_Flatrate_Term.class)
+				.addOnlyActiveRecordsFilter()
+				.addOnlyContextClient()
+				.addEqualsFilter(I_C_Flatrate_Term.COLUMNNAME_Bill_BPartner_ID, bpartnerId.getRepoId())
```
metas-ts, line 208:
```suggestion
				.addEqualsFilter(I_C_Flatrate_Term.COLUMNNAME_Bill_BPartner_ID, bpartnerId)
```
the filter supports RepoIdAwares directly

metas-ts (approved):
lgtm. for next time you could consider to make an ad_sysconfig like this one overridable on org level.
but in the current case we can do this if&when required..

PR #11183, opened 2021-05-27
After adding new entry, table quick input should only focus on first field

siemiatj:
This also lets us simplify the component itself and get rid of state.

Related to #10860

### DIFF
--- a/frontend/src/components/table/TableQuickInput.js
+++ b/frontend/src/components/table/TableQuickInput.js
@@ -17,15 +17,8 @@ import WidgetWrapper from '../../containers/WidgetWrapper';
 class TableQuickInput extends Component {
   // promise with patching for queuing form submission after patch is done
   patchPromise;
-
-  constructor(props) {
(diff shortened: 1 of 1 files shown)


petrican (changes requested):
there is an issue after first jump to the 2nd field. When inputing chars it goes back to the initial field which is NOK. Pls fix thx

siemiatj:
> there is an issue after first jump to the 2nd field. When inputing chars it goes back to the initial field which is NOK. Pls fix thx

I've checked against the old `dev1050` and tbh to me it looks exactly the same. When using `IFCO` product everything works ok. When trying `convenience` product the backend doesn't return any data in the current revision (also the product name is different so maybe it's caused by that).

https://recordit.co/V2s4QzBypY

petrican (approved):
as discussed. It seems to be working exactly as it used to be. Any fine tunning of this should be done in sep issue. @siemiatj

Issue #11015, opened 2021-04-22
Move frontend to react-router v5


siemiatj:
### Is this a bug or feature request?

Feature

PR #11200, opened 2021-05-29
Complete overhaul of routing and authentication + router v5

siemiatj:
Squashed my commits to have an easier job merging this against master.

Related to #11015 #11469

Tested normal login/login to a page/outside logout/token login flows: https://recordit.co/VIm1TPWS95


siemiatj:
@petrican can you start looking into this while I fix the tests ?

frontend/package.json
```diff
@@ -194,6 +196,7 @@
     "throttle-debounce": "2.1.0",
     "timer-machine": "1.1.0",
     "uuid": "3.3.2",
-    "webpack-git-hash": "1.0.2"
+    "webpack-git-hash": "1.0.2",
+    "why-did-you-update": "^1.0.8"
```
petrican, line 200: Isn't that one deprecated ? I recall we switched to https://github.com/metasfresh/metasfresh/blob/master/frontend/package.json#L64

frontend/src/components/app/LoginForm.js
```diff
@@ -200,7 +199,7 @@ class LoginForm extends Component {
         if (roleSelect) {
           return loginCompletionRequest(role)
             .then(() => {
-              dispatch(loginSuccess(auth));
+              auth.login();
```
petrican, line 202: Not sure I follow exactly why we are ditching the actions dispatching and not doing the logic in the redux actions where this should reside.

petrican: How will we "know" we are authenticated at any point in time throughout the app at any level ? Do we need to pass the auth anywhere ? Can't we use just a flag from the redux store ?

frontend/src/containers/App.js
```diff
@@ -12,45 +10,53 @@ import {
...
+ */
+const App = (props) => {
+  // const [pluginsLoading, setPluginsLoading] = useState(!!APP_PLUGINS.length);
+  const auth = useAuth();
+  const dispatch = useDispatch();
 
-    this.state = {
-      pluginsLoading: !!APP_PLUGINS.length,
-    };
+  useWhyDidYouUpdate('App', props);
```
petrican, line 54: are we using this also in prod ?

petrican (changes requested):
What I find a bit odd about this implementation is the fact that we are moving away from having the  state in one single store(the redux store) which I believe it is our goal if we want to have a neat and easy to follow codebase. Imagine a new developer that  is going to work on this codebase expects everything to be in one place. Probably if he will see the authentication logic outside of it he/she will most probable is going to be confused. Also I would try as much of possible not to use extra libs that add more size to the overall bundle(qhistory, qs ..) as we already have quite some heavy one. Thinking also that one of the next steps for refactoring would be to also reduce this size by ditching what we consider not to be really necessary.  

I really appreciate the effort you put into this and  I like that it is working fine but I believe  we should also consult @teosarca regarding this  change. Teo can you pls through an eye over and let us know your thoughts. Thanks

petrican:
If we put side by side the benefits and downsides. I can conclude

Benefits:
 1) less code written due to hooks used
Downsides: 
 1) does not follow redux philosophy 
 2) adds extra size to the bundle due to extra libs used
 3) does not match with the existing app structure as it should (our goal is to use redux though the entire app) 

Feel free to add comments regarding this. Thanks

frontend/src/components/app/LoginForm.js
siemiatj: I don't think we've ever used this flag tbh. We were using the value from the local storage. So it's a duplication of concerns. If you'll check how `auth.login` works, it does pretty much the same but in a clean way. All the requests and actions, that were called before are still fired.

frontend/src/containers/App.js
siemiatj: Nope

siemiatj:
Ok, so to address your concerns:

###Sizing:
With the new solution we're using query-string which was already a part of the package plus
qhistory, which is 23kb.

In contrary, connected-react-router is 444kb (they don't give size of the react-router-redux but I would assume it's similar).

So if anything, we're actually making our package smaller.

###Structure
Maybe I'm missing something but when it was decided to store everything in the redux store ? As far as I remember we were always saying that store should have the data that is shared across the app and can't be shared easily in other way. For a while now there's a tendency in the js world to limit the usage of stores and move to the context api. It's faster and doesn't require to wrap components in the `connect` HOC. Moreover keeping the size of the store is also crucial, as the bigger it gets - the slower it gets. We've just moved filters and some independent components data there, so I think it makes sense to take something out. Especially that the `isLogged` flag stays (as it's a part of the user session response) and we're only really removing routing. Since we've never used the routing from the store for anything else than reading the url, I see no reason for it to be there. Behind the scenes it still uses the browser history API but adds additional redraws when the state changes. In my opinion it's a no-brainer for it to go away. 

###Summary
I agree, that adapting the tests (especially e2e) is an additional work but in the process I've already found out, that some were false-positive.

The biggest benefit (apart of having 10x less redraws, as described in the me03 issue) is that with those changes we'll be closer to updating to react 17. Some of the libs we've been using are not maintained for 4 years and there are no plans to update them.

petrican (approved):
Since the goal is to be up to date with the package deps and also React core version I am willing to make a compromise on this implementation and we can move on. Please document the reason we took that approach also in the code (hooks file probable is more suitable). Also make sure this gets tested roughly before we merge it into master, dev instance, e2e tests, manual e2e aso. Thanks cc: @teosarca

siemiatj:
e2e tests finally passing after merging to master. @petrican can you have another look, so that we could hopefully merge this on Monday ?

petrican:
There seem to be some conflicts in the package.json pls fix them. Thanks

PR #11204, opened 2021-05-31
Master alberta attachment ep

fp-ps:


backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/attachment/AttachmentRestService.java
```diff
@@ -0,0 +1,172 @@
...
+
+		final ExternalIdentifier targetIdentifier = ExternalIdentifier.of(externalTargetIdentifier);
+
+		switch (targetIdentifier.getType())
+		{
+			case METASFRESH_ID:
+				return targetIdentifier.asMetasfreshId();
+			case EXTERNAL_REFERENCE:
+				return externalReferenceService
+						.getJsonMetasfreshIdFromExternalReference((OrgId)null, targetIdentifier, externalReferenceType)
```
fp-ps, line 139: I'm not sure this is ok as the `OrgId` is not present. This is the proposed solution.

metas-ts (commented):
generally looks ok so far

metas-ts (approved):

PR #11222, opened 2021-06-02
match audit config against org.ANY

cp-ps:
refs: https://github.com/metasfresh/metasfresh/issues/11051


metas-ts (approved):

metas-ts:
lgtm; will integrate after successful build

PR #11246, opened 2021-06-09
Blonde monkey uat audit cucumber

cp-ps:


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/audit/ApiAuditLoggable.java
```diff
@@ -77,17 +79,17 @@ public ILoggable addLog(final String msg, final Object... msgParameters)
...
 
-		if (buffer.size() >= bufferSize)
-		{
-			flush();
-		}
+		return this;
+	}
+
+	@Override
+	public ILoggable addTableRecordReferenceLog(final @NonNull ITableRecordReference recordRef, final @NonNull String type, final @NonNull String trxName)
```
metas-ts, line 88: i think it makes sense to call this method with `StateType` isntead of `String`

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/SystemTime_StepDef.java
```diff
@@ -32,4 +33,10 @@ public void metasfresh_has_the_local_time(String zonedDateTime)
 	{
 		SystemTime.setFixedTimeSource(zonedDateTime);
 	}
+
+	@And("we wait for {string} ms")
```
metas-ts, line 37: AFAIU this can be {int}. then you don't need the parse

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/apiauditfilter/ApiAuditFilter_StepDef.java
```diff
@@ -0,0 +1,305 @@
...
+	private final IQueryBL queryBL = Services.get(IQueryBL.class);
+	private final IADTableDAO tableDAO = Services.get(IADTableDAO.class);
+	private final TestContext testContext;
+
+	public ApiAuditFilter_StepDef(final TestContext testContext)
+	{
+		this.testContext = testContext;
+	}
+
+	@And("all the data is reset to default")
```
metas-ts, line 72: better call it something like "all the API audit data is reset"

backend/de.metas.util/src/main/java/de/metas/util/ILoggable.java
```diff
@@ -36,8 +39,16 @@
...
+	/**
+	 * Flush any buffered logs
+	 */
 	default void flush()
 	{
 	}
+
+	default ILoggable addTableRecordReferenceLog(final ITableRecordReference recordRef, final String type, final String trxName)
+	{
+		return this;
```
metas-ts, line 51: please "render" a log message string and call `addLog()` with it

metas-ts (changes requested):
Generally looks good! please check my minor request.
About the zoom-to-problem, it's a bit too late for me now...

I _guess_ you could make an `AD_Relation_Type` that points not to the subtab with its `API_Request_Audit_Log`, but to the main window, using `API_Request_Audit_Log.API_Request_Audit_ID`.

PS: two more things :zap: 
* wdyt about not logging the references to API_Request_Audit and API_Response_Audit ? i think they are quite redundant in there and don't rpobide and benefit.
* did you also test something lik a POST or PUT, where e.g. a C_BPartner (at any rate not API_Request_Audit and API_Response_Audit) was created or altered? if not, please do

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/audit/ApiAuditLoggable.java
dm-ps: I agree, but I can't use StateType in ILoggable as it implies a circular dependency with `de.metas.adempiere.adempiere.base`.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/SystemTime_StepDef.java
dm-ps: Sure, will change

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/apiauditfilter/ApiAuditFilter_StepDef.java
dm-ps: Sure

backend/de.metas.util/src/main/java/de/metas/util/ILoggable.java
dm-ps: Sure

metas-ts (approved):
lgtm; AFAI the cucmber-record-reference tests are now not part of the test, but that's OK for me for now. let's see where we can add them later

Issue #11281, opened 2021-06-14
Make user friendly the reading of QR code


petrican:
### Is this a bug or feature request?
F
### What is the expected or desired behavior?
There should be a Scan button inline that would allow scanning of the QR code

PR #11287, opened 2021-06-15
ScanBtn on the right to improve UI

petrican:
related to https://github.com/metasfresh/metasfresh/issues/11281

Screenshot 2021-06-15 at 14 13 28

Screenshot 2021-06-15 at 14 14 36

Screenshot 2021-06-15 at 14 15 41

Screenshot 2021-06-15 at 14 42 50

After the detection the field is populated with the result of the QR_code

Screenshot 2021-06-15 at 14 45 29

### DIFF
--- a/frontend/src/assets/css/inputs.css
+++ b/frontend/src/assets/css/inputs.css
@@ -1004,3 +1004,6 @@ table {
 }
 
 
+.input-qr-scanner-wrapper {
+    padding: 20px 30px 20px 28px;
+}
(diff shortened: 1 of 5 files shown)


siemiatj (approved):
I would add padding between the input and the scanner component if possible, but other than that looks good.

petrican:
> I would add padding between the input and the scanner component if possible, but other than that looks good.

I added those in the last commit. The images from the initial PR are not reflecting the last changes. Thanks

PR #11294, opened 2021-06-15
Support HU_Item_Storage being out of sync with HU_Storage

cp-ps:
…rage

refs: https://github.com/metasfresh/metasfresh/issues/11240


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/ShipmentSchedule.java
```diff
@@ -105,14 +105,24 @@ public boolean hasAttributes(
...
 			@NonNull final IAttributeDAO attributeDAO)
 	{
+		final ImmutableSet<AttributeSetInstanceId> nonNullTargetAsiIds = targetAsiIds
+				.stream()
+				.filter(asiId -> !AttributeSetInstanceId.NONE.equals(asiId))
+				.collect(ImmutableSet.toImmutableSet());
+
+		if (nonNullTargetAsiIds.isEmpty())
+		{
+			return true;
```
metas-ts, line 115:
```suggestion
			return true; // targetAsiIds was effectively empty, so we return true
```

PR #11304, opened 2021-06-16
Just alt+Enter to submit in prompt

petrican:
Go to a SO and select some order lines
Screenshot 2021-06-16 at 13 46 31

## 1) Use the ALT+Y combination
Screenshot 2021-06-16 at 13 47 37


siemiatj (approved):
👍

Issue #11305, opened 2021-06-16
401 from the backend after updating row/using barcode search


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
FE doesn't deal nicely with the err that it gets
#### Which are the steps to reproduce?
picking tray clearing
### What is the expected or desired behavior?
FE should show a nicer error not that document not found

PR #11306, opened 2021-06-16
Remove the action that leads to 404

petrican:
https://www.loom.com/share/cdc42b735f2b403d810b43650d875d21

## 1) Go to picking tray clearing
Screenshot 2021-06-16 at 16 32 06

## 2) Barcode on 2nd panel
Screenshot 2021-06-16 at 16 32 57

## 3) Input something
Screenshot 2021-06-16 at 16 33 28

## 4) Should see the error but the UI should remain unchanged. no 404 on the screen
Screenshot 2021-06-16 at 16 33 54

### DIFF
--- a/frontend/src/actions/ViewActions.js
+++ b/frontend/src/actions/ViewActions.js
@@ -22,7 +22,6 @@ import {
   FETCH_LAYOUT_ERROR,
   FILTER_VIEW_PENDING,
   FILTER_VIEW_SUCCESS,
-  FILTER_VIEW_ERROR,
   FETCH_LOCATION_CONFIG_SUCCESS,
   FETCH_LOCATION_CONFIG_ERROR,
   RESET_VIEW,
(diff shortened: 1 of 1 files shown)


siemiatj:
Until backend solves this, looks good to me

petrican:
> @petrican @siemiatj any reason to not remove it from viewHandler.js and from all other files where FILTER_VIEW_ERROR is present?
> 
> image

Its indeed not used but I would leave it for now as it is. Reason: in the near future we'll most probably need to adapt it to deal with the refined errors we get from the BE and it is going to be a simpler to just see it on spot.

teosarca:
@petrican understand.

that's not a good practice according to clean code principles (and not only, e.g. https://refactoring.guru/smells/dead-code, https://softwareengineering.stackexchange.com/questions/103031/should-i-remove-unreferenced-code),

but I can live with that now.

metas-Kay:
related to https://github.com/metasfresh/metasfresh/issues/11305

Issue #11285, opened 2021-06-14
Active tab not refreshed on ws event


siemiatj:
### Is this a bug or feature request?

Bug

### What is the current behavior?

When document's status is changed via an action, there's a websocket event, that should trigger refresh of the whole document, including active tab.

PR #11309, opened 2021-06-16
Don't reject promise on successfull response

siemiatj:
Related to #11285


siemiatj:
This has already been approved on master so merging.

Issue #11338, opened 2021-06-18
Update inline tab's valid status on PATCH


siemiatj:
### Is this a bug or feature request?

Bug

PR #11339, opened 2021-06-18
Update inline tab's data with requests

siemiatj:
Related to #11338


siemiatj:
Got a green light to merge from Teo

PR #11342, opened 2021-06-19
Invoke ExternalSystem via RabbitMQ

cp-ps:


metas-ts (approved):

metas-Kay:
related to https://github.com/metasfresh/metasfresh/issues/11341

Issue #11360, opened 2021-06-23
prevent reversed credit memo from making invoice candidate invoiceable again


adi-stefan:
### Is this a bug or feature request?
Bug
### What is the current behavior?
If credit memo is reversed, invoice candidate is made invoiceable again.
#### Which are the steps to reproduce?
* you create a credit memo for an invoice.
  * that invoice came from invoice candidates
  * when creating the credit memo you tell MF that you don't want to invoice the ICs once again
  * => all fine until now :tada: 
* now you reverse the credit memo
  * :bomb: now the IC is invoiceable again. it shall still not be!

PR #11361, opened 2021-06-23
#gh11360: prevent credit memo reversal from making the IC invoiceable again

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/11360

### DIFF
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandBL.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandBL.java
@@ -40,8 +40,8 @@ import de.metas.bpartner.BPartnerId;
 import de.metas.bpartner.BPartnerLocationId;
 import de.metas.bpartner.service.IBPartnerBL;
 import de.metas.bpartner.service.IBPartnerDAO;
-import de.metas.common.util.time.SystemTime;
 import de.metas.cache.CCache;
+import de.metas.common.util.time.SystemTime;
 import de.metas.currency.Currency;
 import de.metas.currency.CurrencyPrecision;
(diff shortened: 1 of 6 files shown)


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandBL.java
```diff
@@ -1456,7 +1456,7 @@ public void handleReversalForInvoice(final org.compiere.model.I_C_Invoice invoic
 					note = "@C_InvoiceLine@  @QtyInvoiced@ = " + il.getQtyInvoiced() + " @IsCreditedInvoiceReinvoicable@='Y'; ignoring overlap, because credit memo";
 				}
 
-				else if (creditMemo && !creditedInvoiceReinvoicable && creditedInvoiceIsReversed)
+				else if (creditMemo && !creditedInvoiceReinvoicable)
```
metas-ts, line 1459: ` && creditedInvoiceIsReversed` was there for a reason..
From the variable naming I assume that it was there to cover the case that you have an invoice, then create a credit-memo and then revert the invoice.

Update: I used git blame and found that this is the reason: https://github.com/metasfresh/metasfresh/issues/6540
Luckily Ruxi made a test in InvoiceCandBLHandlCompleteAndReversalTest that now fails :-P
Please check out those tests and add one for this issue's case as well, when it's fixed

metas-ts (changes requested):

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandBL.java
adi-stefan: Fixed.

metas-ts (approved):

PR #11363, opened 2021-06-23
11298 e2e mid june2021

petrican:
https://github.com/metasfresh/metasfresh/issues/11298

Done package updates on e2e also. 

Considered updating the packages without any impact

ran also the e2e suites on a dev instance and it looks fine


siemiatj:
Why we don't use fixed versions anymore ?

petrican:
Good point. I'll change those to be fixed. Thanks

siemiatj (approved):

Issue #11366, opened 2021-06-23
Add GLN to address field for documents


cristinamghita:
### Is this a bug or feature request?
FR

PR #11367, opened 2021-06-23
#11366 Add GLN to address field for documents

cristinamghita:
#11366 Add GLN to address field for documents


backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/75-de.metas.fresh.reports/5594360_sys_AddGLN_ToAddresInFunctionsForDocs.sql
```diff
@@ -0,0 +1,202 @@
...
+                                    COALESCE(loc.postal || ' ', '') ||
+                                    COALESCE(loc.city, '')
+                    )                      as org_addressline,
+                COALESCE(loc.address1, '') as address1,
+                COALESCE(loc.postal, '')   as postal,
+                COALESCE(loc.city, '')     as city,
+                c.Name                     as country
+         FROM ad_orginfo oi
+                  JOIN c_bpartner_location org_bpl
+                       ON org_bpl.c_bpartner_location_ID = oi.orgbp_location_id AND org_bpl.isActive = 'Y'
```
metas-ts, line 79: I think there shouldn't be an IsActive=Y condition. There is just one org_bpl, and it's needed as per the parameter.
=> on a quick glimpse, this applied to almost all `.isActive = 'Y'`conditions that we have here

Btw, can you also please prepend `p_`to the parameters and then replace the `$n` with param-names?

metas-ts (changes requested):
please make this old function a bit more robust (boyscout-rule)

backend/metasfresh-dist/metasfresh-webui-api-remote-via-ssh-tunnel.launch
```diff
@@ -12,7 +12,7 @@
 <booleanAttribute key="org.eclipse.jdt.launching.ALLOW_TERMINATE" value="false"/>
 <mapAttribute key="org.eclipse.jdt.launching.CONNECT_MAP">
 <mapEntry key="hostname" value="localhost"/>
-<mapEntry key="port" value="18789"/>
+<mapEntry key="port" value="8789"/>
```
metas-ts, line 15: i recommend to stick with 18789 and rather adopt your ssh config...but eiother way, ok with me

metas-ts (commented):
generelly lgtm, but 
th e parameters can't be renamed like this:
```
12:18:35  [INFO ] ScriptsApplier - Applying LocalScript(projectName=75-de.metas.fresh.reports, fileName=5594360_sys_AddGLN_ToAddresInFunctionsForDocs.sql, type=ScriptType(fileExtension=sql), localFile=/tmp/dist/sql/75-de.metas.fresh.reports/5594360_sys_AddGLN_ToAddresInFunctionsForDocs.sql, lastDurationMillis=-1)
12:18:35  ERROR:  cannot change name of input parameter "org_id"
```

metas-ts (approved):

Issue #11386, opened 2021-06-25
webui frontend: show last computed text to target indicators


petrican:
### Is this a bug or feature request?
F

PR #11401, opened 2021-06-28
11386 fuzzy time

petrican:
https://www.loom.com/share/a596da476cc54096b7345c7c732cd670

Screenshot 2021-06-28 at 14 20 00
Screenshot 2021-06-28 at 14 25 30

https://github.com/metasfresh/metasfresh/issues/11386


frontend/src/components/charts/Indicator.js
```diff
@@ -7,6 +7,34 @@ import moment from 'moment';
...
+      .add(1, 'seconds')
+      .format('hh:mm A');
+    return this.setState({ localComputedTimestamp: newTime });
+  };
+
+  /**
+   * @summary lifecycle in which we are calling the funtion that updates the local timestamp on one second
+   */
+  componentDidMount() {
+    this.interval = setInterval(() => this.updateTimestamp(), 1000);
```
teosarca, line 37: IMHO updating each minute is enough.
1sec it's too eager and brings no value for the user.

petrican: Ok I can change that

frontend/src/components/charts/Indicator.js
```diff
@@ -7,6 +7,34 @@ import moment from 'moment';
...
+    return { localComputedTimestamp: computedTimestamp };
+  }
+
+  /**
+   * @summary sets the local timestamp + 1 second, called on one second interval after the component mounts
+   */
+  updateTimestamp = () => {
+    const { localComputedTimestamp } = this.state;
+    const newTime = moment(localComputedTimestamp)
+      .add(1, 'seconds')
```
teosarca, line 28: imho this workaroundish code is not needed and looks ugly.

petrican: why you say this is workaroundish ? I changed to update the localComputerTimestamp with one min

petrican:
@teosarca pls let me know if this can be merged after I done the last fixes

petrican:
This is how the box will look like. Details is smaller and the refresh icon too
Screenshot 2021-06-28 at 15 15 19

frontend/src/components/charts/Indicator.js
```diff
@@ -7,6 +7,34 @@ import moment from 'moment';
...
+   * @returns
+   */
+  static getDerivedStateFromProps({ data: { computedTimestamp } }) {
+    return { localComputedTimestamp: computedTimestamp };
+  }
+
+  /**
+   * @summary sets the local timestamp + 1 min, called on one min interval after the component mounts
+   */
+  updateTimestamp = () => {
```
teosarca, line 25: @siemiatj  do u have any suggestion how can we improve the clarify of this code because IMHO I find it  quite weird.

IMHO we shall not increment localComputedTimestamp but just set it when we get it from WS.

render() will just display the fuzzy time elapsed from `localComputedTimestamp` until `now`.

But in this case I think there is something more we have to do. We have to trigger somehow the component render. Not sure if we can do this programmatically....
Another idea would be to introduce some "counter" into the state and increment the counter from updateTimestamp() function.

I hope i made my concern clear.
@petrican  @siemiatj  wdyt?

teosarca: PS: also, current impl is very error phrone. at least the refresh interval (i.e. 1min) shall be extracted to a constant and used in updateTimestamp and componentDidMount.

but pls don't do this one. let's focus to get a better and cleaner code.

petrican: @teosarca 

1) "IMHO we shall not increment localComputedTimestamp but just set it when we get it from WS." - the BE is not sending the data on every minute via WS. If it were to be sent on the FE side we should not need to have any incrementation of the computed timestamp
Screenshot 2021-06-29 at 07 06 44
Note: when the data is present via WS that localComputedTimestamp is also updated with the computedTimestamp within the data(from props).


2) "render() will just display the fuzzy time elapsed from localComputedTimestamp until now." - this is what it actually happens. I am using fromNow of the moment lib -> https://momentjs.com/docs/#/displaying/fromnow/ 
Screenshot 2021-06-29 at 07 11 21. And the good part is already translated.

3) "But in this case I think there is something more we have to do. We have to trigger somehow the component render. Not sure if we can do this programmatically...." - The component state is updated at the end of the updateTimestamp function when localComputedTimestamp is updated. That change in the state is triggering the component to render after that update took place.

4) "Another idea would be to introduce some "counter" into the state and increment the counter from updateTimestamp() function." - this "counter" in my implementation is the actual timestamp which is incremented every minute with a min.
Screenshot 2021-06-29 at 07 22 22

@siemiatj I am open for other ideas. Currently I don't have other better solutions that this. :(

teosarca: @petrican  pushed 2 commits to show you what I ment

PR #11414, opened 2021-06-30
11413 errors rendering

petrican:
Screenshot 2021-06-30 at 10 01 26


teosarca:
@petrican  code looks good.

Regarding the layout:
image

Can u pls:
* align it with the last computed text
* make sure they have the same size
?

PR #11434, opened 2021-07-02
Mass shipment: not all shipments created bc of problem with oldTopLevelHuId

metas-ts:


metas-ts (commented):
dev-tested locally & works

Issue #11426, opened 2021-07-01
Product attachments should be visible, even if attached to parent products


adi-stefan:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
- User wants to be able to define relationships between products (currently only parent-child). Discussed with Mark that we only consider 2 levels, so no transitivity concerns.
- User wants to be able to see & download attachments uploaded to both this record and its parent (if exists)

PR #11435, opened 2021-07-02
#gh11426: created M_Product_Relationship and M_Product_AttachmentEntry_ReferencedRecord_v

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/11426


metas-ts (approved):

metas-ts (commented):
one think that popped into my mind only later:
Do we need something like `AD_SQLColumn_SourceTableColumn`
So that the view is updated when a new attachment ist added?

I guess if teo is too busy right now, we can clarify this tomorrow and integrate today, but please macke sure to clarify with him

Issue #11438, opened 2021-07-02
`01. Suchauswahl abw. Werte setzen` should either set both bpartner & location, or none at all


adi-stefan:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Currently the user is able to override a bpartner while retaining the original location. This might result in inconsistencies when the OL_Cand is processed.
### What is the expected or desired behavior?
- If bpartner override is provided, then a location needs to also be provided.
- If both bpartner and location are provided, then Ad_User_ID should set to null.

PR #11439, opened 2021-07-02
#gh11438: If bpartner override is provided, then a location needs to also be provided. If both bpartner and location are provided, then Ad_User_ID should set to null.

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/11438

### DIFF
--- a/backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/process/OLCandSetOverrideValues.java
+++ b/backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/process/OLCandSetOverrideValues.java
@@ -48,7 +48,6 @@ public class OLCandSetOverrideValues extends JavaProcess
 {
 	private final static String PARAM_BPartner = I_C_OLCand.COLUMNNAME_C_BPartner_Override_ID;
 	private final static String PARAM_Location = I_C_OLCand.COLUMNNAME_C_BP_Location_Override_ID;
-	private final static String PARAM_User = I_C_OLCand.COLUMNNAME_AD_User_ID;
 
 	private IParams params = null;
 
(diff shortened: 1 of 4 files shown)


backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/process/OLCandSetOverrideValues.java
```diff
@@ -22,30 +22,51 @@
...
 	private IParams params = null;
 
 	@Override
 	protected void prepare()
 	{
-		this.params = getParameterAsIParams();
+		final List<ProcessInfoParameter> parameterList = new ArrayList<>(getProcessInfo().getParameter());
+		final Map<String, ProcessInfoParameter> parameters = parameterList.stream().collect(Collectors.toMap(ProcessInfoParameter::getParameterName, param -> param));
+
+		if (!Check.isBlank(parameters.get(PARAM_BPartner).getParameterAsString()))
```
metas-ts, line 61:
```suggestion
		if (Check.isNotBlank(parameters.get(PARAM_BPartner).getParameterAsString()))
```

metas-ts (changes requested):
I can't see how this is supposed to work;
AFAIU: this process runs `IOLCandUpdateBL#updateOLCands` which calls `OLCandUpdater#process`. In that method there is no `AD_User_ID` parameter.

instead of adding one, i suggest to change `IOLCandUpdateBL#updateOLCands` to
* see if C_OLCand.AD_User_ID is still "compatible" with the override-bpartner and location. if not, only then set it to null and all a Loggables.withLogger(Level.INFO) log message.

metas-ts (approved):

PR #11443, opened 2021-07-05
Payment Discount not allocated when using Prepay Order fixed

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/11442

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/payment/C_Payment.java
+++ b/backend/de.metas.business/src/main/java/de/metas/payment/C_Payment.java
@@ -1,0 +1,103 @@
+/*
+ * #%L
+ * de.metas.business
+ * %%
+ * Copyright (C) 2021 metas GmbH
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+package de.metas.payment;
+
+import de.metas.currency.Currency;
+import de.metas.currency.CurrencyRepository;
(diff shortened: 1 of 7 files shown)


backend/de.metas.business/src/main/java/de/metas/payment/C_Payment.java
```diff
@@ -0,0 +1,103 @@
...
+		final Percent paymentTermDiscountPercent = paymentTermRepository.getPaymentTermDiscount(paymentTermId);
+		final Currency currency = currencyRepository.getById(record.getC_Currency_ID());
+
+		final BigDecimal priceActual = paymentTermDiscountPercent.subtractFromBase(order.getGrandTotal(), currency.getPrecision().toInt());
+		final BigDecimal discountAmount = paymentTermDiscountPercent.computePercentageOf(order.getGrandTotal(), currency.getPrecision().toInt());
+
+		record.setC_Invoice(null);
+		record.setC_Charge_ID(0);
+		record.setIsPrepayment(true);
+		record.setDiscountAmt(BigDecimal.ZERO);
```
metas-ts, line 93:
```suggestion
```

backend/de.metas.business/src/main/java/de/metas/payment/C_Payment.java
```diff
@@ -0,0 +1,103 @@
...
+		this.currencyRepository = currencyRepository;
+	}
+
+	@Init
+	public void registerCallout()
+	{
+		Services.get(IProgramaticCalloutProvider.class).registerAnnotatedCallout(this);
+	}
+
+	@CalloutMethod(columnNames = I_C_Payment.COLUMNNAME_C_Order_ID)
```
metas-ts, line 68:
```suggestion
	/**
	  * Noteth that because of C_Payment.C_Order_ID's validation rule we can one have only prepay orders in this field.
	  * Before the payment is made, the prepay-order is not even completed, so the payment's date is effectively also the order's date.
 	  * So in case of a prepay order there is no need to take the term's DiscountDays into account.
      */
	@CalloutMethod(columnNames = I_C_Payment.COLUMNNAME_C_Order_ID)
```

backend/de.metas.business/src/main/java/de/metas/payment/C_Payment.java
```diff
@@ -0,0 +1,103 @@
...
+	public void updateDiscountAndPaymentAmount(@NonNull final I_C_Payment record)
```
metas-ts, line 69: suggestion: also make this a `@ModelChange` method

metas-ts (changes requested):
nice work :pray:
please check my minor remarks

metas-ts (approved):

Issue #11444, opened 2021-07-05
Investigate/fix the issue with the loading not showing while filtering is pending


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
When using Barcode to filter and is taking longer there is no pending shown on UI
#### Which are the steps to reproduce?
When using Barcode to filter and is taking longer there is no pending shown on UI
### What is the expected or desired behavior?
Pending shouyld be visible while the BE request is querying the info

PR #11446, opened 2021-07-05
Fix triggering spinner logic

petrican:
https://www.loom.com/share/711cc5b789204b4dbce397c1daca56f5

Screenshot 2021-07-05 at 16 15 51

If this fix is fine it should be CPed to scary_plum_uat also


siemiatj (approved):

metas-Kay:
related to https://github.com/metasfresh/metasfresh/issues/11444

PR #11456, opened 2021-07-06
mf connect to ebay

fp-ps:


metas-ts:
Hi there @wgaulke (cc @fp-ps),
team at @metasfresh/pevesoft is about to finish the metasfresh part of things.
They made changes also in the camel-part, so that it takes care of getting the API-Token. That way, metasfresh itself can leave accessing eBay entirely to camel.

Can you please take a look? 
Meanwhile i'm looking into that build problem you mentioned..

metas-ts:
@wgaulke one question:
The field names are taken from the ebay-webui. but we are not sure if the names are the same when looking at them as a "german" person..we didn'T find a way to set the language to DE in that UI..so are the field names always like this?
image

background: it's important to me to make sure the user has no problems recognizing which metasfresh field is which in ebay..

wgaulke:
@metas-ts @fp-ps 

Changes look good to me, thank you. Also I thought user interaction to get a token is a must. Good to see that there seems to be a way without user interaction.

Regarding Naming: I did not find a way to change the language for the ebay developer center, so I think they are fixed to english. See the naming in the screenshot:

Screenshot 2021-07-07 at 08 42 44

Issue #11458, opened 2021-07-06
Create C_BP_Relation.IsBranchOffice and display it in Partner Relation window


BodorAnnamaria:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
- create column C_BP_Relation 
- display column in window 313 under flag `Aufladeort `

PR #11459, opened 2021-07-06
#11458 Create C_BP_Relation.IsBranchOffice and display it in PartnerRelation window

BodorAnnamaria:
https://github.com/metasfresh/metasfresh/issues/11458


metas-ts (changes requested):
sorry, i didn't think of it when i reviewed the concept, but ofc we also need an `AD_SQLColumn_SourceTableColumn` record for the new SQL-column, to make sure the bpartner window refreshed when the bp_relation changes

metas-ts (approved):

PR #11464, opened 2021-07-07
Gh11436 supplier approval

metas-rc:
https://github.com/metasfresh/metasfresh/issues/11436


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/BPartnerSupplierApprovalRepository.java
```diff
@@ -0,0 +1,119 @@
...
+		bpSupplierApprovalRecord.setSupplierApproval_Date(supplierApprovalDate);
+
+		save(bpSupplierApprovalRecord);
+
+	}
+
+	public ImmutableList<I_C_BP_SupplierApproval> retrieveBPSupplierApprovalsAboutToExpire()
+	{
+		final LocalDate today = SystemTime.asLocalDate();
+		final LocalDate threeMonthsFromNow = today.plusMonths(3);
```
metas-ts, line 82: please make the month-interval `3` a parameter

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/BPartnerSupplierApprovalRepository.java
```diff
@@ -0,0 +1,119 @@
...
+import javax.annotation.Nullable;
+import java.sql.Timestamp;
+import java.time.Instant;
+import java.time.LocalDate;
+
+import static org.adempiere.model.InterfaceWrapperHelper.load;
+import static org.adempiere.model.InterfaceWrapperHelper.save;
+
+@Repository
+public class BPartnerSupplierApprovalRepository
```
metas-ts, line 45: Note: to me this is not a repository, but a DAO because it doesn't work with domain objects. 
But i might be too puritanic here, so OK for me
Anyways, next time please consider introducting a SupplierApproval domain object and avoid directly working with generated model classes.

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/BPartnerSupplierApprovalRepository.java
metas-rc: Added a constant for it and propagated it as parameter. 
Maybe, if somebody ever needs another number of months, we should set it in some kind of ``C_BP_SupplierApproval_Notification_Config`` structure.

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/BPartnerSupplierApprovalRepository.java
metas-rc: Thank you. Will keep in mind.

metas-ts (approved):

Issue #11470, opened 2021-07-08
`PP_Order_RecordWork` should support decimal duration values


adi-stefan:
### Is this a bug or feature request?
FR
### What is the current behavior?
Currently `PP_Order_RecordWork` allows decimal values, but always rounds that value up. So 0.5h => 1h
### What is the expected or desired behavior?
There should be no rounding, 0.5h => 30m

PR #11471, opened 2021-07-08
#gh11470: added support for converting decimal durations into integer durations of a smaller TemporalUnit

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/11470

### DIFF
--- a/backend/de.metas.util/src/main/java/de/metas/util/time/DurationUtils.java
+++ b/backend/de.metas.util/src/main/java/de/metas/util/time/DurationUtils.java
@@ -2,6 +2,7 @@ package de.metas.util.time;
 
 import com.google.common.collect.ImmutableList;
 import de.metas.util.Check;
+import de.metas.util.NumberUtils;
 import de.metas.util.StringUtils;
 import lombok.NonNull;
 import lombok.experimental.UtilityClass;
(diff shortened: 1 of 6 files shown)


backend/de.metas.util/src/main/java/de/metas/util/time/DurationUtils.java
```diff
@@ -35,10 +38,32 @@
...
+		BigDecimal duration = durationBD;
+		TemporalUnit currentUnit = unit;
+		long durationLong = -1;
+		while (durationLong < 0)
+		{
+			try
+			{
+				durationLong = duration.longValueExact();
+			}
+			catch (final ArithmeticException ae)
```
teosarca, line 54: avoid using exceptions to control the flow.

pls check if `NumberUtils.stripTrailingDecimalZeros(duration).scale() == 0` might do the job.

backend/de.metas.util/src/main/java/de/metas/util/time/DurationUtils.java
```diff
@@ -75,4 +100,25 @@ else if (unit == ChronoUnit.DAYS)
 		}
 	}
 
+	private static BigDecimal getEquivalentInSmallerTemporalUnit(@NonNull final BigDecimal durationBD, @NonNull final TemporalUnit unit)
+	{
+		if (unit == ChronoUnit.DAYS)
+		{
+			return durationBD.multiply(BigDecimal.valueOf(8));// This refers to work hours, not calendar hours
```
teosarca, line 107: why work hours and not simply hours?

teosarca: i.e. multiply by 24.

teosarca (changes requested):
@adi-stefan  @metas-ts  check my comments. wdyt?

backend/de.metas.util/src/main/java/de/metas/util/time/DurationUtils.java
adi-stefan: These calculations are used in a work hours context, that's why I also changed the method name: `toDuration` -> `toWorkDuration`.

backend/de.metas.util/src/main/java/de/metas/util/time/DurationUtils.java
metas-ts, line 54: don't lean on exception. instead (found in `de.metas.util.NumberUtils#stripTrailingDecimalZeros`), i think you can do 
`stripTrailingZeros()` and then check the scale

backend/de.metas.util/src/main/java/de/metas/util/time/DurationUtils.java
```diff
@@ -35,10 +38,32 @@
...
 {
-	public static Duration toDuration(@NonNull final BigDecimal durationBD, @NonNull final TemporalUnit unit)
+	private static final List<TemporalUnit> supportedTemporalUnits = ImmutableList.of(ChronoUnit.YEARS, ChronoUnit.MONTHS, ChronoUnit.DAYS, ChronoUnit.HOURS, ChronoUnit.MINUTES, ChronoUnit.SECONDS, ChronoUnit.NANOS);
+
+	public static Duration toWorkDuration(final @NonNull BigDecimal durationBD, final @NonNull TemporalUnit unit)
+	{
+		BigDecimal duration = durationBD;
+		TemporalUnit currentUnit = unit;
+		long durationLong = -1;
+		while (durationLong < 0)
```
metas-ts, line 48: i'm a bit nervous about this while loop. what we we end up with durationLong < 0 and it never changes anymore?
why not instead have a for-loop 
or a counter? we can anyways reasonably just iterate as often as we have supportedTemporalUnits, right?

metas-ts: case in point: 
image

metas-ts (changes requested):

backend/de.metas.util/src/main/java/de/metas/util/time/DurationUtils.java
metas-ts: ..or `scale() <= 0`

backend/de.metas.util/src/main/java/de/metas/util/time/DurationUtils.java
```diff
@@ -35,10 +38,32 @@
 @UtilityClass
 public class DurationUtils
 {
-	public static Duration toDuration(@NonNull final BigDecimal durationBD, @NonNull final TemporalUnit unit)
+	private static final List<TemporalUnit> supportedTemporalUnits = ImmutableList.of(ChronoUnit.YEARS, ChronoUnit.MONTHS, ChronoUnit.DAYS, ChronoUnit.HOURS, ChronoUnit.MINUTES, ChronoUnit.SECONDS, ChronoUnit.NANOS);
+
+	public static Duration toWorkDuration(final @NonNull BigDecimal durationBD, final @NonNull TemporalUnit unit)
```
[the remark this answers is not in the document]
teosarca: and add a link to https://en.wikipedia.org/wiki/International_Workers%27_Day too :))

backend/de.metas.util/src/main/java/de/metas/util/time/DurationUtils.java
adi-stefan: makes sense, changed.

backend/de.metas.util/src/main/java/de/metas/util/time/DurationUtils.java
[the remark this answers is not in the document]
adi-stefan: defined, and used in ` UOMConversionBL#getTimeConversionRateAsBigDecimal`

backend/de.metas.util/src/main/java/de/metas/util/time/DurationUtils.java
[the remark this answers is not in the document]
adi-stefan: Did my best to explain "work".

backend/de.metas.util/src/main/java/de/metas/util/time/DurationUtils.java
adi-stefan: changed.

backend/de.metas.util/src/main/java/de/metas/util/time/DurationUtils.java
adi-stefan: Changed to prevent negative duration from being passed in, and also added a second hard stop condition regarding the ChronoUnit.

metas-ts (approved):

PR #11499, opened 2021-07-14
make sure that also inoutLine-only ICs have a C_DocTypeInvoice_ID

metas-ts:


metas-ts (commented):
lgtm; also tested by Annamaria

PR #11505, opened 2021-07-15
div. Alberta Tweaks

metas-ts:
* document-API: 
  * query with the last document's updatedAt as opposed to createdAt
  * assume the document to be always PDF
* product-API: fix a bug in metasfresh


metas-ts (commented):
lgtm

Issue #11508, opened 2021-07-16
Implement ESR payment action - duplicate payment


cristinamghita:
### Is this a bug or feature request?
FR

PR #11511, opened 2021-07-16
#11508 Implement ESR payment action - duplicate payment

cristinamghita:
#11508 Implement ESR payment action - duplicate payment


backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/api/impl/ESRImportDAO.java
```diff
@@ -367,4 +379,27 @@ public I_ESR_ImportLine fetchLineForESRLineText(@NonNull final I_ESR_Import esrI
 				.collect(ImmutableSet.toImmutableSet());
 		return notReconciledESRImportIds;
 	}
+	
+	@Override
+	public Set<PaymentId> findExistentPaymentIds(@NonNull final I_ESR_ImportLine esrLine)
```
metas-ts, line 384: if the method will return max 1 id, i better to make it return Optional or `..OrNull`

metas-ts (approved):

PR #11532, opened 2021-07-21
Tell camel from metasfresh to open http-EP with particular auth-token

fp-ps:


backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/woocommerce/housekeeping/ExternalSystemWooCommerceHouseKeepingTask.java
```diff
@@ -0,0 +1,81 @@
...
+		if (processId == null)
+		{
+			Loggables.withLogger(logger, Level.DEBUG).addLog("Nothing to do!");
+			return;
+
+		}
+
+		final ImmutableList<ExternalSystemParentConfig> parentConfigList = externalSystemConfigDAO.getAllByType(ExternalSystemType.WOO);
+
+		parentConfigList.forEach((config -> ProcessInfo.builder()
```
metas-ts, line 73: please also do a `Loggables.withLogger(logger, Level.DEBUG).addLog("...");` for each for the woocommerce-config that you fire up

metas-ts (changes requested):

backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/woocommerce/housekeeping/ExternalSystemWooCommerceHouseKeepingTask.java
fp-ps: logger added

metas-ts (approved):

PR #11538, opened 2021-07-23
DB-Function for customizable DB-changes on DB-transfer

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/11535


backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/ddl/functions/scramble_metasfresh.sql
```diff
@@ -1,7 +1,7 @@
 
 --## Function to scramble all metasfresh-tables
 
-CREATE OR REPLACE FUNCTION public.scramble_metasfresh(
+CREATE OR REPLACE FUNCTION ops.scramble_metasfresh(
```
metas-ts, line 4: Please add a folder `src/main/sql/postgresql/ddl/public` for and move the existing stuff to it.
Then add a folder `src/main/sql/postgresql/ddl/ops/functions` and move the scramble stuff for that.
(I.e. one folder per schema)

metas-ts (changes requested):

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/ddl/functions/scramble_metasfresh.sql
dm-ps: done

metas-ts (approved):

Issue #10831, opened 2021-03-17
Update frontend vulnerable  packages


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
some packages are vulnerable
#### Which are the steps to reproduce?
npm install
### What is the expected or desired behavior?
no vuln packages at the end of running npm install

petrican:
Note -> https://github.com/metasfresh/metasfresh/pull/10844 

Immer upgrade breaks things and had to revert it. (breadcrumbs)

PR #11542, opened 2021-07-23
Update `react-tether` to React 17 compatible

siemiatj:
Related to #10831 #11554

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -180,7 +180,7 @@
     "react-router": "5.2.0",
     "react-router-dom": "5.2.0",
     "react-tagsinput": "3.19.0",
-    "react-tether": "1.0.4",
+    "react-tether": "2.0.8",
     "react-transition-group": "4.4.1",
     "react-translate-component": "0.15.1",
     "redux": "4.1.0",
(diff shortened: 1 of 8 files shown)


petrican:
The build is failing

e2e/cypress/support/index.js
```diff
@@ -41,14 +42,28 @@ Cypress.on('window:alert', (text) => {
...
 
 before(function () {
-  const autoLogin = function () {
-    return cy.loginViaForm();
-  };
-  autoLogin().then((msg) => {
-    cy.log(msg);
+  cy.clearLocalStorageSnapshot();
+
+  cy.loginViaAPI().then(() => {
```
petrican, line 47: this loginViaAPI was causing tests to fail randomly. This is the reason I switched to loginViaForm. Have you tested this roughly (like running the entire e2e suite 4 times and don't have any test failed as a consequence of login) ?

petrican (changes requested):

e2e/cypress/support/index.js
siemiatj: I've tried loginViaForm and it's also flaky as it's always about how fast the test is running, how fast the API is responding and the timeouts we set. And API authentication is way more reliable than going through the form.

PR #11544, opened 2021-07-24
clean up the legacy MRP Product Info code #11543

metas-ts:
it's unused for years now, but still creates and processes async-workpackages to keep it all up to date


metas-ts (commented):
lgtm

PR #11564, opened 2021-07-30
Training cypress

petrican:
- Adds the component testing on top of our existing e2e structure
- Besides the `npm run cypress:open` which opens the e2e test runner you can now also 
- Start the test runner for Component Testing  with `npm run cypress:open-ct`
- Integration tests for the components can be written (they need to reside under `e2e\cypress\integration\components`) and the components will be rendered and tested in browser
- "Component Test Runner is separate from the Cypress End-to-End (E2E) Test Runner and they can be used either separately or together. They share the same driver, server, and commands. The primary difference is that with E2E you visit a webserver with cy.visit() and with component testing you mount your components directly with mount."
- Motivation why this was added
- 
Screenshot 2021-07-30 at 12 08 24
(more info on https://www.cypress.io/blog/2021/04/06/introducing-the-cypress-component-test-runner/)

How the test runner will look like:
Screenshot 2021-07-30 at 12 10 49


siemiatj (approved):

petrican:
I'll merge this after I do some more test runs within Jenkins

Issue #11571, opened 2021-08-03
Packages update August 2021


petrican:
### Is this a bug or feature request?
F
### What is the current behavior?
Some packages need to be updated
#### Which are the steps to reproduce?
package.json
### What is the expected or desired behavior?
Have packages updated, especially the vulnerable ones without any impact in the app

PR #11577, opened 2021-08-04
11571 pkg update aug

petrican:
- updates some deprecated packages but most importantly is it doing the transition to webpack 5
- can be tested locally with either `npm run start` or by doing a `npm run build-prod`
- also replaced the webpack-git-hash (5 years since it wasn't updated - and of course incompatible with webpack version 5
- will update some more packages in next iterations


siemiatj (approved):
Looks good !

metas-Kay:
related to https://github.com/metasfresh/metasfresh/issues/11571

Issue #11575, opened 2021-08-03
Remove all uses of immutable.js


siemiatj:
### Is this a bug or feature request?

Feature

PR #11584, opened 2021-08-05
Remove immutable.js completely

siemiatj:
Related to #11575

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -144,7 +144,6 @@
     "history": "4.10.1",
     "immer": "9.0.1",
     "immutability-helper": "2.1.2",
-    "immutable": "4.0.0-rc.12",
     "intro.js": "2.9.3",
     "intro.js-react": "0.3.0",
     "lodash": "4.17.21",
--- a/frontend/src/__tests__/actions/WindowActions.test.js
+++ b/frontend/src/__tests__/actions/WindowActions.test.js
@@ -1,7 +1,6 @@
 import thunk from 'redux-thunk';
 import nock from 'nock';
 import configureStore from 'redux-mock-store';
-import { Set } from 'immutable';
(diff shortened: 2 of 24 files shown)


petrican (changes requested):
- please also remove the immutable from the package.json such that it won't be bundled anymore
- just a thought: Won't own implementation of List serve better to the purpose immutable.js initially provided instead of using just the array ? 
Overall looks good. Pls test this roughly by deploying a dev instance and running e2e/manual tests over it before merging. 
Thanks

update: (Didn't spot the ran e2e suite initially and the dev instance. After checking again this morning I saw a bunch failing so there still things to be fixed).

siemiatj:
>     * please also remove the immutable from the package.json such that it won't be bundled anymore
> 
>     * just a thought: Won't own implementation of List serve better to the purpose immutable.js initially provided instead of using just the array ?
>       Overall looks good. Pls test this roughly by deploying a dev instance and running e2e/manual tests over it before merging.
>       Thanks

I don't really think there's a need for that. It's hard to be quicker than the built-in objects, we don't mix types in arrays and use only simple operations. The motivation behind using immutable was to have an easier way of testing prev/next values for changes without a have `toString` or `JSON` operations. But now, since pretty much everything is in the store, we get the updates automatically only when they actually happened. In other places where for some reason immutable checks didn't work, we've had a fallback to using unique hashes. And it stayed that way for `state` stored values if needed.

> 
> update: (Didn't spot the ran e2e suite initially and the dev instance. After checking again this morning I saw a bunch failing so there still things to be fixed).

No idea what Jenkins broke with this dev instance, but it's fine now: https://jenkins.metasfresh.com/job/ops/job/run_e2e_tests/1367/

petrican:
lgtm. Merging it. Thanks

PR #11596, opened 2021-08-09
fix duplicate contracts

cp-ps:
refs: https://github.com/metasfresh/metasfresh/issues/11488

### DIFF
--- a/backend/de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
+++ b/backend/de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
@@ -25,8 +25,11 @@ package de.metas.contracts.impl;
 import ch.qos.logback.classic.Level;
 import de.metas.acct.api.IProductAcctDAO;
 import de.metas.bpartner.service.IBPartnerDAO;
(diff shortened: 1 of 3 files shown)


metas-ts (commented):
i don't really understand the fix...
particulary the cache-invalidation-request should not be neccesary.

does this fix prevent the problem? i didn't see any test docu.
Btw, also note that this should go to `blonde_monkey_uat`..we can CP it there as well, but in your  `blonde_monkey_uat` dev-instance you would las have existing data to dev-test with.

cp-ps:
@metas-ts I can document dev tests no problem... I don't really get why you're saying the `cache-invalidation-request` it's not necessary, basically the result of the method loading the `c_flatrate_term` records, i.e. `de.metas.contracts.impl.FlatrateDAO#retrieveTerms(java.util.Properties, de.metas.organization.OrgId, com.google.common.collect.ImmutableList<java.lang.Integer>, java.sql.Timestamp, int, int, int, java.lang.String)` is cached.. so any new contract gets ignored until the cache is refreshed

metas-ts:
@cp-ps right, i think i got it now :+1:

PR #11610, opened 2021-08-12
Frontend routing fixes

siemiatj:
Related to #11015


frontend/src/hooks/useAuth.js
```diff
@@ -47,24 +52,40 @@ function useProvideAuth() {
...
+        } else {
+          _logoutSuccess();
+        }
 
-      return Promise.resolve(resp.data);
-    });
+        return Promise.resolve(resp.data);
+      })
+      .catch((error) => {
+        if (error.response.status !== 500) {
```
petrican, line 68: What would happen if error is 500 ?

petrican (commented):
I have just a question regarding how we'll deal with 500 err. Other than that lgtm.

PR #11613, opened 2021-08-12
Process to sync products from shopware to metasfresh

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/11612


metas-ts (commented):
i think this is fine as a first increment. let's the the other things that you meantioned in a follow-up.
can you please just add the trl for "getProducts" which i meanwhile spezified?

metas-ts (approved):

PR #11616, opened 2021-08-16
11615 fields bpartner status

petrican:
Screenshot 2021-08-16 at 12 51 00

Flag single filed from the composite field when this is mandatory and is invalidated by the backend.

Demo
https://www.loom.com/share/c925c4c674f3415d92b09d3e46c22ef0


siemiatj (approved):
Looks good

PR #11621, opened 2021-08-17
Gh11617 option to always allow new bPartner

metas-rc:
https://github.com/metasfresh/metasfresh/issues/11617


siemiatj (approved):
FE OK

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/session/UserSession.java
```diff
@@ -477,6 +477,12 @@ public int getHttpCacheMaxAge()
 		};
 	}
 
+	private static final String SYSCONFIG_isAlwaysShowNewBPartner = "de.metas.ui.web.session.UserSession.IsAlwaysDisplayNewBPartner";
```
metas-ts, line 480: * suggestion: stick with "Display" instead of "Show"
* please add a javadoc which indicates that this is about an item / option thingie in the UI (AFAIU it is, right?)

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/address/AddressRestController.java
```diff
@@ -128,7 +128,7 @@ public JSONLookupValuesPage getAttributeTypeahead(
 
 		return addressRepo.getAddressDocumentForReading(docId)
 				.getFieldLookupValuesForQuery(attributeName, query)
-				.transform(page -> JSONLookupValuesPage.of(page, userSession.getAD_Language()));
+				.transform(page -> JSONLookupValuesPage.of(page, userSession.getAD_Language(),false));
```
metas-ts, line 131: instead of adding `, false` everywhere, i think we should add a method `JSONLookupValuesPage.of` without this parameter and there delegate to `JSONLookupValuesPage.of(page, language, false)`

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/datatypes/json/JSONLookupValuesPage.java
```diff
@@ -44,24 +44,31 @@
...
 	}
 
 	private JSONLookupValuesPage()
 	{
 		values = ImmutableList.of();
 		hasMoreResults = false;
+		isAlwaysDisplayNewBPartner = false;
 	}
 
 	public static JSONLookupValuesPage of(
```
metas-ts, line 68:
```suggestion
	public static JSONLookupValuesPage of(
			@Nullable final LookupValuesPage page,
			@NonNull final String adLanguage)
	{
		return of(page, adLanguage, false);
	}

	public static JSONLookupValuesPage of(
```

metas-ts (changes requested):

metas-ts (approved):

PR #11646, opened 2021-08-23
11627 fixed size popup

petrican:


metas-ts (approved):
The JS ofc doesn't tell me much, but according to the demo it's OK.
I think that as a follow-up we need a text label that sais something like "no references" though

Issue #11712, opened 2021-09-10
Filter in Picking Terminal v2 is not working 


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
Clicking on filter leads to white screen
#### Which are the steps to reproduce?
Filter in Picking Terminal v2 is not working . Clicking on filter leads to white screen
### What is the expected or desired behavior?
It should open the filter

PR #11714, opened 2021-09-10
11712 filter picking terminalv2

petrican:
https://www.loom.com/share/30975a7706cc4ae2b718a3216d59c0cc rel to https://github.com/metasfresh/metasfresh/issues/11712


petrican:
this is not a solution. This seems to be introduced most probable by react-tether upgrade. Going to close this

Issue #11729, opened 2021-09-14
Changes in advanced edit not reflected in the underlying view


siemiatj:
### Is this a bug or feature request?

Bug

PR #11736, opened 2021-09-16
Fix checkboxes's communication between the layers

siemiatj:
Related to #11729


petrican (changes requested):
Screenshot 2021-09-16 at 07 49 30
This is affecting the default filters logic for filters. Look how it is now on releasetestit:
Screenshot 2021-09-16 at 07 54 54

siemiatj:
I thought they work the same in all filters but I guess not. Pushed a fix

petrican (approved):
lgtm. thanks

Issue #11744, opened 2021-09-17
Create Unique indexes for Fachrichtung tabels


BodorAnnamaria:
### Is this a bug or feature request?
FR

PR #11745, opened 2021-09-17
#11744 Create Unique indexes for Fachrichtung tabels

BodorAnnamaria:
https://github.com/metasfresh/metasfresh/issues/11744


metas-ts (changes requested):
it's good to anticipate and try to avoid rollout-problems, but we shouldn't skip over this migration-SQL. it's better to fail and solve the problem on the respective database

metas-ts (approved):

PR #11751, opened 2021-09-17
Update CONTRIBUTING.md

cbkurz:
Changed gitter chat reference to forum reference
Chat announces to be deprecated and to use the forum.


metas-ts (approved):
🙏 @cbkurz

PR #11788, opened 2021-09-27
Extend behavior for OLCand with sales-reps

fp-ps:


backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/salesrep/interceptor/C_Order.java
```diff
@@ -75,7 +72,9 @@ public void updateSalesPartnerFromCustomer(@NonNull final I_C_Order orderRecord,
 		// If on a new C_Order record both SalesPartnerCode and the BPartner-IDs were set at the same time,
 		// then don't override the sales-partner-id from the BPartners' mater data, but assume that the sales-partner-id shall remain the way it was set on the new record.
 		// This implies that the master data might be updated on thid C_Order's after-new modelinterceptor method
-		final boolean currentPartnerCodeShallPrevail = type.isNew() && Check.isNotBlank(orderRecord.getSalesPartnerCode());
+		final boolean currentPartnerCodeShallPrevail = type.isNew()
+				&& (Check.isNotBlank(orderRecord.getSalesPartnerCode())
+				|| orderRecord.getC_BPartner_SalesRep_ID() > 0);
```
metas-ts, line 77: ok, but what if a user enters a different SalesRepCode in the UI? I believe this would get broken by this change

metas-ts (changes requested):
please check my remark

PR #11790, opened 2021-09-27
Extend behavior for OLCand with sales-reps

fp-ps:


metas-ts (commented):
in the old PR we have this new block: [`6e0b399` (#11788)](https://github.com/metasfresh/metasfresh/pull/11788/commits/6e0b399963fe5120cf538170c9b6f7cd3f6fa45e#diff-4e6f07f10cb77fc78b029a066771d9cad0b313e35f69bfafc374e55cfe6231fcR79)

do we still need that in here?

backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/salesrep/interceptor/C_Order.java
```diff
@@ -75,7 +76,27 @@ public void updateSalesPartnerFromCustomer(@NonNull final I_C_Order orderRecord,
 		// If on a new C_Order record both SalesPartnerCode and the BPartner-IDs were set at the same time,
 		// then don't override the sales-partner-id from the BPartners' mater data, but assume that the sales-partner-id shall remain the way it was set on the new record.
 		// This implies that the master data might be updated on thid C_Order's after-new modelinterceptor method
-		final boolean currentPartnerCodeShallPrevail = type.isNew() && Check.isNotBlank(orderRecord.getSalesPartnerCode());
+		final BPartnerId salesRepId = BPartnerId.ofRepoIdOrNull(orderRecord.getC_BPartner_SalesRep_ID());
```
metas-ts, line 79: Sorry, i don't get it.
(To put it into context, i didn't understand the first proposed change either..i just thought that it broke something, so I didn't think into it too deeply).

Does this block somehow relate to the comment from lines 76-78?
pleas either make the code clearer or/and add more comments to understand what we are looking at here

fp-ps: If salesPartnerCode is defined we need to take into account also the salesRepId, otherwise it will end up overwriting the salesRepId set on order with the one from customer.

metas-ts (approved):

PR #11801, opened 2021-09-30
Invalidate API-cache when user-role-label is changed

fp-ps:


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/user/role/repository/UserRoleRepository.java
```diff
@@ -0,0 +1,102 @@
...
+			.<UserRoleId, I_C_User_Role>builder()
+			.tableName(I_C_User_Role.Table_Name)
+			.build();
+
+	@NonNull
+	public List<UserRole> getUserRoles(@NonNull final UserId userId)
+	{
+		final List<UserAssignedRoleId> assignedRoleIds = assignedUserRoleCache.getOrLoad(userId, this::getAssignedRoleIds);
+
+		return assignedRoleIds.stream()
```
metas-ts, line 60: it's ok, assumingn that we don't have 1000s of roles, still, this might have been a good place to use the getAllOrLoad() method that i used in the issue description

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v1/bpartner/bpartnercomposite/BPartnerCompositeCacheIndex.java
```diff
@@ -138,4 +148,32 @@ public BPartnerId extractDataItemId(@NonNull final BPartnerComposite dataItem)
 			return result;
 		}
 	}
+
+	@Override
+	public boolean isResetAll(@NonNull final TableRecordReference recordRef)
```
metas-ts, line 153: not 100% sure, but: if we now have `getRecordRefsRoles`, do we still need to return true if the given recordRef is a `_C_User_Assigned_Role` or `C_User_Role`?

metas-ts (commented):
please check my question; also, for the code alone i'm not sure if the case will also be proverly invalidated if you **remove** a role...please check that too if you didn't yet :-)

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v1/bpartner/bpartnercomposite/BPartnerCompositeCacheIndex.java
fp-ps: yes, we do... as in case of new role being assigned, that new `assigned_role_id` will not be part of the cached key (as it wasn't there before) thus the cache is not refreshed. this behavior is reproducing for every child record where the parent doesn't trigger a refresh when changes occur. @cp-ps will give you more info about this scenario.

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v1/bpartner/bpartnercomposite/BPartnerCompositeCacheIndex.java
```diff
@@ -138,4 +148,32 @@ public BPartnerId extractDataItemId(@NonNull final BPartnerComposite dataItem)
 			return result;
 		}
 	}
+
+	@Override
```
metas-ts, line 152:
```suggestion
	/**
	  * @return true if the given recordRef is a C_User_Assigned_Role or C_User_Role.
	  *    background: in case of new role being assigned, that new assigned_role_id will not be part of the cached key (as it wasn't there before) thus the cache is not refreshed.
	  */
	@Override
```

metas-ts (approved):

metas-ts:

Issue #11809, opened 2021-10-04
Window content doesn't change when opening another window 


petrican:
### Is this a bug or feature request?
b
### What is the current behavior?
Changing location the content does not refresh
#### Which are the steps to reproduce?
Go to Dashboard -> Sales order then to BP (listing in table remains from sales order)

PR #11813, opened 2021-10-04
Fix navigation from the header menu

siemiatj:
Related to #11809


siemiatj:
Fixed the main issue. Working on the filters fix now so not ready for merge yet

PR #11815, opened 2021-10-05
material cockipt

cp-ps:
refs: https://github.com/metasfresh/metasfresh/issues/11814


metas-ts (commented):
going to integrate now, review&improve later

PR #11823, opened 2021-10-05
Fix known commission-bugs

fp-ps:


backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/commissioninstance/services/hierarchy/CommissionHierarchyFactory.java
```diff
@@ -36,20 +36,36 @@
...
+ */
 @Service
 public class CommissionHierarchyFactory
 {
-	// very crude but simple implementation; expand and make more efficient as needed
-	public Hierarchy createFor(@NonNull final BPartnerId bPartnerId)
+	@NonNull
+	public Hierarchy createForCustomer(@NonNull final BPartnerId bPartnerId, @NonNull final BPartnerId salesRepId)
 	{
+		if (bPartnerId.equals(salesRepId))
```
metas-ts, line 48: not sure, this is just a hunch: why don't we recurse with the sales rep-id but directly return?

backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/commissioninstance/services/hierarchy/CommissionHierarchyFactory.java
```diff
@@ -36,20 +36,36 @@
...
-/** Crates a hierarchy of BPartners by recursively following {@code C_BPartner.C_BPartner_SalesRep_ID} references. */
+/**
+ * Crates a hierarchy of BPartners by recursively following {@code C_BPartner.C_BPartner_SalesRep_ID} references.
+ */
 @Service
 public class CommissionHierarchyFactory
 {
-	// very crude but simple implementation; expand and make more efficient as needed
-	public Hierarchy createFor(@NonNull final BPartnerId bPartnerId)
+	@NonNull
```
metas-ts, line 45:
```suggestion
	// very crude but simple implementation; expand and make more efficient as needed
	@NonNull
```

sorry to say, but imho it's still crude :-) as we load one bpartner after the other, instead of finding ways to load them all at once

backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/commissioninstance/services/hierarchy/CommissionHierarchyFactory.java
fp-ps: good point, I'll change the logic to support also this scenario. Did you got a change to look over the docu, I've addressed a question there.

metas-ts (approved):

Issue #11812, opened 2021-10-04
Implemented vendor attribute


cristinamghita:
### Is this a bug or feature request?
FR

PR #11830, opened 2021-10-07
#11812 Attribute vendor

cristinamghita:


cristinamghita:
@teosarca cleared all

PR #11870, opened 2021-10-18
extend externalsystems-camel to write an audit trail

fp-ps:


misc/services/camel/de-metas-camel-externalsystems/core/src/main/resources/application.properties
```diff
@@ -71,6 +71,7 @@ metasfresh.attachments.v2.api.uri={{metasfresh.api.v2.baseurl}}/attachment
 
 metasfresh.enqueue-purchases-candidate-v2.api.uri={{metasfresh.api.v2.baseurl}}/order/purchase/enqueueForOrdering
 
+metasfresh.seqno.v2.api.uri={{metasfresh.api.v2.baseurl}}/seqNo
```
metas-ts, line 74: please also remove the raw-data properties for which you removed the constants

misc/de-metas-common/de-metas-common-rest_api/src/main/java/de/metas/common/rest_api/v1/SwaggerDocConstants.java
```diff
@@ -71,4 +71,6 @@
 
 	public static final String PARENT_SYNC_ADVISE_DOC = "Defaults to the parent resource's sync advise, if not specified";
 
+	public static final String AD_SEQ_NO_ID_DOC = "AD_SeqNo_ID for the seqNo to be retrieved";
```
metas-ts, line 74: please add this to the v2 constants

metas-ts (changes requested):

metas-ts (approved):

metas-ts:
can now integrate this to `release` becuase right now we don't have an RC

PR #11910, opened 2021-11-01
Unregister service worker based on config

petrican:
Screenshot 2021-11-01 at 10 46 15

### DIFF
--- a/misc/services/mobile-webui/mobile-webui-frontend/src/serviceWorkerRegistration.js
+++ b/misc/services/mobile-webui/mobile-webui-frontend/src/serviceWorkerRegistration.js
@@ -10,6 +10,8 @@
 // To learn more about the benefits of this model and instructions on how to
 // opt-in, read https://cra.link/PWA
 
+import { REGISTER_SERVICE_WORKER } from './constants';
+
 const isLocalhost = Boolean(
   window.location.hostname === 'localhost' ||
     // [::1] is the IPv6 localhost address.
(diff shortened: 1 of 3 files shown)


misc/services/mobile-webui/mobile-webui-frontend/src/serviceWorkerRegistration.js
```diff
@@ -40,7 +42,9 @@ export function register(config) {
     } else {
       // Is not localhost. Just register service worker
       console.log('[ServiceWorker] - Registering on Prod server');
```
teosarca, line 44: i thikn that console.log shall go inside the `if `

misc/services/mobile-webui/mobile-webui-frontend/src/serviceWorkerRegistration.js
```diff
@@ -40,7 +42,9 @@ export function register(config) {
...
-      registerValidSW(swUrl, config);
+      if (REGISTER_SERVICE_WORKER) {
```
teosarca, line 45: why we need this here when we already have an `if (REGISTER_SERVICE_WORKER)` then call this `register` function.

teosarca:
@petrican  another question: how the code will behave in the following scenario:

* we already rolled out with REGISTER_SERVICE_WORKER=true
* now, we decide to disable service workers, wo we set REGISTER_SERVICE_WORKER=false and we do another rollout


Will it unregister currently registered service workers?

petrican:
> @petrican another question: how the code will behave in the following scenario:
> 
> * we already rolled out with REGISTER_SERVICE_WORKER=true
> * now, we decide to disable service workers, wo we set REGISTER_SERVICE_WORKER=false and we do another rollout
> 
> Will it unregister currently registered service workers?

Yes as there is a routine that unregisters those
Screenshot 2021-11-01 at 11 32 29

petrican:
You were right @teosarca I should do one single check and that within the register routine within the serviceWorkerRegistration. Fixed

teosarca:
@petrican  lgtm

PR #11923, opened 2021-11-02
Abort workflow button with prompt

siemiatj:
Modified the ConfirmButton to be more reusable. Also we couldn't confirm the workflow so fixed this as well and fetched the applications on app init.

Had to change target and got tons of conflicts so decided to just create a new PR. Here's the old one:
https://github.com/metasfresh/metasfresh/pull/11889

### DIFF
--- a/misc/services/mobile-webui/mobile-webui-frontend/src/reducers/headers.js
+++ b/misc/services/mobile-webui/mobile-webui-frontend/src/reducers/headers.js
@@ -10,6 +10,8 @@ export const initialState = {
   ],
 };
 
+const launchersUrlRegExp = /\/launchers\/[A-Z]\w+/gi;
+
 export default function reducer(state = initialState, action) {
   const { payload } = action;
 
@@ -48,11 +50,18 @@ export default function reducer(state = initialState, action) {
         location: { pathname },
       } = payload;
 
-      const newEntries = removeEntries({
-        entriesArray: state.entries,
-        startLocation: pathname,
-        inclusive: false,
-      });
+      let newEntries = null;
+
+      // clear header on main urls
+      if (pathname === '/' || launchersUrlRegExp.test(pathname)) {
+        newEntries = [...initialState.entries];
+      } else {
+        newEntries = removeEntries({
+          entriesArray: state.entries,
+          startLocation: pathname,
+          inclusive: false,
+        });
+      }
 
       // console.log('@@router/LOCATION_CHANGE: pathname=%o', pathname);
       // console.log('@@router/LOCATION_CHANGE=>newEntries:', newEntries);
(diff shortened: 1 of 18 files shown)


petrican (approved):
I saw reselect pkg is used. I have nothing against that due to its benefits. But I also saw that we are using lodash. Was wondering if we could not achieve the findings without even using it with just raw JS. You see the goal is to keep the bundle as thin as possible even lodash adds some extra weight. Other than that looks lgtm. Thanks

misc/services/mobile-webui/mobile-webui-frontend/src/reducers/headers.js
```diff
@@ -48,11 +50,18 @@ export default function reducer(state = initialState, action) {
...
 
-      const newEntries = removeEntries({
-        entriesArray: state.entries,
-        startLocation: pathname,
-        inclusive: false,
-      });
+      let newEntries = null;
+
+      // clear header on main urls
+      if (pathname === '/' || launchersUrlRegExp.test(pathname)) {
```
teosarca, line 56: cant we get rid of hardcoded lauchersUrlRegex?

siemiatj:
> I saw reselect pkg is used. I have nothing against that due to its benefits. But I also saw that we are using lodash. Was wondering if we could not achieve the findings without even using it with just raw JS. You see the goal is to keep the bundle as thin as possible even lodash adds some extra weight. Other than that looks lgtm. Thanks

You are aware, that babel-core uses lodash internally ?

misc/services/mobile-webui/mobile-webui-frontend/src/reducers/headers.js
siemiatj: Not really, no. Unless we want to have the header persist:

Screenshot 2021-11-03 at 18 42 32

Maybe it worked before when we had the launchers screen at `/`, but now it's `/launchers/<appId>`, and there's also the app selection screen. I tried changing the initialState's entry to what we use right now but it's not working. So I expect there's more digging needed to get this to work the same way it did before. That's why I think this approach is easier.

petrican:
> > I saw reselect pkg is used. I have nothing against that due to its benefits. But I also saw that we are using lodash. Was wondering if we could not achieve the findings without even using it with just raw JS. You see the goal is to keep the bundle as thin as possible even lodash adds some extra weight. Other than that looks lgtm. Thanks
> 
> You are aware, that babel-core uses lodash internally ?

Yep. Not much we can do about it.

misc/services/mobile-webui/mobile-webui-frontend/src/reducers/headers.js
teosarca: understand

teosarca:
@siemiatj  let's integrate it as it is now and improve later.

Issue #11931, opened 2021-11-04
Address is not saved when focus is on other tab


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
Address is not saved when focus is on other tab

PR #11932, opened 2021-11-04
Fix saving address when clicking outside

petrican:
Rel to https://github.com/metasfresh/metasfresh/issues/11931

Demo: https://www.loom.com/share/4884fe908eb444c1863a3e3fbc4cf021


siemiatj:
If this doesn't break anything, I'm all down for such solution !

siemiatj (approved):

Issue #11939, opened 2021-11-05
mobile UI frontend: don't call /version if service worker are disabled


petrican:
### Is this a bug or feature request?
B

PR #11940, opened 2021-11-05
Dont call the version check if there is no service worker registered

petrican:
Rel to https://github.com/metasfresh/metasfresh/issues/11939

no calls to version present anymore 
Screenshot 2021-11-05 at 08 53 10

### DIFF
--- a/misc/services/mobile-webui/mobile-webui-frontend/src/components/UpdateCheck.jsx
+++ b/misc/services/mobile-webui/mobile-webui-frontend/src/components/UpdateCheck.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { getServerVersion } from '../api/update';
 import { connect } from 'react-redux';
 import { setVersion } from '../actions/UpdateActions';
-import { REGISTER_SERVICE_WORKER } from '../constants/index';
+import { REGISTER_SERVICE_WORKER, UPDATE_CHECK_INTERVAL } from '../constants/index';
 class UpdateCheck extends Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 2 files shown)


misc/services/mobile-webui/mobile-webui-frontend/src/components/UpdateCheck.jsx
```diff
@@ -49,9 +50,11 @@ class UpdateCheck extends Component {
...
 
   componentDidMount() {
-    const { updateInterval } = this.props;
-    let intervalId = setInterval(this.checkServerVersion, updateInterval);
-    this.setState({ intervalId: intervalId });
+    if (REGISTER_SERVICE_WORKER) {
+      const { updateInterval } = this.props;
+      let intervalId = setInterval(this.checkServerVersion, updateInterval);
+      this.setState({ intervalId: intervalId });
+    }
```
teosarca, line 57: why not here?
image

teosarca (changes requested):
check my comments

misc/services/mobile-webui/mobile-webui-frontend/src/components/UpdateCheck.jsx
petrican: It will still mount the component and it will execute the code from componentDidMount wihin the UpdateCheck component.
So something like 

{!REGISTER_SERVICE_WORKER && <UpdateCheck updateInterval={UPDATE_CHECK_INTERVAL} />} won't work.

Also this check setting is specific to UpdateCheck component and I find the correct place to be is within it

teosarca: > {!REGISTER_SERVICE_WORKER && <UpdateCheck updateInterval={UPDATE_CHECK_INTERVAL} />} won't work.

why it won't work?

> Also this check setting is specific to UpdateCheck component and I find the correct place to be is within it

going with the same thinking, why we have `updateInterval` property. Why the UPDATE_CHECK_INTERVAL is not checked directly inside UpdateCheck component?

not suggesting to do so, but i need to understand, from design point of view, why UPDATE_CHECK_INTERVAL is checked outside of `UpdateCheck` component and `REGISTER_SERVICE_WORKER` is checked inside the component?

Issue #11941, opened 2021-11-05
Prefer isShipToDefault/isBillToDefault addresses when creating new order


adi-stefan:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Currently the pre-populated shipping and billing locations of an order are selected from the `isShipTo`/`isBillTo` locations, even if another location is marked as `isShipToDefault`/`isBillToDefault`.
#### Which are the steps to reproduce?
Have a business partner with multiple locations.
Create a sales order for it. Notice the locations defaulted in when selecting the business partner.
Change the `isShipToDefault`/`isBillToDefault` values for the business partner locations.
Create another sales order for it. Notice the defaulted locations have not changed.

PR #11942, opened 2021-11-05
#11941 prefer locations with isShipToDefault/isBillToDefault when defaulting for an order

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/11941

### DIFF
--- a/backend/de.metas.business/src/main/java-legacy/org/compiere/model/MOrder.java
+++ b/backend/de.metas.business/src/main/java-legacy/org/compiere/model/MOrder.java
@@ -23,9 +23,12 @@ package org.compiere.model;
 
 import com.google.common.collect.ImmutableList;
 import de.metas.acct.api.IFactAcctDAO;
+import de.metas.bpartner.BPartnerLocationId;
 import de.metas.bpartner.exceptions.BPartnerNoBillToAddressException;
 import de.metas.bpartner.exceptions.BPartnerNoShipToAddressException;
 import de.metas.bpartner.service.IBPartnerDAO;
+import de.metas.bpartner.service.IBPartnerDAO.BPartnerLocationQuery;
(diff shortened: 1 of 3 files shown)


backend/de.metas.business/src/main/java-legacy/org/compiere/model/MOrder.java
```diff
@@ -461,36 +464,25 @@ public void setBPartner(final I_C_BPartner bp)
...
-			}
-		}
-		// set to first
-		//No longer setting any location when no shipping location exists for the bpartner
-		if (getBill_Location_ID() == 0 && !locs.isEmpty())
-		{
-			setBill_Location_ID(locs.get(0).getC_BPartner_Location_ID());
-		}
-		if (getC_BPartner_Location_ID() == 0)
+		final BPartnerLocationId shipToLocationId = bPartnerDAO.retrieveBPartnerLocationId(BPartnerLocationQuery.builder()
```
metas-ts, line 465: further up this method i found `// FIXME: keep in sync / merge with de.metas.adempiere.service.impl.OrderBL.setBPartner(I_C_Order, I_C_BPartner)`

that OrderBL is today in a different package..I moved "fixme" to java-doc-level and also fixed the package-name..

but either way, i think that at least you need to
- add the `bPartnerDAO.retrieveBPartnerLocationId` call there
- add the BPartnerLocationAndCaptureId stuff here.

metas-ts (changes requested):

metas-ts (approved):

backend/de.metas.business/src/main/java-legacy/org/compiere/model/MOrder.java
adi-stefan: synced.

Issue #11895, opened 2021-10-27
PartnerLocation dropdown not taking full width


siemiatj:
### Is this a bug or feature request?

Bug

PR #11976, opened 2021-11-12
Use whole width of the lookup field

siemiatj:
Related to #11895 
Screenshot 2021-11-12 at 11 55 03


teosarca:
@siemiatj  how the dropdown for BPartner and Contact looks too? If that's OK then IMHO it's OK.
Are other dropdowns affected?

siemiatj:
Yup, all look good

Screenshot 2021-11-15 at 10 33 56
Screenshot 2021-11-15 at 10 33 35

PR #11987, opened 2021-11-15
Address regressions around "Put InvoiceVoidAndRecreateWorkpackageProcessor into working oder"

av-ps:


backend/de.metas.business/src/main/java/de/metas/allocation/api/impl/AllocationBL.java
```diff
@@ -58,6 +58,13 @@ public I_C_AllocationHdr autoAllocateAvailablePayments(final I_C_Invoice invoice
 			return null;
 		}
 
+		//dev-note: don't allocate payments to reversal documents
```
cp-ps, line 61: @metas-ts note: with this change we made partial payments work so we didn't skip the partial invoices anymore, if you think we should actually keep skipping them let me know and I'll quickly add that

metas-ts (approved):

PR #11992, opened 2021-11-17
Send BPartners metasfresh to GRS

fp-ps:


backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/data/DataExportAudit_StepDef.java
```diff
@@ -311,4 +311,18 @@ public void change_bpartner_location(@NonNull final DataTable dataTable)
 
 		InterfaceWrapperHelper.save(bPartnerLocation);
 	}
+
+	@And("the following c_bpartner is changed")
+	public void change_bpartner(@NonNull final DataTable dataTable)
```
metas-ts, line 316: AFAIU this stepdef
* doesn't require a table, because it about one row and one column.
* doesn't belong into DataExportAudit_StepDef, but rather into C_BPartner_StepDef

LE: feel free to keep it a table if you want, but then please, iterate over all rows, and add a 2nd column for `Name2`, so it can be extended as needed later

backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/grssignum/ExportToGRSService.java
```diff
@@ -0,0 +1,105 @@
...
+		final Map<String, String> parameters = new HashMap<>();
+
+		final String authToken = grsSignumConfig.getAuthToken();
+
+		if (Check.isBlank(authToken))
+		{
+			throw new AdempiereException("Missing authToken for ExternalSystemGRSSignumConfig")
+					.appendParametersToMessage()
+					.setParameter("ExternalSystemGRSSignumConfigId", grsSignumConfig.getId());
+		}
```
metas-ts, line 75:
```suggestion
```

we don't know that the Auth-Token is actually a thing, that's why i spec'ed it as optional

metas-ts (changes requested):

metas-ts (approved):

PR #11999, opened 2021-11-18
use the current attachment's uploadDate as API param for the next call

metas-ts:
...see comment in AttachmentProcessor for details


metas-ts (commented):
lgtm

PR #12040, opened 2021-11-23
Calculate activity issue line's quantity on step change

siemiatj:
https://recordit.co/BtBMu90cAC

Just noticed stupid gif is too short
qty


teosarca (changes requested):
lgtm

PR #12042, opened 2021-11-23
Production dispo

dm-ps:


metas-ts (commented):
intermediate review; i'm roughtly 30% through, but need to move on. will be back later today

backend/de.metas.manufacturing/src/main/java/org/eevolution/api/PPOrderCreateRequest.java
```diff
@@ -99,8 +99,6 @@
 			@Nullable final BPartnerId customerId,
 			@Nullable final ProjectId projectId,
 			//
-			final boolean pickingOrder,
```
metas-ts, line 102: i think we shouldn't remove it (bc of the follow-up-issue)

backend/de.metas.manufacturing/src/main/java/org/eevolution/productioncandidate/model/interceptor/PP_OrderCandidate_PP_Order.java
```diff
@@ -0,0 +1,80 @@
...
+	{
+		final PPOrderCandidateId ppOrderCandidateId = PPOrderCandidateId.ofRepoId(ppOrderAllocation.getPP_Order_Candidate_ID());
+
+		final I_PP_Order_Candidate ppOrderCandidate = ppOrderCandidatesDAO.getById(ppOrderCandidateId);
+
+		final ImmutableList<I_PP_OrderCandidate_PP_Order> orderAllocations = ppOrderCandidatesDAO.getOrderAllocations(ppOrderCandidateId);
+
+		final BigDecimal qtyProcessed = orderAllocations.stream()
+				.peek(allocation -> {
+					if (ppOrderCandidate.getC_UOM_ID() != ppOrderAllocation.getC_UOM_ID())
```
metas-ts, line 62: can be we sure those legitimately never happens?
wdyt about using `Quantitys` add the values in an UOM-safe matter?

backend/de.metas.manufacturing/src/main/java/org/eevolution/productioncandidate/model/interceptor/PP_Order_Candidate.java
```diff
@@ -0,0 +1,167 @@
...
+	@ModelChange(timings = { ModelValidator.TYPE_BEFORE_CHANGE, ModelValidator.TYPE_BEFORE_NEW }, ifColumnsChanged = I_PP_Order_Candidate.COLUMNNAME_QtyToProcess)
+	public void validateQtyToProcess(@NonNull final I_PP_Order_Candidate ppOrderCandidateRecord)
+	{
+		final BigDecimal qtyEntered = ppOrderCandidateRecord.getQtyEntered();
+		final BigDecimal qtyProcessed = ppOrderCandidateRecord.getQtyProcessed();
+		final BigDecimal actualQtyLeftToBeProcessed = qtyEntered.subtract(qtyProcessed);
+
+		if (ppOrderCandidateRecord.getQtyToProcess().compareTo(actualQtyLeftToBeProcessed) > 0)
+		{
+			throw new AdempiereException("QtyToProcess cannot be grater then the actual qty left to be processed!")
```
metas-ts, line 122:
```suggestion
			throw new AdempiereException("QtyToProcess cannot be greater then the actual qty left to be processed!")
```

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/order/inoutcandidate/OrderLineShipmentScheduleHandler.java
```diff
@@ -392,8 +392,6 @@ private PPOrderId createPickingOrderIfNeeded(@NonNull final I_C_OrderLine salesO
 																	   .salesOrderLineId(OrderLineId.ofRepoId(salesOrderLine.getC_OrderLine_ID()))
 																	   .customerId(BPartnerId.ofRepoId(salesOrderLine.getC_BPartner_ID()))
 																	   //
-																	   .pickingOrder(true)
```
metas-ts, line 395: as before, i suggest to leave this for the follow-up issue

metas-ts (changes requested):

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/createProductionOrder.feature
```diff
@@ -0,0 +1,423 @@
...
+    And after not more than 10s, PP_Order_BomLines are found
+      | PP_Order_ID.Identifier | Identifier | M_Product_ID.Identifier | QtyEntered | IsQtyPercentage | C_UOM_ID.X12DE355 | ComponentType |
+      | ppo_1                  | ppol_1     | p_2                     | 100        | false           | PCE               | CO            |
+    And after not more than 10s, PP_OrderCandidate_PP_Order are found
+      | PP_Order_Candidate_ID.Identifier | PP_Order_ID.Identifier | QtyEntered | C_UOM_ID.X12DE355 |
+      | ocP_1                            | ppo_1                  | 10         | PCE               |
+    And after not more than 10s, MD_Candidates are found
+      | Identifier | Type   | BusinessCase | M_Product_ID.Identifier | Time                 | DisplayQty | ATP |
+      | c_3        | SUPPLY | PRODUCTION   | p_1                     | 2021-04-16T21:00:00Z | 10         | 0   |
+    And the following MD_Candidates are validated
```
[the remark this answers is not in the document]
dm-ps: `"after not more than 30s, MD_Candidates are found"` searches for the latest MD_Candidate of a particular type and its related STOCK MD_Candidate ( no candidate Id involved )
 
`"the following MD_Candidates are validated"` gets the MD_Candidate of a particular type by id.

We need them both because depending on the step we are aware or not of the particular candidate id.

regarding column names:
  - on `"the following MD_Candidates are validated"` I've used the real db column names for the fields, so I'll update `"after not more than 30s, MD_Candidates are found"` in the same manner

backend/de.metas.manufacturing/src/main/java/org/eevolution/api/PPOrderCreateRequest.java
dm-ps: the field was actually present only on the builder (probably forgotten after certain refactor work ).. so no functional changes here, only clean up...
the `pickingOrder` flag from `PP_Order` ends up there directly from pp_product_planning, see: `org/eevolution/api/impl/CreateOrderCommand.java:158` 

image

backend/de.metas.manufacturing/src/main/java/org/eevolution/productioncandidate/model/interceptor/PP_OrderCandidate_PP_Order.java
dm-ps: currently we can't change the uom on ppOrderCandidate... so it will never happen... 

> wdyt about using Quantitys add the values in an UOM-safe matter?

we could do that, but it would imply that the pp_order_candidate.c_uom could change... which opens other scenarios, so, since we don't currently allow uom changes, I'd say we are better off not implying that here

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/StepDefUtil.java
```diff
@@ -60,4 +61,11 @@ public int extractId(@NonNull final String idOrIdentifier, @NonNull final StepDe
 			return InterfaceWrapperHelper.getId(model);
 		}
 	}
+
+	@And("^wait for (.*)s$")
+	public void waitFor(final int waitingTimeSec) throws InterruptedException
```
metas-ts, line 66: uh-ooh

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/createProductionOrder.feature
```diff
@@ -39,47 +40,49 @@ Feature: create production order
...
-    And after not more than 10s, PP_OrderCandidate_PP_Order are found
+    And after not more than 30s, PP_OrderCandidate_PP_Order are found
       | PP_Order_Candidate_ID.Identifier | PP_Order_ID.Identifier | QtyEntered | C_UOM_ID.X12DE355 |
       | ocP_1                            | ppo_1                  | 10         | PCE               |
-    And after not more than 10s, MD_Candidates are found
-      | Identifier | Type   | BusinessCase | M_Product_ID.Identifier | Time                 | DisplayQty | ATP |
-      | c_3        | SUPPLY | PRODUCTION   | p_1                     | 2021-04-16T21:00:00Z | 10         | 0   |
+    # this is mostly about de.metas.material.dispo.service.candidatechange.StockCandidateService.applyDeltaToMatchingLaterStockCandidates finishing work before checking the candidates
+    And wait for 30s
+    And after not more than 30s, MD_Candidates are found
```
metas-ts, line 76: why hardcode the test to wait for 30s and then wait up to 30s more? i mean, if metasfresh sometimes really takes that long, why not have just `And after not more than 60s, MD_Candidates are found` ?

metas-ts (changes requested):

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/createProductionOrder.feature
metas-ts: Later edit: 
I noticed
`# this is mostly about de.metas.material.dispo.service.candidatechange.StockCandidateService.applyDeltaToMatchingLaterStockCandidates finishing work before checking the candidates`

if the we need a different solution.
The `after not more than 30s, MD_Candidates are found`  should wait for the other results, too. 
And/or maybe cucumber can wait for the handler to be finished (`wait/notify`, `CountDownLatch`)
But we can't delay the build for another 4 additional minutesin one feature file.

PR #12061, opened 2021-11-25
Gh12059 new columns bp quick input

metas-rc:
https://github.com/metasfresh/metasfresh/issues/12059


backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
```diff
@@ -471,19 +471,23 @@ public void updateOrderLineAddressesFromOrder(final org.compiere.model.I_C_Order
...
 	{
 		final BPartnerId shipBPartnerId = BPartnerId.ofRepoIdOrNull(orderRecord.getC_BPartner_ID());
 		final BPartnerId billBPartnerId = BPartnerId.ofRepoIdOrNull(coalesce(
 				orderRecord.getBill_BPartner_ID(),
 				orderRecord.getC_BPartner_ID()));
-
+		if(shipBPartnerId==null || billBPartnerId==null)
+		{
+			return Optional.empty(); // orderOrecord is not yet ready
+		}
```
metas-ts, line 489:
```suggestion
		if(shipBPartnerId == null || billBPartnerId == null)
		{
			return Optional.empty(); // orderRecord is not yet ready
		}
```

metas-ts (approved):

Issue #12066, opened 2021-11-29
add `C_User_Role.IsCustomUserRestriction` column


adi-stefan:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
add `C_User_Role.IsCustomUserRestriction` column, and virtual column `AD_User.IsCustomUserRestriction` depending on associated `C_User_Role.IsCustomUserRestriction`.

PR #12067, opened 2021-11-29
#12066 added C_User_Role.IsCustomUserRestriction and AD_User.IsCustomUserRestriction

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/12066


metas-ts (commented):
i didn't check the AD_SQLColumn_SourceTableColumn table_id, but don't we need two `AD_SQLColumn_SourceTableColumn` records?
Otherwise i don't see how the email-fields' read-only status can change when the flag in C_User_Role is changed :thinking:

adi-stefan:
> i didn't check the AD_SQLColumn_SourceTableColumn table_id, but don't we need two `AD_SQLColumn_SourceTableColumn` records? Otherwise i don't see how the email-fields' read-only status can change when the flag in C_User_Role is changed 🤔

Not sure why it wouldn't work
vxGjTJURRB

metas-ts (approved):

PR #12078, opened 2021-12-01
invoice candidate updating now can also set the invoice's C_Tax_ID

metas-ts:
..so the IC does not need to be deleted and recreated.

- implemented for orderLines, inoutLines and subscriptions
- the class PriceAndTax actually can contain the TaxId now


metas-ts (commented):
lgtm

Issue #11732, opened 2021-09-15
webui frontend: improve how the 502 error is displayed to user 


petrican:
### Is this a bug or feature request?
F
### What is the current behavior?
FE freaks out when 502 is present in the BE API response
#### Which are the steps to reproduce?
Restart instance and access 
### What is the expected or desired behavior?
There should be an overlay similar to offline mode indicating that there is an issue

PR #12082, opened 2021-12-02
11732 bad gateway

petrican:
- fixes the extra stores issue
- deals with the 502 when you are on the login screen and try to authenticate 
- deals with 502 when you are already authenticated and within the app 

related to: https://github.com/metasfresh/metasfresh/issues/11732


frontend/src/components/Container.js
```diff
@@ -295,6 +302,8 @@ Container.propTypes = {
   windowId: PropTypes.string,
   hasComments: PropTypes.bool,
   showSpinner: PropTypes.bool,
+  //functions
+  badGateway: PropTypes.func,
```
teosarca, line 306: cannot see where is this used

frontend/src/components/app/LoginForm.js
```diff
@@ -152,6 +157,9 @@ class LoginForm extends Component {
 
     request
       .then((response) => {
+        response.data.status === 502
+          ? this.props.dispatch(badGateway(BAD_GATEWAY_ERROR))
```
teosarca, line 161: instead of having `badGateway` action why not having `connectionError` action which takes `errorType` as parameter?

frontend/src/constants/Constants.js
```diff
@@ -1,3 +1,20 @@
+/**
+ * @constant
+ * @type {string} Type of `connectionErrorType` to be used in the redux store
+ */
+export const BAD_GATEWAY_ERROR = 'badGateway';
+
+/**
+ * @type {integer} The number of milliseconds to wait before retrying to ping the server
+ */
+export const PING_INTERVAL_BAD_GATEWAY = 5000;
```
teosarca, line 10: this can be renamed to `CONNECTION_ERROR_RETRY_INTERVAL_MILLIS`.

frontend/src/constants/Constants.js
```diff
@@ -1,3 +1,20 @@
...
+
+/**
+ * @constant
+ * @type {string} Type of connectionErrorType when no error is present
+ */
+export const BAD_GATEWAY_EMPTY = '';
```
teosarca, line 16: after applying my suggestion above, this constant might not be needed.

teosarca (changes requested):
check my comments

frontend/src/components/Container.js
```diff
@@ -108,7 +110,10 @@ class Container extends PureComponent {
           />
         )}
 
+        {/* generic error */}
         {connectionError && <ErrorScreen />}
+        {/* when we have specific error message we render the error screen with the appropriate type of error */}
+        {connectionErrorType && <ErrorScreen errorType={connectionErrorType} />}
```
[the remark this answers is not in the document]
petrican: one is boolean (connectionError) - when there is simply no connection

the other (connectionErrorType) is string  - specific case of connection error as it is our case (badGateway)

frontend/src/components/notifications/Notification.js
```diff
@@ -120,8 +120,11 @@ class Notification extends Component {
           />
         </div>
         <div className="notification-content">
-          {shortMsg && msg.length > SHOW_READ_MORE_FROM ? shortMsg + ' ' : msg}
-          {msg.length > SHOW_READ_MORE_FROM &&
+          {shortMsg && msg && msg.length > SHOW_READ_MORE_FROM
+            ? shortMsg + ' '
```
siemiatj, line 124: use string literal ?

frontend/src/components/app/ErrorScreen.js
```diff
@@ -1,25 +1,99 @@
...
+      case 'badGateway':
+        return counterpart.translate(`window.error.badGateway.${targetLine}`);
+      default:
+        return counterpart.translate(`window.error.noStatus.${targetLine}`);
+    }
+  };
+
+  checkNavigatorLanguage = () => {
+    const navigatorLanguage = navigator.language;
+    if (['de', 'de-CH', 'de-AT', 'de-LU', 'de-LI'].includes(navigatorLanguage))
```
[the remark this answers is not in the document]
teosarca: we have an example mobile UI:
image

frontend/src/components/Container.js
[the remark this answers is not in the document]
teosarca: (read below)

frontend/src/utils/websockets.js
```diff
@@ -43,25 +44,32 @@ export function connectWS(topic, onMessageCallback) {
...
+    let badGatewayStatus = false;
     this.sockClient = new Client({
       brokerURL: config.WS_URL,
       debug(strMessage) {
         // console.log('debug: ', strMessage);
         // -- detect reconnect and increment the reconnect counter
         if (strMessage.includes('reconnect')) {
           getUserSession()
-            .then(({ data }) => {
+            .then((userSessionResp) => {
```
teosarca, line 55: why not simply `.then({ data, status }) => { .....` ?
(instead of introducing that `userSessionResp`)

frontend/src/utils/websockets.js
```diff
@@ -43,25 +44,32 @@ export function connectWS(topic, onMessageCallback) {
...
             });
         }
+        // update the store flag
+        badGatewayStatus && store.dispatch(badGateway(BAD_GATEWAY_ERROR));
+
         // -- if more than max allowed reconnect times  ->  deactivate
         if (reconnectCounter > maxReconnectTimesNo) {
           this.reconnectDelay = 0; // 0 - deactivates the sockClient
-          store.dispatch(noConnection(true));
+          !badGatewayStatus && store.dispatch(noConnection(true));
```
teosarca, line 72: as wrote above... IMHO we shall unify different connection errors into one single event.
The actual reason why there is a connection error (e.g. bad gateway or internet down) can be a type.

frontend/src/utils/offlineMessages.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * These are used as fallback when the app is in offline mode. (no messages api fetched from the BE)
+ */
+module.exports = {
+  offline: {
+    de_DE: {
```
siemiatj, line 6: Do we really need to be this specific with those languages ? can't just have `de` & `en` ? This would simplify the whole thing.

frontend/src/reducers/windowHandler.js
```diff
@@ -109,6 +110,7 @@ const initialModalState = {
  */
 export const initialState = {
   connectionError: false,
```
siemiatj, line 112: As teo mentioned above  - can't we just have `connectionError` and have type as value ?

frontend/src/components/Container.js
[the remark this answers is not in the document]
siemiatj: But I guess we could just have `connectionErrorType` and feed there some default value for no connection and something specific when we get the error message, right ?

siemiatj (changes requested):

frontend/src/components/app/ErrorScreen.js
```diff
@@ -1,25 +1,99 @@
...
+
   render() {
-    let line1 = counterpart.translate('window.error.noStatus.title');
-    let line2 = counterpart.translate('window.error.noStatus.description');
+    let line1 = this.getLineFromCounterpart('title');
+    let line2 = this.getLineFromCounterpart('description');
 
-    if (!line1) {
-      line1 = OFFLINE_MESSAGE_LINE1;
+    if (line1.includes('{lang.')) {
```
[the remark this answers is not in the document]
petrican: counterpart returns that string when does not find the key. This is what I used as a fallback indicator. I'' refactor this part once I put the logic in place to use the 'errorType' ->  `counterpart.translate(window.error.${errorType}.title);`

frontend/src/components/app/ErrorScreen.js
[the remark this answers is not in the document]
petrican: Screenshot 2021-12-08 at 10 13 01

I believe we can just use what we already have instead of adding getBrowserLanguage.
Screenshot 2021-12-08 at 10 13 01

frontend/src/constants/Constants.js
```diff
@@ -8,7 +8,7 @@ export const BAD_GATEWAY_ERROR = 'badGateway';
  * @constant
  * @type {string} Type of connectionError when no error is present (internet down)
  */
-export const NO_CONNECTION_ERROR = 'noConnection';
+export const NO_CONNECTION_ERROR = 'noStatus';
```
teosarca, line 11: i find `noStatus` misleading. Why not calling it `offline` or something?

frontend/src/components/app/ErrorScreen.js
```diff
@@ -3,10 +3,15 @@ import React, { Component } from 'react';
...
 import { CONNECTION_ERROR_RETRY_INTERVAL_MILLIS } from '../../constants/Constants';
+import {
+  getCurrentActiveLocale,
+  extractLanguageFromLocale,
+} from '../../utils/locale';
 class ErrorScreen extends Component {
   constructor(props) {
     super(props);
     this.intervalId = null;
+    this.activeLang = extractLanguageFromLocale(getCurrentActiveLocale());
```
teosarca, line 14: instead of `extractLanguageFromLocale(getCurrentActiveLocale())` 
i would suggest to introduce getCurrentActiveLanguage() function.

frontend/src/utils/locale.js
```diff
@@ -20,6 +20,20 @@ export function getCurrentActiveLocale() {
   return localStorage.getItem(LOCAL_LANG);
 }
 
+/**
+ * @summary extracts the language from the locale
+ * @param {string} locale in form of 'en_US', 'de_DE
+ * @returns {string}
```
teosarca, line 26: best practices for writting javadoc / jsdoc:

> @summary extracts the language from the locale

DRY, basically that's exactly what the function name says too.
And that's great, because a good name which is intention revealing does not need to write a function summary.

> @returns {string}

Basically here you need to write how it returns, e.g.
```
@returns {string} extracted language (e.g. 'en', 'de' etc)
```

frontend/src/utils/offlineMessages.js
```diff
@@ -3,19 +3,27 @@
  */
 module.exports = {
```
teosarca, line 4: why this is called `module.exports`?

frontend/src/utils/offlineMessages.js
```diff
@@ -3,19 +3,27 @@
...
   offline: {
-    de_DE: {
-      OFFLINE_MESSAGE_LINE1: 'Verbindung unterbrochen',
-      OFFLINE_MESSAGE_LINE2:
-        'Es treten Verbindungsprobleme auf. Bitte kontrolliere die Verbindung und lade die Seite neu.',
-      BAD_GATEWAY_LINE1: 'Instanz ist nicht verfügbar',
-      BAD_GATEWAY_LINE2: 'Es treten Verbindungsprobleme auf.',
+    de: {
```
teosarca, line 6: consider moving this part in some offline_de.json file.

frontend/src/utils/offlineMessages.js
```diff
@@ -3,19 +3,27 @@
...
+          'Es treten Verbindungsprobleme auf. Bitte kontrolliere die Verbindung und lade die Seite neu.',
+      },
     },
-    en_US: {
-      OFFLINE_MESSAGE_LINE1: 'Connection lost.',
-      OFFLINE_MESSAGE_LINE2:
-        'There are some connection issues. Check connection and try to refresh the page',
-      BAD_GATEWAY_LINE1: 'Instance is not available',
-      BAD_GATEWAY_LINE2: 'There are some connection issues.',
+    en: {
```
teosarca, line 17: consider moving this part in some offline_en.json file.

teosarca: here is an example: https://github.com/metasfresh/metasfresh/blob/tasty_ham_uat/misc/services/mobile-webui/mobile-webui-frontend/src/apps/inventoryDisposal/index.js#L4

frontend/src/constants/Constants.js
teosarca: also, wondering, wouldn't be better to move such constants to some OfflineActionTypes.js file?

petrican: Screenshot 2021-12-08 at 11 54 51
because this is how that key comes from the BE  and to be able to use it according to the errorType 
Screenshot 2021-12-08 at 11 59 24

frontend/src/utils/offlineMessages.js
petrican: Screenshot 2021-12-08 at 12 26 58
things are a bit more complex as there are some configs that do not exist in webUI setup. I would leave as it is in one single file and after that import thing is fixed we can do this. Agree ?

frontend/src/utils/offlineMessages.js
petrican: pls see https://github.com/metasfresh/metasfresh/pull/12082#discussion_r764737620

frontend/src/utils/offlineMessages.js
petrican: https://github.com/metasfresh/metasfresh/pull/12082#discussion_r764737620

frontend/src/actions/WindowActions.js
```diff
@@ -312,9 +313,16 @@ export function updateDataFieldProperty(property, item, scope) {
...
-export function noConnection(status) {
+export function noConnection({ errorType }) {
   return {
-    type: NO_CONNECTION,
+    type: CONNECTION_ERROR,
+    errorType,
+  };
+}
+
+export function badGateway(status) {
```
teosarca, line 323: imho you shall unify those actions into ONE action: connectionError which takes a errorType param which can be:
* OFFLINE
* BAD_GATEWAY
* maybe more to be added in future

frontend/src/constants/Constants.js
teosarca: add a comment about this.... because:
* it's not obvious
* would prevent next developer to break it

frontend/src/reducers/windowHandler.js
```diff
@@ -314,10 +314,10 @@ export const getMasterDocStatus = createSelector(getData, (data) => {
 
 export default function windowHandler(state = initialState, action) {
   switch (action.type) {
-    case NO_CONNECTION:
+    case CONNECTION_ERROR:
       return {
         ...state,
-        connectionError: action.status,
+        connectionError: action.errorType,
```
teosarca, line 320: rename `connectionError` to `errorType`.

frontend/src/utils/websockets.js
```diff
@@ -43,25 +44,33 @@ export function connectWS(topic, onMessageCallback) {
...
+        // update the store flag
+        badGatewayStatus &&
+          store.dispatch(noConnection({ errorType: BAD_GATEWAY_ERROR }));
+
         // -- if more than max allowed reconnect times  ->  deactivate
         if (reconnectCounter > maxReconnectTimesNo) {
           this.reconnectDelay = 0; // 0 - deactivates the sockClient
-          store.dispatch(noConnection(true));
+          !badGatewayStatus &&
+            store.dispatch(noConnection({ errorType: NO_CONNECTION_ERROR }));
```
teosarca, line 73: why do we have to treat NO_CONNECTION_ERROR and BAD_GATEWAY_ERROR differently?

teosarca (changes requested):
check my comments (2)

frontend/src/actions/WindowActions.js
petrican: we have the `noConnection({ errorType }` I forgot to delete this one

frontend/src/utils/websockets.js
petrican: I thought we want separate messages for each case

teosarca: i mean, errors might be different but we need to have the same approach like retry etc etc

frontend/src/reducers/windowHandler.js
petrican: done

frontend/src/utils/websockets.js
petrican: ok. got it

frontend/src/components/app/ErrorScreen.js
```diff
@@ -1,26 +1,62 @@
...
     }
+  }
 
-    if (!line2) {
-      line2 = OFFLINE_MESSAGE_LINE2;
+  componentWillUnmount() {
+    this.intervalId && clearInterval(this.intervalId);
+  }
+
+  getLineFromCounterpart = (targetLine) => {
```
teosarca, line 38: i think this is no longer needed

frontend/src/utils/websockets.js
```diff
@@ -43,25 +44,35 @@ export function connectWS(topic, onMessageCallback) {
 
   const connect = () => {
     let reconnectCounter = 0;
+    let badGatewayStatus = false;
```
teosarca, line 47: why not having `errorType` (string) instead of the boolean?

frontend/src/actions/WindowActions.js
```diff
@@ -312,10 +312,10 @@ export function updateDataFieldProperty(property, item, scope) {
   };
 }
 
-export function noConnection(status) {
+export function noConnection({ errorType }) {
```
teosarca, line 315: rename the function to `connectionError` to have a name consistent with the action type.

frontend/src/containers/Login.js
```diff
@@ -58,6 +64,7 @@ Login.propTypes = {
   splat: PropTypes.string,
   token: PropTypes.string,
   auth: PropTypes.object,
+  errorType: PropTypes.string,
```
teosarca, line 67: in this context, calling it `connectionErrorType` would be a better name,
 because else it's misleading... the reader could think it's some login error.

frontend/src/reducers/windowHandler.js
```diff
@@ -108,7 +108,7 @@ const initialModalState = {
  * - inlineTab keys ${windowId}_{$tabId}_${rowId}
  */
 export const initialState = {
-  connectionError: false,
+  errorType: '',
```
teosarca, line 111: move this to `appHandler.js` because the connection error is more about the whole application and not about just the window itself.

also i would rename it to `connectionErrorType`.
PS: sorry for asking to rename it earlier, i thought it's stored in a deeper context path instead of in (almost) the root of the context where ofc `errorType` makes no sense at all.

frontend/src/components/app/ErrorScreen.js
```diff
@@ -1,26 +1,62 @@
...
+      case 'badGateway':
+        return counterpart.translate(`window.error.badGateway.${targetLine}`);
+      default:
+        return counterpart.translate(`window.error.noStatus.${targetLine}`);
     }
+  };
+
+  render() {
+    const { errorType } = this.props;
+    let line1 = counterpart.translate(`window.error.${errorType}.title`, {
```
teosarca, line 50: u can call it title instead of line1.

teosarca: and it can be `const`.

frontend/src/components/app/ErrorScreen.js
```diff
@@ -1,26 +1,62 @@
...
+      fallback: counterpart.translate(
+        `offline.${this.activeLang}.${errorType}.title`
+      ),
+    });
+    let line2 = counterpart.translate(`window.error.${errorType}.description`, {
```
teosarca, line 55: same tihng like for `line1`.

teosarca (changes requested):
check my comments (3)

frontend/src/utils/websockets.js
petrican: errorType is a string (the one passed to the noConnection action). 

badGatewayStatus on the other hand is just a flag in here

frontend/src/components/app/ErrorScreen.js
```diff
@@ -1,34 +1,67 @@
...
+  }
+
+  componentWillUnmount() {
+    this.intervalId && clearInterval(this.intervalId);
+  }
+
+  render() {
+    const { errorType } = this.props;
+    const title = counterpart.translate(`window.error.${errorType}.title`, {
+      fallback: counterpart.translate(
```
teosarca, line 41: why this is needed?

i thought that we are plugining in multiple sources to counterpart
and counterpart will do the magic.

At this level i would not expect to bother about from where the message comes.

If we have to do like this for any other case, our whole code will be poluted.

frontend/src/components/app/LoginForm.js
```diff
@@ -152,6 +154,8 @@ class LoginForm extends Component {
 
     request
       .then((response) => {
+        const errorType = response.data.status === 502 ? BAD_GATEWAY_ERROR : '';
```
teosarca, line 157: why we don't have a proper error code for offline error? why empty string?

frontend/src/utils/websockets.js
teosarca: that's obvious for me too.

What i am trying to suggest is, instead of having that flag, which IMHO it's pointless,

have the `connectionErrorType`
which will be initialized with the corresponding connection error type.
i.e. BAD_GATEWAY_ERROR, NO_CONNECTION_ERROR

What would happen if u would have more errors in the future? will u introduce a flag for each of them?

frontend/src/utils/websockets.js
```diff
@@ -43,25 +44,35 @@ export function connectWS(topic, onMessageCallback) {
...
             });
         }
+
         // -- if more than max allowed reconnect times  ->  deactivate
         if (reconnectCounter > maxReconnectTimesNo) {
           this.reconnectDelay = 0; // 0 - deactivates the sockClient
-          store.dispatch(noConnection(true));
+          store.dispatch(
+            connectionError({
+              errorType: badGatewayStatus
```
teosarca, line 71: read above

frontend/src/components/app/LoginForm.js
petrican: in the redux store when the  connectionErrorType is empty string like '' it will mean that there is no error.

teosarca (changes requested):
check my comments (4)

frontend/src/components/app/ErrorScreen.js
petrican: I don't understand. the counterpart is doing the magic. If there is no key window.error.${errorType}.title we are going to use the offline existing key using the counterpart fallback mechanism

Screenshot 2021-12-09 at 12 18 24

frontend/src/utils/websockets.js
petrican: I replied above

frontend/src/utils/websockets.js
teosarca: to introduce and use `let connectionErrorType` instead of that badGatewayStatus flag.

petrican: got it. Thanks for explaining me. Pushed the changes

frontend/src/components/app/ErrorScreen.js
teosarca: image

PR #12091, opened 2021-12-03
Fix for picking HUs on the fly during automated shipping 

metas-ts:
When picking HUs on the fly during automatic shipping, we need to create **TUs** with the shipment-schedule's packing instruction.
In contrast, so far on-the-fly-picking HUs were always created as **CUs**, because we wanted to only decrease the stock and nothing else.

#11890


backend/de-metas-salesorder/src/main/java/de/metas/salesorder/service/AutoProcessingOrderService.java
```diff
@@ -118,6 +118,10 @@ public void completeShipAndInvoice(@NonNull final OrderId orderId)
 		final GenerateShipmentsForSchedulesRequest request = GenerateShipmentsForSchedulesRequest.builder()
 				.scheduleIds(scheduleIds)
 				.quantityTypeToUse(M_ShipmentSchedule_QuantityTypeToUse.TYPE_QTY_TO_DELIVER)
+				
+				// Usually we want only *CUs* to be picked on-the-fly, because we don't know or care which HUs are shipped and don't want to make assumptions regarding their TU-Packaging.
```
metas-ts, line 122:
```suggestion
				// Usually we want only *CUs* to be picked on-the-fly, because we don't know and care which HUs are shipped and don't want to make assumptions regarding their TU-Packaging.
```

metas-ts (commented):
lgtm

PR #12097, opened 2021-12-03
#12094 M_Product.discontinuedFrom

dm-ps:


misc/de-metas-common/de-metas-common-product/src/main/java/de/metas/common/product/v2/response/JsonProduct.java
```diff
@@ -103,15 +105,24 @@
...
 			allowEmptyValue = true, //
 			dataType = "java.lang.String", //
 			value = "This translates to `M_Product.ManufacturerArticleNumber`.")
 	@Nullable
 	@JsonProperty("manufacturerNumber")
 	String manufacturerNumber;
-	
+
+	@ApiModelProperty( //
+			allowEmptyValue = true, //
```
metas-ts, line 118: are we sure it makes sense to allow an empty value here? what does "empty" mean here?
note that empty is different from null..

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/descriptor/sql/ProductLookupDescriptor.java
```diff
@@ -391,9 +395,20 @@ private static void appendFilterByIsActive(final StringBuilder sqlWhereClause, f
...
+		final Timestamp priceDate = TimeUtil.asTimestamp(getEffectivePricingDate(evalCtx));
+
+		sqlWhereClause.append("\n AND ")
+		//@formatter:off
+				.append(" ( ")
+					.append(" p.").append(I_M_Product_Lookup_V.COLUMNNAME_Discontinued).append(" = ").append(sqlWhereClauseParams.placeholder(false))
+				.append(" OR ")
+					.append(" p.").append(I_M_Product_Lookup_V.COLUMNNAME_Discontinued).append(" = ").append(sqlWhereClauseParams.placeholder(true))
+				.append(" AND ")
+					.append(" p.").append(I_M_Product_Lookup_V.COLUMNNAME_DiscontinuedFrom).append(" > ").append(sqlWhereClauseParams.placeholder(priceDate))
```
metas-ts, line 409: hm, i think we might run into trouble if DiscontinuedFrom is `null`, don't we?

metas-ts: Maybe our life is easier if we make DiscontinuedFrom, mandatory after all, if Discontinued = true. In this case we would need
- a migration script to set missing values to e.g. `2010-01-01`
- not just callout but a model interceptor to set the value to `Env.getDate`

metas-ts: PS 2: **And** wdyt about using a coalesce in `M_Product_Lookup_V`, to make sure that DiscontinuedFrom is never null?

cp-ps: Hmm, I know it might be a bit confusing... but actually we consider `Discontinued` == `Y` && `DiscontinuedFrom` == `null` as "always discontinued"... same logic is present in `IProductBL#isDiscontinuedAt`... so with that in mind, I would say the above query works fine with `null` as we are querying for those products that are `NOT` discontinued... 

I'll try to give more context: 
- our query would (more or less) translate to:
  ```SQL
  select *
  from m_product_lookup_v p
  where (p.discontinued = 'N' OR p.discontinued = 'Y' AND p.discontinuedFrom > '2021-12-03'::timestamp);
  ```
- if, `p.discontinuedFrom is 'null'` it will be `null > '2021-12-03'::timestamp` which won't match and the query would see the product as `discontinued` and it will filter it out, which is exactly what we want
- p.s. ofc `p.discontinued = 'Y'` could be omitted but I felt it helps a bit the "understanding process"

misc/de-metas-common/de-metas-common-product/src/main/java/de/metas/common/product/v2/response/JsonProduct.java
cp-ps: right, we need to remove that 👍

dm-ps: removed

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/descriptor/sql/ProductLookupDescriptor.java
dm-ps: re-tested and it works exactly how cp-ps described

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/descriptor/sql/ProductLookupDescriptor.java
```diff
@@ -391,9 +395,20 @@ private static void appendFilterByIsActive(final StringBuilder sqlWhereClause, f
...
 	{
-		sqlWhereClause.append("\n AND p.").append(I_M_Product_Lookup_V.COLUMNNAME_Discontinued).append("=").append(sqlWhereClauseParams.placeholder(false));
+		final Timestamp priceDate = TimeUtil.asTimestamp(getEffectivePricingDate(evalCtx));
+
+		sqlWhereClause.append("\n AND ")
+		//@formatter:off
+				.append(" ( ")
+					.append(" p.").append(I_M_Product_Lookup_V.COLUMNNAME_Discontinued).append(" = ").append(sqlWhereClauseParams.placeholder(false))
+				.append(" OR ")
+					.append(" p.").append(I_M_Product_Lookup_V.COLUMNNAME_Discontinued).append(" = ").append(sqlWhereClauseParams.placeholder(true))
```
metas-ts, line 407:
```suggestion
					.append(" p.").append(I_M_Product_Lookup_V.COLUMNNAME_Discontinued).append(" = ").append(sqlWhereClauseParams.placeholder(true))
					// note: if DiscontinuedFrom='Y' and DiscontinuedFrom is null, then the SQL does not match, which is what we want
```

metas-ts (approved):

Issue #12114, opened 2021-12-08
Add IsPrintWhenPackingMaterial flag in Product


cristinamghita:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?

Add a new flag in product that will be used to display or not packing materials as separate lines in docuemnts

PR #12115, opened 2021-12-08
#12114 Add IsPrintWhenPackingMaterial flag in Product 

cristinamghita:
#12114 Add IsPrintWhenPackingMaterial flag in Product


backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/75-de.metas.fresh.reports/5617251_sys_gh12114_Add_IsPrintWhenPackingMaterial_in_Docs_Sales_InOut_Details_HU.sql
```diff
@@ -0,0 +1,55 @@
...
+	COALESCE(pt.Name, p.name)		AS Name,
+	COALESCE(uomt.UOMSymbol, uom.UOMSymbol)	AS UOMSymbol,
+	iol.Description,
+	p.IsPrintWhenPackingMaterial
+FROM
+	M_InOut io
+	INNER JOIN M_InOutLine iol 			ON io.M_InOut_ID = iol.M_InOut_ID AND iol.isActive = 'Y'
+	-- Product and its translation
+	LEFT OUTER JOIN M_Product p 			ON iol.M_Product_ID = p.M_Product_ID AND p.isActive = 'Y'
+	LEFT OUTER JOIN M_Product_Trl pt 		ON iol.M_Product_ID = pt.M_Product_ID AND pt.AD_Language = $2 AND pt.isActive = 'Y'
```
metas-ts, line 29: please prefix the parameter names with `p_` and use them in the function body, intead of `$1`, `$2` etc

metas-ts (changes requested):
from looking at the code, i don't really see how it works..shouldn't there be some change also in the jrxml. also please check my request regarding DB function params

metas-ts:
PS: please also use the new flag in the standard metasfresh jasper files..if we do display tha packaging materials in there..

metas-ts (approved):

Issue #12111, opened 2021-12-07
Use Partner name and greeting when rendering address only if there is greeting set


cristinamghita:
### Is this a bug or feature request?
FR

PR #12123, opened 2021-12-09
#12111 Use Partner name and greeting when rendering address only if there is greeting set 

cristinamghita:


metas-ts (approved):
the changes are over my head, but since we have a bunch of tests and now you added another one, i guess it's fine

PR #12136, opened 2021-12-10
supplement missing aggregation item for the bill-contact

metas-ts:


metas-ts:
already included in https://github.com/metasfresh/metasfresh/pull/12135

PR #12153, opened 2021-12-11
fix around shipment schedule creation and allocation

metas-ts:
otherwise it might get allocated too much stock


metas-ts (commented):
lgtm

Issue #12160, opened 2021-12-13
Custom filters without parameters are not displayed


siemiatj:
### Is this a bug or feature request?

Bug

PR #12188, opened 2021-12-15
Properly display custom active filters to the user

siemiatj:
Related to #12160

### DIFF
--- a/frontend/src/__tests__/components/filters/Filters.test.js
+++ b/frontend/src/__tests__/components/filters/Filters.test.js
@@ -18,6 +18,7 @@ import filtersActive from '../../../../test_setup/fixtures/filters/filtersActive
 import filtersStoreOne from '../../../../test_setup/fixtures/filters/filtersStoreOne.json';
 import filtersStoreTwo from '../../../../test_setup/fixtures/filters/filtersStoreTwo.json';
 import filtersStoreThree from '../../../../test_setup/fixtures/filters/filtersStoreThree.json';
+import filtersStoreFour from '../../../../test_setup/fixtures/filters/filtersStoreFour.json';
 const mockStore = configureStore([]);
 
 const createStore = function (state = {}) {
@@ -96,6 +97,33 @@ describe('Filters tests', () => {
 
     wrapper.find(
       '.filter-wrapper button[title="Akontozahlung, Completed, Error"]'
+    );
+  });
+
+  it('renders active filters caption for filters without parameters', () => {
+    const updateDocListListener = jest.fn();
+    const dummyProps = createInitialProps(undefined, {
+      updateDocList: updateDocListListener,
+    });
+    const initialState = createStore({
+      windowHandler: {
+        allowShortcut: true,
+        modal: {
+          visible: false,
+        },
+      },
+      filters: filtersStoreFour,
+    });
+
+    const store = mockStore(initialState);
+    const wrapper = shallow(
+      <Provider store={store}>
+        <Filters {...dummyProps} />
+      </Provider>
+    );
+
+    wrapper.find(
+      '.filter-wrapper button[title="Abrechnung_offen_normal"]'
     );
   });
 
@@ -258,7 +286,7 @@ describe('Filters tests', () => {
       expect(updateDocListListener).toBeCalledWith(filterResult);
     });
 
-    it('supports filters without parameters', () => {
+    it('supports selecting filters without parameters', () => {
       const updateDocListListener = jest.fn();
       const dummyProps = createInitialProps(undefined, {
         updateDocList: updateDocListListener,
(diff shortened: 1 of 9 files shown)


frontend/src/components/filters/Filters.js
```diff
@@ -137,6 +140,7 @@ class Filters extends PureComponent {
             : filterItem.value === null;
         });
       }
+      return false;
```
teosarca, line 143: isn't this redundant?

frontend/src/utils/filterHelpers.js
```diff
@@ -164,15 +164,26 @@ export function populateFiltersCaptions(filters) {
...
+          }
+        });
+
         captionsArray = [originalFilter.caption, originalFilter.caption];
       }
 
       if (captionsArray.join('').length) {
-        filtersCaptions[filter.filterId] = captionsArray;
         filtersCaptions[filterId] = captionsArray;
+        filtersCaptions[filterIdx] = captionsArray;
```
teosarca, line 186: why mix ID keys and Index keys in the same map? what's the reason behind it?

frontend/src/__tests__/components/filters/Filters.test.js
```diff
@@ -99,6 +100,33 @@ describe('Filters tests', () => {
...
+    const store = mockStore(initialState);
+    const wrapper = shallow(
+      <Provider store={store}>
+        <Filters {...dummyProps} />
+      </Provider>
+    );
+
+    wrapper.find(
+      '.filter-wrapper button[title="Abrechnung_offen_normal"]'
+    );
```
teosarca, line 127: don't see where is the assumpion/expectation.... or this test is not done yet...

frontend/src/utils/filterHelpers.js
```diff
@@ -16,40 +16,38 @@ import { getFormatForDateField, getFormattedDate } from './widgetHelpers';
...
-
-            filterParameter = {
-              parameterName,
-              value: date,
-              valueTo: dateTo,
-            };
-          }
+    if (filter.parameters && filter.parameters.length) {
+      filter.parameters = filter.parameters.map((filterParameter) => {
+        const { parameterName, value, valueTo } = filterParameter;
```
teosarca, line 27: scout boy rule: i think u can extract the body of `map` into a new function
e.g.
```
function formatFilterParameter(parameter) { ... }
```

frontend/test_setup/fixtures/filters.json
```diff
@@ -69,18 +69,6 @@
                 "displayed": true,
                 "readonly": false,
                 "type": "primary"
-              },
-              {
```
teosarca, line 73: why removing this from the test case?

frontend/test_setup/fixtures/filters.json
```diff
@@ -301,18 +289,6 @@
...
-                "caption": "Error",
```
teosarca, line 306: why removing this from the test case?

frontend/test_setup/fixtures/filters/filtersStoreThree.json
```diff
@@ -39,26 +39,11 @@
...
             "parametersLayoutType": "panel",
-            "debugProperties": {}
+            "debugProperties": { "sortNo": 20002 }
           }
         ]
       }
     ],
-    "filtersActive": [
-      {
-        "filterId": "default",
```
teosarca, line 49: why removing this from the test case?

teosarca (changes requested):
check my comments

frontend/src/utils/filterHelpers.js
siemiatj: I have literally no idea. It was like that before, just used a variable here. There might be some shenanigans when rendering but I haven't written that so did not want to break it.

frontend/test_setup/fixtures/filters.json
siemiatj: Shoot, this was pulled by mistake. Reverting

frontend/test_setup/fixtures/filters/filtersStoreThree.json
siemiatj: It's an active filter that does not exist in the data coming from backend, so it can't really be applied. Frontend allows that as we're treating whatever comes from the server as the source of truth

frontend/src/utils/filterHelpers.js
teosarca: understand

frontend/src/utils/filterHelpers.js
siemiatj: Well sortof. We have this `filterData` that is used in this map function, so in order to pass it there you have to create another anonymous function.

frontend/src/__tests__/components/filters/Filters.test.js
siemiatj: It works the same way as `expect(html).toContain('whatever')` but uses selector instead of string query, so it's supposed to be more efficient. Will fail the same way.

teosarca:
@siemiatj  lgtm, but BEFORE integrating it can u pls provide a test for inline filters and facet filters too?
i.e.
* https://releasetestit.metasfresh.com/window/540375
* https://releasetestit.metasfresh.com/window/540092

teosarca (approved):

siemiatj:
Also fixed the position of the `Apply` button so that it's always visible:

Screenshot 2021-12-21 at 00 53 45

PR #12193, opened 2021-12-16
add auth for externalsystem '/actuator' endpoints

cp-ps:


misc/services/camel/de-metas-camel-externalsystems/core/src/main/resources/application.properties
```diff
@@ -22,7 +25,8 @@ externalSystem.raw-data.uri=file://C://raw_data
 # some port that's not yet taken on my local IDE
 server.port=8095
 
-management.endpoints.web.exposure.include=info,health,mappings,metrics,shutdown
+management.endpoints.web.exposure.include=info,health,mappings,metrics,shutdown,loggers
+management.endpoint.loggers.enabled=true
```
metas-ts, line 30:
```suggestion

# Set de.metas.camel.externalsystems.actuator.auth to specify the token with which to access the actuator endpoint 
# When calling the actuator EP, set the http-parameter X-Auth-Token to this value
# The token can also be specified via env variable DE_METAS_CAMEL_EXTERNALSYSTEMS_ACTUATOR_AUTH

```

metas-ts (approved):

PR #12198, opened 2021-12-17
fix for e2e tests

petrican:
Fallback to en in case no locale passed

Without this fix cypress encounters thrown error and stops testing

Screenshot 2021-12-17 at 16 21 59

this happens because at the mom of the call to get the LOCAL_LANG from localStorage this could not yet be set in the local storage (cypress is super fast) so we need to provide a fallback mechanism.


teosarca (approved):
@petrican  pls integrate it

PR #12199, opened 2021-12-17
Cucumber test for over picked quantity

dm-ps:


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentScheduleUpdater.java
```diff
@@ -782,8 +783,9 @@ void updateProcessedFlag(@NonNull final I_M_ShipmentSchedule sched)
 		}
 		final boolean noQtyOverride = sched.getQtyToDeliver_Override().signum() <= 0;
 		final boolean noQtyReserved = sched.getQtyReserved().signum() <= 0;
+		final boolean noQtyPickedAndNotDelivered = getPendingQtyPicked(sched).signum() <= 0;
 
-		sched.setProcessed(noQtyOverride && noQtyReserved);
+		sched.setProcessed(noQtyOverride && noQtyReserved && noQtyPickedAndNotDelivered);
```
metas-ts, line 788: i think this extra DB quiry&extra-code is not needed. 
why not check using `sched.getQtyDelivered()` and `sched.getQtyPickList()`?

metas-ts (approved):

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentScheduleUpdater.java
cp-ps: yep, ur right, I only saw now the `de.metas.inoutcandidate.api.impl.ShipmentScheduleUpdater#generate_SecondRun` call made right before we try to mark the schedule as processed.. so yeah, we can drop that call to db and rely on `schedule.qtyPicklist`. thx! ✌️

metas-ts (approved):

PR #12201, opened 2021-12-17
Create and Export AD_User.Email2 and EMail3

av-ps:
refs: https://github.com/metasfresh/metasfresh/issues/12200


metas-ts (approved):

metas-ts (changes requested):
sorry, i handn'T thought about creating the actual DDL columns.
unfortunately your approach doesn't work with our postgres-version (btw i recommend using the image from `dev-support`)
See https://jenkins.metasfresh.com/job/metasfresh/job/blonde_monkey_uat_gh12200/1/console

What you can do instead is follow this example (anon plpgsql function): https://github.com/metasfresh/metasfresh/blob/3fe802d7ffd116d87440ec9c9f4149d52924c17b/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5478550_sys_gh705webui_AD_Column_more_filtering_options_DDL.sql

metas-ts (approved):

PR #12213, opened 2021-12-20
Focus first field in the filters panel

siemiatj:
Related to #12206

### DIFF
--- a/frontend/src/components/filters/FiltersItem.js
+++ b/frontend/src/components/filters/FiltersItem.js
@@ -457,6 +457,7 @@ class FiltersItem extends PureComponent {
     } = this.props;
     const { filter, isTooltipShow, maxWidth, maxHeight } = this.state;
     const style = {};
+    let autoFocusedField = false;
 
     if (maxWidth) {
       style.width = maxWidth;
@@ -508,7 +509,17 @@ class FiltersItem extends PureComponent {
                 {filter.parameters &&
                   filter.parameters.map((item, index) => {
                     const { widgetType } = item;
+                    let autoFocus = false;
                     item.field = item.parameterName;
+
+                    // don't focus date/time related fields
+                    if (
+                      !autoFocusedField &&
+                      !Object.keys(DATE_FIELD_FORMATS).includes(widgetType)
+                    ) {
+                      autoFocusedField = true;
+                      autoFocus = true;
+                    }
 
                     return (
                       <WidgetWrapper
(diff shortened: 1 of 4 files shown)


frontend/src/components/filters/FiltersItem.js
```diff
@@ -508,8 +509,18 @@ class FiltersItem extends PureComponent {
                 {filter.parameters &&
                   filter.parameters.map((item, index) => {
                     const { widgetType } = item;
+                    let autoFocus = false;
                     item.field = item.parameterName;
 
+                    // don't focus date/time related fields
+                    if (
+                      !autoFocusedField &&
+                      !Object.keys(DATE_FIELD_FORMATS).includes(widgetType)
```
teosarca, line 518: what's the reason to avoid giving focus to date/time components?

teosarca (changes requested):
pls check my comments

frontend/src/components/filters/FiltersItem.js
siemiatj: Because it's a custom widget, there's no input field we can immediately focus. If I comment this out, field won't be focus until user clicks `[tab]` again

0DOSbjw6Eg

teosarca: @siemiatj ok, but then why not extracting this to some `isFocusableWidgetType(widgetType)` function?
There u could add a comment that dates are not focusable because those components don't have an input field (yet)...

frontend/src/utils/widgetHelpers.js
```diff
@@ -188,3 +189,7 @@ export function shouldPatch({
 export function getWidgetField({ filterWidget = false, fields }) {
   return filterWidget ? fields[0].parameterName : fields[0].field;
 }
+
+export function isFocusableWidgetType(widgetType) {
+  return Object.keys(DATE_FIELD_FORMATS).includes(widgetType);
```
teosarca, line 194: i think it's NOT that...

frontend/src/components/filters/FiltersItem.js
```diff
@@ -515,7 +516,7 @@ class FiltersItem extends PureComponent {
                     // don't focus date/time related fields
```
teosarca, line 516: remove that comment because is no longer valid

frontend/src/components/filters/FiltersItem.js
```diff
@@ -515,7 +516,7 @@ class FiltersItem extends PureComponent {
...
                     if (
                       !autoFocusedField &&
-                      !Object.keys(DATE_FIELD_FORMATS).includes(widgetType)
+                      !isFocusableWidgetType(widgetType)
```
teosarca, line 519: i think it's without NOT

teosarca (changes requested):
check my comments (2)

teosarca (approved):

Issue #12185, opened 2021-12-15
Support for v4.5 datenaustausch  xml schema


adi-stefan:
### Is this a bug or feature request?
Feature
### What is the current behavior?
Currently we only support datenaustausch schema version 4.4 xml files for both import and export.
### What is the expected or desired behavior?
We should also support v4.5, and have the ability to convert to the newer format. so 4.4 -> metasfresh -> 4.5.

PR #12214, opened 2021-12-20
#12185 create correct XmlService values when migrating from 4.4

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/12185


backend/vertical-healthcare_ch/forum_datenaustausch_ch.invoice_450_request/src/main/java/de/metas/vertical/healthcare_ch/forum_datenaustausch_ch/invoice_450/Invoice450FromCrossVersionModelTool.java
```diff
@@ -1163,69 +1197,230 @@ private ServicesType createServicesType(@NonNull final List<XmlService> services
...
+		if (zsrToEanPartyMap.containsKey(zsrOrEanPartyId))
+		{
+			return zsrToEanPartyMap.get(zsrOrEanPartyId);
+		}
+		else if (!StringUtils.isNumber(zsrOrEanPartyId))
+		{
+			Check.fail("Invalid ean: {}", zsrOrEanPartyId);
+		}
+		return zsrOrEanPartyId;
+	}
```
adi-stefan, line 1367: This is honestly a shot in the dark, no idea how other to solve this problem. If we're translating from 4.4 and we get a value as zsr that we didn't previously pick up along with an EanParty (to know where to map it to), we're dead in the water and can't convert from 4.4

metas-ts: Suggestions:
- explain the problem to the customer. Maybe they can ensure on their side to always deliver just EANs.
- here are some screenshots you might use:

image
image

**If no**
- (do C&E first pls)
- Get the windows-help-file https://www.forum-datenaustausch.ch/fileadmin/fda_documents/4.5/Rechnungen/XML/generalInvoiceRequest450.chm_210304.zip and check out what those EANs are about
  - note that in order to read it on a sane windows system, you need to "unlock" it in the properties: 
image
  - spoiler: I took a quick look, but could not see whether or not we can generally expect that the zrs/ean combos show up earlier in the XML.
- Extend `XmlRecordService` to have either a dedicated ZSR and EAN field, or to have just an additional EAN field in addition to the existing "both"-field.
- Extend `XmlService.ServiceModWithSelector` so they can be set within the X-model by metasfresh
- Implement a ZSR => EAN mapping table and window in metasfresh that the user needs to maintain
  - Use the descriptions and terminology from the `chm`-file for the fields' names and descriptions.
- If there is a ZSR, then set it's respective EAN from this table to the x-model.

metas-ts (approved):
generally LGTM to integate and deploy to UAT. That way, the customer can also play with it and check if anything at all needs to be done.

metas-ts (approved):

Issue #11961, opened 2021-11-09
Ability to link sales invoices to purchase invoices


adi-stefan:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
Need ability to create relation between one or multiple sales invoices and one purchase invoices.

PR #12219, opened 2021-12-21
#11961 changed relation type for invoices via ref_invoice_id to only consider invoices with opposite SoTrx

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/11961


adi-stefan:
Daniela found this oddity while testing.
image
To me it doesn't make sense, since `ref_invoice_id` is used mostly with counter docs, which (afaik) are almost always of different `SoTrx` values. So I've changed the relation types to reference opposite `SoTrx` values.
image

metas-ts (approved):

PR #12231, opened 2021-12-22
Gh12230 marketing contact persons

metas-rc:
https://github.com/metasfresh/metasfresh/issues/12230

### DIFF
--- a/backend/de.metas.marketing/base/src/main/java/de/metas/marketing/base/process/MKTG_ContactPerson_ProcessBase.java
+++ b/backend/de.metas.marketing/base/src/main/java/de/metas/marketing/base/process/MKTG_ContactPerson_ProcessBase.java
@@ -1,15 +1,5 @@
 package de.metas.marketing.base.process;
 
-import java.util.stream.Stream;
-
-import de.metas.process.PInstanceId;
-import org.adempiere.ad.dao.IQueryBL;
-import org.adempiere.ad.dao.IQueryFilter;
-import org.compiere.model.IQuery;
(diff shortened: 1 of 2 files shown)


backend/de.metas.marketing/base/src/main/java/de/metas/marketing/base/process/MKTG_ContactPerson_ProcessBase.java
```diff
@@ -93,13 +94,18 @@ public void createContactPersonsForPartner(final MKTG_ContactPerson_ProcessParam
...
 				.filter(params.getSelectionFilter())
 				.andCollectChildren(I_AD_User.COLUMNNAME_C_BPartner_ID, I_AD_User.class)
 				.addNotInSubQueryFilter(I_AD_User.COLUMNNAME_AD_User_ID, I_MKTG_Campaign_ContactPerson.COLUMNNAME_AD_User_ID, linkTableQuery)
 				.create()
+				.createSelection();
+
+		final Stream<User> usersToAdd = queryBL.createQueryBuilder(I_AD_User.class)
+				.setOnlySelection(selectionId)
+				.create()
 				.setOption(IQuery.OPTION_GuaranteedIteratorRequired, false)
```
metas-ts, line 109: actually you would get the same result with 
```suggestion
				.setOption(IQuery.OPTION_GuaranteedIteratorRequired, true)
```

to keep it simpler i think you should rather go with that

metas-ts (changes requested):

PR #12242, opened 2021-12-27
enhance `CheckProcessedAsynBatchWorkpackageProcessor`

cp-ps:
refs: https://github.com/metasfresh/metasfresh/issues/12241

### DIFF
--- a/backend/de.metas.async/src/main/java/de/metas/async/api/IAsyncBatchBL.java
+++ b/backend/de.metas.async/src/main/java/de/metas/async/api/IAsyncBatchBL.java
@@ -87,5 +87,5 @@ public interface IAsyncBatchBL extends ISingletonService
 
 	Optional<Integer> getDelayUntilCheckingProcessedState(I_C_Async_Batch asyncBatch);
 
-	boolean shouldBeProcessedManually(AsyncBatchId asyncBatchId);
+	boolean shouldBeManuallyMarkedAsProcessed(AsyncBatchId asyncBatchId);
 }
--- a/backend/de.metas.async/src/main/java/de/metas/async/api/IAsyncBatchBL.java
+++ b/backend/de.metas.async/src/main/java/de/metas/async/api/IAsyncBatchBL.java
@@ -9,6 +9,7 @@ import de.metas.async.model.I_C_Queue_WorkPackage;
 import de.metas.async.model.I_C_Queue_WorkPackage_Notified;
 import de.metas.util.ISingletonService;
 
+import java.time.Duration;
 import java.util.Optional;
 
 /**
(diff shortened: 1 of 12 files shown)


backend/de.metas.async/src/main/java/de/metas/async/api/impl/AsyncBatchBL.java
```diff
@@ -213,145 +171,92 @@ public void enqueueAsyncBatch(@NonNull final AsyncBatchId asyncBatchId)
...
-		}
-		final int countEnqueued = asyncBatch.getCountEnqueued();
-		return countEnqueued > 0;
-	}
-	
-	private int getProcessedTimeOffsetMillis()
-	{
-		return Services.get(ISysConfigBL.class).getIntValue("de.metas.async.api.impl.AsyncBatchBL_ProcessedOffsetMillis", 1);
+		// If we reach this point, we can move on and check if the async batch is processed
+		return Optional.empty();
```
metas-ts, line 259: so i guess, instead of having an extra Optional, we could also return `Duration.ZERO`. i think that would make things clearer

backend/de.metas.async/src/main/java/de/metas/async/api/IAsyncBatchBL.java
```diff
@@ -84,4 +84,8 @@
 	void updateProcessedFromMilestones(AsyncBatchId asyncBatchId);
 
 	AsyncBatchId newAsyncBatch(String asyncBatchType);
+
+	Optional<Integer> getDelayUntilCheckingProcessedState(I_C_Async_Batch asyncBatch);
+
+	boolean shouldBeProcessedManually(AsyncBatchId asyncBatchId);
```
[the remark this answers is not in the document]
metas-ts: PS: in recent code-versions we don't have async-batch-milestones, so maybe it doesn't make sense to put effort in this

backend/de.metas.async/src/main/java/de/metas/async/processor/impl/WorkpackageProcessorTask.java
```diff
@@ -555,7 +555,7 @@ private void markSkipped(
 		final String msg = StringUtils.formatMessage("Skipped while processing workpackage by processor {}; workpackage={}", processorName, workPackage);
 
 		// log error (for later audit), but note that this exception can actually be quite frequent. so in order to not clutter the log, just log it on debug level
-		logger.debug(msg, skipException);
+		logger.error(msg, skipException);
```
metas-ts, line 558: check my comment please :-)
if the problem is now so infrequest that we don't need the "debug" anymore, please adapt the comment accordingly

metas-ts (changes requested):

backend/de.metas.async/src/main/java/de/metas/async/processor/impl/WorkpackageProcessorTask.java
```diff
@@ -554,8 +554,8 @@ private void markSkipped(
 		}
 		final String msg = StringUtils.formatMessage("Skipped while processing workpackage by processor {}; workpackage={}", processorName, workPackage);
 
-		// log error (for later audit), but note that this exception can actually be quite frequent. so in order to not clutter the log, just log it on debug level
-		logger.debug(msg, skipException);
+		// switch to log debug if this exception turns out too frequently (to not clutter the log)
+		logger.error(msg, skipException);
```
metas-ts, line 558:
```suggestion
		logger.info(msg, skipException);
```

changing the level to info since it's not generally an error

metas-ts (approved):

metas-ts:

metas-ts (approved):

Issue #11728, opened 2021-09-14
Allow contract discounts based on time of subscription creation


adi-stefan:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
When selecting Contract terms, it should be possible to select subscription discounts based on the time of the subscription creation.

PR #12243, opened 2021-12-28
#11728 prevent setting discount if one has already been calculated. Reverse labels for IfForeignDiscountsExist values

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/11728

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/pricing/rules/Discount.java
+++ b/backend/de.metas.business/src/main/java/de/metas/pricing/rules/Discount.java
@@ -80,6 +80,12 @@ public class Discount implements IPricingRule
 			return false;
 		}
 
+		if (result.isDiscountCalculated())
+		{
+			log.debug("Discount was already calculated - {}", result);
+			return false;
+		}
+
 		if (pricingCtx.getBPartnerId() == null)
 		{
 			return false;
@@ -180,10 +186,6 @@ public class Discount implements IPricingRule
 		if (pricingConditionsResult == null)
 		{
 			return;
-		}
-		if (!pricingResult.isDiscountCalculated())
-		{
-			pricingResult.setDiscount(pricingConditionsResult.getDiscount());
 		}
 
 		final BigDecimal priceStdOverride = pricingConditionsResult.getPriceStdOverride();
--- a/backend/de.metas.business/src/main/java/de/metas/pricing/rules/Discount.java
+++ b/backend/de.metas.business/src/main/java/de/metas/pricing/rules/Discount.java
@@ -187,6 +187,7 @@ public class Discount implements IPricingRule
 		{
 			return;
 		}
+		pricingResult.setDiscount(pricingConditionsResult.getDiscount());
 
 		final BigDecimal priceStdOverride = pricingConditionsResult.getPriceStdOverride();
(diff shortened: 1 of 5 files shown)


backend/de.metas.business/src/main/java/de/metas/pricing/rules/Discount.java
```diff
@@ -189,8 +181,10 @@ private static void updatePricingResultFromPricingConditionsResult(
 		{
 			return;
 		}
-
-		pricingResult.setDiscount(pricingConditionsResult.getDiscount());
+		if (!pricingResult.isDiscountCalculated())
```
metas-ts, line 184: which shoul (also?) be in the `apply` method

backend/de.metas.contracts/src/main/java/de/metas/contracts/pricing/SubscriptionPricingRule.java
```diff
@@ -119,7 +119,8 @@ private IPricingResult aggregateSubscriptionDiscount(
 	{
 		if (subscriptionDiscountLine != null
 				&& !subscriptionPricingResult.isDisallowDiscount()
-				&& (!subscriptionPricingResult.isDiscountCalculated() || subscriptionDiscountLine.isPrioritiseOwnDiscount()))
+				&& !subscriptionDiscountLine.getDiscount().isZero()
```
metas-ts, line 122: why ignore zero-discount?

metas-ts (changes requested):

backend/de.metas.contracts/src/main/java/de/metas/contracts/pricing/SubscriptionPricingRule.java
adi-stefan: I thought in a scenario where:
- `C_SubscrDiscount_Line.Discount` = 0
- `C_SubscrDiscount_Line.IfForeignDiscountsExist` = `Use_Our_Discount`
- Non-zero discount coming in from a different rule

In this case, should we apply:
1. Always apply the discount in `C_SubscrDiscount_Line`, because the flag says so
2. Because there's not really a discount here, apply the non-zero one instead.

I found merit in both choices, and went with the option 2. Should I go with 1 instead?

metas-ts: i think 1. makes more sense

backend/de.metas.business/src/main/java/de/metas/pricing/rules/Discount.java
```diff
@@ -190,8 +188,6 @@ private static void updatePricingResultFromPricingConditionsResult(
 			return;
 		}
 
-		pricingResult.setDiscount(pricingConditionsResult.getDiscount());
```
metas-ts, line 193: ..i guess this is not what you wanted, right?

metas-ts (changes requested):

backend/de.metas.business/src/main/java/de/metas/pricing/rules/Discount.java
adi-stefan: right, corrected.

metas-ts (approved):

PR #12249, opened 2022-01-03
Don't consider terms as overlapping that have different AD_Org_IDs

metas-ts:


backend/de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
```diff
@@ -1780,6 +1780,14 @@ public boolean hasOverlappingTerms(final I_C_Flatrate_Term newTerm)
 				continue;
 			}
 
+			// Only consider terms with the same org.
+			// C_Flatrate_Term has access-level=Org, so there is no term with Org=*
+			// Also note that when finding a term for an invoice-candidate, that ICs org is used as a matching criterion
```
metas-ts, line 1785:
```suggestion
			// Also note that when finding a term for an invoice-candidate, that IC's org is used as a matching criterion
```

metas-ts (commented):
lgtm

PR #12255, opened 2022-01-04
HUShippingFacade: avoid packaging-IC to end up in separate invoice

metas-ts:
other classes: just minor improvements


metas-ts (commented):
lgtm

PR #12256, opened 2022-01-04
#11728 correct IfForeignDiscountsExist names, not just translations. Use isDiscountEditable to also drive order of discounts being applied

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/11728

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/pricing/rules/Discount.java
+++ b/backend/de.metas.business/src/main/java/de/metas/pricing/rules/Discount.java
@@ -80,9 +80,9 @@ public class Discount implements IPricingRule
 			return false;
 		}
 
-		if (result.isDiscountCalculated())
+		if (!result.isDiscountEditable())
 		{
-			log.debug("Discount was already calculated - {}", result);
+			log.debug("Discount is not editable - {}", result);
 			return false;
 		}
 
--- a/backend/de.metas.business/src/main/java/de/metas/pricing/rules/Discount.java
+++ b/backend/de.metas.business/src/main/java/de/metas/pricing/rules/Discount.java
@@ -80,9 +80,9 @@ public class Discount implements IPricingRule
 			return false;
 		}
 
-		if (!result.isDiscountEditable())
+		if (result.isDontOverrideDiscountAdvice())
 		{
-			log.debug("Discount is not editable - {}", result);
+			log.debug("Discount should not be modified - {}", result);
 			return false;
 		}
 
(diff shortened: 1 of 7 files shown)


backend/de.metas.contracts/src/main/java/de/metas/contracts/pricing/SubscriptionPricingRule.java
```diff
@@ -119,9 +119,13 @@ private IPricingResult aggregateSubscriptionDiscount(
 	{
 		if (subscriptionDiscountLine != null
 				&& !subscriptionPricingResult.isDisallowDiscount()
-				&& (subscriptionDiscountLine.isPrioritiseOwnDiscount() || !subscriptionPricingResult.isDiscountCalculated()) )
+				&& (subscriptionDiscountLine.isPrioritiseOwnDiscount() || !subscriptionPricingResult.isDiscountCalculated()))
 		{
 			subscriptionPricingResult.setDiscount(subscriptionDiscountLine.getDiscount());
+			if (subscriptionDiscountLine.isPrioritiseOwnDiscount())
+			{
+				subscriptionPricingResult.setDiscountEditable(false);
```
metas-ts, line 127: ok, so i guess that's the answer to my question above. 
but as i wrote, i think `IsDiscountEditable` is related to the user, not to following rules.
Therefore i think we need an additional property like `dontOverrideDiscountAdvice` or something

backend/de.metas.contracts/src/main/java/de/metas/contracts/pricing/SubscriptionPricingRule.java
```diff
@@ -228,7 +231,11 @@ private static void copyDiscountIntoResultIfAllowedByPricingContext(
 	{
 		if (!pricingCtx.isDisallowDiscount())
 		{
-			result.setDiscount(subscriptionPricingResult.getDiscount());
+			if (result.isDiscountEditable())
+			{
+				result.setDiscount(subscriptionPricingResult.getDiscount());
+				result.setDiscountEditable(subscriptionPricingResult.isDiscountEditable());
```
metas-ts, line 237: i would say this `result.setDiscountEditable` still makes sense..and as a 3rd line, here we should also copy the `dontOverrideDiscountAdvice` or whatever property you end up with.

metas-ts (changes requested):

backend/de.metas.business/src/main/java/de/metas/pricing/rules/Discount.java
```diff
@@ -80,9 +80,9 @@ public boolean applies(final IPricingContext pricingCtx, final IPricingResult re
 			return false;
 		}
 
-		if (result.isDiscountCalculated())
+		if (!result.isDiscountEditable())
```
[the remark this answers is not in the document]
adi-stefan: Changed to `isDontOverrideDiscountAdvice`

backend/de.metas.contracts/src/main/java/de/metas/contracts/pricing/SubscriptionPricingRule.java
adi-stefan: added `isDontOverrideDiscountAdvice`

backend/de.metas.contracts/src/main/java/de/metas/contracts/pricing/SubscriptionPricingRule.java
adi-stefan: added.

metas-ts (approved):

PR #12259, opened 2022-01-04
C_Flatrate_Term_Extend_And_Notify_User extends multiple terms per trx

metas-ts:


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandDAO.java
```diff
@@ -1047,6 +1047,9 @@ public void selectionId(final PInstanceId selectionId)
 			});
 		}
 
+		// make sure the order of invoice candidate updating is somewhat predictable
```
metas-ts, line 1050:
```suggestion
		// make sure the order of invoice candidate recomputing is somewhat predictable
```

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandRecomputeTagger.java
```diff
@@ -30,6 +30,7 @@
 import de.metas.util.Check;
 import lombok.NonNull;
 import lombok.ToString;
+import org.adempiere.ad.dao.QueryLimit;
```
metas-ts, line 33:
```suggestion
```

metas-ts (commented):
lgtm so far

Issue #12265, opened 2022-01-05
on HU Editor, transforming a CU by moving part of it to another CU results in the existing CU disappearing


adi-stefan:
### Is this a bug or feature request?
Bug
### What is the current behavior?
-HU Editor
- When splitting CUs in HU editor for return, the original HU is not displayed anymore after the transform.

PR #12266, opened 2022-01-05
#12265 prevent error when doing Material withdrawal (move). Show original CUs after splitting

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/12265


metas-ts (approved):
both changes look plausible, buut
- which of the problems reported by @metas-dh are addressed here?
- did you test if the problems daniela documented work now (material withdrawal, customer returns, split off in HU editor). if you say yes i don't need screenshots this time, though

adi-stefan:
>problem remains for customer returns (created from shipment): when splitting of CUs in HU editor for return, the original HU is not displayed anymore after the transform: https://recordit.co/oSpEcnpQzN ❌

is fixed in `HandlingUnitsBL#isDestroyed`

> also not working anymore: Material withdrawal (move) not working anymore mf15#1196 (not sure if this is related with this task, though) ❓

fixed by change in `WEBUI_M_HU_MoveTUsToDirectWarehouse#doIt`

PR #12267, opened 2022-01-05
Improve retrieval performance of workpackage items for WP processors

metas-ts:
See deprecation notice in IQueueDAO for details.
Also, not strictly related: replace int with QueryLimit


metas-ts (commented):
good to integrate after successful build

PR #12271, opened 2022-01-06
attempt to add visibility and fix possible flaky commission tests

cp-ps:


backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
```diff
@@ -152,8 +156,11 @@ public void recompute_invoice_candidates_if_required(@NonNull final DataTable da
 
 			if (recompute)
 			{
-				final InvoiceCandidateIdsSelection onlyInvoiceCandidateIds = InvoiceCandidateIdsSelection.ofIdsSet(
-						ImmutableSet.of(InvoiceCandidateId.ofRepoId(invoiceCandidate.getC_Invoice_Candidate_ID())));
+				final ImmutableSet<InvoiceCandidateId> invoiceCandidateIds = ImmutableSet.of(InvoiceCandidateId.ofRepoId(invoiceCandidate.getC_Invoice_Candidate_ID()));
+
+				invoiceCandDAO.invalidateCandsFor(invoiceCandidateIds);
+
```
pevesoft, line 162: dev-note: normally, we don't need this for our cases due to `de/metas/invoicecandidate/async/spi/impl/InvoiceCandWorkpackageProcessor.java:121` i.e.:

```JAVA
invoiceCandDAO.invalidateCands(candidatesOfPackage);
```

but I guess there might be cases in the future where we would use this step def to recompute candidates at a different timing so I thought it makes sense

cp-ps, line 162: dev-note: normally, we don't need this for our cases due to `de/metas/invoicecandidate/async/spi/impl/InvoiceCandWorkpackageProcessor.java:121` i.e.:

```JAVA
invoiceCandDAO.invalidateCands(candidatesOfPackage);
```

but there might be cases in the future where we would use this step def to recompute candidates at a different timing so I think it makes sense to have it

metas-ts: I think i don't get it...this seems to be the only *relevant* change.
if we don't need it now, then there is no change at all to fix the flakyness.
And *if* we need it, it AFAIS means that the ICs are incorrect in metasfresh, unless a cucmber-test or something unexpected happens to invalidate them.
In other words, this is actually the flakyness i'm looking forward to be settled.

In that context i now see the "whole" recompute step much more critical. It seems to indicate that the whole thing might fail and we might have wrong data in the production-ICs, unless something unrelated causes them to be recomputed.

Brainfart 1: maybe the recomputing is going on in parallel. In that case, this step needs to make sure to wait for it to end. That can happen by either waiting for and corresponding WPs to be finished, or by waiting for all C_Invoice_Candidate_Reecpute records to be gone.

metas-ts (changes requested):

metas-ts:
PS: if there is a good reason for the recompute, then i think we need a ` recompute ||`, i.e.:

```java
[....]
			final BigDecimal netAmountInvoiced = DataTableUtil.extractBigDecimalOrNullForColumnName(row, "OPT." + COLUMNNAME_NetAmtInvoiced);
			if (netAmountInvoiced != null)
			{
				recompute = recompute || invoiceCandidate.getNetAmtInvoiced().compareTo(netAmountInvoiced) != 0;
			}

			final BigDecimal netAmtToInvoice = DataTableUtil.extractBigDecimalOrNullForColumnName(row, "OPT." + COLUMNNAME_NetAmtToInvoice);
			if (netAmtToInvoice != null)
			{
				recompute = recompute || invoiceCandidate.getNetAmtToInvoice().compareTo(netAmtToInvoice) != 0;
			}
[....]
```

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
```diff
@@ -162,6 +169,10 @@ public void recompute_invoice_candidates_if_required(@NonNull final DataTable da
 						.setOnlyInvoiceCandidateIds(onlyInvoiceCandidateIds)
 						.update();
 			}
+
+			InterfaceWrapperHelper.refresh(invoiceCandidate);
+
+			logger.info("*** C_Invoice_Candidate after recompute: " + invoiceCandidate);
```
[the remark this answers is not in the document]
cp-ps: > AFAIU I_C_InvoiceCandidate.toString() shows little more than the I_C_InvoiceCandidate_ID (and maybe the trx-name)? So I'm not sure this log message makes sense.

thx, didn't check that, I need the log to show more than that, basically all the relevant `qty` columns

PR #12273, opened 2022-01-07
Activity Fix

metas-ts:
Use respective record's ClientId and OrgId when getting the Activity.
We can't assume that the correct AD_Client_ID and AD_Org_ID are always in the context.
E.g. when processing workpackages they are not.


teosarca (approved):
lgtm

PR #12274, opened 2022-01-07
Remove the divider on top of Apply

petrican:
rel to: https://github.com/metasfresh/metasfresh/issues/12160

Screenshot 2022-01-07 at 10 59 26

was added by Kuba as a separator. Agreed with Daniela that we should have consistent layout and that div should not be added.


teosarca (approved):
@petrican  lgtm

Issue #12278, opened 2022-01-07
Show delivery address on customs invoice if there are several shiopments


cristinamghita:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
Show delivery address on customs invoice if there are several shipments, but all have same delivery address

PR #12279, opened 2022-01-07
#12278 Show delivery address on customs invoice if there are several …

cristinamghita:
…shiopments

#12278 Show delivery address on customs invoice if there are several shiopments


teosarca (approved):
lgtm

PR #12331, opened 2022-01-14
send extended bpartnerName to /orders/sales/candidates

av-ps:
refs: https://github.com/metasfresh/metasfresh/issues/12314


metas-ts (dismissed):

metas-ts (changes requested):
please include an api doc at JsonOLCandCreateRequest#bpartnerName that explains how the name will be fall-backed if omitted in here

metas-ts (approved):

Issue #12349, opened 2022-01-18
webui frontend: When loading a single document, if the network is slow, you will see 404 for a second.


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
Load a single document, a sales order or something which loads slow
Before the document gets rendered on screen, you will see that FE renders "404" in the middle of the screen for a while.
That message shall be rendered for documents that were not found but not for documents that are currently loading.
#### Which are the steps to reproduce?
Load a single document, a sales order or something which loads slow
Before the document gets rendered on screen, you will see that FE renders "404" in the middle of the screen for a while.
That message shall be rendered for documents that were not found but not for documents that are currently loading.
### What is the expected or desired behavior?
Expectation: while loading, instead of 404 message, better render "Loading...".

PR #12350, opened 2022-01-18
webui frontend: render loading indicator while included tab data is loading

petrican:
rel to: https://github.com/metasfresh/metasfresh/issues/12349

Demo

https://www.loom.com/share/29b44c6b70d74a82bb2cc8b4171a5901

Done with Slow 3G network throttling (visible starting from 2:42 min)

### DIFF
--- a/frontend/src/__tests__/actions/TablesActions.test.js
+++ b/frontend/src/__tests__/actions/TablesActions.test.js
@@ -43,7 +43,6 @@ import gridRowFixtures from '../../../test_setup/fixtures/grid/row_data.json';
 
 const middlewares = [thunk];
 const mockStore = configureStore(middlewares);
-const pending = true;
 
 const createState = function(state = {}) {
   return merge(
@@ -483,12 +482,12 @@ describe('TableActions tab', () => {
         ...tab,
       };
 
-      dispatchedActions.push(store.dispatch(updateTabTable({ tableId, tableResponse: dataResponse, pending })));
+      dispatchedActions.push(store.dispatch(updateTabTable({ tableId, tableResponse: dataResponse, pending: true })));
       expectedActions.push({
         type: ACTION_TYPES.UPDATE_TABLE,
         payload: {
           id: tableId,
-          data: createTableData({ ...dataResponse, keyProperty: 'rowId', pending }),
+          data: createTableData({ ...dataResponse, keyProperty: 'rowId', pending: true }),
         }
       });
     });
@@ -546,11 +545,11 @@ describe('TableActions tab', () => {
     const tableData = createTableData({
       result: rowDataResponse,
       keyProperty: 'rowId',
-      pending,
+      pending: true,
     });
     tableData.rows = flattenRows(tableData.rows);
 
-    await store.dispatch(updateTabTable({ tableId, tableResponse: { result: rowDataResponse }, pending }));
+    await store.dispatch(updateTabTable({ tableId, tableResponse: { result: rowDataResponse }, pending: true }));
     expect(store.getActions()).toContainEqual(
       {
         type: ACTION_TYPES.UPDATE_TABLE,
--- a/frontend/src/__tests__/actions/TablesActions.test.js
+++ b/frontend/src/__tests__/actions/TablesActions.test.js
@@ -482,12 +482,13 @@ describe('TableActions tab', () => {
         ...tab,
       };
 
-      dispatchedActions.push(store.dispatch(updateTabTable(tableId, dataResponse)));
+      const pending = true;
+      dispatchedActions.push(store.dispatch(updateTabTable({ tableId, tableResponse: dataResponse, pending })));
       expectedActions.push({
         type: ACTION_TYPES.UPDATE_TABLE,
         payload: {
           id: tableId,
-          data: createTableData({ ...dataResponse, keyProperty: 'rowId' }),
+          data: createTableData({ ...dataResponse, keyProperty: 'rowId', pending }),
         }
       });
     });
@@ -539,16 +540,18 @@ describe('TableActions tab', () => {
       },
     });
 
+    const pending = true;
     const store = mockStore(initialState);
     const tabId = rowDataResponse[0].tabId;
     const tableId = getTableId({ windowId: windowType, docId, tabId });
     const tableData = createTableData({
       result: rowDataResponse,
       keyProperty: 'rowId',
+      pending,
     });
     tableData.rows = flattenRows(tableData.rows);
 
-    await store.dispatch(updateTabTable(tableId, { result: rowDataResponse }));
+    await store.dispatch(updateTabTable({ tableId, tableResponse: { result: rowDataResponse }, pending }));
     expect(store.getActions()).toContainEqual(
       {
         type: ACTION_TYPES.UPDATE_TABLE,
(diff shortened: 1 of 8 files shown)


frontend/src/actions/WindowActions.js
```diff
@@ -479,17 +479,21 @@ export function initWindow(windowType, docId, tabId, rowId = null, isAdvanced) {
...
-              data: {},
-              docId: 'notfound',
-              includedTabsInfo: {},
-              scope: 'master',
-              saveStatus: { saved: true },
-              standardActions: [],
-              validStatus: {},
-            })
-          );
+          if (e.response.status === 404) {
```
teosarca, line 482: @petrican in the issue requirements, i was talking about when the document is loading first time.
Before getting the actual data (or layout?) i think somewhere we have hardcoded the default values to be something like not found....

petrican: That is the only place where this is set for the document based on the description you gave. I've tried to replicate on slow instances like BM or SP but without success. I can dig deeper into this.

teosarca: @petrican  now i cannot reproduce it either. OK, let's keep an eye on it.

teosarca: @petrican  found it, it's in included tabs

image


we need there a big rotating gear.

about the change you did here, i would revert it.

petrican: the change is not merged. Yes I can revert it in my branch. But can you give me an instance where it happens ? i see it is on your localhost. Thanks

teosarca: Anywhere @petrican, you just have to throttle your network connection to make it more obvious...

petrican: All right. Thanks! Will check

frontend/src/components/table/Table.js
```diff
@@ -452,7 +452,13 @@ export default class Table extends PureComponent {
   renderEmptyInfo = (rows) => {
     const { emptyText, emptyHint, pending } = this.props;
     if (pending) {
-      return <div>LOADING...</div>;
+      return (
+        <div className="spinner-wrapper-in-tab">
+          <div>
```
teosarca, line 457: why so many divs?

petrican: because it needs to be wrapped and positioned correctly within the table

teosarca: Q: couldn't this be done inside the SpinnerOverlay? I see that there we have already handled the case of "modal" and applied the right CSS class. 
Why this case is different?

petrican: Not quite as we will end up with something like this:
Screenshot 2022-01-20 at 11 58 08
This is different because we need a wrapper that will do the positioning of the spinner ( Imagine I need to put the spinner on the right of a div, in center or whatever, within a tooltip . The spinner needs to keep its size only his position relative to the surrounding items will change).


We do not need to embed logic specific of the Tab within the SpinnerOverlay.
Screenshot 2022-01-20 at 12 04 36

teosarca: ok, understand

frontend/src/__tests__/actions/TablesActions.test.js
```diff
@@ -482,12 +482,13 @@ describe('TableActions tab', () => {
         ...tab,
       };
 
-      dispatchedActions.push(store.dispatch(updateTabTable(tableId, dataResponse)));
+      const pending = true;
```
teosarca, line 485: in tests, pls inline the `pending` const

frontend/src/__tests__/actions/TablesActions.test.js
```diff
@@ -539,16 +540,18 @@ describe('TableActions tab', () => {
       },
     });
 
+    const pending = true;
```
teosarca, line 543: in tests, pls inline the `pending` const

frontend/src/actions/WindowActions.js
```diff
@@ -398,12 +398,15 @@ export function indicatorState(state) {
  */
 export function fetchTab({ tabId, windowId, docId, orderBy }) {
   return (dispatch) => {
+    const tableId = getTableId({ windowId, docId, tabId });
```
teosarca, line 401: use the params in their logical order: windowId, tabId, docId

frontend/src/actions/WindowActions.js
```diff
@@ -398,12 +398,15 @@ export function indicatorState(state) {
...
+    dispatch(updateTabTable({ tableId, tableResponse: {}, pending: true }));
```
teosarca, line 402: as described in the issue, don't clear the rows before getting new ones

petrican: There's a messy logic in there with the creation of the table. I used this workaround to set the pending before. Seems it is not ok. Need to think for a solution. Thx

teosarca: @petrican  still not OK, and u know it. u don;t have to ask me the obvious :(
image

teosarca: @petrican  what i would suggest to do here is to call the function without tableRespose param.
i.e.
```
updateTabTable({ tableId, pending: true })
```

and then change the updateTabTable or where is needed and if the `tableRespose` is missing, simply don't update/change the rows.

makes sense?

teosarca (changes requested):
change my comments

frontend/src/actions/WindowActions.js
petrican: Ok it make sense. Got it. But just to dbl check the final result should be as in this snapshot correct ?

Screenshot 2022-01-21 at 10 32 29

teosarca: @petrican 
image


makes sense now?
any suggestions?

petrican: ok. ty.

petrican: Pushed the changes. This is how it works now https://www.loom.com/share/6007d237bea84505974c822fe706a02f

petrican: Screenshot 2022-01-24 at 09 07 35

petrican:
@teosarca I've done changes. Pls Re-check. Ty

teosarca:
@petrican  checked your code and it becomes more and more complex and polluted because of that `pending` which is stored in redux.

To me, it becomes clear that `pending` flag shall be stored in the local state of that component. 
Do you see any reason why we need to pollute the redux state with this transitory information?

petrican:
> @petrican checked your code and it becomes more and more complex and polluted because of that `pending` which is stored in redux.
> 
> To me, it becomes clear that `pending` flag shall be stored in the local state of that component. Do you see any reason why we need to pollute the redux state with this transitory information?

The goal of a redux app is to have a single source of truth, and that is what we should follow within the app. How the data flows through the redux store is easy to follow and grasp. We need to have the data updated by the correct action and the data flowing from the top on this complex component. Most obvious reason is that having the pending somewhere at component level will make the logic even more complex. ( btw at what component level are you planing to store the `pending` in the state, and which will be the function that will set that ?). Also having the data in the local state makes that data not quite testable for the case of unit tests. If props are passed from the top (as in the case of Redux) that allows easy testing.

teosarca:
@petrican I beg to differ (see for example https://redux.js.org/faq/organizing-state#organizing-state),

but leave it like it is for now.
If it works, let's move forward.

Issue #12351, opened 2022-01-18
webui frontend: Columns shifting on Components of the BOM & Formula


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
Columns shifting

PR #12353, opened 2022-01-18
Add truly unique key for the cells

petrican:
rel to: https://github.com/metasfresh/metasfresh/issues/12351

https://www.loom.com/share/36b89aef4d7b41a2bc45f03c614bedd5

### DIFF
--- a/frontend/src/components/table/TableRow.js
+++ b/frontend/src/components/table/TableRow.js
@@ -15,7 +15,7 @@ import {
   nestedSelect,
   getTooltipWidget,
 } from '../../utils/tableHelpers';
-
+import { v4 as uuidv4 } from 'uuid';
 import TableCell from './TableCell';
 import WithMobileDoubleTap from '../WithMobileDoubleTap';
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/table/TableRow.js
```diff
@@ -560,7 +560,7 @@ class TableRow extends PureComponent {
                 colIndex={idx}
                 hasComments={!!(hasComments && idx === 0)}
                 cellExtended={cellsExtended}
-                key={`${rowId}-${property}`}
+                key={`${rowId}-${property}-${uuidv4()}`}
```
teosarca, line 563: @petrican  Q: why not using the `${idx}` instead of `${uuidv4()}`?

petrican: @teosarca using indexes from map is an anti-pattern. 

Screenshot 2022-01-19 at 09 02 52

https://medium.com/wesionary-team/using-index-as-a-key-is-an-anti-pattern-in-react-8e5db3aea3a9

teosarca: @petrican  I wasn't suggesting to use just the `idx`. I was suggesting to use `${rowId}-${property}-${idx}`.

Ahh, and now that i am looking deeper, i am even not sure why the `${rowId}` needs to be part of the key because as far as I understand the `key` needs to be unique inside the list which contains it.

In our case, the list consists of the cells which are part of a given TableRow.
So, ofc, all cells from that row will share the same `rowId`.
So it's pointless to have it there as a discriminator because it's the same.

To conclude, I would go with `${property}-${idx}`.

petrican: Yeah. That works too. Can the PR be merged ?

PR #12354, opened 2022-01-18
always set invoice-doctype when creating an IC for olc, ol and iol

metas-ts:
because otherwise, ICs that have no doctype but will get the standard-invoice-doctype are not added to the same invoices as similar ICs that already explicitly have this doctype


metas-ts (commented):
lgtm

Issue #12360, opened 2022-01-19
Allow setting shipping costs by country area, not just by country.


adi-stefan:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
Allow setting shipping costs by country area, not just by country.

PR #12370, opened 2022-01-20
#12360 implemented freight cost by country area, not just by country

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/12360

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/freighcost/FreightCostRepository.java
+++ b/backend/de.metas.business/src/main/java/de/metas/freighcost/FreightCostRepository.java
@@ -1,20 +1,5 @@
 package de.metas.freighcost;
 
-import java.util.Collection;
-import java.util.Comparator;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
-
-import org.adempiere.ad.dao.IQueryBL;
-import org.adempiere.exceptions.AdempiereException;
-import org.adempiere.model.I_M_FreightCost;
-import org.adempiere.model.I_M_FreightCostDetail;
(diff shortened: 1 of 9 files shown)


backend/de.metas.business/src/main/java/de/metas/freighcost/FreightCostRepository.java
```diff
@@ -209,11 +208,13 @@ private static FreightCostShipper toFreightCostShipper(
 	private FreightCostBreak toFreightCostBreak(final I_M_FreightCostDetail record)
 	{
 		final CountryId countryId = CountryId.ofRepoIdOrNull(record.getC_Country_ID());
+		final CountryAreaId countryAreaId = CountryAreaId.ofRepoIdOrNull(record.getC_CountryArea_ID());
```
metas-ts, line 211: AFAIU users could currently still have multiple breaks for the same shipmentValue and shipper and idk what kinds of problems that will cause down the road.
therefore i suggest to add a SeqNo (default value expression with a "max(...)+10"-SQL ) with a unique constraint and take the matching break with the smallest shipmentValue **and seqNo**

metas-ts (changes requested):

backend/de.metas.business/src/main/java/de/metas/freighcost/FreightCostRepository.java
adi-stefan: added `M_FreightCostShipper.SeqNo` and changed sort order to use it (records with a null SeqNo will have least priority).

metas-ts (approved):

Issue #12363, opened 2022-01-20
webui frontend: When searching in menu don't show There are no results while loading


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
When searching in menu don't show There are no results while loading
#### Which are the steps to reproduce?
When searching in menu don't show There are no results while loading

PR #12380, opened 2022-01-21
12363 hide no results while loading

petrican:
rel to : https://github.com/metasfresh/metasfresh/issues/12363


frontend/src/assets/css/overlayfield.css
```diff
@@ -99,4 +99,16 @@
         display: flex;
         padding: 20px;
     }
-}
\ No newline at end of file
+}
+
+.menu-overlay-spinner {
```
teosarca, line 104: why not have this in a dedicated menu.css or in already existing header.css?

frontend/src/components/header/MenuOverlay.js
```diff
@@ -96,18 +97,21 @@ class MenuOverlay extends Component {
     if (targetValue) {
       this.setState({
         query: e.target.value,
+        pendingQuery: e.target.value ? true : false,
       });
 
       queryPathsRequest(e.target.value, 9)
         .then((response) => {
           this.setState({
```
teosarca, line 105: (not in requirements but as a scout boy rule)

While testing using Slow 3G i observed that 
* if i search for something => request(1) is sent to backend
* i don't wait to finish the request but i clear the search and write something else  => request(2) is sent to backend

In some cases if the the request(1) is finishing last, so my search is displaying the results of request(1) even though i would expect the results of request(2).

Suggestion to fix:
before updating the state, check if the query from the state is the one from our request, to make sure we get the right one.

frontend/src/components/header/MenuOverlay.js
```diff
@@ -675,8 +681,18 @@ class MenuOverlay extends Component {
                     />
                   ))}
 
-                {queriedResults.length === 0 && query !== '' && (
-                  <span>There are no results</span>
+                {queriedResults.length === 0 &&
+                  query !== '' &&
```
teosarca, line 685: IMHO, the `query !== ''` expression is not needed.

frontend/src/components/header/MenuOverlay.js
```diff
@@ -675,8 +681,18 @@ class MenuOverlay extends Component {
...
+                  !pendingQuery && (
+                    <span>
+                      {counterpart.translate('window.noResults.caption')}
+                    </span>
+                  )}
+
+                {pendingQuery && queriedResults.length === 0 && (
```
teosarca, line 692: IMHO the `queriedResults.length === 0` is not needed

frontend/src/containers/NavigationTree.js
```diff
@@ -107,13 +107,15 @@ class NavigationTree extends Component {
   };
 
   queryRequest = async (value) => {
+    this.setState({ pendingQuery: true });
     try {
       const response = await queryPathsRequest(value, '', true);
 
       await new Promise((resolve) =>
         this.setState(
```
teosarca, line 115: pls implement the same protection mechanism to make sure we get the last result.
(like [here](https://github.com/metasfresh/metasfresh/pull/12380/files#r789665656))

frontend/src/containers/NavigationTree.js
```diff
@@ -353,24 +356,33 @@ class NavigationTree extends Component {
...
+                )}
+            </div>
+            <div className="sitemap-column-right">
+              {sitemapRightColItems &&
+                sitemapRightColItems.map((subitem, subindex) =>
+                  this.renderMenuOverlayContainer(subitem, subindex)
+                )}
+            </div>
+
+            {queriedResults.length === 0 && query !== '' && (
```
teosarca, line 374: IMHO the expression shall be `!pendingQuery && queriedResults.length === 0`

frontend/src/assets/css/overlayfield.css
petrican: Will add it to header.css. ty

frontend/src/components/header/MenuOverlay.js
petrican: Actually it is needed otherwise you will see this message
Screenshot 2022-01-24 at 09 45 11

frontend/src/components/header/MenuOverlay.js
petrican: In this case yes it is not needed. Thanks

frontend/src/containers/NavigationTree.js
petrican: Same case as for the MenuOverlay with the remark that here the no results message will be visible only for a split second  (as the tables with sitemap data are rendered visible). So in this case we also need to have query!== ''

frontend/src/components/header/MenuOverlay.js
teosarca: ahh, understand.... so `query !== ''` means "we already queried for some results".

frontend/src/containers/NavigationTree.js
petrican: added

frontend/src/components/header/MenuOverlay.js
petrican: Added here too. But in this case things get more complicated as there is a debouncer used. So even with the changes done(using the suggestion) due to that debouncer we won't get the desired result. Exceeded the budget for the issue. Probably this should be tackled in a separate issue.

frontend/src/components/header/MenuOverlay.js
petrican: yep

frontend/src/components/header/MenuOverlay.js
```diff
@@ -651,6 +655,7 @@ class MenuOverlay extends Component {
                       this.handleKeyDown,
                       DEBOUNCE_TIME_SEARCH
                     )}
+                    onKeyUp={this.setLastQuery}
```
teosarca, line 658: what's the reason behind this?

teosarca: i mean, why is not enough to capture the last query exacly when we set the result?

petrican: have the last query string recorded so it wont be debounced

teosarca: so now it's OK to not have it in redux, right? :)

petrican: as the documentation that you pointed emphasises it is a balance. For the case of the Table because it is a complex component I chosen that redux update approach. In this case this is more appropriate (note that is not updating the state such that it doesn't cause re-rendering, it's updating local scoped flag)

teosarca: Q: why are u clearing it when u get the results?
IMHO u shall keep it in the state as follows:
* queriedResult - no change here, keep it like it is now
* query - the query which comes with the queriedResult or which is issued when we start to query ..... so there we always have the last query
* pendingQuery - flag to indicate if the `query` is currently running or not

why do we need to that `this.lastQuery` don't understand?
it feels like spaghetti just born...

frontend/src/components/header/MenuOverlay.js
```diff
@@ -96,18 +97,23 @@ class MenuOverlay extends Component {
...
         query: e.target.value,
+        pendingQuery: e.target.value ? true : false,
       });
 
       queryPathsRequest(e.target.value, 9)
         .then((response) => {
-          this.setState({
-            queriedResults: flattenLastElem(response.data),
-          });
+          if (e.target.value === this.queryString) {
```
teosarca, line 105: extract `e.target.value` to some `query` const... just to have a better naming for it.
also IMHO u shall compare it with the query from the state and not with that weirdly call and introduced field (this.queryString), see below...

frontend/src/components/header/MenuOverlay.js
```diff
@@ -96,18 +97,23 @@ class MenuOverlay extends Component {
...
+            this.setState({
+              queriedResults: flattenLastElem(response.data),
+              pendingQuery: false,
```
teosarca, line 108: set also `query: query`

teosarca: to know which is the query that was executed in order to get those `queriedResults`

frontend/src/components/header/MenuOverlay.js
```diff
@@ -96,18 +97,23 @@ class MenuOverlay extends Component {
...
+            });
+          }
         })
         .catch((err) => {
           if (err.response && err.response.status === 404) {
             this.setState({
               queriedResults: [],
+              pendingQuery: false,
```
teosarca, line 116: `query:''`?

frontend/src/components/header/MenuOverlay.js
```diff
@@ -575,6 +582,8 @@ class MenuOverlay extends Component {
     }
   }
 
+  setLastQuery = (e) => (this.queryString = e.target.value);
```
teosarca, line 585: imho not needed

teosarca (changes requested):
check my comments

frontend/src/components/header/MenuOverlay.js
petrican: I am setting it to the value of the input on keyUpEvent to have the string used to be queried stored without causing an extra  re-render due to state change. If I were to set this in the state using setState most probably there will be also times when it will be updated in the state after the completion of the query. Note that setState is async and wanted to avoid this. Now the issue that I am having is with the debouncing.

petrican: Screenshot 2022-01-25 at 14 56 23
Somehow the logic I want to have does not work as I will have all the time equality between the string from the input (e.target.value and the `this.queryString`)

petrican: I think we also need to increase the debounce time

frontend/src/components/header/MenuOverlay.js
```diff
@@ -96,18 +97,23 @@ class MenuOverlay extends Component {
     if (targetValue) {
       this.setState({
         query: e.target.value,
+        pendingQuery: e.target.value ? true : false,
```
teosarca, line 100: why not using `targetValue` instead of `e.target.value`.

Also, i would rename `targetValue` (what a name?!) to something like `query`.

frontend/src/components/header/MenuOverlay.js
teosarca: as I said, set the query too... because we are running async and we want to be consistent/on the safe side.

frontend/src/components/header/MenuOverlay.js
teosarca: compare the query with the one from the state and set it only if they are equal

frontend/src/containers/NavigationTree.js
teosarca: pls apply the same pattern u did on MenuOverlay.js .

frontend/src/components/header/MenuOverlay.js
petrican: query already set on l:99 so no need to set it again in here in case of error

frontend/src/components/header/MenuOverlay.js
petrican: done

frontend/src/components/header/MenuOverlay.js
teosarca: ohh boy, let me explain it to you by voice

frontend/src/containers/NavigationTree.js
```diff
@@ -107,13 +107,17 @@ class NavigationTree extends Component {
   };
 
   queryRequest = async (value) => {
+    this.setState({ pendingQuery: true, query: value });
+    const { query } = this.state;
     try {
       const response = await queryPathsRequest(value, '', true);
 
       await new Promise((resolve) =>
```
teosarca, line 115: pls take out this c**p: `await new Promise((resolve)`

frontend/src/containers/NavigationTree.js
```diff
@@ -149,7 +145,7 @@ class NavigationTree extends Component {
     const input = document.getElementById('search-input');
     const firstMenuItem = document.getElementsByClassName('js-menu-item')[0];
     let prevParentSibling = document.activeElement.previousSibling;
-
+    const elementToEnter = document.activeElement.childNodes[0];
```
teosarca, line 148: why not having it at line 178, right were is used?

teosarca (changes requested):
check my minor comment.
besides that lgtm, but before integrating it pls deploy it somewhere and let me check

frontend/src/containers/NavigationTree.js
petrican: eslint would complain that we have def within case blocks

teosarca (approved):

PR #12405, opened 2022-01-26
12317 search icons

petrican:
rel to: https://github.com/metasfresh/metasfresh/issues/12317

add icons to the search query results as well 

Screenshot 2022-01-26 at 14 22 17


petrican:
Updated such that we don't have the icon underlined
Screenshot 2022-01-27 at 09 19 25

PR #12412, opened 2022-01-27
fix CZ accountnumberlength; add additional tests

metas-ts:
fix CZ accountnumberlength; add additional failing tests to IBANValidationBLTest


metas-ts (commented):
lgtm! 
did you also verify that the C_Country records in the DB are configured the way they are set up in `IBANValidationBLTest`?

Chemssou:
@metas-ts :
Verification done, iban settings in code versus DB
Here is verification file: 
[checked_countries.ods](https://github.com/metasfresh/metasfresh/files/7973021/checked_countries.ods)

PR #12423, opened 2022-01-27
externalsystems https EP shall provide better Error-infos

cp-ps:
refs: https://github.com/metasfresh/metasfresh/issues/12422


metas-ts (approved):

PR #12427, opened 2022-01-27
Add sysconfig that allows to set the offset for initial PreparationDa…

riadmen:
…te calculation in de.metas.tourplanning.api.impl.OrderDeliveryDayBL

### DIFF
--- a/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/payment_allocation/InvoiceRow.java
+++ b/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/payment_allocation/InvoiceRow.java
@@ -61,7 +61,7 @@ public class InvoiceRow implements IViewRow
 	@Getter
 	private final String documentNo;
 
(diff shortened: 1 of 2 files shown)


backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/payment_allocation/InvoiceRow.java
```diff
@@ -61,6 +61,10 @@
 	@Getter
 	private final String documentNo;
 
+	@ViewColumn(seqNo = 20, widgetType = DocumentFieldWidgetType.Text, widgetSize = WidgetSize.Small, captionKey = "POReference")
```
metas-ts, line 64:
```suggestion
	@ViewColumn(seqNo = 25, widgetType = DocumentFieldWidgetType.Text, widgetSize = WidgetSize.Small, captionKey = "POReference")
```

..don't want to have seqNo=20 twice

metas-ts (approved):

Issue #12372, opened 2022-01-20
Support import formats linked to banks for bank statements


metas-rc:
### Is this a bug or feature request?

* FR
### What is the current behavior?

* Currently, the bank statement import is done based on an import format set in a config file. It is the same for any bank.

### What is the expected or desired behavior?
* Have the option to define different import formats for different banks

PR #12440, opened 2022-01-30
Data Import Per Bank

metas-rc:
https://github.com/metasfresh/metasfresh/issues/12372

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/banking/api/BankRepository.java
+++ b/backend/de.metas.business/src/main/java/de/metas/banking/api/BankRepository.java
@@ -22,39 +22,40 @@
 
 package de.metas.banking.api;
 
-import static org.adempiere.model.InterfaceWrapperHelper.newInstance;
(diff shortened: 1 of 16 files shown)


backend/de.metas.business/src/main/java/de/metas/banking/api/BankRepository.java
```diff
@@ -124,8 +136,25 @@ public Bank createBank(@NonNull final BankCreateRequest request)
...
+		if (bankId == null)
+		{
+			return retrieveDefaultBankDataImportConfigId();
+		}
+
+		final Bank bank = getById(bankId);
+
+		final DataImportConfigId dataImportConfigId = bank.getDataImportConfigId();
+
+		return dataImportConfigId != null ? dataImportConfigId : retrieveDefaultBankDataImportConfigId();
```
metas-ts, line 157: - here you can use CoalesceUtils.coalesceSupplierNotNull afais
- also AFAIU you should annotate the method with `@NonNull` to help intellij helping us

metas-ts (approved):

PR #12453, opened 2022-02-01
Alberta - order&product changes

fp-ps:


metas-ts (commented):
Thx, lgtm, i'm going to integrate, 
❗ but please also make sure to update the
https://github.com/metasfresh/metasfresh/blob/cloned_cereal_uat/misc/services/camel/de-metas-camel-externalsystems/alberta/README.md

PR #12466, opened 2022-02-03
order's effective bill-contact - only fall back if consistent

metas-ts:
only fall back to the order's AD_User as bill-contact if it's consisten with the order's Bill_Partner

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/order/IOrderBL.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/IOrderBL.java
@@ -80,11 +80,19 @@ public interface IOrderBL extends ISingletonService
 	 */
 	I_AD_User getShipToUser(I_C_Order order);
 
+	/**
+	 * @return the order's bill location <b>OR</b> falls back to the "general" contact ({@code C_Order.C_BParter_Location_ID}).
+	 */
+	@NonNull
 	BPartnerLocationAndCaptureId getBillToLocationId(I_C_Order order);
 
 	@Nullable
 	BPartnerId getEffectiveBillPartnerId(@NonNull I_C_Order orderRecord);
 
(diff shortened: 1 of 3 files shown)


backend/de.metas.business/src/main/java/de/metas/order/IOrderBL.java
```diff
@@ -80,11 +80,19 @@
 	 */
 	I_AD_User getShipToUser(I_C_Order order);
 
+	/**
+	 * @return the order's bill location <b>OR</b> falls back to the "general" contact ({@code C_Order.C_BParter_Location_ID}).
```
metas-ts, line 84:
```suggestion
	 * @return the order's bill location <b>OR</b> falls back to the "general" location ({@code C_Order.C_BParter_Location_ID}).
```

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
```diff
@@ -933,12 +937,28 @@ public boolean hasBillToContactId(@NonNull final I_C_Order order)
...
-				? billToContactId
-				: BPartnerContactId.ofRepoIdOrNull(order.getC_BPartner_ID(), order.getAD_User_ID());
+		if (billToContactId != null)
+		{
+			return billToContactId; // we are done
+		}
+
+		if (order.getAD_User_ID() <= 0)
+		{
+			return null; // nothing we can fall back to
```
metas-ts, line 951:
```suggestion
			return null; // nothing we could fall back to
```

metas-ts (commented):
lgtm

metas-ts:
going to integrate now; will keep a keen eye on the scary_plum_hotfix build

PR #12500, opened 2022-02-09
Gh49

riadmen:
https://github.com/metasfresh/ph171/issues/49


metas-ts (changes requested):
instead, please 
- add the payment-rule column to c_invoice_candidate
- propagate the payment-rule to the C_Invoice_Candidate when creating/updating the IC from an order line (see de.metas.order.invoicecandidate.C_OrderLine_Handler#setOrderedData(de.metas.invoicecandidate.model.I_C_Invoice_Candidate))
  - in the non, order cases, the payment-rule can be set to the default value (from bpartner, i guess)
  - add the payment rule to the default invoice aggregator
  - propagate it to the invoice, via invoiceHeader

Issue #12518, opened 2022-02-14
Copy incoterm location from order to invoice and location


cristinamghita:
### Is this a bug or feature request?
FR

PR #12519, opened 2022-02-14
#12518 Copy incoterm location from order to invoice and location 

cristinamghita:
#12518 Copy incoterm location from order to invoice and location


metas-ts (changes requested):
lgtm, the only thing missing AFAIU is to add the incoterm-location to the default ("minimal") invoice header aggregation

metas-ts (approved):

PR #12524, opened 2022-02-14
Gh12372 iban same org

metas-rc:
https://github.com/metasfresh/metasfresh/issues/12372

### DIFF
--- a/backend/de.metas.banking/de.metas.banking.base/src/main/java/de/metas/banking/impexp/BankStatementImportTableSqlUpdater.java
+++ b/backend/de.metas.banking/de.metas.banking.base/src/main/java/de/metas/banking/impexp/BankStatementImportTableSqlUpdater.java
@@ -1,18 +1,5 @@
 package de.metas.banking.impexp;
 
-import java.sql.PreparedStatement;
-import java.sql.ResultSet;
-import java.time.LocalDate;
-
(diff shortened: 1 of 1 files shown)


backend/de.metas.banking/de.metas.banking.base/src/main/java/de/metas/banking/impexp/BankStatementImportTableSqlUpdater.java
```diff
@@ -169,7 +169,10 @@ private void updateOrgBankAccount(final BankAccountId orgBankAccountId, final Im
 																+ " WHERE a." + I_C_BP_BankAccount.COLUMNNAME_IBAN
 																+ " = i." + I_I_BankStatement.COLUMNNAME_IBAN
 																+ " AND a.AD_Client_ID=i.AD_Client_ID "
-																+ " )"
+																+ " AND i." + I_I_BankStatement.COLUMNNAME_AD_Org_ID
```
metas-ts, line 172: please take into account that a bpartner and thus its bankaccount might also have AD_Org_ID=0

metas-ts (approved):
I think this will work for mad_orange, so if it's urgend, feel free to proceed.
But also please make sure to address my comment

PR #12531, opened 2022-02-15
shopware: receive customer => auto-send to rabbitmq

fp-ps:


metas-ts (changes requested):
generally LGTM. please just check out the minor layout request that i had

metas-ts (approved):

metas-ts:
going to integrate after sucessful build

PR #12556, opened 2022-02-18
shopware: renumber shopware-OLCand Lines

fp-ps:


backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_OrderLine.java
```diff
@@ -404,4 +405,16 @@ public void updateProductDocumentNote(final I_C_OrderLine orderLine)
 	{
 		orderLineBL.updateProductDocumentNote(orderLine);
 	}
+
+	@ModelChange(timings = { ModelValidator.TYPE_AFTER_NEW, ModelValidator.TYPE_AFTER_CHANGE },
+			ifColumnsChanged = { I_C_OrderLine.COLUMNNAME_IsGroupCompensationLine, I_C_OrderLine.COLUMNNAME_C_Order_CompensationGroup_ID })
+	public void renumberLinesIfCompensationGroupChanged(@NonNull final I_C_OrderLine orderLine)
```
metas-ts, line 411: this MI which fires for each order line makes me a bit nervous tbh. why do we need it? also, why with both after-new and after-change timings?

metas-ts (approved):

metas-ts (approved):

Issue #12571, opened 2022-02-21
Shopware - Call using config's orgId instead of user role's orgId


adi-stefan:
### Is this a bug or feature request?
FR
### What is the current behavior?
Currently, when calling Shopware, the orgId of the user's role is used. As such, imports may be done in the wrong organization.
Instead, the organization to use should be the one in the Shopware config: 
`SELECT ad_org_id FROM externalsystem_config_shopware6`

PR #12572, opened 2022-02-21
#12571 invoke Shopware actions using Shopware config's orgId. Cleanup

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/12571


backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/process/InvokeShopware6Action.java
```diff
@@ -177,71 +162,11 @@ protected ExternalSystemType getExternalSystemType()
 		return ExternalSystemType.Shopware6;
 	}
 
-	@NonNull
-	private String getConfigMappings(final ExternalSystemShopware6Config shopware6Config)
```
metas-ts, line 181: nice catch

metas-ts (approved):

Issue #12580, opened 2022-02-22
In OLCandSetOverrideValues Infer location by GLN, for given bpartner


adi-stefan:
### Is this a bug or feature request?
FR
### What is the current behavior?
Currently when overriding `C_OLCand.C_BPartner_Override_ID` then `C_OLCand.C_BP_Location_Override_ID` is also required.
### What is the expected or desired behavior?
The location should be inferred via the `C_OLCand.C_BPartner_Location_ID`'s GLN

PR #12582, opened 2022-02-22
#12580 renamed OLCandSetOverrideValues to C_OLCand_SetOverrideValues. Attempt to calculate C_BP_Location_Override_ID from C_BPartner_Override_ID

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/12580


backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/process/C_OLCand_SetOverrideValues.java
```diff
@@ -56,12 +75,35 @@ protected void prepare()
...
+			parameters.put(PARAM_Location, ProcessInfoParameter.of(PARAM_Location, getBPartnerLocationId(bpartnerId).getRepoId()));
+		}
+		params = new ProcessParams(ImmutableList.copyOf(parameters.values()));
+	}
 
-		if (Check.isNotBlank(parameters.get(PARAM_BPartner).getParameterAsString()) && Check.isBlank(parameters.get(PARAM_Location).getParameterAsString()))
+	private BPartnerLocationId getBPartnerLocationId(final int bpartnerId)
+	{
+		final Set<GLN> glns = createQuery().stream()
+				.map(record -> BPartnerLocationId.ofRepoId(record.getC_BPartner_ID(), record.getC_BPartner_Location_ID()))
```
metas-ts, line 89: i think this creates an `n+1`problem.
To avoid it, you can use one of the `IQueryBuilder#andCollect*` methods and get all the bpartner-locations  right away.

Also note that when looking for the *target* bpartner-location (location with the same GLN, but of the overriding-bpartner), please don't consider deactivatedd ones

metas-ts (commented):
AFAIU we need to build a map with old-bpartner-location-id => new-bpartner-location-id
and apply that map to the selected OLCands to get the right override-location for each of them.
..i can't see how this works in the PR

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/process/C_OLCand_SetOverrideValues.java
adi-stefan: rewritten to get everything in a single go.

adi-stefan:
I was under the impression that all selected records should have the same GLN. Which in hindsight makes no sense. Rewritten.

metas-ts (approved):

adi-stefan:
Originating issue was closed.

PR #12589, opened 2022-02-22
#12571: populate current timestamp when receiving external system log message

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/12571


metas-ts (approved):

metas-ts:
lgtm. please proceed. i checked the failed test; it's flaky and will be fixed soon, so you can ignore it here.

PR #12593, opened 2022-02-23
Fixes for dropship delivery and material tracking

metas-ts:

PR #12596, opened 2022-02-23
fix NPE when checking if AbstractProductionMaterial is a byProduct

metas-ts:


metas-ts:
merged into https://github.com/metasfresh/metasfresh/pull/12593

PR #12626, opened 2022-03-03
augment the message send to metasfresh when a shopware order is skipped

metas-ts:


backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/process/InvokeExternalSystemProcess.java
```diff
@@ -146,12 +146,13 @@ public Object getParameterDefaultValue(@NonNull final IProcessDefaultParameter p
 	}
 
 	/**
-	 * Needed so we also have a "since" when the process is run via AD_Scheduler
+	 * Needed so we also have a "since" when the process is run via AD_Scheduler.
+	 * This might be the process's last invocation time. Note that oftentimes, there is also a runtime-parameter with the actual valud used by the external system.
```
metas-ts, line 150:
```suggestion
	 * This might be the process's last invocation time. Note that oftentimes, there is also a runtime-parameter with the actual value used by the external system.
```

metas-ts (commented):
lgtm

Issue #12506, opened 2022-02-10
Inherit Invoice PaymentRule from Order


riadmen:
### Is this a bug or feature request?
BUG
### What is the expected or desired behavior?

Fix the problem that payment type is not inherited from order to invoice

PR #12627, opened 2022-03-03
when cloning an order, payment rule should be set as in the original …

riadmen:

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/invoice/interceptor/C_Invoice.java
+++ b/backend/de.metas.business/src/main/java/de/metas/invoice/interceptor/C_Invoice.java
@@ -177,7 +177,7 @@ public class C_Invoice // 03771
 	@CalloutMethod(columnNames = I_C_Invoice.COLUMNNAME_C_BPartner_ID)
 	public void setPaymentRule(final I_C_Invoice invoice)
 	{
-		if (!InterfaceWrapperHelper.isCopying(invoice))
+		if (!InterfaceWrapperHelper.isCopying(invoice) && !InterfaceWrapperHelper.isNew(invoice))
 		{
 			final I_C_BPartner bpartner = bpartnerDAO.getById(invoice.getC_BPartner_ID());
 			final PaymentRule paymentRule;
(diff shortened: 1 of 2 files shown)


backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
```diff
@@ -230,17 +230,20 @@ public void setDeliveryViaRule(final I_C_Order order)
...
 	public void setPaymentRule(final I_C_Order order)
 	{
-		final I_C_BPartner bpartner = order.getC_BPartner();
-		final PaymentRule paymentRule;
-		if (bpartner != null && bpartner.getPaymentRule() != null)
-		{
-			paymentRule = PaymentRule.ofCode(bpartner.getPaymentRule());
-		}
-		else
+		if (!InterfaceWrapperHelper.isCopying(order))
```
metas-ts, line 233: shouldn't we then have this if also in `backend/de.metas.business/src/main/java/de/metas/invoice/interceptor/C_Invoice.java` ?

metas-ts: shouldn't we also have `&& !InterfaceWrapperHelper.isNew(invoice)` here?

riadmen: If I understood correctly from Daniela::

- A new order is supposed to inherit the payment rule from the BPartener:
- A new invoice is supposed to inherit the payment rule from the related order.

so i think we shouldn't add it to `de.metas.order.model.interceptor.C_Order`

backend/de.metas.business/src/main/java/de/metas/invoice/interceptor/C_Invoice.java
```diff
@@ -177,17 +177,20 @@ public void removeMaterialLinesNotCorrespondingToPriceList(final I_C_Invoice inv
 	@CalloutMethod(columnNames = I_C_Invoice.COLUMNNAME_C_BPartner_ID)
 	public void setPaymentRule(final I_C_Invoice invoice)
 	{
-		final I_C_BPartner bpartner = bpartnerDAO.getById(invoice.getC_BPartner_ID());
-		final PaymentRule paymentRule;
-		if (bpartner != null && bpartner.getPaymentRule() != null)
+		if (!InterfaceWrapperHelper.isCopying(invoice) && !InterfaceWrapperHelper.isNew(invoice))
```
metas-ts, line 180: please add javadoc to explain why this method only acts if (!InterfaceWrapperHelper.isCopying(invoice) && !InterfaceWrapperHelper.isNew(invoice))`

❗ btw, what about the case that a new invoice without an order is created in the webui?

backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
metas-ts, line 233: please add javadoc to explain why this method only acts `if (!InterfaceWrapperHelper.isCopying(order))`

metas-ts (changes requested):

metas-ts (approved):

PR #12629, opened 2022-03-07
differentiate between SO and PO BPartner payment rule

riadmen:


metas-ts (approved):
lgtm (i think the null checks are probably not needed, but don't hurt either).
fine for me to integratre, but please create the test docu that @metas-dh asked for, before returning it back to the testing lane

Issue #12647, opened 2022-03-11
Do not allow modifying the product's uom if it already has uom conversions defined


riadmen:
### Is this a bug or feature request?
BUG
### What is the current behavior?

The problem is that the UOM conversion keeps the old initial UOM, even if it has been changed in the product.

#### Which are the steps to reproduce?

- Create a product, set a UOM to it. Let's call it `Product_UOM`
- Add conversion to another UOM `Conversion_UOM` 
- Change the `Product_UOM`to another value
- Check the conversion  => The old value of `Product_UOM` is displayed.

### What is the expected or desired behavior?

Do not allow modifying the product's uom if it already has uom conversions defined for the old one

PR #12648, opened 2022-03-11
Do not allow modifying the product's uom if it already has uom conver…

riadmen:
https://github.com/metasfresh/metasfresh/issues/12647

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/product/model/interceptor/M_Product.java
+++ b/backend/de.metas.business/src/main/java/de/metas/product/model/interceptor/M_Product.java
@@ -5,6 +5,8 @@ import de.metas.organization.OrgId;
 import de.metas.product.IProductPlanningSchemaBL;
 import de.metas.product.ProductId;
 import de.metas.product.ProductPlanningSchemaSelector;
+import de.metas.uom.IUOMConversionDAO;
+import de.metas.uom.UOMConversionsMap;
 import de.metas.util.Services;
 import lombok.NonNull;
 import org.adempiere.ad.callout.annotations.CalloutMethod;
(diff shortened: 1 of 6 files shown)


backend/de.metas.business/src/main/java/de/metas/product/model/interceptor/M_Product.java
```diff
@@ -66,4 +73,37 @@ private void createOrUpdateProductPlanningsForSelector(final @NonNull I_M_Produc
 		productPlanningSchemaBL.createOrUpdateProductPlanningsForSelector(productId, orgId, productPlanningSchemaSelector);
 	}
 
+	@ModelChange(timings = { ModelValidator.TYPE_BEFORE_NEW, ModelValidator.TYPE_BEFORE_CHANGE },
```
metas-ts, line 76:
```suggestion
	@ModelChange(timings = ModelValidator.TYPE_BEFORE_CHANGE ,
```
if the product is new, there can't be any `C_UOM_Conversion` yet..

backend/de.metas.business/src/main/java/de/metas/product/model/interceptor/M_Product.java
```diff
@@ -66,4 +73,37 @@ private void createOrUpdateProductPlanningsForSelector(final @NonNull I_M_Produc
...
+		final AdMessageKey errorMessage = checkExistingUOMConversions(product);
+
+		if (errorMessage != null)
+		{
+			throw new AdempiereException(errorMessage);
+		}
+	}
+
+	@Nullable
+	private AdMessageKey checkExistingUOMConversions(final I_M_Product product)
```
metas-ts, line 90: instead, please use `IUOMConversionDAO.getProductConversions` and `UOMConversionsMap.isEmpty` to figure out it there is a conversion

metas-ts (changes requested):

metas-ts (approved):

PR #12668, opened 2022-03-16
HU-Clearance support for aggregated HUs

av-ps:
refs: https://github.com/metasfresh/metasfresh/issues/12662

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/allocation/transfer/HUTransformService.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/allocation/transfer/HUTransformService.java
@@ -1161,6 +1161,9 @@ public class HUTransformService
 		return result.build();
 	}
 
+	/**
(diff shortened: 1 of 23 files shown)


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/allocation/transfer/HUTransformService.java
```diff
@@ -1161,6 +1161,26 @@ public I_M_HU huToNewSingleCU(@NonNull final HUsToNewCUsRequest newCURequest)
 		return result.build();
 	}
 
+	@NonNull
```
metas-ts, line 1164:
```suggestion
	/**
	  * If the given HU is aggregated, split out one TU and assign the given QR-code to it. Otherwise do nothing and jsut return the given {@code huId}.
	  */
	@NonNull
```

metas-ts (approved):

metas-ts:
going to integrate after sucessful build

PR #12681, opened 2022-03-18
GRS - Support multiple attachments in raw materials

av-ps:
refs: https://github.com/metasfresh/metasfresh/issues/12675


metas-ts (approved):

metas-ts:
the PR looks good to me. will try another full jenkins build

PR #12691, opened 2022-03-20
Gh12506 new payment rules correction

riadmen:


metas-ts (approved):

metas-ts (changes requested):
please extract the "UPDATE AD_Ref_List SET Value=..." statements into a different file, because we need to execute it also on the custom instances in any case.

metas-ts (approved):

PR #12700, opened 2022-03-22
#process proto changes

Chemssou:


backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/pporder/PPOrderLinesViewFactory.java
```diff
@@ -155,7 +155,8 @@ private ViewLayout createViewLayout(final WindowId windowId)
 	{
 		return ImmutableList.of(
 				createProcessDescriptorForIssueReceiptWindow(de.metas.ui.web.pporder.process.WEBUI_PP_Order_Receipt.class),
-				createProcessDescriptorForIssueReceiptWindow(de.metas.ui.web.handlingunits.process.WEBUI_M_HU_Pick.class),
```
metas-ts, line 158: what is `WEBUI_M_HU_Pick` about? is it now obsolete? if yes, please remove it altogether

metas-ts: ok, so I guess `WEBUI_M_HU_Pick` will stay in its own right, jsut won't be accessible from the production modal, right?

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/pporder/process/WEBUI_PP_Order_Pick_HU.java
```diff
@@ -0,0 +1,290 @@
...
+		// NOTE: we are not moving the HU to shipment schedule's locator.
+		pickingCandidateService.processForHUIds(ImmutableSet.of(huId),
+												shipmentScheduleId,
+												OnOverDelivery.ofTakeWholeHUFlag(isTakeWholeHU),
+												getView().getPpOrderId());
+	}
+
+	private Set<HuId> retrieveSelectedHuIds()
+	{
+		final DocumentIdsSelection selectedRowIds = getSelectedRowIds();
```
metas-ts, line 192: please extract the code that is common with WEBUI_PP_Order_PrintLabel into a super-class, utility-class or elsewhere, to deduplicate it

metas-ts: PS: also please look into doing the same for the code that is common with WEBUI_M_HU_Pick

Chemssou: >> ok, so I guess WEBUI_M_HU_Pick will stay in its own right, jsut won't be accessible from the production modal, right?
Yes indeed

PR #12722, opened 2022-03-25
Create Cucumber Tests for checking the effect of Override fields in invoice candidate windows

dm-ps:


metas-ts (approved):
lgtm; @metas-dh do you also want to fale a look, e.g. at the feature files

metas-ts (approved):

PR #12729, opened 2022-03-28
Set an ASI to the receipt schedule even if the order didn't have one

metas-rc:
https://github.com/metasfresh/metasfresh/issues/12728

### DIFF
--- a/backend/de.metas.business/src/main/java/org/adempiere/mm/attributes/api/IAttributeSetInstanceBL.java
+++ b/backend/de.metas.business/src/main/java/org/adempiere/mm/attributes/api/IAttributeSetInstanceBL.java
@@ -87,6 +87,8 @@ public interface IAttributeSetInstanceBL extends ISingletonService
 		return createASIFromAttributeSet(attributeSet, Predicates.alwaysTrue());
 	}
 
+	void cloneOrCreateASI(Object to, Object from);
+
 	I_M_AttributeSetInstance createASIFromAttributeSet(IAttributeSet attributeSet, Predicate<I_M_Attribute> filter);
 
 	I_M_AttributeSetInstance createASIWithASFromProductAndInsertAttributeSet(ProductId productId, IAttributeSet attributeSet);
--- a/backend/de.metas.business/src/main/java/org/adempiere/mm/attributes/api/IAttributeSetInstanceBL.java
+++ b/backend/de.metas.business/src/main/java/org/adempiere/mm/attributes/api/IAttributeSetInstanceBL.java
@@ -12,22 +12,24 @@ import org.compiere.model.I_M_Attribute;
 import org.compiere.model.I_M_AttributeInstance;
 import org.compiere.model.I_M_AttributeSetInstance;
 
+import javax.annotation.Nullable;
 import java.util.function.Predicate;
 
 /**
  * Service to create and update AttributeInstances and AttributeSetInstances.
  *
  * @author metas-dev <dev@metasfresh.com>
- *
  */
 public interface IAttributeSetInstanceBL extends ISingletonService
 {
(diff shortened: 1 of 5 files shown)


backend/de.metas.business/src/main/java/org/adempiere/mm/attributes/api/impl/AttributeSetInstanceBL.java
```diff
@@ -242,6 +242,41 @@ public void cloneASI(final Object to, final Object from)
 		}
 	}
 
+
+	@Override
+	public void cloneOrCreateASI(final Object to, final Object from)
```
metas-ts, line 247: please annotate them as `@Nullable`, both there an in the interface

backend/de.metas.business/src/main/java/org/adempiere/mm/attributes/api/impl/AttributeSetInstanceBL.java
```diff
@@ -242,6 +242,41 @@ public void cloneASI(final Object to, final Object from)
...
+
+		final IAttributeSetInstanceAware toASIAware = attributeSetInstanceAwareFactoryService.createOrNull(to);
+		if (toASIAware == null)
+		{
+			return;
+		}
+		final IAttributeSetInstanceAware fromASIAware = attributeSetInstanceAwareFactoryService.createOrNull(from);
+		if (fromASIAware == null)
+		{
+			return;
```
metas-ts, line 260: wdyt about creating a new ASI also in this case?

backend/de.metas.business/src/main/java/org/adempiere/mm/attributes/api/IAttributeSetInstanceBL.java
```diff
@@ -87,6 +87,8 @@ default I_M_AttributeSetInstance createASIFromAttributeSet(IAttributeSet attribu
 		return createASIFromAttributeSet(attributeSet, Predicates.alwaysTrue());
 	}
 
+	void cloneOrCreateASI(Object to, Object from);
```
metas-ts, line 90:
```suggestion
	/**
	  * Clone the ASI from the given {@code from} to the given {@code to}, or just create a new create empty ASI in {@code to} in case from is null or "none"
	  */
	void cloneOrCreateASI(Object to, Object from);
```

metas-ts (commented):
the ordering doesn't look wrong, but i can't see it's bednefit. so in future, everyone might be afraid of removing it. So please explain with a comment, why why we want to order the HUs.

I don't understand the code very well, but on a quick look at the code, it seems as if might be just to few serial numbers in the list. At least there seems to be no check or error, message or something along those lines, to handle that case..
so i think i would rather look into that direction..
Isn't that more likely to be the problem

metas-rc:
> the ordering doesn't look wrong, but i can't see it's bednefit. so in future, everyone might be afraid of removing it. So please explain with a comment, why why we want to order the HUs.


* Done


> I don't understand the code very well, but on a quick look at the code, it seems as if might be just to few serial numbers in the list. At least there seems to be no check or error, message or something along those lines, to handle that case..
so i think i would rather look into that direction..
Isn't that more likely to be the problem

* If there are a X serial numbers in the list and X<N (number of initial pieces in the CU) , then X CUs of qty 1 will be split out of the initial CU.
* If there are more serial numbers than N, N-1 will be split into new CUs, and one piece will remain in the initial CU:

Example

* CU has 4 pieces
image


* Insert 5 serial numbers

image


=>


* 3 CUs with qty 1 are split, the original one stays, but with qty 1, the first serial numbers were attributed to them

image

metas-ts (approved):

Issue #12772, opened 2022-04-06
Write a process that causes an OOME


LarsSander:
### Is this a bug or feature request?
Feature request
### What is the expected or desired behavior?
We should have a process that we can start, that causes an OutOfMemoryException

PR #12773, opened 2022-04-06
Cause out of memory exception

LarsSander:
https://github.com/metasfresh/metasfresh/issues/12772


metas-ts (approved):

metas-ts (changes requested):
very nice, but please change CauseOutOfMemoryException to CauseOutOfMemoryError.
Otherwise ppl might have FUD whether it's the real thing that can kill and application or something like a "soft" or "simulated" OOME

metas-ts (approved):

PR #12796, opened 2022-04-08
procurement-webui-frontend builds fail with TypeScript error

riadmen:
https://github.com/metasfresh/me03/issues/11578


metas-ts (approved):
feel free to integrate if build builds for you like this

PR #12813, opened 2022-04-12
Master db init dockerimage always run migrationscript

mrie04:


backend/metasfresh-dist/dist/src/main/docker/db-init/docker-entrypoint_init_only-14.2.sh
```diff
@@ -159,6 +159,34 @@ if [ "$1" = 'postgres' ]; then
 		echo
 		echo 'PostgreSQL init process complete; ready for start up.'
 		echo
+	else 
+		# always run migration script
```
metas-ts, line 163:
```suggestion
		# Always run migration scripts - in out case that's provision_metasfresh_db.sh.
		# For this we need to start the DB also if $PGDATA/PG_VERSION is there, then run the SQL and then stop the DB again.
```

metas-ts (approved):

Issue #12785, opened 2022-04-07
Shopware6 - Prevent NPE when unsetting companyName & error when different addresses point to same metasfresh address


adi-stefan:
### Is this a bug or feature request?
Bug
#### Which are the steps to reproduce?
1. Import order from Shopware6 with bpartner having 2 distinct locations which have the same `metasfreshLocationId`
2. Import order from Shopware6 for bpartner no longer having a `companyName`

PR #12819, opened 2022-04-13
#12785 always prefer metasfreshId over bodymedUserId as bpartner identifier

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/12785

### DIFF
--- a/misc/services/camel/de-metas-camel-externalsystems/de-metas-camel-shopware6/src/main/java/de/metas/camel/externalsystems/shopware6/api/model/order/OrderCandidate.java
+++ b/misc/services/camel/de-metas-camel-externalsystems/de-metas-camel-shopware6/src/main/java/de/metas/camel/externalsystems/shopware6/api/model/order/OrderCandidate.java
@@ -53,6 +53,6 @@ public class OrderCandidate
 	{
 		final JsonNode customIdNode = orderNode.at(customPath);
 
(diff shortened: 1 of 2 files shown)


misc/services/camel/de-metas-camel-externalsystems/de-metas-camel-shopware6/src/main/java/de/metas/camel/externalsystems/shopware6/api/model/order/OrderCandidate.java
```diff
@@ -49,16 +48,11 @@
...
 	{
 		final JsonNode customIdNode = orderNode.at(customPath);
 
-		if (customIdNode == null || Check.isBlank(customIdNode.asText()))
-		{
-			throw new RuntimeException("Failed to process order " + jsonOrder.getOrderNumber() + " (ID=" + jsonOrder.getId() + "); Nothing found on the given customPath: " + customPath);
-		}
-
-		return customIdNode.asText();
+		return (customIdNode == null) ? null : customIdNode.asText();
```
metas-ts:
```suggestion
		return (customIdNode == null || customIdNode.isNull()) ? null : customIdNode.asText();
```

if the property was there but with value `null` then afaiu we would otherwise return the string `"null"`

metas-ts (approved):

metas-ts (approved):

Issue #12830, opened 2022-04-14
Set compensated group id to orderLines in an ordered fashion


adi-stefan:
### Is this a bug or feature request?
bug
### What is the current behavior?
`C_OrderLine.C_Order_CompensationGroup_ID` is set to lines in the order in which they're created in the Orderline. If the first item is compensated, the group will not be created, as there must be at least one non-compensated product in each group.

PR #12823, opened 2022-04-13
#12785 fix tests, prevent error when userId is missing.

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/12785

PR #12831, opened 2022-04-14
#12830 sort order lines before setting groupId, so as to prevent having a group with only compensated products.

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/12830


metas-ts (approved):

adi-stefan:
merging, failing JUnits will be fixed in https://github.com/metasfresh/metasfresh/pull/12823

PR #12906, opened 2022-04-28
Add offset to preparation date

riadmen:


metas-ts (approved):
i think it's a step ahead, but please check my internal remark

PR #12917, opened 2022-05-03
don't attempt to upsert the contact if the bpartner is RO

metas-ts:
…rted

Also use the bpartner's sync-advice also for the contact, and not the location's sysnc advise


metas-ts (commented):
lgtm

PR #12927, opened 2022-05-05
Alberta products: update last date in external system

metas-rc:
https://github.com/metasfresh/metasfresh/issues/12926


metas-ts (commented):
looks to me as if it goes into the right direction 👍

PR #12950, opened 2022-05-10
Master docker build 060522

cp-ps:


metas-ts (commented):
the code looks plausible. 
please also make sure and document that 
- [x] it starts in your IDE
- [x] after successful build it can be deployed to and stat on a dev-instance (i can help there)
- it starts from documer (document the startup-cmd etc) when build on
  - [x] windows
  - [x] linux (i.e. github?)

mpeelen:
It starts on kubernetes with the image build under linux by github actions - which was the goal of this exercise. :)
From my point of view, we can merge and complete this.

PR #12954, opened 2022-05-10
Gh12953 invoicing support aggregating different values where possible

Chemssou:


metas-ts (approved):
good to integrate as an intermediate step

PR #12962, opened 2022-05-11
fix ProducerTemplate dependency injection

cp-ps:


misc/services/camel/de-metas-camel-externalsystems/core/src/main/java/de/metas/camel/externalsystems/core/AppConfiguration.java
```diff
@@ -55,7 +57,7 @@ public ProducerTemplate producerTemplate()
 	public void auditEventNotifier()
 	{
 		camelContext.getManagementStrategy()
-				.addEventNotifier(new AuditEventNotifier(context.getBean(ProducerTemplate.class)));
+				.addEventNotifier(new AuditEventNotifier(producerTemplate()));
```
metas-ts, line 60: hm..aren't we creating two ProducerTemplate instances now? this seems to be not the way..
did you also try
`public void auditEventNotifier(ProducerTemplate producerTemplate)` ? maybe, then spring will to some magic?
o, maybe put the ProducerTemplate into it's own bean

cp-ps: nope, we are not... anytime `producerTemaplte()` is called Spring will check first in the container and only if the bean is not already there will call the actual method and create a new one.. see docs extracts:

image

link to spring doc: https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans

(also tested to make sure we don't create multiple beans)

misc/services/camel/de-metas-camel-externalsystems/core/src/main/java/de/metas/camel/externalsystems/core/AppConfiguration.java
```diff
@@ -55,7 +57,7 @@ public ProducerTemplate producerTemplate()
 	public void auditEventNotifier()
 	{
 		camelContext.getManagementStrategy()
```
metas-ts, line 59:
```suggestion
		// note that calling producerTemplate() here does *not* mean wh create an additional instance. See https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans for details
		camelContext.getManagementStrategy()
```

metas-ts (approved):

Issue #12972, opened 2022-05-12
When removing `C_Commission_Share` records, also delete relevant `C_Commission_Fact` records


adi-stefan:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Error thrown when an invoice candidate with commission share is recalculated.

PR #12973, opened 2022-05-12
#12972 remove `C_Commission_Fact` related records when deleting `C_Commission_Share`s

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/12972

### DIFF
--- a/backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/commissioninstance/services/repos/CommissionInstanceRepository.java
+++ b/backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/commissioninstance/services/repos/CommissionInstanceRepository.java
@@ -1,5 +1,6 @@
 package de.metas.contracts.commission.commissioninstance.services.repos;
 
+import ch.qos.logback.classic.Level;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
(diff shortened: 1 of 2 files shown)


backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/commissioninstance/services/repos/CommissionInstanceRepository.java
```diff
@@ -380,11 +385,24 @@ private ImmutableMap<CommissionShare, I_C_Commission_Share> syncShareRecords(
 			result.put(share, shareRecord);
 		}
 
-		shareRecordsToDelete.forEach(InterfaceWrapperHelper::delete);
+		deleteShareRecords(shareRecordsToDelete);
```
metas-ts, line 388: i think it would be better to add a "before-delete" method to `de.metas.contracts.commission.commissioninstance.interceptor.C_Commission_Share`.
The hardcoded SQL is faster, but i think too dirty.

also, 
- please log to Loggable when a share is deleted (with commission-instance, sales rep, level and contract )
- replace the lambda with an old-school-for

metas-ts (changes requested):

backend/de.metas.contracts/src/main/java/de/metas/contracts/commission/commissioninstance/services/repos/CommissionInstanceRepository.java
```diff
@@ -380,11 +386,21 @@ private ImmutableMap<CommissionShare, I_C_Commission_Share> syncShareRecords(
...
 
-		shareRecordsToDelete.forEach(InterfaceWrapperHelper::delete);
+		shareRecords.forEach(this::logAndDeleteComissionShare);
 
 		return result.build();
 	}
 
+	private void logAndDeleteComissionShare(@NonNull final I_C_Commission_Share commissionShare)
+	{
+		Loggables.withLogger(logger, Level.INFO).addLog("Deleting commission share having instance: {}, salesRep: {}, level: {}, contract{}",
```
metas-ts, line 396:
```suggestion
		Loggables.withLogger(logger, Level.INFO).addLog("Deleting commission share having C_Commission_Instance_ID={}, C_BPartner_SalesRep_ID={}, LevelHierarchy={}, C_Flatrate_Term_ID={}",
```

metas-ts (approved):

PR #12975, opened 2022-05-13
Default paymentRule for packing item causes >1 invoice for the same o…

dm-ps:
…rder

refs: https://github.com/metasfresh/metasfresh/issues/12974


metas-ts (approved):
please proceed after successful build

PR #12981, opened 2022-05-13
Scale Price not used despite usescaleprice=S

mi-ps:


metas-ts (commented):
the code looks OK to me; going to wait for the automated tests though

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pricing/scalePrices.feature
```diff
@@ -0,0 +1,107 @@
...
+      | ps_3       | pp_1              | 0          | 0         | 4        | 100 |
+    And metasfresh contains C_Orders:
+      | Identifier | IsSOTrx | C_BPartner_ID.Identifier | DateOrdered |
+      | o_1        | true    | bpartner_1               | 2022-05-17  |
+    And metasfresh contains C_OrderLines:
+      | Identifier | C_Order_ID.Identifier | M_Product_ID.Identifier | QtyEntered |
+      | ol_1       | o_1                   | salesProduct_180520225  | 8          |
+      | ol_2       | o_1                   | salesProduct_180520225  | 15         |
+      | ol_3       | o_1                   | salesProduct_180520225  | 150        |
+    When update prices for order line
```
metas-ts, line 42: sooo...why do we need this step? doesn't that mean that the  scale prices would not be applied in the real world either?
I would have thought that there is some ModelInterceptor in chanrge of updating the order-line-prices automatically 🤔

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pricing/scalePrices.feature
```diff
@@ -0,0 +1,107 @@
...
```
metas-ts, line 72: (same ❓ )

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pricing/scalePrices.feature
```diff
@@ -0,0 +1,107 @@
...
+      | pp_1       | defaultPLV                        | salesProduct_180520225  | 5.00     | PCE               | Normal                        | Y                 |
+    And metasfresh contains C_Orders:
+      | Identifier | IsSOTrx | C_BPartner_ID.Identifier | DateOrdered |
+      | o_1        | true    | bpartner_1               | 2022-05-17  |
+    And metasfresh contains C_OrderLines:
+      | Identifier | C_Order_ID.Identifier | M_Product_ID.Identifier | QtyEntered |
+      | ol_1       | o_1                   | salesProduct_180520225  | 8          |
+      | ol_2       | o_1                   | salesProduct_180520225  | 15         |
+      | ol_3       | o_1                   | salesProduct_180520225  | 150        |
+    When update prices for order line
```
metas-ts, line 97: (same ❓ )

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pricing/scalePrices.feature
```diff
@@ -0,0 +1,107 @@
...
+      | C_OrderLine_ID.Identifier |
+      | ol_1                      |
+      | ol_2                      |
+      | ol_3                      |
+    And validate C_OrderLine:
+      | C_OrderLine_ID.Identifier | C_Order_ID.Identifier | dateordered | M_Product_ID.Identifier | qtyordered | qtydelivered | qtyinvoiced | price | discount | currencyCode | processed |
+      | ol_1                      | o_1                   | 2022-05-17  | salesProduct_180520225  | 8          | 0            | 0           | 5     | 0        | EUR          | false     |
+      | ol_2                      | o_1                   | 2022-05-17  | salesProduct_180520225  | 15         | 0            | 0           | 5     | 0        | EUR          | false     |
+      | ol_3                      | o_1                   | 2022-05-17  | salesProduct_180520225  | 150        | 0            | 0           | 5     | 0        | EUR          | false     |
```
metas-ts, line 106: i think this feature file is a good start, but I can't see the connection between it and the moving parts of HUPricing.java where the actual fix took place.
Afais the order-lines under test don't even have an attribute or HUPI after all..

metas-ts (changes requested):

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pricing/scalePrices.feature
cp-ps: > Afais the order-lines under test don't even have an attribute or HUPI after all..

yep, but that's what was actually broken with the HUPricing, namely if the order line didn't have any HUPI set, the HUPricing was looking at all product prices for that product

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pricing/scalePrices.feature
cp-ps: good point, the ones I saw were callouts but truth to be told, I didn't check all of them... so @mi-ps pls check if the price calculator will be called for a line created in cucumber

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pricing/scalePrices.feature
metas-ts: ok then it's fine with me

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pricing/scalePrices.feature
metas-ts: LE: so, i would like ask you to look into changing those callouts into MIs

mi-ps: @metas-ts @cp-ps yes, you are right, the updateOrderLine method gets called in the process of creating a new order line. 
So there is a method org.compiere.model.MOrderLine#setPrice() that eventually calls de.metas.order.impl.OrderLinePriceCalculator#updateOrderLine() and that gets called before saving the orderline as well as from various interceptors. The interceptor that we found in this case is de.metas.order.model.interceptor.C_OrderLine#updatePricesOverrideExistingDiscounts()

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pricing/scalePrices.feature
mi-ps: @metas-ts same answer as above

metas-ts (approved):

PR #12983, opened 2022-05-13
Korrektur action in material receipt candidates is not considered cor…

dm-ps:
…rectly in Material Dispo fix

refs: https://github.com/metasfresh/metasfresh/issues/12982


metas-ts (changes requested):
- looking at the fist, it seems to me as if we need the same fix for reversed shipments.
- also, i'm looking forward to the automated tests
- - @pvpurcarcosmin please consult with me if this goes beyond the scope

PR #12987, opened 2022-05-16
Password displayed when hovering over the *** with mouse, w/o using the eye icon

fp-ps:


metas-ts (changes requested):
the change looks plausible to me, but the build fails (https://jenkins.metasfresh.com/job/metasfresh/job/release_gh12986/1/console). please check

metas-ts (approved):

PR #12988, opened 2022-05-16
Export PP_Orders with extended infos

av-ps:


misc/de-metas-common/de-metas-common-product/src/main/java/de/metas/common/product/v2/request/constants/SwaggerDocConstants.java
```diff
@@ -24,9 +24,10 @@
 
 public class SwaggerDocConstants
 {
-	public static final String PRODUCT_IDENTIFIER_DOC = "Identifier of the product in question. Can be\n"
+	public static final String PRODUCT_IDENTIFIER_DOC = "Identifier of the product for the price in question. Can be\\n"
```
metas-ts, line 27: i superficially looked at the places where this constant is used, and i think that "for the price" doesn't really fit..or did i check too briefly?

metas-ts (changes requested):
minor stuff in here and some more in the preceeding review - pls check them out :-)

misc/de-metas-common/de-metas-common-product/src/main/java/de/metas/common/product/v2/request/constants/SwaggerDocConstants.java
av-ps: yes, you're right. my mistake

metas-ts (approved):

PR #12992, opened 2022-05-16
For Picking Profile "Group by Order" the order and bpartner are not displayed in Picking Terminal

fp-ps:


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/picking/modelvalidator/M_Picking_Config.java
```diff
@@ -38,4 +42,13 @@ public void onInit()
 	{
 		CacheMgt.get().enableRemoteCacheInvalidationForTableName(I_M_Picking_Config.Table_Name);
 	}
+
+	@ModelChange(timings = ModelValidator.TYPE_AFTER_CHANGE, ifColumnsChanged = I_M_Picking_Config.COLUMNNAME_WEBUI_PickingTerminal_ViewProfile)
+	public void resetCacheAfterViewProfileChange(final I_M_Picking_Config pickingConfig)
+	{
+		final CacheInvalidateMultiRequest invalidateMultiRequest = CacheInvalidateMultiRequest
```
metas-ts, line 49: this looks plausible, though i don't fully understand what it does..
...shouldn't it be enough to just invalidate the window with `AD_Window_ID` = `de.metas.ui.web.picking.PickingConstants#WINDOWID_PackageableView`?

fp-ps: this is the cache that needs to be invalidated `de.metas.ui.web.window.descriptor.factory.standard.DefaultDocumentDescriptorFactory#documentDescriptorsByWindowId`, but invalidating the window with `AD_Window_ID = de.metas.ui.web.picking.PickingConstants#WINDOWID_PackageableView` should also do the work. I'll make the change and test it. thx for the suggestion

PR #13003, opened 2022-05-17
Error on customer return

fp-ps:


metas-ts (changes requested):
i'm afraid that message is not helpful enough for most users.
It might be possible to include the reference to the bpartner in the notification. that woud be nice. But at any rate, please include the bpartner's value and name in the error message

metas-ts (approved):

PR #13005, opened 2022-05-18
attempt to fix cucumber tests that are failing
extract de.metas.bpartner.service.impl.BPartnerDAO.getContactLocationEmail

adi-stefan:
https://github.com/metasfresh/me03/issues/10564 - linking to this issue since it's about cucumber fixes on blonde_monkey_uat.


adi-stefan:
no longer relevant.

PR #13031, opened 2022-05-20
change order lines' project when the order's project is changed

metas-ts:


backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
```diff
@@ -200,6 +200,22 @@ public void updateReserved(final I_C_Order order)
 		}
 	}
 
+	@ModelChange(timings = ModelValidator.TYPE_BEFORE_CHANGE,
+			ifColumnsChanged = I_C_Order.COLUMNNAME_C_Project_ID)
+	public void updateProjectFromOrder(@NonNull final I_C_Order order)
+	{
+		if (order.getC_Project_ID() <= 0)
+		{
+			return; // let possilbe the existing project assignment be
```
metas-ts, line 209:
```suggestion
			return; // let possible existing project assignment be
```

metas-ts (commented):
lgtm

PR #13036, opened 2022-05-23
Give user-friendly error if bpartner has credit-stop

fp-ps:


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/event/InOutUserNotificationsProducer.java
```diff
@@ -189,11 +189,11 @@ public InOutUserNotificationsProducer notifyShipmentError(final String sourceInf
 
 	private void postNotification(final UserNotificationRequest notification)
 	{
-		notificationBL.sendAfterCommit(notification);
+		notificationBL.send(notification);
```
metas-ts, line 192: Y not sendAfterCommit?

fp-ps: when the wp fails, the trx is rolled back and with `sendAfterCommit` there is no error notification displayed.

metas-ts (approved):

PR #13051, opened 2022-05-26
fix cucumber scale prices tests after merge

fp-ps:


backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/automateOrderCandToOrderAndShipmentAndInvoice.feature
```diff
@@ -58,8 +58,8 @@ Feature: Process order candidate and automatically generate shipment and invoice
       | order_1               | 1444       | 2156425       | 2205175                | 2021-07-20  | SOO         | EUR          | F            | S               | po_ref_mock | true      | CL        |
 
     And validate the created order lines
```
metas-ts, line 60: eek, the lowercase qtyordered probably has to be changed back when we merge towards master, but in this branch i guess it's better this way..

metas-ts (approved):

PR #13105, opened 2022-06-03
Exclude Password Reset process from DebugMailTo

mi-ps:


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/email/EMail.java
```diff
@@ -1111,6 +1106,29 @@ public List<EMailAttachment> getAttachments()
 		return ImmutableList.copyOf(_attachments);
 	}
 
+	public void forceRealEmailRecipients()
```
metas-ts, line 1109:
```suggestion
	/**
	   *  Do send the mail to the respective mail address, even if the {@code DebugMailTo} SysConfig is set.
	   */
	public void forceRealEmailRecipients()
```

metas-ts (approved):

PR #13115, opened 2022-06-07
Provide default flatrate terms and boiler plate in the user Profile

metas-rc:
https://github.com/metasfresh/metasfresh/issues/13114

### DIFF
--- a/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/datatypes/json/JSONLookupValuesList.java
+++ b/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/datatypes/json/JSONLookupValuesList.java
@@ -143,7 +143,7 @@ public class JSONLookupValuesList
 
 	@JsonProperty("defaultValue")
 	@JsonInclude(JsonInclude.Include.NON_ABSENT)
-	private String defaultValue; // TODO rename to defaultId in a separate commit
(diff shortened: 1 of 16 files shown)


backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/datatypes/json/JSONLookupValuesList.java
```diff
@@ -194,14 +194,14 @@ public void putOtherProperty(final String name, final String jsonValue)
 	}
 
 	@JsonSetter
-	public JSONLookupValuesList setDefaultValue(final String defaultValue)
+	public JSONLookupValuesList setDefaultId(final String defaultId)
```
metas-ts, line 197: why rename to `defaultId`?

metas-rc: I renamed it after consulting with Teo. A String parameter called "defaultValue" might trick the dev into thinking that the expected unique identifier is the "Value" column, which is usually a unique String and is present in a lot of our tables. 
Instead, the expected default there is the ID, e.g. the AD_BoilerPlate_ID . I think this name is less confusing.

metas-ts (approved):

Issue #12948, opened 2022-05-10
Improve performance for report.fresh_qty_statistics_report_kg_week


cristinamghita:
### Is this a bug or feature request?
FR

PR #13117, opened 2022-06-08
#12948  Improve performance report.fresh_statistics_kg_week

cristinamghita:
#12948 Improve performance


metas-ts (commented):
the SQL lgtm. did you also already look into extending the timeout?

Issue #13158, opened 2022-06-15
Adjust Window Manufacturing candidate


riadmen:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?

Adjust Window `Produktionsdisposition`

- Add new field `Sequenz Nr.` (SeqNo)
  - Add this field to product planning too. when produktionsdispo data is generated by material dispo, then also copy that fieldcontend from product planning to produktionsdispo.
- Add field `Summe pro Tag`. 
  - Add this field into product planning too. when produktionsdispo data is generated by material dispo, then also copy that fieldcontend from product planning to produktionsdispo.
- Set default order by `Zugesagter Termin`, `Ressource`, `SeqNo`

PR #13159, opened 2022-06-16
Gh13158 adjust manufacturing candidate

riadmen:
https://github.com/metasfresh/metasfresh/issues/13158

### DIFF
--- a/backend/de.metas.manufacturing/src/main/java/org/eevolution/productioncandidate/service/CreateOrderCandidateCommand.java
+++ b/backend/de.metas.manufacturing/src/main/java/org/eevolution/productioncandidate/service/CreateOrderCandidateCommand.java
@@ -44,6 +44,7 @@ import org.eevolution.model.I_PP_Product_Planning;
 import org.eevolution.productioncandidate.model.dao.PPOrderCandidateDAO;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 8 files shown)


backend/de.metas.manufacturing/src/main/java/org/eevolution/productioncandidate/service/CreateOrderCandidateCommand.java
```diff
@@ -98,6 +99,11 @@ public I_PP_Order_Candidate execute()
 		ppOrderCandidateRecord.setC_OrderLine_ID(OrderLineId.toRepoId(request.getSalesOrderLineId()));
 		ppOrderCandidateRecord.setM_ShipmentSchedule_ID(ShipmentScheduleId.toRepoId(request.getShipmentScheduleId()));
 
+		final I_PP_Product_Planning productPlanning = productPlanningsRepo.getById(request.getProductPlanningId());
+		final BigDecimal qtyProcessed_OnDate = productPlanning.getQtyProcessed_OnDate();
```
metas-ts, line 103: this is kindow a shortcut..but i guess it's OK..mabe we need to go the full circle via material-dispo later though, wen we better understand how the two columns are used

metas-ts (approved):

PR #13167, opened 2022-06-17
Fix bug when looking up mail-config where docBaseAndSubType was ignored

metas-ts:
also improve logging if no AD_MailConfig was found and metasfresh falls back to the AD_Client's config


metas-ts (commented):
lgtm

PR #13173, opened 2022-06-17
Available for sales qty from metasfresh to shopware

dm-ps:


backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/availableforsales/AvailableForSalesQuery.java
```diff
@@ -33,20 +35,45 @@
...
+	public AvailableForSalesQuery(
+			@NonNull final ProductId productId,
+			@Nullable final WarehouseId warehouseId,
+			@NonNull final AttributesKeyPattern storageAttributesKeyPattern,
+			@NonNull final OrgId orgId,
+			@NonNull final Instant dateOfInterest,
+			final int shipmentDateLookAheadHours,
+			final int salesOrderLookBehindHours)
+	{
+		Check.assumeGreaterThanZero(orgId.getRepoId(), "OrgId must be regular!");
```
metas-ts, line 69: - you can actually use Check.errorUnless(orgId.isRegular(), ...) (if you want)
- the check is fine in that it makes clear we need a "real" org-id and not "*", but we can add more context to give a possible supporter a better chance to see which ShipmentSchedule etc this is all about. Suggestion:
```java
		Check.errorUnless(orgId.isRegular(), "AD_Org_Id={} must be regular! M_Product_ID={}, M_Warehouse_ID={}, AttributesKey={}, dateOfInterest=",
						  OrgId.toRepoId(orgId), ProductId.toRepoId(productId), WarehouseId.toRepoId(warehouseId), storageAttributesKeyPattern, dateOfInterest);
```

backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/availableforsales/AvailableForSalesRepository.java
```diff
@@ -70,40 +84,121 @@ public AvailableForSalesMultiResult getBy(@NonNull final AvailableForSalesMultiQ
...
+			@NonNull final ProductId productId,
+			@NonNull final OrgId orgId)
+	{
+		return queryBL.createQueryBuilder(I_MD_Available_For_Sales.class)
+				.addOnlyActiveRecordsFilter()
+				.addEqualsFilter(I_MD_Available_For_Sales.COLUMNNAME_M_Product_ID, productId)
+				.addEqualsFilter(I_MD_Available_For_Sales.COLUMNNAME_AD_Org_ID, orgId)
+				.create()
+				.listImmutable(I_MD_Available_For_Sales.class);
+	}
```
metas-ts, line 136: i think these three methods can be combined into just one (`getRecordsByQuery`)

backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/availableforsales/AvailableForSalesService.java
```diff
@@ -0,0 +1,249 @@
...
+					availableForSalesRepository.save(record);
+				});
+
+		final Set<I_MD_Available_For_Sales> recordsToBeDeleted = new HashSet<>(availableForSalesRecords);
+		recordsToBeDeleted.removeAll(recordsToBeUpdated.keySet());
+		recordsToBeDeleted.forEach(availableForSalesRepository::delete);
+
+		final Set<AvailableForSalesResult> recordsToBeCreated = new HashSet<>(availableForSalesResults);
+		recordsToBeCreated.removeAll(recordsToBeUpdated.values());
+		recordsToBeCreated.stream()
```
metas-ts, line 177: please use normal for-loop here. this stream is hard to debug

backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/availableforsales/AvailableForSalesService.java
```diff
@@ -0,0 +1,249 @@
...
+		final ImmutableList<AvailableForSalesResult> availableForSalesResults = computeAvailableForSales(AvailableForSalesMultiQuery.of(availableForSalesQuery))
+				.getAvailableForSalesResults();
+
+		final ImmutableList<I_MD_Available_For_Sales> availableForSalesRecords = availableForSalesRepository.getRecordsByQuery(availableForSalesQuery);
+
+		final ImmutableMap<I_MD_Available_For_Sales, AvailableForSalesResult> recordsToBeUpdated =
+				mapRecordsWithResultBasedOnASI(availableForSalesRecords, availableForSalesResults);
+
+		recordsToBeUpdated
+				.forEach((record, result) -> {
```
metas-ts, line 164: i don't really see the point of using forEach here, but since i can set my breakpoint in this lambda, it's fine with me :-)

metas-ts (changes requested):

backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/availableforsales/AvailableForSalesService.java
```diff
@@ -170,13 +172,21 @@ private void syncAvailableForSalesTable(@NonNull final AvailableForSalesQuery av
 
 		final Set<I_MD_Available_For_Sales> recordsToBeDeleted = new HashSet<>(availableForSalesRecords);
 		recordsToBeDeleted.removeAll(recordsToBeUpdated.keySet());
```
metas-ts, line 174: eek, i hadn't seen this before. it makes me nervous because i don't really know how hashcode is implemented.
i'm fine if you can explain how this is safe..otherwise please use some way that works with Integers, RepoIdAwares (e.g. instances of `Map<Integer, I_MD_Available_For_Sales>`) where we can be sure that it works

metas-ts (changes requested):

PR #13217, opened 2022-06-27
Tasty ham uat gh 13210

mi-ps:


metas-ts (changes requested):
hi, please check my remarks.
is the part about
> also make sure inactive records are not displayed in MD_Candidate window or production simulation modal

covered?

metas-ts (approved):

PR #13224, opened 2022-06-27
12287 show only projects matching organisation

Chemssou:


Chemssou:
Test table  updated taking in consideration projects defined in *.

PR #13233, opened 2022-06-30
Tell camel which auth token to use when connecting to MF

mi-ps:


backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/rabbitmq/authorization/CustomMFToExternalSystemQueueConfig.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.context.annotation.Configuration;
+
+import static de.metas.common.externalsystem.ExternalSystemConstants.QUEUE_NAME_MF_TO_ES_AUTH;
+
+@Configuration
+@EnableRabbit
+public class CustomMFToExternalSystemQueueConfig
+{
+	@Bean(name = QUEUE_NAME_MF_TO_ES_AUTH)
+	public Queue metasfreshFromAuthProcurementWebQueue()
```
cp-ps, line 37: same as above

backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/rabbitmq/authorization/listener/CustomAuthorizationRabbitMQListener.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.amqp.rabbit.annotation.RabbitListener;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import static de.metas.common.externalsystem.ExternalSystemConstants.QUEUE_NAME_ES_TO_MF_AUTH;
+
+@Component
+public class CustomAuthorizationRabbitMQListener
+{
+	private static final Logger logger = LoggerFactory.getLogger(CustomAuthorizationRabbitMQListener.class);
```
cp-ps, line 17: pls remove it as it's not used

backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/rabbitmq/authorization/listener/CustomAuthorizationRabbitMQListener.java
```diff
@@ -0,0 +1,31 @@
...
+
+	@Autowired
+	private ExternalSystemAuthorizationService authorizationService;
```
cp-ps, line 20: inject via constructor

backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/rabbitmq/authorization/listener/CustomAuthorizationRabbitMQListener.java
```diff
@@ -0,0 +1,31 @@
...
+
+	@RabbitListener(queues = QUEUE_NAME_ES_TO_MF_AUTH)
+	public void onRequestAuthorizationEvent(final JsonExternalSystemMessage message)
+	{
+		if (message.getType().equals(ExternalSystemConstants.TO_MF_AUTHORIZATION_REQUEST_MESSAGE_TYPE))
+		{
+			authorizationService.postAuthorizationReply();
+		}
+	}
```
cp-ps, line 29: 1. rename to `onMessage()` -> as this is not designed to be solely about authorization, hence the request types and so on

2. `final JsonExternalSystemMessage message` can be @NonNull

misc/de-metas-common/de-metas-common-externalsystem/src/main/java/de/metas/common/externalsystem/JsonExternalSystemMessage.java
```diff
@@ -0,0 +1,53 @@
...
+
+	@JsonProperty("payload")
+	JsonExternalSystemMessagePayload payload;
+
+	@Builder
+	@JsonCreator
+	public JsonExternalSystemMessage(
+			@NonNull @JsonProperty("type") final String type,
+			@Nullable @JsonProperty("payload") final JsonExternalSystemMessagePayload payload
+	)
```
cp-ps, line 47: as mentioned before:

- type -> enum
- payload -> string

misc/de-metas-common/de-metas-common-externalsystem/src/main/java/de/metas/common/externalsystem/JsonExternalSystemMessagePayload.java
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	@JsonCreator
+	public JsonExternalSystemMessagePayload(
+			@NonNull final String authToken
+	)
+	{
+		this.authToken = authToken;
+	}
+}
```
cp-ps, line 48: I think 

```
@JsonValue
	String authToken;
```

it's not such a great idea, wrapping the string with {"authToken":""} might help for clarity when checking messages

-> I suggest we go with a @Builder and @JsonDeserialize(builder)

misc/services/camel/de-metas-camel-externalsystems/core/src/main/java/de/metas/camel/externalsystems/core/AppConfiguration.java
```diff
@@ -61,19 +64,44 @@ public void auditEventNotifier()
 				.addEventNotifier(new AuditEventNotifier(producerTemplate()));
 	}
 
+	@PostConstruct
+	public void authorizationTokenNotifier()
+	{
+		final String defaultAuthToken = context.getEnvironment().getProperty(ExternalSystemCamelConstants.MF_API_AUTHORIZATION_TOKEN_PROPERTY);
+		// note that calling producerTemp8late() here does *not* mean wh create an additional instance. See https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans for details
```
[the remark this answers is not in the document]
cp-ps: also, instead of injecting the `context`, inject only the `MetasfreshAuthProvider`... since this in on a `postContruct ` handler, the bean should be there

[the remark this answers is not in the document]
cp-ps: also, I don't think we need the producer template here as the `MetasfreshAuthorizationTokenNotifier` should not be calling any routes

misc/services/camel/de-metas-camel-externalsystems/core/src/main/java/de/metas/camel/externalsystems/core/AppConfiguration.java
```diff
@@ -61,19 +64,44 @@ public void auditEventNotifier()
...
+				try
+				{
+					camelContext.getRouteController().startRoute(CUSTOM_TO_MF_ROUTE_ID);
+					camelContext.getRouteController().startRoute(CUSTOM_FROM_MF_ROUTE_ID);
+
+					context.getBean(ProducerTemplate.class)
+							.sendBody("direct:" + CUSTOM_TO_MF_ROUTE_ID, "trigger external system authentication for metasfresh!");
+
+					// context.getBean(ProducerTemplate.class)
+					// 		.sendBody("direct:" + HANDLE_EXTERNAL_SYSTEM_SERVICES_ROUTE_ID, "trigger rest api handler!");
```
cp-ps, line 99: ❗ this should be now called after we get the auth token from metasfresh => a nice way to do that would be to change the `HANDLE_EXTERNAL_SYSTEM_SERVICES_ROUTE_ID` definition, from `direct` to a `from("timer://runOnce?repeatCount=1)` that way, when enabled, it will automatically run.. that way it's completely decoupled from the authorization route

-> also, another thing to consider is the start-up order of the routes, as `HANDLE_EXTERNAL_SYSTEM_SERVICES_ROUTE_ID` needs all the other routes to be up and running.. so to achieve that we can take advantage of `org.apache.camel.model.RouteDefinition#startupOrder` -> and then, in our `CustomRouteController.startAllRoutes()` to take that into account

- even more, to have clear evidence of the ordering, I suggest we add an Enum for this rather than point blank setting some integers

misc/services/camel/de-metas-camel-externalsystems/core/src/main/java/de/metas/camel/externalsystems/core/authorizationmf/MetasfreshAuthorizationTokenNotifier.java
```diff
@@ -0,0 +1,109 @@
...
+	private final ProducerTemplate producerTemplate;
+
+	private final String defaultAuthToken;
+
+	public MetasfreshAuthorizationTokenNotifier(@NonNull final ApplicationContext context, @NonNull final ProducerTemplate producerTemplate, @Nullable final String defaultAuthToken)
+	{
+		this.producerTemplate = producerTemplate;
+		this.defaultAuthToken = defaultAuthToken;
+		this.context = context;
+	}
```
cp-ps, line 59: I've left a comment above about what this notifier should receive

misc/services/camel/de-metas-camel-externalsystems/core/src/main/java/de/metas/camel/externalsystems/core/authorizationmf/MetasfreshAuthorizationTokenNotifier.java
```diff
@@ -0,0 +1,109 @@
...
+				logger.log(Level.SEVERE, "Audit failed! and no pInstance could be obtained from source exchange!");
+				return;
+			}
+
+			final Map<String, Object> headers = new HashMap<>();
+			headers.put(HEADER_PINSTANCE_ID, sourceExchange.getIn().getHeader(HEADER_PINSTANCE_ID, Object.class));
+			headers.put(Exchange.EXCEPTION_CAUGHT, exception);
+
+			producerTemplate.sendBodyAndHeaders("direct:" + ERROR_WRITE_TO_ADISSUE, null, headers);
+		}
```
cp-ps, line 90: In case of an exception here we cannot do anything else but log the error as calling `ERROR_WRITE_TO_ADISSUE` will try to call MF-api and enter in an infinite loop...

One thing we could do, would be to shut down all routes if `MetasfreshAuthProvider` doesn't have an auth token for us... and send a custom message to request one.. @metas-ts wdyt?

misc/services/camel/de-metas-camel-externalsystems/core/src/main/java/de/metas/camel/externalsystems/core/authorizationmf/ToMFAuthorizationRouteBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+package de.metas.camel.externalsystems.core.authorizationmf;
+
```
cp-ps, line 24: `authorizationmf` -> `authorization`

misc/services/camel/de-metas-camel-externalsystems/core/src/main/java/de/metas/camel/externalsystems/core/to_mf/v2/SeqNoRouteBuilder.java
```diff
@@ -50,7 +49,7 @@ public void configure() throws Exception
 				.streamCaching()
 				.log("Route invoked!")
 				.removeHeaders("CamelHttp*")
-				.setHeader(CoreConstants.AUTHORIZATION, simple(CoreConstants.AUTHORIZATION_TOKEN))
+				// .setHeader(CoreConstants.AUTHORIZATION, simple(CoreConstants.AUTHORIZATION_TOKEN))
 				.setHeader(Exchange.HTTP_METHOD, constant(HttpEndpointBuilderFactory.HttpMethods.GET))
```
cp-ps, line 53: remove all comments like this :)

cp-ps (changes requested):

misc/services/camel/de-metas-camel-externalsystems/core/src/main/java/de/metas/camel/externalsystems/core/authorizationmf/MetasfreshAuthorizationTokenNotifier.java
metas-ts: aggree

backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/externalservice/authorization/ExternalSystemAuthorizationService.java
```diff
@@ -0,0 +1,87 @@
...
+				fromMFMessageSender.send(message);
+			} catch (Exception e)
+			{
+				final UserNotificationRequest request = UserNotificationRequest.builder()
+						.recipientUserId(Env.getLoggedUserId())
+						.subjectPlain("Metasfresh Authorization Token for External Systems")
+						.contentPlain("The authorization token could not be verified by metasfresh using sysconfig. This is the error that was thrown : " + e.getMessage() + " .")
+						.noEmail(false)
+						.build();
+				Services.get(INotificationBL.class).send(request);
```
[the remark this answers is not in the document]
cp-ps: 🤓 let's user a dedicated user group instead

backend/de.metas.externalsystem/src/main/sql/postgresql/system/80-de.metas.externalsystem/5645370_sys_gh13198_External_System_Auth_Token.sql
```diff
@@ -0,0 +1,83 @@
+-- 2022-06-28T13:55:04.441Z
+INSERT INTO AD_SysConfig (AD_Client_ID,AD_Org_ID,AD_SysConfig_ID,ConfigurationLevel,Created,CreatedBy,EntityType,IsActive,Name,Updated,UpdatedBy,Value) VALUES (0,0,541464,'S',TO_TIMESTAMP('2022-06-28 16:55:04','YYYY-MM-DD HH24:MI:SS'),100,'D','Y','EXTERNAL_SYSTEM_AUTHORIZATION_TOKEN',TO_TIMESTAMP('2022-06-28 16:55:04','YYYY-MM-DD HH24:MI:SS'),100,'2ac8d2ac2c0d485c970db5d96f89d809')
+;
+
+-- 2022-06-30T14:59:46.179Z
+UPDATE AD_SysConfig SET ConfigurationLevel='S', Description='This is a sys config with an authorization token that is used to check that metasfresh has the appropriate authorization token.', Name='de.metas.externalsystem.externalservice.authorization.authToken',Updated=TO_TIMESTAMP('2022-06-30 17:59:46','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_SysConfig_ID=541464
```
metas-ts, line 6: i don't unserstand the description. how is this value used? is is set to some `AD_User_AuthToken` value ? if yes - to which? (metasfresh / AD_USer_ID=100 ?) 

also, please 
- outline the relation between this value and the auth-token in `application.properties`
- if we don't remove the token properties from `application.properties` (why not?), then please also add an informative comment there

backend/de.metas.externalsystem/src/main/sql/postgresql/system/80-de.metas.externalsystem/5645370_sys_gh13198_External_System_Auth_Token.sql
```diff
@@ -0,0 +1,83 @@
...
+;
+
+-- 2022-07-01T14:27:12.019Z
+INSERT INTO AD_UserGroup (AD_Client_ID,AD_Org_ID,AD_UserGroup_ID,Created,CreatedBy,IsActive,Name,Updated,UpdatedBy) VALUES (1000000,1000000,540005,TO_TIMESTAMP('2022-07-01 17:27:12','YYYY-MM-DD HH24:MI:SS'),100,'Y','API-Setup',TO_TIMESTAMP('2022-07-01 17:27:12','YYYY-MM-DD HH24:MI:SS'),100)
```
metas-ts, line 10: please add an informative description to this group and also to the respective AD_SysConfig

metas-ts (changes requested):

backend/de.metas.externalsystem/src/main/sql/postgresql/system/80-de.metas.externalsystem/5645370_sys_gh13198_External_System_Auth_Token.sql
mi-ps: @metas-ts yes, right now it is it is set to an AD_User_AuthToken value, which, as per the prerequisites of the documentation is called "local-dev". I'm not sure which user is that assigned to.

Also, there is no relation between this value and the auth-token from application.properties. The value from application.properties is only used when sending a request to Metasfresh, if the auth-token is not set from the newly added logic.

@cp-ps @metas-ts should we remove it from application.properties

metas-ts: hi @mi-ps 
>  I'm not sure which user is that assigned to.

ok, then please find out how your feature works :-), and document it

> should we remove it from application.properties?

I understand from talks with cosmin that externalSystems will try getting a token from MF until it receives one. So, yes, please remove it. it's generating FUD.

metas-ts:
@mi-ps also, please 
- put a meaningful name as PR name - got example the internal issue's name
- but the metasfresh-issue's URL into the PR description

metas-ts (changes requested):

PR #13239, opened 2022-07-01
Showpare - percentage that is deducated from the available-to-sales qty

metas-ts:
refs https://github.com/metasfresh/metasfresh/issues/13238


metas-ts (commented):
lgtm

Issue #13237, opened 2022-07-01
Shopware - import business partners


adi-stefan:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
Import business partners regardless of number of orders associated to them.
Use pagination when importing.

PR #13245, opened 2022-07-04
#gh13237 add pagination to customer import

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/13237


misc/services/camel/de-metas-camel-externalsystems/de-metas-camel-shopware6/src/test/java/de/metas/camel/externalsystems/shopware6/customer/GetCustomersRouteBuilderTest.java
```diff
@@ -196,6 +196,7 @@ public void process(final Exchange exchange) throws IOException
 					.updatedAfter(updatedAfter)
 					.priceListBasicInfo(GetProductsRouteHelper.getTargetPriceListInfo(request))
 					.skipNextImportTimestamp(skipNextImportTimestamp)
+					.pageLimit(2)
```
metas-ts, line 199:
```suggestion
					.pageLimit(2) // this is a bit of a hack; we go with a page-size of two, knowing that our mocked paginated endpoint with return just one partner. so we'll invoke that EP just once
```

metas-ts (approved):

PR #13251, opened 2022-07-05
Shopware product variants  gh 13250

wgaulke:
I created a branch with an initial try to resolve the shopware variant import issue.

Approach:

- added parentId to JsonProduct
- added  filter step in getProductRoute to switch between variant and parent/flat product
- fetch parentProduct details
- merge parent with variant to fill missing fields
- upsert merged 

I added an extensive testcase based on a real api call to validate.

This solution basically creates exploded variants in mf. I spoke with @metasnw that proper variant support could be the next step.


metas-ts (approved):

wgaulke:
@metas-ts Thanks for approving. If there are any more changes necessary, please ping me :)

metas-ts:
@wgaulke thx for your patience!
FYI, we are going to cherry-pick your work to the release-branch so that it's going to be part of the upcoming version

Issue #13228, opened 2022-06-29
Basic style for the `Article Statistics` Report and certain Parameters mandatory


pauljakobseidenstuecker:
### Is this a bug or feature request?
FR
### What is the current behavior?
The Article statistics is not up to style guidelines and should be updated.
In the selection for the report the year should be mandatory and also show the current year
#### Which are the steps to reproduce?
From the navigation menu search for "Article Statistics" and select it. Enter an organisation, a year (it is empty in the beginning) and a period and click start to view the process.
### What is the expected or desired behavior?
When selecting the article statistics, the year should be filled (currently 2022) and deleting it should hinder the report from being started. When opening the report it should be follow the style guide (all Arial, alternating yellow lines, date and page count at the bottom, grey area around the cloumn header, etc.).

PR #13254, opened 2022-07-06
Remove the word alle from article statistics

pauljakobseidenstuecker:
- remove the word alle/all from the report in article statistics, instead show nothing (for empty reports)
  - rewritten expressions where it was used
  - removed it from language properties

for #13228


cristinamghita (approved):
lgtm

PR #13264, opened 2022-07-07
remove duplicated ebay dir.

wgaulke:


metas-ts (approved):
hi @wgaulke I might have intended to rename the folder `ebay` to `de-metas-camel-ebay`..
but doesn't matter. thx for the leanup

PR #13288, opened 2022-07-14
Create REST Endpoint to read and write WorkOrder- and Budget-C_Projects

fp-ps:


metas-ts:
FYI, i'm working on the WOProjectOrder code. please coordinate with me in case your also want to do changes in that area.

Issue #13227, opened 2022-06-29
Picking QA Step


cristinamghita:
### Is this a bug or feature request?
FR

PR #13289, opened 2022-07-14
#gh13227 prevent shipping orders with in-review picking jobs

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/13227


cristinamghita (approved):
lgtm

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/ShipmentScheduleWithHUService.java
```diff
@@ -654,6 +668,15 @@ private List<I_M_ShipmentSchedule_QtyPicked> retrieveQtyPickedRecords(final I_M_
 		return unshippedHUs;
 	}
 
+	/**
+	 * @param pickingJob the picking job to check
+	 * @return true if the picking job needs to be approved, but not approval is yet given.
```
metas-ts, line 673:
```suggestion
	 * @return true if the picking job needs to be approved, but no approval is yet given.
```

metas-ts (approved):

PR #13411, opened 2022-08-08
Iron plant uat gh13283 sq 2

dm-ps:


backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/project/budget/BudgetProjectRestService.java
```diff
@@ -0,0 +1,301 @@
...
+						.resourceIdentifier(String.valueOf(projectId.getRepoId()))
+						.build());
+
+		return toJsonBudgetProjectResponse(projectData);
+	}
+
+	@NonNull
+	public JsonBudgetProjectUpsertResponse upsertBudgetProject(@NonNull final JsonBudgetProjectUpsertRequest request)
+	{
+		return trxManager.callInNewTrx(() -> upsertBudgetProjectWithinTrx(request));
```
teosarca, line 109: why `callInNewTrx` and not `callInThreadInheritedTrx`?

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/project/budget/BudgetProjectRestService.java
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	@NonNull
+	private JsonBudgetProjectUpsertResponse upsertBudgetProjectWithinTrx(@NonNull final JsonBudgetProjectUpsertRequest request)
+	{
+		final SyncAdvise budgetProjectSyncAdvise = request.getSyncAdvise();
+		if (budgetProjectSyncAdvise == null)
```
teosarca, line 116: why not making syncAdvice mandatory and fail when JsonBudgetProjectUpsertRequest is constructed?

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/project/budget/BudgetProjectRestService.java
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	@NonNull
+	private Optional<BudgetProject> getExistingBudgetProject(
+			@NonNull final JsonBudgetProjectUpsertRequest request,
+			@NonNull final OrgId orgId)
+	{
+		final IdentifierString projectIdentifier = IdentifierString.of(request.getProjectIdentifier());
+
+		if (IdentifierString.Type.METASFRESH_ID.equals(projectIdentifier.getType()))
```
teosarca, line 239: improvement suggestion: introduce and use IdentifierString.isMetasfreshId()

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/project/resource/ResourceRestService.java
```diff
@@ -0,0 +1,83 @@
...
+	{
+		final Predicate<Resource> resourcePredicate;
+		switch (resourceIdentifier.getType())
+		{
+			case METASFRESH_ID:
+				resourcePredicate = r -> ResourceId.toRepoId(r.getResourceId()) == resourceIdentifier.asMetasfreshId().getValue();
+				break;
+			case VALUE:
+				resourcePredicate = r -> // make sure that org and value match
+						(r.getOrgId().isAny() || orgId.isAny() || Objects.equals(r.getOrgId(), orgId))
```
teosarca, line 61: pls use OrgId.equals(id1, id2) instead of Objects.equals

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/project/resource/ResourceRestService.java
```diff
@@ -0,0 +1,83 @@
...
+								&& Objects.equals(r.getValue(), resourceIdentifier.asValue());
+				break;
+			case INTERNALNAME:
+				resourcePredicate = r -> // make sure that org and value match
+						(r.getOrgId().isAny() || orgId.isAny() || Objects.equals(r.getOrgId(), orgId))
```
teosarca, line 66: pls use OrgId.equals(id1, id2) instead of Objects.equals

backend/de.metas.business/src/main/java/de/metas/project/budget/BudgetProjectRepository.java
```diff
@@ -23,28 +23,46 @@
...
 import java.util.Optional;
 
+import static org.adempiere.model.InterfaceWrapperHelper.saveRecord;
+
 @Repository
 public class BudgetProjectRepository
 {
-	final IQueryBL queryBL = Services.get(IQueryBL.class);
+	private final IQueryBL queryBL = Services.get(IQueryBL.class);
+	private final BudgetProjectResourceRepository budgetProjectResourceRepository;
```
teosarca, line 58: using a Repository from another Repository is usually a bad practice,
... and it's also a sign that:

Option 1. you might want to merge both repositories, BudgetProjectRepository and BudgetProjectResourceRepository, and consider BudgetProject as your aggregate root.
That means you are not allowed to retrieve `BudgetProjectResource` directly, but only by retrieving its aggregate root (BudgetProject).

Looking at BudgetProjectResourceRepository I think we don't want that because when we paint the visual calendar we have to fetch BudgetProjectResources from many projects.
So `BudgetProjectResource` it's an aggregate root on its own.

Option 2. Handle `BudgetProject` and `BudgetProjectResource` as two distinct aggregate roots (as we did until this change).
That means, 
* Each of BudgetProject and BudgetProjectResource shall have their own repository
* BudgetProject shall not contain BudgetProjectResource because from BudgetProject you are not able to load the BudgetProjectResource
* if you still need to have both of them somehow assembled (think about it, maybe you don't!), consider introducing BudgetProjectWithResources and have a service that queries both repositories and assembles your object.

wdyt?
cc @metas-ts

backend/de.metas.business/src/main/java/de/metas/project/budget/BudgetProjectRepository.java
```diff
@@ -78,12 +97,97 @@ public static Optional<BudgetProject> fromRecord(final I_C_Project record)
...
+		projectRecord.setDateContract(TimeUtil.asTimestamp(budgetProjectData.getDateContract()));
+		projectRecord.setDateFinish(TimeUtil.asTimestamp(budgetProjectData.getDateFinish()));
+		projectRecord.setC_Project_Reference_Ext(budgetProjectData.getProjectReferenceExt());
+
+		saveRecord(projectRecord);
+
+		final ProjectId projectId = ProjectId.ofRepoId(projectRecord.getC_Project_ID());
+
+		upsertBudgetProjectRequest.getProjectResourceRequests()
+				.forEach(upsertBudgetProjectResourceRequest -> budgetProjectResourceRepository.save(upsertBudgetProjectResourceRequest, projectId));
```
teosarca, line 188: here we have SQL N+1 performance issue.
(if you are not familiar, a description of that can be found here: https://vladmihalcea.com/n-plus-1-query-problem/ ... and in many other places).

Basically, if your `upsertBudgetProjectRequest.getProjectResourceRequests()` has 100 items, it means you will issue 100 SELECTs to load the database record.

How to solve it?
* send the whole list `upsertBudgetProjectRequest.getProjectResourceRequests()` to budgetProjectResourceRepository.save(..). You can even rename it as `saveAll` or better `createOrUpdateAll` because basically that's what it does
* The new `createOrUpdateAll` shall:
   * load all existing records all together (in one select)
   * iterate the upserts and update existing record if any or create a new record
   * save each record, one by one (on save side, we don't have batching... so saving one by one it's the only option :( ).

Here is an example on how to do that: https://github.com/metasfresh/metasfresh/blob/inner_silence_uat/backend/de.metas.business/src/main/java/de/metas/costrevaluation/CostRevaluationRepository.java#L76-L104

cc @metas-ts

backend/de.metas.business/src/main/java/de/metas/project/budget/BudgetProjectResource.java
```diff
@@ -22,60 +22,24 @@
...
-	@NonNull UomId durationUomId;
-	@NonNull Money plannedAmount;
-	@NonNull Quantity plannedDuration;
-	@NonNull Money pricePerDurationUnit;
+	@NonNull
+	BudgetProjectResourceId id;
 
-	@NonNull CalendarDateRange dateRange;
+	@NonNull
+	BudgetProjectResourceData budgetProjectResourceData;
```
teosarca, line 38: IMHO introducing this "Data" object to be used in the actual model and also in upsert Request it's not the best design.

Why?
* because not all the fields of the model are updatable from outside. Maybe they are now, but they won't be in the future. So what will you do then? You will move the not-updatable field from BudgetProjectResourceData directly into BudgetProjectResource and the whole BudgetProjectResource will be a mambo-jambo of fields here and there. Does not feel right.
* the modal is very tightly coupled to the data importing. So changing/improving the data import will influence the code which actually works with those models and vice-versa.
* While it feels super easy to work with imports, it's a nightmare for developers which have to deal with actual data.
* import/export shall not shape the model design

I would suggest:
* have the model as it was before (without the data object)
* have dedicated methods which updates the records from model or from request. Even if it looks like copy-paste, and feels like breaking DRY principle, IMHO this is one of the cases you have too. Btw, here is a nice video explaining a bit more about DRY pros/cons: https://www.youtube.com/watch?v=cqKGDpnE4eY&t=1s

cc @metas-ts

backend/de.metas.business/src/main/java/de/metas/project/budget/BudgetProjectResourceRepository.java
```diff
@@ -193,4 +209,32 @@ public InSetPredicate<ProjectId> getProjectIdsPredicateByResourceGroupIds(
 		final ImmutableSet<ProjectId> projectIdsEffective = ProjectId.ofRepoIds(projectRepoIdsEffective);
 		return InSetPredicate.only(projectIdsEffective);
 	}
+
+	@NonNull
+	public BudgetProjectResource save(
```
teosarca, line 214: wrote about this somewhere above. consider implementing createOrUpdateAll(List<UpsertBudgetProjectResourceRequest> requests, ProjectId projected)

backend/de.metas.business/src/main/java/de/metas/project/budget/BudgetProjectResourceService.java
```diff
@@ -0,0 +1,114 @@
...
+
+	private final ResourceService resourceService;
+
+	public BudgetProjectResourceService(@NonNull final ResourceService resourceService)
+	{
+		this.resourceService = resourceService;
+	}
+
+	@NonNull
+	public Optional<BigDecimal> computePlannedDuration(@NonNull final I_C_Project_Resource_Budget budgetResource)
```
teosarca, line 56: why not returning `Duration` instead of `BigDecimal`?

teosarca: NOTE: in DurationUtils you have a couple of helper methods to deal with time conversion, Duration and BigDecimal

backend/de.metas.business/src/main/java/de/metas/project/budget/BudgetProjectResources.java
```diff
@@ -100,4 +102,12 @@ public BudgetProjectResource getBudgetById(@NonNull final BudgetProjectResourceI
 				.orElseThrow(() -> new AdempiereException("No budget found for " + id));
 	}
 
+	@NonNull
+	public Optional<BudgetProjectResource> findBudget(@NonNull final ResourceId resourceId)
+	{
+		return budgets.stream()
+				.filter(budgetProjectResource -> budgetProjectResource.getBudgetProjectResourceData().getResourceId() != null)
```
teosarca, line 109: remark, as I described above, now all the code as poluted by `budgetProjectResource.getBudgetProjectResourceData()`.

backend/de.metas.business/src/main/java/de/metas/project/budget/BudgetProjectRepository.java
```diff
@@ -69,7 +87,8 @@ public List<BudgetProject> queryAllActiveProjects(@NonNull final InSetPredicate<
 				.collect(ImmutableList.toImmutableList());
 	}
 
-	public static Optional<BudgetProject> fromRecord(final I_C_Project record)
+	@NonNull
+	public Optional<BudgetProject> fromRecord(final I_C_Project record)
```
teosarca, line 91: usually, if you cannot make this method a pure function (basically make it static, don't use external services, no side effects), it's a sign of bad design.

backend/de.metas.business/src/main/java/de/metas/project/budget/BudgetProjectResourceData.java
```diff
@@ -0,0 +1,104 @@
...
+import de.metas.uom.UomId;
+import de.metas.util.lang.ExternalId;
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
+
+import javax.annotation.Nullable;
+
+@Value
+public class BudgetProjectResourceData
```
teosarca, line 40: IMHO we shall get rid of this.

backend/de.metas.business/src/main/java/de/metas/project/budget/interceptor/C_Project_Resource_Budget.java
```diff
@@ -40,23 +44,25 @@
...
 	{
 		this.multiCalendarService = multiCalendarService;
+		this.budgetProjectResourceService = budgetProjectResourceService;
+		this.resourceService = resourceService;
 	}
 
-	@ModelChange(timings = { ModelValidator.TYPE_BEFORE_NEW, ModelValidator.TYPE_BEFORE_CHANGE })
-	public void beforeSave(final I_C_Project_Resource_Budget record)
+	@ModelChange(timings = { ModelValidator.TYPE_AFTER_NEW, ModelValidator.TYPE_AFTER_CHANGE })
+	public void afterSave(final I_C_Project_Resource_Budget record, final ModelChangeType changeType)
```
teosarca, line 61: do you still need `, final ModelChangeType changeType` param?

backend/de.metas.business/src/main/java/de/metas/project/budget/BudgetProjectResourceService.java
teosarca: update: now I saw that it was refactored from some private method and moved here and it's convenient to return BigDecimal.
In this case i would suggest to write a small javadoc and document which is the UOM of returned BigDecimal.

backend/de.metas.business/src/main/java/de/metas/project/shared/ProjectSharedService.java
```diff
@@ -0,0 +1,96 @@
...
+import de.metas.util.Services;
+import lombok.NonNull;
+import org.compiere.util.Env;
+import org.compiere.util.Evaluatees;
+import org.springframework.stereotype.Service;
+
+import javax.annotation.Nullable;
+
+@Service
+public class ProjectSharedService
```
teosarca, line 42: What's the purpose of this ProjectSharedService?
Why not using ProjectService?

backend/de.metas.business/src/main/java/de/metas/project/service/ProjectRepository.java
```diff
@@ -144,39 +130,15 @@ public ImmutableSet<ProjectId> getProjectIdsDownStream(@NonNull final Collection
...
-		}
-
-		final ProjectType projectType = projectTypeRepository.getById(projectTypeId);
-		projectRecord.setProjectCategory(projectType.getProjectCategory().getCode());
-	}
-
 	@NonNull
 	private Project save(
 			@NonNull final I_C_Project projectRecord,
 			@NonNull final ProjectData projectData)
```
teosarca, line 136: i have the feeling projectData has the same problem as budgetProjectData

backend/de.metas.business/src/main/java/de/metas/project/shared/ProjectSharedService.java
```diff
@@ -0,0 +1,96 @@
...
+		if (docSequenceId == null)
+		{
+			return null;
+		}
+
+		final DocumentSequenceInfo documentSequenceInfo = documentSequenceDAO.retriveDocumentSequenceInfo(docSequenceId);
+
+		return documentNoBuilderFactory.createDocumentNoBuilder()
+				.setDocumentSequenceInfo(documentSequenceInfo)
+				.setClientId(Env.getClientId())
```
teosarca, line 91: pls avoid Env.getClientId(). Introduce it as parameter or used it project ProjectType.

backend/de.metas.business/src/main/java/de/metas/project/shared/ProjectSharedService.java
```diff
@@ -0,0 +1,96 @@
...
+				.setEvaluationContext(Evaluatees.ofCtx(Env.getCtx()))
```
teosarca, line 92: why do we need that EvaluationContext. Why not leave it empty?
Basically, i am trying to avoid using global contexts.

backend/de.metas.business/src/main/java/de/metas/project/shared/ProjectSharedService.java
```diff
@@ -0,0 +1,96 @@
...
+		{
+			return null;
+		}
+
+		final ProjectType projectType = projectTypeRepository.getById(projectTypeId);
+		return projectType.getProjectCategory();
+	}
+
+	@Nullable
+	public String getValueForProjectType(@Nullable final ProjectTypeId projectTypeId)
```
teosarca, line 72: i suggest to rename it to `getAndIncrementNextProjectValue` or `getNextProjectValue`

backend/de.metas.business/src/main/java/de/metas/project/shared/ProjectSharedService.java
```diff
@@ -0,0 +1,96 @@
...
+	public ProjectSharedService(
+			@NonNull final IDocumentNoBuilderFactory documentNoBuilderFactory,
+			@NonNull final ProjectTypeRepository projectTypeRepository)
+	{
+		this.documentNoBuilderFactory = documentNoBuilderFactory;
+		this.projectTypeRepository = projectTypeRepository;
+	}
+
+	@Nullable
+	public ProjectCategory getProjectCategoryFromProjectType(@Nullable final ProjectTypeId projectTypeId)
```
teosarca, line 60: pls avoid invalid params like:
* accepting nullable parameter (@Nullable final ProjectTypeId projectTypeId))
* first thing in the method is `if is null then return null`.

Basically,
* the param shall be annotated with `@NonNull`
* the caller shall make sure that it's calling this method with non null argument.

Spreading this kind of null check everywhere
* makes the code less readable
* spreads FUD
* bad design

cc @metas-ts

backend/de.metas.business/src/main/java/de/metas/project/workorder/calendar/ToCalendarEntryConverter.java
```diff
@@ -26,8 +28,10 @@
 class ToCalendarEntryConverter
 {
 	private static final Logger logger = LogManager.getLogger(ToCalendarEntryConverter.class);
-	@NonNull private final IUOMDAO uomDAO = Services.get(IUOMDAO.class);
-	@NonNull private final ProjectFrontendURLsProvider frontendURLs = new ProjectFrontendURLsProvider();
+	@NonNull
+	private final IUOMDAO uomDAO = Services.get(IUOMDAO.class);
```
teosarca, line 32: pls check your intelliJ code style:
image

@metas-ts  don't we have these settings in our team coding styles?

misc/de-metas-common/de-metas-common-rest_api/src/main/java/de/metas/common/rest_api/v2/money/JsonMoney.java
```diff
@@ -0,0 +1,53 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
+
+import java.math.BigDecimal;
+
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY, getterVisibility = JsonAutoDetect.Visibility.NONE, isGetterVisibility = JsonAutoDetect.Visibility.NONE, setterVisibility = JsonAutoDetect.Visibility.NONE)
+@Value
+public class JsonMoney
```
teosarca, line 36: you can simply annotate this class with `@Value`, `@Builder` and `@Jacksonized`.

and you can remove:
* ctor
* `@JsonAutoDetect`
* `@JsonProperty`
* `@JsonCreator`

Just have a skinny class.
`@Jacksonized` does everything.

misc/de-metas-common/de-metas-common-rest_api/src/main/java/de/metas/common/rest_api/v2/project/budget/JsonBudgetProjectResourceResponse.java
```diff
@@ -0,0 +1,156 @@
...
+import lombok.NonNull;
+import lombok.Value;
+
+import javax.annotation.Nullable;
+import java.math.BigDecimal;
+import java.time.LocalDate;
+
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY, getterVisibility = JsonAutoDetect.Visibility.NONE, isGetterVisibility = JsonAutoDetect.Visibility.NONE, setterVisibility = JsonAutoDetect.Visibility.NONE)
+@Value
+public class JsonBudgetProjectResourceResponse
```
teosarca, line 42: consider using `@Jacksonized`.
Search for examples in our code.

misc/de-metas-common/de-metas-common-rest_api/src/main/java/de/metas/common/rest_api/v2/project/budget/JsonBudgetProjectResponse.java
```diff
@@ -0,0 +1,174 @@
...
+import lombok.Singular;
+import lombok.Value;
+
+import javax.annotation.Nullable;
+import java.time.LocalDate;
+import java.util.List;
+
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY, getterVisibility = JsonAutoDetect.Visibility.NONE, isGetterVisibility = JsonAutoDetect.Visibility.NONE, setterVisibility = JsonAutoDetect.Visibility.NONE)
+@Value
+public class JsonBudgetProjectResponse
```
teosarca, line 43: use `@Jacksonized`

misc/de-metas-common/de-metas-common-rest_api/src/main/java/de/metas/common/rest_api/v2/project/budget/JsonBudgetProjectUpsertResponse.java
```diff
@@ -0,0 +1,60 @@
...
+import de.metas.common.rest_api.v2.JsonResponseUpsertItem;
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Singular;
+import lombok.Value;
+
+import java.util.List;
+
+@Value
+public class JsonBudgetProjectUpsertResponse
```
teosarca, line 37: use `@Jacksonized`

misc/de-metas-common/de-metas-common-rest_api/src/main/java/de/metas/common/rest_api/v2/project/workorder/JsonWorkOrderProjectResponse.java
```diff
@@ -41,7 +41,7 @@
 	@JsonProperty("projectId")
 	JsonMetasfreshId projectId;
 
-	@NonNull
+	@Nullable
```
teosarca, line 44: why nullable?

misc/de-metas-common/de-metas-common-rest_api/src/main/java/de/metas/common/rest_api/v2/project/workorder/JsonWorkOrderStepResponse.java
```diff
@@ -64,20 +64,25 @@
...
+
 	@Nullable
 	@JsonProperty("resources")
 	List<JsonWorkOrderResourceResponse> resources;
 	
 	@JsonCreator
 	@Builder
 	public JsonWorkOrderStepResponse(
-			@NonNull @JsonProperty("stepId") final JsonMetasfreshId stepId,
+			@Nullable @JsonProperty("stepId") final JsonMetasfreshId stepId,
```
teosarca, line 78: why nullable?

misc/de-metas-common/de-metas-common-rest_api/src/main/java/de/metas/common/rest_api/v2/project/workorder/JsonWorkOrderStepResponse.java
```diff
@@ -64,20 +64,25 @@
...
 			@NonNull @JsonProperty("name") final String name,
-			@NonNull @JsonProperty("projectId") final JsonMetasfreshId projectId,
+			@Nullable @JsonProperty("projectId") final JsonMetasfreshId projectId,
```
teosarca, line 80: why nullable?

teosarca (changes requested):
check my comments

backend/de.metas.business/src/main/java/de/metas/project/shared/ProjectSharedService.java
metas-ts: aggree most of the time (incl here)

backend/de.metas.business/src/main/java/de/metas/project/workorder/calendar/ToCalendarEntryConverter.java
metas-ts: not sure...at any rate, my settings also say "Field annotations" => always wrap

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/project/budget/BudgetProjectRestService.java
dm-ps: bc JsonBudgetProjectUpsertRequest is not Immutable && we build it via setters in camel so I think it will be weird to give some fields via constructor/builder and others via setters.. (we use setters in order to know which fields were actually set - i.e. `isNameSet` )

backend/de.metas.business/src/main/java/de/metas/project/budget/interceptor/C_Project_Resource_Budget.java
dm-ps: Yes, it's still needed in de.metas.project.budget.interceptor.C_Project_Resource_Budget#notifyIfUserChange

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/project/ProjectRestService.java
```diff
@@ -66,10 +70,14 @@
 	private final ICurrencyDAO currenciesRepo = Services.get(ICurrencyDAO.class);
 
 	private final ProjectRepository projectRepository;
+	private final ProjectService projectService;
 
-	public ProjectRestService(@NonNull final ProjectRepository projectRepository)
+	public ProjectRestService(
+			@NonNull final ProjectRepository projectRepository,
+			@NonNull final ProjectService projectService)
```
teosarca, line 77: leave it like this now,
but usualy when u see repo and service one near the other, 
it's a sign that you shall use only the service and let the service expose the needed repo method(s).

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/project/budget/BudgetProjectRestService.java
teosarca: understand

backend/de.metas.business/src/main/java/de/metas/project/budget/BudgetProjectRepository.java
teosarca: i saw you marked this as resolved, but there is no change. the repo is still there.

backend/de.metas.business/src/main/java/de/metas/project/service/ProjectRepository.java
teosarca: what's the resolution here?

teosarca (changes requested):
check my comments (2)

backend/de.metas.business/src/main/java/de/metas/project/service/ProjectRepository.java
cp-ps: out of scope :-s ....

teosarca: true. 
but pls, let's keep in mind next time to not build more on top of ProjectData. 
We would need to refactor it sooner or later.

backend/de.metas.business/src/main/java/de/metas/project/budget/BudgetProjectRepository.java
dm-ps: It was not used, I just forgot to remove it, removed now, thx

PR #13441, opened 2022-08-16
Gh12358 customize doctype purchaseorder

MHNV:
- default doctype flags in windows fields
- default `IsAutoApplyValidationRule` = Y

https://github.com/metasfresh/ic114/issues/1487


cristinamghita:
The pull request shall be against intensive_care_hotfix.
And I do not see where A`D_Column.IsAutoApplyValidationRule=Y ` is addressed.

PR #13451, opened 2022-08-17
customize default doctype

MHNV:
customize default doctype

https://github.com/metasfresh/ic114/issues/1487


metas-ts:
some users might *not* want these doctypes two to be the default. 
they might e.g. want every order to be a prepay-order.

Issue #13461, opened 2022-08-18
Additional fields to Partner and contact


cristinamghita:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?

* Metasfresh new columns in Partner:
  - Is Section Partner
  - Is Section Group Partner
  - Section Group Partner
  - Mother Company
  - Is Sales Partner
  - Is Forwarder
  - Is Default Forwarder
  - Is Storage/Warehouse
  - Is Running Contracts
  - Actis Rating
  - RIM Category
   
* Metasfresh new fields in User:
  - Is Logistics Contact
  - Is Backoffice Contact

PR #13462, opened 2022-08-18
#13461 Additional fields to Partner and contact 

cristinamghita:
https://github.com/metasfresh/metasfresh/issues/13461


metas-ts (commented):
sorry to slow down this ithis issue but i'm really a bit worried about the new fields and terms that we introduce. i think they need descriptions. Like, what is a section group bypartner about? Or what is even a section?
Remember that this fields will end up in general metasfresh.
wdyt? I'm totally fine if we make sure to add those descriptions later..for example when corresponding code is added...if you can make sure with the issue-give that it won'T be forgotten in those future issues

Issue #13524, opened 2022-08-31
Add new Document Types `Call Order SO` and `Frame Order SO`


riadmen:
### What is the expected or desired behavior?

Add new Document Types `Call Order SO` and `Frame Order SO` based on standard `Sales Order`

PR #13525, opened 2022-08-31
Add new Document Types Call Order SO and Frame Order SO

riadmen:
https://github.com/metasfresh/metasfresh/issues/13524


metas-rc:
I am pretty sure you don't need a new "Call Order SO" doc type. It already exists on vanilla: 

image

metas-rc:
Please, make sure you set the correct docSubType in the Frame Order SO . Take Frame Order PO as example. 

image

riadmen:
> I am pretty sure you don't need a new "Call Order SO" doc type. It already exists on vanilla:
> 
> image

- Ok: Rename existing `Abrufauftrag` DT to `Call Order SO` just for ab182

PR #13534, opened 2022-09-01
BUG: I_C_InvoiceCandidate.QtyDelivered wrongly computed for manual invoice candidates

av-ps:
QtyToInvoice properly set on C_Invoice_Candidate

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/inout/IInOutDAO.java
+++ b/backend/de.metas.business/src/main/java/de/metas/inout/IInOutDAO.java
@@ -132,4 +132,6 @@ public interface IInOutDAO extends ISingletonService
 	<T extends I_M_InOut> Map<InOutId, T> getShipmentsByIds(Set<InOutId> inOutIds, Class<T> modelClass);
 
 	Collection<InOutAndLineId> retrieveLineIdsForOrderLineIdAvailableForInterimInvoice(OrderLineId orderLine);
+
+	<T extends I_M_InOutLine> T getLineByIdInTrx(@NonNull InOutLineId inoutLineId, Class<T> modelClass);
 }
--- a/backend/de.metas.business/src/main/java/de/metas/inout/IInOutDAO.java
+++ b/backend/de.metas.business/src/main/java/de/metas/inout/IInOutDAO.java
@@ -133,5 +133,5 @@ public interface IInOutDAO extends ISingletonService
 
 	Collection<InOutAndLineId> retrieveLineIdsForOrderLineIdAvailableForInterimInvoice(OrderLineId orderLine);
 
-	<T extends I_M_InOutLine> T getLineByIdInTrx(@NonNull InOutLineId inoutLineId, Class<T> modelClass);
+	<T extends I_M_InOutLine> T getLineByIdInTrx(@NonNull InOutLineId inoutLineId, @NonNull Class<T> modelClass);
 }
--- a/backend/de.metas.business/src/main/java/de/metas/inout/impl/InOutDAO.java
+++ b/backend/de.metas.business/src/main/java/de/metas/inout/impl/InOutDAO.java
@@ -476,4 +476,10 @@ public class InOutDAO implements IInOutDAO
 				.collect(ImmutableList.toImmutableList());
 
 	}
+
+	@Override
+	public <T extends I_M_InOutLine> T getLineByIdInTrx(@NonNull final InOutLineId inoutLineId, final Class<T> modelClass)
+	{
+		return load(inoutLineId.getRepoId(), modelClass);
+	}
 }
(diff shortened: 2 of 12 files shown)


backend/de.metas.business/src/main/java/de/metas/inout/impl/InOutDAO.java
```diff
@@ -476,4 +476,10 @@ public Collection<InOutAndLineId> retrieveLineIdsForOrderLineIdAvailableForInter
 				.collect(ImmutableList.toImmutableList());
 
 	}
+
+	@Override
+	public <T extends I_M_InOutLine> T getLineByIdInTrx(@NonNull final InOutLineId inoutLineId, final Class<T> modelClass)
```
metas-ts, line 481:
```suggestion
	public <T extends I_M_InOutLine> T getLineByIdInTrx(@NonNull final InOutLineId inoutLineId, @NonNull final Class<T> modelClass)
```

backend/de.metas.business/src/main/java/de/metas/inout/IInOutDAO.java
```diff
@@ -132,4 +132,6 @@ public interface IInOutDAO extends ISingletonService
 	<T extends I_M_InOut> Map<InOutId, T> getShipmentsByIds(Set<InOutId> inOutIds, Class<T> modelClass);
 
 	Collection<InOutAndLineId> retrieveLineIdsForOrderLineIdAvailableForInterimInvoice(OrderLineId orderLine);
+
+	<T extends I_M_InOutLine> T getLineByIdInTrx(@NonNull InOutLineId inoutLineId, Class<T> modelClass);
```
metas-ts, line 136:
```suggestion
	<T extends I_M_InOutLine> T getLineByIdInTrx(@NonNull InOutLineId inoutLineId, @NonNull Class<T> modelClass);
```

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandDAO.java
```diff
@@ -422,6 +422,18 @@ private boolean isInOutCompletedOrClosed(@NonNull final I_C_InvoiceCandidate_InO
...
 	}
 
+	@Override
+	public List<I_C_InvoiceCandidate_InOutLine> retrieveICIOLAssociationsFor(@NonNull final InvoiceCandidateId invoiceCandidateId)
+	{
+		return queryBL.createQueryBuilder(I_C_InvoiceCandidate_InOutLine.class)
+				.addEqualsFilter(I_C_InvoiceCandidate_InOutLine.COLUMN_C_Invoice_Candidate_ID, invoiceCandidateId)
+				.addOnlyActiveRecordsFilter()
+				.orderBy(I_C_InvoiceCandidate_InOutLine.COLUMN_M_InOutLine_ID)
+				.orderBy(I_C_InvoiceCandidate_InOutLine.COLUMN_C_Invoice_Candidate_ID)
```
metas-ts, line 432:
```suggestion
```

no need to order by those..because of the `.addEqualsFilter(I_C_InvoiceCandidate_InOutLine.COLUMN_C_Invoice_Candidate_ID, invoiceCandidateId)`

metas-ts (commented):
hi @av-ps ,
what makes me a bit nervous about this PR is i don't understand why it was not working before os what code-changes do the fix..
- i checked in this branch with git blame and imho the stuff that *should* cause this thing to behave correctly was already in place since 2019'ish times

..and so i also can't see how this PR - besides adding a previously missing cucumber-test ofc - fixes a problem.
I do see that instead of just calling `retrieveICIOLAssociationsExclRE` you now get them alland then later skip the ones that are not completed or closed...but AFAIU it shouldn't matter because there are no `icIolAssociationRecords` in either way..

so, could you point me please?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/internalbusinesslogic/ShipmentData.java
```diff
@@ -79,11 +70,7 @@ private ShipmentData(
 
 	public StockQtyAndUOMQty computeInvoicableQtyDelivered(@NonNull final InvoicableQtyBasedOn invoicableQtyBasedOn)
 	{
-		if(empty)
-		{
-			return StockQtyAndUOMQtys.createZero(productId, getQtyNominal().getUomId());
-		}
-		Quantity deliveredInUom;
+		final Quantity deliveredInUom;
 		switch (invoicableQtyBasedOn)
```
av-ps, line 74: hi @metas-ts , the changes cover 2 scenarios:
1. removing this line is what allows our scenario to fallback to default QtyDelivered/QtyToInvoice (meaning existing values from C_Invoice_Candidate record) => we want to fallback to default if no ICIOLs are available
2. loading all ICIOLs prevents falling back to default values if there are any iciols out there (basically covers the reversed & reopen scenarios)

PR #13540, opened 2022-09-02
Intensive care uat desdvpack 010922

cp-ps:
https://github.com/metasfresh/metasfresh/issues/13538


metas-ts (changes requested):
generally lgtm, but please check out my cucumber-ramblings

PR #13547, opened 2022-09-05
WO-Project-JSON in-Depth QA

fp-ps:


backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/project/workorder/WorkOrderProjectResourceRestService.java
```diff
@@ -504,11 +507,17 @@ private static JsonWorkOrderResourceResponse toJsonWorkOrderResourceResponse(
...
-
 		@NonNull
 		WOProjectStepId stepId;
 
 		@NonNull
 		IdentifierString resourceIdentifier;
 	}
+
+	@NonNull
+	private static Predicate<WOProjectResourceIdentifier> distinctByResourceIdentifier(@NonNull final Function<WOProjectResourceIdentifier, IdentifierString> keyExtractor)
```
metas-ts, line 518: it took me a few minutes to wrap my had around this...i guess it's legit though. wdyt, @teosarca

metas-ts: LE/sidenote: it took me another few minutes to believe that there is no `distinctBy(keyextrator)` method in  java.util..but it seems there isn't

metas-ts (approved):

PR #13552, opened 2022-09-05
Unify de.metas.project.workorder && de.metas.project.workorder.data

av-ps:


backend/de.metas.business/src/main/java/de/metas/project/workorder/project/WOProjectRepository.java
```diff
@@ -154,10 +156,98 @@ public List<WOProject> listBy(@NonNull final WOProjectQuery query)
...
+		return queryBuilder.create();
+	}
+
+	@NonNull
+	private IQuery<I_C_Project> computeProjectCalendarQuery(@NonNull final WOProjectCalendarQuery query)
+	{
+		final IQueryBuilder<I_C_Project> queryBuilder = queryBL
+				.createQueryBuilder(I_C_Project.class)
+				.addOnlyActiveRecordsFilter()
+				.addEqualsFilter(I_C_Project.COLUMNNAME_ProjectCategory, ProjectCategory.WorkOrderJob)
```
metas-ts, line 233: i checked org.adempiere.ad.dao.impl.CompareQueryFilter#toString and i don't think that is will "unbox" WorkOrderJob's code in a reliable manner, so AFAIU you need to use `ProjectCategory.WorkOrderJob.getCode()`, no?

PS: if you know it does, please augment the javadoc of de.metas.util.lang.ReferenceListAwareEnum accordingly

av-ps: I've just tested this method and the unboxing worked fine (also checked org.adempiere.ad.dao.impl.CompareQueryFilter#accept and when the values are compared you can see that  org.adempiere.ad.dao.impl.CompareQueryFilter#normalizeValue method covers also the `ReferenceListAwareEnum` case )

PR #13560, opened 2022-09-07
#13423 fix for material receipts with quality issues

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/13423

### DIFF
--- a/backend/de.metas.contracts/src/main/java/de/metas/contacts/invoice/interim/invoicecandidatehandler/FlatrateTermInterimInvoice_Handler.java
+++ b/backend/de.metas.contracts/src/main/java/de/metas/contacts/invoice/interim/invoicecandidatehandler/FlatrateTermInterimInvoice_Handler.java
@@ -198,7 +198,7 @@ public class FlatrateTermInterimInvoice_Handler implements ConditionTypeSpecific
 
 		final I_C_Invoice_Candidate invoiceCandidate = HandlerTools.createIcAndSetCommonFields(term);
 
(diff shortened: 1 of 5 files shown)


backend/de.metas.contracts/src/main/java/de/metas/contacts/invoice/interim/invoicecandidatehandler/FlatrateTermInterimInvoice_Handler.java
```diff
@@ -198,6 +198,12 @@ private I_C_Invoice_Candidate createBaseIC(final @NonNull I_C_Flatrate_Term term
 
 		final I_C_Invoice_Candidate invoiceCandidate = HandlerTools.createIcAndSetCommonFields(term);
 
+		// Currently relying on C_Flatrate_Conditions.InvoiceRule instead of setting it programmatically.
```
metas-ts, line 201: relying on C_Flatrate_Conditions.InvoiceRule sounds plausible, but *where* are we relyinging on it and in what fashion?
why don't we rely on it by settings it to the IC in here?

adi-stefan: The InvoiceRule is already set on the IC from the conditions in `de.metas.contracts.invoicecandidate.FlatrateTerm_Handler#createCandidateForTerm`. Will rephrase the comment to make that clearer.

metas-ts (approved):

PR #13573, opened 2022-09-08
Follow - Generic importer for invoice candidates

av-ps:
Follow - import invoice candidates


backend/de.metas.swat/de.metas.swat.base/src/main/java-gen/de/metas/invoicecandidate/model/I_I_Invoice_Candidate.java
```diff
@@ -297,6 +297,28 @@ public interface I_I_Invoice_Candidate
...
+	void setDefault_OrgCode (java.lang.String Default_OrgCode);
+
+	/**
+	 * Get Default Org Code.
+	 *
+	 * <br>Type: String
+	 * <br>Mandatory: true
+	 * <br>Virtual Column: false
+	 */
+	java.lang.String getDefault_OrgCode();
```
metas-ts, line 317: please add a nice field description which says that this column is supposed to be set from a constant in the import format and that it serves as a fallback for the case that the org-code is not set

metas-ts (changes requested):
LGTM, please

Issue #13578, opened 2022-09-09
When creating a counter doc C_Order, take into account the dropship information


adi-stefan:
### Is this a bug or feature request?
Bug
### What is the current behavior?
When creating a counter doc C_order, the dropship information is currently not taken into account

PR #13581, opened 2022-09-09
#13578 populate `M_InOut` with dropship information (if available) from `M_ShipmentSchedule`

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/13578


adi-stefan:
not needed.

PR #13588, opened 2022-09-13
Extend the tests to catch the problem and fix it with a cherry-pick

metas-ts:
https://github.com/metasfresh/metasfresh/issues/13282


metas-ts (commented):
lgtm

Issue #13595, opened 2022-09-13
Create  an attribute value handler for Business Partner 


Chemssou:
### Is this a bug or feature request?
FR
#### Which are the steps to reproduce?
/
### What is the expected or desired behavior?
Having an attribute value handlerfor business partner

PR #13596, opened 2022-09-13
gh13595 add business partner attribute value handler

Chemssou:
https://github.com/metasfresh/metasfresh/issues/13595


metas-ts (changes requested):
as written in MM; please document some more that it works (dev-test), or explain how it works, so that i can understand it..

PR #13612, opened 2022-09-15
Allow different ICs going into an invoice have different ExternalIds

metas-ts:


backend/de.metas.util/src/main/java/de/metas/util/lang/ExternalId.java
```diff
@@ -34,6 +33,11 @@
 @Value
 public class ExternalId implements Comparable<ExternalId>
 {
+	/**
+	 * Can be used to distinguish between an external ID that was not yet set and one that already set to {@code null}.
```
metas-ts, line 37:
```suggestion
	 * Can be used to distinguish between an external ID that was not yet set and one that was already set to "can't be used or set".
```

backend/de.metas.util/src/main/java/de/metas/util/lang/ExternalId.java
```diff
@@ -63,6 +67,18 @@ public static String toValue(@Nullable final ExternalId externalId)
 		return externalId.getValue();
 	}
 
+	/**
+	 * @return {@code true} if the given {@code externalId} is the same at {@link #INVALID}.
```
metas-ts, line 71:
```suggestion
	 * @return {@code true} if the given {@code externalId} is the same as {@link #INVALID}.
```

metas-ts (commented):
lgtm

metas-ts:
@MergifyIO copy middle_ages_uat

PR #13613, opened 2022-09-15
IC Handle the case of respective record going "into limbo"

av-ps:


metas-ts (approved):

metas-ts:
Hi @av-ps please also always add the metasfresh-URL to the PR description

PR #13618, opened 2022-09-16
ClientId fix + extended logging

cp-ps:
refs: https://github.com/metasfresh/metasfresh/issues/13198


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/impl/ADProcessLoggable.java
```diff
@@ -0,0 +1,140 @@
...
+import lombok.Builder;
+import lombok.NonNull;
+import org.slf4j.Logger;
+
+import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.locks.ReentrantLock;
+
+public class ADProcessLoggable implements ILoggable
```
metas-ts, line 40:
```suggestion
/**
   * This logabble can writes logs to {@code AD_PInstance_Log} just like a JavaProcess. It's supposed to be used when you have an {@code AD_PInstance_ID} and want to log to it, but you are not inside a java-process.
   */
public class ADProcessLoggable implements ILoggable
```

metas-ts (approved):

PR #13630, opened 2022-09-20
Investigate and fix flaky Checking the effect of override fields on invoice candidate

av-ps:
Work package progress


metas-ts (commented):
i can't claim that i understand it all, but looks plausible.
however, something else popped into my mind: maybe we need to use the `volatile` keeword? Also see https://www.baeldung.com/java-volatile

PR #13638, opened 2022-09-20
Set Clearingstatus on HU-Editor selection

av-ps:
Set clearance status process


metas-ts (approved):
..but please check out my remark about the language of the param values

Issue #13639, opened 2022-09-20
Section Code Propagation


riadmen:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?

The `Section Code` must be propagated in all documents: sales order, purchase order, invoice candidate, shipment sched, receipt sched, payment allocation, etc..

metas-ts:
(so far this is implemented partially)

PR #13640, opened 2022-09-20
Section Code Propagation

riadmen:
https://github.com/metasfresh/metasfresh/issues/13639


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/spi/impl/InOutProducerFromShipmentScheduleWithHU.java
```diff
@@ -327,7 +327,8 @@ private I_M_InOut createShipmentHeader(final ShipmentScheduleWithHU candidate, f
 		final I_M_InOut shipment = InterfaceWrapperHelper.newInstance(I_M_InOut.class, processorCtx);
 		shipment.setAD_Org_ID(shipmentSchedule.getAD_Org_ID());
 
-		//
+		shipment.setM_SectionCode_ID(candidate.getM_ShipmentSchedule().getM_SectionCode_ID());
```
metas-ts, line 330: At yesterday's scrum we talked about the possibility that a BPartner's section code might change.
That means that you can have in two shipment-schedules the same `C_BPartner_ID` but different `M_SectionCode_ID`s.
Those scheds need to go into different shipments, and therefore the `M_SectionCode_ID` needs to be made a part of the `HeaderAggregationKey`

metas-ts (approved):

PR #13643, opened 2022-09-21
Do not check for ESR accoutno if we deal with a QR account

cristinamghita:


backend/de.metas.payment.esr/src/main/java/de/metas/payment/api/impl/QRPaymentStringDataProvider.java
```diff
@@ -83,8 +83,9 @@ public I_C_BP_BankAccount createNewC_BP_BankAccount(final IContextAware contextP
 		bpBankAccount.setA_Name(bPartnerName);
 		bpBankAccount.setName(bPartnerName);
 
+		bpBankAccount.setQR_IBAN(paymentString.getIBAN());
 		bpBankAccount.setAccountNo(paymentString.getInnerAccountNo());
-		bpBankAccount.setESR_RenderedAccountNo(paymentString.getPostAccountNo());
+		bpBankAccount.setESR_RenderedAccountNo(paymentString.getPostAccountNo()); // we can not know it
```
metas-ts, line 88: not sure i get the comment. is it still current?

metas-ts (approved):

PR #13652, opened 2022-09-22
Allow invoking "Other" ExternalSystem via API

av-ps:
invoke "Other" ExternalSystem via API

### DIFF
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/externalsystem/invokeExternalSystem.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/externalsystem/invokeExternalSystem.feature
@@ -58,4 +58,5 @@ Feature: external system invocation using metasfresh api
       | JsonExternalSystemRequest.parameters.Key | JsonExternalSystemRequest.parameters.Value |
       | TestName1                                | TestValue1                                 |
       | TestName2                                | TestValue2                                 |
+      | External_Request                         | test                                       |
     And a new metasfresh AD_PInstance_Log is stored for the external system 'Other' invocation
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/externalsystem/invokeExternalSystem.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/externalsystem/invokeExternalSystem.feature
@@ -46,11 +46,16 @@ Feature: external system invocation using metasfresh api
   @from:cucumber
   Scenario: The request is good and the external Other is invoked via the correct process
     Given add Other external system config with identifier: otherConfigIdentifier
-      | Name | Value |
-      | Test | Test  |
+      | Name      | Value      |
+      | TestName1 | TestValue1 |
+      | TestName2 | TestValue2 |
 
     And store endpointPath /api/externalsystem/Other/:otherConfigIdentifier/test in context, resolving placeholder as ExternalSystem.value
 
     When a 'POST' request is sent to metasfresh REST-API with endpointPath from context and fulfills with '200' status code
 
-    Then a new metasfresh AD_PInstance_Log is stored for the external system 'Other' invocation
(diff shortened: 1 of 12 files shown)


backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/externalsystem/invokeExternalSystem.feature
```diff
@@ -41,3 +41,16 @@ Feature: external system invocation using metasfresh api
...
       | OrderNo   | 222         |
+
+
+  @from:cucumber
+  Scenario: The request is good and the external Other is invoked via the correct process
+    Given add Other external system config with identifier: otherConfigIdentifier
+      | Name | Value |
+      | Test | Test  |
+
+    And store endpointPath /api/externalsystem/Other/:otherConfigIdentifier/test in context, resolving placeholder as ExternalSystem.value
```
metas-ts, line 52: is the `:` correct? shouldn't it be just `/api/externalsystem/Other/otherConfigIdentifier/test` ?

metas-ts: i guess i got it wrongly..re-checking..

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/externalsystem/invokeExternalSystem.feature
```diff
@@ -41,3 +41,16 @@ Feature: external system invocation using metasfresh api
       | param.key | param.value |
       | OrderId   | 111         |
       | OrderNo   | 222         |
+
+
+  @from:cucumber
+  Scenario: The request is good and the external Other is invoked via the correct process
+    Given add Other external system config with identifier: otherConfigIdentifier
+      | Name | Value |
+      | Test | Test  |
```
metas-ts, line 50: as we know, one "other" can have any number of parameters.
please demonstracte that it also works with >1 of those

metas-ts (changes requested):

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/externalsystem/invokeExternalSystem.feature
```diff
@@ -41,3 +41,16 @@ Feature: external system invocation using metasfresh api
...
+
+    And store endpointPath /api/externalsystem/Other/:otherConfigIdentifier/test in context, resolving placeholder as ExternalSystem.value
+
+    When a 'POST' request is sent to metasfresh REST-API with endpointPath from context and fulfills with '200' status code
```
metas-ts, line 54: also, please send some parameters and verify that those params make it to the external-systems invocation - together this the ones from the config

metas-ts (changes requested):

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/externalsystem/invokeExternalSystem.feature
cp-ps: just to be sure we are on the same page here:
  - the parameters we can send to the invoke ES rest-endpoint (i.e. `de.metas.common.externalsystem.JsonInvokeExternalSystemParams` ) are about AD_Process_Para, i.e. parameters used during the java process which do not necessarily end up in `de.metas.common.externalsystem.JsonExternalSystemRequest#parameters` (i.e. parameters sent over to the external systems)... so what we could verify is:
    - `ExternalSystem_Other_ConfigParameter` records we add will end up in `de.metas.common.externalsystem.JsonExternalSystemRequest#parameters`
    - `de.metas.common.externalsystem.JsonInvokeExternalSystemParams` we send will end up in `AD_PInstance_Para`

metas-ts (approved):

PR #13665, opened 2022-09-26
Gh13154 add flo id product table

MHNV:
Add FLO ID to the product window

https://github.com/metasfresh/th173/issues/446


backend/de.metas.adempiere.adempiere/base/src/main/java-gen/org/compiere/model/I_M_Product.java
```diff
@@ -1824,6 +1823,29 @@ public interface I_M_Product
...
 
+	/**
+	 * Set FLO ID.
+	 * ID granted by FLOCERT for companies to put it on their products.
+	 *
+	 * <br>Type: String
+	 * <br>Mandatory: false
+	 * <br>Virtual Column: false
+	 */
+	void setth173_FLO_ID (@Nullable java.lang.String th173_FLO_ID);
```
metas-ts, line 1834: as talked => column should be named FLO_Identifier or sth similar (not `th173_`, not `_ID`)

metas-ts (changes requested):

metas-ts (approved):

PR #13676, opened 2022-09-27
Registering another listener within a listener's event handling code might be a development error and that other listener might not be fired.

cp-ps:


backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
```diff
@@ -1127,7 +1127,9 @@ private boolean load_C_Invoice_Candidate(@NonNull final Map<String, String> row)
 			invCandQueryBuilder.addEqualsFilter(COLUMNNAME_C_Invoice_Candidate_ID, invoiceCandidateInOutLine.getC_Invoice_Candidate_ID());
 		}
 
-		final String orderLineIdentifier = DataTableUtil.extractNullableStringForColumnName(row, I_C_Invoice_Candidate.COLUMNNAME_C_OrderLine_ID + "." + TABLECOLUMN_IDENTIFIER);
+		final String orderLineIdentifier = Optional.ofNullable(DataTableUtil.extractNullableStringForColumnName(row, I_C_Invoice_Candidate.COLUMNNAME_C_OrderLine_ID + "." + TABLECOLUMN_IDENTIFIER))
+				.orElseGet(() -> DataTableUtil.extractNullableStringForColumnName(row, "OPT." + I_C_Invoice_Candidate.COLUMNNAME_C_OrderLine_ID + "." + TABLECOLUMN_IDENTIFIER));
```
metas-ts, line 1131: i might have overseen them,, but to me it looks like there are no `OPT.C_OrderLine_ID.Identifier` columns.
however, maybe the should be..so why not make them `OPT.` in the feature files?
if it too much hassle, maybe do it just in master (=>follow-up)?

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/impl/TrxListenerManager.java
```diff
@@ -96,16 +80,22 @@ private void verifyEventTiming(@NonNull final RegisterListenerRequest listener)
 		final TrxEventTiming eventTimingOfListener = listener.getTiming();
 
 		final TrxEventTiming currentTiming = getCurrentTiming();
+
+		if (currentTiming == eventTimingOfListener && listener.isForceAfterNextCommit())
```
metas-ts, line 84: i don't really understand this boolean expression; maybe you can try making it clearer by extracking local vars or by adding a comment?

metas-ts (changes requested):

backend/de.metas.async/src/main/java/de/metas/async/model/validator/C_Queue_WorkPackage.java
```diff
@@ -21,7 +21,7 @@
 import org.slf4j.Logger;
 import org.springframework.stereotype.Component;
 
-import org.springframework.stereotype.Component;
+import java.time.Instant;
```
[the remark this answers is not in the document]
cp-ps: yep 👍

PR #13700, opened 2022-09-30
after merge fixes

fp-ps:


backend/de.metas.payment.sepa/base/src/main/java/de/metas/payment/sepa/sepamarshaller/impl/SEPAVendorCreditTransferMarshaler_Pain_001_001_03_CH_02.java
```diff
@@ -483,22 +483,6 @@ private CreditTransferTransactionInformation10CH createCreditTransferTransaction
 	{
 		final CreditTransferTransactionInformation10CH cdtTrfTxInf = objectFactory.createCreditTransferTransactionInformation10CH();
 
-		//
-		// Payment ID
-		// EndToEndId: A unique key generated by the system for each payment.
```
metas-ts, line 488: uhm - are we sure we want to drop this?
asking because we have this block in both intensive_care_uat and master..

fp-ps: the code is still there to handle `paymentID`, see https://github.com/metasfresh/metasfresh/blob/intensive_care_5.175_fix_30092022/backend/de.metas.payment.sepa/base/src/main/java/de/metas/payment/sepa/sepamarshaller/impl/SEPAVendorCreditTransferMarshaler_Pain_001_001_03_CH_02.java#:~:text=//-,%7B,cdtTrfTxInf.setPmtId(pmtId)%3B,-%7D. setting `paymentID` ended up being duplicate in `intensive_care_uat`. this fixes `SEPAVendorCreditTransferMarshaler_Pain_001_001_03_CH_02_Test` unit test, which is currently failing in both branches.

PR #13701, opened 2022-09-30
Propagate Section_ID from OLCand

fp-ps:


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/sectionCode/SectionCodeRepository.java
```diff
@@ -0,0 +1,52 @@
...
+
+	@NonNull
+	public Optional<SectionCodeId> getSectionCodeIdByValue(@NonNull final OrgId orgId, @NonNull final String value)
+	{
+		return queryBL.createQueryBuilder(I_M_SectionCode.class)
+				.addOnlyActiveRecordsFilter()
+				.addEqualsFilter(I_M_SectionCode.COLUMNNAME_AD_Org_ID, orgId)
+				.addEqualsFilter(I_M_SectionCode.COLUMNNAME_Value, value)
+				.create()
+				.listIds()
```
metas-ts, line 47: please assume that there can be one active and fail if that's not the case.
also, if there is no unique-constraint on (Value, AD_Org_ID), please add one

backend/de.metas.dunning/src/main/sql/postgresql/system/50-de.metas.dunning/5657330_sys_gh13648_C_Dunning_Candidate_M_SectionCode_ID_column.sql
```diff
@@ -0,0 +1,49 @@
...
+-- 2022-09-23T10:04:19.396Z
+INSERT INTO AD_Index_Column (AD_Client_ID,AD_Column_ID,AD_Index_Column_ID,AD_Index_Table_ID,AD_Org_ID,Created,CreatedBy,EntityType,IsActive,SeqNo,Updated,UpdatedBy) VALUES (0,583967,541274,540709,0,TO_TIMESTAMP('2022-09-23 13:04:19','YYYY-MM-DD HH24:MI:SS'),100,'D','Y',20,TO_TIMESTAMP('2022-09-23 13:04:19','YYYY-MM-DD HH24:MI:SS'),100)
+;
+
+-- 2022-09-23T10:04:21.984Z
+DROP INDEX IF EXISTS idx_m_sectioncode_unique_value
+;
+
+-- 2022-09-23T10:04:21.986Z
+CREATE UNIQUE INDEX IDX_M_SectionCode_unique_value ON M_SectionCode (Value,AD_Org_ID)
```
metas-ts, line 47: 👍

metas-ts (changes requested):
just a minor thing..

PR #13721, opened 2022-10-06
Micrometer Performance Monitoring

mrie04:
Add Micrometer Performance Monitoring
https://github.com/metasfresh/metasfresh/issues/13746


backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/PO.java
```diff
@@ -1951,6 +1965,22 @@ private boolean loadColumn(final int index)
 	 * @return true if loaded
 	 */
 	protected final boolean load(final HashMap<String, String> hmIn)
+	{
+		final PerformanceMonitoringService service = SpringContextHolder.instance.getBeanOr(
```
teosarca, line 1969: discussed with @metas-ts the issues from these lines. pls clarify with him tomorrow.

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/monitoringannotation/aspect/MonitorAspect.java
```diff
@@ -0,0 +1,135 @@
...
+
+		Method method = ((MethodSignature) pjp.getSignature()).getMethod();
+		Monitor monitorAnnotation = method.getAnnotation(Monitor.class);
+
+		if(monitorAnnotation.type() == PerformanceMonitoringService.Type.REST_CONTROLLER_WITH_WINDOW_ID)
+		{
+			HttpServletRequest request = ((ServletRequestAttributes) RequestContextHolder.currentRequestAttributes()).getRequest();
+			Map pathVariables = (Map) request.getAttribute(HandlerMapping.URI_TEMPLATE_VARIABLES_ATTRIBUTE);
+			String windowId = (String)pathVariables.get("windowId");
+			String window = (Services.get(IADWindowDAO.class).retrieveWindowName(AdWindowId.ofRepoId(Integer.parseInt(windowId)))).getDefaultValue();
```
teosarca, line 80: extract IADWindowDAO service as class field.
extact this code snippet to `String getWindowName(HttpServletRequest request)` method.

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/monitoringannotation/aspect/MonitorAspect.java
```diff
@@ -0,0 +1,135 @@
...
+		this.service = service;
+	}
+
+	@Value("${performance.monitoring.enable:false}")
+	private Boolean perfMonEnvVar;
+
+	@Around("execution(* *(..)) && @annotation(de.metas.monitoringannotation.annotation.Monitor)")
+	public Object monitorMethod(ProceedingJoinPoint pjp) throws Throwable
+	{
+		final PerformanceMonitoringService.Metadata metadata;
```
teosarca, line 64: pls check as soon as possible if the monitoring is off, so you can stop here and avoid spending processing resources.

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/monitoringannotation/aspect/MonitorAspect.java
```diff
@@ -0,0 +1,135 @@
...
+				{
+					return pjp.proceed();
+				}
+				catch (Exception e)
+				{
+					throw e;
+				}
+				catch (Throwable t)
+				{
+					throw new RuntimeException(t);
```
teosarca, line 120: why not `throw AdempiereException.wrapIfNeeded(t)`?
...or why catching non-Exception Throwables (i.e. Error) ? what's the reason?

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/monitoringannotation/aspect/MonitorAspect.java
```diff
@@ -0,0 +1,135 @@
...
+public class MonitorAspect
+{
+	private final MicrometerPerformanceMonitoringService service;
+
+	public MonitorAspect(@NonNull final MicrometerPerformanceMonitoringService service)
+	{
+		this.service = service;
+	}
+
+	@Value("${performance.monitoring.enable:false}")
```
teosarca, line 58: why not usings sysconfig? a sysconfig could be changed on the fly...

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/monitoringannotation/aspect/MonitorAspect.java
```diff
@@ -0,0 +1,135 @@
...
+	private Boolean perfMonEnvVar;
+
+	@Around("execution(* *(..)) && @annotation(de.metas.monitoringannotation.annotation.Monitor)")
+	public Object monitorMethod(ProceedingJoinPoint pjp) throws Throwable
+	{
+		final PerformanceMonitoringService.Metadata metadata;
+		final Callable callable = getCallableFromProceedingJoinPoint( pjp );
```
teosarca, line 65: use java5 generics, i.e. Callable<?> or something...

also, instead of 
> final Callable callable = getCallableFromProceedingJoinPoint( pjp );

why `pp::processed` is not enough?

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/monitoringannotation/aspect/MonitorAspect.java
```diff
@@ -0,0 +1,135 @@
...
+					.type(monitorAnnotation.type())
+					.action(method.getName())
+					.build();
+		}
+
+		return service.monitor(callable, metadata);
+
+	}
+
+	private Callable<Object> getCallableFromProceedingJoinPoint(final ProceedingJoinPoint pjp)
```
teosarca, line 102: i guess this method is not needed.
anyways, if is needed, pls make it static and rename it to something like 'wrapAsCallable` because basically that's what it does

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -42,58 +44,92 @@
...
 {
 	private final MeterRegistry meterRegistry;
-	private final Optional<APMPerformanceMonitoringService> apmPerformanceMonitoringService;
+	private final ThreadLocal<Integer> depth = ThreadLocal.withInitial(() -> 0);
+	private final ThreadLocal<String> initiator = ThreadLocal.withInitial(() -> "");
+	private final ThreadLocal<String> initiatorWindow = ThreadLocal.withInitial(() -> "");
+	private final ThreadLocal<ArrayList<String>> calledBy = ThreadLocal.withInitial(() -> new ArrayList<>());
+	private final ThreadLocal<Boolean> isInitiatorLabelActive = ThreadLocal.withInitial(() -> false);
+
+	@Value("${performance.monitoring.enable:false}")
```
teosarca, line 53: why not using sysconfig?

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -42,58 +44,92 @@
...
 	}
 
 	@Override
-	public <V> V monitorTransaction(
+	public <V> V monitor(
 			@NonNull final Callable<V> callable,
-			@NonNull final TransactionMetadata metadata)
+			@NonNull final PerformanceMonitoringService.Metadata metadata)
 	{
+		if(perfMonEnvVar = false)
```
teosarca, line 67: > perfMonEnvVar = false

that will be always false :)
you wanted to use "==" but ... better, pls use sysconfig

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -42,58 +44,92 @@
 public class MicrometerPerformanceMonitoringService implements PerformanceMonitoringService
 {
 	private final MeterRegistry meterRegistry;
-	private final Optional<APMPerformanceMonitoringService> apmPerformanceMonitoringService;
+	private final ThreadLocal<Integer> depth = ThreadLocal.withInitial(() -> 0);
+	private final ThreadLocal<String> initiator = ThreadLocal.withInitial(() -> "");
+	private final ThreadLocal<String> initiatorWindow = ThreadLocal.withInitial(() -> "");
+	private final ThreadLocal<ArrayList<String>> calledBy = ThreadLocal.withInitial(() -> new ArrayList<>());
+	private final ThreadLocal<Boolean> isInitiatorLabelActive = ThreadLocal.withInitial(() -> false);
```
teosarca, line 51: considering introducing an class which holds depth, initiator etc data.
The class shall have proper methods for adding, removing things which reveal ur intentions.
Here you could have only one ThreadLocal of that class.

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -132,4 +168,11 @@ private <V> V recordCallable(final Callable<V> callable, final ArrayList<Tag> ta
 			throw PerformanceMonitoringServiceUtil.asRTE(e);
 		}
 	}
+
+	private IAutoCloseable reduceDepth()
```
teosarca, line 172: pretty weird name....
usually when u have a method which returns a closeable the expectation is to do something in that method and return an autocloseable which reverses the effect.

Basically, to have the logic which does and the logic which undoes in the same place.

But in ur case, u refactored here only the undo logic... which makes the code harder to read and maintain.
In this case i suggest to undo logic in some finally block of caller.

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/PerformanceMonitoringService.java
```diff
@@ -58,54 +58,28 @@ default void monitorSpan(final Runnable runnable, final SpanMetadata metadata)
...
-	@Value
-	@Builder
-	class TransactionMetadata
-	{
-		@NonNull
-		String name;
+		String window;
 
 		@Singular
 		Map<String, String> labels;
```
teosarca, line 79: can we use ImmutableMap?

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/PerformanceMonitoringService.java
```diff
@@ -58,54 +58,28 @@ default void monitorSpan(final Runnable runnable, final SpanMetadata metadata)
...
 
-		@NonNull
-		Type type;
+		@Nullable
+		BiConsumer<String, String> distributedHeadersInjector;
 
 		@Singular
 		Map<String, String> distributedTransactionHeaders;
```
teosarca, line 85: can we use ImmutableMap?

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/PerformanceMonitoringService.java
```diff
@@ -58,54 +58,28 @@ default void monitorSpan(final Runnable runnable, final SpanMetadata metadata)
...
-				request);
-	}
-
-	@Value
-	@Builder
-	class TransactionMetadata
-	{
-		@NonNull
-		String name;
+		String window;
```
teosarca, line 76: is this the "windowInfo"?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandidateHandlerBL.java
```diff
@@ -302,11 +302,12 @@ private ImmutableList<I_C_Invoice_Candidate> createForModel(
 			final IInvoiceCandidateHandler invoiceCandiateHandler)
 	{
 		final PerformanceMonitoringService performanceMonitoringService = SpringContextHolder.instance.getBean(PerformanceMonitoringService.class);
-		final SpanMetadata request = SpanMetadata.builder()
-				.type("createMissingInvoiceCandidates")
+		final Metadata request = Metadata.builder()
```
teosarca, line 305: can we refactor this as a constant?

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/DefaultView.java
```diff
@@ -357,6 +360,24 @@ public ViewResult getPage(
 			final int firstRow,
 			final int pageLength,
 			final ViewRowsOrderBy orderBy)
+	{
+		final PerformanceMonitoringService service = SpringContextHolder.instance.getBeanOr(
```
teosarca, line 364: pls avoid SpringContextHolder.instance.getBean* as much as possible.
get PerformanceMonitoringService as construction time.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/DefaultView.java
```diff
@@ -357,6 +360,24 @@ public ViewResult getPage(
...
+				PerformanceMonitoringService.class,
+				NoopPerformanceMonitoringService.INSTANCE);
+		return service.monitor(
+				() -> getPage0(firstRow, pageLength, orderBy),
+				PerformanceMonitoringService.Metadata
```
teosarca, line 369: providing the `viewId` would be a valuable information.

teosarca: also, why do we have to measure here when we measure the ViewRestCOntroller.getViewData?

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/PO.java
teosarca: briefly:
* move to PerformanceMonitoringService service to POServicesFacade
* consider converting `PerformanceMonitoringService.Metadata.builder()...build()` to constant

teosarca (changes requested):
pls check my comments

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/monitoringannotation/aspect/MonitorAspect.java
mrie04: because of override i needed to add this catch

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/monitoringannotation/aspect/MonitorAspect.java
mrie04: i used it here because i need a callable at the end anyway (monitor Function in return)

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/PerformanceMonitoringService.java
mrie04: yes, window name + id (changed to windowNameAndId for better understanding)

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandidateHandlerBL.java
mrie04: for this metric type currently no metrics get created and it's no visualisation realized. So i removed it

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/DefaultView.java
mrie04: in the dashboard you can see how long the underlying function inside a HTTP-Request take. This functions get handled together. Because of this they all must have the same Labels. If i add the 'viewId' here it would be ignored, because the other functions monitored in this Meter don't have a 'viewId' label.

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/PO.java
```diff
@@ -1688,6 +1691,14 @@ protected final void load(final int ID, final String trxName)
 		}
 	}    // load
 
+	private static final PerformanceMonitoringService.Metadata loadMetadata =
+			PerformanceMonitoringService.Metadata
+					.builder()
+					.name("PO")
+					.type(PerformanceMonitoringService.Type.REST_API_PROCESSING)
```
teosarca, line 1698: not sure but i guess u would want to use some "CRUD" type instead of REST_API_PROCESSING

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/ModelValidationEngine.java
```diff
@@ -101,6 +101,10 @@
 public class ModelValidationEngine implements IModelValidationEngine
 {
 
+	private static final String PERF_MON_SYSCONFIG_NAME = "de.metas.monitoring.modelInterceptor.enable";
+	private static final boolean SYS_CONFIG_DEFAULT_VALUE = false;
+	private final ISysConfigBL SYS_CONFIG_BL = Services.get(ISysConfigBL.class);
```
teosarca, line 106: pls name it sysconfigBL instead of SYS_CONFIG_BL.
it's not a constant.

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/ModelValidationEngine.java
```diff
@@ -691,21 +695,29 @@ public void removeModelChange(String tableName, IModelInterceptor interceptor)
...
-						.label(PerformanceMonitoringService.LABEL_RECORD_ID, Integer.toString(po.get_ID()))
-						.build());
+		final boolean perfMonIsActive = SYS_CONFIG_BL.getBooleanValue(PERF_MON_SYSCONFIG_NAME, SYS_CONFIG_DEFAULT_VALUE);
+		if(!perfMonIsActive)
+		{
+			fireModelChange0(po, changeType);
+		}
+		else
+		{
+			final PerformanceMonitoringService performanceMonitoringService = SpringContextHolder.instance.getBeanOr(PerformanceMonitoringService.class, NoopPerformanceMonitoringService.INSTANCE);
```
teosarca, line 705: check if we can have performanceMonitoringService as a class field? (even if lazy)

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/ModelValidationEngine.java
```diff
@@ -691,21 +695,29 @@ public void removeModelChange(String tableName, IModelInterceptor interceptor)
...
-				Metadata
-						.builder()
-						.name("ModelValidationEngine")
-						.type(Type.MODEL_INTERCEPTOR)
-						.action("fireModelChange")
-						.label("changeType", changeTypeStr)
-						.label("tableName", tableName)
-						.label(PerformanceMonitoringService.LABEL_RECORD_ID, Integer.toString(po.get_ID()))
-						.build());
+		final boolean perfMonIsActive = SYS_CONFIG_BL.getBooleanValue(PERF_MON_SYSCONFIG_NAME, SYS_CONFIG_DEFAULT_VALUE);
```
teosarca, line 698: refactor this to a new private method called `isPerformanceMonitorEnabled` (or something like that).

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/util/DB.java
```diff
@@ -483,6 +490,17 @@ public CPreparedStatement prepareStatement(final String sql)
 	}    // prepareStatement
 
 	public CPreparedStatement prepareStatement(final String sql, @Nullable final String trxName)
+	{
+		final boolean isPerfMonActive = dbPerformanceMonitoringHelper.isPerfMonActive();
```
teosarca, line 494: minor: IMHO u can inline this

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/util/DBPerformanceMonitoringHelper.java
```diff
@@ -0,0 +1,105 @@
...
+			performanceMonitoringService = _performanceMonitoringService = SpringContextHolder.instance.getBeanOr(
+					PerformanceMonitoringService.class,
+					NoopPerformanceMonitoringService.INSTANCE);
+		}
+		return performanceMonitoringService;
+	}
+
+	public boolean isPerfMonActive()
+	{
+		if(!isGettingSysconfig.get())
```
teosarca, line 91: why do we need this "synchronization"?

mrie04: this part is about avoiding recursion asap (`sysConfigBL().getBooleanValue()` -> ... -> `DB.prepareStatement()` -> `isPerfMonactive()` -> `sysConfigBL().getBooleanValue()`)

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/monitoring/aspect/MonitorAspect.java
```diff
@@ -0,0 +1,148 @@
...
+		};
+
+		return callable;
+	}
+
+	private String getWindowNameAndId(){
+		final HttpServletRequest request = ((ServletRequestAttributes) RequestContextHolder.currentRequestAttributes()).getRequest();
+		final Map pathVariables = (Map) request.getAttribute(HandlerMapping.URI_TEMPLATE_VARIABLES_ATTRIBUTE);
+		final String windowId = (String)pathVariables.get("windowId");
+		final IADWindowDAO iadWindowDAO = Services.get(IADWindowDAO.class);
```
teosarca, line 136: if u cannot extract this service as a field, pls add a comment here shortly explaining why

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -120,6 +144,11 @@ private Optional<Tag> mkTagIfNotNull(@Nullable final String name, @Nullable fina
 		return Optional.of(Tag.of(name, value));
 	}
 
+	private boolean isAnyRestControllerType(final Metadata metadata)
+	{
+		return metadata.getType() == Type.REST_CONTROLLER || metadata.getType() == Type.REST_CONTROLLER_WITH_WINDOW_ID;
```
teosarca, line 149: this method can go into Type enum

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -42,63 +42,88 @@
 public class MicrometerPerformanceMonitoringService implements PerformanceMonitoringService
 {
 	private final MeterRegistry meterRegistry;
-	private final Optional<APMPerformanceMonitoringService> apmPerformanceMonitoringService;
+	private static final ThreadLocal<PerformanceMonitoringData> perfMonDataTL = ThreadLocal.withInitial(PerformanceMonitoringData::new);
```
teosarca, line 45: why u create a new instance as initial value?

are you aware that you will get a new instance everytime a thread is created in our system (and we have quite a lot) even if u need this performance monitoring or not?

suggestion: can't we go with null initial value and create the instance only when it's needed?

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -42,63 +42,88 @@
...
-		mkTagIfNotNull("Name", metadata.getName()).ifPresent(tags::add);
 
-		final Callable<V> callableToUse;
-		if (apmPerformanceMonitoringService.isPresent())
+		final PerformanceMonitoringData perfMonData = perfMonDataTL.get();
+		final ArrayList<Tag> tags = new ArrayList<>();
+
+		mkTagIfNotNull("name", metadata.getClassName()).ifPresent(tags::add);
+		mkTagIfNotNull("action", metadata.getFunctionName()).ifPresent(tags::add);
+
```
teosarca, line 64: don't understand the cases from the if/else  below.

some random thoughts:
* aren't REST_CONTROLLER_WITH_WINDOW_ID and REST_CONTROLLER quite similar
* instead of having whole mambo-jambo logics like 
```
metadata.getType() == Type.REST_CONTROLLER_WITH_WINDOW_ID
				&& metadata.getFunctionName() != null
				&& metadata.getWindowNameAndId() != null)
				````
couldn't we have a nice method into metadata which have a descriptive name?

File: backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java [hunk 28]
```diff
@@ -42,63 +42,88 @@
...
-		final Callable<V> callableToUse;
-		if (apmPerformanceMonitoringService.isPresent())
+		else if(perfMonData.getDepth() == 0)
 		{
-			callableToUse = () -> apmPerformanceMonitoringService.get().monitorSpan(callable, metadata);
+			perfMonData.setIsInitiatorLabelActive(false);
 		}
-		else
+
+		if(perfMonData.getIsInitiatorLabelActive())
```
teosarca, line 90: java naming conventions are saying that (primitive) boolean returning methods shall start with `is` and not `getIs`

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -42,63 +42,88 @@
...
 		{
-			callableToUse = callable;
+			mkTagIfNotNull("depth", String.valueOf(perfMonData.getDepth())).ifPresent(tags::add);
```
teosarca, line 92: i kind of don't like this construction at this level of application:

> mkTagIfNotNull("depth", String.valueOf(perfMonData.getDepth())).ifPresent(tags::add);

why not having a method called

addTagIfNotNull(ArrayList<Tag> tags, String name, String value)

which avoids creating unnecessary objects (like Optionals) 
and which simply does what it says.

NOTE: on other levels of application, using Optionals when needed is perfectly OK, but here, we must make sure our measuring overhead is MINIMAL.

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -42,63 +42,88 @@
...
+			mkTagIfNotNull("initiator", perfMonData.getInitiator()).ifPresent(tags::add);
+			mkTagIfNotNull("window", perfMonData.getInitiatorWindow()).ifPresent(tags::add);
+			mkTagIfNotNull("callerName", metadata.getClassName() + " - " + metadata.getFunctionName()).ifPresent(tags::add);
+			mkTagIfNotNull("calledBy", perfMonData.getCalledBy().get(perfMonData.getDepth())).ifPresent(tags::add);
+			perfMonData.getCalledBy().add(perfMonData.getDepth() + 1 ,metadata.getClassName() + " - " + metadata.getFunctionName());
+
+			final Type type;
+			if(isAnyRestControllerType(metadata))
```
teosarca, line 100: isAnyRestControllerType can be a method of metadata
... which can deletage it to Type.isAnyRestControllerType

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -42,63 +42,88 @@
...
+			{
+				type = metadata.getType();
+			}
+			else
+			{
+				type = Type.REST_API_PROCESSING;
+			}
+
+			perfMonData.setDepth(perfMonData.getDepth() + 1);
```
teosarca, line 109: why not introduce a method called incrementDepth() instead of setDepth.
same, decrementDepth() if needed

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -42,63 +42,88 @@
...
 		}
 
-		return recordCallable(callableToUse, tags, "mf." + metadata.getType());
+		createTagsFromLabels(tags, metadata.getLabels());
+		return recordCallable(callable, tags, "mf." + metadata.getType().getCode());
+
 	}
 
-	private ArrayList<Tag> createTagsFromLabels(@NonNull final Map<String, String> labels)
+	private void createTagsFromLabels(final ArrayList<Tag> tags, @NonNull final Map<String, String> labels)
```
teosarca, line 125: i would call it `appendLabelsToTags` because that's what it does

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -42,63 +42,88 @@
...
 
-		final Callable<V> callableToUse;
-		if (apmPerformanceMonitoringService.isPresent())
+		final PerformanceMonitoringData perfMonData = perfMonDataTL.get();
+		final ArrayList<Tag> tags = new ArrayList<>();
+
+		mkTagIfNotNull("name", metadata.getClassName()).ifPresent(tags::add);
+		mkTagIfNotNull("action", metadata.getFunctionName()).ifPresent(tags::add);
+
+		if(perfMonData.getDepth() == 0
```
teosarca, line 65: > perfMonData.getDepth() == 0

introduce a method in perfMonData class for that logic.
something like isFirstCall or isFirstDepth ... i don't like my suggestions... trying finding a better name.

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -42,63 +42,88 @@
...
-		else
+		else if(perfMonData.getDepth() == 0
+				&& metadata.getType() == Type.REST_CONTROLLER
+				&& metadata.getFunctionName() != null)
 		{
-			callableToUse = callable;
+			perfMonData.setIsInitiatorLabelActive(true);
+			perfMonData.setInitiator(metadata.getClassName() + " - " + metadata.getFunctionName());
+			perfMonData.setInitiatorWindow("NONE");
+			perfMonData.getCalledBy().add(0, "HTTP Request");
```
teosarca, line 83: > perfMonData.getCalledBy().add(0, "HTTP Request");

in perfMonData class,
introduce a method called something like `addCalledByAsFirst`.

Btw, why are u inserting it at first position?
If depth is ZERO isn't it supposed to be the first call?

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/PerformanceMonitoringService.java
```diff
@@ -58,57 +56,22 @@ default void monitorSpan(final Runnable runnable, final SpanMetadata metadata)
 	
 	@Value
 	@Builder
-	class SpanMetadata
+	class Metadata
```
teosarca, line 59: if this class grows (like adding more methods) consider extracting it to it's own file

nevertheless, if u are keeping it here, pls make sure it's `static`!

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/PerformanceMonitoringData.java
```diff
@@ -27,14 +20,20 @@
...
-		return SpanMetadata.builder()
-				.name("REST-API - " + restEndpointName)
-				.type(Type.REST_API_PROCESSING.getCode())
-				.build();
-	}
+	private int depth = 0;
+	private String initiator = "";
+	private String initiatorWindow = "";
+	private ArrayList<String> calledBy = new ArrayList<>();
+	private Boolean isInitiatorLabelActive = false;
```
teosarca, line 38: why Boolean object?

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/PerformanceMonitoringData.java
```diff
@@ -27,14 +20,20 @@
...
-public class ApiAPMHelper
+package de.metas.monitoring.adapter;
+
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.ArrayList;
+
+@Getter
+@Setter
```
teosarca, line 31: if u are adding the methods i've advised, 
it might me that u would need some `Getter/Setter` annotation on couple of fields only,
most of the fields won't need it.

PS: anyways in case u need getter and setter annotation of class level, it might be that you want to use `@Data` annotation.

backend/de.metas.async/src/main/java/de/metas/async/processor/impl/WorkpackageProcessorTask.java
```diff
@@ -165,18 +170,26 @@ private Properties createProcessingCtx()
...
-						.label("de.metas.async.queueProcessor.name", queueProcessor.getName())
-						.label(PerformanceMonitoringService.LABEL_WORKPACKAGE_ID, Integer.toString(workPackage.getC_Queue_WorkPackage_ID()))
-						.build());
+		final boolean perfMonIsActive = SYS_CONFIG_BL.getBooleanValue(PERF_MON_SYSCONFIG_NAME, SYS_CONFIG_DEFAULT_VALUE);
+		if(!perfMonIsActive){
+			run0();
+		}
+		else
+		{
+			final PerformanceMonitoringService service = SpringContextHolder.instance.getBeanOr(
```
teosarca, line 179: can't you get `final PerformanceMonitoringService service` as ctor parameter?

backend/de.metas.async/src/main/java/de/metas/async/processor/impl/WorkpackageProcessorTask.java
```diff
@@ -135,6 +136,10 @@ class WorkpackageProcessorTask implements Runnable
 	// task 09933 just adding this member for now, because it's unclear if in future we want to or have to extend on it or not.
 	private final boolean retryOnDeadLock = true;
 
+	private final ISysConfigBL SYS_CONFIG_BL = Services.get(ISysConfigBL.class);
```
teosarca, line 139: why not `sysConfigBL`?

backend/de.metas.adempiere.adempiere/serverRoot/de.metas.adempiere.adempiere.serverRoot.base/src/main/java-legacy/org/compiere/server/Scheduler.java
```diff
@@ -143,6 +143,10 @@ public Scheduler(final MScheduler model)
 	private it.sauronsoftware.cron4j.Scheduler cronScheduler;
 	private Predictor predictor;
 
+	private static final String PERF_MON_SYSCONFIG_NAME = "de.metas.monitoring.scheduler.enable";
+	private static final boolean SYS_CONFIG_DEFAULT_VALUE = false;
+	private final ISysConfigBL SYS_CONFIG_BL = Services.get(ISysConfigBL.class);
```
teosarca, line 148: why not sysConfigBL?

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/monitoring/aspect/MonitorAspect.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.concurrent.Callable;
+
+@Aspect
+@Component
+public class MonitorAspect
+{
+	private final MicrometerPerformanceMonitoringService service;
+	private static final String PERF_MON_SYSCONFIG_NAME = "de.metas.monitoring.annotation.enable";
+	private static final boolean SYS_CONFIG_DEFAULT_VALUE = false;
+	private final ISysConfigBL SYS_CONFIG_BL = Services.get(ISysConfigBL.class);
```
teosarca, line 55: sysConfigBL

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/util/DBPerformanceMonitoringHelper.java
```diff
@@ -0,0 +1,121 @@
...
+import de.metas.monitoring.adapter.NoopPerformanceMonitoringService;
+import de.metas.monitoring.adapter.PerformanceMonitoringService;
+import de.metas.util.Services;
+import lombok.NonNull;
+import org.adempiere.service.ISysConfigBL;
+import org.compiere.SpringContextHolder;
+
+import java.util.concurrent.Callable;
+
+public class DBPerformanceMonitoringHelper
```
teosarca, line 34: we already have org.adempiere.ad.dao.impl.QueryStatisticsLogger which is measuring performances....
what we would have to do is to improve it and make it work with micrometer.

The big advantage of QueryStatisticsLogger is that
* it's quite performant
* already works
* it's converting all executed SQLs..... which this code does not, e.g. not sure if ResultSet.executeQuery() is converted...

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/POServicesFacade.java
```diff
@@ -34,6 +38,25 @@
 	private ITrxManager _trxManager;
 	private ADReferenceService _adReferenceService;
 
+	private static PerformanceMonitoringService _performanceMonitoringService;
+	private static final String PM_METADATA_CLASS_NAME = "PO";
+	private static final String PM_METADATA_SAVE_EX_ACTION = "saveEx";
+	private static final String PM_METADATA_LOAD_ACTION = "load";
```
teosarca, line 44: pls inline those 3 constants.... it's quite pointless to have them as constants.
PM_METADATA_CLASS_NAME = "PO";
private static final String PM_METADATA_SAVE_EX_ACTION = "saveEx";
private static final String PM_METADATA_LOAD_ACTION

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/PO.java
```diff
@@ -265,6 +265,12 @@ public PO(final Properties ctx, final PO source, final int AD_Client_ID, final i
 		setAD_Org_ID(AD_Org_ID);
 	}    // PO
 
+	/**
+	 * Performance Monitoring sysconfig
+	 */
+	private static final String PERF_MON_SYSCONFIG_NAME = "de.metas.monitoring.po.enable";
+	private static final boolean SYS_CONFIG_DEFAULT_VALUE = false;
```
teosarca, line 272: why do we have these constants here (and also the `services.getSysConfigBooleanValue(PERF_MON_SYSCONFIG_NAME, SYS_CONFIG_DEFAULT_VALUE)` checking)

but the actual monitoring methods are in POServicesFacade.

Consider:
* either having all perf related things here
* either having all perf related things in POServicesFacade

but pls don't spread them.

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
mrie04: it's a generated getter method, i'll adjust boolean name to 'initiatorLabelIsActive'

teosarca: @mrie04  first, pls check why u need to use Boolean and not primitive boolean.

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
mrie04: my first thought is to replace it with `!isCalledByMonitoredFunction()`

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
mrie04: `calledBy` label for depth=0 is needed for the filter inside the dashboard. So every function can find the label-value at it's own depth and not depth-1 and sets the value at depth+1.

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/PerformanceMonitoringService.java
mrie04: modifier `static` is redundant for inner classes of interfaces

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -55,49 +55,37 @@ public <V> V monitor(
 			@NonNull final Callable<V> callable,
 			@NonNull final PerformanceMonitoringService.Metadata metadata)
 	{
+		if(perfMonDataTL == null)
+		{
+			perfMonDataTL = ThreadLocal.withInitial(PerformanceMonitoringData::new);
+		}
```
teosarca, line 61: didn't get this one.

my suggesiton was to use `ThreadLocal perfMonDataTL = new ThreadLocal()`

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -131,22 +120,26 @@ private void createTagsFromLabels(final ArrayList<Tag> tags, @NonNull final Map<
 				// Avoid OOME: if we included e.g. the recordId, then every recordId would cause a new meter to be created.
 				continue;
 			}
-			mkTagIfNotNull(entry.getKey(), entry.getValue()).ifPresent(tags::add);
+			addTagIfNotNull(entry.getKey(), entry.getValue(), tags);
 		}
 	}
 	
-	private Optional<Tag> mkTagIfNotNull(@Nullable final String name, @Nullable final String value)
+	private void addTagIfNotNull(@Nullable final String name, @Nullable final String value, @NonNull ArrayList<Tag> tags)
```
teosarca, line 127: static or inline, as u feel

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -131,22 +120,26 @@ private void createTagsFromLabels(final ArrayList<Tag> tags, @NonNull final Map<
...
-		return Optional.of(Tag.of(name, value));
 	}
 
-	private boolean isAnyRestControllerType(final Metadata metadata)
+	private void incrementDepth(PerformanceMonitoringData perfMonData)
+	{
+		perfMonData.setDepth(perfMonData.getDepth() + 1);
+	}
+
+	private void decrementDepth(PerformanceMonitoringData perfMonData)
```
teosarca, line 140: move decrementDepth to PerformanceMonitoringData class

teosarca: also avoid exposing PerformanceMonitoringData.setDepth method

teosarca: also, as a general rule, avoid exposing setters as much as possible.

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/PerformanceMonitoringData.java
```diff
@@ -28,12 +28,36 @@
...
 
 @Getter
-@Setter
 public class PerformanceMonitoringData
 {
+	@Setter
 	private int depth = 0;
 	private String initiator = "";
 	private String initiatorWindow = "";
+	@Setter
```
teosarca, line 37: why do u need this setter?

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -36,13 +36,15 @@
...
 import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
 
 @Service
 @Primary
 public class MicrometerPerformanceMonitoringService implements PerformanceMonitoringService
 {
 	private final MeterRegistry meterRegistry;
-	private static ThreadLocal<PerformanceMonitoringData> perfMonDataTL = null;
+	private static ThreadLocal<PerformanceMonitoringData> perfMonDataTL = new ThreadLocal<>();
```
teosarca, line 46: final

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/util/DB.java
```diff
@@ -539,6 +528,21 @@ public CPreparedStatement prepareStatement(final String sql,
 											   final int resultSetType,
 											   final int resultSetConcurrency,
 											   final String trxName)
+	{
+		{
+			if (dbPerformanceMonitoringHelper.isPerfMonActive())
```
teosarca, line 533: do we still need this since we are measuring at QueryStatisticsLogger?

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/PerformanceMonitoringService.java
```diff
@@ -53,6 +54,11 @@ default void monitor(final Runnable runnable, final Metadata metadata)
 				},
 				metadata);
 	}
+
+	default void monitor(final long duration, TimeUnit unit, final Metadata metadata)
```
teosarca, line 58: why not call it something like `record` or `recordTimeElapsed`?

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/PerformanceMonitoringData.java
```diff
@@ -60,4 +58,12 @@ public boolean isCalledByMonitoredFunction()
 	{
 		return depth != 0;
 	}
+
+	public void incrementDepth(){
```
teosarca, line 62: pls check ur intelliJ code formatting

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/util/DBPerformanceMonitoringHelper.java
mrie04: DB.executeUpdate() includes some ResultSet.executeQuery(), but in general it isn't monitored. After QuerryStatisticLogger is added DB.executeUpdate monitoring i think can be removed. 
DB.prepareStatement() isn't included in QuerryStatisticLogger and we could keep it and the basic Framework for more DB.java monitoring, if needed later

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/util/DB.java
mrie04: https://github.com/metasfresh/metasfresh/pull/13721#discussion_r1016312726

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/util/DBPerformanceMonitoringHelper.java
teosarca: disagree... let's talk

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/MicrometerPerformanceMonitoringService.java
```diff
@@ -124,22 +143,38 @@ private void appendLabelsToTags(@NonNull final ArrayList<Tag> tags, @NonNull fin
...
-		perfMonData.setDepth(perfMonData.getDepth() + 1);
+		addTagIfNotNull("depth", String.valueOf(perfMonData.getDepth()), tags);
+		addTagIfNotNull("initiator", perfMonData.getInitiator(), tags);
+		addTagIfNotNull("window", perfMonData.getInitiatorWindow(), tags);
+		addTagIfNotNull("callerName", metadata.getClassName() + " - " + metadata.getFunctionName(), tags);
+		addTagIfNotNull("calledBy", perfMonData.getCalledBy().get(perfMonData.getDepth()), tags);
 	}
 
-	private void decrementDepth(PerformanceMonitoringData perfMonData)
+	private Type getTypeForMeterName(Metadata metadata)
```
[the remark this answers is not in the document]
mrie04: in general the type is used for the meter-name (meter = set of metrics),
but all functions called by an RestController-function are grouped in Type.REST_API_PROCESSING for dashboard use (so for example type.db gets replaced).
So this function replaces the type or keeps the restController type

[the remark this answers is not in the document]
teosarca: hmmm.... this is smelly. 

why not group everything under the first caller? I mean, the first caller can tell if is a REST API or not.

For me this method does not make sense.
1. the name is wrong... has nothing to do with what it does
2. the logic is "all about REST"  (i.e. if the type is something related to REST then OK keep it if not, i don't care, i consider is about REST).... and that's wrong.

You shall know if is a rest or NOT if the initiator is REST or not. 

am I missing something?

cc @metas-ts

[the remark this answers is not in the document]
mrie04: >why not group everything under the first caller?
this would make the use of `Prometheus Query Language` in the Grafana Dashboard much more complicated and they would take longer. RestController metrics and underlaying functions metrics are handled separately.

>1. the name is wrong... has nothing to do with what it does
- the meter name is set like METER_PREFIX + type.getCode(), which results in mf_rest_controller...
- without this function every Type would get it's own name (like mf_PO, mf_DB, mf_modelInterceptor...)
- so this function gives me the type that should be used for the name
  - RestController functions keep their name
  - underlying functions get mf_rest_api_processing as name
  - functions not called by a rest controller keep their name mf_db, mf_po (else clause: `!perfMonData.isInitiatorLabelActive()`)

backend/de.metas.monitoring/src/main/java/de/metas/monitoring/adapter/PerformanceMonitoringData.java
```diff
@@ -37,11 +37,18 @@
...
 	public String getInitiatorFunctionNameFQ()
 	{
-		return !calledBy.isEmpty() ? calledBy.get(0).getFunctionNameFQ() : "";
+		if (calledBy.isEmpty())
+		{
+			return "";
+		}
+		else
+		{
+			return calledBy.size() > 1 ? calledBy.get(1).getFunctionNameFQ() : calledBy.get(0).getFunctionNameFQ();
```
teosarca, line 46: > calledBy.size() > 1

this does not look good to me. It's a hidden feature which could be broken anytime in the future.

I suggest to:
* introduce a flag in PerformanceMonitoringService.Metadata, e.g. `isGrouppingPlaceholder`. A flag which you will set when you will create the dummy "HTTP request" grouping metadata.

* in this method i think u could take the first one which has isGrouppingPlaceholder=false

PR #13762, opened 2022-10-17
fix qr sepa export

cristinamghita:


metas-ts (changes requested):
the code looks plausible, but what's missing imho is docu. 
particularly i think the descriptions of the two IBAN fields need to indicate the coalesce that you have at one place. maybe it also helps for me to get a glimpse on the UAT-docu that you are going to send to the user

metas-ts (commented):
going to integrate now; please supplement the docu asap

Issue #13772, opened 2022-10-18
Set "-" in MAC Address instead of null


metas-rc:
### Is this a bug or feature request?
* FR
### What is the current behavior?
* If n<6 MAC Addresses are required for a BOM, the attribute values are not generated for the ones that are not.
#### Which are the steps to reproduce?
* Have a BOM with 4 MAC Addresses
* Create and complete a manufacturing order for it
* Run the process to Issue/Receipt
* Select the MAC Address line and run the process  ``Issue Services``
* => Only 4 MAC addresses are generated, the last 3 being left null
### What is the expected or desired behavior?
* Instead of null, set the value of the unneeded MAC Addresses to "-".

PR #13773, opened 2022-10-18
MAC Address: set "-" instead of leaving empty

metas-rc:
https://github.com/metasfresh/metasfresh/issues/13772

### DIFF
--- a/backend/de.metas.manufacturing/src/main/java/de/metas/manufacturing/generatedcomponents/MacAddressGenerator.java
+++ b/backend/de.metas.manufacturing/src/main/java/de/metas/manufacturing/generatedcomponents/MacAddressGenerator.java
@@ -113,7 +113,8 @@ public class MacAddressGenerator implements IComponentGenerator
 		{
 			return ImmutableAttributeSet.EMPTY;
 		}
-		if (countRemainingToGenerate > attributesAvailableToGenerate.size())
+		final int availableSize = attributesAvailableToGenerate.size();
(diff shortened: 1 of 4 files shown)


backend/de.metas.manufacturing/src/main/java/de/metas/manufacturing/generatedcomponents/MacAddressGenerator.java
```diff
@@ -111,7 +113,8 @@ public ImmutableAttributeSet generate(@NonNull final ComponentGeneratorContext c
 		{
 			return ImmutableAttributeSet.EMPTY;
 		}
-		if (countRemainingToGenerate > attributesAvailableToGenerate.size())
+		final int availableSize = attributesAvailableToGenerate.size();
```
metas-ts, line 116: hi, it generally looks OK to me, i'm just a bit concerned about the understandability..so wdyt about
- does it make sense to create the `-`  Mac-Addresses only for mandatory attributes?
- add a javadoc to this method that describes that missing Mac-Addresses with be `-`ed

metas-rc: * The value "-" is set for the MAC Attributes even if they are not mandatory for manufacturing.
* Column "Description"added to PP_ComponentGenerator and displayed in the window

metas-ts (approved):

PR #13782, opened 2022-10-20
Look up C_OLcand.M_HU_PI_Item_Product_ID also via StoreGLN

av-ps:


backend/de.metas.edi/src/main/sql/postgresql/ddl/views/EDI_C_BPartner_Lookup_BPL_GLN_v.sql
```diff
@@ -1,46 +1,44 @@
...
-;
+    (
+        SELECT
+            bp.C_BPartner_ID,
+            bp.IsActive,
+            --
+            -- Note: The GLN is unique per BPL.
+            -- We're just filtering by two locations here and will add them to the whereclause in the EXP_Format_Line filter.
+            --
+            TRIM(BOTH ' ' FROM l_main.GLN) AS GLN, -- The Selector's GLN
```
metas-ts, line 15: please do it like we do in edi_cctop_invoic_500_v_view.sql.
because that also deals with e.g. tabs

PR #13785, opened 2022-10-20
Import material (product) data from SAP

mi-ps:
refs: https://github.com/metasfresh/metasfresh/issues/13745


misc/services/camel/de-metas-camel-externalsystems/de-metas-camel-sap/src/main/java/de/metas/camel/externalsystems/sap/api/model/product/ProductRow.java
```diff
@@ -0,0 +1,53 @@
...
+	@DataField(pos = 1)
+	private String materialCode;
+
+	@DataField(pos = 3)
+	private String sectionCode;
+
+	@DataField(pos = 4)
+	private String name;
+
+	//FIXME: pending mappings
```
metas-ts, line 42: is this still current? seems to me that you do use the value

metas-ts:
also, please add the metasfresh-Issue-URL to this PR's description

misc/services/camel/de-metas-camel-externalsystems/de-metas-camel-sap/src/main/java/de/metas/camel/externalsystems/sap/api/model/product/ProductRow.java
cp-ps: yep, this is still a thing as according to specs the UOM is in some custom format => so we might need custom mappings

metas-ts (changes requested):
lgtm so far, but please check my remarks regarding Proposed TRLs

Issue #13812, opened 2022-10-24
HU Clearance status: Test Pending


metas-rc:
### Is this a bug or feature request?
* FR
### What is the current behavior?
* 
#### Which are the steps to reproduce?
*
### What is the expected or desired behavior?
* New status "test pending" (DE: "zur Prüfung")
* HUs with that status are not allowed to be used for mobile UI Manufacturing and picking

PR #13815, opened 2022-10-24
HU Clearance: Status Test Pending

metas-rc:
https://github.com/metasfresh/metasfresh/issues/13812

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/impl/HandlingUnitsBL.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/impl/HandlingUnitsBL.java
@@ -97,6 +97,7 @@ import org.compiere.model.I_M_Attribute;
 import org.compiere.model.I_M_AttributeSetInstance;
 import org.compiere.model.I_M_Product;
 import org.compiere.model.I_M_Transaction;
+import org.compiere.util.TimeUtil;
 import org.slf4j.Logger;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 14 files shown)


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/impl/HandlingUnitsBL.java
```diff
@@ -1118,6 +1119,7 @@ public void setClearanceStatusRecursively(
 
 		hu.setClearanceStatus(clearanceStatusInfo.getClearanceStatus().getCode());
 		hu.setClearanceNote(clearanceStatusInfo.getClearanceNote());
+		hu.setClearanceDate(TimeUtil.asTimestamp(clearanceStatusInfo.getClearanceDate()));
```
metas-ts, line 1122: i think in such a case it's very risky to not take the timezone (org into account).
Please either make sure to have the OrgId and use OrgDAO.getTimezone,
Or probably even better, use InstantAndOrgId

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/pporder/api/impl/AbstractPPOrderReceiptHUProducer.java
```diff
@@ -384,7 +385,12 @@ private IAllocationRequest createAllocationRequest(final IHUContext huContext, f
 		if (clearanceStatus != null)
 		{
 			final String language = getOrgUserOrLoggedInUSerLanguage(referencedModel);
-			clearanceStatusInfo = ClearanceStatusInfo.of(clearanceStatus, msgBL.getMsg(language, MESSAGE_ClearanceStatusInfo_Manufactured));
+			clearanceStatusInfo = ClearanceStatusInfo.builder()
+					.clearanceStatus(clearanceStatus)
+					.clearanceNote(msgBL.getMsg(language, MESSAGE_ClearanceStatusInfo_Manufactured))
+					.clearanceDate(TimeUtil.asInstant(date))
+					.build();
```
metas-ts, line 392: add the org-id (or the new InstanceAndOrg-thingie?)
so that we don't end up in the wrong timezone

metas-ts (changes requested):

metas-ts (approved):

PR #13820, opened 2022-10-25
Add Soehnle (model 3025/3035) scale integration

adi-stefan:


backend/de.metas.device.scales/src/main/java/de/metas/device/scales/impl/soehenle/ISoehenleCmd.java
```diff
@@ -0,0 +1,47 @@
...
+
+package de.metas.device.scales.impl.soehenle;
+
+import de.metas.device.scales.impl.ICmd;
+
+import java.util.LinkedHashMap;
+
+public interface ISoehenleCmd extends ICmd
+{
+	String CMD_TERMINATOR = "\r\n"; // we need to terminal the command string with "Carriage Return, Line Feed"
```
metas-ts, line 31:
```suggestion
	String CMD_TERMINATOR = "\r\n"; // we need to terminate the command string with "Carriage Return, Line Feed"
```

metas-ts (changes requested):
- please add some using tests for the parser code
- please add some (inactive?) sysconfigs so we know the soehnle-support is there...i think we have something similar for the existing two scale-types

metas-ts (approved):

Issue #13824, opened 2022-10-25
Fix `BOM Recursive Report`


riadmen:
### Is this a bug or feature request?
BUG
### What is the current behavior?

The Excel export ignore the latest BOM versions

### What is the expected or desired behavior?

The excel export process should consider the latest BOM version

PR #13825, opened 2022-10-25
The excel export process should consider the latest BOM version

riadmen:
https://github.com/metasfresh/metasfresh/issues/13824

### DIFF
--- a/backend/de.metas.manufacturing/src/main/sql/postgresql/ddl/functions/PP_Product_BOM_Recursive.sql
+++ b/backend/de.metas.manufacturing/src/main/sql/postgresql/ddl/functions/PP_Product_BOM_Recursive.sql
@@ -61,7 +61,7 @@ $BODY$
 						where bom.M_Product_ID=bomLineProduct.M_Product_ID
 						and bom.IsActive='Y'
 						and bom.Value=bomLineProduct.Value
-						order by bom.PP_Product_BOM_ID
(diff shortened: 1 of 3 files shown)


backend/de.metas.manufacturing/src/main/sql/postgresql/ddl/functions/PP_Product_BOM_Recursive.sql
```diff
@@ -61,7 +61,7 @@ $BODY$
 						where bom.M_Product_ID=bomLineProduct.M_Product_ID
 						and bom.IsActive='Y'
 						and bom.Value=bomLineProduct.Value
-						order by bom.PP_Product_BOM_ID
+                        ORDER BY bom.validfrom DESC
```
metas-rc, line 64: Please, also order by the old PP_Product_BOM_ID (or other unique thing)  after ordering by validFrom so we cover the cases where 2 BOMs are valid from the same date.

metas-rc (changes requested):
Please, check my small comment

backend/de.metas.manufacturing/src/main/sql/postgresql/ddl/functions/PP_Product_BOM_Recursive.sql
riadmen: Modification done

PR #13838, opened 2022-10-27
The excel export process should consider the latest BOM version and exclude the oldest version with validto date

riadmen:
https://github.com/metasfresh/metasfresh/issues/13824


adi-stefan (approved):
lgtm

Issue #13843, opened 2022-10-28
Use Sequence from PP Order Dispo to create PP Orders


riadmen:
### Is this a bug or feature request?
FR

### What is the expected or desired behavior?

The PP Order document No shall be in line with the sequence because the mobile UI is sorting by PP Order Document No ascending

PR #13847, opened 2022-10-31
Gh13843 pp order seq no

riadmen:


backend/de.metas.manufacturing/src/main/java/org/eevolution/api/PPOrderCreateRequest.java
```diff
@@ -107,7 +109,8 @@
 			@Nullable final ProjectId projectId,
 			//
 			@Nullable final Boolean completeDocument,
-			@Nullable final HUPIItemProductId packingMaterialId)
+			@Nullable final HUPIItemProductId packingMaterialId,
+			final int seqNo)
```
metas-ts, line 113: it's OK for me to call the new column `OrdercandidateSeqNo` (as opposed to renaming the exiging one to `ProductPlanningSeqNo`), but then please also call the fields and params `orderCandidateSeqNo` as well, to reduce FUD when looking at both SeqNos

backend/de.metas.manufacturing/src/main/java/org/eevolution/productioncandidate/service/CreateOrderCandidateCommand.java
```diff
@@ -118,6 +122,17 @@ public I_PP_Order_Candidate execute()
 		return ppOrderCandidateRecord;
 	}
 
+	private int calculatePPOrderCandidateSeqNo(final I_PP_Order_Candidate ppOrderCandidateRecord)
+	{
+		final int lastSeqNo = Services.get(IQueryBL.class)
+				.createQueryBuilder(I_PP_Order_Candidate.class, ppOrderCandidateRecord)
+				.addEqualsFilter(I_PP_Order_Candidate.COLUMNNAME_M_Product_ID, ppOrderCandidateRecord.getM_Product_ID())
```
metas-ts, line 129: i can't see where the requirement regarding "same product" is coming from. to me it seems useless at best and might be detrimental if the user wants to run different products through the same machine (which many users don't but some might)
related idea: why don't we 
- make `OrderCandidateSeqNo` an SQL-column with the value being `PP_Order_Candidate_ID * 10`
- and sort the processable candidates by `PP_Order_Candidate_ID` 

please **clarify this question** and **write a functional concept** before requesting further code-reviews from me

backend/de.metas.manufacturing/src/main/java/org/eevolution/api/PPOrderCreateRequest.java
metas-ts: LE: actually the name OrdercandidateSeqNo does not really help - according to the name the seqno of the order candidate 
...as opposed to the other column of the same table that's just named `SeqNo`.
If anything, wdyt about calling is `ProcessingSeqNo`

metas-ts (changes requested):

PR #13854, opened 2022-11-01
Set incoterms from partners only if has a value

cristinamghita:


backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
```diff
@@ -225,18 +225,21 @@ public void setIncoterms(final I_C_Order order)
...
+			incotermsId = bpartner.getC_Incoterms_Customer_ID();
 		}
 		else
 		{
-			c_Incoterms = bpartner.getC_Incoterms_Vendor_ID();
+			incotermsId = bpartner.getC_Incoterms_Vendor_ID();
 		}
 
-		order.setC_Incoterms_ID(c_Incoterms);
+		if (incotermsId > 0)
```
metas-ts, line 239: the dev-test sais

> - Clone an order with Incoterms set
>    - the new order has the incoterm copied

....but from this snippet here, i can easily construct a test-case that still fails
- create a sales-order with incoterm A
- set the bpartner's `C_Incoterms_Customer_ID` to incoterm B
- clone the sales order
- the clone has incoterm B ❗

PR #13865, opened 2022-11-02
Tenacious d uat unique attribute

metas-rc:
https://github.com/metasfresh/metasfresh/issues/13850


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/attribute/impl/HUUniqueAttributesRepository.java
```diff
@@ -0,0 +1,136 @@
...
+
+	private I_M_HU_UniqueAttribute retrieveHUUniqueAttributeForHU(@NonNull final HuId huId)
+	{
+		return queryBL.createQueryBuilder(I_M_HU_UniqueAttribute.class)
+				.addEqualsFilter(I_M_HU_UniqueAttribute.COLUMNNAME_M_HU_ID, huId)
+				.create()
+				.firstOnly(I_M_HU_UniqueAttribute.class);
+	}
+
+	public I_M_HU_UniqueAttribute retrieveHUUniqueAttributeForProductAndAttributeValue(@NonNull final HUUniqueAttributeParameters huUniqueAttributeParameters)
```
metas-ts, line 81: i don't think we can or should (or have to) assume that an HU has just one unique attribute ❗ 
Afaiu it makes sense to select using M_HU_ID and M_Attribute_ID

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/attribute/impl/HUUniqueAttributesRepository.java
```diff
@@ -0,0 +1,136 @@
...
+				.create()
+				.firstOnly(I_M_HU_UniqueAttribute.class);
+	}
+
+	public List<I_M_HU_Attribute> retrieveHUAttributes(@NonNull final HuPackingInstructionsAttributeId huPIAttributeId)
+	{
+		return queryBL.createQueryBuilder(I_M_HU_Attribute.class)
+				.addEqualsFilter(I_M_HU_Attribute.COLUMNNAME_M_HU_PI_Attribute_ID, huPIAttributeId)
+				.create()
+				.list(I_M_HU_Attribute.class);
```
metas-ts, line 102: there might be a 110million M_HU_Attributes. Even if this is rarely changed, please return an Iterator to avoid OOMEs (i think having the user wait for quite some time is justifyable - they will du this very rearly and will know what they do)

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/model/validator/M_HU.java
```diff
@@ -83,6 +83,23 @@ public void validateStatusChange(@NonNull final I_M_HU hu)
 		huStatusBL.assertStatusChangeIsAllowed(hu, oldHu.getHUStatus(), hu.getHUStatus());
 	}
 
+	@ModelChange(timings = ModelValidator.TYPE_BEFORE_CHANGE, ifColumnsChanged = I_M_HU.COLUMNNAME_HUStatus)
+	public void handleHUUniqueAttributes(@NonNull final I_M_HU hu)
+	{
+		final IHUStatusBL huStatusBL = Services.get(IHUStatusBL.class);
+		final HUUniqueAttributesService huUniqueAttributesService = SpringContextHolder.instance.getBean(HUUniqueAttributesService.class);
```
metas-ts, line 90: please inject the service...either via `backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/model/validator/Main.java`
or by turning it into a `@Component`

metas-ts (changes requested):

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/attribute/impl/HUUniqueAttributesRepository.java
metas-rc: I will also add the product in the query. I don't know if the storages could contain different products for the same HU but codewise they could.

metas-ts (approved):

PR #13905, opened 2022-11-08
Follow-Up  Creating PP Order by machine size

av-ps:


backend/de.metas.manufacturing/src/main/java/org/eevolution/productioncandidate/async/GeneratePPOrderFromPPOrderCandidate.java
```diff
@@ -61,23 +80,19 @@ public Result processWorkPackage(@NonNull final I_C_Queue_WorkPackage workPackag
...
-		if (isDocComplete == null)
-		{
-			return null;
-		}
-
-		return StringUtils.toBoolean(isDocComplete);
+		return candidateStream.filter(orderCandidate -> !orderCandidate.isProcessed())
+				.map(PPOrderCandidateToAllocate::of)
+				.sorted(Comparator.comparing(PPOrderCandidateToAllocate::getHeaderAggregationKey))
+				.collect(ImmutableList.toImmutableList());
```
metas-ts, line 96: We had very recently this PR: https://github.com/metasfresh/metasfresh/pull/13896

It's about  making sure that if a number of PP_Order_Candiadtes are enqueued together, then the order in which the PP_Orders are created (*and thus their order if you sort them by DocumentNo!*), is according to the PP_Order_Candidates' SeqNo.

Please make sure that (still) works :-)

metas-ts (approved):

cristinastefan90:
- [ ] customer is mentioned in the UAT docu
- [ ] the customer issue is move to the "ready-for-UAT" column in the customer-dashboard

# 📖👑 UAT Documentation

- Performed on instance_name / version

image
//testing in progress
- ...

PR #13915, opened 2022-11-09
get rid of int AD_User_ID and replace it with UserId

metas-rc:
https://github.com/metasfresh/metasfresh/issues/13904


adi-stefan (approved):
lgtm.

PR #13932, opened 2022-11-14
prevent adding documents from metasfresh to KT and KV sub-typed invoices

adi-stefan:


metas-ts (approved):
lgtm, assuming that our question to the user is answered the way we guess it will be :-)

PR #13985, opened 2022-11-23
the sha1 below doesn't always work with our migration tool;

metas-ts:
Commit https://github.com/metasfresh/metasfresh/commit/c4b045f3b618a886bf043856b1a949a220d0c85f doesn't work because the migration tool only reads the SQLs to ignore at the start; so one can't ignore files that are picked up in the same run of the tool


metas-ts:
when CPing this, be sure to CP the commit https://github.com/metasfresh/metasfresh/commit/c4b045f3b618a886bf043856b1a949a220d0c85f before it

Issue #14042, opened 2022-11-28
Create a generic function to update the `Fact_Acct.DateAcct`


riadmen:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?

Create a generic function to update the `Fact_Acct.DateAcct` before date

1. Update all `Fact_Acct` for records with `DateAcct < p_DateAcct` and change the `DateAcct = p_DateAcct`. 
2. Update the `DateAcct` for the correspondents documents.
3. Make a backup of the adjusted tables before.

PR #14046, opened 2022-11-28
Create a generic function to update the Fact_Acct.DateAcct before date

riadmen:
https://github.com/metasfresh/metasfresh/issues/14042


metas-ts (approved):
LGTM. I guess this SQL is identical to the one in https://github.com/metasfresh/metasfresh/pull/14043, so i think it would make the most sense to 
- integrate this PR
- close the other
- merge _hotfix 👉 _uat branch

integrating both and then merge hotfix to uat will also work, but probably just with an additional manual handling of (trivial) conflicts

PR #14051, opened 2022-11-29
New Qty Columns in M_Delivery_Planning 

metas-rc:
https://github.com/metasfresh/metasfresh/issues/14050

### DIFF
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningCreateRequest.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningCreateRequest.java
@@ -119,6 +119,12 @@ public class DeliveryPlanningCreateRequest
 	@NonNull
 	Quantity actualDeliveredQty;
 
+	@NonNull Quantity plannedLoadedQty;
+
+	@NonNull Quantity plannedDischargeQty;
+
+	@NonNull Quantity actualDischargeQty;
+
 	@Nullable
 	CountryId originCountryId;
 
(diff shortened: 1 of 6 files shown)


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningCreateRequest.java
```diff
@@ -119,6 +119,12 @@
 	@NonNull
 	Quantity actualDeliveredQty;
 
+	@NonNull Quantity plannedLoadedQty;
```
metas-ts, line 122: i guess this is the pendant to `actualLoadQty`, right? if so, please pick either `load`or `loaded` 🤓

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningCreateRequest.java
```diff
@@ -119,6 +119,12 @@
...
+
+	@NonNull Quantity plannedDischargeQty;
```
metas-ts, line 124: i think i get  `plannedDischargeQty` and `actualDischargeQty`, but then what is `actualDeliveredQty`?
i mean what's the difference between `actualDeliveredQty` and `actualDischargeQty` ?

metas-rc: > what's the difference between `actualDeliveredQty` and `actualDischargeQty` ?

Asked @metasnw

metasnw: @metas-ts depending on the incoterm the good issues (in our language "shipment") is either when the goods are loaded to  a ship or discharged at a warehouse.
Nevertheless we want to save both qtys in case they change (they do actually) from loading to discharge.

metas-rc: thanks

Issue #14062, opened 2022-11-30
`Flat Rate Term` extension uses wrong `Bill Partner`


riadmen:
### Is this a bug or feature request?
BUG
### What is the expected or desired behavior?

All subscriptions with a different billing partner have the wrong billing partner after renewal

PR #14065, opened 2022-12-01
Flat Rate Term extension uses wrong Bill Partner

riadmen:
https://github.com/metasfresh/metasfresh/issues/14062


backend/de.metas.contracts/src/main/java/de/metas/contracts/interceptor/C_Flatrate_Term.java
```diff
@@ -184,7 +184,10 @@ public void setFlatrateDataProcessed(final I_C_Flatrate_Term term)
 	{
 		final I_C_Flatrate_Data flatrateData = term.getC_Flatrate_Data();
 
-		term.setBill_BPartner_ID(flatrateData.getC_BPartner_ID());
+		if (term.getBill_BPartner_ID() == 0)
+		{
```
metas-ts, line 188:
```suggestion
		{ // make sure not to override a Bill_BPartner_ID that is already set via extend-flatrate-term
```

metas-ts (approved):

PR #14067, opened 2022-12-01
Extend CopyRecordSupport to ignore IsCalculated in *some* columns and use it when voiding orders

metas-ts:
… use it when voiding orders


mrie04:
https://github.com/metasfresh/metasfresh/pull/14073

Issue #14072, opened 2022-12-02
Fix Sales Order - void selected order with related documents


mrie04:
### Is this a bug or feature request?
bug
### What is the current behavior?
Sales Order - void selected order with related documents process fails with order-date in the future and product not on currently valid pricelistversion ( product is on future pricelist-version )
### What is the expected or desired behavior?
process works with  order-date in the future and product not on currently valid pricelist-version

PR #14078, opened 2022-12-05
Fix Sales Order - void selected order with related documents (2)

mrie04:
https://github.com/metasfresh/metasfresh/issues/14072


cristinamghita (approved):
LGTM!

Issue #14092, opened 2022-12-06
Allow EDI-Invoice without order-id, if it isn't a credit memo 


mrie04:
### Is this a bug or feature request?
feature
### What is the current behavior?
EDI-Error if Invoice that isn't a Credit Memo doesn't have a order-id
### What is the expected or desired behavior?
no error

PR #14093, opened 2022-12-06
remove EDI isValidInvoice check for order id

mrie04:
https://github.com/metasfresh/metasfresh/issues/14092


metas-ts (approved):
as talked, the XSD didn't change in the end. so, lgtm

PR #14095, opened 2022-12-06
Sum amounts per vatcode

cristinamghita:
Summing up only those amounts that are on tax accounts


cristinamghita:
> the SQL generally looks plausible to me. i have one concern though, regarding the use of IsActive='Y' conditions. We in a DB-function that takes an AD_Org_ID as parameter (there is one somewhere in this PR): why don't we want to filte out inactive orgs?


Yes! Agree! I will take  it out all over!

PR #14121, opened 2022-12-12
Iron plant uat gh14038

cp-ps:


misc/services/camel/de-metas-camel-externalsystems/common/src/main/java/de/metas/camel/externalsystems/common/ExternalSystemCamelConstants.java
```diff
@@ -138,4 +138,7 @@
 
 	public static final String MF_UPSERT_BUDGET_PROJECT_V2_ROUTE_ID = "To-MF_UpsertBudgetProjects-Route";
 	public static final String MF_UPSERT_BUDGET_PROJECT_V2_CAMEL_URI = "metasfresh.upsert-budget-project-v2.camel.uri";
+
+	public static final String RAA_GET_WORK_ORDER_PROJECT_ROUTE_ID = "To-RAA_GetWorkOrderProject-Route";
```
metas-ts, line 142: i think these constants makes no sense over here.. 🤔

metas-ts (changes requested):

misc/services/camel/de-metas-camel-externalsystems/common/src/main/java/de/metas/camel/externalsystems/common/ExternalSystemCamelConstants.java
cp-ps: srry missed that 👍

metas-ts (approved):

PR #14200, opened 2022-12-21
Product API bpartnerProductItems update fails

fp-ps:


backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/product/ProductRestService.java
```diff
@@ -363,6 +363,12 @@ private void createOrUpdateBpartnerProduct(
 			if (effectiveSyncAdvise.getIfExists().isUpdate())
 			{
 				final BPartnerProduct bPartnerProduct = syncBPartnerProductWithJson(jsonRequestBPartnerProductUpsert, existingBPartnerProduct, bPartnerId);
+
+				if (Boolean.TRUE.equals(bPartnerProduct.getCurrentVendor()))
+				{
+					productRepository.resetCurrentVendorFor(productId);
```
metas-ts, line 369: afais this reset now also happens even if the if `bPartnerProduct.getCurrentVendor() == true` already. please du the update only if needed

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/product/ProductRestService.java
```diff
@@ -377,6 +383,12 @@ else if (effectiveSyncAdvise.isFailIfNotExists())
 		else
 		{
 			final CreateBPartnerProductRequest createBPartnerProductRequest = getCreateBPartnerProductRequest(jsonRequestBPartnerProductUpsert, productId, bPartnerId);
+
+			if (Boolean.TRUE.equals(createBPartnerProductRequest.getCurrentVendor()))
+			{
+				productRepository.resetCurrentVendorFor(productId);
```
metas-ts, line 389: ..same as above

metas-ts (changes requested):

backend/de.metas.business/src/main/java/de/metas/product/ProductRepository.java
```diff
@@ -280,6 +280,18 @@ public static BPartnerProduct ofBPartnerProductRecord(@NonNull final I_C_BPartne
 				.usedForVendor(record.isUsedForVendor())
 				.build();
 	}
+
+	public void resetCurrentVendorFor(@NonNull final ProductId productId)
+	{
+		queryBL.createQueryBuilder(I_C_BPartner_Product.class)
+				.addEqualsFilter(I_C_BPartner_Product.COLUMNNAME_M_Product_ID, productId)
+				.addOnlyActiveRecordsFilter()
```
metas-ts, line 288: ..also please don't update all `I_C_BPartner_Product`s  ut just the one that *needs* updating

metas-ts (changes requested):

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/product/ProductRestService.java
fp-ps: this is on create flow. we validate on `createBPartnerProductRequest` as there is no existing record

metas-ts (approved):

PR #14236, opened 2023-01-02
Enhance REST-API with cache control

cp-ps:


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/cache/ThreadLocalCacheController.java
```diff
@@ -0,0 +1,95 @@
...
+			return false;
+		}
+
+		final boolean anyNonApplicationDictionaryTableNames = labels.stream().anyMatch(label -> isNonApplicationDictionaryTableName(label, cacheName));
+		return !anyNonApplicationDictionaryTableNames;
+	}
+
+	private static boolean isNonApplicationDictionaryTableName(final CacheLabel label, final String cacheName)
+	{
+		return label.equalsByName(cacheName) //ignore the label created from this.cacheName as it's not necessary a table name
```
cp-ps, line 59: I think we are missing a `!` -> should be `return !label.equalsByName(cacheName)`

metas-ts (approved):

PR #14250, opened 2023-01-04
Add isPackingMaterialProductCategoryID function for checking all packing material categories

cristinamghita:


metas-ts (approved):
fine with me to integrate, because it's urgent. 
But please create a follow-up to make sure that we search recursively and get the right result with more nested product-category-hierarchies

Issue #14257, opened 2023-01-06
add V2 APIs for Invoice Candidate operations


adi-stefan:
### Is this a bug or feature request?
FR

### What is the expected or desired behavior?
Allow creation of ICs by using external references for product/bpartner/bp location/user

PR #14258, opened 2023-01-06
add V2 APIs for IC operations

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/14257


backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/invoicecandidates/impl/CreateInvoiceCandidatesService.java
```diff
@@ -0,0 +1,575 @@
...
+					.build();
+			result.responseItem(responseItem);
+		}
+
+		//ensure NetAmtToInvoice and other related fields are correctly populated.
+		invoiceCandBL.updateInvalid()
+				.setContext(Env.getCtx(), ITrx.TRXNAME_ThreadInherited)
+				.setTaggedWithAnyTag()
+				.setOnlyInvoiceCandidateIds(InvoiceCandidateIdsSelection.ofIdsSet(invoiceCandidateIds.build()))
+				.update();
```
adi-stefan, line 185: I'm not entirely sure this is wanted, but I was not able to create and enqueue for invoicing an IC from API without this. (Enqueueing failed because the IC would be changed during it. See de.metas.invoicecandidate.api.impl.InvoiceCandidateEnqueuer#enqueueSelectionInTrx )

metas-ts: I see the risk of running into race-condidions while the same code is executed via async-workpackage.
I think if make a breakpoint that stops the whole program, it should turn out that the ICs are already flagged for recompute (=> `c_invoice_candidate_recompute`), at this point, right?
So you could wait for the workpackage-processor to do the work, like `de.metas.cucumber.stepdefs.StepDefUtil#tryAndWaitForItem(long, long, de.metas.cucumber.stepdefs.ItemProvider<T>)` does it.

adi-stefan: yup, looks like I just wasn't waiting enough in my cucumber test. removed

metas-ts (approved):

PR #14380, opened 2023-01-17
Fix HUReceiptScheduleBL by not collecting the HUs of inactive RSAs

metas-ts:
Also rename IHUReceiptScheduleDAO.retrieveHandlingUnitAllocations to ...retrieve**All**HandlingUnitAllocations for more clarify


metas-ts:
@Mergifyio copy master

Issue #14361, opened 2023-01-13
`Purchase Order` incl taxes but generate `Invoice` without taxes


riadmen:
### Is this a bug or feature request?
BUG
#### Which are the steps to reproduce?

1. Clone and complete `PO`: 

<img width="824" alt="image" src="https://user-images.githubusercontent.com/93593841/212169203-74606836-4961-42fd-929e-acd3f011362d.png">

2. Generate directly the invoice:

<img width="823" alt="image" src="https://user-images.githubusercontent.com/93593841/212169053-1e7b86c4-7816-4923-add8-05cd56503476.png">

### What is the expected or desired behavior?

The generated must have the same `PO` taxes

PR #14386, opened 2023-01-17
Purchase Order incl taxes but generate Invoice without taxes

riadmen:
https://github.com/metasfresh/metasfresh/issues/14361

### DIFF
--- a/backend/de.metas.business/src/main/java-legacy/org/compiere/model/CalloutOrder.java
+++ b/backend/de.metas.business/src/main/java-legacy/org/compiere/model/CalloutOrder.java
@@ -1623,7 +1623,7 @@ public class CalloutOrder extends CalloutEngine
 		}
 		else
 		{
-			if (!order.isDropShip())
+			if (order.isDropShip())
 			{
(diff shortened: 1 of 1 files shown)


backend/de.metas.business/src/main/java-legacy/org/compiere/model/CalloutOrder.java
```diff
@@ -1623,7 +1623,7 @@ public String DropShipPartner(final ICalloutField calloutField)
 		}
 		else
 		{
-			if (!order.isDropShip())
+			if (order.isDropShip())
```
metas-ts, line 1626: that `!` is here for a long time. I have problems grasping what the code actually does.
Please elaborate how removing the `!` fixes the current issue without adding other problems

metas-ts: LE: please add those explanations as java-comments

metas-ts: LE II but i still don't get it. if the purchase order is a dropship-order then why should
the drops-ship-location be set to the current org's warehouse after all?

PR #14404, opened 2023-01-18
refs: https://github.com/metasfresh/metasfresh/issues/14223

cp-ps:


misc/services/camel/de-metas-camel-externalsystems/de-metas-camel-grssignum/src/main/java/de/metas/camel/externalsystems/grssignum/from_grs/product/processor/PushRawMaterialsProcessor.java
```diff
@@ -104,19 +113,49 @@ private ProductUpsertCamelRequest getProductUpsertCamelRequest(@NonNull final Js
 				.build();
 	}
 
+	@NonNull
+	private Optional<JsonRequestUpsertProductAllergen> getAllergens(@NonNull final JsonProduct jsonProduct)
+	{
+		if (jsonProduct.getBPartnerProducts() == null || jsonProduct.getBPartnerProducts().size() != 1)
+		{
+			return Optional.empty();
```
metas-ts, line 121: i think it will just be a matter of time until we get a support request asking why stuff is not properly imported. then instead of having to
- analyse the behaviour and find this line in the code
- use git-blame to find the issue
- read the requequirements and uat-docu to find out it was aggreed like this

....please send a log message to metasfresh
(I would prefer giving an HTTP-error back to GRS, but maybe it's too late in the cycle for that)

metas-ts (changes requested):

PR #14409, opened 2023-01-19
ci(Mergify): configuration update

metas-ts:
This change has been made by @metas-ts from Mergify config editor.


metas-ts:
add additional base branches

metas-ts:
add additional base branches

PR #14465, opened 2023-01-30
updated taxcode column

cristinastefan90:
https://github.com/metasfresh/me03/issues/14270

### DIFF
--- a/backend/de.metas.datev/src/main/sql/postgresql/ddl/views/RV_DATEV_Export_Fact_Acct_Invoice.sql
+++ b/backend/de.metas.datev/src/main/sql/postgresql/ddl/views/RV_DATEV_Export_Fact_Acct_Invoice.sql
@@ -1,9 +1,14 @@
+DROP VIEW IF EXISTS RV_DATEV_Export_Fact_Acct_Invoice_PerTax
+;
+DROP VIEW IF EXISTS RV_DATEV_Export_Fact_Acct_Invoice_All
+;
 DROP VIEW IF EXISTS RV_DATEV_Export_Fact_Acct_Invoice
 ;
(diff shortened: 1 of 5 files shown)


backend/de.metas.datev/src/main/sql/postgresql/ddl/views/RV_DATEV_Export_Fact_Acct_Invoice.sql
```diff
@@ -1,9 +1,14 @@
+DROP VIEW IF EXISTS RV_DATEV_Export_Fact_Acct_Invoice_PerTax
+;
+DROP VIEW IF EXISTS RV_DATEV_Export_Fact_Acct_Invoice_All
+;
```
adi-stefan, line 4:
```suggestion
```

These are for git traceability purposes only and are never executed as migrations scripts. So changes here will not be reflected to the DDL.

In addition to these 3 files, you should also add 3 files with the exact same contents in `backend/de.metas.datev/src/main/sql/postgresql/system/`. That's where the Drop Views should happen.
Also, since you're creating a new column in the view, you should add it to `AD_Column`

adi-stefan (changes requested):
Please see below.

backend/de.metas.datev/src/main/sql/postgresql/ddl/views/RV_DATEV_Export_Fact_Acct_Invoice.sql
cristinastefan90: done.

adi-stefan (approved):
lgtm

PR #14495, opened 2023-02-02
make sure that HUs are removed from sourceHU when draft-issued

metas-ts:
#14493


metas-ts:
@MergifyIO copy release

PR #14497, opened 2023-02-03
Fix and test /invoices/payment

dm-ps:


backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/payment/paymentAPI.feature
```diff
@@ -0,0 +1,1439 @@
...
+      | C_DocType_ID.Identifier | DocBaseType |
+      | outboundDocType         | APP         |
+
+    And load C_Currency:
+      | C_Currency_ID.Identifier | OPT.C_Currency_ID |
+      | eur                      | 102               |
+      | chf                      | 318               |
+
+    And load C_Conversion_Rate:
+      | C_Conversion_Rate_ID.Identifier | C_Currency_ID.Identifier | C_Currency_ID_To.Identifier | ValidFrom  | ValidTo    |
```
metas-ts, line 1303: i'm not sure what to make of these two conversion-rates.
- aren't both valid a the same time?
- which one is used by the conversion?

In my mind the tests should make sure that depending on the dates, in one case one conversion is used and in the other case the other conversion

metas-ts (changes requested):

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/payment/paymentAPI.feature
dm-ps: The conversion rate with identifier `currency_rate_1` is validFrom `2015-05-21` and the rate with identifier `currency_rate_2` is validFrom `2019-05-21`.
The `transactionDate` in the request is set to `2016-05-21` which means that `currency_rate_1` should be taken into consideration.
In the next step we validate the `DivideRate` for the two conversion rates, and those are:
 - `1.045236` for `currency_rate_1`
 - `1.13` for `currency_rate_2`
 
The `GrandTotal` of the invoice is `23.8 EUR`
The `amount` in the request is set to `24.88 CHF`
We validate that the invoice is paid in the end, which means that the `currency_rate_1` was taken into consideration (`24.88 CHF`/`1.045236` = `23.8 EUR`),
if it had been `currency_rate_2` the invoice would not have been paid (`24.88 CHF`/`1.13` = `22.01 EUR`).

That's how I was thinking about it.
Is it correct or am I missing something?

metas-ts: hi @dm-ps yes - makes sense, thx for explaining the reasoning. 
what confused me was the `ValidTo` dates. I now checked the code (`CurrencyDAO#retrieveRateQuery`) and found that both `validFrom` and `validTo` are filtered by, and if more than one rate matches a date, then the one with the "youngest" (biggest) `validFrom` is used.
Please formulate this info and add it as a comment to the feature-file and the step-def. 
The goal being that i don't have to ask this again in 2 months :-)

metas-ts (approved):

PR #14511, opened 2023-02-07
Create Cucumber for accounting

dm-ps:
https://github.com/metasfresh/metasfresh/issues/14508


metas-ts (commented):
I don't know a lot about accounting, but shouldn't the CR and DR sides be "balanced" in the cucumber-tests?

metas-ts (approved):

Issue #14506, opened 2023-02-06
New Fields on `Sales Order` - `Invoice Candidate` - `Invoice`


riadmen:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?

- New Fields on Sales Order `Additional Text for Invoice` and `Do not show Country of Origin`
- Invoicing copies the field `Additional Text for Invoice` and `Do not show Country of Origin` from `Sales Order` to `Sales invoice`

PR #14517, opened 2023-02-07
Gh14506

riadmen:
https://github.com/metasfresh/metasfresh/issues/14506

### DIFF
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java-gen/org/compiere/model/I_C_Order.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java-gen/org/compiere/model/I_C_Order.java
@@ -1442,7 +1442,7 @@ public interface I_C_Order
 	String COLUMNNAME_EMail = "EMail";
 
 	/**
-	 * Set External ID.
+	 * Set SAP PayT ID.
 	 *
 	 * <br>Type: String
 	 * <br>Mandatory: false
@@ -1708,6 +1708,49 @@ public interface I_C_Order
 	ModelColumn<I_C_Order, Object> COLUMN_IncotermLocation = new ModelColumn<>(I_C_Order.class, "IncotermLocation", null);
 	String COLUMNNAME_IncotermLocation = "IncotermLocation";
 
+	/**
+	 * Set Additional Text for Invoice.
+	 *
+	 * <br>Type: Text
+	 * <br>Mandatory: false
+	 * <br>Virtual Column: false
+	 */
+	void setInvoiceAdditionalText (@Nullable java.lang.String InvoiceAdditionalText);
+
+	/**
+	 * Get Additional Text for Invoice.
+	 *
+	 * <br>Type: Text
+	 * <br>Mandatory: false
+	 * <br>Virtual Column: false
+	 */
+	@Nullable java.lang.String getInvoiceAdditionalText();
+
+	ModelColumn<I_C_Order, Object> COLUMN_InvoiceAdditionalText = new ModelColumn<>(I_C_Order.class, "InvoiceAdditionalText", null);
+	String COLUMNNAME_InvoiceAdditionalText = "InvoiceAdditionalText";
+
+	/**
+	 * Set Do not show Country of Origin.
+	 * If is NO, then the Country of Origin of the products is displayed in the invoice report
+	 *
+	 * <br>Type: YesNo
+	 * <br>Mandatory: true
+	 * <br>Virtual Column: false
+	 */
+	void setIsNotShowOriginCountry (boolean IsNotShowOriginCountry);
+
+	/**
+	 * Get Do not show Country of Origin.
+	 * If is NO, then the Country of Origin of the products is displayed in the invoice report
+	 *
+	 * <br>Type: YesNo
+	 * <br>Mandatory: true
+	 * <br>Virtual Column: false
+	 */
+	boolean isNotShowOriginCountry();
+
+	ModelColumn<I_C_Order, Object> COLUMN_IsNotShowOriginCountry = new ModelColumn<>(I_C_Order.class, "IsNotShowOriginCountry", null);
+	String COLUMNNAME_IsNotShowOriginCountry = "IsNotShowOriginCountry";
 	/**
 	 * Set Invoice Rule.
 	 * Frequency and method of invoicing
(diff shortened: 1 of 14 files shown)


backend/de.metas.adempiere.adempiere/base/src/main/java-gen/org/compiere/model/I_C_Order.java
```diff
@@ -1442,7 +1442,7 @@ public interface I_C_Order
 	String COLUMNNAME_EMail = "EMail";
 
 	/**
-	 * Set External ID.
+	 * Set SAP PayT ID.
```
adi-stefan, line 1445: I'm not sure this is correct. Why did the vanilla AD_Element associated with ExternalId change?

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/70-de.metas.fresh/5675830_sys_gh14506_IsNotShowOriginCountry.sql
```diff
@@ -0,0 +1,107 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
```
adi-stefan, line 64: Why is this in the middle of the file?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceHeaderImpl.java
```diff
@@ -400,4 +404,27 @@ public SectionCodeId getM_SectionCode_ID()
...
+
+	public void setInvoiceAdditionalText(final String invoiceAdditionalText)
+	{
+		this.invoiceAdditionalText = invoiceAdditionalText;
+	}
+
+	public void setNotShowOriginCountry(final boolean notShowOriginCountry)
+	{
+		this.notShowOriginCountry = notShowOriginCountry;
+	}
```
adi-stefan, line 429: Why is this here? You are not aggregating by this field.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandBLCreateInvoices.java
```diff
@@ -433,6 +433,10 @@ private I_C_Invoice createInvoice(
 
 			invoice.setPaymentRule(invoiceHeader.getPaymentRule());
 			invoice.setM_SectionCode_ID(SectionCodeId.toRepoId(invoiceHeader.getM_SectionCode_ID()));
+
+			invoice.setInvoiceAdditionalText(invoiceHeader.getInvoiceAdditionalText());
+			invoice.setIsNotShowOriginCountry(invoiceHeader.isNotShowOriginCountry());
```
adi-stefan, line 438: Instead of here, you should be populating this field in: `de.metas.invoice.service.impl.AbstractInvoiceBL#createInvoiceFromOrder`. That way, there's no need for it to be in the InvoiceHeader.

adi-stefan (changes requested):
See my comments.

backend/de.metas.adempiere.adempiere/base/src/main/java-gen/org/compiere/model/I_C_Order.java
riadmen: corrected

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/70-de.metas.fresh/5675830_sys_gh14506_IsNotShowOriginCountry.sql
riadmen: corrected

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceHeaderImpl.java
riadmen: To copy the new fields from Invoice candidate to the invoice

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandBLCreateInvoices.java
riadmen: To copy the new fields from Invoice candidate to the invoice

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceHeaderImpl.java
adi-stefan: What I meant is that, because you're not ensuring uniqueness of the flag for all ICs (via aggregation), it's always taken from the first/last processed IC.
Is this intended?

metas-ts: @riadmen (@adi-stefan ) right, i also had this concern in the TC review, but then wrote that you are handling it further down the TC. But unfortunately just just do it in the user's custom intance which is not sufficient.
you need to add `InvoiceAdditionalText` to the *standard* metasfresh's `invoicing-agg-std` aggregator.

and with respect to `notShowOriginCountry` =>if this is going into C_Invoice (and not into C_ÌnvoiceLine), then it *also* needs to go into the *standard* metasfresh's `invoicing-agg-std` aggregator.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandBLCreateInvoices.java
metas-ts: Besides `AbstractInvoiceBL#createInvoiceFromOrder` (which I didn't dig into now), i would like those properties to enter the invoice via C_Invoice_Candidate and not directly coming form order 
(because we have many sources for invoice candidates and in future they might also want to control these two properties)

adi-stefan (approved):
lgtm.

Issue #14478, opened 2023-02-01
When creating PO from SO allow grouping of PO lines by DatePromised


adi-stefan:
### Is this a bug or feature request?
FR

PR #14527, opened 2023-02-09
#14478 fix DatePromised TRLs for Create PO from SO process

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/14478


adi-stefan:
trl only, integrating without approval.

PR #14549, opened 2023-02-11
Import Exchange Rate from .dat file

metas-ts:


metas-ts:
checked the code - it looks OK to me to integrate for now; @cp-ps @av-ps feel free to push additional improvements to another PR

PR #14566, opened 2023-02-14
fix invoiceAPI test

av-ps:


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoice/ManualInvoiceService.java
```diff
@@ -235,70 +236,77 @@ private CreateManualInvoiceLineRequest buildManualInvoiceLine(
...
 		final ProductPrice manualPrice = request.getPriceEntered(requestLine).orElse(null);
 
+		TaxId taxId = Optional.ofNullable(requestLine.getVatCodeId())
+				.map(this::getTaxByVatId)
+				.orElse(null);
+
+		final IPricingResult pricingResult = Optional.of(taxId == null || manualPrice == null)
+				.filter(calculatePrice -> calculatePrice)
+				.map(ignored -> getPricingResult(request.getHeader(), requestLine, priceListId, countryId, zoneId))
+				.orElse(null);
```
metas-ts, line 248:
```suggestion
		final boolean calculatePrice = taxId == null || manualPrice == null;
		final IPricingResult pricingResult  = calculatePrice ? getPricingResult(request.getHeader(), requestLine, priceListId, countryId, zoneId) : null;
```

i think my version is more readable

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoice/ManualInvoiceService.java
```diff
@@ -235,70 +236,77 @@ private CreateManualInvoiceLineRequest buildManualInvoiceLine(
...
-															  requestLine,
-															  priceListId,
-															  countryId,
-															  zoneId);
-
 		final ProductPrice manualPrice = request.getPriceEntered(requestLine).orElse(null);
 
+		TaxId taxId = Optional.ofNullable(requestLine.getVatCodeId())
+				.map(this::getTaxByVatId)
+				.orElse(null);
```
metas-ts, line 243:
```suggestion
		final VatCodeId vatCodeId = requestLine.getVatCodeId();
		TaxId taxId = (vatCodeId != null) ? getTaxByVatId(vatCodeId) : null;
```

not everything that *can* be cone with streams&lambdas *should* be done with streams&lambdas..

metas-ts (changes requested):
generally lgtm, but maybe just because i'm old...please consider my suggestions 🙃

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoice/ManualInvoiceService.java
cp-ps: agree, but here I don't see why one option is better than the other :-?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoice/ManualInvoiceService.java
cp-ps: maybe 😊

metas-ts: so this one really gave me a headache to read at first..like..."are they really doing an if-then-else here?"
I mean it's not wrong, per se (unless you visualize the stacktrace below the `getPricingResult` maybe ),
but that `.filter(calculatePrice -> calculatePrice)`....
So, let's just say i would prefer the code to be boring on this low level, and creativity to be applied to the higher levels

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoice/ManualInvoiceService.java
metas-ts: I aggree, this one is OK, but I propose to change it so that we have `if ? then : else` twice in the method.
So that the reader, after having wrapped her/his head around the first  `if ? then : else` doesn't have to take on an `Optional.map.orElse;` expression next.

Issue #14569, opened 2023-02-14
Generate Invoice with Allowing Override Due Date


cristinamghita:
### Is this a bug or feature request?
FR

PR #14574, opened 2023-02-14
#14569 Generate Invoice with Allowing Override Due Date

cristinamghita:
#14569 Generate Invoice with Allowing Override Due Date


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/AggregationEngine.java
```diff
@@ -552,6 +565,37 @@ private LocalDate computeDateAcct(@NonNull final I_C_Invoice_Candidate ic)
 				});
 	}
 
+	private LocalDate computeOverrideDueDate(@NonNull final I_C_Invoice_Candidate ic)
+	{
+		return CoalesceUtil.coalesceSuppliers(
+				() -> {
+					if (overrideDueDateParam != null)
+					{
+						logger.debug("computeOverrideDueDate - returning aggregator's overrideDueDateParam={} as overrideDueDate", overrideDueDateParam);
```
metas-ts, line 574: this logger message seems odd - after all you might still return `null` in the end

metas-ts (approved):

PR #14646, opened 2023-02-20
Order Window Ivict adaptions (Billto Default, Drop Shipment)

av-ps:
automatically mark order as DropShip if dropShip warehouse was selected

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/order/IOrderBL.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/IOrderBL.java
@@ -283,5 +283,8 @@ public interface IOrderBL extends ISingletonService
 
 	void updateIsOnConsignmentFromLines(OrderId orderId);
 
+	/**
+	 * @return {@code true} if metasfresh should use the default-billTo-location for {@code C_Order.C_BPartner_Location_ID}
+	 */
 	boolean isUseDefaultBillToLocationForBPartner(@NonNull I_C_Order order);
 }
--- a/backend/de.metas.business/src/main/java/de/metas/order/IOrderBL.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/IOrderBL.java
@@ -283,8 +283,4 @@ public interface IOrderBL extends ISingletonService
 
 	void updateIsOnConsignmentFromLines(OrderId orderId);
 
-	void markAsDropShipIfDropShipWarehouse(I_C_Order order);
-
-	void setBillToDefaultLocationAsDefaultIfDropShipSO(I_C_Order order);
-
 }
--- a/backend/de.metas.business/src/main/java/de/metas/order/IOrderBL.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/IOrderBL.java
@@ -283,4 +283,5 @@ public interface IOrderBL extends ISingletonService
 
 	void updateIsOnConsignmentFromLines(OrderId orderId);
 
+	boolean isUseDefaultBillToLocationForBPartner(@NonNull I_C_Order order);
 }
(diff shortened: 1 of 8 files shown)


backend/de.metas.business/src/main/java/de/metas/order/IOrderBL.java
```diff
@@ -283,4 +283,5 @@ default void setDocTypeTargetIdAndUpdateDescription(final I_C_Order order, final
 
 	void updateIsOnConsignmentFromLines(OrderId orderId);
 
+	boolean isUseDefaultBillToLocationForBPartner(@NonNull I_C_Order order);
```
metas-ts, line 286:
```suggestion
	/**
	 * @return {@code true} if metasfresh should use the default-billTo-location for {@code C_Order.C_BPartner_Location_ID}
	 */
	boolean isUseDefaultBillToLocationForBPartner(@NonNull I_C_Order order);
```

backend/de.metas.business/src/main/sql/postgresql/system/20-de.metas.business/5679120_sys_gh14645_UseDefaultBillToLocationAsOrderDefaultLocation_sys_config.sql
```diff
@@ -0,0 +1,4 @@
+-- 2023-02-23T13:01:53.284Z
+INSERT INTO AD_SysConfig (AD_Client_ID,AD_Org_ID,AD_SysConfig_ID,ConfigurationLevel,Created,CreatedBy,EntityType,IsActive,Name,Updated,UpdatedBy,Value) VALUES (0,0,541597,'S',TO_TIMESTAMP('2023-02-23 15:01:53','YYYY-MM-DD HH24:MI:SS'),100,'U','Y','de.metas.order.impl.UseDefaultBillToLocationAsOrderDefaultLocation',TO_TIMESTAMP('2023-02-23 15:01:53','YYYY-MM-DD HH24:MI:SS'),100,'N')
```
metas-ts, line 2: please add a description to the asysconfig, such that the user knows
- what it is about
- how the system actually behaves for Y and for N

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
```diff
@@ -1250,4 +1253,31 @@ public void updateIsOnConsignmentFromLines(OrderId orderId)
 		order.setIsOnConsignment(isOnConsignment);
 		save(order);
 	}
+
+	public boolean isUseDefaultBillToLocationForBPartner(@NonNull final I_C_Order order)
+	{
+		if (!sysConfigBL.getBooleanValue(SYSCONFIG_USE_DEFAULT_BILL_TO_LOCATION_AS_ORDER_DEFAULT_LOCATION, false))
```
metas-ts, line 1259: please let this sysconfig depend also on client and org (of the order).

backend/de.metas.business/src/main/sql/postgresql/system/20-de.metas.business/5679120_sys_gh14645_UseDefaultBillToLocationAsOrderDefaultLocation_sys_config.sql
metas-ts: please use a proper entitiy-type - either an order-specific one or D

metas-ts: please set the ConfigurationLevel to "Org". then one doesn't need to update this single record, but can override it with different values for different orgs

Issue #14686, opened 2023-02-23
Do not reset `BillToAddress` whenever Delivery Address changed


chemsnouioua:
### Is this a bug or feature request?
Bug

### What is the current behavior?
/
#### Which are the steps to reproduce?
/
### What is the expected or desired behavior?
/

PR #14687, opened 2023-02-23
#14686

chemsnouioua:
#14686

### DIFF
--- a/backend/de.metas.business/src/main/java-legacy/org/compiere/model/CalloutOrder.java
+++ b/backend/de.metas.business/src/main/java-legacy/org/compiere/model/CalloutOrder.java
@@ -332,17 +332,13 @@ public class CalloutOrder extends CalloutEngine
 		}
 
 
-		if(order.getBill_Location_ID() <= 0)
+		if (!orderBL.setBillLocation(order))
 		{
-			final boolean isBillLocationSet = orderBL.setBillLocation(order);
-			if (!isBillLocationSet)
-			{
 				final String localizedMessage = new BPartnerNoBillToAddressException(bpartner).getLocalizedMessage();
 				calloutField.fireDataStatusEEvent(
 						localizedMessage,
 						localizedMessage, // this appears onHover
 						true);
-			}
 		}
 
 		return NO_ERROR;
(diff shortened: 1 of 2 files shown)


backend/de.metas.business/src/main/java-legacy/org/compiere/model/CalloutOrder.java
```diff
@@ -331,13 +331,18 @@ public String bPartnerLocation(final ICalloutField calloutField)
 			orderBL.setBPLocation(order, bpartner);
 		}
 
-		if (!orderBL.setBillLocation(order))
+
+		if(order.getBill_Location_ID() <= 0)
 		{
```
cristinamghita, line 336: @chemsnouioua I am not ok with this. Pls debug here `de.metas.order.impl.OrderBL#setBillLocation` why this set to null and to the proper address.

cristinamghita (changes requested):
See my comment

backend/de.metas.business/src/main/java-legacy/org/compiere/model/CalloutOrder.java
chemsnouioua: @cristinamghita 
After the change of the delivery address, the callout  `org.compiere.model.CalloutOrder#bPartnerLocation` starts.
When de.metas.order.impl.OrderBL#setBillLocation whe do not calculate nor pass the `BillToAddress` :eye: screenshot
image

imho, my fix is still valid, try to get it only when its  null

cristinamghita: @chemsnouioua 
Yes, we do it. Here: `de.metas.document.location.adapter.IDocumentBillLocationAdapter#setFrom(de.metas.document.location.DocumentLocation)`

chemsnouioua: replied with DM

cristinamghita: @chemsnouioua See my fix. Please, test it. I did not!

cristinamghita (changes requested):
Dev test it!

chemsnouioua (commented):
works fine, thx @cristinamghita

backend/de.metas.business/src/main/java-legacy/org/compiere/model/CalloutOrder.java
chemsnouioua: works fine, thx @cristinamghita

PR #14702, opened 2023-02-28
Make sure that ICs are always enqueued in their intended order (_uat)

metas-ts:


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandidateEnqueuer.java
```diff
@@ -80,9 +84,36 @@ public IInvoiceCandidateEnqueueResult enqueueInvoiceCandidateIds(final Set<Invoi
...
+		// Create invoice candidates changes checker.
+		final IInvoiceCandidatesChangesChecker icChangesChecker = newInvoiceCandidatesChangesChecker();
+		icChangesChecker.setBeforeChanges(unorderedICs);
+
+		//
+		// Prepare them in a dedicated trx so that the update-WP-processor "sees" them
+		trxManager.runInNewTrx(() -> updateSelectionBeforeEnqueueing(pInstanceId));
+		// NOTE: after running that method we expect some invoice candidates to be invalidated, but that's not a problem because:
+		// * the ones which are in our selection, we will update right now (see below)
+		// * the other ones will be updated later, asynchronously
```
metas-ts, line 107:
```suggestion
```

metas-ts:
## Commit-Message-Draft

```
* !!! Make sure that ICs are always enqueued in their intended order
  We achieve this by making sure that the Iterator returned by IInvoiceCandDAO.retrieveIcForSelectionStableOrdering has a fixed ordering that won't change, even if it's underlying ICs might be updated.
...and then we use that method in InvoiceCandidateEnqueuer.enqueueSelectionInTrx()
* !!! Avoid concurrently revalidating the invoice-candidates we just have updated - leave that job to the workpackage-processor
  I.e. don't let the method called by the supplier incidentally enqueue other Workpackages.
* !!! Add both UpdateInvalidInvoiceCandidatesWorkpackageProcessor and InvoiceCandWorkpackageProcessor (which creates invoices) to the same Queue-Processor, so they are not any longer executed concurrently.
* Fix issues around unlocking the WP and its elements;
* also, further minor improvements:
  * allow totalNetAmtToInvoiceChecksum to be null as it's implemented
  * generally relax timeouts from 30s to 60s - note that this is not a CP of https://github.com/metasfresh/metasfresh/commit/95d4129c839e69788b17841d7b7ac889cb5e999b (because there were many conflicts), but a "fresh" search&replace
  * add hard_encoded_uat env file and launch configs
  * add other env-files and fix run-config (from master brnch)
  * make cucumber tests start faster (via IntelliJ run-config) and without that ugly error

#14708
```

PR #14712, opened 2023-03-01
avoid concurrently updating the invoice-candidates we just have updated

metas-ts:
#14708


metas-ts:
Closing because: 
- there is a workaround and we don't need a hotfix-rollout anymore
- we have a more thorrow improvement in https://github.com/metasfresh/metasfresh/pull/14702

PR #14800, opened 2023-03-08
Improve stability of the invoicing process (#14702)

metas-ts:
* !!! Make sure that ICs are always enqueued in their intended order We achieve this by making sure that the Iterator returned by IInvoiceCandDAO.retrieveIcForSelectionStableOrdering has a fixed ordering that won't change, even if it's underlying ICs might be updated. ...and then we use that method in InvoiceCandidateEnqueuer.enqueueSelectionInTrx()
* !!! Avoid concurrently revalidating the invoice-candidates we just have updated - leave that job to the workpackage-processor I.e. don't let the method called by the supplier incidentally enqueue other Workpackages.
* !!! Add both UpdateInvalidInvoiceCandidatesWorkpackageProcessor and InvoiceCandWorkpackageProcessor (which creates invoices) to the same Queue-Processor, so they are not any longer executed concurrently.
* Fix issues around unlocking the WP and its elements;
* also, further minor improvements:
  * allow totalNetAmtToInvoiceChecksum to be null as it's implemented
  * generally relax timeouts from 30s to 60s - note that this is not a CP of https://github.com/metasfresh/metasfresh/commit/95d4129c839e69788b17841d7b7ac889cb5e999b (because there were many conflicts), but a "fresh" search&replace
  * add hard_encoded_uat env file and launch configs
  * add other env-files and fix run-config (from master brnch)
  * make cucumber tests start faster (via IntelliJ run-config) and without that ugly error

#14708

(cherry picked from commit 93bf802cf147bf5896d63d3e422a31c74db67dc9)

### DIFF
--- a/backend/de-metas-salesorder/src/main/java/de/metas/salesorder/candidate/ProcessOLCandsWorkpackageProcessor.java
+++ b/backend/de-metas-salesorder/src/main/java/de/metas/salesorder/candidate/ProcessOLCandsWorkpackageProcessor.java
@@ -26,6 +26,7 @@ import de.metas.async.model.I_C_Queue_WorkPackage;
 import de.metas.async.spi.WorkpackageProcessorAdapter;
 import de.metas.process.PInstanceId;
 import de.metas.util.Check;
+import lombok.NonNull;
 import org.compiere.SpringContextHolder;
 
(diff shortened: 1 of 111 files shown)


metas-ts:
@MergifyIO big_coconut_uat

metas-ts:

Issue #14814, opened 2023-03-09
Accounting Export Update


riadmen:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?

- When creating data in the Transactions window export process
- Only for Creditors
- When it's a EU Transaction (intra-community delivery IGL)
- Then the respective Austrian domestic tax rate is to be recorded in BUHA Export (reduced, standard)
- and be issued with booking code 9

PR #14815, opened 2023-03-09
Tax rate modification in accounting export

riadmen:
https://github.com/metasfresh/metasfresh/issues/14814


cristinamghita (approved):
LGTM!

PR #14858, opened 2023-03-14
Do not prevent Goods Receipt if missing delivery instructions

av-ps:
https://github.com/metasfresh/metasfresh/issues/14843


adi-stefan (approved):
lgtm

PR #14875, opened 2023-03-15
Script correction after merge from release

metas-rc:


backend/de.metas.async/src/main/sql/postgresql/system/80-async-dlm/5624381_sys_gh12591_migrate_c_queue_package_processor_ID.sql
```diff
@@ -1,3 +1,68 @@
...
+             FROM workPackageIDsToDelete wpd WHERE wp.c_queue_workpackage_id = wpd.c_queue_workpackage_ID)
+;
+
+DELETE
+FROM c_queue_block b
+WHERE NOT EXISTS(SELECT 1 FROM c_queue_packageprocessor WHERE c_queue_packageprocessor_id = b.c_queue_packageprocessor_id)
+;
+
+
+DELETE
```
metas-ts, line 55: No need to delete these `c_queue_workpackage_archived`.
AFAIS they have no foreign-key constraints

backend/de.metas.async/src/main/sql/postgresql/system/80-async-dlm/5624381_sys_gh12591_migrate_c_queue_package_processor_ID.sql
```diff
@@ -1,3 +1,68 @@
...
+
+CREATE TABLE backup.BKP_c_queue_element_15032023 AS
+SELECT *
+FROM c_queue_element
+;
+
+
+-- DELETE invalid data
+
+CREATE TEMP TABLE workPackageIDsToDelete AS
```
metas-ts, line 28: It's fine like this, but in general i would consider to not have this as a temp table. 
Instead create the table as `backup.workPackageIDsToDelete_15032023`.
Then you can later see what was deleted and maybe x-check with a backup or archive data.

metas-ts (approved):
anyways, lgtm

Issue #15086, opened 2023-04-06
Field `JobRouter Instance ID` in SO-PO invoices


riadmen:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
Adding `JobRouter Instance ID` field in SO-PO invoices:
- Read-only if the invoice was created via API for purchase invoices, but read write for sales invoices

PR #15106, opened 2023-04-12
Field Externe ID in SO-PO invoices

riadmen:
https://github.com/metasfresh/metasfresh/issues/15086


mrie04 (changes requested):
pls check failing cucumber tests related to this changes for example https://github.com/metasfresh/metasfresh/actions/runs/4690874256/jobs/8315145096?pr=15106#step:9:419

mrie04 (commented):
LGTM now

Issue #15119, opened 2023-04-18
start_esb-camel_docker.sh' returned a non-zero code: 139


metas-marcus:
https://jenkins.metasfresh.com/job/metasfresh/job/master/24320/console
Jenkins build of master fails with 
```
13:03:37  Step 7/10 : RUN dos2unix /opt/metasfresh-esb-camel/start_esb-camel_docker.sh
13:03:37   ---> Running in 43001852d723
13:03:38  The command '/bin/sh -c dos2unix /opt/metasfresh-esb-camel/start_esb-camel_docker.sh' returned a non-zero code: 139
```

PR #15120, opened 2023-04-18
removed dos2unix to rule out potential cause for jenkins build fail

metas-marcus:
https://github.com/metasfresh/metasfresh/issues/15119

Jenkins build without dos2unix command succeeded: https://jenkins.metasfresh.com/job/metasfresh/job/master_gh15119/2/

### DIFF
--- a/misc/services/camel/de-metas-camel-edi/Dockerfile
+++ b/misc/services/camel/de-metas-camel-edi/Dockerfile
@@ -6,7 +6,7 @@ FROM openjdk:17-jdk-alpine
 # thx to https://github.com/moby/moby/issues/1996#issuecomment-185872769
 ARG CACHEBUST=1
 
-RUN apk update && apk upgrade && apk add bash && apk add curl && apk add dos2unix --update-cache --repository http://dl-3.alpinelinux.org/alpine/edge/community/ --allow-untrusted
+RUN apk update && apk upgrade && apk add bash && apk add curl --update-cache --repository http://dl-3.alpinelinux.org/alpine/edge/community/ --allow-untrusted
 
 COPY ./target/de-metas-camel-edi-exec.jar /opt/metasfresh-esb-camel/
 COPY ./start_esb-camel_docker.sh /opt/metasfresh-esb-camel/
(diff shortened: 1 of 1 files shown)


metas-ts (approved):

misc/services/camel/de-metas-camel-edi/Dockerfile
```diff
@@ -6,7 +6,7 @@ FROM openjdk:17-jdk-alpine
 # thx to https://github.com/moby/moby/issues/1996#issuecomment-185872769
 ARG CACHEBUST=1
 
-RUN apk update && apk upgrade && apk add bash && apk add curl && apk add dos2unix --update-cache --repository http://dl-3.alpinelinux.org/alpine/edge/community/ --allow-untrusted
+RUN apk update && apk upgrade && apk add bash && apk add curl --update-cache --repository http://dl-3.alpinelinux.org/alpine/edge/community/ --allow-untrusted
 
 COPY ./target/de-metas-camel-edi-exec.jar /opt/metasfresh-esb-camel/
```
cp-ps, line 11: I don't think that's the way to go here, we had dos2unix in place to take care of the line separators conversion between windows and linux ( at least that's what I've understood), so I'd imagine we still need that...

also, another thing to consider is that there are other Dockerfiles where we do the exact same thing for other shell scripts... and there everything goes well, the only difference being the package manager (i.e. apt instead of apk) so most probably we end up with different versions of the dos2unix utility.

now, I've looked at https://dl-3.alpinelinux.org/alpine/edge/community, the `dos2unix` source in `de-metas-camel-edi/Dockerfile` and I've noticed the package was updated on 13th of April 2023 => most probably that's the root cause of the failure.

cp-ps (changes requested):

misc/services/camel/de-metas-camel-edi/Dockerfile
cp-ps: but ofc, if it's fine to just drop it, then let's save some effort and just drop it :)

metas-marcus: I was told by @metas-ts that dos2unix is most likely no longer needed. 
When I attempted the docker build myself it ran just fine. Only jenkins produced the 139 Error every single time. Maybe its due to the nested and temporary directories and they no longer play fine with the latest version of the dos2unix tool? Just an assumption.

metas-ts:
@MergifyIO copy iron_plant_uat

PR #15164, opened 2023-04-24
DESADV ist fehlerhaft - Lieferscheinnummer anstelle von DESADV Nummer übertragen

fp-ps:


cp-ps (changes requested):
IMHO the implementation is not completed, as from my understanding we should export the `mInoutsDocumentNo` as `DocumentNo` in the Desadv XML, and ofc, this should not be the default behavior but something configurable via sys config... but pls clarify with the PO

metas-ts:
hi @fp-ps @cp-ps (cc @metasnw ),

> IMHO the implementation is not complete, as from my understanding we should export the mInoutsDocumentNo as the existing DocumentNo field in the Desadv XML... and ofc, this should not be the default behavior but something configurable via sys config... but pls clarify with the PO

No, it's alright to have a dedicated field for the shipmentDocumentNo(s).
Desadv has its own documentno and there might be multiple shipments within metasfresh. If we use the existing column (and possible a sysconfig), we enter the world of FUD

cp-ps (approved):

Issue #15163, opened 2023-04-24
Display Incoterm/Incoterm Location in  SO Window


chemsnouioua:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Not displayed in main view
#### Which are the steps to reproduce?
/
### What is the expected or desired behavior?
Display  Incoterm/Incoterm Location in SO Window, main window

PR #15165, opened 2023-04-24
#15163 display Incoterm/Incoterm Location in SO window

chemsnouioua:
#15163


metas-ts:
@mergifyio copy release

metas-ts:
(ok, probably this already is in release)

PR #15173, opened 2023-04-25
Do not create HU Trace for top HU in advance

cristinamghita:


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/trace/HUTraceEventsService.java
```diff
@@ -661,7 +661,7 @@ void createAndAddEventsForManufacturingIssue(
 					.eventTime(ppCostCollector.getUpdated().toInstant())
 					.topLevelHuId(huId);
 
-			createTraceForProductionIssueHU(builder, ppOrderId, huRecord);
+			// createTraceForProductionIssueHU(builder, ppOrderId, huRecord);
```
metas-ts, line 664:
```suggestion
			// createTraceForProductionIssueHU(builder, ppOrderId, huRecord); the traces are already created for the vhus which  includes the huRecord. Making this call would result in duplicates.
```

metas-ts (approved):

Issue #15201, opened 2023-04-28
Set Drop ship partner and location from from flatrate term when creating printing queue


cristinamghita:
### Is this a bug or feature request?
FR

PR #15202, opened 2023-04-28
#15201 Set Drop ship partner and location from from flatrate term when creating printing queue 

cristinamghita:
#15201

### DIFF
--- a/backend/de.metas.contracts/src/main/java/de/metas/contracts/printing/impl/FlatrateTermPrintingQueueHandler.java
+++ b/backend/de.metas.contracts/src/main/java/de/metas/contracts/printing/impl/FlatrateTermPrintingQueueHandler.java
@@ -1,0 +1,89 @@
+/*
+ * #%L
+ * de.metas.contracts
+ * %%
+ * Copyright (C) 2023 metas GmbH
+ * %%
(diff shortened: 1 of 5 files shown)


backend/de.metas.contracts/src/main/java/de/metas/contracts/printing/impl/FlatrateTermPrintingQueueHandler.java
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+
+	@Override
+	public void afterEnqueueBeforeSave(final I_C_Printing_Queue queueItem, final I_AD_Archive archive)
+	{
+		logger.debug("C_Printing_Queue={}; AD_Archive {} has [AD_Table_ID={}, Record_ID={}, C_BPartner_ID={}];",
+				new Object[] { queueItem, archive, archive.getAD_Table_ID(), archive.getRecord_ID(), archive.getC_BPartner_ID() });
+
+		final Object archiveRefencedModel = Services.get(IArchiveDAO.class).retrieveReferencedModel(archive, Object.class);
```
metas-ts, line 67: Y not have this service also as a field?

metas-ts (approved):

PR #15227, opened 2023-05-02
HU traces report

metas-rc:


backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/ddl/public/functions/getCurrentStorageStock.sql
```diff
@@ -1,11 +1,13 @@
 DROP FUNCTION IF EXISTS getCurrentStorageStock(p_M_Product_ID   numeric,
+                                               p_C_UOM_ID       numeric,
```
metas-ts, line 2: wdyt about adding this DDL to de.metas.handlingunits?

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/ddl/public/functions/M_HU_Trace_Report.sql
```diff
@@ -18,14 +18,15 @@ AS
 $BODY$
 
 
-SELECT  t.LotNumber              AS LotNumber,
+SELECT t.LotNumber              AS LotNumber,
```
metas-ts, line 21: wdyt about adding this DDL to de.metas.handlingunits as well?

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/inventory/InventoryRepository.java
```diff
@@ -716,4 +722,55 @@ public Inventory createInventoryLine(@NonNull final InventoryLineCreateRequest r
...
+				.filter(filter)
+				.create()
+				.stream()
+				.map(huAssignment -> TableRecordReference.ofReferenced(huAssignment))
+				.sorted(Comparator.comparing(TableRecordReference::getRecord_ID))
+				.distinct()
+				.map(ref -> ref.getModel(ctxAware, I_M_InventoryLine.class))
+				.collect(Collectors.toList());
+
+		return ImmutableSet.<I_M_InventoryLine>builder()
```
metas-ts, line 769: when adding model classes to sets like this, i'm always a bit uneasy because I have a big amound of FUD towards `hashCode()` & `equals()`: if we load the same `M_InventoryLine` record from DB twice - will the corresponding `I_M_InventoryLine`s be èqual to each other?

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/ddl/public/functions/M_HU_Trace_Report.sql
metas-rc: I moved this too but it's not visible here as "Outdated". I don't know why it is 🤔

metas-ts (approved):

PR #15233, opened 2023-05-03
new class for projectRefSeq Changes

MHNV:
new class for projectRefSeq Changes

https://github.com/metasfresh/ip180/issues/210


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/document/sequenceno/ExternalProjectRefSequenceChanger.java
```diff
@@ -0,0 +1,40 @@
...
+
+		final String decimalPattern = documentSequenceInfo.getDecimalPattern();
+		if (decimalPattern == null) {
+			return incrementalSeqNo + "-" + customPart;
+		}
+
+		final int incrementalSeqNoInt = Integer.parseInt(incrementalSeqNo);
+
+		return new DecimalFormat(decimalPattern).format(incrementalSeqNoInt) + "-" + customPart;
+	}
```
cristinamghita, line 33: So this is the only thing different from `ExternalProjectRefSequenceNoProvider `, right?
Then pls refactor it to not duplicate the code.
And also rename it from `ExternalProjectRefSequenceChanger`. Is still a provider.
So maybe something like `ExternalProjectRefBeginingSequenceNoProvider` Or something similar, meaningful..

cristinamghita (changes requested):
see my comment

PR #15271, opened 2023-05-09
SAP FI interface: GL entries to be NOT interfaced into SAP

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/15260


backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/invoice/impl/JsonInvoiceService.java
```diff
@@ -839,6 +847,32 @@ private static void validateInvoiceTypeMatchesTrx(@NonNull final JsonCreateInvoi
...
+
+	@NonNull
+	private static InputDataSourceId getInputDataSourceId(
+			@NonNull final JsonCreateInvoiceRequestItemHeader invoiceHeader,
+			@NonNull final MasterdataProvider masterdataProvider,
+			@NonNull final OrgId orgId)
+	{
+		final String dataSourceInternalNameToUse = coalesce(
+				invoiceHeader.getDataSource(),
+				"int-" + DEFAULT_DATA_SOURCE_INTERNAL_NAME);
```
cp-ps, line 859: this is not necessarily an internal name, so just go with `dataSourceIdentifier`

PR #15306, opened 2023-05-12
Fix the REST-API method to PUT

metas-ts:
...to be in line with the API's expected method


metas-ts (commented):
lgtm, trivial fix, locally run unit tests are green => going to integrate

PR #15328, opened 2023-05-17
use OrgId from PurchaseCandidate instead of user's OrgId when processing purchase candidates into orders

adi-stefan:


backend/de.metas.purchasecandidate.base/src/main/java/de/metas/purchasecandidate/purchaseordercreation/remoteorder/RealVendorGatewayInvoker.java
```diff
@@ -154,6 +151,19 @@ public List<PurchaseItem> placeRemotePurchaseOrder(
 		return result.build();
 	}
 
+	@NonNull
+	private OrgId getOrgIdFromPurchaseCandidates(final Collection<PurchaseCandidate> purchaseCandidates)
+	{
+		final OrgId orgId = purchaseCandidates.stream()
+				.map(PurchaseCandidate::getOrgId)
+				.reduce((a, b) -> {
+					throw new AdempiereException("Can only process purchase candidates from a single organization.");
```
metas-ts, line 160: ok, FTR, here i wanted to write "please give more infos so that the user knows just *which* candidates this is about
but i guess that's clear, from the workpackage

backend/de.metas.purchasecandidate.base/src/main/java/de/metas/purchasecandidate/purchaseordercreation/remoteorder/VendorGatewayInvokerFactory.java
```diff
@@ -58,7 +57,6 @@ public VendorGatewayInvoker createForVendorId(@NonNull final BPartnerId vendorId
 		if (vendorGatewayService.isPresent())
 		{
 			return RealVendorGatewayInvoker.builder()
-					.orgId(orgId)
```
metas-ts, line 61: i can't comment at like 51..but don't we now still get the same error? shouldn't likes 50 - 56 just go away?

adi-stefan: obviously. removed.

Issue #15471, opened 2023-06-05
Add TaxCode to Accounting Transaction Details Window


chemsnouioua:
### Is this a bug or feature request?
FR

PR #15472, opened 2023-06-05
Gh15471 add tax code to accounting transaction details window

chemsnouioua:
#15471


metas-ts (commented):
Don't we need `AD_SQLColumn_SourceTableColumn` records so the webui can react to changes?

chemsnouioua:
> Don't we need `AD_SQLColumn_SourceTableColumn` records so the webui can react to changes?

IMHO, no need as this is not a SQL Column and also this  is a BD View  `Fact_Acct_Transactions_View`

teosarca (approved):

Issue #15452, opened 2023-06-01
Materialcockpit 404


metas-ts:

dspoelgen:
# :white_check_mark: Decision Request estimation changed 

🎬 Responsible 
- [ ] inform the Daily PO Members
Decision  
- [ ] case: **approved** +2h inkl. Test
  - [ ] coverage of costs
  - [ ] 👑 approved by customer (documented? Jour Fixe issue and in customer issue)
  - [ ] approved by metas (who?) Mark & Dominik 
  - [ ] current progress time ( 9.5 h)
  - [ ] 🛠 Dev informed 
  - [ ] set a new Bud label
  - [ ] remove Wait4BudgetApproval label 
- [ ] case: **unapproved**
  - [ ] close the issue
  - [ ] document the decision in Jour Fixe issue and in customer issue
  - [ ] set invoicible effort in instances
  - [ ] 🛠 Dev informed

PR #15483, opened 2023-06-07
#15452 prevent changing UOM of a used product

chemsnouioua:
#15452


metas-ts:
@MergifyIO copy release

PR #15514, opened 2023-06-07
Fix ClassCastException if picking process is called from HU editor

metas-ts:


metas-ts:
@Mergifyio copy adorable_badger_uat

PR #15539, opened 2023-06-09
Gh15471 fix display vatcode inplaveof taxcode

chemsnouioua:
#15471


cristinamghita (commented):
@chems, in the previous PR I think you generated models for I_Fact_Acct_Transactions_View.java
You shall do that again.

PR #15559, opened 2023-06-12
import ic date format change

mrie04:
change `invoice candidate import format` `date-formats` from `yyyy-mm-dd` to `dd.mm.yyyy` and extend info in error message for import table upserts


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/impexp/DataImportCommand.java
```diff
@@ -272,7 +272,7 @@ private PInstanceId getOrCreateRecordsToImportSelectionId()
 			_recordsToImportSelectionId = query.createSelection();
 			if (_recordsToImportSelectionId == null)
 			{
-				throw new AdempiereException("No records to import for " + query);
+				throw new AdempiereException("No records to import for " + query + " Errors on import: " + insertIntoImportTableResult.getErrors());
```
adi-stefan, line 275: I'm thinking if it would not make more sense to have just a list of (distinct) error messages shown here instead of the whole message+ lineNo + lineContent.
Reasoning:
1. Individual messages will still be shown on the ImportTable.I_ErrorMsg level.
2. The shown message is about providing feedback that an an error occured in the import process. Parsing a failure message for 3 import lines  in the brief time we show the message would be a challenge for virtually any user. Adding the line contents would make it even harder IMHO.

adi-stefan (changes requested):

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/impexp/DataImportCommand.java
mrie04: 1. When this error message is shown, there was no ImportTable record created (error on insert)
2. because of 1 we don't have access to Informations of the cause of the problem atm

will ask if we have some additional bud to polish it

adi-stefan: Then let's leave it as-is and hope it's understandable for the user.

adi-stefan (approved):

PR #15590, opened 2023-06-15
Fix dlm async queue archived tables

metas-ts:


metas-ts (commented):
the cucumber fails are not due to this SQL - in that case there would be also a test failure in jenkins

PR #15610, opened 2023-06-15
Make sure to update the Updated column when propagating TRL update to base table

cp-ps:


teosarca (approved):

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/i18n/po/POTrlRepository.java
```diff
@@ -276,6 +277,15 @@ public void onTrlRecordChanged(@NonNull final PO trlRecord)
 			return;
 		}
 
+		final boolean isUpdatedColumnPresent = Optional.ofNullable(POInfo.getPOInfo(baseTableName))
+				.map(poInfo -> poInfo.getColumnIndex("Updated") > -1)
+				.orElse(false);
+
+		if (isUpdatedColumnPresent)
+		{
+			sqlSet.append(", Updated").append("=").append(DB.TO_DATE(SystemTime.asTimestamp(), false));
```
teosarca, line 286: for the records, we have to do that because in `update_ad_element_on_ad_element_trl_update` we do:

image

metas-ts:
@Mergifyio copy release

PR #15656, opened 2023-06-21
New Flag in c_bpartner `isManual` and Import Format for Business Partner

fp-ps:


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/banking/api/BankRepository.java
```diff
@@ -160,4 +162,15 @@ public DataImportConfigId retrieveDataImportConfigIdForBank(@Nullable final Bank
...
+	public Set<BankId> retrieveBankIdsByName(final String bankName)
+	{
+		return queryBL.createQueryBuilder(I_C_Bank.class)
+				.addStringLikeFilter(I_C_Bank.COLUMNNAME_Name, bankName, false)
+				.addOnlyActiveRecordsFilter()
+				.iterateAndStream()
+				.map(I_C_Bank::getC_Bank_ID)
+				.map(BankId::ofRepoId)
+				.collect(ImmutableSet.toImmutableSet());
+	}
```
cp-ps, line 175: - doesn't make much sense to use iterateAndStream if we are anyway collecting all results in a Set,  you could use directly `stream()` in such scenarios
- in this particular case you can directly use `.listIds()`

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/impl/BPBankAccountDAO.java
```diff
@@ -140,6 +141,24 @@ else if (query.getInvoiceId() != null)
...
+			queryBuilder.addEqualsFilter(I_C_BP_BankAccount.COLUMNNAME_C_Bank_ID, query.getBankId());
+		}
+		if(Check.isNotBlank(query.getIban()))
+		{
+			queryBuilder.addStringLikeFilter(I_C_BP_BankAccount.COLUMNNAME_IBAN, query.getIban(), false);
+		}
+		if(Check.isNotBlank(query.getQrIban()))
+		{
+			queryBuilder.addStringLikeFilter(I_C_BP_BankAccount.COLUMNNAME_QR_IBAN, query.getQrIban(), false);
+		}
```
cp-ps, line 160: - fix formatting
- also, I'd suggest to use ignore case = true here as well

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/impl/BPBankAccountDAO.java
```diff
@@ -140,6 +141,24 @@ else if (query.getInvoiceId() != null)
 							.addEqualsFilter(I_C_Invoice.COLUMNNAME_C_Invoice_ID, query.getInvoiceId())
 							.create());
 		}
+
+		if (Check.isNotBlank(query.getDescription()))
+		{
+			queryBuilder.addStringLikeFilter(I_C_BP_BankAccount.COLUMNNAME_Description, query.getDescription(), true);
+		}
```
cp-ps, line 148: - I don't think it makes much sense to make this lookup, I mean, `description` is not reliable enough to do lookups

backend/de.metas.business/src/main/java/de/metas/bpartner/impexp/BPartnerBankAccountImportHelper.java
```diff
@@ -93,8 +100,101 @@ else if (!Check.isEmpty(importRecord.getSwiftCode(), true) && !Check.isEmpty(imp
 
 			importRecord.setC_BP_BankAccount_ID(bankAccount.getC_BP_BankAccount_ID());
 		}
+		else if (Check.isNotBlank(importRecord.getBankDetails()))
+		{
+			bankAccount = initializeBankAccountFromBankDetails(importRecord.getBankDetails());
+		}
+
```
cp-ps, line 107: - we should not create a bank account without at least an IBAN or QR-IBAN

backend/de.metas.business/src/main/java/de/metas/bpartner/impexp/BPartnerBankAccountImportHelper.java
```diff
@@ -93,8 +100,101 @@ else if (!Check.isEmpty(importRecord.getSwiftCode(), true) && !Check.isEmpty(imp
...
+				bankAccount.setName(bankDetails);
+			}
+
+			if (Check.isBlank(bankAccount.getDescription()))
+			{
+				bankAccount.setDescription(bankDetails);
+			}
+		}
 
 		return bankAccount;
```
cp-ps, line 181: - as mentioned before, we should not query by `bpartnerAccount.description`
- the way I see it, I would only use this method to locate a `C_Bank` by `bankDetails`, if one can be found, then propagate the C_Bank_ID to BPAccount

backend/de.metas.business/src/main/java/de/metas/bpartner/impexp/BPartnerBankAccountImportHelper.java
```diff
@@ -93,8 +100,101 @@ else if (!Check.isEmpty(importRecord.getSwiftCode(), true) && !Check.isEmpty(imp
...
+	}
+
+	@NonNull
+	private I_C_BP_BankAccount handleBPBankAccountCollection(final List<BPartnerBankAccount> bpBankAccountCollection)
+	{
+		return Optional.ofNullable(CollectionUtils.singleElementOrNull(bpBankAccountCollection))
+				.map(BPartnerBankAccount::getId)
+				.map(id -> InterfaceWrapperHelper.load(id, I_C_BP_BankAccount.class))
+				.orElseGet(() -> InterfaceWrapperHelper.newInstance(I_C_BP_BankAccount.class));
+	}
```
cp-ps, line 199: - I'd refactor to `getSingleBPBankAccount(): Optional<I_C_BP_BankAccount>` and keep the method only about loading the I_C_BP_BankAccount form the `List<BPartnerBankAccount>`
  - I'd actually remove this method all together and add a new method in `bankAccountDAO`
- another thing, we should not use `InterfaceWrapperHelper` to `load` or `create a new instance` outside of DAO/Repo classes, so pls go via `BankAccountDAO`

cp-ps (changes requested):

cp-ps (changes requested):
as discussed

cp-ps (approved):

PR #15657, opened 2023-06-21
Datefields in InvoiceCandidates

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/15653

### DIFF
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/acct/api/AccountDimension.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/acct/api/AccountDimension.java
@@ -22,17 +22,18 @@ package de.metas.acct.api;
  * #L%
  */
 
-import java.util.HashMap;
-import java.util.Map;
-
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
-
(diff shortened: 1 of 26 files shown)


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/acct/api/AccountDimension.java
```diff
@@ -256,15 +269,18 @@ public Builder setAlias(final String alias)
...
-			if(value instanceof Integer)
+			if (value instanceof Integer)
 			{
 				final int intValue = NumberUtils.asInt(value, 0);
 				segmentValues.put(segmentType, intValue);
 			}
+			else if (value instanceof Instant)
+			{
+				Optional.ofNullable(TimeUtil.asInstant(value))
+						.ifPresent(instantValue -> segmentValues.put(segmentType, instantValue));
```
cp-ps, line 282: no need to use optional here, if value is null => `value instanceof Instant` will be false

cp-ps: what we need instead, is to make sure we can handle the "null" string when reading dates from segmentValues

cp-ps (changes requested):

cp-ps (approved):

PR #15664, opened 2023-06-22
extend bpartner import

mrie04:
- add accountno, accountowner, lacationBPartnerName, locationName to BPartner Import

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/bpartner/impexp/BPartnerBankAccountImportHelper.java
+++ b/backend/de.metas.business/src/main/java/de/metas/bpartner/impexp/BPartnerBankAccountImportHelper.java
@@ -3,6 +3,7 @@ package de.metas.bpartner.impexp;
 import de.metas.banking.BankAccountId;
 import de.metas.banking.BankId;
 import de.metas.banking.api.BankRepository;
+import de.metas.common.util.CoalesceUtil;
 import de.metas.currency.ICurrencyBL;
 import de.metas.impexp.processing.IImportInterceptor;
 import de.metas.invoice_gateway.spi.model.BPartnerId;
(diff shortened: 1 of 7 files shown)


backend/de.metas.business/src/main/java/de/metas/bpartner/impexp/BPartnerBankAccountImportHelper.java
```diff
@@ -71,6 +69,9 @@ public I_C_BP_BankAccount importRecord(final I_I_BPartner importRecord)
 		if (bankAccount != null)
 		{
 			bankAccount.setIBAN(importRecord.getIBAN());
+			final String accountName = importRecord.getA_Name();
+			bankAccount.setA_Name(Check.isEmpty(accountName) ? importRecord.getSwiftCode() : accountName);
```
adi-stefan, line 73: Why not coalescing accountName, SwiftCode, then bpartnerName (swiftCode is also nullable, and bpartnerName is also used as value, see de.metas.payment.api.impl.QRPaymentStringDataProvider#createNewC_BP_BankAccount).

backend/de.metas.business/src/main/java/de/metas/bpartner/impexp/BPartnerBankAccountImportHelper.java
```diff
@@ -79,7 +80,9 @@ else if (!Check.isEmpty(importRecord.getSwiftCode(), true) && !Check.isEmpty(imp
 			bankAccount = InterfaceWrapperHelper.newInstance(I_C_BP_BankAccount.class);
 			bankAccount.setC_BPartner_ID(bpartnerId.getRepoId());
 			bankAccount.setIBAN(importRecord.getIBAN());
-			bankAccount.setA_Name(importRecord.getSwiftCode());
+			final String accountName = importRecord.getA_Name();
+			bankAccount.setA_Name(Check.isEmpty(accountName) ? importRecord.getSwiftCode() : accountName);
```
adi-stefan, line 84: same as above, though here the SwiftCode is not-null.

adi-stefan (changes requested):
See above.

backend/de.metas.business/src/main/java/de/metas/bpartner/impexp/BPartnerBankAccountImportHelper.java
mrie04: In german this field is labeled as `Name auf Kreditkarte` (Name on CreditCard). This is why i didn't do coalescing. I don't know why Swift-Code was used here.
What do you think about using bppartner.name as fallback instead of Swift-Code?

adi-stefan: I see the bpartner name used in more recent code, so let's go with that then.

adi-stefan (approved):

PR #15677, opened 2023-06-23
Invoice.DueDate: computed from ActualLoadingDate instead of PlannedLoadingDate

cp-ps:


cristinamghita (approved):
lgtm

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/invoicecandidate/DeliveryPlanningICHandlerExtension.java
```diff
@@ -0,0 +1,46 @@
...
+@Component
+@RequiredArgsConstructor
+public class DeliveryPlanningICHandlerExtension implements OrderLineHandlerExtension
+{
+	@NonNull
+	private final DeliveryPlanningService deliveryPlanningService;
+
+	@Override
+	public void setDeliveryRelatedData(@NonNull final OrderLineId orderLineId, @NonNull final I_C_Invoice_Candidate invoiceCandidate)
+	{
```
cristinamghita, line 42: I think you shall set it to null you did not find any Delivery instruction

PR #15689, opened 2023-06-23
Module Contract tables and columns renaming

metas-rc:


adi-stefan (dismissed):
Lgtm, but I just remembered: did we decide on keeping the (harvest) Year on the Settings level, or did we decide to move it to an intermediate entity, PLV style?

backend/de.metas.adempiere.adempiere/base/src/main/java-gen/org/compiere/model/X_Contract_Module_Log.java
```diff
@@ -1,246 +0,0 @@
-// Generated Model - DO NOT CHANGE
```
adi-stefan, line 1: Shouldn't you also remove the interfaces as well?

adi-stefan (changes requested):
Mostly lgtm, but I just remembered: did we decide on keeping the (harvest) Year on the Settings level, or did we decide to move it to an intermediate entity, PLV style?

backend/de.metas.adempiere.adempiere/base/src/main/java-gen/org/compiere/model/X_Contract_Module_Log.java
metas-ts: yes. lemmi check, if was thinking i did that

metas-ts:
> Mostly lgtm, but I just remembered: did we decide on keeping the (harvest) Year on the Settings level, or did we decide to move it to an intermediate entity, PLV style?

yes we did. but it's AFAIR part of @cristinamghita 's concept

adi-stefan:
> > Mostly lgtm, but I just remembered: did we decide on keeping the (harvest) Year on the Settings level, or did we decide to move it to an intermediate entity, PLV style?
> 
> yes we did. but it's AFAIR part of @cristinamghita 's concept

Then just removing the interface classes and it will lgtm.

cristinamghita:
> > Mostly lgtm, but I just remembered: did we decide on keeping the (harvest) Year on the Settings level, or did we decide to move it to an intermediate entity, PLV style?
> 
> yes we did. but it's AFAIR part of @cristinamghita 's concept

@metas-ts (cc @adi-stefan ) yes, we decide it move it.
Is one of my ACs.
See AC3 here: https://github.com/metasfresh/me03/issues/16108#issuecomment-1600621986

adi-stefan (approved):

PR #15697, opened 2023-06-26
country based docno tax departure country

mrie04:
- Sales Invoice window tax departure country
- country based docNo tax departure country instead of billTo country
- country based docNo cucumber


adi-stefan (commented):
Minor nitpick on the constant, otherwise looks ok.

adi-stefan (approved):

PR #15695, opened 2023-06-25
15694 fenster bestellanforderung iteration 2

chemsnouioua:
#15694

Issue #15694, opened 2023-06-25
Fenster Bestellanforderung - Iteration 2


chemsnouioua:
### Is this a bug or feature request?
FR

PR #15763, opened 2023-06-30
fix junit requisition

mrie04:
https://github.com/metasfresh/metasfresh/pull/15695
https://github.com/metasfresh/metasfresh/issues/15694


cristinamghita (approved):
lgtm

PR #15805, opened 2023-07-05
DeliveryPlanningICHandlerExtension: ActLoadDate from Planning docs with no Instructions

cp-ps:
DeliveryPlanningICHandlerExtension: consider delivery plannings without instructions in case of soTrx = Purchase


cristinamghita (approved):
lgtm

PR #15836, opened 2023-07-06
Inner silence uat gh15819

dm-ps:
https://github.com/metasfresh/metasfresh/issues/15819


backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/impl/InventoryLineModularContractHandler.java
```diff
@@ -0,0 +1,191 @@
...
+		}
+
+		final Optional<ModularContractTypeId> modularContractTypeId = modularContractSettings.getModuleConfigs()
+				.stream()
+				.filter(config -> config.isMatchingClassName(InventoryLineModularContractHandler.class.getName()))
+				.map(ModuleConfig::getModularContractType)
+				.map(ModularContractType::getId)
+				.findFirst();
+
+		final Quantity quantity = inventoryBL.getMovementQty(inventoryLine);
```
metas-ts, line 123: **💡 can be done later**
if modularContractTypeId is empty, we can spare ourselves the additional work and return Optional.empty() right here.
This is also clearer IMO, and we can avoid the `modularContractTypeId.map` down the road, which is once again clearer

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/interceptor/M_InventoryLine.java
```diff
@@ -0,0 +1,58 @@
...
+	@ModelChange(timings = ModelValidator.TYPE_BEFORE_CHANGE, ifColumnsChanged = I_M_InventoryLine.COLUMNNAME_Modular_Flatrate_Term_ID)
+	public void validateModularContract(@NonNull final I_M_InventoryLine inventoryLineRecord)
+	{
+		final FlatrateTermId modularContractId = FlatrateTermId.ofRepoIdOrNull(inventoryLineRecord.getModular_Flatrate_Term_ID());
+		if (modularContractId == null)
+		{
+			return;
+		}
+
+		if (!flatrateBL.isModularContract(modularContractId))
```
metas-ts, line 53: **💡 can be done later**

unless it's already the case we need 
- either a validation rule so the user can select only modular contracts
- or a user-friendly error-message

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/impl/InventoryLineModularContractHandler.java
```diff
@@ -0,0 +1,191 @@
...
+		final InventoryId reversalId = InventoryId.ofRepoIdOrNull(inventory.getReversal_ID());
+		if (reversalId != null)
+		{
+			return Optional.empty();
+		}
+
+		return Optional.of(LogEntryReverseRequest.builder()
+								   .referencedModel(TableRecordReference.of(inventoryLine))
+								   .flatrateTermId(modularContractId)
+								   .build());
```
metas-rc, line 163: 💡 can be done later

I think we should provide a different description for the reversal log line, to make it obvious this inventory was reversed.

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/log/ModularContractLogDAO.java
```diff
@@ -156,7 +156,10 @@ public ModularContractLogEntryId reverse(@NonNull final LogEntryReverseRequest r
 			reversedLog.setAmount(reversedLog.getAmount().negate());
 		}
 
-		reversedLog.setDescription(request.description());
+		if (request.description() != null)
+		{
+			reversedLog.setDescription(request.description());
```
metas-rc, line 161: 💡 can be done later

Again, I think the reverse description should be different, to make it obvious we are dealing with a reversal.

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/interceptor/M_InventoryLine.java
dm-ps: A validation rule so that the user can select only modular contracts is already in place

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/impl/InventoryLineModularContractHandler.java
dm-ps: The used description is `A shortage/excess stock of {0} was counted during inventory.`, and the parameter is always positive (so that it is applicable in both cases - complete and reverse). Shall we change the description and use a different one?

metas-rc: I see your point regarding the description. I am not sure how detailed our customer would like that description to be but I have the feeling they would like to see if the document was reversed. I can bring it up in a future meeting with Mark and will document my findings in another issue.
For now, let's just leave it like it is

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/log/ModularContractLogDAO.java
metas-rc: see above https://github.com/metasfresh/metasfresh/pull/15836#discussion_r1255408642

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/impl/InventoryLineModularContractHandler.java
dm-ps: done

cp-ps (approved):

PR #15902, opened 2023-07-18
print via externalsystem store as file

mrie04:


misc/de-metas-common/de-metas-common-externalsystem/src/main/java/de/metas/common/externalsystem/ExternalSystemConstants.java
```diff
@@ -30,7 +30,7 @@ public class ExternalSystemConstants
 	public static final String PARAM_API_KEY = "APIKey";
 	public static final String PARAM_BASE_PATH = "BasePath";
 	public static final String PARAM_TENANT = "Tenant";
-	public static final String PARAM_BPARTNER_GROUP = "BPartnerGroup";
```
metas-ts, line 33: (also not used in `master` branch :+1:)

metas-ts (approved):
i took some notes for the follow-up issue though

Issue #15920, opened 2023-07-20
`Create PO from Requisition` process `Trx name` problem 


riadmen:
### Is this a bug or feature request?
BUG
### What is the expected or desired behavior?

<img width="158" alt="image" src="https://github.com/metasfresh/metasfresh/assets/93593841/d2fb1a3a-e2a2-47c3-9f87-8b7ba727d873">

PR #15922, opened 2023-07-20
Trx name problem

riadmen:
https://github.com/metasfresh/metasfresh/issues/15920

### DIFF
--- a/backend/de.metas.business/src/main/java-legacy/org/compiere/process/RequisitionPOCreate.java
+++ b/backend/de.metas.business/src/main/java-legacy/org/compiere/process/RequisitionPOCreate.java
@@ -56,56 +56,86 @@ import java.util.List;
 
 /**
  * Create PO from Requisition
- * 
- * 
+ *
  * @author Jorg Janke
- * @version $Id: RequisitionPOCreate.java,v 1.2 2006/07/30 00:51:01 jjanke Exp $
- * 
  * @author Teo Sarca, www.arhipac.ro
- *         <li>BF [ 2609760 ] RequisitionPOCreate not using DateRequired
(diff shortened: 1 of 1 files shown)


backend/de.metas.business/src/main/java-legacy/org/compiere/process/RequisitionPOCreate.java
```diff
@@ -543,9 +543,9 @@ private boolean isVendorForProduct(final int C_BPartner_ID, final MProduct produ
 
 	}
 
-	private boolean isPOPriceList(final int M_PriceList_ID)
+	private boolean isPOPriceList(final int M_PriceList_ID, final I_M_Requisition requisition)
 	{
-		return Services.get(IQueryBL.class).createQueryBuilder(I_M_PriceList.class, getProcessInfo())
+		return Services.get(IQueryBL.class).createQueryBuilder(I_M_PriceList.class, requisition)
 				.addEqualsFilter(I_M_PriceList.COLUMNNAME_M_PriceList_ID, M_PriceList_ID)
```
cristinamghita, line 549: Please take out `Services.get(IQueryBL.class) `to the top and adjust the entire class.
The, use the query builder without giving it any object as a parameter. Shall get the trx from the thread if you use that.
So you do not need to give a new parameter only for the transaction.

cristinamghita (changes requested):
see my comment

backend/de.metas.business/src/main/java-legacy/org/compiere/process/RequisitionPOCreate.java
riadmen: modification done

metas-rc (approved):

cristinamghita (approved):

PR #15941, opened 2023-07-25
Extend process M_HU_Report_QRCode to support several jasper processes

cristinamghita:


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/IADProcessDAO.java
```diff
@@ -127,4 +127,6 @@ default void registerTableProcess(final String tableName, final AdProcessId adPr
 	void updateColumnNameByAdElementId(
 			@NonNull AdElementId adElementId,
 			@Nullable String newColumnName);
+
+	boolean isJasperJSONProcess(@NonNull AdProcessId processId);
 }
```
metas-ts, line 132: i think it would be better to have retrieveProcessType that returns the enum. otherwise the next person will add a method isJasperSQLProcess and so on..

backend/de.metas.global_qrcode.services/src/main/java/de/metas/global_qrcodes/service/CreatePDFCommand.java
```diff
@@ -21,31 +22,53 @@
...
 
 public class CreatePDFCommand
 {
-	private static final AdProcessId qrCodeProcessId = AdProcessId.ofRepoId(584977); // hard coded process id
+
 	private final IADPInstanceDAO adPInstanceDAO = Services.get(IADPInstanceDAO.class);
+	private final IADProcessDAO adProcessDAO = Services.get(IADProcessDAO.class);
 	private final ImmutableList<PrintableQRCode> qrCodes;
+	private final ImmutableList<ProcessInfoParameter> processParams;
+	private AdProcessId qrCodeProcessId ;
```
metas-ts, line 36: make `final`

backend/de.metas.handlingunits.base/src/main/sql/postgresql/ddl/functions/hu_label.sql
```diff
@@ -0,0 +1,68 @@
...
+    /** receipt infos */
+         LEFT OUTER JOIN (
+    SELECT DISTINCT ON (tu.M_HU_ID) tu.M_HU_ID,
+                                    bp.name         AS vendorName,
+                                    o.DocumentNo    AS PurchaseOrderNo,
+                                    io.movementdate AS receiptdate
+    FROM M_HU tu
+             LEFT OUTER JOIN M_HU_Item lui ON lui.M_HU_Item_ID = tu.M_HU_Item_Parent_ID AND lui.isActive = 'Y'
+             LEFT OUTER JOIN M_HU lu ON lu.M_HU_ID = lui.M_HU_ID
+             LEFT OUTER JOIN M_HU_PI_Version piv
```
metas-ts, line 52: if we print the label to a random HU - why would we *not*  want to print the infos from the  the HU's `M_HU_PI_Version` if that piv is inactive? it's not like the HU has a better, more recent one, right?

metas-ts (changes requested):
minor requests..

metas-ts (approved):
thx, looks like this *also* addresses the additional concerns i wrote after the first review 
But I didn't check the userfriendlyness of the IsDirectPrint) param.

The user should know what happens if it's ticked (=> PDF send to metasfresh direct printing framework) vs not ticked (PDF downloaded/opened in the briowser)

⚡ Also i think in the  intensive-care-custom repo IsDirectPrint should be true by default, because that's the vast majority of cases

PR #15958, opened 2023-07-27
Import Bank Accounts

cp-ps:
- import bank accounts via standard file import (tsv, DataImportConfig, import window & process)
- add bank account id to generate purchase invoice process

refs: https://github.com/metasfresh/metasfresh/issues/15957


backend/de.metas.invoice.webui/src/main/java/de/metas/invoicecandidate/process/C_Invoice_Candidate_EnqueueSelectionForInvoicing.java
```diff
@@ -188,6 +206,32 @@ public LookupValuesList getAvailableForexContracts()
 				: LookupValuesList.EMPTY;
 	}
 
+	@ProcessParamLookupValuesProvider(parameterName = InvoicingParams.PARA_BP_BankAccountId, numericKey = true, lookupSource = DocumentLayoutElementFieldDescriptor.LookupSource.lookup)
+	public LookupValuesList getAvailableBankAccountIds(@NonNull final LookupDataSourceContext context)
```
metas-ts, line 210: won't this thrown an exception if `getInvoicingParams().isInvoicingSingleBPartner()==false` and the user tries to invoice multiple partners?

cp-ps: nope, only if the user tries to fill in a Bank Account, which is not supported for multiple BPs

cp-ps: also, the Bank Account parameter is not displayed in standard metasfresh

metas-ts (approved):

Issue #15975, opened 2023-07-28
`Shipment` -> `Contract Module Log`


riadmen:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?
`Shipment`-> `contract module log` relation implementation

PR #15976, opened 2023-07-28
Shipment-> contract module log relation implementation

riadmen:
https://github.com/metasfresh/metasfresh/issues/15975

### DIFF
--- a/backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/impl/ShipmentLineModularContractHandler.java
+++ b/backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/impl/ShipmentLineModularContractHandler.java
@@ -41,6 +41,8 @@ import de.metas.contracts.modular.settings.ModularContractType;
 import de.metas.contracts.modular.settings.ModularContractTypeId;
 import de.metas.contracts.modular.settings.ModuleConfig;
 import de.metas.i18n.AdMessageKey;
+import de.metas.i18n.IMsgBL;
+import de.metas.i18n.ITranslatableString;
 import de.metas.inout.IInOutDAO;
 import de.metas.inout.InOutId;
 import de.metas.lang.SOTrx;
@@ -63,10 +65,12 @@ import org.compiere.model.I_C_Order;
 import org.compiere.model.I_C_UOM;
 import org.compiere.model.I_M_InOut;
 import org.compiere.model.I_M_InOutLine;
+import org.compiere.util.Env;
 import org.springframework.stereotype.Component;
 
 import java.util.Optional;
 import java.util.stream.Stream;
+
 @Component
 public class ShipmentLineModularContractHandler implements IModularContractTypeHandler<I_M_InOutLine>
 {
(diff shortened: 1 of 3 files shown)


backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/impl/ShipmentLineModularContractHandler.java
```diff
@@ -0,0 +1,190 @@
...
+				.build());
+	}
+
+	@Override
+	public @NonNull Optional<LogEntryReverseRequest> createLogEntryReverseRequest(final @NonNull I_M_InOutLine inOutLineRecord, final @NonNull FlatrateTermId flatrateTermId)
+	{
+		return Optional.of(LogEntryReverseRequest.builder()
+								   .referencedModel(TableRecordReference.of(I_M_InOutLine.Table_Name, inOutLineRecord.getM_InOutLine_ID()))
+								   .flatrateTermId(flatrateTermId)
+								   .description("Shipment reversed")
```
metas-ts, line 150: no hardcoded description

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/impl/ShipmentLineModularContractHandler.java
```diff
@@ -0,0 +1,190 @@
...
+								   .build());
+	}
+
+	@Override
+	public @NonNull Stream<FlatrateTermId> streamContractIds(@NonNull final I_M_InOutLine inOutLineRecord)
+	{
+		final I_C_Order order = orderBL.getById(OrderId.ofRepoId(inOutLineRecord.getC_Order_ID()));
```
metas-ts, line 157: don't we also need to check for SOTrx?

metas-ts (commented):
(only took a fleeting glimpse for now)

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/impl/ShipmentLineModularContractHandler.java
```diff
@@ -0,0 +1,190 @@
...
+
+		final WarehouseId warehouseId = WarehouseId.ofRepoIdOrNull(order.getM_Warehouse_ID());
+		assert warehouseId != null;
```
metas-ts, line 160: we don't use java-1.4. asserts, i don't even know that they are (always) swtiched on. 
Since order.getM_Warehouse_ID is a mandatory column, please just use `WarehouseId.ofRepoId(....)`

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/impl/ShipmentLineModularContractHandler.java
```diff
@@ -0,0 +1,190 @@
...
+
+		final YearId harvestingYearId = YearId.ofRepoIdOrNull(order.getHarvesting_Year_ID());
```
metas-ts, line 162: Why want to assert that order.getHarvestingYear_ID is > 0 ?
AFAIU it might be null, n which case we don'T expect an assertion-error, but simply no FlatrateTerm to be returned..AFAIU

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/impl/ShipmentLineModularContractHandler.java
metas-ts, line 157: also note, that the inoutLine simply might not have an order..AFAIU there simply shouldn't be a FlatrateTermId to be returned, but no error

metas-ts: LE: or the applies method should not apply..

PR #15982, opened 2023-07-31
local currency conversion needs to look at tax departure country

mi-ps:
metasfresh issue : https://github.com/metasfresh/metasfresh/issues/15980


mrie04 (approved):
LGTM!

PR #16013, opened 2023-08-03
improve DESADV matching with BPartnerId

mrie04:


metas-ts (commented):
lgtm, pls just add that migration-SQL here..

metas-ts (approved):

PR #16053, opened 2023-08-14
Product search (filter) - name

mi-ps:
metasfresh issue : https://github.com/metasfresh/metasfresh/issues/16052

### DIFF
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/product/ProductId.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/product/ProductId.java
@@ -1,0 +1,107 @@
+/*
+ * #%L
+ * de.metas.adempiere.adempiere.base
+ * %%
+ * Copyright (C) 2023 metas GmbH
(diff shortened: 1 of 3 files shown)


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/product/ProductId.java
```diff
@@ -0,0 +1,107 @@
...
+import org.adempiere.util.lang.impl.TableRecordReference;
+
+import javax.annotation.Nullable;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+
+@Value
+public class ProductId implements RepoIdAware
```
metas-ts, line 41: if you press `shift-shift` and enter `ProductId`, you'll find that we already have a class with the same classname. It's not allowed to have the same classname twice.
If you need ProductId in adempiere.base, you need to *move*  (not  *copy*) it there

metas-ts (approved):

PR #16099, opened 2023-08-25
Master with entropy

cp-ps:


mpeelen (changes requested):
The file `docker-builds/version.info` should not get overwritten, pls make sure the version does not change. In this case `5.177` is the correct version for master.

The other pipeline changes are from my pov fine to have in 'master'.

PR #16127, opened 2023-08-30
Improve ESR-Performance and fix QuerySelectionToDeleteHelper

metas-ts:
**ESR-Performance**
- cherry-pick https://github.com/metasfresh/metasfresh/pull/12946
- as an additional fix, don't even look for `ESR_ImportLine`s if there are no similar payments

**QuerySelectionToDeleteHelper**
Do not enqueue an UUID for deletion if it is already enqueued or if there are no corresponding `T_Query_Selection` or `T_Query_Selection_Pagination` records
We need this because `GuaranteedPOBufferedIterator#finalize` invokes the schedule method a lot and might inserts 100Ks of record that should not be there

**`5618160_sys_gh12070_Add_new_additives_window.sql`**
Minor fix to avoid a DB-deadlock down the line

-------------

Refs: https://github.com/metasfresh/metasfresh/issues/16128


mrie04:
@Mergifyio copy yoyo_uat

PR #16135, opened 2023-09-01
fixx collection point bpartner

mi-ps:


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/modular/workpackage/PPCostCollectorLogHandler.java
```diff
@@ -99,6 +100,8 @@ public LogAction getLogAction(@NonNull final HandleLogsRequest<I_PP_Cost_Collect
 		final ModularContractSettings modularContractSettings = request.getModularContractSettings();
 
 		final I_C_Flatrate_Term modularContractRecord = flatrateDAO.getById(contractId);
+		final I_C_Flatrate_Data flatrateDataRecord = flatrateDAO.retrieveFlatrateData(modularContractRecord);
+
```
metas-rc, line 104: why not using

de.metas.contracts.impl.FlatrateDAO#getById(de.metas.contracts.FlatrateTermId) to extract the flatrate term and then take the C_Bpartner directly from it?

Issue #16164, opened 2023-09-06
Update the `de_metas_endcustomer_fresh_reports.hasFreeEUTaxes` Function


riadmen:
### Is this a bug or feature request?
BUG
### What is the expected or desired behavior?

 The `IsToEULocation` column is ***Deprecated*** and must be replaced by `TypeOfDestCountry`

PR #16165, opened 2023-09-06
Update the de_metas_endcustomer_fresh_reports.hasFreeEUTaxes Function

riadmen:
https://github.com/metasfresh/metasfresh/issues/16164


backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/70-de.metas.fresh/5701720_sys_gh16164_C_Invoice_V.sql
```diff
@@ -0,0 +1,47 @@
+CREATE VIEW de_metas_endcustomer_fresh_reports.c_invoice_v
+            (
+             value,
```
cristinamghita, line 3: Why are you adjusting this?

cristinamghita (changes requested):
See my question.

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/70-de.metas.fresh/5701720_sys_gh16164_C_Invoice_V.sql
riadmen: `c_invoice_v` depend on `de_metas_endcustomer_fresh_reports.hasFreeEUTaxes` so i need to drop this view before updating the function, and then create it egain

cristinamghita: @riadmen but there is no drop. You just recreate the view.
Pls, do that is the same file with function:
Drop the view, drop the function, recreate the function, recreate the view.

cristinamghita (changes requested):
Check my comment

cristinamghita (approved):
lgtm

PR #16290, opened 2023-09-27
Fix saldobilanz_report to consider first day of fiscal year (as opposed to first day of calendar year) for year-to-date calculations

adi-stefan:


cristinamghita (approved):
lgtm

PR #16298, opened 2023-09-28
Fix saldobilanz_report to consider first day of previous fiscal year (as opposed to first day of current fiscal year year) for previous year calculations

adi-stefan:


cristinamghita (approved):
lgtm

PR #16332, opened 2023-10-02
Modular Contract: proForma Sales Order

mrie04:


backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
```diff
@@ -363,10 +363,13 @@ public void update_C_Invoice_Candidate(@NonNull final DataTable dataTable)
 	}
 
 	@And("^there is no C_Invoice_Candidate for C_Order (.*)$")
-	public void validate_no_C_Invoice_Candidate_created(@NonNull final String orderIdentifier)
+	public void validate_no_C_Invoice_Candidate_created(@NonNull final String orderIdentifier) throws InterruptedException
 	{
 		final I_C_Order order = orderTable.get(orderIdentifier);
 
+		// give IC handlers some time
+		Thread.sleep(1000);
```
adi-stefan, line 371: Rather than waiting a fixed amount, please use StepDefUtil.tryAndWait(maxWaitSeconds, checkingIntervalMs, worker) instead.
Waiting for a fixed amount is a recipe for either long runtimes or flaky tests, because the the processes on which you're waiting did not have enough time to do their expected job.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
```diff
@@ -377,6 +380,24 @@ public void validate_no_C_Invoice_Candidate_created(@NonNull final String orderI
 		Assertions.assertThat(invoiceCandidates).isEmpty();
 	}
 
+	@And("^there is no C_Invoice_Candidate for M_InOutLine (.*)$")
+	public void validate_no_C_Invoice_Candidate_created_for_InOutLine(@NonNull final String inOutLineIdentifier) throws InterruptedException
+	{
+		final I_M_InOutLine inOutLineRecord = inoutLineTable.get(inOutLineIdentifier);
+
+		// give IC handlers some time
+		Thread.sleep(1000);
```
adi-stefan, line 389: Same as above.

adi-stefan (changes requested):
please see comments

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
mrie04: As talked: `StepDefUtil.tryAndWait` doesn't work because we wait and check for something NOT created. We couldn't find a better solution, so we stay with this for now.

Note: We're waiting for IC-handlers that create missing ICs and are called from ModelInterceptors

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
mrie04: same as above

adi-stefan (approved):

Issue #16558, opened 2023-10-27
Use the price precision in jasper reports


riadmen:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?

- Use the price precision taken from the Price list in jasper reports

PR #16559, opened 2023-10-27
Use the price precision in jasper reports

riadmen:
https://github.com/metasfresh/metasfresh/issues/16558


cristinamghita (changes requested):
check my comment

cristinamghita (approved):
lgtm

PR #16579, opened 2023-11-01
Consider the number of matching attributes in product proposal when deciding which is the linked orderline

metas-rc:


adi-stefan (approved):

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/order/products_proposal/service/OrderLine.java
```diff
@@ -76,5 +77,4 @@ boolean isMatching(
 				HUPIItemProductId.nullToVirtual(packingMaterialId))
 				&& asiMatcher.test(this);
```
adi-stefan, line 78: Rather than this, why not use: `org.adempiere.mm.attributes.api.AttributesKeys#createAttributesKeyFromASIStorageAttributes` to extract the `AttributeKeys` and then just `Objects.equals(attributeKey1, attributeKey2)` ?

PR #16669, opened 2023-11-14
Reactivate queue processors

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/16660


metas-rc (approved):
Approving just to have the master green again

PR #16792, opened 2023-11-24
Fix UOM-conversion bug when reopening orderline with TU-UOM

metas-ts:


backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
```diff
@@ -822,9 +820,7 @@ public void reopenLine(@NonNull final org.compiere.model.I_C_OrderLine orderLine
 	{
 		//
 		// Calculate QtyOrdered as QtyEntered converted to stocking UOM
-		final ProductId productId = ProductId.ofRepoId(orderLine.getM_Product_ID());
-		final Quantity qtyEntered = orderLineBL.getQtyEntered(orderLine);
-		final Quantity qtyOrdered = uomConversionBL.convertToProductUOM(qtyEntered, productId);
+		final Quantity qtyOrdered = orderLineBL.convertQtyEnteredToStockUOM(orderLine);
```
metas-ts, line 823: this is the actual fix

metas-ts:
@mergifyio copy release intensive_care_uat

PR #16850, opened 2023-12-04
Zählbestand Reihenfolge is not transferred to Material Cockpit

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/16791


metas-ts (approved):
LGTM

Issue #16879, opened 2023-12-06
Improve Base Price List Validation


riadmen:
### Is this a bug or feature request?
BUG
### What is the expected or desired behavior?

- Improve Base Price List Validation

PR #16880, opened 2023-12-06
Scary plum uat gh16879

riadmen:
https://github.com/metasfresh/metasfresh/issues/16879


mrie04 (changes requested):
Change lgtm, but should be implemented in soft_panda and in IMHO soft_panda_hotfix

Issue #16995, opened 2023-12-20
Update the font in jasper report to `Arial`


riadmen:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?

 - Update the font in jasper report to `Arial`

PR #16996, opened 2023-12-20
Update the font in jasper report to Arial

riadmen:
https://github.com/metasfresh/metasfresh/issues/16995


cristinamghita (approved):
lgtm

PR #17049, opened 2024-01-02
play with jetbrains Contract annotation to avoid pointless warnings

metas-ts:
E.g. avoid a warning saying that a parameter might be null although Check.isNotBlank returned false for that parameter


metas-ts:

PR #17069, opened 2024-01-09
Inner silence uat material cockpit warehouse (PROTO)

metas-rc:


backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/MaterialCockpitRow.java
```diff
@@ -387,6 +399,7 @@ private MaterialCockpitRow(
 			final Quantity qtyStockCurrentAtDate,
 			final Quantity qtyOnHandStock,
 			@NonNull final ProductId productId,
+			final WarehouseId warehouseId,
```
metas-ts, line 402: also `@Nullable` ?

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/MaterialCockpitRowsLoader.java
```diff
@@ -100,15 +103,20 @@ public List<MaterialCockpitRow> getMaterialCockpitRows(
 		final List<I_QtyDemand_QtySupply_V> quantitiesRecords = QtyDemandSupplyFilters
 				.createQuantitiesQueryFor(filters)
 				.list();
+
+		final List<WarehouseId> warehouseIds = quantitiesRecords.stream()
+				.map(qtyRecord -> WarehouseId.ofRepoId(qtyRecord.getM_Warehouse_ID()))
```
metas-ts, line 108: do we know that `M_Warehouse_ID` is never `null`? if yes, its AD_Column could be flagged as mandatory

metas-ts (commented):
the code looks legit to me

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/MaterialCockpitRowsLoader.java
metas-rc: no longer needed in the end

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/MaterialCockpitRow.java
metas-rc: no longer needed in the end

metas-ts (approved):

PR #17092, opened 2024-01-10
Fix migration script so that we can roll out

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/17088


mrie04:
@Mergifyio copy hard_encoded_hotfix hard_encoded_uat

PR #17108, opened 2024-01-11
add index for better perf if there are many unprocessed workpackages

metas-ts:


metas-ts:
@MergifyIO copy release_5.175

mrie04:
@Mergifyio copy inner_silence_uat

PR #17115, opened 2024-01-12
Cherry-Pick fix against NPE

metas-ts:


metas-ts:
@MergifyIO copy release_5.175

PR #17120, opened 2024-01-12
fix existing older Bill_User_IDs in c_invoice_candidate

metas-ts:


metas-ts:
@MergifyIO copy release_5.175

PR #17217, opened 2024-01-31
Fix Manufacturing Order -> Maturing Candidate zoom relations
Made some fields on Maturing Candidate window read-only

adi-stefan:
https://github.com/metasfresh/metasfresh/issues/17194


cristinamghita (approved):
lgtm

PR #17238, opened 2024-02-01
Do no return Spare Part as BOM Type

cristinamghita:


metas-ts:
@mergifyio copy release

PR #17291, opened 2024-02-06
Fix AD_PrinterRouting matching and NumberOfCopies

metas-ts:
## Fix AD_PrinterRouting matching

Without this fix, the routings that **I** care about were never picked:

<details>
<summary>📷 Screenshot </summary>

image
</details>

I could have given them a lower SeqNo, but then the "Doctype"-Routings would never have been picked

## NumberOfCopies

Without this fix, `C_Printing_Queue.Copies` for records created by `M_HU_Report_QRCode.java` was always 1, no matter the process parameter value.


metas-ts:
@mergifyio copy release

Issue #17366, opened 2024-02-13
`SetUp Wizard` must change currency in `Dunning Type`


riadmen:
### Is this a bug or feature request?
BUG
### What is the expected or desired behavior?

- `SetUp Wizard` process does not change currency in dunning type

PR #17367, opened 2024-02-13
`SetUp Wizard` must change currency in `Dunning Type`

riadmen:
https://github.com/metasfresh/metasfresh/issues/17366

### DIFF
--- a/backend/de.metas.fresh/de.metas.fresh.base/src/main/java/de/metas/fresh/setup/process/ClientSetup.java
+++ b/backend/de.metas.fresh/de.metas.fresh.base/src/main/java/de/metas/fresh/setup/process/ClientSetup.java
@@ -271,7 +271,7 @@ class ClientSetup
 
 		priceList_None.setC_Currency_ID(currencyId.getRepoId());
 
-		dunningDAO.retrieveDunnings().forEach(dunning -> {
+		dunningDAO.retrieveDunnings().stream().filter(dunning -> dunning.getAD_Org_ID() == adOrg.getAD_Org_ID()).forEach(dunning -> {
 			dunning.setC_Currency_ID(currencyId.getRepoId());
 			dunningDAO.save(dunning);
 		});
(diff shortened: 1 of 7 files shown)


backend/de.metas.fresh/de.metas.fresh.base/src/main/java/de/metas/fresh/setup/process/ClientSetup.java
```diff
@@ -269,6 +271,11 @@ public ClientSetup setCurrencyId(final CurrencyId currencyId)
 
 		priceList_None.setC_Currency_ID(currencyId.getRepoId());
 
+		dunningDAO.retrieveDunnings().stream().filter(dunning -> dunning.getAD_Org_ID() == adOrg.getAD_Org_ID()).forEach(dunning -> {
```
dm-ps, line 274: Please create a new method in `de.metas.dunning.api.impl.AbstractDunningDAO`, something like `retrieveDunningsByOrg(OrgId orgId)` which is going to return only the dunnings for the respective org (i.e. fiter the dunnings directly in the query).
Also, IMHO, we should load these records in the constructor, have them as a field in this class, change the currency for them in this method and and save them in `saveInTrx()`. The same way we do for other parameters (fields) in this class.

dm-ps (changes requested):

backend/de.metas.fresh/de.metas.fresh.base/src/main/java/de/metas/fresh/setup/process/ClientSetup.java
dm-ps: Or a better way to do it (IMHO), better than keeping the list as a field, is the following:
add a new field: `Runnable updateDunningsOnSaveAction;`. Initialize this field in the `setCurrencyId`, something like: 
```
updateDunningsOnSaveAction = (ITrx.TRXNAME_ThreadInherited) => 
{ 
retrieveDunningsByOrg(OrgId orgId); 
set currency for the fetched records; 
save(ITrx.TRXNAME_ThreadInherited);
}
```
and then in `de.metas.fresh.setup.process.ClientSetup#saveInTrx`, have something like:

```
if (updateDunningsOnSaveAction != null)
{
  updateDunningsOnSaveAction.run();
}
```

dm-ps (approved):
LGTM

PR #17394, opened 2024-02-15
Display due date in sales invoice window

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/17393

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/invoice/callout/C_Invoice.java
+++ b/backend/de.metas.business/src/main/java/de/metas/invoice/callout/C_Invoice.java
@@ -12,6 +12,9 @@ import de.metas.invoice.location.adapter.InvoiceDocumentLocationAdapterFactory;
 import de.metas.invoice.service.IInvoiceBL;
 import de.metas.lang.SOTrx;
 import de.metas.payment.PaymentRule;
+import de.metas.payment.paymentterm.IPaymentTermRepository;
+import de.metas.payment.paymentterm.PaymentTerm;
+import de.metas.payment.paymentterm.PaymentTermId;
 import de.metas.pricing.PricingSystemId;
 import de.metas.pricing.service.IPriceListBL;
 import de.metas.util.Services;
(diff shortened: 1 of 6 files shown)


backend/de.metas.business/src/main/java/de/metas/invoice/callout/C_Invoice.java
```diff
@@ -182,4 +186,17 @@ public void updateBPartnerAddressForceUpdateCapturedLocation(final I_C_Invoice i
 		documentLocationBL.updateCapturedLocation(InvoiceDocumentLocationAdapterFactory.locationAdapter(invoice));
 
 	}
+
+	@CalloutMethod(columnNames = { I_C_Invoice.COLUMNNAME_C_PaymentTerm_ID, I_C_Invoice.COLUMNNAME_DateInvoiced }, skipIfCopying = true)
```
metas-ts, line 190: is it OK to do this on DateInvoiced, too?
i checked PaymentTerm.computeDueDate and it seems that the iver date would be overridden, even if `IsAllowOverrideDueDate` is `true`

metas-ts: ok, i guess the DueDate is never set via UI-Field-change..then i think it might be better to remove the dateInvoiced from the `@CalloutMethod` anotation, in order to reduce FUD

dm-ps: I think in this case, we're not really talking about overriding the `DueDate`, but setting the "initial" value for it (based on PaymentTerm and DateInvoiced). So, if we change the `DateInvoiced`, we'll set the initial (calculated) value of the `DueDate` for that `DateInvoiced`, even though, the PaymentTerm has `IsAllowOverrideDueDate` = `false`.
Then, if we really want to override the initial value, we will use the process and verify that the payment term allows it.

metas-ts (approved):

PR #17428, opened 2024-02-21
Intensive care uat without desadv changes

metas-ts:


backend/de.metas.edi/src/main/java-gen/de/metas/esb/edi/model/I_EDI_Desadv_Pack_Item.java
metas-ts, line 1: identical with `intensive_care_hotfix` ✅

misc/services/camel/de-metas-camel-edi/src/main/xsd/metasfresh/schema.xsd
metas-ts, line 1: identical with `intensive_care_hotfix` ✅

PR #17451, opened 2024-02-22
Fix paySelectionLine's pay-amount in case of payment-discount

metas-ts:
(shall go to uat now, despite the branch name)


metas-ts:
@mergifyio copy release

PR #17570, opened 2024-03-11
Do not do anything if there is not HU needed to be created

cristinamghita:
If qty book is negative and qty count is zero, that means that we do not need to create any HU or to distribute any quantity


teosarca (approved):
lgtm

PR #17578, opened 2024-03-11
Fix missing bank account details

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/17577


metas-ts (changes requested):
the code looks good, and the behaviour - AFIAU - is ok, buut, please look into the exsiting unit-tests and make sure to cover the new behavior

metas-ts (approved):

PR #17712, opened 2024-03-26
Retrieve Postfinance Results

metas-rc:


backend/de.metas.postfinance/src/main/java/de/metas/postfinance/docoutboundlog/PostFinanceLogRepository.java
```diff
@@ -70,11 +71,12 @@ private PostFinanceLog toPostFinanceLog(@NonNull final I_C_Doc_Outbound_Log_Post
...
 	}
 
-	public Optional<PostFinanceLog> retrieveLatestLogWithTransactionId(@NonNull final String transactionId)
+	public Optional<PostFinanceLog> retrieveLatestLogWithTransactionIdAndOrgId(@NonNull final String transactionId, @NonNull final OrgId orgId)
 	{
 		return queryBL.createQueryBuilder(I_C_Doc_Outbound_Log_PostFinance_Log.class)
 				.addOnlyActiveRecordsFilter()
-				.addEqualsFilter(I_C_Doc_Outbound_Log_PostFinance_Log.COLUMN_PostFinance_Transaction_Id, transactionId)
+				.addStringLikeFilter(I_C_Doc_Outbound_Log_PostFinance_Log.COLUMNNAME_PostFinance_Transaction_Id, transactionId, false)
+				.addEqualsFilter(I_C_Doc_Outbound_Log_PostFinance_Log.COLUMNNAME_AD_Org_ID, orgId)
```
mrie04, line 79: I think we don't need the OrgId here, because the transactionId contains <InvoiceId/DunningId><DocBaseType><DocSubType><DocumentNo> and because of this should be unique

backend/de.metas.postfinance/src/main/java/de/metas/postfinance/customerregistration/util/XMLUtil.java
```diff
@@ -100,4 +102,21 @@ private String toString(@NonNull final Document document) throws TransformerExce
...
+			final String documentAsString = toString(document);
+
+			final XmlMapper xmlMapper = ConfiguredXmlMapper.get();
+			return xmlMapper.readValue(documentAsString, Envelope.class);
+		}
+		catch (final Exception e)
+		{
+			throw new AdempiereException(e);
+		}
+	}
```
mrie04, line 121: reminder to cleanup :)

backend/de.metas.postfinance/src/main/java/de/metas/postfinance/document/results/GetResultsFromPostFinanceService.java
```diff
@@ -0,0 +1,185 @@
...
+
+			try
+			{
+				JAXBContext jc = JAXBContext.newInstance(Envelope.class);
+				final Unmarshaller unmarshaller = jc.createUnmarshaller();
+				final ByteArrayInputStream inputStream = new ByteArrayInputStream(file.getData().getValue());
+
+				final Envelope envelope = (Envelope)unmarshaller.unmarshal(inputStream);
+
+				envelope.getBody().getDeliveryDate().forEach(deliveryDate -> handleDeliveryDate(deliveryDate, attachmentEntryCreateRequest, orgId));
```
mrie04, line 83: It may be better to only attach billType relevant to C_Doc_outboundlog instead of the complete file in  attachmentEntryCreateRequest

backend/de.metas.postfinance/src/main/java/de/metas/postfinance/docoutboundlog/PostFinanceLogRepository.java
metas-rc: Because this is the earliest steps where I can exclude entries that are not for the parameter org.

mrie04: You already filtered by org, when you request processResults with orgs BillerId. And if you filter the results after download, if 2 Orgs use the same BillerId ou aren't able to download them again without changing isArchive

backend/de.metas.postfinance/src/main/java/de/metas/postfinance/document/results/GetResultsFromPostFinanceService.java
mrie04: As talked, it's ok to keep it like this

backend/de.metas.postfinance/src/main/java/de/metas/postfinance/docoutboundlog/PostFinanceLogRepository.java
metas-rc: right. Removed the org parameter.

mrie04 (changes requested):
pls check comments

backend/de.metas.postfinance/src/main/java/de/metas/postfinance/customerregistration/util/XMLUtil.java
```diff
@@ -24,8 +24,9 @@
 
 import com.fasterxml.jackson.dataformat.xml.XmlMapper;
 import de.metas.postfinance.ConfiguredXmlMapper;
-import de.metas.postfinance.jaxb.DownloadFile;
 import de.metas.postfinance.customerregistration.model.XmlCustomerRegistrationMessage;
+import de.metas.postfinance.jaxb.DownloadFile;
+import de.metas.postfinance.processprotocol.Envelope;
```
mrie04, line 29: unused I think

mrie04 (approved):
lgtm

PR #17720, opened 2024-03-27
Recreate edi xsd from export-formats

metas-ts:
❗ When merging towards `master`, make sure not to regress ❗
Actually, expect to discard the changes made in this PR


metas-ts:
@MergifyIO copy release_5.175

PR #17828, opened 2024-04-11
Avoid NPE on webapi startup

metas-ts:
The NPE isn't seen by the user, but if something else goes wrong, this is a red herring that just shouldn't happen


metas-ts:
@MergifyIO copy release_5.175

PR #17837, opened 2024-04-11
Prevent changing a product's C_UOM_ID when there is already a costdetail

metas-ts:


teosarca (approved):
as quick/hot fix, i approve it

PR #17861, opened 2024-04-15
Fix creation of one invoice from orders with different externalHeaderIds

metas-ts:


metas-ts:
💡 This is needed in addition to https://github.com/metasfresh/metasfresh/pull/17855/

PR #17870, opened 2024-04-16
PCM - don't attempt to import purchase order items if there are still masterdata files

metas-ts:


metas-ts:
closing; upon writinig tests I noticed that this doesn't work. 
The actual fix is included in https://github.com/metasfresh/metasfresh/pull/17879 .

PR #17942, opened 2024-04-24
fix reverse log logic

cp-ps:
https://github.com/metasfresh/metasfresh/issues/17902


backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/workpackage/ModularContractLogHandler.java
```diff
@@ -137,13 +136,13 @@ private void handleLogs(
 					.typeId(moduleConfig.getModularContractType().getId())
```
mrie04, line 136: minor, this should be improved, but I also didn't have time in my last task to do it and I try toDo it in my task. Just wanted to write it down

PR #17945, opened 2024-04-24
The interim invoice logs should point to the modular contract

metas-rc:


mrie04 (changes requested):
pls check my comments

mrie04 (approved):
lgtm

PR #18175, opened 2024-06-06
Mahndisposition - keine Mahnung erstellt

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/18173


metas-ts (approved):
it's OK now, but next time, please CP the existing fix.
in this case, this would have amounted to:
- going to https://github.com/metasfresh/metasfresh/pull/16137
- writing `@mergifyO copy pink_rooster_hotfix`

Issue #18184, opened 2024-06-07
Add `QtyPattern` field and set it as `Pattern Expression` in Jasper


riadmen:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?

- Add `QtyPattern` field and set it as `Pattern Expression` in Jasper

PR #18185, opened 2024-06-07
Add `QtyPattern` field and set it as `Pattern Expression`

riadmen:
https://github.com/metasfresh/metasfresh/issues/18184


cristinamghita (approved):
lgtm

PR #18199, opened 2024-06-10
fix shipment without harvesting details

mrie04:


backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/ModularContractProvider.java
```diff
@@ -288,20 +271,9 @@ public Stream<FlatrateTermId> streamModularPurchaseContractsForShipmentLine(@Non
...
-				.checkHasCompletedModularCondition(true)
-				.build());
-
-		final ProductId settingsProductId = CollectionUtils.extractSingleElementOrDefault(settings, ModularContractSettings::getRawProductId, null);
-		final ProductId productIdToUse = CoalesceUtil.coalesceNotNull(settingsProductId, productId);
-
 		final ModularFlatrateTermQuery query = ModularFlatrateTermQuery.builder()
 				.bPartnerId(warehouseBL.getBPartnerId(warehouseId))
-				.productId(productIdToUse)
+				.productId(productId)
```
mrie04, line 276: I think we don't need this, because of `de.metas.contracts.impl.FlatrateDAO#buildModularContractSettingsQueryFilter`

PR #18213, opened 2024-06-11
JsonPersisterService: Fix bug when IBAN is updated

metas-ts:
..also do minor improvements elsewhere


metas-ts:
@Mergify copy middle_ages_uat release release_5.175 yoyo_uat

PR #18260, opened 2024-06-18
One DESADV per shipment

cp-ps:
refs: https://github.com/metasfresh/metasfresh/issues/18259


backend/de.metas.edi/src/main/java-gen/de/metas/esb/edi/model/I_EDI_DesadvLine_InOutLine.java
```diff
@@ -0,0 +1,370 @@
+package de.metas.esb.edi.model;
+
+import org.adempiere.model.ModelColumn;
+
+import javax.annotation.Nullable;
+import java.math.BigDecimal;
```
metas-ts, line 6: why do we need an `m:n`-Table in addition to  `M_InOutLine.EDI_DesadvLine_ID`

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/DesadvDAO.java
```diff
@@ -331,4 +334,27 @@ public BPartnerId retrieveBPartnerFromEdiDesadvPackId(final int desadvLinePackID
...
+				.list(I_M_InOut.class);
+	}
+
+	@Override
+	@NonNull
+	public I_M_InOut_Desadv_V getInOutDesadvByInOutId(@NonNull final InOutId shipmentId)
+	{
+		return queryBL.createQueryBuilder(I_M_InOut_Desadv_V.class)
+				.addOnlyActiveRecordsFilter()
+				.addEqualsFilter(I_M_InOut_Desadv_V.COLUMNNAME_M_InOut_ID, shipmentId.getRepoId())
```
metas-ts, line 356: FYI you can add a `RepoIdAware` directly as a parameter

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EDIDesadvInOutLineDAO.java
```diff
@@ -0,0 +1,64 @@
...
+@Repository
+public class EDIDesadvInOutLineDAO
+{
+	private final IQueryBL queryBL = Services.get(IQueryBL.class);
+
+	@NonNull
+	public Optional<I_EDI_DesadvLine_InOutLine> getByInOutLineId(@NonNull final InOutLineId shipmentLineId)
+	{
+		return queryBL.createQueryBuilder(I_EDI_DesadvLine_InOutLine.class)
+				.addEqualsFilter(I_EDI_DesadvLine_InOutLine.COLUMNNAME_M_InOutLine_ID, shipmentLineId)
```
metas-ts, line 44: if we need this table, please add an index for this column

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/DesadvDAO.java
```diff
@@ -331,4 +334,27 @@ public BPartnerId retrieveBPartnerFromEdiDesadvPackId(final int desadvLinePackID
...
 				.firstId(BPartnerId::ofRepoId);
 	}
+
+	@Override
+	@NonNull
+	public List<I_M_InOut> retrieveShipmentsPendingExport(@NonNull final I_EDI_Desadv desadv)
+	{
+		return queryBL.createQueryBuilder(I_M_InOut.class, desadv)
+				.addOnlyActiveRecordsFilter()
+				.addEqualsFilter(I_M_InOut.COLUMNNAME_EDI_Desadv_ID, desadv.getEDI_Desadv_ID())
```
metas-ts, line 344: please make sure there in an idenx on this column

backend/de.metas.edi/src/main/sql/postgresql/system/80-de.metas.edi/5726254_sys_gh18259_add_M_InOut_DesadvLine_V.sql
```diff
@@ -0,0 +1,64 @@
+drop view if exists M_InOut_DesadvLine_V
+;
+
+create or replace view M_InOut_DesadvLine_V as
+select shipment.m_inout_id                                                       as M_InOut_Desadv_ID,
+       case
+           when desadvInOutLine.edi_desadvline_id > 0 then
+               desadvInOutLine.edi_desadvline_id || '-' || desadvInOutLine.m_inoutline_id
+                                                      else
+               uuid_generate_v4() || ''
```
metas-ts, line 10: seems as if this is different each time, so it probably won't work. 
when could `desadvInOutLine.edi_desadvline_id` be `<0` or `null` anyways

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/DesadvBL.java
```diff
@@ -1112,11 +1157,11 @@ private Optional<ITranslatableString> createSingleMsg(
 	}
 
 	@NonNull
-	private Optional<BigDecimal> computeDeliveredQtyInBPartnerUOM(
+	private Optional<Quantity> computeDeliveredQtyInBPartnerUOM(
```
metas-ts, line 1160: AFAIS this don't have to be optional, as it's never empty, or is  it?

metas-ts (changes requested):
..we also need the xsd-file..

backend/de.metas.edi/src/main/java-gen/de/metas/esb/edi/model/I_EDI_DesadvLine_InOutLine.java
cp-ps: actually, it's not an m:n, it's a 1:1, I just didn't  want to add the new columns to shipment line

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EDIDesadvInOutLineDAO.java
cp-ps: it will be a unique indexm as the table it's just an extension of the current m_inoutline table

backend/de.metas.edi/src/main/sql/postgresql/system/80-de.metas.edi/5726254_sys_gh18259_add_M_InOut_DesadvLine_V.sql
cp-ps: actually, it works just fine :) this is about those desadv lines that are not part of the current shipment, so we don't want to capture the packs for them

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/DesadvBL.java
cp-ps: it is:

```Java
if (orderLine.getC_UOM_BPartner_ID() <= 0)
		{
			return Optional.empty();
		}
```

backend/de.metas.edi/src/main/sql/postgresql/system/80-de.metas.edi/5726254_sys_gh18259_add_M_InOut_DesadvLine_V.sql
metas-ts: ah, ok, please add a comment to that effect to the DDL then

metas-ts (approved):
lgtm

PR #18263, opened 2024-06-18
Inactive products with qty in warehouse

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/18254

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/IHUQueryBuilder.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/IHUQueryBuilder.java
@@ -475,10 +475,4 @@ public interface IHUQueryBuilder
 	 * If false or not set = all HUs are included (the field {@link org.compiere.model.I_M_Product#COLUMNNAME_IsStocked} is ignored when filtering the HUs)
 	 */
 	IHUQueryBuilder setOnlyStockedProducts(final boolean onlyStockedProducts);
-
-	/**
-	 * Do not consider {@link I_M_HU#COLUMNNAME_C_BPartner_ID} in the query, so we do not filter based on this column at all
-	 * Otherwise, if {@link #addOnlyInBPartnerId(BPartnerId)} gets <code>null</code> we will search only for those records with {@link I_M_HU#COLUMNNAME_C_BPartner_ID} <code>null</code>.
-	 */
-	IHUQueryBuilder setIgnoreBPartnerColumn();
 }
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/IHUQueryBuilder.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/IHUQueryBuilder.java
@@ -476,5 +476,9 @@ public interface IHUQueryBuilder
 	 */
 	IHUQueryBuilder setOnlyStockedProducts(final boolean onlyStockedProducts);
 
-
+	/**
+	 * Do not consider {@link I_M_HU#COLUMNNAME_C_BPartner_ID} in the query, so we do not filter based on this column at all
+	 * Otherwise, if {@link #addOnlyInBPartnerId(BPartnerId)} gets <code>null</code> we will search only for those records with {@link I_M_HU#COLUMNNAME_C_BPartner_ID} <code>null</code>.
+	 */
+	IHUQueryBuilder setIgnoreBPartnerColumn();
 }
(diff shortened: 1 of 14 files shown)


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/IHUQueryBuilder.java
```diff
@@ -476,5 +476,9 @@ public interface IHUQueryBuilder
 	 */
 	IHUQueryBuilder setOnlyStockedProducts(final boolean onlyStockedProducts);
 
-
+	/**
+	 * Do not consider {@link I_M_HU#COLUMNNAME_C_BPartner_ID} in the query, so we do not filter based on this column at all
+	 * Otherwise, if {@link #addOnlyInBPartnerId(BPartnerId)} gets <code>null</code> we will search only for those records with {@link I_M_HU#COLUMNNAME_C_BPartner_ID} <code>null</code>.
+	 */
+	IHUQueryBuilder setIgnoreBPartnerColumn();
```
teosarca, line 483: why do we need this flag?
precisely which is the case we cannot achieve by onlyIfAssignedToBPartner and OnlyInBPartnerIds fields.

PS: I have the feeling we are overcomplicating things for future devs.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/model/validator/M_Product.java
```diff
@@ -58,4 +68,31 @@ public void updateAllM_HU_PI_Item_Products(final I_M_Product product)
...
+		final IStorageQuery storageQuery = storageEngine
+				.newStorageQuery()
+				.addProductId(ProductId.ofRepoId(product.getM_Product_ID()))
+				.setExcludeAfterPickingLocator(false)
+				.setOnlyStockedProducts(true)
+				.setOnlyActiveHUs(true)
+				.setIgnoreBPartnerColumn();
+
+		final IContextAware context = PlainContextAware.createUsingOutOfTransaction();
+		final boolean isEmptyStock = storageEngine.retrieveStorageRecords(context, storageQuery).isEmpty();
```
teosarca, line 91: why not introducing storageEngine.anyMatching(storageQuery) method?

i.e. why shall we load all records when we just want to check if there is any matching.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/filters/ProductFilterUtil.java
```diff
@@ -89,6 +87,18 @@ public static DocumentFilterDescriptor createFilterDescriptor()
 				.setWidgetType(DocumentFieldWidgetType.YesNo)
 				.setOperator(Operator.EQUAL);
 
+		final DocumentFilterParamDescriptor.Builder isActive = DocumentFilterParamDescriptor.builder()
+				.setFieldName(I_M_Product.COLUMNNAME_IsActive)
+				.setDisplayName(Services.get(IMsgBL.class).translatable(I_M_Product.COLUMNNAME_IsActive))
```
teosarca, line 92: 1. either extract `Services.get(IMsgBL.class)` as a local var and use it in all 3 cases u have,
2. or use de.metas.i18n.TranslatableStrings#adElementOrMessage

teosarca (changes requested):
check my comments

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/IHUQueryBuilder.java
dm-ps: When we have a Picked HU for which the BPartner is set, but I don't know that BPartner directly from the product (interceptor), so I want to entirely omit `M_HU.C_BPartner_ID` when fetching the records.
By default in `de.metas.storage.spi.hu.impl.HUStorageQuery#HUStorageQuery`, `OnlyIfAssignedToBPartner` is set to `false` and `addOnlyInBPartnerId` to null.
Therefore, in `de.metas.handlingunits.impl.HUQueryBuilder#createQueryFilter` we'll end up with a Set which is not empty, and we'll filter only for those record with `C_BPartner_ID IS NULL`.
image

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/model/validator/M_Product.java
dm-ps: will do, thx

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/filters/ProductFilterUtil.java
dm-ps: will do the first option, thx

teosarca (approved):

PR #18380, opened 2024-07-03
Warehouse BPartner + Allow location

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/18379

CP from: https://github.com/metasfresh/metasfresh/commit/cd7a5bd8bdb5544c905946fab19b387311c47396


metas-ts (approved):
the change was done two year ago and is in use, so LGTM

Issue #18381, opened 2024-07-03
Remove `Maturing Candidate` scripts


riadmen:
### Is this a bug or feature request?
FR

### What is the expected or desired behavior?

- Remove `Maturing Candidate` scripts

PR #18389, opened 2024-07-04
Remove `Maturing Candidate Window` scripts

riadmen:
https://github.com/metasfresh/metasfresh/issues/18381


teosarca (changes requested):
don't integrate it until we clarify it

PR #18409, opened 2024-07-05
Add C_BPartner.Lookup_Label as additional lookup-criterion when selecing BPartners during EDI-ORDERS import

metas-ts:
#18408


metas-ts:
closing because the slashes in the branch-name don't work well with the legacy build&deploy process

PR #18413, opened 2024-07-05
Add C_BPartner.Lookup_Label as additional lookup-criterion when selecing BPartners during EDI-ORDERS import

metas-ts:
#18408


metas-ts (commented):
lgtm

PR #18416, opened 2024-07-08
Remove the overriding for the `Maturing Candidate Window`

riadmen:
https://github.com/metasfresh/metasfresh/issues/18381


backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5713345_AddMaturingTablesAndConfigs.sql
```diff
@@ -0,0 +1,3 @@
+-- Window: Reiferdisposition, Overrides_Window_ID=NULL
+UPDATE AD_Window SET Overrides_Window_ID=NULL WHERE AD_Window_ID=541756
+;
```
mrie04, line 3: Like this it will still fail on rollout, if an custom pp_order_candidate window is created in an script with lower number. You also need to remove it from existing insert.

mrie04 (approved):
As update script already exists lgtm
( https://github.com/metasfresh/metasfresh/blob/92383bfd99476350e42ba30f759c6f8582eb6b26/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5715000_sys_maturing_fixes.sql )

mrie04:
@Mergifyio copy release

PR #18418, opened 2024-07-08
Mobile UI: HU Manager - make visible attributes configurable

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/18417


teosarca (changes requested):
pls check & test my QA commits

teosarca (approved):

PR #18431, opened 2024-07-10
Added AD_Tab.InsertLogic and logic to regulate when creation of new records is not a simple yes/no flag

adi-stefan:


backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/GridTabVO.java
```diff
@@ -241,6 +241,9 @@ private static boolean loadTabDetails(final GridTabVO vo, final ResultSet rs) th
 			{
 				vo.IsInsertRecord = false;
 			}
+			final String InsertLogic = rs.getString("InsertLogic");
```
metas-rc, line 244: why the capital letter?

PR #18434, opened 2024-07-10
merging after rollout pink_rooster_uat_6 into pink_rooster_hotfix

metas-marcus:
merging after pink_rooster_uat_6 into pink_rooster_hotfix


metas-marcus:
PR assumed okay as it is an update from 5.174

PR #18490, opened 2024-07-16
Make sure we are taking first dunning user and the rest

cristinamghita:


backend/de.metas.dunning/src/main/java/de/metas/dunning/document/archive/DunningDocOutboundLogMailRecipientProvider.java
```diff
@@ -97,6 +98,9 @@ public Optional<DocOutBoundRecipient> provideMailRecipient(@NonNull final DocOut
 		final List<I_C_Invoice> dunnedInvoices = dunningService.retrieveDunnedInvoices(dunningDocId);
 		final int singleCommonInvoiceContactId = CollectionUtils.extractSingleElementOrDefault(dunnedInvoices, I_C_Invoice::getAD_User_ID, -1);
 
+		final I_AD_User dunningContact =  dunningRecord.getC_Dunning_Contact();
```
adi-stefan, line 101: please use: de.metas.user.api.impl.UserDAO#getById, which uses a cache

backend/de.metas.dunning/src/main/java/de/metas/dunning/document/archive/DunningDocOutboundLogMailRecipientProvider.java
```diff
@@ -105,6 +109,43 @@ public Optional<DocOutBoundRecipient> provideMailRecipient(@NonNull final DocOut
...
+			if (Check.isNotBlank(dunningRecipient.getEmailAddress()))
+			{
+				Loggables.addLog("The dunned invoices all have invoiceUser={}, so we take that user as the dunning mail's participant", dunningRecipient);
+				return Optional.of(dunningRecipient);
+			}
+
+			if (Check.isNotBlank(locationEmail))
+			{
+				Loggables.addLog("The dunned invoices all have invoiceUser={} and the location has the email {} so we take that user as the dunning mail's participant, with this email address", dunningRecipient, locationEmail);
+				return Optional.of(dunningRecipient.withEmailAddress(invoiceEmail));
```
adi-stefan, line 141:
```suggestion
				return Optional.of(dunningRecipient.withEmailAddress(locationEmail));
```
Shouldn't this be using `locationEmail`, which is what we're checking as not-null? If `invoiceEmail` would be usable, it should have been used already 2 ifs above this one

backend/de.metas.dunning/src/test/java/de/metas/dunning/document/archive/DunningDocOutboundLogMailRecipientProviderTest.java
```diff
@@ -273,12 +368,16 @@ public void provideMailRecipient_dunned_invoices_without_common_email()
 	}
 
 	@SuppressWarnings("SameParameterValue")
-	private org.compiere.model.I_AD_User createUserRecord(final String eMail)
+	private org.compiere.model.I_AD_User createUserRecord(final String eMail, final boolean... isDunningUser)
```
adi-stefan, line 371:
```suggestion
	private org.compiere.model.I_AD_User createUserRecord(final String eMail)
	{
		return createUserRecord(eMail, false);
	}	
	private org.compiere.model.I_AD_User createUserRecord(final String eMail, boolean isDunningUser)	
```

backend/de.metas.dunning/src/test/java/de/metas/dunning/document/archive/DunningDocOutboundLogMailRecipientProviderTest.java
```diff
@@ -273,12 +368,16 @@ public void provideMailRecipient_dunned_invoices_without_common_email()
...
 	{
 		final org.compiere.model.I_AD_User userRecord = newInstance(I_AD_User.class);
 		userRecord.setName("userRecord");
 		userRecord.setEMail(eMail);
 		userRecord.setC_BPartner_ID(bPartnerRecord.getC_BPartner_ID());
+		if (isDunningUser.length > 0)
+		{
+			userRecord.setIsDunningContact(isDunningUser[0]);
+		}
```
adi-stefan, line 380:
```suggestion
		userRecord.setIsDunningContact(isDunningUser);
```

adi-stefan (changes requested):
Please see above

backend/de.metas.dunning/src/main/java/de/metas/dunning/document/archive/DunningDocOutboundLogMailRecipientProvider.java
cristinamghita: Yes, right!
Was a bug in the code below, that I did not notice, and I copied here as well.
Fixed here and in the bellow if.

adi-stefan (approved):

metas-ts:
@MergifyIO copy intensive_care_hotfix

PR #18495, opened 2024-07-17
Add Language fr_CH

mrie04:


mrie04:
languages also loaded from db

Issue #18518, opened 2024-07-21
Calculate the TU-Qty in `Split Shipment` as in `Partial Shipment`


riadmen:
### Is this a bug or feature request?
BUG

### What is the expected or desired behavior?
 - Calculate the TU-Qty in `Split Shipment` as in `Partial Shipment`

PR #18519, opened 2024-07-21
Calculate the TU-Qty in `Split Shipment` as in `Partial Shipment`

riadmen:
https://github.com/metasfresh/metasfresh/issues/18518


mrie04 (approved):
lgtm

PR #18526, opened 2024-07-23
Fix regression from conflict resolving done at PR #18420

metas-ts:


mrie04:
@Mergifyio copy release

PR #18540, opened 2024-07-24
start proto

metas-rc:


backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/interest/InterestComputationCommand.java
```diff
@@ -371,37 +371,37 @@ private void distributeInterest(
...
 	{
-		Check.assume(LogEntryDocumentType.CONTRACT_PREFINANCING == interimContractEntry.getDocumentType(),
-					 "Expecting DocumentType = " + LogEntryDocumentType.CONTRACT_PREFINANCING +
-							 " but got " + interimContractEntry.getDocumentType() + "! LogId=" + interimContractEntry.getId());
-		Check.assumeNotNull(interimContractEntry.getAmount(), "Invoices with no amount should've been skipped already! LogId="
-				+ interimContractEntry.getId());
+		Check.assume(LogEntryDocumentType.PURCHASE_MODULAR_CONTRACT == modularContractLogEntry.getDocumentType(),
+					 "Expecting DocumentType = " + LogEntryDocumentType.PURCHASE_MODULAR_CONTRACT +
+							 " but got " + modularContractLogEntry.getDocumentType() + "! LogId=" + modularContractLogEntry.getId());
+		Check.assumeNotNull(modularContractLogEntry.getAmount(), "Invoices with no amount should've been skipped already! LogId="
```
adi-stefan, line 382:
```suggestion
		Check.assumeNotNull(modularContractLogEntry.getAmount(), "Contracts with no amount should've been skipped already! LogId="
```

I'm not even sure this check makes sense, because if the contract is for amount 0, I question their business model.

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/interest/ModularContractAllocations.java
```diff
@@ -145,7 +145,7 @@ private Instant getInterimDate()
 			return cachedInterimContractDate;
 		}
 
-		cachedInterimContractDate = interimContractEntry.getTransactionDate().toInstant(orgDAO::getTimeZone);
+		cachedInterimContractDate = modularContractLogEntry.getTransactionDate().toInstant(orgDAO::getTimeZone);
```
adi-stefan, line 148: Does this still make sense? This is now the modular contract date, not the interim contract date (which did not make sense either, because the contract date != interim invoice date).
Maybe we should think about moving to use the interim date parameter from the process?

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/log/ModularContractLogDAO.java
```diff
@@ -482,6 +482,14 @@ public ModularContractLogEntriesList getModularContractLogEntries(@NonNull final
 				.collect(ModularContractLogEntriesList.collect());
 	}
 
+	@NonNull
+	public Optional<ModularContractLogEntry> getModularContractLogEntry(@NonNull final ModularContractLogQuery query)
```
adi-stefan, line 486: Should we javadoc that the query should have an orderBy clause?

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/interest/ModularContractAllocations.java
metas-rc: done

adi-stefan (approved):

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/interest/BonusAndInterestTimeInterval.java
```diff
@@ -31,7 +31,7 @@
 
 @Value
 @Builder
-public class BonusComputationTimeInterval
+public class BonusAndInterestTimeInterval
```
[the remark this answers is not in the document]
metas-rc: let's leave it like this for now, because of the time pressure.

PR #18549, opened 2024-07-26
activate flaky material cucumbers

mrie04:


mrie04:
Included in https://github.com/metasfresh/metasfresh/pull/18706

PR #18562, opened 2024-07-30
Merge to intensive_care_uat from ts/intensive_care_uat/merge-from/intensive_care_hotfix

metas-ts:


metas-ts:
Marcus was faster :-)

PR #18570, opened 2024-07-31
port unit test fix from master

metas-ts:


metas-ts:

PR #18585, opened 2024-08-02
Follow up - One DESADV per shipment - Invoice per shipment 

cp-ps:
https://github.com/metasfresh/metasfresh/issues/18583


metas-ts (approved):

metas-ts:
@cp-ps but please add the new aggregation-rule's migrations-script..

PR #18595, opened 2024-08-07
Make sure we are taking the dunning user and setting it in the dunning doc

cristinamghita:


metas-ts:
@MergifyIO copy intensive_care_hotfix

PR #18624, opened 2024-08-12
Fix NPE when a subtrahend is not zero but null

metas-ts:


metas-ts:
@MergifyIO copy master

metas-ts:

PR #18639, opened 2024-08-13
Fix invoice-candidate invoicing UOM problem

metas-ts:
Changes done during the debugging-session
- `InvoiceCandidateWithInOutLineAggregator`: little hack so that `qtyShippedButNotInvoiced` is equal to `qtyLeftToInvoice`, we continue with `qtyLeftToInvoice`, to avoid the need to compare different UOMs in this case
- `InvoiceCandidateWithInOutLine`: make sure that at least `getQtysAlreadyShipped` is coming in the IC's UOM
- minor improvements


metas-ts:
@mergifyio copy release

dm-ps (approved):
LGTM

PR #18672, opened 2024-08-15
Show Qty in TU UOM in Material Cockpit (based on default packing capacity

cp-ps:
https://github.com/metasfresh/metasfresh/issues/18671

### DIFF
--- a/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/QtyTUConvertor.java
+++ b/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/QtyTUConvertor.java
@@ -61,7 +61,7 @@ public class QtyTUConvertor implements QtyConvertor
 		final Capacity capacity = getCapacityForUomId(quantity.getUOM());
 
 		final BigDecimal qtyTUQty = quantity.toBigDecimal()
-				.divide(capacity.toBigDecimal(), 2, RoundingMode.HALF_UP);
+				.divide(capacity.toBigDecimal(), tuUOM.getStdPrecision(), RoundingMode.HALF_UP);
 
 		return Quantity.of(qtyTUQty, tuUOM);
 	}
--- a/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/QtyTUConvertor.java
+++ b/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/QtyTUConvertor.java
@@ -1,0 +1,81 @@
+/*
+ * #%L
+ * de.metas.ui.web.base
+ * %%
+ * Copyright (C) 2024 metas GmbH
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as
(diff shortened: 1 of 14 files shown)


backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/QtyTUConvertor.java
```diff
@@ -0,0 +1,81 @@
...
+	{
+		if (quantity == null)
+		{
+			return null;
+		}
+
+		final Capacity capacity = getCapacityForUomId(quantity.getUOM());
+
+		final BigDecimal qtyTUQty = quantity.toBigDecimal()
+				.divide(capacity.toBigDecimal(), 2, RoundingMode.HALF_UP);
```
metas-ts, line 64: do always up. event if a box is not filled, it's still one box

backend/de.metas.ui.web.base/src/main/sql/postgresql/system/41-de.metas.ui.web.base/5731360_sys_gh_add_new_sys_config.sql
```diff
@@ -0,0 +1,4 @@
+-- 2024-08-15T10:22:49.704Z
+INSERT INTO AD_SysConfig (AD_Client_ID,AD_Org_ID,AD_SysConfig_ID,ConfigurationLevel,Created,CreatedBy,EntityType,IsActive,Name,Updated,UpdatedBy,Value) VALUES (0,0,541729,'S',TO_TIMESTAMP('2024-08-15 13:22:49','YYYY-MM-DD HH24:MI:SS'),100,'de.metas.ui.web','Y','de.metas.ui.web.material.cockpit.SYS_CONFIG_SHOW_QTY_IN_TU_UOM',TO_TIMESTAMP('2024-08-15 13:22:49','YYYY-MM-DD HH24:MI:SS'),100,'N')
```
metas-ts, line 2: please ad a nice&userfriendly description

metas-ts (changes requested):
see comments :-)

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/QtyTUConvertor.java
cp-ps: hmmmmm, are we sure about this? I mean, you wouldn't be able to tell how much qty do you actually have

metas-ts: we never do, that's what rounding is about.
but if you have your stuff in 5PCE-boxes and you have anything at all, then you have at least one box.

cp-ps: alright, I thought this would be a bit different since they are checking the stock :-? but ok :D

metas-ts (approved):
as talked, when the issues are addressed, it's fine to integrate

PS: ...and when the build is green 🤓

PR #18674, opened 2024-08-16
Fix camt payment matching by using equals instead of ILIKE %invoiceDocOrEsrRefNoCandidate%

mrie04:


cristinamghita (approved):
lgtm

mrie04:
@Mergifyio copy release blue_horizon_uat

PR #18681, opened 2024-08-19
Fix ViewExcelExporter in case of column not displayed because of displayLogic

mrie04:


metas-ts (approved):

mrie04:
@Mergifyio copy release

PR #18690, opened 2024-08-20
remove Callout-Code that was called too early; call it at the right time

metas-ts:
OrderPricingSystem: the two callouts were fired before C_Order.C_BPartner_Location_Value_ID was updated and therefore always set the M_PriceList_ID for the country of the location that we are changing **away** from just now.

Instead, set the pricelist now in C_Order => no need to assume the right ordering of callouts


metas-ts:
@MergifyIO copy release

PR #18691, opened 2024-08-20
The documentno shall be added to description when the invoice gets voided only if the sys config says so

cristinamghita:


backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5731410_AddSysConfigForDocNoWhenVoidingInvoices.sql
```diff
@@ -0,0 +1,13 @@
+-- SysConfig Name: org.compiere.model.MInvoice.ANNOTATE_DOCNO_INVOICE
+-- SysConfig Value: Y
+-- 2024-08-20T14:52:42.284Z
+INSERT INTO AD_SysConfig (AD_Client_ID,AD_Org_ID,AD_SysConfig_ID,ConfigurationLevel,Created,CreatedBy,Description,EntityType,IsActive,Name,Updated,UpdatedBy,Value) VALUES (0,0,541730,'S',TO_TIMESTAMP('2024-08-20 17:52:42.139','YYYY-MM-DD HH24:MI:SS.US'),100,'If set to Y, then metasfresh will  he documentno will be added to description when the invoice gets voided.','D','Y','org.compiere.model.MInvoice.ANNOTATE_DOCNO_INVOICE',TO_TIMESTAMP('2024-08-20 17:52:42.139','YYYY-MM-DD HH24:MI:SS.US'),100,'Y')
+;
```
metas-rc, line 5: Please, check the description. There's a gap in it.

metas-rc: I would add the name "Description" to the name of this sys config. like in the method from the code: annotateDocNoToDescription . I think this way a user would have better chance to find the sys config by searching by the obvious field that it changes: "Description"

metas-rc (approved):

PR #18700, opened 2024-08-22
Improvement to the error-message in case an Upsert-Item contains a new partner but an existing bankaccount

metas-ts:


metas-ts (commented):
lgtm

PR #18710, opened 2024-08-22
Master modular contracts link with pp order

metas-rc:


backend/de.metas.contracts/src/main/sql/postgresql/system/50-de.metas.contracts/5731630_sys_PP_Order_Link_ModularContract.sql
```diff
@@ -0,0 +1,36 @@
+select backup_table('ad_reference');
+
+select backup_table('ad_relationtype');
+
+
+DELETE FROM ad_relationtype WHERE NAME = 'PP_Order -> ModCntr_Log';
+
+
+DELETE from ad_reference where name = 'ModCntr_Log_Target_PP_Order';
```
mrie04, line 9: Why by name? What if the name is changed until next release and delete doesn't fail if id isn't found?

mrie04: And why was this needed at all, don't get it as name is unique

metas-rc: I see your point but, in this particular case, I think the name is way more obvious than the ID. I think that adding this relation type and reference on the private repo only was a big mistake. The rest of the relations that follow the name pattern `` tablename -> ModCntr_Log`` are already on public level 

image

mrie04: I think something like this would be more clear, with using same inserts/updates and then also adding a note in customisations scripts that it was moved, so it's clear there also
````sql
DO
$$
    BEGIN
        -- moved from customer repo into metasfresh/metasfresh
        INSERT INTO ...
    EXCEPTION
        WHEN unique_violation THEN -- avoid error on rollout
            RAISE NOTICE 'already exists  for this customer (introduced here)';
    END
$$
;
```

mrie04: as talked, keep it as it is this time, as we don't expect changes in this area in the near future

PR #18712, opened 2024-08-23
Reduce flakyness of effortControlPurchaseInvoiceCandidate.feature by fixing AdRefListRepositoryOverJdbc

metas-ts:
AdRefListRepositoryOverJdbc:
When calling createRefListItem + getById + createRefListItem for the same ref-list value fast (in cucumber), then sometimes getById does not contain the previously created ref-list and then the 2nd createRefListItem call fails with a DBUniqueConstraintException. The only reasons i can think of are
 * that cache.clear() doesn't cause the cache to be really empty **before** it returns
 * that there is some other element of concurrency between reading and writing

 Also make S_Issue_StepDef output in which row it failed (in case the problem persists)


metas-ts (commented):
lgtm

metas-ts:
@MergifyIO copy release

PR #18717, opened 2024-08-26
Display C_InvoiceSchedule.InvoiceDistance because it's used in the BL; Also:

metas-ts:
- make the fields which are relevant to C_Invoice_Candidates not-updatable, because updating them can cause a respectable number of ICs to be flagged as invalid, and this alone can make the UI seem to hang
- instead, have a dedicated process to set those values

#18716


metas-ts:
@MergifyIO copy release

PR #18724, opened 2024-08-28
Extend docoutbound

cristinamghita:


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/report/DocOutboundConfigRepository.java
```diff
@@ -0,0 +1,169 @@
...
+public class DocOutboundConfigRepository
+{
+	public static final IQueryBL queryBL = Services.get(IQueryBL.class);
+
+	public I_C_Doc_Outbound_Config getConfigById(final int docOutboundConfigId)
+	{
+		return InterfaceWrapperHelper.load(docOutboundConfigId, I_C_Doc_Outbound_Config.class);
+	}
+
+	// note that this method doesn't directly access the DB. Therefore, a unit test DAO implementation can extend this
```
metas-ts, line 60: LE: ok, i later saw you moved this from `de.metas.document.archive.api.impl.DocOutboundDAO`
I think think calling it repo is a bit unfortunate..i would expect a repo to read and write domain-objects. but i don't have a better idea right now

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/report/DocOutboundConfigRepository.java
```diff
@@ -0,0 +1,169 @@
...
+	public final I_C_Doc_Outbound_Config retrieveConfigForModel(@NonNull final Object model)
+	{
+		final Properties ctx = InterfaceWrapperHelper.getCtx(model);
+		final int adTableId = InterfaceWrapperHelper.getModelTableId(model);
+
+		return retrieveConfig(ctx, adTableId);
+	}
+
+	@Cached(cacheName = I_C_Doc_Outbound_Config_CC.Table_Name + "#AD_PrintFormatAll")
+	public List<PrintFormatId> retrieveAllPrintFormatIds(final int docOutboundConfigId)
```
metas-ts, line 116: generally, for all these list-returning methods: better return `com.google.common.collect.ImmutableList`

metas-ts (approved):

PR #18725, opened 2024-08-28
Fix material receipt NPE

metas-ts:


metas-ts:

PR #18784, opened 2024-09-05
HU-Unreserve on order void or reverse (or reactivate)

metas-ts:
Add a little framework so that different modules can register own logic.
Also
- convert the existing code in service-repair-order module accordingly
- create a default-handler which covers void or reversal of "normal" order


metas-ts (commented):
checked the code, particularly `C_Order.java` and `RepairOrderUnProcessHandler.java`, lgtm

PR #18821, opened 2024-09-10
Adjust for partial deliveries with individual invoices

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/18820

### DIFF
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandDAO.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandDAO.java
@@ -4,9 +4,6 @@ import ch.qos.logback.classic.Level;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import de.metas.aggregation.model.I_C_Aggregation;
-import de.metas.aggregation.model.I_C_AggregationItem;
-import de.metas.aggregation.model.I_C_Aggregation_Attribute;
-import de.metas.aggregation.model.X_C_AggregationItem;
 import de.metas.async.AsyncBatchId;
 import de.metas.bpartner.BPartnerId;
 import de.metas.bpartner.service.IBPartnerDAO;
@@ -90,6 +87,7 @@ import org.compiere.model.I_C_InvoiceSchedule;
 import org.compiere.model.I_C_OrderLine;
 import org.compiere.model.I_M_InOut;
 import org.compiere.model.I_M_InOutLine;
+import org.compiere.model.I_M_MatchInv;
 import org.compiere.util.DB;
 import org.compiere.util.Env;
 import org.compiere.util.TimeUtil;
@@ -113,9 +111,7 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.UUID;
 
-import static de.metas.invoicecandidate.api.impl.AggregationKeyEvaluationContext.ATTRIBUTE_CODE_AggregatePer_M_InOut_ID;
 import static org.adempiere.model.InterfaceWrapperHelper.delete;
-import static org.compiere.model.I_M_InOut.COLUMNNAME_M_InOut_ID;
 
 /*
  * #%L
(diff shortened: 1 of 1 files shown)


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandDAO.java
```diff
@@ -470,20 +475,26 @@ public List<I_C_InvoiceCandidate_InOutLine> retrieveICIOLAssociationsFor(@NonNul
...
+			return queryBL.createQueryBuilder(I_C_Invoice.class)
+					.addOnlyActiveRecordsFilter()
+					.addEqualsFilter(COLUMNNAME_M_InOut_ID, inOutId)
+					.addInArrayOrAllFilter(I_C_Invoice.COLUMNNAME_DocStatus, IDocument.STATUS_Closed, IDocument.STATUS_Completed) // DocStatus in ('CO', 'CL')
+					.create()
+					.anyMatch();
+		}
+		else
+		{
+			return invoiceLineAllocQueryBuilder
```
metas-ts, line 491: doesn't ` invoiceLineAllocQueryBuilder` at this point have all the filters and "andCollect"s that were added in the above if-clause?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandDAO.java
```diff
@@ -2013,4 +2024,37 @@ private ICompositeQueryFilter<I_C_Invoice_Candidate> toFilter(@NonNull final Inv
...
+
+	private boolean isAggregationWithAttributeAggregatePerInOut(@NonNull final IQueryBuilder<I_C_Invoice_Line_Alloc> invoiceLineAllocQueryBuilder)
+	{
+		return invoiceLineAllocQueryBuilder
+				.andCollect(I_C_Invoice_Candidate.COLUMNNAME_C_Invoice_Candidate_ID, I_C_Invoice_Candidate.class)
+				.andCollect(I_C_Invoice_Candidate.COLUMNNAME_HeaderAggregationKeyBuilder_ID, I_C_Aggregation.class)
+				.andCollectChildren(I_C_AggregationItem.COLUMNNAME_C_Aggregation_ID, I_C_AggregationItem.class)
+				.addEqualsFilter(I_C_AggregationItem.COLUMNNAME_Type, X_C_AggregationItem.TYPE_Attribute)
+				.andCollect(I_C_AggregationItem.COLUMNNAME_C_Aggregation_Attribute_ID, I_C_Aggregation_Attribute.class)
+				.addEqualsFilter(I_C_Aggregation_Attribute.COLUMNNAME_Code, ATTRIBUTE_CODE_AggregatePer_M_InOut_ID)
```
metas-ts, line 2056: from a quick glimpse IDK where exactly, but i think there have to be filters for IsActive, because the propery might be there, but not deactivated

metas-ts (changes requested):
pls check my comments. Still besides, I don't think it's a reliable approach as 
- aggregation-rules can change over time.
- it smells like a workaround..and I'm not really certain this works in all cases

I thinks it would be better to use `I_M_MatchInv` to check if an inout is related to an invoice

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandDAO.java
dm-ps: You're right about `M_MatchInv`, changed the implementation to use it

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandDAO.java
dm-ps: Changed the implementation to use `M_MatchInv`

metas-ts (approved):

PR #18861, opened 2024-09-12
fix unrelated error on building mobile ui

dm-ps:


dm-ps:
Will be done with https://github.com/metasfresh/metasfresh/pull/18866

PR #18911, opened 2024-09-17
Waschproben Additional Fee Raw Material

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/18910


metas-ts (changes requested):
lgtm, just one small request that will make the result much smoother

metas-ts (approved):

PR #18919, opened 2024-09-18
fix Abrechnung Waschproben BIO paymentterm_ID

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/18910


metas-ts:
@mergifyio copy new_dawn_uat

PR #18969, opened 2024-09-26
DocOutbound mail recipient respect flag isInvoiceMailEnabled

mrie04:


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/IBPartnerBL.java
```diff
@@ -219,6 +219,9 @@ public enum IfNotFound
 		@NonNull
 		Comparator<User> comparator = Comparator.comparing(User::getName);
 
+		@Default
+		boolean onlyMailEnabled = false;
```
metas-ts, line 223: i think the field should rather be named `onlyIfInvoiceEmailEnabled`

metas-ts (changes requested):

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/IBPartnerBL.java
dm-ps: done

metas-ts (approved):

PR #18986, opened 2024-09-27
Intensive care uat materialdispo tweaks

metas-ts:
- extend the test to cover scenario with UOM-conversions
- correctly handle MD_Candidates if an order is voided
- add relationtype for PP_Order_candidate-parent and tweak display-columns


metas-ts (commented):
looked over my changes, still lgtm

PR #19030, opened 2024-10-02
Tasty ham hotfix hu trace

metas-rc:


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/trace/HUTraceEventsService.java
```diff
@@ -100,24 +100,16 @@ public class HUTraceEventsService
 			I_M_MovementLine.Table_Name,
 			I_PP_Cost_Collector.Table_Name,
 			I_M_ShipmentSchedule_QtyPicked.Table_Name);
-
+	final IHandlingUnitsBL handlingUnitsBL = Services.get(IHandlingUnitsBL.class);
```
cp-ps, line 103: I guess you forgot `private` :)

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/trace/HUTraceEventsService.java
```diff
@@ -225,6 +217,11 @@ void createAndAddEventsForInventoryLines(
 			for (final InventoryLineHU inventoryLineHU : inventoryLineHUs)
 			{
 				final HuId huId = inventoryLineHU.getHuId();
+				if(huId == null)
```
cp-ps, line 220: this one can't really happen, if it does I believe we should throw an error as we would be missing the events for that inventory like see: 

image

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/trace/HUTraceEventsService.java
```diff
@@ -727,14 +721,19 @@ void createAndAddEventsForPOIssueOrReceipt(
 			else
 			{
 				builder.topLevelHuId(huId);
-				createTraceForPOIssueOrReceiptHU(builder, ppCostCollector, huRecord);
 			}
-
 			final List<I_M_HU> vhus = huAccessService.retrieveVhus(huId);
-
-			for (final I_M_HU vhu : vhus)
+			if (vhus.isEmpty())
```
cp-ps, line 726: I'd rather go with something like: isCU(huId) => create the events, else retrieve the vhus and create the events for them

PR #19107, opened 2024-10-11
Handle desadv lines with same prod picked onto same HU

metas-ts:


metas-rc:
@Mergifyio copy intensive_care_uat middle_ages_uat gray_knight_uat oasis_sun_uat

PR #19189, opened 2024-10-18
Fix the invoicing problem related to override due date

metas-ts:
If there are two different overrideDueDates, take the later one. 
Also add a bit of `@Contract` and get rid of some annoying warnings


cristinamghita (approved):
lgtm

PR #19190, opened 2024-10-18
EDI_Desadv_GenerateSSCCLabels creates DESADV-Pack with Packagin-GTINs and -Codes

metas-ts:
Prototype


dm-ps (approved):
Tested on local env. LGTM

PR #19197, opened 2024-10-21
fix duplicate printing

cp-ps:
refs: https://github.com/metasfresh/metasfresh/issues/19186


metas-ts:
@MergifyIO copy new_dawn_uat

Issue #19231, opened 2024-10-25
Show `LotNo` and `Best before date` on `Shipment Note`


riadmen:
### Is this a bug or feature request?
FR
### What is the expected or desired behavior?

- Show `LotNo` and `Best before date` on `Shipment Note`

PR #19232, opened 2024-10-25
Show `LotNo` and `Best before date` on `Shipment Note`

riadmen:
https://github.com/metasfresh/metasfresh/issues/19231


backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/Docs_Sales_InOut_Details.sql
```diff
@@ -160,26 +155,20 @@ FROM M_InOutLine iol
...
-                                                        AS Attributes,
-
-           at.M_AttributeSetInstance_ID,
-           STRING_AGG(REPLACE(at.ai_value, 'MHD: ', ''), ', ')
-           FILTER (WHERE at.at_value LIKE 'HU_BestBeforeDate')
-                                                        AS best_before_date,
-           STRING_AGG(ai_value, ', ')
-           FILTER (WHERE at.at_value LIKE 'Lot-Nummer') AS lotno
+         LEFT OUTER JOIN (SELECT STRING_AGG((at.at_Name || ': ' || REPLACE(at.ai_value, 'MHD: ', '')), '<br> ' ORDER BY at.at_Value DESC) AS Attributes,
+                                 at.M_AttributeSetInstance_ID,
```
cristinamghita, line 159: Why did you add this break line?
From what I can see Lotno and Bestebefore date shall already be visible.

cristinamghita (changes requested):
PLs check my comments

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/Docs_Sales_InOut_Details.sql
riadmen: i just take the same logic and display from the existing customers: 

- breack are used to make the attributes more orgenized rather than just `, `
- this 2 attributes was not displayed because of this filter in the standard function:
```SQL
FILTER (WHERE at.at_value NOT IN ('HU_BestBeforeDate', 'Lot-Nummer')
```

cristinamghita: but you have the attribute 2 lines bellow, for lotno and best before date.
And regarding the break, pls do not put a break and leave the comma there as it is.
You break standard functionality with this.
If there are 10 attributes, they will be displayed in 10 lines! 
So, I think there is not need to change this SQL.
Only in the jasper, bellow attribute, add lotno and bestbefore date.

cristinamghita (approved):
lgtm

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/Docs_Sales_InOut_Details.sql
riadmen: done

PR #19235, opened 2024-10-25
Function that create Partner, Project and external project reference

dm-ps:


dm-ps:
Not needed here (it is instances specific)

PR #19263, opened 2024-10-31
Import Invoice Candidate External Reference Price List Version Name

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/19262


metas-ts (commented):
what I see LGTM, but is the PLV actually displayed in the invoice-candidate window?
if not, i think it should be, at least in advanced edit 🤔

metas-ts (approved):
...as written

PR #19338, opened 2024-11-12
Modular contract proforma handling and definitive invoice adjustments

mrie04:


backend/de.metas.business/src/main/java/de/metas/order/OrderLineQuery.java
```diff
@@ -0,0 +1,36 @@
...
+
+import de.metas.contracts.FlatrateTermId;
+import lombok.Builder;
+import lombok.Value;
+
+import javax.annotation.Nullable;
+
+@Builder
+@Value
+public class OrderLineQuery
```
metas-rc, line 33: Is this query with one field really needed?

backend/de.metas.business/src/main/java/de/metas/order/impl/AbstractOrderDAO.java
```diff
@@ -524,4 +525,17 @@ public Stream<I_C_Order> streamOrders(@NonNull final IQueryFilter<I_C_Order> ord
 				.create()
 				.iterateAndStream();
 	}
+
+	@Override
+	public Stream<I_C_OrderLine> streamOrderLines(@NonNull final OrderLineQuery query)
```
metas-rc, line 530: I would use something like "streamByModularPurchaseContractId" and pass only the flatrate term ID instead of the query.

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/ModularContractService.java
```diff
@@ -228,11 +252,84 @@ public void afterInvoiceReverse(@NonNull final I_C_Invoice invoiceRecord, @NonNu
...
+			final boolean isErrorIfMoreThanOneFound)
+	{
+		final OrderId orderId = OrderId.ofRepoId(orderLineRecord.getC_Order_ID());
+		final ProductId productId = ProductId.ofRepoId(orderLineRecord.getM_Product_ID());
+		final Set<FlatrateTermId> contractIds = contractProvider.getInitialPurchaseModularContractCandidatesForSalesOrderLine(orderId, productId);
+		final FlatrateTermId currentFlatrateTermId = FlatrateTermId.ofRepoIdOrNull(orderLineRecord.getPurchase_Modular_Flatrate_Term_ID());
+
+		if (contractIds.isEmpty())
+		{
+			orderLineRecord.setPurchase_Modular_Flatrate_Term_ID(-1);
```
metas-rc, line 311: This could delete a contract form an order line that is not related with modular contracts! 
We need a way to make sure we only apply this logic to relevant lines.

mrie04: this is a new column for modular contract only

backend/de.metas.business/src/main/java/de/metas/order/impl/AbstractOrderDAO.java
mrie04: I did this, so it is extendable as it returns a common usable I_C_OrderLine stream

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/interceptor/C_Order.java
```diff
@@ -79,6 +79,18 @@ public class C_Order
...
+	@DocValidate(timings = ModelValidator.TIMING_BEFORE_COMPLETE)
+	public void beforeComplete(@NonNull final I_C_Order orderRecord)
+	{
+		contractService.setPurchaseModularContractIdsIfExists(orderRecord, true);
+	}
+
+	@ModelChange(timings = ModelValidator.TYPE_AFTER_CHANGE, ifColumnsChanged = { I_C_Order.COLUMNNAME_Harvesting_Year_ID, I_C_Order.COLUMNNAME_M_Warehouse_ID })
+	public void afterChange(@NonNull final I_C_Order orderRecord)
+	{
+		contractService.setPurchaseModularContractIdsIfExists(orderRecord, false);
```
metas-rc, line 91: Again, I think we should make sure we only affect modular related orders. For example, don't try to go forward if there's no harvesting year set.

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/interceptor/C_OrderLine.java
```diff
@@ -57,8 +53,14 @@ public void beforeDelete(@NonNull final I_C_OrderLine orderLine)
...
-			throw new AdempiereException(msgBL.getTranslatableMsgText(MSG_ERR_DELETION_NOT_ALLOWED))
+			throw new AdempiereException(MSG_ERR_DELETION_NOT_ALLOWED)
 					.markAsUserValidationError();
 		}
 	}
+
+	@ModelChange(timings = { ModelValidator.TYPE_BEFORE_NEW, ModelValidator.TYPE_BEFORE_CHANGE },ifColumnsChanged = { I_C_OrderLine.COLUMNNAME_M_Product_ID })
+	public void updatePurchaseModularContractId(@NonNull final I_C_OrderLine orderLine)
+	{
+		contractService.updatePurchaseModularContractId(orderLine, false);
```
metas-rc, line 64: As mentioned above, we must make sure we don't affect non-modular order lines. We can do it either here or in the method ``de.metas.contracts.modular.ModularContractService#updatePurchaseModularContractId`` but we should do it somewhere

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/log/LogsRecomputationService.java
```diff
@@ -82,6 +92,86 @@ public class LogsRecomputationService
...
+														   .orderIds(salesOrderIds)
+														   .build())
+					.forEach(shippingNotification -> recomputeForShippingNotificationLines(shippingNotification.getLines()));
+
+			ppOrderDAO.streamManufacturingOrders(ManufacturingOrderQuery.builder()
+														 .modularFlatrateTermId(flatrateTermId)
+														 .build())
+					.map(I_PP_Order::getPP_Order_ID)
+					.map(PPOrderId::ofRepoId)
+					.forEach(this::recomputeForPPOrder);
```
metas-rc, line 134: I think we should also recompute this log ``de.metas.contracts.modular.computing.purchasecontract.addedvalue.interim.PurchaseModularContractLog``

mrie04: This always exists before final invoice, so no need to recompute it

metas-rc: I am thinking about this de.metas.contracts.modular.computing.purchasecontract.addedvalue.interim.AVInterimComputingMethod#splitLogsIfNeeded but fine by me to leave it out for now and fix it later when we test/repair interests

metas-rc (approved):

PR #19367, opened 2024-11-18
Fix an error creating migration-logs

metas-ts:


metas-ts:
@MergifyIO copy release

PR #19375, opened 2024-11-19
Merge to master from origin/release

metas-ts:


metas-ts:
(jasper conflicts solved under cristina's direction)

PR #19400, opened 2024-11-22
Fix calculation of the number of LUs on deliveries without actual picking

metas-ts:


backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/pack/EDIDesadvPackService.java
```diff
@@ -280,6 +283,9 @@ else if (!lutuConfigurationInStockUOM.isInfiniteQtyCU())
 
 		StockQtyAndUOMQty remainingQty = qtyToAdd;
 
+		// About the "remainingQty.signum() > 0":
+		// although it should be fine to rely on requiredLUCount, there was a case - which I didn't manage to reproduce - 
+		// where a superflous pack with movementQty=0 was created.
```
metas-ts, line 288:
```suggestion
```

PR #19447, opened 2024-11-27
Add the docstatus-condition to JsonInvoiceReviewService

metas-ts:
Also docu-improvements


metas-ts:
created from the wrong base-branch

PR #19560, opened 2024-12-09
Support SCOR type when Importing esr  files

cristinamghita:


backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/dataimporter/ESRImportEnqueuer.java
```diff
@@ -217,50 +214,85 @@ private void createImportFileFromSingleAttachment(@NonNull final I_ESR_Import es
 	}
 
 	private void createImportFilesFromZips(@NonNull final I_ESR_Import esrImport,
-			@NonNull final ByteArrayInputStream in)
+										   @NonNull final ByteArrayInputStream in)
 	{
-		final List<ZipFileResource> unzippedFiles = getZipFileResources(in);
+		final List<ZipFileResource> unzippedFiles = getZipFileResourcesSafely(in);
```
metas-ts, line 219: AFAIS we now try to import all zipped files and don't stop everything on the first error.
But does this work in practice? 
If a zipped file is partially imported, how does the user proceed once the problem was addressed?
I mean, without importing the same files twice? or do we have a check for that?

cristinamghita: tbh, not sure what he can do, except importing separately the the files that failed

PR #19575, opened 2024-12-10
Fix AD_Field_v(t) and make sure next time there is a helpful exception

metas-ts:


metas-ts:
Merged to master as part of https://github.com/metasfresh/metasfresh/pull/19600

PR #19579, opened 2024-12-10
Fix locking failure in C_OLCandEnqueueForSalesOrderCreation 

metas-ts:


metas-ts:
Merged to master as part of https://github.com/metasfresh/metasfresh/pull/19600

PR #19650, opened 2024-12-16
Make `Product` field Mandatory in `Packing Instruction` level

riadmen:
https://github.com/metasfresh/metasfresh/issues/19649


cristinamghita (approved):
lgtm

PR #19655, opened 2024-12-16
C_Doc_Outbound_Log EDI Fields changes

metas-rc:


backend/de.metas.edi/src/main/java/de/metas/edi/api/EDIDocOutBoundLogService.java
```diff
@@ -77,4 +77,16 @@ public void save(@NonNull final de.metas.edi.model.I_C_Doc_Outbound_Log docOutbo
 	{
 		InterfaceWrapperHelper.save(docOutboundLog);
 	}
+
+	public void setEdiExportStatusFromInvoiceRecord(final @NonNull I_C_Doc_Outbound_Log docOutboundLogRecord)
```
metas-ts, line 81: now there are two very similar setEdiExportStatusFromInvoiceRecord method in the service, each one is called once.
please check if one public method is enough. if not, please please deduplicate the logic.

metas-ts (changes requested):

metas-ts (approved):

Issue #19735, opened 2025-01-08
Make sure pricelist-comparison DB-functions are fixed


riadmen:
### Is this a bug or feature request?
BUG
### What is the expected or desired behavior?

- The problem was that in the DB, some DB-functions contained `...ORDER BY true,...` which is not supported anymore with pg-15.

PR #19745, opened 2025-01-08
Make sure pricelist-comparison DB-functions are fixed

riadmen:
https://github.com/metasfresh/metasfresh/issues/19735


cristinamghita (approved):
Order by is the same for all - lgtm

PR #19763, opened 2025-01-10
Bestelldispo: UOM conversion not used correctly for products where stock UOM != price UOM

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/19759


metas-ts (approved):
generally LGTM, ❗ please just fix the formatting before integrating

PR #19808, opened 2025-01-15
Window Doctype - bring fields from advanced edit to first page

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/19807


dm-ps:
@Mergifyio copy hard_encoded_uat

metas-ts (commented):
from the ducomented dev-test i'm a bit nervous about the changes.
I can see why we want the flags on the pront-page, but i suspect that many don't have proper field descriptions and it's not clear what they do.

From the dev-test-docu i'm also not sure which field is shown when (and if i knew, maybe this would have to be overhauled, too)

finally, i wonder if those fields shouldn't be grouped together..or some of them.

I'm afraid that given them to anyone like this will cause additional confusion. Please discuss this with torby. if he's still OK with the current PR, then fine with me

dm-ps:
@Mergifyio copy hard_encoded_uat

dm-ps:
@Mergifyio copy hard_encoded_uat

dm-ps:
@Mergifyio copy hard_encoded_uat

PR #19876, opened 2025-01-21
No NPE when creating an account that is not summary

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/19875


backend/de.metas.acct.base/src/main/java/de/metas/acct/interceptor/C_ElementValue.java
```diff
@@ -98,7 +95,7 @@ protected void createValidCombinationIfNeeded(final I_C_ElementValue elementValu
 		final ChartOfAccountsId chartOfAccountsId = ChartOfAccountsId.ofRepoId(elementValue.getC_Element_ID());
 		for (final AcctSchema acctSchema : acctSchemasRepo.getByChartOfAccountsId(chartOfAccountsId))
 		{
-			accountDAO.getOrCreate(accountDimensionTemplate.setAcctSchemaId(acctSchema.getId()).build());
```
metas-ts, line 101: please check instead if `IAccountDAO.getOrCreate` is used by others. 
If not, please make it work within the trx, deprecate  `MAccount.getCreate` and add a hint towards `IAccountDAO.getOrCreate`. 
And if `IAccountDAO.getOrCreate` is needed elsewhere, and has to be out-of-trx, please create a new method `IAccountDAO.getOrCreateWithinTrx` and use that

metas-ts (dismissed):

cp-ps (approved):

PR #19935, opened 2025-01-27
Fix MD_Stock_Update_From_M_HUs in case of M_Warehouse.AD_Org_ID=0

metas-ts:


metas-ts:
warehouses should neverever have ad_org_id=0. this case was due to some old datahack.
will instead have a DB-constraint

PR #19952, opened 2025-01-29
consider LU destroyed if unpicking the last TU from it

adi-stefan:


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/service/commands/PickingJobUnPickCommand.java
```diff
@@ -251,7 +251,7 @@ private LUPickingTarget reinitializeLUPickingTarget(@NonNull final LUPickingTarg
 		}
 
 		final I_M_HU lu = handlingUnitsBL.getById(luId);
-		if (!handlingUnitsBL.isDestroyed(lu))
+		if (!handlingUnitsBL.isEmptyStorage(lu))
```
cp-ps, line 254: @teosarca suggestion: let's introduce handlingUnitsBL.isDestroyOrEmpty(I_M_HU)

PR #20099, opened 2025-02-18
Propagate edi export status to shipments EDI_Exp_Desadv

dm-ps:


metas-ts (dismissed):
i think we rather need a model-interceptor, because when the EDI_ExportStatus is changed to "Send" or "Error", we need to update the shipments accordingly.

PR #20116, opened 2025-02-20
Fix spelling of sender and receiver GLN in inhouse INVOIC

metas-ts:


metas-ts:
@MergifyIO copy new_dawn_uat_soft_panda_release

PR #20236, opened 2025-03-04
make sure the reversal has the same totals as the original invoice, with opposite sign

metas-rc:


teosarca (approved):
fine for me in given time&budget

PR #20245, opened 2025-03-04
Don't show process to create prices with no schema selected

metas-rc:


backend/de.metas.business/src/main/java-legacy/org/compiere/process/M_PriceList_Create.java
```diff
@@ -22,11 +28,31 @@
...
+	private  final transient IPriceListDAO priceListDAO = Services.get(IPriceListDAO.class);
+
+	@Override
+	public ProcessPreconditionsResolution checkPreconditionsApplicable(final @NonNull IProcessPreconditionsContext context)
+	{
+		final PriceListVersionId priceListVersionId = PriceListVersionId.ofRepoIdOrNull(context.getSingleSelectedRecordId());
+		if (priceListVersionId == null)
+		{
+			return ProcessPreconditionsResolution.rejectBecauseNotSingleSelection().toInternal();
+		}
```
adi-stefan, line 45:
```suggestion
		if (context.isMoreThanOneSelected())
		{
			return ProcessPreconditionsResolution.rejectBecauseNotSingleSelection().toInternal();
		}
		final PriceListVersionId priceListVersionId = PriceListVersionId.ofRepoIdOrNull(context.getSingleSelectedRecordId());
```

because `context.getSingleSelectedRecordId()`, based on its javadoc, will throw an error in case of multiple selection

adi-stefan (changes requested):
See my comment

adi-stefan (approved):

PR #20285, opened 2025-03-07
Reuse MoveHUCommand for putting back HUs to initial locator

cp-ps:
https://github.com/metasfresh/metasfresh/issues/20262


teosarca (approved):
check my QA commit. besides that lgtm

PR #20295, opened 2025-03-07
New dawn uat merge soft panda hotfix

mrie04:


mrie04:
manual rerun because of internal server error on pushing maven dependencies https://github.com/metasfresh/metasfresh/actions/runs/13725881961

PR #20333, opened 2025-03-13
Performance Improvements - Shipment Complete with EDI (DESADV)

mrie04:


backend/de.metas.edi/src/main/sql/postgresql/system/80-de.metas.edi/5749050_indices_for_preformance_on_shipment_completion.sql
```diff
@@ -0,0 +1,28 @@
+CREATE INDEX IF NOT EXISTS EDI_Desadv_Pack_M_HU_ID on edi_desadv_pack (m_hu_id)
+;
+
+CREATE INDEX IF NOT EXISTS M_Package_HU_HU_ID on M_Package_HU (m_hu_id)
+;
+
+CREATE INDEX IF NOT EXISTS m_hu_m_hu_pi_version_id ON m_hu(m_hu_pi_version_id)
```
metas-ts, line 7: not sure about this index; for each `m_hu_pi_version_id` there will be a large amount of HUs, wo the DB will habe to scan a large part of that index

backend/de.metas.edi/src/main/sql/postgresql/system/80-de.metas.edi/5749050_indices_for_preformance_on_shipment_completion.sql
```diff
@@ -0,0 +1,28 @@
...
+;
+
+-- old
+-- CREATE INDEX m_hu_commonsearch
+--     ON m_hu (ad_client_id, m_locator_id, c_bpartner_id, hustatus, m_hu_pi_version_id)
+--     WHERE (isactive = 'Y'::bpchar)
+-- ;
+DROP INDEX IF EXISTS m_hu_commonsearch;
+CREATE INDEX m_hu_commonsearch
+    ON m_hu (m_hu_pi_version_id, hustatus, m_locator_id,  c_bpartner_id, ad_client_id )
```
metas-ts, line 23: not knowing the SQL(s) this index will support, i generally think it would be better to have m_locator_id or c_bpartner_id at the front. reasoning similar to `m_hu_m_hu_pi_version_id`. and/or maybe have more indices than just this one.

mrie04: tested with explain and this was the fastest order

backend/de.metas.edi/src/main/sql/postgresql/system/80-de.metas.edi/5749050_indices_for_preformance_on_shipment_completion.sql
mrie04: as talked, we will keep it

PR #20356, opened 2025-03-14
Bring Bank Account Address to SEPA File

dm-ps:
refs: https://github.com/metasfresh/metasfresh/issues/20353

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/banking/BankAccount.java
+++ b/backend/de.metas.business/src/main/java/de/metas/banking/BankAccount.java
@@ -2,6 +2,7 @@ package de.metas.banking;
 
 import de.metas.money.CurrencyId;
 import de.metas.organization.OrgId;
+import de.metas.util.Check;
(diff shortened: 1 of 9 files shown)


backend/de.metas.business/src/main/java/de/metas/banking/BankAccount.java
```diff
@@ -93,4 +106,12 @@ public boolean isAccountNoMatching(@NonNull final String accountNo)
 				|| postAcctNoCleaned.equals(SEPA_CreditorIdentifier);
 
 	}
+
+	public boolean isAddressComplete()
+	{
+		return Check.isNotBlank(accountStreet) &&
+				Check.isNotBlank(accountZip) &&
+				Check.isNotBlank(accountCity) &&
+				Check.isNotBlank(accountCountry);
```
metas-ts, line 115: not sure about accountCountry..might be empty if only domestic transactions are anticipated..

metas-ts (approved):

PR #20397, opened 2025-03-20
Fix problem where the "LocationChanged" error was always thrown

metas-ts:


metas-ts:
@mergifyio copy new_dawn_uat

PR #20416, opened 2025-03-21
Fix min quantity for qtyCUsPerTopLevelHU (#20412)

dm-ps:
* Fix min quantity for qtyCUsPerTopLevelHU refs: https://github.com/metasfresh/metasfresh/issues/20411


mrie04 (changes requested):
@dm-ps why cp instead of merge

dm-ps:
Will merge instead of CP, thx Max

PR #20484, opened 2025-03-28
use the new method "Fact_Acct records are matching" in the cucumber

metas-rc:


metas-rc:
Had to manually run the build . It was successful: https://github.com/metasfresh/metasfresh/actions/runs/14130888213

PR #20509, opened 2025-04-02
remove orphan invoice ad_archive and doc_outbound_log records

dm-ps:


dm-ps:
Not needed, run on the specific system

PR #20538, opened 2025-04-04
bugfix fresh_Attributes: override the attributes printed value only if the attribute has a non null value

cristinamghita:


teosarca (changes requested):
check my comments

teosarca (approved):

Issue #20558, opened 2025-04-08
Fix `Delivery to Address` in `Sales Order`


riadmen:
### Is this a bug or feature request?

- BUG

### What is the expected or desired behavior?

- Fix `Delivery to Address` in `Sales Order`

PR #20559, opened 2025-04-08
Fix `Delivery to Address` in `Sales Order` report description

riadmen:
https://github.com/metasfresh/metasfresh/issues/20558


backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/Docs_Sales_Order_Description.sql
```diff
@@ -74,7 +74,7 @@ SELECT o.description                             AS description,
        COALESCE(dtt.Description, dt.Description) AS dt_description,
        offer.documentno                          AS offer_documentno,
        REPLACE(
-               REPLACE(o.deliverytoaddress, E'\r\n', ' | '),
+               REPLACE(o.BPartnerAddress, E'\r\n', ' | '),
                E'\n', ' | '
```
cristinamghita, line 78: why?
the delivery address is the delivery address
If there is no dropship, shall be the same as BPArtneraddess
This shall not be fixed from SQL
See exactly why is not filled up as expected

cristinamghita (changes requested):
check my comments

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/Docs_Sales_Order_Description.sql
riadmen: - In the report description sub report, the `$F{deliverytoaddress}` should be visible only when `new Boolean($F{showdeliverytoaddress}.equals( "Y" ))`
- But when the sales order have `IsDropShip = N` we have `DropShip_Location_ID = NULL`
- That mean the actual logic in the function return `ShowDeliveryToAddress = Y`:
```SQL
CASE
           WHEN o.c_bpartner_location_id = o.dropship_location_id
               THEN 'N'
               ELSE 'Y'
       END                                       AS showdeliverytoaddress
```
- So IMHO, we nedd to update this logic to:
```SQL
o.IsDropShip                              AS ShowDeliveryToAddress 
```

cristinamghita (approved):
lgtm

Issue #20561, opened 2025-04-09
Fix `Zoom` into `PLV` from `Product Window`


riadmen:
### Is this a bug or feature request?

- BUG

### What is the expected or desired behavior?

- Fix `Zoom` into `PLV` from `Product Window`

PR #20562, opened 2025-04-09
Fix `Zoom` into `PLV` from `Product Window`

riadmen:
https://github.com/metasfresh/metasfresh/issues/20561


cristinamghita (approved):
lgtm

Issue #20592, opened 2025-04-14
Update `Weight` fields in `Product` window


riadmen:
### Is this a bug or feature request?

- FR

### What is the expected or desired behavior?

- Fix the german and english name for M_Product.Weight and call it "Net Weight (kg)"
- Make sure GrossWeight field is displayed near the M_Product.Weight field
- Make sure GrossWeight_UOM_ID field is displayed and is mandatory if GrossWeight is set and is not zero.

PR #20625, opened 2025-04-16
Rename `GrossWeight_UOM_ID` and adding weight validation rule

riadmen:
#20592


riadmen:
@teosarca validation rule renamed

teosarca (approved):

PR #20731, opened 2025-04-28
Don't use the bankAccount's A_Country, because we need an ISO-3166 text in this field

cristinamghita:


backend/de.metas.payment.sepa/base/src/main/java/de/metas/payment/sepa/model/validator/C_BP_BankAccount.java
```diff
@@ -74,4 +78,35 @@ public void validateIBAN(final I_C_BP_BankAccount bp_bankAccount)
...
 
+	@ModelChange(
+			timings = { ModelValidator.TYPE_BEFORE_CHANGE, ModelValidator.TYPE_AFTER_NEW },
+			ifColumnsChanged = { I_C_BP_BankAccount.COLUMNNAME_A_Country_ID}
+	)
+	public void setA_Country(final I_C_BP_BankAccount bp_bankAccount)
+	{
+		final int countryId = bp_bankAccount.getA_Country_ID();
+		if (countryId > 0)
+		{
```
metas-ts, line 89: shouldn'T we set the code to null if the country-id is set to null?

backend/de.metas.payment.sepa/base/src/main/java/de/metas/payment/sepa/model/validator/C_BP_BankAccount.java
```diff
@@ -74,4 +78,35 @@ public void validateIBAN(final I_C_BP_BankAccount bp_bankAccount)
...
+	}
+
+	@ModelChange(
+			timings = { ModelValidator.TYPE_BEFORE_CHANGE, ModelValidator.TYPE_AFTER_NEW },
+			ifColumnsChanged = { I_C_BP_BankAccount.COLUMNNAME_A_Country}
+	)
+	public void validateA_Country(final I_C_BP_BankAccount bp_bankAccount)
+	{
+		final String A_Country = bp_bankAccount.getA_Country();
+		if (StringUtils.trimBlankToNull(A_Country) != null)
```
metas-ts, line 102: i believe the code would be clearer with `Check.isBlank(A_Country)`

backend/de.metas.payment.sepa/base/src/main/java/de/metas/payment/sepa/model/validator/C_BP_BankAccount.java
```diff
@@ -74,4 +78,35 @@ public void validateIBAN(final I_C_BP_BankAccount bp_bankAccount)
...
+		{
+			final CountryId countryId = countryDAO.getCountryIdByCountryCodeOrNull(A_Country);
+			if (countryId == null)
```
metas-ts, line 105: how is the A_Country_ID set if if was found using the code?

metas-ts (commented):
i didn't manage to check the changes in `CountryDAO.java` yet.
going to do it later or on friday.
are you going to test the API manually? don'T we have existing cucmber-tests you could extend instead?

backend/de.metas.payment.sepa/base/src/main/java/de/metas/payment/sepa/model/validator/C_BP_BankAccount.java
cristinamghita: We do not set it back if it was via API, by setting A_Country.
We validate that A_Country is a country code, regardless of how it was set.

metas-ts (changes requested):
I think we need to extend the automated API-Test and clarify my quest wrt the C_BP_BankAccount.java`-MI

metas-ts (approved):

PR #20741, opened 2025-04-29
Business Rule: C_OrderLine require full LU - Use function when searching for the tuOnLU

metas-rc:


backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5753110_BusinessRule_C_OrderLine_FullLURequired_WithFunction.sql
```diff
@@ -0,0 +1,28 @@
...
+                       WHERE qtyenteredtu % tu.qty != 0)
+
+        /* Multiple M_HU_PI_Items exist for the TU on a current LU version, both fitting and not fitting the qty */
+        OR (
+        EXISTS (SELECT 1
+                FROM getTUItemsOnCurrentLUVersions(C_OrderLine.c_orderline_id) tu
+                WHERE qtyenteredtu % tu.qty != 0)
+            AND EXISTS (SELECT 1
+                        FROM getTUItemsOnCurrentLUVersions(C_OrderLine.c_orderline_id) tu
+                        WHERE qtyenteredtu % tu.qty = 0)
```
adi-stefan, line 26:
```suggestion
        OR NOT EXISTS (SELECT 1
                       FROM getTUItemsOnCurrentLUVersions(C_OrderLine.c_orderline_id) tu
                       WHERE qtyenteredtu % tu.qty != 0)

        /* Multiple M_HU_PI_Items exist for the TU on a current LU version, both fitting and not fitting the qty */
        OR EXISTS (SELECT 1
                        FROM getTUItemsOnCurrentLUVersions(C_OrderLine.c_orderline_id) tu
                        WHERE qtyenteredtu % tu.qty = 0)
```

this is `!A || (A & B)`, which can be simplified to `!A || B`

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5753110_BusinessRule_C_OrderLine_FullLURequired_WithFunction.sql
```diff
@@ -0,0 +1,28 @@
+-- Name: C_OrderLine Require Full LU
+-- 2025-04-29T20:35:20.696Z
+UPDATE AD_Val_Rule SET Code=' /* partner does not require a full LU */
```
adi-stefan, line 3: qtyenteredTU is nullable (and based on `de.metas.adempiere.gui.search.impl.HUPackingAwareBL#calculateQtyTU` null is ok on several scenarios). Should we also exclude those cases from throwing up a validation error?

adi-stefan (approved):

PR #20784, opened 2025-05-08
Implement accounting for Allocations between two PO-Invoices

metas-ts:


metas-ts:
this code is part of PR https://github.com/metasfresh/metasfresh/pull/20781

PR #20811, opened 2025-05-13
merge inner_silence_hotfix -> inner_silence_uat

dm-ps:


backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/log/ModularContractLogService.java
```diff
@@ -392,7 +397,13 @@ public void unprocessModularContractLogs(@NonNull final InvoiceId invoiceId, @No
...
 		 if (docTypeBL.isDefinitiveInvoiceOrDefinitiveCreditMemo(docTypeId))
 		 {
-			 flatrateBL.reverseDefinitiveInvoice(ImmutableSet.of(Check.assumePresent(flatrateBL.getIdByInvoiceId(invoiceId), "FlatrateTermId should be present")));
+			 final Set<FlatrateTermId> contractIds = invoiceBL.getLines(invoiceId)
+					 .stream()
+					 .map(org.compiere.model.I_C_InvoiceLine::getC_Flatrate_Term_ID)
+					 .map(FlatrateTermId::ofRepoIdOrNull) // invoice might have been imported from a csv-file and have no flatrate term 
+					 .filter(Objects::nonNull)
+					 .collect(Collectors.toSet());
+			 flatrateBL.reverseDefinitiveInvoice(contractIds);
```
mrie04, line 406: Why drop `flatrateBL.getIdByInvoiceId(invoiceId)` which does the same and also ensures that we don't have multiple Ids in the same Invoice?

mrie04 (changes requested):

backend/de.metas.contracts/src/main/java/de/metas/contracts/modular/log/ModularContractLogService.java
mrie04: nvm, `de.metas.util.collections.CollectionUtils#extractDistinctElements` excludes nulls, which I didn't notice

mrie04 (approved):

PR #20839, opened 2025-05-19
Introduce M_Product.Commentary

dm-ps:


metas-ts (approved):
not superhappy with the naming, because `Compentatry` doesn't imply that is about the discontinuing, but i think the extra-effort to change it now is too much. please just consider in next time

PR #20897, opened 2025-06-01
Fix preloaded db image

mrie04:


mrie04:
@Mergifyio copy inner_silence_uat

PR #20927, opened 2025-06-10
keep single Purchase_Candidate.Price_UOM_ID

dm-ps:


backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5757190_sys_gh23706_keep_single_Purchase_Candidate_Price_UOM_ID.sql
```diff
@@ -0,0 +1,47 @@
...
+;
```
mrie04, line 18: This will fail on systems, where `5742770_sys_gh19759_C_PurchaseCandidate_Price_UOM_ID` was executed

mrie04 (changes requested):
pls check my comments

PR #20982, opened 2025-06-19
Make `IsSystemUser = Y` as default in `User` window filters

riadmen:
https://github.com/metasfresh/me03/issues/23878


cristinamghita (approved):
lgtm

PR #20998, opened 2025-06-24
Allow postgrest-processes to export exactly one json-object, not an array

metas-ts:


metas-ts (commented):
LGTM, but github wouldn't let me approve my own PR

PR #21015, opened 2025-06-27
Adjust paymentterm unique index

mrie04:


backend/de.metas.business/src/main/java/de/metas/payment/paymentterm/impl/PaymentTermRepository.java
```diff
@@ -138,14 +138,15 @@ public Optional<PaymentTermId> retrievePaymentTermId(@NonNull final PaymentTermQ
 			final IQueryBuilder<I_C_BPartner> tmp = queryBL
 					.createQueryBuilder(I_C_BPartner.class)
 					.addEqualsFilter(I_C_BPartner.COLUMN_C_BPartner_ID, query.getBPartnerId());
-			if (query.getSoTrx() != null && query.getSoTrx().isPurchase())
+			if (query.getSoTrx().isPurchase())
 			{
```
cristinamghita, line 142: @mrie04 why did you check null checking?
SoTrx could be null. For example, in this case: `de.metas.rest_api.v1.ordercandidates.impl.MasterdataProvider#getPaymentTermId`

cristinamghita (changes requested):
check my comment, pls

cristinamghita (approved):
lgtm

PR #21067, opened 2025-07-08
All UOMs field are ordered alphabetically

riadmen:
https://github.com/metasfresh/me03/issues/24166


metas-rc:
@riadmen 

To be on the safe side, I suggest creating a new script with a lower ID (e.g. ``5760299_sys_gh24166_AD_Column_Backup.sql``) where to call ``select table_backup('AD_Column');`` just in case we want to see later what references we replaced.

metas-rc (approved):

PR #21120, opened 2025-07-22
Set Shipper for each delivery address / define a Leitweg Nr

cristinamghita:


backend/de.metas.business/src/main/java/de/metas/inout/impl/InOutBL.java
```diff
@@ -804,4 +805,30 @@ public ImmutableSet<I_M_InOut> getNotVoidedNotReversedForOrderId(@NonNull final
...
+	{
+
+		if (inout.getDropShip_BPartner_ID() > 0 && inout.getDropShip_Location_ID() > 0)
+		{
+			final Optional<ShipperId> deliveryAddressShipperId = bpartnerDAO.getShipperIdByBPLocationId(BPartnerLocationId.ofRepoId(inout.getDropShip_BPartner_ID(), inout.getDropShip_Location_ID()));
+			if (deliveryAddressShipperId.isPresent())
+			{
+				return deliveryAddressShipperId.get(); // we are done
+			}
+		}
```
mrie04, line 826: Is the dropship location and partner always set or isn't it set if we use the default ship to address of the order? How do we get the location specific shipper then?

backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
```diff
@@ -494,6 +494,8 @@ public void onDropShipLocation(final I_C_Order order)
 		}
 
 		orderBL.setPriceList(order);
+
+		orderBL.setShipperId(order);
```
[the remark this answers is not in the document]
cristinamghita: @mrie04 yes, is enough, because if the partner is changed, then the location is changed and the validator is triggered

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
```diff
@@ -1359,4 +1360,26 @@ public String getDescriptionById(@NonNull final OrderId orderId)
...
+	private ShipperId findShipperId(@NonNull final I_C_Order orderRecord)
+	{
+		if (orderRecord.getDropShip_BPartner_ID() > 0 && orderRecord.getDropShip_Location_ID() > 0)
+		{
+			final Optional<ShipperId> deliveryAddressShipperId = partnerDAO.getShipperIdByBPLocationId(BPartnerLocationId.ofRepoId(orderRecord.getDropShip_BPartner_ID(), orderRecord.getDropShip_Location_ID()));
+			if (deliveryAddressShipperId.isPresent())
+			{
+				return deliveryAddressShipperId.get(); 
+			}
+		}
```
mrie04, line 1379: same as on inout

backend/de.metas.business/src/main/java/de/metas/inout/impl/InOutBL.java
cristinamghita: if the Dropship location is not set, we get the shipper from dropship partner; this was the requirement.

PR #21173, opened 2025-08-07
Collect LU packing materials into order lines, cap OrderLine.QtyLU via sysconfig. Allow editing OrderLine.QtyLU.

adi-stefan:


mrie04 (changes requested):
pls check my comments

PR #21199, opened 2025-08-21
Make sure Order Location callout methods are executed in the correct order

dm-ps:
Make sure Order Location callout methods are executed in the correct order

### DIFF
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/document/location/RecordBasedLocationAdapter.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/document/location/RecordBasedLocationAdapter.java
@@ -32,6 +32,8 @@ public interface RecordBasedLocationAdapter<SELF extends RecordBasedLocationAdap
 
 	void setRenderedAddressAndCapturedLocation(RenderedAddressAndCapturedLocation from);
 
+	void setRenderedAddress(RenderedAddressAndCapturedLocation from);
+
 	Optional<DocumentLocation> toPlainDocumentLocation(IDocumentLocationBL documentLocationBL);
 
 	SELF toOldValues();
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/document/location/adapter/IDocumentBillLocationAdapter.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/document/location/adapter/IDocumentBillLocationAdapter.java
@@ -33,7 +33,6 @@ import de.metas.location.LocationId;
 import lombok.NonNull;
 import org.adempiere.exceptions.AdempiereException;
 
-import javax.annotation.Nullable;
 import java.util.Optional;
 
 public interface IDocumentBillLocationAdapter extends IDocumentLocationAdapterTemplate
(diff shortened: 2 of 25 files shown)


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/order/callout/C_Order.java
```diff
@@ -197,12 +192,13 @@ public void setContactDetails(final I_C_Order order)
...
 		}
 	}
+
 	@CalloutMethod(columnNames = {
 			I_C_Order.COLUMNNAME_M_Shipper_ID },
 			skipIfCopying = true)
 	public void updateDeliveryViaRule(final I_C_Order order)
 	{
-		if(order.getM_Shipper_ID() > 0)
+		if (order.getM_Shipper_ID() > 0)
```
mrie04, line 201: pls use ShipperId.ofRepoIdOrNull

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/order/callout/C_Order.java
```diff
@@ -101,7 +96,7 @@ public void updateBPartnerAddressForceUpdateCapturedLocation(final I_C_Order ord
 	{
 		documentLocationBL.updateCapturedLocation(OrderDocumentLocationAdapterFactory.locationAdapter(order));
 
-		orderBL.setM_PricingSystem_ID(order, true); // changed partner/location might imply a changed PS/PL
+		updateRenderedAddressCapturedLocationAndPricingSystem(order);
```
mrie04, line 99: I think `de.metas.document.location.impl.DocumentLocationBL.updateRenderedAddressAndCapturedLocation(de.metas.document.location.adapter.IDocumentLocationAdapter)` should be changed to
```java
@Override
	public void updateRenderedAddressAndCapturedLocation(final IDocumentLocationAdapter locationAdapter)
	{
		toPlainDocumentLocation(locationAdapter)
				.map(this::withUpdatedLocationId)
				.map(this::computeRenderedAddress) //newly added
				.ifPresent(locationAdapter::setRenderedAddressAndCapturedLocation);
	}
```
So it is doing what the function name implies. Then we only need this + pricingSystem and don't need to call documentLocationBL twice.

I think the same applies for other Location types. And as the Problem from my understanding is in this function, the same Problem should be present in other documents, as it is used there in callouts too

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/order/callout/C_Order.java
```diff
@@ -81,16 +82,10 @@ public void setDropShipWarehouse(final I_C_Order order, final ICalloutField fiel
...
-			I_C_Order.COLUMNNAME_C_BPartner_Location_ID,
-			I_C_Order.COLUMNNAME_AD_User_ID },
-			skipIfCopying = true)
+	@CalloutMethod(columnNames = I_C_Order.COLUMNNAME_AD_User_ID, skipIfCopying = true)
 	public void updateBPartnerAddress(final I_C_Order order)
 	{
-		documentLocationBL.updateRenderedAddressAndCapturedLocation(OrderDocumentLocationAdapterFactory.locationAdapter(order));
-		
-		orderBL.setM_PricingSystem_ID(order, true); // changed partner/location might imply a changed PS/PL
+		updateRenderedAddressCapturedLocationAndPricingSystem(order);
```
mrie04, line 88: Consider introducing documentLocationBL.updateRenderedAddress AFAIS this is the only relevant part here

mrie04 (approved):
lgtm

PR #21211, opened 2025-08-26
`HU Display` reports correction

riadmen:


backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/Docs_Sales_InOut_Root.sql
```diff
@@ -32,44 +29,41 @@ $$SELECT
 		THEN dt.printname
 		ELSE COALESCE(dtt.printname,dt.printname)
 	END AS printname,
-	CASE WHEN io.DocStatus IN ('DR', 'IP') THEN 'de_CH' ELSE $2 END AS AD_Language,
+	CASE WHEN io.DocStatus IN ('DR', 'IP') THEN 'de_CH' ELSE p_ad_language END AS AD_Language,
 	isHidePackingMaterialInShipmentPrint,
```
cristinamghita, line 33: when we introduce this?
and why?
does not make any sense!
How DocStatus is linked with language?
And why do we need ad_language to be returned here?

cristinamghita (changes requested):
check my comments

PR #21311, opened 2025-09-15
don't update the invoice document number on doctype change when the same sequence is used

metas-rc:


metas-ts (approved):

metas-rc:
@Mergifyio copy science_vessel_release

PR #21345, opened 2025-09-22
Cucumbers for API Endpoint to receive response by requestId

dm-ps:


metas-ts (approved):
ok, what i had in mind - and should have written, sorry - was to use `StringExpressionCompiler` like `de.metas.cucumber.stepdefs.REST_API_StepDef#resolveContextVariables` does.

But we can do that later.
Please make a note and change it accordingly in a future issue - there are more to come.

PR #21346, opened 2025-09-22
New dawn uat purchase cand date ordred

metas-rc:


metas-ts (commented):
lgtm so far. just please also include the other thing we wrote about

metas-ts (approved):

PR #21349, opened 2025-09-23
Take warehouse partner as supplier in edi invoic export

dm-ps:
if the warehouse from which the goods are sent has a partner and location set, then we need to use that partner as supplier


metas-ts (approved):

metas-ts:

PR #21360, opened 2025-09-24
Allow shipment-scheds with reverted shipment to be shipped again

dm-ps:
Fix M_Packageable_v and use the same EnqueuerQueryFilters

### DIFF
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/sql/postgresql/ddl/de_metas_inoutcandidate/M_Packageable_v.sql
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/sql/postgresql/ddl/de_metas_inoutcandidate/M_Packageable_v.sql
@@ -151,7 +151,7 @@ FROM (SELECT
       WHERE TRUE
         AND s.Processed = 'N'
         AND s.IsActive = 'Y'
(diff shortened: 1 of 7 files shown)


backend/de.metas.swat/de.metas.swat.base/src/main/sql/postgresql/ddl/de_metas_inoutcandidate/M_Packageable_v.sql
```diff
@@ -151,7 +151,7 @@ FROM (SELECT
       WHERE TRUE
         AND s.Processed = 'N'
         AND s.IsActive = 'Y'
-        AND s.QtyToDeliver > 0
+        AND s.QtyToDeliver > 0 OR s.qtypicklist > 0
```
metas-ts, line 154:
```suggestion
        AND (s.QtyToDeliver > 0 OR s.qtypicklist > 0)
```

metas-ts (changes requested):
please check my suggestions.
also, pls consult with teo on whether `M_Packageable_v` is used in the mobileUI and now fully picked scheds show up in there although they shouldn't

metas-ts (approved):

PR #21394, opened 2025-09-30
Stock/Availability infos extend with warehouse code

dm-ps:


backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/modelvalidator/M_ShipmentSchedule.java
```diff
@@ -216,6 +220,8 @@ private boolean isExtraSyncNeededForOldShipmentSchedule(
 			@NonNull final I_M_ShipmentSchedule oldShipmentScheduleRecord)
 	{
 		return oldShipmentScheduleRecord.getM_Product_ID() != currentShipmentScheduleRecord.getM_Product_ID()
+				|| oldShipmentScheduleRecord.getM_Warehouse_Override_ID() != currentShipmentScheduleRecord.getM_Warehouse_Override_ID()
```
metas-ts, line 223: AFAIU this is not right..we need to compare the "coalesce" of the two schedules' warehouse/warehouse_orderrides

metas-ts (approved):

PR #21457, opened 2025-10-09
Fix ColumnSQLs for M_ReceiptsSchedule

metas-ts:


metas-rc:
@Mergifyio copy memorable_shiny_hotfix

PR #21546, opened 2025-10-27
Prevent creating duplicate M_ShippingPackages when PO is reactivated/completed.

adi-stefan:


backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationRepository.java
```diff
@@ -185,4 +186,43 @@ private static PackageItem toPackageItem(@NonNull final I_M_InOutLine inOutLine)
 				.build();
 	}
 
+	@Nullable
+	public ShipperTransportationId getByQueryOrNull(@NonNull final ShipperTransportationQuery query)
```
metas-ts: not entirely sure what this class is or should be...anyways i think this method should be in IShipperTransportationDAO. If you move it there, please make sure from the name and/or javadoc if it's about all orders or just purchase-orders.
And while i think it's ok to use a query object, better don't introduce and support query-fields that we currently don't even need. after you removed the unsued bits, please decide if you still need a query-object at this point.

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
```diff
@@ -159,19 +159,27 @@ private void addPurchaseOrderToShipperTransportation(final @NonNull org.compiere
...
-			repo.addPurchaseOrderToShipperTransportation(requestTemplate
-					// .sscc(sscc18CodeBL.generate(orgId)) //No requirements currently ask for this
+			final ShipperTransportationId existingShipperTransportationId = repo.getByQueryOrNull(ShipperTransportationQuery.builder()
+					.orderId(orderId)
+					.onlyRecordsWithNullOrderLineId(true)
 					.build());
+			if (existingShipperTransportationId == null)
+			{
+				//create a generic package for all order lines without LUQty set on them
+				repo.addPurchaseOrderToShipperTransportation(requestTemplate
```
metas-ts: this method is called a 2nd time inside thiw service - in another method. do we need some checking there as well?

metas-ts (commented):
pls check my request&question

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
adi-stefan: No, we don't. There we're already handling the difference in expected vs actual M_ShippingPackages.

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationRepository.java
adi-stefan: Renamed to `PurchaseShippingPackageQuery` to better reflect its usage. Also simplified a bit the logic, since we're interested in a yes/no answer from the repo.

metas-ts (approved):

PR #21553, opened 2025-10-28
Scripted Import Conversion Route

dm-ps:


metas-ts (changes requested):
the code lgtm, but please add tests

misc/services/camel/de-metas-camel-externalsystems/de-metas-camel-scriptedadapter/src/main/java/de/metas/camel/externalsystems/scriptedadapter/convertmsg/to_mf/ScriptedImportConversionDynamicRouteBuilder.java
```diff
@@ -0,0 +1,168 @@
...
+			final CamelServiceRouteIdWithRequestType camelRouteIdWithRequestType = CamelServiceRouteIdWithRequestType.ofRouteId(request.getCamelServiceRouteID());
+			final Object payload = JsonObjectMapperHolder.sharedJsonObjectMapper()
+					.readValue(request.getRequestBody(), camelRouteIdWithRequestType.getRequestType());
+
+			final String response = producerTemplate.requestBody(resolveCamelEndpointUri(camelRouteIdWithRequestType), payload, String.class);
+			exchange.getMessage().setBody(response);
+		}
+		catch (final Exception e)
+		{
+			exchange.getMessage().setBody(e.getMessage());
```
metas-ts, line 110: i'm not 100% sure how it will look in practice.
if an exception is caught. it probably doesn't make sense to clotter the console with a stacktrace, but there should be a log-message to help a supported see which endpoint had the problem. if this as already happening (maybe `stopOnException()` makes sure outputs?), then please add a comment to that effect

misc/services/camel/de-metas-camel-externalsystems/de-metas-camel-scriptedadapter/src/main/java/de/metas/camel/externalsystems/scriptedadapter/convertmsg/to_mf/ScriptedImportConversionRestAPIRouteBuilder.java
```diff
@@ -0,0 +1,354 @@
...
+				.process(exchange -> this.prepareExternalStatusCreateRequest(exchange, JsonExternalStatus.Inactive)).id(DISABLE_PREPARE_EXTERNAL_STATUS_CREATE_REQ_PROCESSOR_ID)
+				.to("{{" + ExternalSystemCamelConstants.MF_CREATE_EXTERNAL_SYSTEM_STATUS_V2_CAMEL_URI + "}}")
+				.end();
+		//@formatter:on
+	}
+
+	private void prepareResponse(@NonNull final Exchange exchange)
+	{
+		@SuppressWarnings("unchecked") final List<Object> list = exchange.getIn().getBody(List.class);
+		if (list.size() == 1)
```
metas-ts, line 174: pls verify that the list i not null, 
so if it is, we have a better starting point for the troubleshooting

metas-ts (commented):
check additional requests

misc/services/camel/de-metas-camel-externalsystems/de-metas-camel-scriptedadapter/src/main/java/de/metas/camel/externalsystems/scriptedadapter/convertmsg/to_mf/ScriptedImportConversionDynamicRouteBuilder.java
dm-ps: added a log statement

metas-ts (approved):

PR #21564, opened 2025-10-30
Fix Product - CU-TU allocation - Zoom

dm-ps:


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/ad_reference/ADReferenceService.java
```diff
@@ -209,6 +217,9 @@ private ADRefTable retrieveTableDirectRefInfo(@NonNull final String columnName)
 				.keyColumn(keyColumn)
 				.autoComplete(autoComplete)
 				.tooltipType(tooltipType)
+				.zoomSO_Window_ID(soWindowId)
+				.zoomPO_Window_ID(poWindowId)
+				.zoomAD_Window_ID_Override(CoalesceUtil.coalesce(soWindowId, poWindowId))
```
metas-ts, line 223: i don't understand the fix, particularly not this line. 
is this a port from another branch?

dm-ps: no, it is not a port
the field from which we zoom in is `M_HU_PI_Item_ID` in table `M_HU_PI_Item_Product`, but we don't have a window for table `M_HU_PI_Item`, because it is a sub-tab in Packing Instruction Version window (ID `540344`). However, in the `AD_Table` in the record for `M_HU_PI_Item`, we have `M_HU_PI_Item.AD_Window_ID` set to  `540344`.
Also setting the `zoomAD_Window_ID_Override` in this case (as the first nonNull value between these two), because that is the one used when building SqlLookupDescriptor `de.metas.ui.web.window.descriptor.sql.SqlLookupDescriptorFactory#build` and the value that ends up in `de.metas.ui.web.window.descriptor.sql.SqlLookupDescriptor#zoomIntoWindowId`

metas-ts (approved):

PR #21671, opened 2025-11-13
Make sure that the correct payment term is used in cucumber, regardless of the customer branch

dm-ps:


backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/report/baseandcustom/report.feature
```diff
@@ -119,9 +119,13 @@ Feature: Jasper Report Tests
...
       | dunningLevel_S0471_200 | dunning_S0471_200 | 0            |
+    # dev-note: make sure that the correct payment term is used, regardless of the branch
+    And metasfresh contains C_PaymentTerm
+      | Identifier            |
+      | paymentTerm_S0471_200 |
     And update C_BPartner:
-      | Identifier | C_Dunning_ID      |
-      | customer   | dunning_S0471_200 |
+      | Identifier | C_Dunning_ID      | C_PaymentTerm_ID      |
+      | customer   | dunning_S0471_200 | paymentTerm_S0471_200 |
```
mrie04: As the partner is created on Background I would prefer to either set it there or overwrite it on order

mrie04 (changes requested):
pls check my comment

mrie04 (approved):
lgtm

PR #21672, opened 2025-11-13
Make asyncBatchId-propagation optional in OLCand=>Order processing and fix C_Async_Batch stats updates

metas-ts:


metas-ts (commented):
checked all classes once more. if ther is a probem in new_dawn_uat i shall address it

PR #21686, opened 2025-11-14
Tax exempt: support tax exempt inherithed from C_BParnter.IsTaxExempt (sales side)

cristinamghita:


backend/de.metas.business/src/main/java/de/metas/order/impl/OrderLineBL.java
```diff
@@ -1033,4 +1044,58 @@ public Optional<BPartnerId> getBPartnerId(@NonNull final OrderAndLineId orderLin
...
+
+		final boolean isSOTrx = orderRecord.isSOTrx();
+		final Timestamp taxDate = orderLine.getDatePromised();
+
+		final BPartnerId effectiveBillPartnerId = orderBL().getEffectiveBillPartnerId(orderRecord);
+		final Boolean isTaxExempt;
+		if (effectiveBillPartnerId != null)
+		{
+			final I_C_BPartner billBPartnerRecord = bpartnerDAO.getById(effectiveBillPartnerId);
+			isTaxExempt = billBPartnerRecord.isTaxExempt();
```
metas-ts: We can't **dissallow** tax-exempt `C_Taxes` in case the `C_B_Partner` has the flag not set.
Y => filter for tax-exempt tax
N => like null, do not filter

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/tax/orderTaxCalculation.feature
```diff
@@ -40,18 +41,21 @@ Feature: Validate tax calculation for orders taking into account dropship locati
       | pp_Switzerland_Purchase | plv_Switzerland_Purchase          | product_S0151           | 10.0     | PCE               | Normal                        |
       | pp_Netherlands_Purchase | plv_Netherlands_Purchase          | product_S0151           | 10.0     | PCE               | Normal                        |
     And metasfresh contains C_BPartners without locations:
-      | Identifier   | Name                 | M_PricingSystem_ID.Identifier | OPT.IsVendor | OPT.IsCustomer | OPT.AD_Org_ID.Identifier | OPT.AD_OrgBP_ID.Identifier |
-      | bpartner_1   | BPartnerNameS0151    | ps_1                          | Y            | Y              | switzerland_org          |                            |
-      | org_bpartner | OrgBPartnerNameS0151 | ps_1                          |              |                | switzerland_org          | switzerland_org            |
+      | Identifier   | Name                 | M_PricingSystem_ID | IsVendor | IsCustomer | AD_Org_ID       | AD_OrgBP_ID     | IsTaxExemt |
```
metas-ts: Typo: should be IsTaxExempt

metas-ts (changes requested):

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderLineBL.java
metas-ts: **Also** AFAIU the `isTaxExempt ` may be used only if the order is a sales order!
Adding isSOTrx to the query is not enough because there might be C_Taxes with this property being empty

metas-ts:
```suggestion
			// Only set if TRUE - otherwise leave null (don't filter)
			isTaxExempt = billBPartnerRecord.isTaxExempt() ? Boolean.TRUE : null;
```

metas-ts (approved):

PR #21695, opened 2025-11-15
 Improve performance of async-workpackage poll-and-lock

metas-ts:


metas-ts (commented):
lgtm. ran a loadtest on a dev-system that created Orders and therefore a lot of WPs. manually created an invoice via webui

PR #21728, opened 2025-11-19
Set payment rule from Bill partner

cristinamghita:


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/impl/BPartnerBL.java
```diff
@@ -729,19 +729,30 @@ public Optional<PaymentTermId> getPaymentTermIdForBPartner(@NonNull final BPartn
 
 	@Override
 	public Optional<PaymentRule> getPaymentRuleForBPartner(@NonNull final BPartnerId bpartnerId)
```
metas-ts: who is using this method? maybe it make sense to change them to use the new method, so they need to be concous of the IsSOTrx?

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/impl/BPartnerBL.java
```diff
@@ -729,19 +729,30 @@ public Optional<PaymentTermId> getPaymentTermIdForBPartner(@NonNull final BPartn
...
+	{
+
+		return getPaymentRuleForBPartner(bpartnerId, true);
+	}
+
+	@Override
+	public Optional<PaymentRule> getPaymentRuleForBPartner(@NonNull final BPartnerId bpartnerId, final boolean isSOTrx)
```
metas-ts: we have a  enum for SOTrx, better use that

backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
```diff
@@ -253,26 +254,24 @@ public void setDeliveryViaRule(final I_C_Order order)
 	}
 
 	/**
-	 * When creating a manual order: The new order must inherit the payment rule from the BPartner.
+	 * When creating a manual order: The new order must inherit the payment rule from the BillPartner.
```
metas-ts:
```suggestion
	 * When creating a manual order: The new order must inherit the payment rule from the BillPartner (or BPartner if no BillPartner is set).
```

backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
```diff
@@ -282,6 +281,8 @@ else if (!order.isSOTrx() && bpartner != null && bpartner.getPaymentRulePO() !=
 		order.setPaymentRule(paymentRule.getCode());
 	}
 
+
+
```
metas-ts:
```suggestion
```

metas-ts (changes requested):

metas-ts (approved):

teosarca (approved):

PR #21747, opened 2025-11-21
Introduces the IsUpdateProductCategory field to the I_Product table and model, allowing selective updating of product category during import.

cristinamghita:


backend/de.metas.business/src/main/java/de/metas/product/impexp/ProductImportHelper.java
```diff
@@ -116,9 +116,14 @@ private void updateExistingProductNotSave(@NonNull final I_M_Product product,
 			product.setSKU(from.getSKU());
 		}
 
-		// Set UOM, product category, and classification
+		// set product category is wanted
+		if (from.isUpdateProductCategory())
```
metas-ts, line 120: so if isUpdateProductCategory=N, then new products get the category and exisitng ones don't, correct?

metas-ts (approved):

backend/de.metas.business/src/main/java/de/metas/product/impexp/ProductImportHelper.java
cristinamghita: Exactly, by default, it is Y, to preserve the regular behaviour

PR #21828, opened 2025-12-05
Filter Historical Invoices and Historical Shipments by DocType_Base

dm-ps:


mrie04 (approved):
lgtm, but consider adding documents that actually not matching the filter

PR #21893, opened 2025-12-17
Partner Product and Invoice Full-Text Search with Postgres FTS and pgtrgm (ngram) combined

mrie04:


teosarca (approved):
scrolled over, lgtm

PR #21894, opened 2025-12-17
Add missing SEPA_Export_Line_Ref SQL

dm-ps:


dm-ps:
Solved with: https://github.com/metasfresh/metasfresh/pull/21895
Such that in case the version before this one was already deployed on a system, the new file is going to run on that system and ignored the script on the system I manually ran the SQL

PR #21962, opened 2026-01-11
Fix postgresql operators to work with postgresql 17

riadmen:


metas-ts (approved):
i don't think it makes sense to create these permanet functions...I would have rather imagined anonymous function-blocks. 
but ok for this time

PR #21973, opened 2026-01-12
Added `ProjectValue` M_Attribute and populate it with  C_Project.Value of C_OrderLine (both SO and PO)

adi-stefan:


backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
```diff
@@ -1489,4 +1496,26 @@ public List<I_C_Order> getByQueryFilter(final IQueryFilter<I_C_Order> queryFilte
...
+	@Override
+	public void updateASIFromProjectId(@NonNull final I_C_OrderLine orderLine)
+	{
+		AttributeSetInstanceId asiId = AttributeSetInstanceId.ofRepoIdOrNull(orderLine.getM_AttributeSetInstance_ID());
+		if (asiId == null || asiId.isNone())
+		{
+			final I_M_AttributeSetInstance asi = attributeSetInstanceBL.createASI(ProductId.ofRepoId(orderLine.getM_Product_ID()));
+			asiId = AttributeSetInstanceId.ofRepoId(asi.getM_AttributeSetInstance_ID());
+			orderLine.setM_AttributeSetInstance_ID(asiId.getRepoId());
+		}
```
teosarca, line 1514: IMHO in case the ASI exists, shouldn't u clone it?
asking because ASIs are conceptually immutable.

adi-stefan: changed API to clone ASI when setting the new value.

teosarca (approved):

PR #21982, opened 2026-01-14
fix(db): PostgreSQL 17 compatibility - remove custom minus operators

metas-ts:
## Summary

- Removes custom minus operators from `public` schema that block `pg_upgrade` to PostgreSQL 17
- Replaces operator usage in `c_invoice_candidate_v` with explicit `subtractdays()` function calls
- Adds helper functions for discovering and fixing similar issues in other databases

## Problem

Custom operators in the `public` schema that override the built-in `-` operator cannot be migrated by `pg_upgrade` to PostgreSQL 17:

| Operator | Left Type | Right Type | Function |
|----------|-----------|------------|----------|
| `-` | `numeric` | `timestamptz` | built-in |
| `-` | `numeric` | `interval` | built-in |
| `-` | `timestamptz` | `numeric` | `subtractdays()` |
| `-` | `interval` | `numeric` | `subtractdays()` |

## Solution

The migration script (`5783690_sys_pg17_remove_custom_minus_operators.sql`):

1. Installs helper functions for discovery
2. Checks if custom operators exist
3. Fixes `c_invoice_candidate_v` by replacing `(timestamp + offset) - 1` with `subtractdays(timestamp + offset, 1)`
4. Drops all 4 custom minus operators
5. Verifies all operators are removed

## Helper Functions (DDL)

| Function | Purpose |
|----------|---------|
| `migration_find_all_affected_objects()` | Discovers views/functions using custom operators |
| `migration_generate_view_fix(schema, name)` | Generates fix SQL for a specific view |

## Test plan

- [x] Validated helper functions against `new_dawn_uat` database
- [x] Verified only `c_invoice_candidate_v` requires fixing (no false positives)
- [x] Confirmed generated fix replaces all 6 operator usages with `subtractdays()` calls
- [ ] Run migration on test database
- [ ] Verify `c_invoice_candidate_v` works after migration
- [ ] Verify `pg_upgrade` preflight check passes

🤖 Generated with [Claude Code](https://claude.ai/code)


metas-ts:
@mergifyio copy inner_silence_hotfix

PR #21984, opened 2026-01-14
new_dawn_uat_merge_keen_hawn_hotfix

metas-rc:


backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5783630_Remove_Obsolete_PCE_UOMs.sql
```diff
@@ -0,0 +1,182 @@
...
+        END IF;
+    END
+$$
+;
+
+DELETE FROM c_uom_trl WHERE c_uom_id IN (SELECT c_uom_id FROM deprecated_pce_uoms);
+
+
+DELETE FROM c_uom WHERE c_uom_id IN (SELECT c_uom_id FROM deprecated_pce_uoms);
+
```
metas-ts, line 182: How about adding
```sql
CREATE UNIQUE INDEX c_uom_x12de355_uc
ON c_uom (x12de355)
WHERE x12de355 IN ('PCE', 'KGM');
```

because AFAIU at least *those* are fine now, right?

metas-rc: Yes! Thank you for the good idea.

PR #22007, opened 2026-01-16
Add Supply vs. Demand Product Info report and function 

cristinamghita:
Introduces a new SQL function 'SupplyVsDemand_Product_Info' for comparative analysis of product supply and demand, including vendor and cost price details. Adds migration scripts to register the report and its parameters in the application, supporting export to Excel and translations for multiple languages.


metas-ts (approved):
pls address the things we talked about. then it's good to go

PR #22013, opened 2026-01-17
Fix Product costing when dealing with included tax prices and catch weight

teosarca:
- **minor QA. Refactor variable usage in `InOutBL`. Replaced inline method call for `getQtyDeliveredCatch` with a local variable to improve readability and reuse. Adjusted references accordingly within `getStockQtyAndCatchQty`.**
- **Minor QA. Add `@NonNull` annotations to `InOutAndLineId` fields**
- **Implement support for `StockQtyAndUOMQty` in costing logic**


teosarca:

PR #22049, opened 2026-01-20
Extend Import GL Journal

cristinamghita:


backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportTableSqlUpdater.java
```diff
@@ -0,0 +1,908 @@
...
+
+	/**
+	 * Set OrgTrx from OrgTrxValue
+	 */
+	private void setOrgTrx()
+	{
+		// Set OrgTrx from Value
+		StringBuilder sql = new StringBuilder("UPDATE I_GLJournal i "
+				+ "SET AD_OrgTrx_ID=(SELECT o.AD_Org_ID FROM AD_Org o"
+				+ " WHERE o.Value=i.OrgValue AND o.IsSummary='N' AND i.AD_Client_ID=o.AD_Client_ID) "
```
metas-ts: Should this  be
```suggestion
				+ " WHERE o.Value=i.OrgTrxValue AND o.IsSummary='N' AND i.AD_Client_ID=o.AD_Client_ID) "
```
?
or maybe a coalesce?

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcess.java
```diff
@@ -842,6 +346,79 @@ protected ImportRecordResult importRecord(@NonNull final IMutable<Object> state,
 		line.setQty(importRecord.getQty());
 		//
 
+		// Set/Get Tax Account Combination From (Debit)
+		if (importRecord.getDR_Tax_Acct_ID() > 0 && importRecord.getC_ValidCombinationTaxFrom_ID() == 0)
+		{
+			final AccountDimension taxAcctDim = newMinimalAccountDimension(importRecord, importRecord.getDR_Tax_Acct_ID());
+			final MAccount taxAcct = MAccount.get(getCtx(), taxAcctDim);
+			if (taxAcct.get_ID() == 0)
```
teosarca: AFAIK, MAccount.get is also saving it so no need for saving.
anyways, i would prefer using saveEx

metas-ts (dismissed):
also see the review-DM and and consider which are valid suggestions

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcess.java
```diff
@@ -842,6 +346,79 @@ protected ImportRecordResult importRecord(@NonNull final IMutable<Object> state,
...
+				InterfaceWrapperHelper.save(importRecord);
+			}
+		}
+
+		// Set/Get Tax Account Combination To (Credit)
+		if (importRecord.getCR_Tax_Acct_ID() > 0 && importRecord.getC_ValidCombinationTaxTo_ID() == 0)
+		{
+			final AccountDimension taxAcctDim = newMinimalAccountDimension(importRecord, importRecord.getCR_Tax_Acct_ID());
+			final MAccount taxAcct = MAccount.get(getCtx(), taxAcctDim);
+			if (taxAcct.get_ID() == 0)
```
teosarca: same

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcess.java
```diff
@@ -842,6 +346,79 @@ protected ImportRecordResult importRecord(@NonNull final IMutable<Object> state,
...
+			}
+			else
+			{
+				importRecord.setC_ValidCombinationTaxTo_ID(taxAcct.get_ID());
+				InterfaceWrapperHelper.save(importRecord);
+			}
+		}
+
+		// Add tax accunt and amounts if exists
+		if ((importRecord.getDR_TaxTotalAmt() != null && importRecord.getDR_TaxTotalAmt().signum() != 0)
```
teosarca: AFAIK getDR_TaxTotalAmt never returns null

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcessContext.java
```diff
@@ -0,0 +1,61 @@
...
+import lombok.Value;
+
+import javax.annotation.Nullable;
+import java.sql.Timestamp;
+
+@Value
+@Builder
+public class GLJournalImportProcessContext
+{
+	int adClientId;
```
teosarca: pls use AdClientId, AdOrgId, AcctSchemaId, Instant dateAcct.
also add proper NonNull/Nullable annotations

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcessContext.java
```diff
@@ -0,0 +1,61 @@
...
+	{
+		return GLJournalImportProcessContext.builder()
+				.adClientId(parameters.getParameterAsInt("AD_Client_ID", -1))
+				.adOrgId(parameters.getParameterAsInt("AD_Org_ID", -1))
+				.cAcctSchemaId(parameters.getParameterAsInt("C_AcctSchema_ID", -1))
+				.dateAcct(parameters.getParameterAsTimestamp("DateAcct"))
+				.build();
+	}
+
+	public interface GLJournalImportProcessParameters
```
teosarca: i don't see the purpose of this interface.
why not simply constructing the GLJournalImportProcessContext in the caller class (maybe a separate method),
but this ofParams method that takes GLJournalImportProcessParameters as param is like "let me call you to call me back"

teosarca: or there are multiple implementations of this interface?

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportTableSqlUpdater.java
```diff
@@ -0,0 +1,908 @@
...
+	}
+
+	/**
+	 * Set Period from DateAcct and validate
+	 */
+	private void setPeriod()
+	{
+		// Set Period from DateAcct
+		StringBuilder sql = new StringBuilder("UPDATE I_GLJournal i "
+				+ "SET C_Period_ID=(SELECT MAX(p.C_Period_ID) FROM C_Period p"
```
teosarca, line 409: why not using getC_Period_ID_by_Date db function?

teosarca: (low prio)

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportTableSqlUpdater.java
```diff
@@ -0,0 +1,908 @@
...
+				+ "SET DR_Tax_Acct_ID=(SELECT MAX(ev.C_ElementValue_ID) FROM C_ElementValue ev"
+				+ " INNER JOIN C_Element e ON (e.C_Element_ID=ev.C_Element_ID)"
+				+ " INNER JOIN C_AcctSchema_Element ase ON (e.C_Element_ID=ase.C_Element_ID AND ase.ElementType='AC')"
+				+ " WHERE ev.Value=i.TaxAccountValueFrom AND ev.IsSummary='N'"
+				+ " AND i.C_AcctSchema_ID=ase.C_AcctSchema_ID AND i.AD_Client_ID=ev.AD_Client_ID) "
+				+ "WHERE DR_Tax_Acct_ID IS NULL AND TaxAccountValueFrom IS NOT NULL"
+				+ " AND I_IsImported<>'Y'")
+				.append(selection.toSqlWhereClause("i"));
+		int no = DB.executeUpdateAndThrowExceptionOnFail(sql.toString(), ITrx.TRXNAME_ThreadInherited);
+		log.debug("Set DR_Tax_Acct_ID from Value={}", no);
```
teosarca, line 737: shouldn't we flag as error if TaxAccountValueFrom is not null but DR_Tax_Acct_ID is null?

same for CR side

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportTableSqlUpdater.java
```diff
@@ -0,0 +1,908 @@
...
+	}
+
+	/**
+	 * Calculate Accounted Amounts from Source Amounts and Currency Rate
+	 */
+	private void calculateAccountedAmounts()
+	{
+		// Calculate Accounted DR
+		StringBuilder sql = new StringBuilder("UPDATE I_GLJournal "
+				+ "SET AmtAcctDr = ROUND(AmtSourceDr * CurrencyRate, 2) "    // HARDCODED rounding
```
teosarca, line 886: the hardcoded rounting was it like this before too?

teosarca (changes requested):
check my comments

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportTableSqlUpdater.java
cristinamghita: legacy code;

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportTableSqlUpdater.java
cristinamghita: yes, it was like;
I only moved the sql part and split it in parts to be more easier to read

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcess.java
```diff
@@ -842,6 +346,73 @@ protected ImportRecordResult importRecord(@NonNull final IMutable<Object> state,
 		line.setQty(importRecord.getQty());
 		//
 
+		// Set/Get Tax Account Combination From (Debit)
+		if (importRecord.getDR_Tax_Acct_ID() > 0 && importRecord.getC_ValidCombinationTaxFrom_ID() == 0)
```
teosarca: pls use `< 0` instead of `== 0`

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcess.java
```diff
@@ -842,6 +346,73 @@ protected ImportRecordResult importRecord(@NonNull final IMutable<Object> state,
 		line.setQty(importRecord.getQty());
 		//
 
+		// Set/Get Tax Account Combination From (Debit)
+		if (importRecord.getDR_Tax_Acct_ID() > 0 && importRecord.getC_ValidCombinationTaxFrom_ID() == 0)
+		{
+			final AccountDimension taxAcctDim = newMinimalAccountDimension(importRecord, importRecord.getDR_Tax_Acct_ID());
+			final MAccount taxAcct = MAccount.get(getCtx(), taxAcctDim);
+
+			if (taxAcct.get_ID() == 0)
```
teosarca: i don't see when "taxAcct.get_ID() == 0)" will be true

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcess.java
```diff
@@ -842,6 +346,73 @@ protected ImportRecordResult importRecord(@NonNull final IMutable<Object> state,
...
+			}
+			else
+			{
+				importRecord.setC_ValidCombinationTaxFrom_ID(taxAcct.get_ID());
+				InterfaceWrapperHelper.save(importRecord);
+			}
+		}
+
+		// Set/Get Tax Account Combination To (Credit)
+		if (importRecord.getCR_Tax_Acct_ID() > 0 && importRecord.getC_ValidCombinationTaxTo_ID() == 0)
```
teosarca: same

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcess.java
```diff
@@ -842,6 +346,73 @@ protected ImportRecordResult importRecord(@NonNull final IMutable<Object> state,
...
+		{
+			final AccountDimension taxAcctDim = newMinimalAccountDimension(importRecord, importRecord.getCR_Tax_Acct_ID());
+			final MAccount taxAcct = MAccount.get(getCtx(), taxAcctDim);
+
+			if (taxAcct.get_ID() == 0)
```
teosarca: same

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcess.java
```diff
@@ -900,4 +471,18 @@ protected I_I_GLJournal retrieveImportRecord(Properties ctx, ResultSet rs) throw
...
 		return new X_I_GLJournal(ctx, rs, ITrx.TRXNAME_ThreadInherited);
 	}
-}
+
+	@Value
+	@Builder
+	protected static class GLJournalImportProcessContext
+	{
+		@Nullable
+		ClientId adClientId;
```
teosarca: keep annotations on the same line as the field (will be way easy to read)

teosarca: also why all of these are Nullable? is it right?

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcess.java
```diff
@@ -471,4 +471,18 @@ protected I_I_GLJournal retrieveImportRecord(Properties ctx, ResultSet rs) throw
 	{
 		return new X_I_GLJournal(ctx, rs, ITrx.TRXNAME_ThreadInherited);
 	}
+
+	@Value
+	@Builder
+	protected static class GLJournalImportProcessContext
```
teosarca: since is used in more classes, consider extracting it as a package level class file

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcess.java
```diff
@@ -471,4 +471,18 @@ protected I_I_GLJournal retrieveImportRecord(Properties ctx, ResultSet rs) throw
...
+	@Value
+	@Builder
+	protected static class GLJournalImportProcessContext
+	{
+		@Nullable
+		ClientId adClientId;
+		@Nullable
+		OrgId adOrgId;
+		@Nullable
+		AcctSchemaId cAcctSchemaId;
```
teosarca: pls name it `acctSchemaId` instead of `cAcctSchemaId`. we never name it like that

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcess.java
cristinamghita: But then will not create the account and that set the setC_ValidCombinationTaxFrom_ID
image

Or I am missing something?

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcess.java
cristinamghita: yes, client_id and orgId can be null and they will be taken from context

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcessContext.java
```diff
@@ -0,0 +1,43 @@
...
+import org.adempiere.service.ClientId;
+
+import javax.annotation.Nullable;
+import java.time.Instant;
+
+@Value
+@Builder
+class GLJournalImportProcessContext
+{
+	@Nullable ClientId adClientId;
```
teosarca: why all of those are nullable?

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcess.java
```diff
@@ -56,540 +67,37 @@
...
-		m_C_AcctSchema_ID = getParameters().getParameterAsInt("C_AcctSchema_ID", -1);
-		m_DateAcct = getParameters().getParameterAsTimestamp("DateAcct");
+		final ClientId adClientId = ClientId.ofRepoIdOrNull(getParameters().getParameterAsInt(I_GL_Journal.COLUMNNAME_AD_Client_ID, -1));
+		final OrgId adOrgId = OrgId.ofRepoIdOrNull(getParameters().getParameterAsInt(I_GL_Journal.COLUMNNAME_AD_Org_ID, -1));
+		final AcctSchemaId cAcctSchemaId = AcctSchemaId.ofRepoIdOrNull(getParameters().getParameterAsInt(I_GL_Journal.COLUMNNAME_C_AcctSchema_ID, -1));
+		final Instant dateAcct = getParameters().getParameterAsInstant(I_GL_Journal.COLUMNNAME_DateAcct);
+
+		return GLJournalImportProcessContext.builder()
+				.adClientId(adClientId)
+				.adOrgId(adOrgId)
```
teosarca: is adClientId and adOrgId still needed?

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportProcess.java
```diff
@@ -412,6 +440,14 @@ protected ImportRecordResult importRecord(@NonNull final IMutable<Object> state,
 		return wasInsert ? ImportRecordResult.Inserted : ImportRecordResult.Updated;
 	}
 
+	private CurrencyPrecision getPrecision(@NonNull final I_GL_JournalLine glJournalLine)
+	{
+		final CurrencyId currencyId = CurrencyId.ofRepoIdOrNull(glJournalLine.getC_Currency_ID());
```
teosarca, line 430: fishy but fine for now

backend/de.metas.acct.base/src/main/java/de/metas/acct/impexp/GLJournalImportTableSqlUpdater.java
```diff
@@ -126,15 +127,15 @@ private void setClientFromName()
...
+				+ "SET AD_Client_ID = COALESCE (AD_Client_ID,").append(ClientId.METASFRESH.getRepoId()).append("),"
+				+ " AD_OrgDoc_ID = COALESCE (AD_OrgDoc_ID,").append(OrgId.MAIN.getRepoId()).append("),");
 		if (importProcessContext.getAcctSchemaId() != null)
 		{
 			sql.append(" C_AcctSchema_ID = COALESCE (C_AcctSchema_ID,").append(AcctSchemaId.toRepoId(importProcessContext.getAcctSchemaId())).append("),");
 		}
 		if (importProcessContext.getDateAcct() != null)
 		{
-			sql.append(" DateAcct = COALESCE (DateAcct,").append(DB.TO_DATE(importProcessContext.getDateAcct())).append("),");
+			sql.append(" DateAcct = COALESCE (DateAcct,").append(DB.TO_DATE(Instant.now())).append("),");
```
teosarca: pls use SystemTIme.asInstant

teosarca (changes requested):
check my comments

teosarca (approved):

teosarca (approved):

PR #22062, opened 2026-01-21
mobile UI launchers: use a placeholder or indicators so we make sure they will have the same height

teosarca:


metas-ts:
@mergifyio copy memorable_shiny_release

metas-rc:
@Mergifyio copy intensive_care_release

PR #22078, opened 2026-01-22
Material-Receipt - Fix duplicate M_HU_Trace records when multiple M_HU_Assignments resolve to same VHU

metas-ts:
## Problem

When creating a material receipt, duplicate `M_HU_Trace` records were created for the same VHU.

Root cause: Multiple `M_HU_Assignment` records can point to the same VHU via different paths (e.g., one via `M_HU_ID`, another via `M_TU_HU_ID`). Each
assignment has a different `Updated` timestamp, which is used as `EventTime` in the trace. Since `EventTime` is part of the idempotent duplicate check,
these were not detected as duplicates.

## Solution

Deduplicate VHUs before creating trace events in `HUTraceEventsService`:
- **`createAndAddEvents`** (used by M_InOut, M_Movement): Collect unique VHUs across all assignments using a `LinkedHashMap`. When the same VHU appears
multiple times, keep the most specific assignment (VHU_ID > M_TU_HU_ID > M_LU_HU_ID > M_HU_ID).

## Changes

- `HUTraceEventsService`: Refactored to deduplicate VHUs, added `VhuWithAssignmentInfo` helper class
- `HUTraceRepository`: Minor cleanup
- `CreateReturnedHUsTrxListener`: Minor cleanup


metas-ts:
@mergifyio copy keen_hawk_hotfix

PR #22173, opened 2026-01-28
feat(e2e): add SKIP_TESTS env var to disable tests in custom repos

metas-ts:
## Summary

- Add support for `SKIP_TESTS` environment variable that allows custom repositories (like ms205) to skip specific test files that don't work with their customized DB configurations or window overrides
- Format: `SKIP_TESTS="partial-receipt,other-test"` (comma-separated list of test file names)

## Background

The `partial-receipt.spec.js` test fails on ms205 due to HU Editor modal timing issues caused by custom window overrides (e.g., Purchase Order window 181 → 542052). Rather than modifying the test logic which works fine on standard metasfresh, this provides a clean way for custom repos to opt-out of specific tests.

## Test plan

- [x] Verify test runs normally without SKIP_TESTS set
- [x] Verify test is skipped when `SKIP_TESTS=partial-receipt` is set
- [ ] CI validation

## Related

- Companion PR in ms205 uses this feature to skip the failing test

### DIFF
--- a/e2e/frontend-webui/tests/spec/partial-receipt.spec.js
+++ b/e2e/frontend-webui/tests/spec/partial-receipt.spec.js
@@ -8,6 +8,24 @@ import { PurchaseOrderPage } from '../utils/pages/PurchaseOrderPage';
 import { ReceiptCandidatesPage } from '../utils/pages/ReceiptCandidatesPage';
 
 /**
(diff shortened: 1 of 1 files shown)


metas-ts:
Closing in favor of Playwright-native testIgnore approach in ms205. See https://github.com/metasfresh/ms205/pull/89

PR #22175, opened 2026-01-28
Improve trace report performance

mrie04:


teosarca (approved):
blindly approved because the test data before and after is the same

PR #22268, opened 2026-02-03
More mobile UI Picking playwright tests

teosarca:


teosarca:
implemented by
* https://github.com/metasfresh/metasfresh/pull/22419

PR #22273, opened 2026-02-04
EDI - add new external system config and drop isEDIEnabled (ExportStatus as only source of truth)

mrie04:


backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
```diff
@@ -910,13 +910,20 @@ public BPartnerId getEffectiveBillPartnerId(@NonNull final I_C_Order orderRecord
 				orderRecord.getC_BPartner_ID()));
 	}
 
+	@Override
+	public @NonNull BPartnerId getEffectiveDropshipPartnerId(@NonNull final I_C_Order orderRecord)
+	{
+		return Check.assumeNotNull(getEffectiveDropshipPartnerIdOrNull(orderRecord), "ShipTo Partner should be present");
```
teosarca: provide the order too, 
e.g. ....should be present in {}

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
```diff
@@ -910,13 +910,20 @@ public BPartnerId getEffectiveBillPartnerId(@NonNull final I_C_Order orderRecord
...
+	}
+
 	@Override
 	@Nullable
-	public BPartnerId getEffectiveDropshipPartnerId(@NonNull final I_C_Order orderRecord)
+	public BPartnerId getEffectiveDropshipPartnerIdOrNull(@NonNull final I_C_Order orderRecord)
```
teosarca, line 921: fine by me, but usually consider returning optional from BLs

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EDIBPartnerConfigRepository.java
```diff
@@ -0,0 +1,120 @@
...
+	{
+		private final ImmutableMap<BPartnerId, EDIBPartnerConfig> allEnabledById;
+
+		EDIBPartnerConfigMap(final List<EDIBPartnerConfig> list)
+		{
+			this.allEnabledById = Maps.uniqueIndex(list, EDIBPartnerConfig::getBPartnerId);
+		}
+
+		@Nullable
+		EDIBPartnerConfig getById(@NonNull final BPartnerId bPartnerId) {return allEnabledById.get(bPartnerId);}
```
teosarca, line 118: consider returning optional

teosarca: or renaming it to ...OrNull

backend/de.metas.edi/src/main/sql/postgresql/system/80-de.metas.edi/5786970_c_invoice_candidate_failed_to_update_v_remove_obsolete_edi_columns.sql
```diff
@@ -0,0 +1,873 @@
+/*
```
teosarca, line 1: don't like that too many things are happening in one file

i expected all DDL functions/views etc to be one per file

app dict changes in one file
actually dropping the current column + backup before in one file

backend/de.metas.edi/src/main/java/de/metas/edi/model/validator/C_BPartner.java
```diff
@@ -64,26 +66,4 @@ public void validate(final I_C_BPartner bpartner)
 
 		throw new AdempiereException("Invalid EDI partner " + causes.toString().trim());
 	}
-
-	/**
-	 * Make sure the IsEDIRecipient flag from the invoice candidates of a partner is always up to date
-	 */
-	@ModelChange(timings = { ModelValidator.TYPE_BEFORE_CHANGE }, ifColumnsChanged = { I_C_BPartner.COLUMNNAME_IsEdiInvoicRecipient })
-	public void updateIsEDIInvoicRecipient_InvoiceCandidates(final I_C_BPartner partner)
```
teosarca, line 72: why removing?

backend/de.metas.edi/src/main/java/de/metas/edi/model/validator/C_Invoice_Candidate.java
```diff
@@ -1,69 +0,0 @@
...
-import de.metas.bpartner.BPartnerId;
-import de.metas.bpartner.service.IBPartnerDAO;
-import de.metas.edi.api.IEDIInvoiceCandBL;
-import de.metas.edi.model.I_C_BPartner;
-import de.metas.edi.model.I_C_Invoice_Candidate;
-import de.metas.util.Services;
-
-@Interceptor(I_C_Invoice_Candidate.class)
-@Component
-public class C_Invoice_Candidate
```
teosarca, line 40: why dropping?

mrie04: because both columns are removed

backend/de.metas.edi/src/main/java/de/metas/edi/model/validator/C_BPartner.java
mrie04: column dropped, because it wasn't used. Confirmed with Tobi

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
mrie04: pulled from uat, would create merge conflicts, i would like to avoid here

PR #22370, opened 2026-02-17
Fix duplicate shipment lines from race condition (me03#28143)

metas-ts:
## Summary

- Adds a guard in `GenerateInOutFromShipmentSchedules` that detects when a schedule already has a draft-shipment allocation from a prior workpackage, preventing duplicate `M_InOutLine` creation
- When a user triggers "Generate Shipments" multiple times before the async recompute updates `QtyToDeliver`, the same schedule was processed by multiple workpackages — this guard blocks the duplicates
- Extends `ShipmentScheduleAllocQuery` and `ShipmentScheduleAllocDAO` with a `processed` filter and a new `getScheduleIdsWithDraftShipmentAllocations()` batch method
- New Cucumber feature `duplicateShipmentLineGuard.feature` with 3 scenarios: normal flow (no false positive), duplicate blocked, partial delivery allowed

## Root Cause

`QtyToDeliver` is only updated asynchronously by `UpdateInvalidShipmentSchedulesWorkpackageProcessor`. Between user triggers, the recompute may not have run yet, so subsequent workpackages still see stale `QtyToDeliver > 0` and create duplicate shipment lines.

The guard checks `M_ShipmentSchedule_QtyPicked` for records with `Processed=N` + `M_InOutLine_ID IS NOT NULL` (draft allocations visible immediately after the first WP commits), making it immune to the recompute timing issue.

## Incidents Fixed

| Date | Shipment | Duplicates | Triggers |
|------|----------|------------|----------|
| Jan 31 | LS26-300279 | 2x (3 groups) | 7 triggers over 49 min |
| Feb 14 | LS26-300399 | 3x (2 groups) | 3 triggers over 13 min |

## Test plan

- [x] Unit tests pass: `de.metas.swat.base` (361 tests), `de.metas.handlingunits.base` (637 tests)
- [x] CI pipeline passes (Java build + existing Cucumber tests)
- [x] New Cucumber feature `duplicateShipmentLineGuard.feature` passes (3 scenarios)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,6 +70,9 @@ misc/services/camel/de-metas-camel-shipping/tmp/
 .github/merge_to_master
 .github/integrate
 
+# workspace-local Maven repository
+.m2-local/
+
 # claude related
 .claude
 **/CLAUDE.md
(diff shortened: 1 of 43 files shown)


metas-ts (commented):
lgtm

PR #22373, opened 2026-02-17
Unset IC.DateInvoiced / DateAcct when all associated C_Invoice_Line_Allocs are reversed.

adi-stefan:


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandBL.java
```diff
@@ -1907,6 +1907,12 @@ public void updateProcessedFlag(@NonNull final I_C_Invoice_Candidate ic)
 
 			ic.setApprovalForInvoicing(false);
 		}
+		else if (nonReversedIlas == 0)
```
teosarca: what happens if ic.isError() but in DB we actually have non-reversed ILAs?

suggestion:
* change the type of nonReversedIlas  to "Integer" and have it null.
* before starting to increment it, set it to zero
* here instead of checking nonReversedIlas == 0 check if is not null too

wdyt?

adi-stefan: done

adi-stefan:
@Mergifyio copy keen_hawk_hotfix

metas-rc:
@Mergifyio copy oasis_sun_hotfix

PR #22423, opened 2026-02-23
Allow opening related records (via a relation type defined in an AD_Process) as a modal window.

adi-stefan:
# Pull Request Summary

## Overview
This PR introduces a new **"RelationTypeInOverlay"** process type that enables processes to open related documents in modal overlay windows based on a configured relation type. This provides a standardized way to navigate between related documents in the metasfresh UI.

### Database Schema
- Added `AD_RelationType_ID` column to `AD_Process` table with foreign key to `AD_RelationType`
- Field is conditionally displayed when Process Type = `RelationTypeInOverlay`
- Migration script: `5789740_sys_add_ADProcess_ADRelationType_ID.sql`

### Backend Domain Model
- **New class**: `RelationTypeId` - Value object for type-safe handling of relation type IDs
- **Enhanced**: `ProcessInfo` - Now carries `adRelationTypeId` to support the new process type
- **Enhanced**: `ProcessType` enum - Added `RelationTypeInOverlay` enum value
- **Enhanced**: `RelationTypeRelatedDocumentsProvidersFactory` - Added method to find document providers by `RelationTypeId`
- **Enhanced**: `POZoomSource` - Added null-safety improvements with `@Nullable` annotations

### UI Layer
- **New class**: `RelationTypeInOverlayProcess` - Process implementation that:
  - Retrieves related documents based on configured relation type
  - Creates a grid view with the related documents
  - Opens the view in a modal overlay
- **Enhanced**: Process interceptor - Auto-assigns `RelationTypeInOverlayProcess` class when Type is set to `RelationTypeInOverlay`
- **Test coverage**: `RelationTypeInOverlayProcessTest` - Comprehensive unit tests covering all scenarios

### Code Quality
- Translated German comments to English in generated model classes
- Added `@Nullable` annotations throughout for null-safety
- Proper separation of concerns between backend and UI layers
- Follows existing patterns and conventions in the codebase

## How It Works

1. Administrator configures a process with Type = `RelationTypeInOverlay` and selects an `AD_RelationType`
2. User executes the process on a single record
3. System retrieves related documents using the configured relation type provider
4. Related documents are displayed in a modal overlay grid view
5. User can interact with the related documents without navigating away from the current screen

## Testing
- Unit tests cover all core scenarios:
  - Single selection validation
  - Relation type provider lookup
  - Related documents retrieval
  - Empty results handling
  - Overlay view creation

## Backward Compatibility
✅ Fully backward compatible - new optional field doesn't affect existing processes


backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
```diff
@@ -0,0 +1,185 @@
...
+
+		return POZoomSource.of(sourcePO, getProcessInfo().getAdWindowId());
+	}
+
+	private IView createView(@NonNull final TableRecordReference recordReference, @NonNull final WindowId targetWindowId)
+	{
+
+		final RelatedDocumentsId relatedDocumentsId = RelatedDocumentsId.ofString("AD_RelationType_ID-" + getRelationTypeId().getRepoId());
+
+		final WindowId srcWindowId = WindowId.of(getProcessInfo().getAdWindowId());
```
teosarca: **HIGH — NPE risk**: `getProcessInfo().getAdWindowId()` can return `null` (the `ProcessInfo` field is `@Nullable`). `WindowId.of()` requires `@NonNull` — this will throw an NPE.

The `checkPreconditionsApplicable` guards against null in the UI context, but `doIt()` can be invoked programmatically (e.g., from a workpackage) bypassing preconditions.

Suggested fix:
```java
final AdWindowId adWindowId = Check.assumeNotNull(
    getProcessInfo().getAdWindowId(),
    "AD_Window_ID is required for process {}", getProcessInfo().getAdProcessId());
final WindowId srcWindowId = WindowId.of(adWindowId);
```

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
```diff
@@ -0,0 +1,185 @@
...
+				.map(docProvider -> docProvider.retrieveRelatedDocumentsCandidates(zoomSource, null))
+				.orElse(Collections.emptyList());
+
+		if (relatedDocumentGroups.isEmpty())
+		{
+			throw new AdempiereException(MSG_NO_RELATED_DOCS_FOUND);
+		}
+
+		// Get the first group and create a view from it. We're expecting exactly one group.
+		final RelatedDocumentsCandidateGroup firstGroup = relatedDocumentGroups.get(0);
```
teosarca: **MEDIUM — Silent discard of extra groups**: The comment says "expecting exactly one group" but doesn't enforce it. If a misconfigured relation type produces multiple groups, the extras are silently ignored. Consider logging a warning:
```java
if (relatedDocumentGroups.size() > 1)
{
    log.warn("RelationType {} returned {} groups; using only the first. RelationType may be misconfigured.",
        relationTypeId, relatedDocumentGroups.size());
}
```

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * Process implementation for opening related documents in an overlay window (as grid view).
+ * <p>
+ * This process is automatically assigned when a process has Type = 'RelationTypeInOverlay'.
+ * It retrieves related documents based on the configured AD_RelationType_ID and displays them in a modal overlay.
+ */
+public class RelationTypeInOverlayProcess extends JavaProcess implements IProcessPrecondition
+{
+	private final static AdMessageKey MSG_NO_RELATED_DOCS_FOUND = AdMessageKey.of("NO_RELATED_DOCS_FOUND");
```
teosarca: **HIGH — Duplicate constant**: This same `AdMessageKey.of("NO_RELATED_DOCS_FOUND")` is also defined in `SqlViewFactory.java` (line 79). Should be defined in one place and shared to avoid divergence.

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/model/interceptor/AD_Process.java
```diff
@@ -58,6 +60,9 @@ public void setClassnameForProcessType(@NonNull final I_AD_Process process)
 					throw new FillMandatoryException(I_AD_Process.COLUMNNAME_JSONPath);
 				}
 				break;
+			case RelationTypeInOverlay:
+				process.setClassname(RELATION_TYPE_IN_OVERLAY_PROCESS_CLASSNAME);
+				break;
```
teosarca, line 65: **HIGH — Missing AD_RelationType_ID validation**: The `POSTGREST` case (line 58-61) validates that `JSONPath` is filled. The same pattern should be applied here — validate that `AD_RelationType_ID` is configured, otherwise the process will only fail at runtime.

The column has `MandatoryLogic` in AD, which helps in the UI, but the interceptor should enforce it server-side too:
```java
case RelationTypeInOverlay:
    process.setClassname(RELATION_TYPE_IN_OVERLAY_PROCESS_CLASSNAME);
    if (process.getAD_RelationType_ID() <= 0)
    {
        throw new FillMandatoryException(I_AD_Process.COLUMNNAME_AD_RelationType_ID);
    }
    break;
```

backend/de.metas.ui.web.base/src/test/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcessTest.java
```diff
@@ -0,0 +1,228 @@
...
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+class RelationTypeInOverlayProcessTest
+{
+	@BeforeEach
+	void beforeEach()
+	{
+		MockitoAnnotations.initMocks(this);
```
teosarca, line 49: **MEDIUM — Deprecated API**: `MockitoAnnotations.initMocks(this)` was deprecated in Mockito 3.4.0. Use `MockitoAnnotations.openMocks(this)` instead.

Also: the `@Mock` annotation on the inner class fields (e.g., line 125-128) combined with `= mock(...)` initializer is redundant. `initMocks`/`openMocks` already creates the mock via `@Mock`. Pick one pattern.

teosarca (commented):
Code review for me03#28057. See inline comments below.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
adi-stefan: This is already guarded against in checkPreconditionsApplicable()

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
adi-stefan: done

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/model/interceptor/AD_Process.java
adi-stefan: This would cause a problem. If the user is changing the type to `RelationTypeInOverlay`, the AD_RelationType_ID is just about to appear on the UI. However, because we're throwing an exception from here it means the UI is not going to be refreshed properly, so the AD_RelationType_ID field is never shown.

Tested in ~~SwingUI~~ webUI.

backend/de.metas.ui.web.base/src/test/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcessTest.java
adi-stefan: we're using mockito-core:2.7.22. `MockitoAnnotations.openMocks(this)` does not exist in this version.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
adi-stefan: added.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
```diff
@@ -0,0 +1,195 @@
...
+		if (relatedDocumentGroups.isEmpty())
+		{
+			throw new AdempiereException(MSG_NO_RELATED_DOCS_FOUND);
+		}
+
+		// Get the first group and create a view from it. We're expecting exactly one group.
+		final RelatedDocumentsCandidateGroup firstGroup = relatedDocumentGroups.get(0);
+		if (relatedDocumentGroups.size() > 1)
+		{
+			Loggables.withLogger(log, Level.WARN).addLog("RelationType {} returned {} groups; using only the first. RelationType may be misconfigured.",
```
teosarca: * u can simply call addLog (JavaProcess has that method)
* minor: i would keep this block together with "if (relatedDocumentGroups.isEmpty())"

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
teosarca: Agreed that checkPreconditionsApplicable guards against it in the UI path. Thanks for adding the `Check.assumeNotNull` anyway — it protects against programmatic invocations.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
teosarca: Looks good, thanks.

backend/de.metas.ui.web.base/src/test/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcessTest.java
teosarca: Right, `openMocks` requires Mockito 3.4+. With 2.7.22 `initMocks` is the correct API. My mistake.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
teosarca: Thanks. Pushed a small follow-up (7820e08) — switched from `Loggables.withLogger` to the simpler `addLog()` and moved the size check right next to the `isEmpty()` check.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
teosarca: Done in 7820e08.

teosarca (approved):

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/model/interceptor/AD_Process.java
teosarca: why not setting MandatoryLogic: show only type is RelationTypeInOverlay (basically shall be the same as DisplayLogic...) ?

adi-stefan: Already doing that: https://github.com/metasfresh/metasfresh/pull/22423/changes#diff-0d2843385d3c7bf51dab1ad684070f0384b258658d0b86d365ebcdb1b6bee7f7R100

adi-stefan: Here's the problem demo:
chrome_6YBJq71B7f
Notice how, for Postgrest the UI does not update to show the `JSONPath` field because of the mandatory field exception.

It works in SwingUI, so I guess that's why no one noticed it before.

teosarca: Thanks for the demo — this is very helpful. So the underlying issue is: the WebUI is not enforcing `MandatoryLogic` before save. When `MandatoryLogic` evaluates to true and the field is empty, the WebUI should block the save attempt entirely (same as regular mandatory fields), so the interceptor never fires. Can you confirm that's the correct understanding? If so, this is a separate WebUI bug affecting any field with `MandatoryLogic`.

adi-stefan: No, it's because it's also a callout method, so it gets invoked even if the save is not done.

teosarca: Got it — looking at this more closely, the issue is that the method is also a **callout** (`@Callout`/`@CalloutMethod`), not just an interceptor. So when the user changes `Type` to `Postgrest`, the callout fires on the field change (before save), validates `JSONPath` is filled, and throws — but the user hasn't had a chance to fill it yet since it just appeared via `DisplayLogic`. The validation belongs in the interceptor context (on save), not the callout context (on field change). `MandatoryLogic` handles the UI enforcement. This is a pre-existing issue with the `Postgrest` case and not something to fix in this PR. Thanks for the demo\!

PR #22429, opened 2026-02-24
User update performance fix - Prevent full postgresFTS reindex with partner not set

mrie04:


mrie04:
@Mergifyio copy inner_silence_hotfix task_force_hotfix

Issue #1630, opened 2017-05-24
Producer Invoices w/ price override after invoice void wrong


metas-mk:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Currently, when setting a price override in invoice candidate, invoicing, voiding the invoice and invoicing again, the price actual in the newly created invoice is wrong.
#### Which are the steps to reproduce?
Create a Purchase Order with a product Line. Do the Material Receipt. Go to Invoice Candidates and override the price. Create the invoice. Void the invoice. Create a new invoice. The Priceactual in the new invoice is wrong (initial Purchase order price). Price entered remains correct, as in Price Override.
### What is the expected or desired behavior?
Price actual shall also have the price override.

metas-rc:
I wasn't able to reproduce the bug any more. Moving task to done.

PR #22437, opened 2026-02-25
Fix C_Flatrate_Term.MasterEndDate not always set (#1630)

metas-ts:
## Summary

- **Interceptor fix**: Removed `!isAutoRenew()` guard from `C_Flatrate_Term.setMasterEndDate()` so all new terms (including auto-renew) get `MasterEndDate = EndDate` on creation
- **FlatrateBL fix**: Removed `ExtendAll`-only restriction from `updateMasterEndDateIfNeeded()` so MasterEndDate is propagated through the chain for all extension types

### Root cause

Two issues caused `MasterEndDate` to remain NULL for auto-renew contracts:

1. The interceptor only set `MasterEndDate = EndDate` for new **non-auto-renew** terms. Auto-renew terms got `null`, and when the next term in the chain was linked, it read `null` from the next term too (chicken-and-egg).

2. `updateMasterEndDateIfNeeded()` (which correctly walks the whole chain and sets the last EndDate) only executed for `EXTENSIONTYPE_ExtendAll`. Regular single-extension auto-renew contracts never hit this code path.

### Data impact

On the customer DB: 16,532 out of 116,008 completed terms have `MasterEndDate = NULL`. All are auto-renew with a next term set. A companion migration script in yo98 fixes the existing data.

## Test plan

- [ ] Create a new auto-renew contract and verify `MasterEndDate` is set to `EndDate`
- [ ] Extend the contract and verify all terms in the chain have `MasterEndDate` = last term's `EndDate`
- [ ] Create a non-auto-renew contract and verify `MasterEndDate` still works correctly
- [ ] Test ExtendAll contracts still work as before

### DIFF
--- a/backend/de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
+++ b/backend/de.metas.contracts/src/main/java/de/metas/contracts/impl/FlatrateBL.java
@@ -1143,6 +1143,17 @@ public class FlatrateBL implements IFlatrateBL
 		}
 		while (X_C_Flatrate_Transition.EXTENSIONTYPE_ExtendAll.equals(nextTransition.getExtensionType()) && nextTransition.getC_Flatrate_Conditions_Next_ID() > 0);
 
+		// Also collect predecessor contracts (those before the initial contract in the chain)
+		// so their MasterEndDate is updated as well.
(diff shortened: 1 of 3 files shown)


adi-stefan:
@claude review

adi-stefan:
@claude review

adi-stefan:
@claude review

PR #22442, opened 2026-02-25
Add bug fix for order-independent parent-child HU reservation logic. Created relevant unit tests.

adi-stefan:


teosarca (changes requested):
## Code Review — HU Reservation Parent-Child Filtering

### CRITICAL Issues (2)

**CRITICAL 1: Transaction Atomicity Broken**

`deleteChildReservations` calls the public `deleteReservationsByVHUIds`, which internally uses `trxManager.runInNewTrx(...)` — an independent auto-committed transaction. But it's called from within `makeReservationInTrx` (running in `callInThreadInheritedTrx`). Child reservation deletions are **committed immediately** in a separate transaction. If the subsequent parent reservation creation fails, child reservations are already gone — data corruption.

**Fix**: Call the private `deleteReservationsByVHUIdsInTrx` directly instead of the public method.

**CRITICAL 2: Premature Deletion — Deletes Before Filtering**

In `getHUsToReserve`, the execution order is:
1. Load all HUs
2. Build ID set
3. **Delete child reservations** (for ALL input HUs)
4. Filter out HUs with ancestors in list
5. Filter out already-reserved HUs
6. Filter out HUs with reserved ancestors

Step 3 runs before filtering. If an input HU is itself already reserved (filtered out at step 5), its children's reservations are already deleted. Legitimate reservations belonging to other orders get destroyed for HUs that won't even be reserved.

**Fix**: Reorder — filter first (steps 4-6), then delete child reservations only for the surviving HUs.

---

### HIGH Issues (3)

**HIGH 1: Dead code** — `isHuUnreservable()` is no longer called by anything after `retrieveUnreservableQty` was replaced. Should be removed.

**HIGH 2: `maxDepth = 3` is misleading** — metasfresh HU hierarchy is LU→TU→VHU = 3 levels but only 2 parent hops. The safety counter of 3 allows one extra hop. More importantly, the tests don't exercise this boundary at all (all removed with "requires 3-level hierarchy"). Consider removing the counter entirely (parent chain is always finite) or documenting the rationale.

**HIGH 3: Undocumented semantic change** in `retrieveActualReservedQty` — the old method counted qty from VHUs with `IsReserved=Y` flag; the new one queries `M_HU_Reservation` records. These can disagree (flag set without record, or vice versa). The behavioral difference affects when the "Delete HU Reservation" action is shown in the UI. Please document the rationale for this change.

---

### MEDIUM Issues (4)

**MEDIUM 1: Test 31 doesn't test what its title claims** — "Parent Reservation Deletes Child Reservations" manually re-implements descendant collection inline instead of calling `deleteChildReservations`. It doesn't verify reservation records are actually deleted.

**MEDIUM 2: Missing import — compilation failure** — `ImmutableList.of(lu)` is used in Test 31 but `ImmutableList` is not imported. Only `ImmutableSet` is imported.

**MEDIUM 3: Unnecessary DB traversal** — `getDescendantHuIds` traverses all descendants for ALL input HUs including ones that will be filtered out later. (Related to CRITICAL 2 — fixing the ordering also fixes this.)

**MEDIUM 4: Test coverage gap** — No end-to-end test for `getHUsToReserve` itself; no test for the scenario where a TU is selected but its parent LU is already reserved (not in the input list).

---

### Positive Observations

- Core algorithm is correct: building the full `allHuIds` set before filtering achieves true order-independence
- `Set<HuId>` for O(1) ancestor lookup is the right data structure choice
- `@VisibleForTesting` annotations follow project conventions
- Rename from `retrieveUnreservableQty` → `retrieveActualReservedQty` is consistently reflected in the caller

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/reservation/HUReservationService.java
```diff
@@ -205,10 +205,127 @@ private Optional<HUReservation> makeReservationInTrx(@NonNull final ReserveHUsRe
...
+		// Step 3: Filter out HUs whose ancestors are in the Set
+		// Step 4: Filter out already reserved HUs
+		// Step 5: Filter out HUs whose ancestors are already reserved (even if ancestor not in input list)
+		final ImmutableList<I_M_HU> toReserveHUs = allHUs.stream()
+				.filter(hu -> !hasAncestorInList(hu, allHuIds))
 				.filter(hu -> !hu.isReserved())
+				.filter(hu -> !hasReservedAncestor(hu))
 				.collect(ImmutableList.toImmutableList());
+
+		// Step 6: Before filtering, delete any existing child reservations for HUs we're about to reserve
```
teosarca: **Nit: Stale comment** — this now says "Before filtering" but the deletion actually happens AFTER filtering (steps 3-5 above). Should be:
```java
// Step 6: After filtering, delete any existing child reservations for HUs we're about to reserve
```

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/reservation/HUReservationService.java
```diff
@@ -205,10 +205,127 @@ private Optional<HUReservation> makeReservationInTrx(@NonNull final ReserveHUsRe
...
+	 */
+	private void deleteChildReservations(final @NonNull List<I_M_HU> hus)
+	{
+		final Set<HuId> childHuIds = hus.stream()
+				.flatMap(hu -> getDescendantHuIds(hu).stream())
+				.collect(ImmutableSet.toImmutableSet());
+
+		if (!childHuIds.isEmpty())
+		{
+			deleteReservationsByVHUIdsInTrx(childHuIds);
```
teosarca, line 248: **Logic concern: cross-order reservation deletion**

`deleteReservationsByVHUIdsInTrx` deletes ALL `M_HU_Reservation` records for the given VHU IDs — it does not filter by order line / document ref.

Consider this scenario:
1. Order **B** reserves child TU (TU has `IsReserved=Y` + reservation record for order B)
2. User selects parent LU for order **A**
3. LU passes all filters: it is not reserved, has no reserved ancestor. The filters only look **up** the hierarchy, never down, so the reserved child is invisible to them.
4. `deleteChildReservations` → finds child TU → deletes order **B**'s reservation
5. LU gets reserved for order **A** — order B silently lost its reservation

If this is intentional business behavior ("parent reservation supersedes child reservations regardless of owner"), please add a comment here documenting it. If not, the deletion should be scoped to only the current document ref, or the filters should also check for reserved descendants.

teosarca (commented):
Thanks for the fixes — both CRITICALs and most of the other issues are properly resolved. Two remaining items below (one nit, one logic concern).

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/reservation/HUReservationService.java
adi-stefan: added logic that prevents such a case from happening. If a child it reserved, it shall not be possible to reserve the parent.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/reservation/HUReservationService.java
adi-stefan: changed.

teosarca (approved):
All review findings addressed. The new `hasReservedDescendant` filter is the right approach — it prevents the cross-order issue at the source rather than trying to clean up after the fact.

One minor observation: with step 6 blocking any HU that has reserved descendants, `deleteChildReservations` at step 7 is now effectively a no-op in normal flow (no HU reaching step 7 can have reserved children). It still serves as a safety net for potential flag/record inconsistencies, so fine to keep.

LGTM.

PR #22447, opened 2026-02-25
Fix - `VAT-ID` may not be displayed as part of the address (#21969)

riadmen:
https://github.com/metasfresh/metasfresh/pull/21969


cristinamghita (approved):
lgtm

PR #22452, opened 2026-02-25
Quick input: Enter key resolves product and advances focus, beep on not found

teosarca:
## Summary

- In quick input (batch entry), pressing Enter in a lookup field now immediately resolves the typed value and auto-advances focus to the next field (e.g., quantity)
- If the product number is not found, a beep sound is played via the Web Audio API as audio feedback
- Enables a fast numpad-only workflow: product number → Enter → quantity → Enter → line added → repeat
- Only affects quick input context (`subentity === 'quickInput'`); regular lookup behavior is unchanged

## Test plan

- [ ] Open a Sales Order and activate batch entry (quick input)
- [ ] Type a valid product number and press Enter — product should resolve and focus should advance to the quantity field
- [ ] Type the quantity and press Enter — the order line should be added and focus should return to the product field
- [ ] Type an invalid product number and press Enter — a beep sound should play
- [ ] Verify regular lookup behavior outside quick input is unchanged (e.g., BPartner lookup in a document)
- [ ] Verify the typeahead dropdown still works normally when typing slowly and selecting with mouse

### DIFF
--- a/e2e/frontend-webui/tests/spec/quick-input-enter.spec.js
+++ b/e2e/frontend-webui/tests/spec/quick-input-enter.spec.js
@@ -1,0 +1,229 @@
+import { expect } from '@playwright/test';
+import { test } from '../../playwright.config';
+import { allure } from 'allure-playwright';
+import { Backend } from '../utils/Backend';
+import { LoginPage } from '../utils/pages/LoginPage';
+import { DashboardPage } from '../utils/pages/DashboardPage';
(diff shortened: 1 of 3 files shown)


teosarca:
Consolidated into #22460 — all unique changes (playBeep, resolveAndSelectOnEnter, buildTypeaheadRequestForQuery, handleAutoSelectAndAdvance, focusNextFieldInForm, Enter key interception) have been merged. E2E test coverage expanded to 5 tests × 2 languages (10 total) including invalid product behavior.

PR #22464, opened 2026-02-26
Fix QR scanning on Zebra MC3300x scanners (DataWedge IME)

teosarca:
## Summary

- Replace `readOnly` with `inputMode="none"` on barcode scanner input field to keep Chrome's InputConnection alive for Zebra DataWedge IME text injection
- Remove readonly guards from `onChange`/`onKeyUp` handlers so IME-injected text is processed
- Always focus the input field (required for InputConnection, safe with `inputMode="none"`)
- Relax `altKey` filter in `useKeyboardBarcodeReader` for Zebra MC3300x firmware that sets `altKey=true` on all scanner keystrokes
- Add `typeViaIME()` E2E test utility and barcode scanner input mode regression tests

## Test plan

- [ ] Run new E2E test: `npx playwright test tests/spec/picking/barcodeScannerModes.spec.js`
- [ ] Run picking regression: `npx playwright test tests/spec/picking/` (23+ existing tests)
- [ ] Verify on Zebra MC3300x device: QR scan via DataWedge IME populates the field and triggers scan processing
- [ ] Verify virtual keyboard does NOT appear on mobile when input field has focus

### DIFF
--- a/misc/services/mobile-webui/mobile-webui-frontend/src/components/BarcodeScannerComponent.jsx
+++ b/misc/services/mobile-webui/mobile-webui-frontend/src/components/BarcodeScannerComponent.jsx
@@ -121,9 +121,7 @@ const BarcodeScannerComponent = ({
       if (isShowVideo) {
         videoRef?.current?.scrollIntoView({ behaviour: 'smooth', block: 'center', inline: 'end' });
       }
-      if (!isInputTextReadonly) {
-        inputTextRef?.current?.focus();
-      }
+      inputTextRef?.current?.focus();
     } /* no deps, call it on each render */
   );
 
@@ -222,8 +220,6 @@ const BarcodeScannerComponent = ({
   );
 
   const handleInputTextChanged = (e) => {
-    if (isInputTextReadonly) return;
-
     const scannedBarcode = e.target.value;
 
     if (
@@ -240,8 +236,6 @@ const BarcodeScannerComponent = ({
   }, [textChangedDebounceMillis]);
 
   const handleInputTextKeyPress = (e) => {
-    if (isInputTextReadonly) return;
-
     if (e.key === 'Enter') {
       const scannedBarcode = e.target.value;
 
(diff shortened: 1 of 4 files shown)


misc/services/mobile-webui/mobile-webui-frontend/src/components/BarcodeScannerComponent.jsx
```diff
@@ -271,7 +265,7 @@ const BarcodeScannerComponent = ({
           className="input-text"
           type={isShowInputText ? 'text' : 'hidden'}
           placeholder={inputPlaceholderText || trl('components.BarcodeScannerComponent.scanTextPlaceholder')}
-          readOnly={isInputTextReadonly}
+          inputMode={isInputTextReadonly ? 'none' : undefined}
```
teosarca, line 268: what's happening when isInputTextReadonly?
i would expect user to be allowed to really type in that field. is it right?

teosarca: When `isInputTextReadonly=true` (the default on mobile/tablet):

- **Before (old code)**: `readOnly` attribute → field is truly read-only, browser blocks all input. Virtual keyboard suppressed. But this also prevents Chrome from establishing an `InputConnection`, so DataWedge IME text injection silently fails.
- **After (this PR)**: `inputMode="none"` → virtual keyboard is suppressed (user can't manually type on mobile — no keyboard appears), but the field stays editable. This keeps `InputConnection` alive, so DataWedge IME injection works.

On mobile, the user still can't manually type because there's no virtual keyboard. Scanner input reaches the app via two paths:
1. **Keyboard events** (scanner in keyboard/wedge mode) → `useKeyboardBarcodeReader` hook → `onReadDone`
2. **IME text injection** (scanner in IME/InputConnection mode, e.g. DataWedge default) → `onChange` → `handleInputTextChanged`

When `isInputTextReadonly=false`: `inputMode` is `undefined` → normal behavior, virtual keyboard shows, user can type freely.

PR #22465, opened 2026-02-26
Extend packing material handling logic to include project associations, ensuring comprehensive project-level tracking and alignment in handling unit processing.

adi-stefan:


teosarca (changes requested):
## Code Review — PR #22465

### BUG (HIGH): Missing `.distinct()` in `getUniqueProjectIdOrNull()`

**File**: `HUPackingMaterialDocumentLineCandidate.java`

```java
public ProjectId getUniqueProjectIdOrNull()
{
    return CollectionUtils.singleElementOrNull(sources.stream()
            .map(IHUPackingMaterialCollectorSource::getProjectId)
            .filter(Objects::nonNull)
            .collect(Collectors.toList()));
}
```

`CollectionUtils.singleElementOrNull()` returns `null` when `collection.size() != 1`. If there are multiple sources sharing the same projectId (the **normal case** — e.g., 3 InOutLines from the same project), the collected list has 3 identical entries → size > 1 → returns **null** → C_Project_ID is **not** propagated to the packing material InOut line.

**Fix**: Add `.distinct()` before `.collect()`:

```java
return CollectionUtils.singleElementOrNull(sources.stream()
        .map(IHUPackingMaterialCollectorSource::getProjectId)
        .filter(Objects::nonNull)
        .distinct()
        .collect(Collectors.toList()));
```

This is called from `HUInOutBL.updatePackingMaterialInOutLine()` which sets the project on the packing material shipment line — without this fix, the project won't propagate to packing material lines in most real-world scenarios.

---

### MEDIUM: Second cucumber scenario missing cleanup step

The first scenario ends with:
```gherkin
    And set project type Sales/Purchase Order to inactive
```

The second scenario ("C_Project_ID propagates to packing material invoice candidates") does **not** have this cleanup step. Since both scenarios share the same Background, the cleanup from scenario 1 runs after it, but scenario 2 leaves state behind that could affect subsequent tests.

---

### MEDIUM: `InOutLineHUPackingMaterialCollectorSource` — unnecessary `projectId` in `@EqualsAndHashCode`

```java
@EqualsAndHashCode(of= {"productId", "recordId", "projectId"})
```

`recordId` is the `M_InOutLine_ID` which already uniquely identifies the record. Adding `projectId` changes equality semantics unnecessarily. While functionally harmless (same record → same project), it's a semantic change that implies two records with the same `M_InOutLine_ID` could have different project IDs.

---

### LOW: Style inconsistency in `M_ShipmentSchedule_StepDef`

The new C_Project_ID validation uses old-style `DataTableUtil`:

```java
final String projectIdentifier = DataTableUtil.extractStringOrNullForColumnName(tableRow, "OPT." + I_M_ShipmentSchedule.COLUMNNAME_C_Project_ID + "." + TABLECOLUMN_IDENTIFIER);
```

All other step defs in this PR use the modern `DataTableRow` API (`row.getAsOptionalIdentifier(...)`). This is because the surrounding method still uses the old `Map<String, String>` parameter, but it's worth noting for consistency.

---

### LOW: `PlainPackingMaterialDocumentLineSource.projectId` missing `@Nullable`

The field is declared without `@Nullable`:
```java
ProjectId projectId;
```

But the constructors accept `@Nullable ProjectId projectId`. The Lombok `@Value` class should annotate the field as `@Nullable` for clarity.

---

### NIT: Formatting

- `OrderLinePackingMaterialDocumentLineSource.getProjectId()` — one-liner brace style inconsistent with the rest of the codebase (should use multi-line format)
- Extra trailing blank lines in `OrderLinePackingMaterialDocumentLine.java` and `InOutLineHUPackingMaterialCollectorSource.java`
- Copyright year 2025 in new `C_Project_StepDef.java` (should be 2026)

---

### Overall

The approach is solid — threading `ProjectId` through the packing material document line interfaces and the `AbstractPackingMaterialDocumentLinesBuilder` key is the right design. The packing material key now correctly groups by `(ProductId, ProjectId)` which means different projects get separate packing material lines. The cucumber test coverage is comprehensive, covering both regular and packing material flows.

The `.distinct()` bug is the blocking issue — without it, the feature won't work correctly when multiple source lines share the same project (which is the primary use case).

teosarca (changes requested):
## Code Review - PR #22465

### BUG (HIGH): Missing .distinct() in getUniqueProjectIdOrNull()

**File**: HUPackingMaterialDocumentLineCandidate.java

CollectionUtils.singleElementOrNull() returns null when collection.size() != 1. If there are multiple sources sharing the same projectId (the normal case, e.g. 3 InOutLines from the same project), the collected list has 3 identical entries, size > 1, returns null, so C_Project_ID is NOT propagated to the packing material InOut line.

**Fix**: Add .distinct() before .collect():
```java
return CollectionUtils.singleElementOrNull(sources.stream()
        .map(IHUPackingMaterialCollectorSource::getProjectId)
        .filter(Objects::nonNull)
        .distinct()
        .collect(Collectors.toList()));
```

This is called from HUInOutBL.updatePackingMaterialInOutLine() which sets the project on the packing material shipment line. Without this fix, the project won't propagate to packing material lines in most real-world scenarios.

---

### MEDIUM: Second cucumber scenario missing cleanup step

The first scenario ends with:
```
And set project type Sales/Purchase Order to inactive
```

The second scenario (C_Project_ID propagates to packing material invoice candidates) does NOT have this cleanup step. This could leave state behind that affects subsequent tests.

---

### MEDIUM: InOutLineHUPackingMaterialCollectorSource - unnecessary projectId in @EqualsAndHashCode

```java
@EqualsAndHashCode(of= {"productId", "recordId", "projectId"})
```

recordId is the M_InOutLine_ID which already uniquely identifies the record. Adding projectId changes equality semantics unnecessarily. While functionally harmless (same record = same project), it implies two records with the same M_InOutLine_ID could have different project IDs.

---

### LOW: Style inconsistency in M_ShipmentSchedule_StepDef

The new C_Project_ID validation uses old-style DataTableUtil while all other step defs in this PR use the modern DataTableRow API (row.getAsOptionalIdentifier(...)). This is because the surrounding method still uses the old Map parameter.

---

### LOW: PlainPackingMaterialDocumentLineSource.projectId missing @Nullable

The field is declared without @Nullable but the constructors accept @Nullable ProjectId. Should annotate the field for clarity.

---

### NIT: Formatting

- OrderLinePackingMaterialDocumentLineSource.getProjectId() one-liner brace style inconsistent with codebase
- Extra trailing blank lines in OrderLinePackingMaterialDocumentLine.java and InOutLineHUPackingMaterialCollectorSource.java  
- Copyright year 2025 in new C_Project_StepDef.java (should be 2026)

---

### Overall

The approach is solid. Threading ProjectId through the packing material document line interfaces and the AbstractPackingMaterialDocumentLinesBuilder key is the right design. The packing material key now correctly groups by (ProductId, ProjectId) which means different projects get separate packing material lines. The cucumber test coverage is comprehensive, covering both regular and packing material flows.

The .distinct() bug is the blocking issue. Without it, the feature won't work correctly when multiple source lines share the same project (which is the primary use case).

teosarca (approved):
All review items addressed in commit 7b248e1c:
- .distinct() added to getUniqueProjectIdOrNull() (critical bug fix)
- @EqualsAndHashCode reverted to original fields
- @Nullable added to projectId field
- Copyright year fixed
- Cosmetic cleanups

CI all green (31/31 pass). LGTM.

Note: the cleanup step was removed from scenario 1 rather than added to scenario 2 — acceptable since it apparently wasn't needed (CI passes without it).

PR #22473, opened 2026-02-27
EDI External Systems - Export Status update improvements and automated tests

mrie04:


teosarca (changes requested):
## Code Review — EDI External Systems: Export Status Improvements

Hi @mrie04, great work on this PR. The design spec is excellent and the architecture is well thought out — the error context propagation chain (Camel → JsonErrorItem → ExternalSystemService → IExternalSystemInvocationErrorListener → EDIExternalSystemErrorListener) is clean and extensible.

I've reviewed all 65 files. Here are the findings, grouped by priority.

---

### Blockers (need fixing before merge)

#### 1. Missing `PersonalDataCategory` in AD_Column inserts
**File:** `5790680_sys_EDI_AD_PInstance_ID_columns.sql`

Both `INSERT INTO AD_Column` statements (for `M_InOut.EDI_AD_PInstance_ID` column ID 592092 and `C_Invoice.EDI_AD_PInstance_ID` column ID 592093) are missing `PersonalDataCategory`. Since these columns store technical `AD_PInstance` references, they should be `'NP'`.

#### 2. Missing `ErrorCode` on AD_Message
**File:** `5790730_EDI_Desadv_Close_process.sql`

The new `AD_Message` with `MsgType='E'` (ID 545637, Value `EDI_Desadv_Close_UnresolvedInOuts`) is missing an `ErrorCode`. Error-type messages need ErrorCode so API consumers can programmatically distinguish errors.

#### 3. Cucumber feature bug — wrong identifier column in Scenario 2
**File:** `ediDesadvExportViaExternalSystem.feature`

In Scenario 2, the final re-enqueue step uses `EDI_Desadv_ID.Identifier` instead of `M_InOut_ID.Identifier`:
```gherkin
And M_InOut is enqueued for EDI export
  | EDI_Desadv_ID.Identifier |   ← should be M_InOut_ID.Identifier
  | d_2                      |
```

Also, the subsequent validation steps use bare `M_InOut_ID` and `EDI_Desadv_ID` without `.Identifier` suffix — looks like a copy-paste error from Scenario 1 where identifiers were correct.

#### 4. Wrong assertion in `ExternalSystem_Error_StepDef.sendErrorResponseForInvoice`
**File:** `ExternalSystem_Error_StepDef.java`

```java
final org.compiere.model.I_C_Invoice inout = invoiceTable.get(invoiceIdentifier);
assertThat(invoiceIdentifier).isNotNull();  // ← asserts identifier, should assert `inout`
```

The assertion checks the identifier (which is never null here), not the retrieved invoice record. Also, the variable is named `inout` instead of `invoice` (copy-paste from the shipment method).

#### 5. Missing `isNoSelection()` check in `EDI_Desadv_Close.checkPreconditionsApplicable`
**File:** `EDI_Desadv_Close.java`

The method only checks `!context.isSingleSelection()` but not `context.isNoSelection()`. Both `EDIExport` and `M_InOut_EnqueueForExport_ExternalSystem` in this same PR correctly check both. Pattern should be:
```java
if (context.isNoSelection()) {
    return ProcessPreconditionsResolution.rejectBecauseNoSelection();
}
if (!context.isSingleSelection()) {
    return ProcessPreconditionsResolution.rejectBecauseNotSingleSelection();
}
```

#### 6. `@Param` missing `mandatory = true` in `ChangeEDI_ExportStatus_M_InOut_SingleView`
The migration script registers the parameter with `IsMandatory='Y'`, but the Java annotation doesn't set `mandatory = true`. The Grid View version correctly has `mandatory = true`.

---

### High Priority (should fix)

#### 7. Potential infinite loop in bottom-up → top-down cascade
The chain `M_InOut.EDI_ExportStatus change → recomputeDesadvStatusFromInOuts → save desadv → EDI_Desadv.onDesadvStatusChanged` is currently safe because `isOneDesadvPerShipment=true` causes an early return. But this is a subtle invariant — if `isOneDesadvPerShipment` ever returns false unexpectedly, the system enters a cascade. Suggestion: add a guard in `recomputeDesadvStatusFromInOuts` that checks if the computed status actually differs from the current status before saving, or at minimum add a comment documenting this invariant.

#### 8. `ChangeEDI_ExportStatus_M_InOut_GridView.getParameterDefaultValue` may throw
`getSelectedInOutIds().iterator().next()` throws `NoSuchElementException` if the set is empty. While `checkPreconditionsApplicable` guards against this, `getParameterDefaultValue` may be called by the framework independently. Add a null-safety guard:
```java
final ImmutableSet<InOutId> selectedIds = getSelectedInOutIds();
if (selectedIds.isEmpty()) return null;
```

#### 9. `buildAggregatedErrorMessage` — unnecessary intermediate collect
```java
.collect(ImmutableList.toImmutableList())
.stream()
.reduce((a, b) -> a + "; " + b)
```
Should be simply `.collect(Collectors.joining("; "))`.

#### 10. `M_InOut_EnqueueForExport_ExternalSystem.doIt` missing `setPriority`
The comparable enqueue in `EDIExport.doIt` sets `.setPriority(IWorkPackageQueue.PRIORITY_AUTO)`. This new process omits it. Also uses `.bindToTrxName(get_TrxName())` where `EDIExport` uses `.bindToThreadInheritedTrx()` — is this difference intentional?

---

### Minor / Code Quality

- **Field naming:** `IExternalSystemInvocationErrorListeners` starts with uppercase `I` (interface convention) — should be `externalSystemInvocationErrorListeners`
- **Missing run-mode header:** `5790740_M_InOut_EnqueueForExport_ExternalSystem_process.sql` is missing `-- Run mode: SWING_CLIENT` header that other scripts in this PR have
- **Redundant `applies()` check:** `EDIExternalSystemErrorListener.onInvocationError` re-checks `applies()` even though the caller already did — either remove or document as defensive
- **Camel v1→v2 import change:** `LogMessageRouteBuilder` changes from v1 to v2 for `CreatePInstanceLogRequest` and `JsonPInstanceLog` — verify no other Camel route still depends on v1 classes

---

### Test Coverage Gap

The design spec covers DESADV with multiple shipments (Mode B/C mixed success/failure). The Cucumber tests only cover single-shipment scenarios. The **"Close DESADV"** process (`EDI_Desadv_Close`) is not covered by any test. Consider adding at least one multi-shipment scenario in a follow-up.

---

### What's Good

- The design spec is thorough and well-structured — it covers all modes, error cases, and recovery paths
- The error context propagation chain is clean and extensible
- The `ExternalSystemInvocationResult` value object properly replaces the previous `List<Exception>` pattern
- The `EDIExportStatus` helper methods (`isSent()`, `isDontSend()`, etc.) improve readability
- Bug fix in `C_BPartner_StepDef` (using correct `setEdiINVOIC_ExternalSystem_Config_ID`) is a genuine pre-existing fix
- Migration scripts correctly use `db_alter_table()` for DDL changes

Looking forward to the fixes — solid foundation here.

teosarca:
## Follow-up on QA Commit (11df93b8)

@mrie04 Great work on the quick turnaround — the QA commit addresses most of the review findings:

### Fixed
- [x] **PersonalDataCategory** — `'NP'` added to both AD_Column inserts
- [x] **Cucumber identifier bug** — Changed to `M_InOut.Identifier`
- [x] **Wrong assertion** — Now asserts `invoice` (not identifier), variable renamed
- [x] **isNoSelection()** — Added to `EDI_Desadv_Close`
- [x] **@Param mandatory** — Added `mandatory = true` to SingleView
- [x] **Infinite loop guard** — Moved `isOneDesadvPerShipment` guard into `propagateEDIStatus` itself — this is cleaner than my suggestion. The "should always be opposite of..." comments clearly document the invariant.
- [x] **GridView empty set** — Added empty check
- [x] **buildAggregatedErrorMessage** — Now uses `Collectors.joining`
- [x] **setPriority** — Added `PRIORITY_AUTO`
- [x] **Field naming** — Renamed to `externalSystemInvocationErrorListeners`
- [x] **Run mode header** — Added `-- Run mode: SWING_CLIENT`
- [x] **Redundant applies()** — Documented as safety check

### Still open

1. **Missing `ErrorCode` on AD_Message** (blocker #2 from original review) — The `AD_Message` with `MsgType='E'` (ID 545637, Value `EDI_Desadv_Close_UnresolvedInOuts`) in `5790730_EDI_Desadv_Close_process.sql` still has no `ErrorCode`. This needs an additional UPDATE statement in the migration script.

2. **Cucumber Scenario 2 validation steps** — The re-enqueue identifier was fixed (`M_InOut.Identifier`), but the subsequent validation `Then` steps still use bare `M_InOut_ID` and `EDI_Desadv_ID` without `.Identifier` suffix:
   ```gherkin
   Then after not more than 60s, M_InOut records have the following export status
     | M_InOut_ID | EDI_ExportStatus |
     | s_2        | S                |
   ```
   Scenario 1 uses `M_InOut_ID.Identifier` / `EDI_Desadv_ID.Identifier`. If the step defs accept both forms, this is fine — just flagging the inconsistency.

Once item 1 is addressed, I'm happy to approve.

teosarca (approved):
All review findings have been addressed across the QA commits (11df93b8, 9ede4513) and the `fix unwanted invoice ediStatusUpdate` commit (b3d9613d).

CI is fully green — all cucumber profiles, Java tests, frontend tests, mobile tests, and health check pass.

Well done @mrie04 — clean implementation with good test coverage.

PR #22475, opened 2026-02-27
SQL-Migrationscript fixes for long-living metasfresh-instances

metas-ts:


metas-ts:
@mergifyio copy soft_panda_release

PR #22487, opened 2026-02-28
gh#28433 Extend IncludedTabTopAction to discover processes from tab's table

teosarca:
## Summary
- **Process discovery**: When fetching top actions for an included tab, also query for processes registered against the tab's own table (not just the root/header table). Deduplicates by process ID.
- **Process execution**: `ExecuteUpdateSQL` now adds `selectedIncludedRecords` to the eval context so `@variable@` placeholders resolve from the selected tab row's columns.
- **5 files changed** across `de.metas.ui.web.base` and `de.metas.adempiere.adempiere/base`, all backward-compatible (new `@Nullable` fields with defaults, empty-set loop guard).

## Test plan
- [ ] Open a window with an included tab that has a process registered on the tab's own table (e.g., Bestellung → Aufteilung tab with "Zeile duplizieren")
- [ ] Verify the process button appears in the tab toolbar
- [ ] Select a row and click the button — verify the process executes correctly
- [ ] Verify existing IncludedTabTopAction processes (registered on the header table) still work unchanged
- [ ] Verify regular document actions and view quick actions are unaffected

### DIFF
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/ExecuteUpdateSQL.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/ExecuteUpdateSQL.java
@@ -95,7 +95,18 @@ public class ExecuteUpdateSQL extends JavaProcess
 		contexts.add(Evaluatees.ofRangeAwareParams(getParameterAsIParams()));
 
 		//
-		// 2: underlying record
+		// 2: selected included records (e.g. tab rows selected when invoking an IncludedTabTopAction)
+		for (final TableRecordReference includedRecordRef : getProcessInfo().getSelectedIncludedRecords())
+		{
+			final Evaluatee evalCtx = Evaluatees.ofTableRecordReference(includedRecordRef);
+			if (evalCtx != null)
+			{
+				contexts.add(evalCtx);
+			}
+		}
+
+		//
+		// 3: underlying (header) record
 		final String recordTableName = getTableName();
 		final int recordId = getRecord_ID();
 		if (recordTableName != null && recordId > 0)
@@ -109,7 +120,7 @@ public class ExecuteUpdateSQL extends JavaProcess
 		}
 
 		//
-		// 3: global context
+		// 4: global context
 		contexts.add(Evaluatees.ofCtx(getCtx()));
 
 		return Evaluatees.compose(contexts);
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/ExecuteUpdateSQL.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/ExecuteUpdateSQL.java
@@ -96,14 +96,7 @@ public class ExecuteUpdateSQL extends JavaProcess
 
 		//
 		// 2: selected included records (e.g. tab rows selected when invoking an IncludedTabTopAction)
-		for (final TableRecordReference includedRecordRef : getProcessInfo().getSelectedIncludedRecords())
-		{
-			final Evaluatee evalCtx = Evaluatees.ofTableRecordReference(includedRecordRef);
-			if (evalCtx != null)
-			{
-				contexts.add(evalCtx);
-			}
-		}
+		contexts.addAll(Evaluatees.ofTableRecordReferences(getProcessInfo().getSelectedIncludedRecords()));
 
 		//
 		// 3: underlying (header) record
--- a/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/process/DocumentPreconditionsAsContext.java
+++ b/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/process/DocumentPreconditionsAsContext.java
@@ -59,7 +59,6 @@ public final class DocumentPreconditionsAsContext implements WebuiPreconditionsC
 	@Getter
 	private final DisplayPlace displayPlace;
 
-	@Getter
 	@Nullable
 	private final String includedTabTableName;
 
@@ -148,5 +147,12 @@ public final class DocumentPreconditionsAsContext implements WebuiPreconditionsC
 	public OptionalBoolean isProcessedDocument()
 	{
 		return OptionalBoolean.ofBoolean(document.isProcessed());
+	}
+
+	@Nullable
+	@Override
+	public String getIncludedTabTableName()
+	{
+		return includedTabTableName;
 	}
 }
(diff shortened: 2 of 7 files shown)


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/ExecuteUpdateSQL.java
```diff
@@ -95,7 +95,18 @@ private Evaluatee getEvalContext()
 		contexts.add(Evaluatees.ofRangeAwareParams(getParameterAsIParams()));
 
 		//
-		// 2: underlying record
+		// 2: selected included records (e.g. tab rows selected when invoking an IncludedTabTopAction)
+		for (final TableRecordReference includedRecordRef : getProcessInfo().getSelectedIncludedRecords())
+		{
+			final Evaluatee evalCtx = Evaluatees.ofTableRecordReference(includedRecordRef);
```
teosarca: we shall avoid the SQL N+1.
If that's not feasible/easy, at least we have to isolate it in a separate method on document that on that method.
In this case we shall introduce org.compiere.util.Evaluatees#ofTableRecordReferences method that takes Collection<TableRecordReference> as param and returns a list of Evaluatee.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/process/adprocess/ADProcessDescriptorsFactory.java
```diff
@@ -128,6 +129,19 @@ public Stream<WebuiRelatedProcessDescriptor> streamDocumentRelatedProcesses(
 			final Stream<RelatedProcessDescriptor> tableRelatedProcessDescriptors = adProcessService.getRelatedProcessDescriptors(adTableId, adWindowId, adTabId)
 					.stream();
 			relatedProcessDescriptors = Stream.concat(relatedProcessDescriptors, tableRelatedProcessDescriptors);
+
+			// Also fetch processes registered against the included tab's own table (gh#28433)
+			final String includedTabTableName = preconditionsContext.getIncludedTabTableName();
+			if (!Check.isEmpty(includedTabTableName))
+			{
+				final AdTableId includedTabTableId = AdTableId.ofRepoId(adTableDAO.retrieveTableId(includedTabTableName));
```
teosarca: you shall use org.adempiere.ad.table.api.impl.ADTableDAO#retrieveAdTableId instead of org.adempiere.ad.table.api.impl.ADTableDAO#retrieveTableId.

also document on org.adempiere.ad.table.api.impl.ADTableDAO#retrieveTableId that developer shall consider using the other one.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/process/adprocess/ADProcessDescriptorsFactory.java
```diff
@@ -128,6 +129,19 @@ public Stream<WebuiRelatedProcessDescriptor> streamDocumentRelatedProcesses(
 			final Stream<RelatedProcessDescriptor> tableRelatedProcessDescriptors = adProcessService.getRelatedProcessDescriptors(adTableId, adWindowId, adTabId)
 					.stream();
 			relatedProcessDescriptors = Stream.concat(relatedProcessDescriptors, tableRelatedProcessDescriptors);
+
+			// Also fetch processes registered against the included tab's own table (gh#28433)
```
teosarca, line 133: in UAT docu, pls also document regression of this change

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/controller/WindowRestController.java
```diff
@@ -886,11 +886,26 @@ private JSONDocumentActionsList getDocumentActions(
 				: WebuiRelatedProcessDescriptor::isEnabled;
 
 		return documentCollection.forDocumentReadonly(documentPath, document -> {
+			// Resolve the included tab's table name so process discovery can also find
```
teosarca: consider extacting this code snippet to a dedicated method, i.e. getIncludedTabTableName

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/process/DocumentPreconditionsAsContext.java
```diff
@@ -59,12 +59,17 @@ public final class DocumentPreconditionsAsContext implements WebuiPreconditionsC
 	@Getter
 	private final DisplayPlace displayPlace;
 
+	@Getter
+	@Nullable
+	private final String includedTabTableName;
```
teosarca, line 63: yes, always prefer using lombok `@Getter`, `@Setter` etc instead of generating the code,
but there is an exception: the generated method overrides/impement the interface/base class method.
In this case we must generate the java method because we must use the `@Override` annotation to make sure the code is rock solid.

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/process/ExecuteUpdateSQL.java
teosarca: Done — created `Evaluatees.ofTableRecordReferences(Collection<? extends ITableRecordReference>)` that shares a single `PlainContextAware` across all loads. `ExecuteUpdateSQL.getEvalContext()` now uses `contexts.addAll(Evaluatees.ofTableRecordReferences(...))` instead of the per-record loop.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/process/adprocess/ADProcessDescriptorsFactory.java
teosarca: Done — changed to `adTableDAO.retrieveAdTableId(includedTabTableName)`. Also added `@apiNote` on `IADTableDAO.retrieveTableId` suggesting developers consider using `retrieveAdTableId` for the typed result.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/process/adprocess/ADProcessDescriptorsFactory.java
teosarca: Done — added two regression test scenarios to the UAT documentation: (4.4) no duplicate process buttons in standard windows, (4.5) process discovery performance check.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/controller/WindowRestController.java
teosarca: Done — extracted to `private static String getIncludedTabTableName(Document, DetailId, DisplayPlace)`.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/process/DocumentPreconditionsAsContext.java
teosarca: Done — removed Lombok `@Getter` and added explicit `@Override` method for `getIncludedTabTableName()`.

teosarca:

PR #22494, opened 2026-02-28
Restore grid row selection after browser-back from detail view

teosarca:
## Summary

- Cache selected row ID in Redux `listHandler` when navigating from grid to detail view
- Restore cached selection when grid reloads after browser-back navigation
- Previously, the first row was always re-selected regardless of which row was active before navigation

## Changes

**Frontend (5 files):**
- `frontend/src/constants/actions/ListTypes.js` — New `SET_LIST_SELECTED` action type
- `frontend/src/actions/ListActions.js` — New `setListSelected(rowId, viewId, windowType)` action creator
- `frontend/src/reducers/listHandler.js` — Store `selectedRow` alongside existing pagination/sorting/viewId cache
- `frontend/src/utils/documentListHelper.js` — Map `listSelectedRow` from Redux to component props + PropTypes
- `frontend/src/containers/DocumentList.js` — Cache selection on navigate, restore on grid reload via `restoreCachedSelection()`

**E2E (12 files):**
- Playwright test infrastructure (brought from `new_dawn_uat`)
- `e2e/frontend-webui/tests/spec/grid-selection-restore.spec.js` — E2E test covering the fix

## How it works

1. When user navigates from grid to detail view (Enter/double-click), `redirectToDocument()` dispatches `setListSelected(rowId, viewId, windowId)` to cache the current selection in Redux `listHandler`
2. When browser-back returns to the grid, `getData()` loads the view rows and then calls `restoreCachedSelection(viewId)`
3. `restoreCachedSelection()` checks that the cached `windowType + viewId` match, then dispatches `updateTableSelection()` to restore the row selection
4. The cache is cleared after each restoration to prevent stale restores

## Test plan

- [ ] Navigate to a grid view (e.g., Organisation window), select a non-first row, press Enter to open detail view, press browser Back — previously selected row should be re-selected (not first row)
- [ ] Navigate to a window via menu (no cached selection) — first row should be selected as before
- [ ] Navigate away from grid to a completely different window, then back via browser history — no stale selection should be restored (viewId mismatch clears cache)


e2e/frontend-webui/package.json
```diff
@@ -0,0 +1,32 @@
+{
```
adi-stefan, line 1: Scope question for a hotfix branch: this PR introduces an entire new Playwright E2E harness under `e2e/frontend-webui/` (~2000 LOC including a 1257-line `package-lock.json`, `.env.example`, config, page objects and a separate login spec). On a customer hotfix branch the usual expectation is the minimal fix; a brand-new test-infrastructure tree is a large merge/maintenance surface here (and the lockfile is exactly the kind of file that's causing the current merge conflict). Is the intent to land the harness on the hotfix, or would it be better introduced separately on the mainline/UAT branch with only the single regression spec (or a cucumber/existing-frontend test) riding along on the hotfix?

frontend/src/containers/DocumentList.js
```diff
@@ -629,6 +634,42 @@ class DocumentListContainer extends Component {
...
+      listSelectedRow.viewId !== viewId
+    ) {
+      return;
+    }
+
+    const tableId = getTableId({ windowId, viewId });
+
+    updateTableSelection({
+      id: tableId,
+      selection: [listSelectedRow.rowId],
```
adi-stefan, line 662: REQ-2 (graceful fallback when the previously-selected row no longer exists) doesn't appear to be explicitly handled: the cached `rowId` is re-selected without checking it's still present in the freshly-loaded rows. If the row was deleted or filtered out between leaving and returning, this selects a non-existent id. Does `updateTableSelection` degrade gracefully (no-op highlight) in that case, or should we guard by confirming the `rowId` exists in the loaded result before selecting? Asking because I haven't verified the downstream behaviour.

frontend/src/containers/DocumentList.js
```diff
@@ -629,6 +634,42 @@ class DocumentListContainer extends Component {
...
+      windowId,
+      viewId,
+      isModal,
+    });
+
+    // Clear the cached selection so it's not reused on fresh navigation
+    setListSelected(null, null, null);
```
adi-stefan, line 669: Minor: clearing via `setListSelected(null, null, null)` leaves `selectedRow = {rowId: null, viewId: null, windowType: null}` rather than the `null` used in `initialState`. The guard above handles it fine (windowType !== windowId), so this is purely a consistency nit — a dedicated clear/reset-to-null would match the initial shape.

adi-stefan (commented):
Peer review of the grid-selection-restore change. CI is green and the core fix is small and reads cleanly. Three points below — none are hard blockers on the production logic, but the first two are worth a decision before merge.

**Merge state:** the PR currently shows as CONFLICTING against the base branch — needs a rebase/merge to resolve before it can go in.

**Production diff (5 files in `frontend/src`):** the redux cache (`selectedRow` in `listHandler`), the `SET_LIST_SELECTED` action, the caching on navigate-to-detail (`redirectToDocument`, correctly gated behind `!isSideListShow` alongside the existing pagination/sort/listId caching), and the `restoreCachedSelection` on data-fetch success all hang together. Logic looks correct.

See inline comments for the specifics.

e2e/frontend-webui/package.json
adi-stefan: **Resolution (dev):** Resolved. The base branch already carries this e2e harness (README + three specs + shared utils), so the duplicate copy was the issue. The merge takes the base harness as canonical and keeps only the self-contained `grid-selection-restore` spec; the duplicate page-objects, the extra login spec, and the 1257-line `package-lock.json` (the conflict source) were dropped. Net footprint over the base branch is now the ~70-LOC frontend fix plus one spec. See https://github.com/metasfresh/metasfresh/commit/88d453cc710

frontend/src/containers/DocumentList.js
adi-stefan: **Resolution (dev):** No code change made. Based on reading the reducer (not yet a runtime test): `updateTableSelection` calls `doesSelectionExist` before applying the selection, so a cached row id that is absent from the reloaded rows should be a silent no-op and the grid should fall back to its default selection — i.e. the graceful fallback is expected to hold at the Redux level. This has NOT been exercised at runtime yet; it will be confirmed in the manual UAT demo (delete/filter the previously selected row, then navigate back). If UAT shows otherwise, an explicit existence guard will be added here.

frontend/src/containers/DocumentList.js
adi-stefan: **Resolution (dev):** Fixed. `listHandler` `SET_LIST_SELECTED` now returns `selectedRow: null` on the clear path (`rowId === null`), matching `initialState` and the `|| null` read site, instead of an all-null object. See https://github.com/metasfresh/metasfresh/commit/90b812a038b

PR #22522, opened 2026-03-02
E2E test: quick input multiple matches keep dropdown open

teosarca:
## Summary

- Add E2E test for `resolveItems()` multiple-match behavior: when typeahead returns 2+ products, dropdown stays open for manual selection (no auto-select)
- Regression test: single match still auto-selects and advances focus
- Both scenarios run in en_US and de_DE
- Update TEST-COVERAGE.md with new test entry (#9)

## Test plan

- [ ] CI passes (frontend webui test shards 1/3, 2/3, 3/3)
- [ ] New test `quick-input-multiple-matches.spec.js` runs in all 3 E2E shards without failures


teosarca:

PR #22646, opened 2026-03-07
gh#28336: fix Intrastat view M_InOut_V weight per line

metas-ts:
## Summary

- **Bug**: `M_InOut_V` used `Docs_Sales_InOut_Sum_Weight()` which returns the **total shipment weight** (sum of all lines). Every commodity group showed the full shipment weight instead of its per-line share.
- **Fix**: Replaced with per-line weight calculation (`catch weight → UOM convert to KG → qtyentered × product weight`) and `SUM(weight)` in the outer query instead of `GROUP BY weight`.
- **Test**: Added cucumber test (`intrastatViewWeight.feature`) that ships two products with different commodity numbers in one shipment and validates per-product weight (25kg / 15kg, not 40kg / 40kg).

## Test plan

- [ ] Migration script `5792720` applies cleanly
- [ ] Cucumber test `intrastatViewWeight.feature` passes
- [ ] Existing Intrastat export still works (manual verification on UAT)


metas-ts (commented):
lgtm

PR #22665, opened 2026-03-08
gh#28565 Fix Promotion Code Evaluation report

metas-ts:
## Summary

- Fix SQL bug in Promotion Code Evaluation report: `@Invoiced/null@` passed unquoted `Y`/`N` to PostgreSQL causing `column "n" does not exist` error. Fixed with `NULLIF` wrapping.
- Fix "Invoiced" parameter showing as "Aktiv" (wrong AD_Element 348 / IsActive). Now correctly shows "Fakturiert" / "Invoiced".
- Add DE/EN descriptions to process and all three parameters — previously all were null.

## Test plan

- [ ] Run report with no filters → returns all orders with promotion codes
- [ ] Run report with Invoiced=Y → returns only invoiced orders
- [ ] Run report with Invoiced=N → returns only non-invoiced orders
- [ ] Verify parameter labels show "Aktionskennzeichen", "Aktionskennzeichen 2", "Fakturiert"
- [ ] Verify process description is visible in the report dialog

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,9 +73,6 @@ misc/services/camel/de-metas-camel-shipping/tmp/
 .github/merge_to_master
 .github/integrate
 
-# workspace-local Maven repository
-.m2-local/
-
 # claude related
 .claude
 **/CLAUDE.md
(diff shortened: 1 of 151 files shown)


metas-ts (commented):
lgtm

PR #22697, opened 2026-03-09
`Package Licensing Movement Details` report update

riadmen:


metas-ts:
@riadmen there is a cucumber test failing. please find out which one. maybe we had a cucumber test for the 3rd process that we now removed.

PR #22757, opened 2026-03-09
gh#28336: add IsIncludeCSVHeaderRow flag to suppress CSV header

metas-ts:
## Summary
- Add `IsIncludeCSVHeaderRow` (YesNo, default Y) column to `AD_Process` to control whether CSV exports include a header row
- Set to `N` for `Intrastat_Export` (AD_Process_ID=585508) so Austrian RTIC no longer rejects submissions due to unexpected column header rows
- Wire through `SpreadsheetExportOptions` → `JdbcCSVExporter` → `CSVWriter`

## Test plan
- [x] 15 unit tests pass locally (5 new tests for header suppression)
- [x] CI green
- [x] Verify Intrastat CSV export produces no header row
- [x] Verify other CSV exports still include header row (backward compatible)


metas-ts (commented):
lgtm

PR #22771, opened 2026-03-09
Fix IsInheritPackingInstruction: German base language translations

metas-ts:
## Summary
- The original migration (5792610) set English text on the base `AD_Element` record, but metasfresh base language is `de_DE`
- This caused the field to display "Inherit Packing Instruction" (English) instead of "Packvorschrift vererben" (German) when logged in with `de_DE`
- Fix: update base `AD_Element` with German text, re-propagate to all downstream AD records

## Test plan
- [ ] Apply migration on test DB
- [ ] Log in with `de_DE` → field shows "Packvorschrift vererben"
- [ ] Log in with `en_US` → field shows "Inherit Packing Instruction"

### DIFF
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/order/C_CompensationGroup_CreateFromSchema_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/order/C_CompensationGroup_CreateFromSchema_StepDef.java
@@ -1,0 +1,166 @@
+/*
+ * #%L
+ * de.metas.cucumber
+ * %%
+ * Copyright (C) 2026 metas GmbH
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+package de.metas.cucumber.stepdefs.order;
+
+import de.metas.bpartner.BPartnerId;
+import de.metas.cucumber.stepdefs.DataTableRows;
+import de.metas.cucumber.stepdefs.hu.M_HU_PI_Item_Product_StepDefData;
+import de.metas.handlingunits.HUPIItemProduct;
+import de.metas.handlingunits.HUPIItemProductId;
+import de.metas.handlingunits.HuPackingInstructionsItemId;
+import de.metas.handlingunits.IHUPIItemProductDAO;
+import de.metas.handlingunits.model.I_M_HU_PI_Item;
+import de.metas.handlingunits.model.I_M_HU_PI_Item_Product;
+import de.metas.order.OrderId;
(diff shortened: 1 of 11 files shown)


backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/quickinput/orderline/OrderLineQuickInputProcessor.java
```diff
@@ -205,11 +209,25 @@ private static void applyPackingInstructionFromQuickInput(final QuickInput quick
 			return;
 		}
 
+		final IHUPIItemProductDAO piItemProductDAO = Services.get(IHUPIItemProductDAO.class);
+		final HUPIItemProduct mainPiItemProduct = piItemProductDAO.getById(piItemProductId);
```
metas-ts: these should be fields

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/order/C_CompensationGroup_CreateFromSchema_StepDef.java
```diff
@@ -0,0 +1,166 @@
...
+			final Group group = orderGroupsRepo.prepareNewGroup()
+					.groupTemplate(groupTemplate)
+					.qty(qty)
+					.createGroup(orderId, null);
+
+			// Apply PI inheritance if the schema flag is set and a PI was provided
+			if (groupTemplate.isInheritPackingInstruction())
+			{
+				row.getAsOptionalIdentifier("M_HU_PI_Item_Product_ID")
+						.ifPresent(piIdentifier -> {
```
metas-ts, line 130: AFAIU this should be *tested* . I mean the code in `OrderLineQuickInputProcessor.java should be under test - not duplicated

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/quickinput/orderline/OrderLineQuickInputProcessor.java
metas-ts: Fixed: Extracted PI inheritance logic into a shared `OrderGroupPIInheritanceService` in `de.metas.handlingunits.base`. `OrderLineQuickInputProcessor` now uses `piInheritanceService` field instead of local `Services.get()` calls.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/order/C_CompensationGroup_CreateFromSchema_StepDef.java
metas-ts: Fixed: Removed the duplicated method from the step def. Both `OrderLineQuickInputProcessor` and the Cucumber step def now call the same `OrderGroupPIInheritanceService.applyPackingInstructionInheritance()` — the logic lives in one place (`de.metas.handlingunits.base`).

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/order/OrderGroupPIInheritanceService.java
```diff
@@ -0,0 +1,93 @@
...
+	private final IHUPIItemProductDAO piItemProductDAO = Services.get(IHUPIItemProductDAO.class);
+
+	/**
+	 * Applies PI inheritance to all regular lines in the given compensation group.
+	 *
+	 * @param order                the target order (used to resolve BPartner for PI lookup)
+	 * @param group                the compensation group whose regular lines will be updated
+	 * @param mainPiItemProductId  the main article's PI Item Product (determines the TU type to match)
+	 */
+	public void applyPackingInstructionInheritance(
```
metas-ts, line 67: please look for a better place for this code. if there is non - fine

PR #22687, opened 2026-03-09
gh#28688 fix executor shutdown timeout: MICROSECONDS -> MILLISECONDS

metas-ts:
## Summary

- Fix `awaitTermination` using `TimeUnit.MICROSECONDS` instead of `MILLISECONDS` in `ThreadPoolQueueProcessor.shutdown0()` and `AsyncProcessorPlanner.shutdownExecutor()`
- The bug caused executor shutdown to wait only ~5ms per retry (25ms total) instead of the intended 5s per retry (25s total)
- This prevented graceful thread pool shutdown, causing thread leaks when pools were recreated — eventually exhausting OS thread limits (`OutOfMemoryError: unable to create new native thread`)

## Context

The bug has been present since 2015 (initial add of `ThreadPoolQueueProcessor`, commit `a9fb744b9538`) and was copy-pasted to `AsyncProcessorPlanner` in 2022 (PR #12603).

Identified during analysis of the keen_hawk app crash on 2026-03-09 where thread count grew unbounded (pool-35/thread-1465+), hitting the OS thread/PID limit.

## Test plan

- [ ] Verify async queue processor shutdown logs show actual wait time (~5s per retry instead of instant)
- [ ] Monitor thread count after deployment — should not grow unbounded over days

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #22783, opened 2026-03-09
Merge: inner_silence_hotfix → inner_silence_uat (gh#28688 thread leak fix)

metas-ts:
Forward merge of thread leak fix from inner_silence_hotfix to inner_silence_uat.

Source: #22705 (merged)
Original fix: #22687

**Must use merge-commit (not squash).**


metas-ts:
@mergifyio queue

PR #22785, opened 2026-03-09
Fix receipt schedule POReference overwrite and transport order loss on PO reactivation

metas-ts:
## Summary
- **POReference preserved**: When updating an existing receipt schedule (e.g. after PO reactivation + re-completion), the POReference is no longer overwritten with the order's value. Only new receipt schedules get the order's POReference.
- **Transport order preserved**: Shipping packages (`M_ShippingPackage`) are no longer deleted when a purchase order is reactivated. This preserves the transport order link on receipt schedules (virtual column via `M_ShippingPackage`). Reactivation is still blocked when the transport order is already processed.

## Test plan
- [x] `PurchaseOrderToShipperTransportationServiceTest` — 5 tests pass (3 existing + 2 new)
- [x] `OrderLineReceiptScheduleProducerTest` — 3 tests pass (1 existing + 2 new)
- [ ] CI green

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
@@ -378,9 +378,12 @@ public class C_Order
 	}
 
 	@DocValidate(timings = ModelValidator.TIMING_BEFORE_REACTIVATE)
-	public void deleteShippingPackageIfPossible(final I_C_Order order)
+	public void blockReactivationIfProcessedTransportationOrder(final I_C_Order order)
 	{
-		if (!purchaseOrderToShipperTransportationService.deleteShippingPackagesForOrderIfPossible(OrderId.ofRepoId(order.getC_Order_ID())))
+		// Block reactivation when the transport order is already processed,
+		// but do NOT delete shipping packages — they must be preserved so the
+		// transport order link on receipt schedules (virtual column via M_ShippingPackage) survives.
+		if (purchaseOrderToShipperTransportationService.hasProcessedShipperTransportation(OrderId.ofRepoId(order.getC_Order_ID())))
 		{
 			throw new AdempiereException(MSG_ORDER_ASSIGNED_TO_PROCESSED_TRANSPORTATION_ORDER);
 		}
--- a/backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
@@ -389,6 +389,16 @@ public class C_Order
 		}
 	}
 
+	@DocValidate(timings = ModelValidator.TIMING_AFTER_COMPLETE)
+	public void syncShippingPackagesFromOrder(final I_C_Order order)
+	{
+		if (order.isSOTrx())
+		{
+			return; // only for purchase orders
+		}
+		purchaseOrderToShipperTransportationService.syncShippingPackagesFromOrder(order);
+	}
+
 	@CalloutMethod(columnNames = I_C_Order.COLUMNNAME_PaymentRule)
 	public void checkPaymentRuleWithReservation(@NonNull final I_C_Order salesOrder, @NonNull final ICalloutField calloutField)
 	{
--- a/backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
@@ -380,6 +380,10 @@ public class C_Order
 	@DocValidate(timings = ModelValidator.TIMING_BEFORE_REACTIVATE)
 	public void blockReactivationIfProcessedTransportationOrder(final I_C_Order order)
 	{
+		if (order.isSOTrx())
+		{
+			return; // only for purchase orders
+		}
 		// Block reactivation when the transport order is already processed,
 		// but do NOT delete shipping packages — they must be preserved so the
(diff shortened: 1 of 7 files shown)


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/spi/impl/OrderLineReceiptScheduleProducer.java
```diff
@@ -170,7 +170,12 @@ else if (receiptScheduleBL.isClosed(receiptSchedule))
 		receiptSchedule.setC_BPartner_Location_ID(line.getC_BPartner_Location_ID());
 		final I_C_Order order = line.getC_Order();
 		receiptSchedule.setAD_User_ID(order.getAD_User_ID());
-		receiptSchedule.setPOReference(order.getPOReference());
+		if (isNewReceiptSchedule)
```
metas-ts: if there are no manual changed, i.e. the schedule's POReference is blank, then please do sync the order's POReference

backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
```diff
@@ -378,9 +378,12 @@ public void checkPricingConditionsInOrderLines(final I_C_Order order)
 	}
 
 	@DocValidate(timings = ModelValidator.TIMING_BEFORE_REACTIVATE)
-	public void deleteShippingPackageIfPossible(final I_C_Order order)
+	public void blockReactivationIfProcessedTransportationOrder(final I_C_Order order)
 	{
-		if (!purchaseOrderToShipperTransportationService.deleteShippingPackagesForOrderIfPossible(OrderId.ofRepoId(order.getC_Order_ID())))
+		// Block reactivation when the transport order is already processed,
```
metas-ts, line 387: if we don't delete the packaged, then we need to make sure that meningful changes are synched to the transportation-order after re-complete

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/spi/impl/OrderLineReceiptScheduleProducer.java
metas-ts: Done. Changed the guard to `if (isNewReceiptSchedule || Check.isBlank(receiptSchedule.getPOReference()))` — now syncs POReference from the order when blank (no manual changes), preserves when non-blank. Added unit test `poReference_isSyncedFromOrderWhenBlank` covering this case.

backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
metas-ts: Done. Added sync logic:

1. New `syncShippingPackagesFromOrder()` method in `PurchaseOrderToShipperTransportationService` — iterates existing `M_ShippingPackage` + `M_Package` records for the order and syncs:
   - `DatePromised` → `M_Package.ShipDate`
   - `C_BPartner_ID` + `C_BPartner_Location_ID` → both `M_ShippingPackage` and `M_Package`

2. New `@DocValidate(TIMING_AFTER_COMPLETE)` hook `syncShippingPackagesFromOrder` on `C_Order` interceptor — calls the sync method for purchase orders on re-completion.

3. Added unit test `syncShippingPackagesFromOrder_syncsDatePromisedAndLocation` and cucumber scenario "Shipping packages survive PO reactivation and are synced on re-complete".

metas-ts: what about properties like product or quenatity that might have changed? what about lines that are deleved or added while the order was reactivated?

metas-ts: Good point. Updated the sync logic to handle line-level changes:

**Removed lines**: `syncShippingPackagesFromOrder` now collects current order line IDs and deletes shipping packages (+ M_Package) for order lines that no longer exist. This covers the case where lines are deleted during reactivation.

**Changed lines** (product/quantity): The shipping packages reference the `C_OrderLine_ID`, which survives the reactivation. Product and quantity live on the order line itself — they don't need syncing to the shipping package since the package just points to the line. The transport order UI shows the current order line data via the FK.

**Added lines**: These are NOT auto-assigned to the transport order. The user must explicitly run `C_Order_AddTo_M_ShipperTransportation` again — same as the initial creation flow. Auto-creating packages for new lines would be surprising since the user might not want all lines on the same transport order.

All unit tests (7/7) and cucumber tests (2/2) pass locally.

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
```diff
@@ -291,18 +295,71 @@ public ReportResultData printSSCC18_Labels(@NonNull final OrderId orderId)
...
+				continue;
+			}
+
+			// Sync header-level fields
+			sp.setC_BPartner_ID(bPartnerId.getRepoId());
+			sp.setC_BPartner_Location_ID(BPartnerLocationId.toRepoId(bPartnerLocationId));
+			InterfaceWrapperHelper.save(sp);
+
+			// Also sync the underlying M_Package
+			final I_M_Package mPackage = InterfaceWrapperHelper.load(sp.getM_Package_ID(), I_M_Package.class);
```
metas-ts: this should be not in the service but in a repo

PR #22947, opened 2026-03-13
feat: Rewrite EPCIS SQL function to HU-based, DESADV-optional

metas-ts:
## Summary
- Rewrote `get_epcis_events_json_fn` to derive pallet/crate/item data from HU hierarchy instead of DESADV pack items
- DESADV is now LEFT JOINed (optional) — function works for shipments without DESADV
- Added `cuGTIN` (from `M_Product.GTIN`) to item output for correct SGTIN/LGTIN generation
- Warehouse GLN sourced from warehouse BPartner location with org fallback
- Handles both individual TUs (itemtype='HU') and aggregated TUs (itemtype='HA') with GRAI expansion
- Added Cucumber integration test validating function execution against seed data
- Migration script 5794490

## Test plan
- [x] SQL function executes without errors on 10 seed shipments
- [x] Cucumber test passes (load shipment → call function → validate JSON)
- [ ] Validate CASE B (aggregated TU) join path with production data containing real LU HUs
- [ ] CI green


metas-ts (commented):
👍

PR #23031, opened 2026-03-17
Overhaul and rename window sales&purchase statistics

metas-ts:
Also add payments related windows to one folder


metas-ts:
@mergifyio copy keen_hawk_hotfix

PR #23049, opened 2026-03-19
Created process to remove an MD_Candidate from ATP calculations

adi-stefan:
This pull request introduces the following changes:

- Implements the `MD_Candidate_Remove_From_ATP` SQL function and related process, enhancing the approach to remove candidates from ATP calculations, including recursive updates and thorough test coverage.


metas-rc (approved):
Verbally approved by Tobi

PR #23064, opened 2026-03-20
Skip fully-delivered shipment schedules in segment-based invalidation

metas-ts:
## Summary

Don't flood the workpackage queue with redundant shipment schedule WPs. One WP is enough to process all pending work — additional WPs just occupy queue slots (~0.5s poll interval each) and exit after finding nothing to do.

- **Commit 1** (reverted in commit 4): ~~SQL filter on segment-based invalidation~~ — removed per review
- **Commit 2**: Port from `new_dawn_uat` (`eeec5ccf5af`): skip enqueue for `CreateMissingShipmentSchedules` if processable WPs already exist
- **Commit 3**: Same pattern for `UpdateInvalidShipmentSchedules`: skip enqueue if processable WPs already exist
- **Commit 4**: Address review — revert SQL filter, add test for queue dedup

## Problem

On 2026-03-20, completing a 57-line sales order on a system with 15,633 shipment schedules took ~10 minutes to create the schedules. Root cause: **~2,250 redundant `UpdateInvalidShipmentSchedules` workpackages** flooded the single-threaded queue processor.

Each WP only takes 5-20ms to process (first one does the work, rest find nothing). But the queue processor has a ~0.5s poll interval per WP, so: **2,250 x 0.5s = ~19 minutes** to drain the queue, blocking all other shipment schedule operations.

## Fix

Check `queueForEnqueuing.size() > 1` before enqueuing both `CreateMissingShipmentSchedules` and `UpdateInvalidShipmentSchedules` WPs. One WP is enough — it processes ALL pending work from `M_ShipmentSchedule_Recompute`.

This is the same pattern already proven in `new_dawn_uat` for `CreateMissing` (commit `eeec5ccf5af`), now also applied to `UpdateInvalid`.

## Test plan

- [x] `UpdateInvalidShipmentSchedulesWorkpackageProcessorTest` (3 tests: empty queue, 1 WP, multiple WPs)
- [x] Existing `ShipmentScheduleUpdaterTest` passes
- [ ] CI green


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/invalidation/impl/ShipmentScheduleInvalidateRepository.java
```diff
@@ -291,6 +291,19 @@ public void invalidateStorageSegments(
 		sqlWhereClause.append(ssAlias + I_M_ShipmentSchedule.COLUMNNAME_Processed).append("=?");
 		sqlParams.add(false);
 
+		// Not Closed
```
metas-ts: these conditions add FUD and are now IMO not needed anymore

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/async/UpdateInvalidShipmentSchedulesWorkpackageProcessor.java
```diff
@@ -68,8 +71,22 @@ public static void schedule(@NonNull final ShipmentSchedulesUpdateSchedulerReque
 		_schedule(request);
 	}
 
+	private static final IWorkPackageQueueFactory workPackageQueueFactory = Services.get(IWorkPackageQueueFactory.class);
+
 	private static void _schedule(@NonNull final ShipmentSchedulesUpdateSchedulerRequest request)
 	{
+		final ILoggable loggable = Loggables.withLogger(logger, Level.DEBUG);
```
metas-ts, line 76: can you come up with a test for this code?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/async/UpdateInvalidShipmentSchedulesWorkpackageProcessor.java
```diff
@@ -23,7 +23,9 @@
  */
```
metas-ts, line 23: Reverted the SQL filter in b90e95e — you are right, the queue dedup (commits 2+3) is the proper fix. The SQL filter was defense-in-depth but adds complexity without clear benefit now.

Added test for the UpdateInvalid queue dedup: `UpdateInvalidShipmentSchedulesWorkpackageProcessorTest` with 3 tests (empty queue, 1 WP, multiple WPs). Also moved `Services.get(IWorkPackageQueueFactory)` into the method body so it can be mocked in tests.

PR #23092, opened 2026-03-21
Fix profile3 Cucumber: move frontendTesting to executor6

metas-ts:
## Summary
Move `frontendTestingOrderLifecycle.feature` from executor3 to executor6 to avoid DB state pollution that causes `productionCandidate.feature` PP_Order_Candidate assertion failures.

## Root cause
The frontendTesting feature creates orders, pricing systems, and price lists on executor3. This pollutes the shared DB state for `productionCandidate.feature` (also executor3), causing `PP_Order_Candidate` validation to fail:
- `QtyToProcess expected: 0 but was: 10`
- `QtyProcessed expected: 10 but was: 0`

## Fix
Change `@ghActions:run_on_executor3` → `@ghActions:run_on_executor6`


metas-ts:
Wrong branch — contains all commits from the merged PR. Creating a clean fix branch.

PR #23105, opened 2026-03-23
Fix mobile UI logout button not visible on small screens

teosarca:
## Summary
- The applications list screen (launcher) on the mobile UI was not scrollable, causing the logout button to be clipped on devices with small screen heights (e.g. ~450px)
- Made the `.applications-list` container a flex column filling the viewport, with the `.section` area scrollable via `overflow-y: auto`
- Fixed a bug in the responsive media query which used `max-width` instead of `max-height` for the small screen height breakpoint

## Test plan
- [ ] Open mobile UI launcher screen on a device or browser with ~450px viewport height
- [ ] Verify all buttons including "Abmelden" (logout) are visible (scroll if needed)
- [ ] Verify the logout button works when tapped
- [ ] Verify normal-height screens still look correct (no unnecessary scrollbar)

### DIFF
--- a/e2e/mobile-webui/tests/spec/login.spec.js
+++ b/e2e/mobile-webui/tests/spec/login.spec.js
@@ -23,7 +23,7 @@ test.describe('Login/Logout', () => {
                 const response = await Backend.createMasterdata({
                     request: {
                         mobileConfig: {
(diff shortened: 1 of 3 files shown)


teosarca:
@claude review

PR #23116, opened 2026-03-23
Robust BPartner lookup in bank statement and DATEV payment imports

metas-ts:
## Summary

- **Bank statement import** (`BankStatementImportTableSqlUpdater`): When multiple active BPartners share the same DebtorId or CreditorId (after relaxing the unique constraint), the subquery now uses `HAVING COUNT(*)=1` to only match when exactly one BPartner fits. Ambiguous matches are flagged with `I_IsImported='E'` and a clear error message. Also adds `IsActive='Y'` filter and fixes an AND/OR precedence bug where the `AD_Org_ID` check only applied to the CreditorId match branch.
- **DATEV payment import** (`CPaymentImportTableSqlUpdater`): Replaces `MAX(C_BPartner_ID)` (which silently picked an arbitrary partner on duplicates) with the same `HAVING COUNT(*)=1` pattern. Adds `IsActive='Y'` filter and ambiguous-match error flagging.
- Both fixes cover DebtorId **and** CreditorId, so if either unique constraint is relaxed, the imports remain safe.

## Test plan

- [ ] Import a bank statement where one import line's `DebitorOrCreditorId` matches exactly one active BPartner → BPartner is correctly assigned
- [ ] Import a bank statement where one import line's `DebitorOrCreditorId` matches two active BPartners with the same DebtorId → line is flagged with `I_IsImported='E'` and error message "Multiple BPartners found..."
- [ ] Import a DATEV payment file where `BPartnerValue` matches exactly one active BPartner's DebtorId/CreditorId → BPartner is correctly assigned
- [ ] Import a DATEV payment file where `BPartnerValue` matches multiple active BPartners → line is flagged with error message
- [ ] Verify inactive BPartners (`IsActive='N'`) are excluded from matching in both importers

### DIFF
--- a/backend/de.metas.banking/de.metas.banking.base/src/main/java/de/metas/banking/impexp/BankStatementImportTableSqlUpdater.java
+++ b/backend/de.metas.banking/de.metas.banking.base/src/main/java/de/metas/banking/impexp/BankStatementImportTableSqlUpdater.java
@@ -259,22 +259,20 @@ public class BankStatementImportTableSqlUpdater
 				+ " HAVING COUNT(*) = 1"
 				+ ") "
 				+ "WHERE i." + I_I_BankStatement.COLUMNNAME_C_BPartner_ID + " IS NULL "
-				+ "AND i.I_IsImported<>'Y' "
-				+ "OR i.I_IsImported IS NULL")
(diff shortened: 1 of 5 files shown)


metas-ts:
@claude review opus

PR #23123, opened 2026-03-24
Fix M_QtyReservation AD metadata on preloaded DBs

teosarca:
## Summary

Idempotent fixup migration for `M_QtyReservation` AD metadata that gets lost when the preloaded DB image already contains `AD_Element` 584594.

**Root cause:** Migration `5790900` runs in a single transaction. If `AD_Element` 584594 already exists (baked into the preloaded image), the INSERT fails → entire transaction rolls back → all 18 `AD_Column` INSERTs are lost. The physical table survives (`CREATE TABLE IF NOT EXISTS`) but the AD metadata is gone.

**Symptom:** Health check fails with `No key columns defined for M_QtyReservation` on any branch that includes the Qty Reservation window (script `5792870`).

**Fix:** Re-insert all AD metadata (element, reference, ref_list, table, 21 columns) with `ON CONFLICT DO NOTHING`. Safe to run on any DB state — no-op if metadata already exists.

## Test plan
- [ ] CI health check passes (was failing on `soft_panda_release_grai`)


teosarca:
closing, it seems is no longer needed

PR #23124, opened 2026-03-24
correct the PP_Order_candidate valrule

metas-rc:


metas-ts (approved):
to me the code looks ok, but please check the failig test reports

metas-rc:
> to me the code looks ok, but please check the failig test reports

Error corrected in the script. Tests are now green.

PR #23165, opened 2026-03-25
Purchase Pro Forma Invoice

mrie04:


mrie04:
@claude review

backend/de.metas.business/src/main/java/de/metas/invoice/service/impl/AbstractInvoiceBL.java
```diff
@@ -573,6 +578,10 @@ public boolean setPaymentStatus(
 			@NonNull final BigDecimal openAmt,
 			@NonNull final InvoicePaymentStatus paymentStatus)
 	{
+		if(!invoice.isFinancial())
+		{
+			invoice.setIsPaid(false);
```
teosarca, line 583: which is the case here. shall we also assume openAmt=0, paymentStatus=NOT_PAID?

teosarca (changes requested):
check my cmments

backend/de.metas.business/src/main/java-legacy/org/compiere/model/MInvoice.java
```diff
@@ -106,7 +106,17 @@
...
-	private final IDocTypeBL docTypeBL = Services.get(IDocTypeBL.class);
+	private IDocTypeBL docTypeBL;
+
+	private IDocTypeBL getDocTypeBL()
+	{
+		if (docTypeBL == null)
+		{
+			docTypeBL = Services.get(IDocTypeBL.class);
+		}
+		return docTypeBL;
```
teosarca, line 118: IDocTypeBL docTypeBL = this._docTypeBL;
if (docTypeBL == null)
{
		docTypeBL = this._docTypeBL = Services.get(IDocTypeBL.class);
}
return docTypeBL;

teosarca (approved):
check my cmments

teosarca:
## Iter-2 follow-up — `proformaInvoiceOpen` superseded by unified `invoiceOpenToDate`

This PR introduced `IsFinancial` on `C_Invoice` and gated `C_Invoice_v` with `IsFinancial='Y'`, which excludes proforma invoices from the financial-invoice consumer set (dunning, BPartner stats, payment-allocation pickers, open-items reports). That gating is **kept**.

Iter-2 https://github.com/metasfresh/metasfresh/pull/23682 originally introduced a separate `proformaInvoiceOpen()` SQL function to compute the open amount for proforma invoices in the pay-selection flow (the only consumer that needs to see proformas). Code review of that PR surfaced two issues:

1. The separate function duplicated the `invoiceOpenToDate` allocation-walk pattern, but proforma invoices have **no `C_AllocationLine` rows by design** (no accounting for proforma payments — hard constraint), so the allocation walk always returned `GrandTotal`. The function was structurally broken.
2. Maintaining two open-amount functions diverged the architecture.

**Resolution** (https://github.com/metasfresh/metasfresh/pull/23682 commit `8a36a7f8885`): `proformaInvoiceOpen()` is dropped. `invoiceOpenToDate` gains an internal early-return branch on `IsFinancial='N'` that uses `C_Payment.Proforma_Invoice_ID + DocStatus IN ('CO','CL')` for paid-detection — the iter-2-introduced linkage. The `IsFinancial='Y'` gate on `C_Invoice_v` is preserved (no changes to dunning / BPstats / allocation-picker behaviour).

The proforma branch:
- Reads invoice metadata directly from `C_Invoice` (not `C_Invoice_v`)
- Uses `SUM(PayAmt) FILTER (Proforma_Invoice_ID = $1 AND DocStatus IN (CO, CL) AND IsActive = 'Y')` for paid-detection
- Honours `p_DateType` / `p_Date` / `p_Result_Currency_ID` / `p_Exclude_Payment_IDs` for parity
- Uses precision-derived snap-to-zero (matching the financial branch) instead of the hardcoded ±0.005 in the now-deleted function
- Returns `HasAllocations='N'` (proformas have no allocation rows by design)

A new full-payment-only guard on `C_Payment` `BEFORE_PREPARE` (AC #16, migration `5799640_*`) ensures `abs(PayAmt) = proforma.GrandTotal` for proforma payments — partial proforma payments are forbidden because there's no allocation mechanism to reconcile residual balances.

The `IsFinancial` mechanism this PR introduced remains the right boundary for "should this consumer see proformas". The unification only changes how the open-amount computation is co-located.

PR #23195, opened 2026-03-27
fix: SFTP endpoints cannot be saved — AuthType/Type NOT NULL

metas-ts:
## Summary

- `AuthType` and `Type` columns on `ExternalSystem_Endpoint` were `IsMandatory=Y` with `NOT NULL` and no `MandatoryLogic`, causing save failures when creating SFTP endpoints (these are HTTP-only fields)
- Both columns are now conditionally mandatory only when `TransportType=HTTP`
- `Type` field is hidden when `TransportType=SFTP` (same as `AuthType` already was)
- Interceptor clears all HTTP auth fields when switching to SFTP transport
- Java domain model: `authType` changed from `@NonNull` to `@Nullable` with null-safe serialization

## Test plan

- [ ] Create new ExternalSystem_Endpoint with TransportType=SFTP — verify record saves without errors
- [ ] Verify `Art` (Type) and `Authentifizierung` (AuthType) fields are hidden when TransportType=SFTP
- [ ] Verify both fields are visible and mandatory when TransportType=HTTP
- [ ] Switch existing HTTP endpoint to SFTP — verify HTTP auth fields are cleared
- [ ] Unit tests pass: `ExternalSystemEndpointTest`, `EndpointAuthTypeTest`, `SftpAuthTypeTest`, `TransportTypeTest`


backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/scriptedexportconversion/ExternalSystemScriptedExportConversionService.java
```diff
@@ -145,6 +153,26 @@ public Map<String, String> getParameters(
...
+	/**
+	 * Tries to read the DocumentNo from the outbound record table.
+	 * Returns null if the table doesn't have a DocumentNo column or the record is not found.
+	 */
+	@Nullable
+	private static String retrieveDocumentNo(@NonNull final String tableName, @NonNull final String recordId)
+	{
+		try
+		{
+			final String sql = "SELECT DocumentNo FROM " + tableName
```
metas-ts: we don't do this- no hardcoded SQL. i'm sure there is a service which dies this

e2e/mobile-webui/tests/utils/screens/huManager/GRAIScreen.js
```diff
@@ -30,7 +30,9 @@ export const GRAIScreen = {
 
     scanGraiBarcode: async ({ barcodeString }) => await test.step(`${NAME} - Scan GRAI barcode`, async () => {
         await GRAIScreen.expectVisible();
-        await BarcodeScannerComponent.type(barcodeString);
+        // Use typeBatch with a single code so Enter is dispatched after the barcode,
+        // triggering immediate processing instead of relying on the interval timer flush.
```
metas-ts, line 34: what is this change about? does it really belong into this change, or did it accidentally slip in?

backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/scriptedexportconversion/ExternalSystemScriptedExportConversionService.java
metas-ts: Fixed in e000a94 — now uses `IDocumentBL.getDocumentNo(ctx, adTableId, recordId)` which handles the DocumentNo/Value/Name fallback chain properly, no hardcoded SQL.

e2e/mobile-webui/tests/utils/screens/huManager/GRAIScreen.js
metas-ts: Intentional fix for flaky CI mobile tests. `scanGraiBarcode` previously used `BarcodeScannerComponent.type()` which dispatches keyboard events **without Enter** — the barcode is only submitted after a 100ms interval flush timer. In CI headless Chrome, this timing is unreliable and caused:
- "Non-GRAI barcode silently ignored" — expected 1 chip, got 0
- "Clear All GRAIs" — expected 0 chips, got 1

Now uses `typeBatch` (single-element array) which dispatches Enter after the barcode, triggering immediate `onReadDone` instead of relying on the timer. The GRAI tests on profile3 were green on the previous CI run but the mobile test shard kept failing — this fix makes the barcode processing deterministic.

PR #23281, opened 2026-03-31
Fix mobileUI overdelivery prompt not appearing during picking

metas-ts:
## Summary
- The overdelivery confirmation prompt ("Do you really want to pack more than ordered?") was never shown in the mobile UI picking workflow, even with `IsShowConfirmationPromptWhenOverPick=Y`
- **Two root causes fixed:**
  1. **Manual qty entry**: `ScanHUAndGetQtyComponent.validateQtyEntered()` enforces a hard `qtyMax` limit that disables the "Done" button when qty > remaining. This prevents the confirmation prompt from ever being reached. **Fix**: skip `qtyMax` validation when the over-pick confirmation prompt is enabled.
  2. **Whole-TU scanning** (`pickWholeTU=true`): The dialog pre-fills qty = remaining, but the backend picks the actual HU storage qty (which may be higher). The frontend didn't know the HU's actual qty, so it couldn't detect overdelivery. **Fix**: When scanning a whole-TU, the frontend now fetches the HU's actual product qty from the backend via `/hu/byScannedCode` (extended with `productNo` param and `productQty`/`productUom` response fields) and uses it as the initial qty, allowing the prompt to fire.

## Test plan
- [ ] Set `IsShowConfirmationPromptWhenOverPick=Y` in `MobileUI_UserProfile_Picking`
- [ ] **Manual entry**: Create a picking job, enter a quantity greater than remaining - verify the over-pick confirmation prompt appears
- [ ] **Whole-TU scanning**: Scan a TU barcode that contains more product than remaining - verify the prompt appears with the actual HU qty shown
- [ ] Click "Yes" on prompt - pick is submitted
- [ ] Click "No" on prompt - dialog stays, user can cancel
- [ ] With `IsShowConfirmationPromptWhenOverPick=N`: verify manual entry still hard-blocks qty above max (existing behavior)

### DIFF
--- a/backend/de.metas.picking.rest-api/src/main/java/de/metas/picking/rest_api/PickingRestController.java
+++ b/backend/de.metas.picking.rest-api/src/main/java/de/metas/picking/rest_api/PickingRestController.java
@@ -253,12 +253,25 @@ public class PickingRestController
 		}
 		final JsonHU hu = hus.get(0);
 
-		return JsonHUInfo.builder()
+		final JsonHUInfo.JsonHUInfoBuilder builder = JsonHUInfo.builder()
 				.id(hu.getId())
 				.unitType(hu.getUnitType())
 				.qtyTUs(hu.getQtyTUs())
-				.huQRCode(hu.getQrCode())
-				.build();
+				.huQRCode(hu.getQrCode());
+
+		// gh#29069: return product qty for overdelivery check
+		final String productNo = request.getProductNo();
+		if (productNo != null && hu.getProducts() != null)
+		{
+			hu.getProducts().stream()
+					.filter(p -> productNo.equals(p.getProductValue()))
+					.findFirst()
+					.ifPresent(p -> builder
+							.productQty(new java.math.BigDecimal(p.getQty()))
+							.productUom(p.getUom()));
+		}
+
+		return builder.build();
 	}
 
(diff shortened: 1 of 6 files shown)


teosarca:
Closing this PR — replacing with a corrected implementation that gates the new behavior on `isShowPromptWhenOverPicking` to avoid the `pick_by_HUId.spec.js` regression (qtyTUs leak into non-LU-TU scenarios causing qtyInitial=1 instead of remaining). Will open a new PR with the fix.

teosarca:
**Resolution (dev):** Addressed — this PR was superseded as described, and the replacement is merged.

The corrected implementation gating the new behaviour on `isShowPromptWhenOverPicking` (avoiding the `pick_by_HUId.spec.js` regression where `qtyTUs` leaked into non-LU-TU scenarios and produced `qtyInitial=1` instead of remaining) shipped as https://github.com/metasfresh/metasfresh/pull/23307 — merged 2026-04-01 onto `intensive_care_hotfix`, commit https://github.com/metasfresh/metasfresh/commit/1f4a4f0f1ae.

Follow-up test hardening: https://github.com/metasfresh/metasfresh/pull/23316 — commit https://github.com/metasfresh/metasfresh/commit/ec17648ec1e.

No further action on this closed PR.

PR #23320, opened 2026-04-02
Cap reservation CU qty at order line QtyOrdered

teosarca:
## Summary
- Prevents over-reservation when the cockpit row's CU/TU ratio differs from the order line's effective CU/TU (partially-filled last TU)
- Caps reservation Qty at the order line's QtyOrdered in `QtyReservationService.makeReservation()`
- Adds `CreateQtyReservationRequest.withQty()` for immutable copy with capped qty

## Context
Real-world example on seefruchtuat: order for 20 Stück in 3 TU (capacity 9.9/TU), but stock TUs average 10 CU each. Without cap: 3 × 10 = 30 CU reserved. With cap: 20 CU (= QtyOrdered).

## Test plan
- [ ] Cucumber test `S_QtyRes_220` in se203 repo validates the capping logic
- [ ] Existing reservation tests still pass (no behavioral change when CU ≤ QtyOrdered)

### DIFF
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/qty_reservation/CreateQtyReservationRequest.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/qty_reservation/CreateQtyReservationRequest.java
@@ -29,4 +29,20 @@ public class CreateQtyReservationRequest
 	@NonNull Quantity qty;
 	@Nullable AttributesKey attributesKey;
 	@Nullable ProjectId projectId;
+
+	public CreateQtyReservationRequest withQty(@NonNull final Quantity newQty)
+	{
+		return builder()
+				.orderAndLineId(orderAndLineId)
+				.productId(productId)
+				.warehouseId(warehouseId)
+				.vendorBPartnerId(vendorBPartnerId)
+				.supplyType(supplyType)
+				.datePromised(datePromised)
+				.qtyTU(qtyTU)
+				.qty(newQty)
+				.attributesKey(attributesKey)
+				.projectId(projectId)
+				.build();
+	}
 }
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/qty_reservation/CreateQtyReservationRequest.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/qty_reservation/CreateQtyReservationRequest.java
@@ -16,7 +16,7 @@ import javax.annotation.Nullable;
 import java.time.Instant;
 
 @Value
-@Builder
+@Builder(toBuilder = true)
 public class CreateQtyReservationRequest
 {
 	@NonNull OrderAndLineId orderAndLineId;
@@ -29,20 +29,4 @@ public class CreateQtyReservationRequest
 	@NonNull Quantity qty;
 	@Nullable AttributesKey attributesKey;
 	@Nullable ProjectId projectId;
-
-	public CreateQtyReservationRequest withQty(@NonNull final Quantity newQty)
-	{
-		return builder()
-				.orderAndLineId(orderAndLineId)
-				.productId(productId)
-				.warehouseId(warehouseId)
-				.vendorBPartnerId(vendorBPartnerId)
-				.supplyType(supplyType)
-				.datePromised(datePromised)
-				.qtyTU(qtyTU)
-				.qty(newQty)
-				.attributesKey(attributesKey)
-				.projectId(projectId)
-				.build();
-	}
 }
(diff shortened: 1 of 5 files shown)


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/qty_reservation/CreateQtyReservationRequest.java
```diff
@@ -29,4 +29,20 @@ public class CreateQtyReservationRequest
 	@NonNull Quantity qty;
 	@Nullable AttributesKey attributesKey;
 	@Nullable ProjectId projectId;
+
+	public CreateQtyReservationRequest withQty(@NonNull final Quantity newQty)
```
teosarca: pls use `@Builder(toBuilder=true)` on class level
and here u have to call toBuilder().qty(newQty).build()

teosarca: or simply use lombok `@With`  annotation on qty field

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/qty_reservation/QtyReservationService.java
```diff
@@ -54,6 +56,27 @@ public QtyReservationId makeReservation(@NonNull final CreateQtyReservationReque
...
 	}
 
+	/**
+	 * Cap the reservation's CU qty so it does not exceed the order line's QtyOrdered.
+	 * <p>
+	 * This prevents over-reservation when the cockpit row's CU/TU ratio differs from
+	 * the order line's effective CU/TU (e.g. partially-filled last TU:
+	 * order needs 20 CU in 3 TU, but stock TUs average 10 CU each → uncapped = 30 CU).
+	 */
+	private CreateQtyReservationRequest capQtyToOrderLine(
```
teosarca: not sure if we shall do this here.... IMHO this is too late, the request shall already have the right qty.
If we do here, is just a hack, because we take care about only one corner case, but we don't take care about how much was alrady reserved on that order etc.... so to me is wrong place and spreading the BL

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/qty_reservation/MakeQtyReservationCommand.java
```diff
@@ -28,6 +31,10 @@ public class MakeQtyReservationCommand
 
 	public QtyReservationId execute()
 	{
+		final Quantity qtyCUFromStock = rowVO.computeQtyCUToReserve(qtyToReserveTU);
+		final Quantity qtyOrdered = Services.get(IOrderLineBL.class).getQtyOrdered(salesOrderAndLineId);
```
teosarca: rule: services shall be loaded and declared as class fields.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/qty_reservation/MakeQtyReservationCommand.java
```diff
@@ -28,6 +31,10 @@ public class MakeQtyReservationCommand
...
+		final Quantity qtyCU = qtyCUFromStock.min(qtyOrdered);
```
teosarca: pls document the reason of this beucase it might not be obvious.


consider extracting these 2 lines in some computeQtyCUToReserve() method in this class, if that would make the things cleaner.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/qty_reservation/CreateQtyReservationRequest.java
```diff
@@ -16,7 +16,7 @@
 import java.time.Instant;
 
 @Value
-@Builder
+@Builder(toBuilder = true)
```
teosarca: i think this is no longer needed because we no longer use toBuilder()

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/v2/reservation/MD_CockpitV2_MakeQtyReservation.java
```diff
@@ -23,6 +24,7 @@ public class MD_CockpitV2_MakeQtyReservation
 		extends MaterialCockpitV2BasedProcess
 		implements IProcessPrecondition, IProcessDefaultParametersProvider
 {
+	@Autowired private IOrderLineBL orderLineBL;
```
teosarca: rule: check if the service can be used with spring auto-wiring or we shall use Services.get().
i.e. 
* a service can be spirng autowired if the class or the implementation class if this is an interface has one of the `@Component` spring annotaions (Service, Repository etc).
* if the interface is implementing ISingletonService and the implementation classs does not have a spring component annotation then the service shall be injected with Services.get
* else there is no way to injected it shall be provided at construction time or as parameter or ask human what to do
---

In this case you shall use Services.get

PR #23400, opened 2026-04-10
fix: return null when TU PI is missing in material tracking invoicing

metas-ts:
## Summary

- Fix `Assumption failure: tuPI not null` crash in `HUHandlingUnitsInfoFactory.createFromInOutLine()` when an InOutLine has no TU HU_PI assigned
- The method now returns `null` (which callers already handle) instead of passing `null` to the `HUHandlingUnitsInfo` constructor which crashes
- Resolves a long-standing TODO in the code ("shall we just return null or throw exception?")

## Context

When creating invoice candidates for a manufacturing order (quality inspection / "Waschprobe"), the system iterates over vendor receipt InOutLines and creates `HUHandlingUnitsInfo` for each. If an InOutLine has no TU PI assigned, `getTU_HU_PI()` returns null, and the constructor's `Check.assume(tuPI != null)` throws an exception. This blocks invoice candidate creation entirely.

The caller `InOutLineAsVendorReceipt.loadQtysIfNeeded()` already has a null check for the returned `IHandlingUnitsInfo` (line 215: `if (handlingUnitsInfo == null) { // do nothing }`), so returning null is safe.

## Test plan

- [ ] CI green
- [ ] Verify that manufacturing orders with InOutLines missing TU PI no longer crash during invoice candidate creation

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/materialtracking/spi/impl/HUHandlingUnitsInfoFactory.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/materialtracking/spi/impl/HUHandlingUnitsInfoFactory.java
@@ -103,7 +103,9 @@ public class HUHandlingUnitsInfoFactory implements IHandlingUnitsInfoFactory
 		final IHandlingUnitsBL handlingUnitsBL = Services.get(IHandlingUnitsBL.class);
 
 		//
-		// Primary: resolve from actual HU assignments
+		// Primary: resolve from actual HU assignments.
+		// Use a plain query (not retrieveTopLevelHUAssignmentsForModel) because we want ALL assignments,
+		// including derived ones that have M_TU_HU_ID set — those let us resolve the TU PI directly.
 		final int adTableId = InterfaceWrapperHelper.getModelTableId(inoutLine);
 		final int recordId = InterfaceWrapperHelper.getId(inoutLine);
 		final List<I_M_HU_Assignment> huAssignments = Services.get(IQueryBL.class)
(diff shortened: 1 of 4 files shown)


metas-ts (commented):
we can't just return `null`. it would just push the problem downstream, to fail later

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/materialtracking/spi/impl/HUHandlingUnitsInfoFactory.java
```diff
@@ -79,20 +83,125 @@ else if (InterfaceWrapperHelper.isInstanceOf(model, I_PP_Order_BOMLine.class))
...
+	@NonNull
+	private I_M_HU_PI resolveTU_HU_PI(final I_M_InOutLine inoutLine)
+	{
+		final IHandlingUnitsBL handlingUnitsBL = Services.get(IHandlingUnitsBL.class);
+
+		//
+		// Primary: resolve from actual HU assignments.
+		// Use a plain query (not retrieveTopLevelHUAssignmentsForModel) because we want ALL assignments,
+		// including derived ones that have M_TU_HU_ID set — those let us resolve the TU PI directly.
+		final int adTableId = InterfaceWrapperHelper.getModelTableId(inoutLine);
```
metas-ts: use TableRecordReference instead

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/materialtracking/spi/impl/HUHandlingUnitsInfoFactory.java
```diff
@@ -79,20 +83,125 @@ else if (InterfaceWrapperHelper.isInstanceOf(model, I_PP_Order_BOMLine.class))
...
+		final int recordId = InterfaceWrapperHelper.getId(inoutLine);
+		final List<I_M_HU_Assignment> huAssignments = Services.get(IQueryBL.class)
+				.createQueryBuilder(I_M_HU_Assignment.class, inoutLine)
+				.addEqualsFilter(I_M_HU_Assignment.COLUMNNAME_AD_Table_ID, adTableId)
```
metas-ts: I'm sure IHUAssignmentBL has something for us! and if not, please add it after x-checking new_dawn_uat

metas-ts: Done — using `IHUAssignmentDAO.retrieveTUHUAssignmentsForModelQuery` (Pass 1) and `retrieveTopLevelHUAssignmentsForModel` (Pass 2). Both methods already existed on this branch.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/materialtracking/spi/impl/HUHandlingUnitsInfoFactory.java
metas-ts: Done — the raw `adTableId`/`recordId` resolution is gone. The DAO methods handle model-to-reference conversion internally.

PR #23401, opened 2026-04-10
fix: return null when TU PI is missing in material tracking invoicing

metas-ts:
## Summary

- Fix `Assumption failure: tuPI not null` crash in `HUHandlingUnitsInfoFactory.createFromInOutLine()` when an InOutLine has no TU HU_PI assigned
- The method now returns `null` (which callers already handle) instead of passing `null` to the `HUHandlingUnitsInfo` constructor which crashes
- Same fix as https://github.com/metasfresh/metasfresh/pull/23400 (targeting a customer hotfix branch)

## Test plan

- [ ] CI green
- [ ] Verify that manufacturing orders with InOutLines missing TU PI no longer crash during invoice candidate creation


metas-ts:
Closing — the fix will reach new_dawn_uat via the normal merge chain from soft_panda_hotfix.

PR #23413, opened 2026-04-15
Fix: schedule ASI always propagates to shipment line

metas-ts:
## Summary
- Shipment schedule ASI (from order line) now always propagates to shipment line, bypassing the `M_ShipmentSchedule_AttributeConfig` whitelist
- HU attributes still filtered through whitelist; schedule ASI takes precedence, HU values fill gaps for empty schedule ASI values
- Extracted `mergeAttributeValues()` as testable pure function with 8 unit tests
- Renamed misleading "HU-Aggregation" tab to "Shipment Line Attribute Config" (AD_Tab_ID=541052)
- Added window 540150 ("Lieferkandidaten - Handler") to WebUI menu tree

## Test plan
- [x] Unit tests: `ShipmentScheduleWithHUTests.MergeAttributeValuesTests` — 8 scenarios for merge logic
- [x] Cucumber: `shipmentLineASIPropagation.feature` — dropship, precedence, fallback (in progress)
- [ ] Verify tab rename in WebUI
- [ ] Verify window accessible in WebUI menu

### DIFF
--- a/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798070_sys_me03_29295_rename_HU_Aggregation_tab.sql
+++ b/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798070_sys_me03_29295_rename_HU_Aggregation_tab.sql
@@ -1,0 +1,13 @@
+-- me03#29295: Rename misleading "HU-Aggregation" tab to "Shipment Line Attribute Config"
+-- AD_Tab_ID=541052 in window 540150 (Lieferkandidaten - Handler)
+
+-- Update base record (de_DE is the base language)
+UPDATE AD_Tab SET Name = 'Lieferzeile Merkmalkonfiguration', Updated = '2026-04-15 14:00', UpdatedBy = 99
+WHERE AD_Tab_ID = 541052;
+
+-- Update translations
+UPDATE AD_Tab_Trl SET Name = 'Shipment Line Attribute Config', IsTranslated = 'Y', Updated = '2026-04-15 14:00', UpdatedBy = 99
+WHERE AD_Tab_ID = 541052 AND AD_Language = 'en_US';
+
+UPDATE AD_Tab_Trl SET Name = 'Lieferzeile Merkmalkonfiguration', Updated = '2026-04-15 14:00', UpdatedBy = 99
+WHERE AD_Tab_ID = 541052 AND AD_Language IN ('de_DE', 'de_CH');
--- a/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798080_sys_me03_29295_add_window_to_webui_menu.sql
+++ b/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798080_sys_me03_29295_add_window_to_webui_menu.sql
@@ -1,15 +1,15 @@
--- me03#29295: Add window "Lieferkandidaten - Handler" (AD_Menu_ID=540383) to the
--- "Lieferung > Einstellungen" folder in the menu tree (ad_tree_id=10).
--- Parent node 1000079 = "Einstellungen" folder under "Lieferung" (node 1000019).
+-- me03#29295: Move window "Lieferkandidaten - Handler" (AD_Menu_ID=540383) to
+-- "Lieferung > Einstellungen" in the menu tree (ad_tree_id=10).
+-- The node already exists in tree 10 under "Admin" (parent 540382) — move it to
+-- "Einstellungen" (parent 1000079) under the "Lieferung" folder (node 1000019).
 
--- Remove wrong placement (if applied from earlier version of this script)
+-- Remove wrong placement from tree 1000019 (if applied from earlier version of this script)
 DELETE FROM AD_TreeNodeMM WHERE AD_Tree_ID = 1000019 AND Node_ID = 540383;
 
--- Insert into the correct location: Lieferung > Einstellungen
-INSERT INTO AD_TreeNodeMM (AD_Client_ID, AD_Org_ID, AD_Tree_ID, Node_ID, Parent_ID, SeqNo, Created, CreatedBy, Updated, UpdatedBy, IsActive)
-SELECT 0, 0, 10, 540383, 1000079,
-       COALESCE((SELECT MAX(SeqNo) FROM AD_TreeNodeMM WHERE AD_Tree_ID = 10 AND Parent_ID = 1000079), 0) + 1,
-       '2026-04-15 14:00', 99, '2026-04-15 14:00', 99, 'Y'
-WHERE NOT EXISTS (
-    SELECT 1 FROM AD_TreeNodeMM WHERE AD_Tree_ID = 10 AND Node_ID = 540383 AND Parent_ID = 1000079
-);
+-- Move existing node from old parent (540382 = Admin) to new parent (1000079 = Einstellungen)
+UPDATE AD_TreeNodeMM
+SET Parent_ID = 1000079,
+    SeqNo = COALESCE((SELECT MAX(SeqNo) FROM AD_TreeNodeMM WHERE AD_Tree_ID = 10 AND Parent_ID = 1000079), 0) + 1,
+    Updated = '2026-04-15 14:00',
+    UpdatedBy = 99
+WHERE AD_Tree_ID = 10 AND Node_ID = 540383;
--- a/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798080_sys_me03_29295_add_window_to_webui_menu.sql
+++ b/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798080_sys_me03_29295_add_window_to_webui_menu.sql
@@ -1,0 +1,10 @@
+-- me03#29295: Add window "Lieferkandidaten - Handler" (AD_Menu_ID=540383) to WebUI menu tree
+-- In the old tree (ad_tree_id=10), path is: Verkauf > Lieferscheine > Admin > Lieferkandidaten - Handler
+-- In the WebUI tree (ad_tree_id=1000019), "Lieferscheine" and "Admin" folders don't exist.
+-- Place under "Einstellungen Verkauf" (node 1000001) which is the sales admin folder under "Verkauf".
+
+INSERT INTO AD_TreeNodeMM (AD_Client_ID, AD_Org_ID, AD_Tree_ID, Node_ID, Parent_ID, SeqNo, Created, CreatedBy, Updated, UpdatedBy, IsActive)
+SELECT 0, 0, 1000019, 540383, 1000001, 8, '2026-04-15 14:00', 99, '2026-04-15 14:00', 99, 'Y'
+WHERE NOT EXISTS (
+    SELECT 1 FROM AD_TreeNodeMM WHERE AD_Tree_ID = 1000019 AND Node_ID = 540383
+);
--- a/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798080_sys_me03_29295_add_window_to_webui_menu.sql
+++ b/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798080_sys_me03_29295_add_window_to_webui_menu.sql
@@ -1,10 +1,15 @@
--- me03#29295: Add window "Lieferkandidaten - Handler" (AD_Menu_ID=540383) to WebUI menu tree
--- In the old tree (ad_tree_id=10), path is: Verkauf > Lieferscheine > Admin > Lieferkandidaten - Handler
--- In the WebUI tree (ad_tree_id=1000019), "Lieferscheine" and "Admin" folders don't exist.
--- Place under "Einstellungen Verkauf" (node 1000001) which is the sales admin folder under "Verkauf".
+-- me03#29295: Add window "Lieferkandidaten - Handler" (AD_Menu_ID=540383) to the
+-- "Lieferung > Einstellungen" folder in the menu tree (ad_tree_id=10).
+-- Parent node 1000079 = "Einstellungen" folder under "Lieferung" (node 1000019).
 
+-- Remove wrong placement (if applied from earlier version of this script)
+DELETE FROM AD_TreeNodeMM WHERE AD_Tree_ID = 1000019 AND Node_ID = 540383;
+
+-- Insert into the correct location: Lieferung > Einstellungen
 INSERT INTO AD_TreeNodeMM (AD_Client_ID, AD_Org_ID, AD_Tree_ID, Node_ID, Parent_ID, SeqNo, Created, CreatedBy, Updated, UpdatedBy, IsActive)
-SELECT 0, 0, 1000019, 540383, 1000001, 8, '2026-04-15 14:00', 99, '2026-04-15 14:00', 99, 'Y'
+SELECT 0, 0, 10, 540383, 1000079,
+       COALESCE((SELECT MAX(SeqNo) FROM AD_TreeNodeMM WHERE AD_Tree_ID = 10 AND Parent_ID = 1000079), 0) + 1,
+       '2026-04-15 14:00', 99, '2026-04-15 14:00', 99, 'Y'
 WHERE NOT EXISTS (
-    SELECT 1 FROM AD_TreeNodeMM WHERE AD_Tree_ID = 1000019 AND Node_ID = 540383
+    SELECT 1 FROM AD_TreeNodeMM WHERE AD_Tree_ID = 10 AND Node_ID = 540383 AND Parent_ID = 1000079
 );
--- a/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798230_sys_me03_29295_IsHUAttributeOverridesASI.sql
+++ b/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798230_sys_me03_29295_IsHUAttributeOverridesASI.sql
@@ -1,4 +1,5 @@
 -- me03#29295: Add IsHUAttributeOverridesASI to M_ShipmentSchedule_AttributeConfig
+-- Part 1: AD metadata only (no DDL on the target table — see 5798290 for DDL)
 --
 -- Controls whether the HU attribute value overwrites the order line's ASI value
 -- on the shipment line. Default Y = backward compatible (HU wins, as before).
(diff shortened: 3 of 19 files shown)


backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798070_sys_me03_29295_rename_HU_Aggregation_tab.sql
```diff
@@ -0,0 +1,13 @@
+-- me03#29295: Rename misleading "HU-Aggregation" tab to "Shipment Line Attribute Config"
+-- AD_Tab_ID=541052 in window 540150 (Lieferkandidaten - Handler)
+
+-- Update base record (de_DE is the base language)
+UPDATE AD_Tab SET Name = 'Lieferzeile Merkmalkonfiguration', Updated = '2026-04-15 14:00', UpdatedBy = 99
```
utafrali, line 5: Hardcoded timestamp `'2026-04-15 14:00'` will be stale the moment this migration runs on any environment after that date. Use `NOW()` instead, consistent with standard metasfresh migration practice:

```sql
UPDATE AD_Tab SET Name = 'Lieferzeile Merkmalkonfiguration', Updated = NOW(), UpdatedBy = 99
```
Same applies to lines 9 and 13.

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798080_sys_me03_29295_add_window_to_webui_menu.sql
```diff
@@ -0,0 +1,10 @@
+-- me03#29295: Add window "Lieferkandidaten - Handler" (AD_Menu_ID=540383) to WebUI menu tree
+-- In the old tree (ad_tree_id=10), path is: Verkauf > Lieferscheine > Admin > Lieferkandidaten - Handler
+-- In the WebUI tree (ad_tree_id=1000019), "Lieferscheine" and "Admin" folders don't exist.
+-- Place under "Einstellungen Verkauf" (node 1000001) which is the sales admin folder under "Verkauf".
+
+INSERT INTO AD_TreeNodeMM (AD_Client_ID, AD_Org_ID, AD_Tree_ID, Node_ID, Parent_ID, SeqNo, Created, CreatedBy, Updated, UpdatedBy, IsActive)
+SELECT 0, 0, 1000019, 540383, 1000001, 8, '2026-04-15 14:00', 99, '2026-04-15 14:00', 99, 'Y'
```
utafrali: Two concerns here:

1. Same hardcoded timestamp issue — use `NOW()`.
2. `SeqNo = 8` is hard-coded without checking what values already exist under `Parent_ID = 1000001`. If another node already occupies `SeqNo = 8`, the ordering will be undefined. Either query `COALESCE(MAX(SeqNo), 0) + 10` or document explicitly which existing node has been verified to not occupy that slot:

```sql
INSERT INTO AD_TreeNodeMM (..., SeqNo, ...)
SELECT 0, 0, 1000019, 540383, 1000001,
       COALESCE((SELECT MAX(SeqNo) FROM AD_TreeNodeMM WHERE AD_Tree_ID = 1000019 AND Parent_ID = 1000001), 0) + 10,
       NOW(), 99, NOW(), 99, 'Y'
WHERE NOT EXISTS (...);
```

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/ShipmentScheduleWithHU.java
```diff
@@ -289,29 +293,85 @@ private List<IAttributeValue> computeAttributeValues()
...
 			final I_M_AttributeSetInstance attributeSetInstance = load(getM_AttributeSetInstance_ID(), I_M_AttributeSetInstance.class);
 			final IAttributeStorage asiAttributeStorage = ASIAttributeStorage.createNew(attributeStorageFactory, attributeSetInstance);
-			allAttributeValues.addAll(asiAttributeStorage.getAttributeValues());
+			schedAsiAttributeValues.addAll(asiAttributeStorage.getAttributeValues());
 
-			// additionally add whatever the attributeStorageFactory's storage implementation has to offer.
 			final IAttributeStorage huAsiAttributeStorage = attributeStorageFactory.getAttributeStorage(attributeSetInstance);
-			allAttributeValues.addAll(huAsiAttributeStorage.getAttributeValues());
+			schedAsiAttributeValues.addAll(huAsiAttributeStorage.getAttributeValues());
 		}
```
utafrali, line 328: The `mergeAttributeValues` method re-checks `if (!schedAttr.isUseInASI()) continue` internally, but the caller already filters `filteredSchedAsiAttributes` by `.filter(IAttributeValue::isUseInASI)` before passing it in. This redundant guard implies the method contract is unclear. Either remove the internal guard and document that the parameter must be pre-filtered, or remove the caller-side filter and let the method own that responsibility. Mixing both creates confusion about where the invariant is enforced.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/ShipmentScheduleWithHU.java
```diff
@@ -289,29 +293,85 @@ private List<IAttributeValue> computeAttributeValues()
...
+		// 2. Collect schedule ASI attributes (NO handler whitelist filter — always pass through)
+		final TreeSet<IAttributeValue> schedAsiAttributeValues = new TreeSet<>(Comparator.comparing(av -> av.getM_Attribute().getM_Attribute_ID()));
+
 		if (getM_AttributeSetInstance_ID() > 0)
 		{
-			/// add all values from the ASI
 			final I_M_AttributeSetInstance attributeSetInstance = load(getM_AttributeSetInstance_ID(), I_M_AttributeSetInstance.class);
 			final IAttributeStorage asiAttributeStorage = ASIAttributeStorage.createNew(attributeStorageFactory, attributeSetInstance);
-			allAttributeValues.addAll(asiAttributeStorage.getAttributeValues());
+			schedAsiAttributeValues.addAll(asiAttributeStorage.getAttributeValues());
```
utafrali, line 323: The `mergeAttributeValues` method appears to have default (package-private) access. The test class is in package `de.metas.handlingunits.shipmentschedule.api.impl`, which is a *different* package from `de.metas.handlingunits.shipmentschedule.api` — Java package-private access does not extend to sub-packages. If the method is truly package-private, the test file will not compile. The method should be explicitly `public` (acceptable here since `@VisibleForTesting` signals the intent) or the test should be moved to the same package. Please verify the actual access modifier and confirm the tests build cleanly.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/ShipmentScheduleWithHU.java
```diff
@@ -279,8 +282,9 @@ private Object computeAttributesAggregationKey()
...
 	private List<IAttributeValue> computeAttributeValues()
 	{
-		final TreeSet<IAttributeValue> allAttributeValues = //
-				new TreeSet<>(Comparator.comparing(av -> av.getM_Attribute().getM_Attribute_ID()));
+		//
+		// 1. Collect HU attributes (filtered by handler whitelist)
+		final TreeSet<IAttributeValue> huAttributeValues = new TreeSet<>(Comparator.comparing(av -> av.getM_Attribute().getM_Attribute_ID()));
 
 		final IAttributeStorageFactory attributeStorageFactory = huContext.getHUAttributeStorageFactory();
 		streamHUHierarchyBottomUp().forEach(hu -> {
```
utafrali, line 293: The `filteredSchedAsiAttributes` stream does not filter out empty values before passing them to `mergeAttributeValues` (unlike `filteredHUAttributes`, which does filter at the stream step). The merge method then re-checks emptiness via `Objects.equals(schedAttr.getValue(), schedAttr.getEmptyValue())`. This works correctly for standard attribute types, but if any `IAttributeValue` implementation returns `null` from `getEmptyValue()` and a non-null-but-semantically-empty value from `getValue()`, the `Objects.equals` check would not detect it as empty, and the schedule ASI value would incorrectly override the HU value. A comment explaining why the empty pre-filter is intentionally omitted here (and why the merge-level check is sufficient) would prevent future maintainers from introducing a bug by "fixing" the apparent inconsistency.

utafrali (commented):
The core logic change is well-reasoned and the extracted `mergeAttributeValues` function is clean and properly tested. However, hardcoded timestamps in both SQL migrations need to be replaced with `NOW()`, the `SeqNo` insertion needs a dynamic max-query to avoid ordering conflicts, and the access modifier on `mergeAttributeValues` must be confirmed to allow the test in the `impl` sub-package to compile.

metas-ts:
@claude review opus

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798230_sys_me03_29295_IsHUAttributeOverridesASI.sql
```diff
@@ -0,0 +1,130 @@
...
+-- take precedence over what was physically picked (e.g., Herkunft/Positions Nr).
+
+-- =============================================================================
+-- 1. AD_Element
+-- =============================================================================
+INSERT INTO AD_Element (AD_Element_ID, AD_Client_ID, AD_Org_ID, IsActive, Created, CreatedBy, Updated, UpdatedBy,
+                        ColumnName, Name, PrintName, Description, Help)
+VALUES (584759 /*From ID Server*/, 0, 0, 'Y', '2026-04-16 18:00', 0, '2026-04-16 18:00', 0,
+        'IsHUAttributeOverridesASI',
+        'ME-Merkmal überschreibt Merkmalsatz',
```
metas-ts: IDK what ME means. from ckentext it might be "HU"? if so, please write HU instead of ME.
we use it also in `de_*`

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798230_sys_me03_29295_IsHUAttributeOverridesASI.sql
```diff
@@ -0,0 +1,130 @@
...
+INSERT INTO AD_Column_Trl (AD_Language, AD_Column_ID, Name, IsTranslated, AD_Client_ID, AD_Org_ID, Created, CreatedBy, Updated, UpdatedBy, IsActive)
+SELECT l.AD_Language, t.AD_Column_ID, t.Name, 'N', t.AD_Client_ID, t.AD_Org_ID, t.Created, t.CreatedBy, t.Updated, t.UpdatedBy, 'Y'
+FROM AD_Language l, AD_Column t
+WHERE l.IsActive = 'Y' AND l.IsSystemLanguage = 'Y' AND t.AD_Column_ID = 592363
+  AND NOT EXISTS (SELECT 1 FROM AD_Column_Trl tt WHERE tt.AD_Language = l.AD_Language AND tt.AD_Column_ID = t.AD_Column_ID);
+
+-- =============================================================================
+-- 3. Physical column (new column, use ALTER TABLE ADD COLUMN)
+-- =============================================================================
+ALTER TABLE M_ShipmentSchedule_AttributeConfig ADD COLUMN IF NOT EXISTS IsHUAttributeOverridesASI CHAR(1) DEFAULT 'Y';
```
metas-ts: this might result in a "pendingg trigger event" error. put DDL and DML (in the same column) into different migration-files

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798290_sys_me03_29295_IsHUAttributeOverridesASI_DDL.sql
```diff
@@ -0,0 +1,6 @@
+-- me03#29295: Add IsHUAttributeOverridesASI physical column
+-- Part 2: DDL + DML (separate from AD metadata to avoid pending trigger events)
```
metas-ts: that's not what i meant. the DDL *and* the DML for the same column need to be separated *from each other*

metas-ts (commented):
we also need a test where  `M_ShipmentSchedule_AttributeConfig.IsHUAttributeOverridesASI` is yes..obviously

metas-ts:
@claude review opus

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798290_sys_me03_29295_IsHUAttributeOverridesASI_DDL.sql
metas-ts: Fixed in a2df155: split into 5798290 (DDL only: `ALTER TABLE ADD COLUMN`) and 5798320 (DML: `UPDATE` + `SET NOT NULL`).

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5798320_sys_me03_29295_IsHUAttributeOverridesASI_DML.sql
```diff
@@ -0,0 +1,4 @@
+-- me03#29295: Populate IsHUAttributeOverridesASI and set NOT NULL (DML, separate from DDL)
+
+UPDATE M_ShipmentSchedule_AttributeConfig SET IsHUAttributeOverridesASI = 'Y' WHERE IsHUAttributeOverridesASI IS NULL;
+ALTER TABLE M_ShipmentSchedule_AttributeConfig ALTER COLUMN IsHUAttributeOverridesASI SET NOT NULL;
```
metas-ts: isn'T the 1st line DML and the 2nd line DDL?

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipment/shipmentLineASIPropagation.feature
```diff
@@ -0,0 +1,266 @@
+@from:cucumber
+@ghActions:run_on_executor3
+Feature: Shipment line ASI propagation
+  The shipment schedule's ASI (from the order line) must always propagate to the shipment line.
+  The M_ShipmentSchedule_AttributeConfig whitelist only filters HU attributes — not the schedule's own ASI.
+  When HU attributes and schedule ASI overlap, the schedule ASI value takes precedence unless it is empty.
```
metas-ts: this is not up2date anymore - right?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/spi/ShipmentScheduleHandler.java
```diff
@@ -207,6 +217,20 @@ public final boolean attributeShallBePartOfShipmentLine(
 		return hasNonNullAttributeListValue(attributeInstance);
 	}
 
+	/**
+	 * Returns {@code true} if the HU attribute value should overwrite the schedule ASI value
+	 * for the given attribute on the shipment line. Defaults to {@code true} (backward compatible)
+	 * if no config record is found.
+	 */
+	public final boolean isHUAttributeOverridesASI(
+			@NonNull final I_M_ShipmentSchedule shipmentSchedule,
```
metas-ts: better go with OrgId if the schedule record doesn't really matter

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/spi/impl/ShipmentLineBuilder.java
```diff
@@ -632,6 +634,27 @@ private void transferAttributesToShipmentLine(@NonNull final I_M_InOutLine shipm
 		});
 	}
 
+	/**
+	 * Filters out attributes where {@code IsHUAttributeOverridesASI=N} in {@link I_M_ShipmentSchedule_AttributeConfig}.
+	 * For those attributes, the schedule ASI value (already set on the shipment line by
+	 * {@link ShipmentScheduleWithHU#computeAttributeValues()}) must not be overwritten by HU attribute transfer.
+	 */
+	private Collection<I_M_Attribute> filterAttributesForHUTransfer(@NonNull final Collection<I_M_Attribute> allAttributes)
```
metas-ts, line 644: please look if you can avoid raw model-classes like I_M_Attribute or I_M_ShipmentSchedule

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/ShipmentScheduleWithHU.java
```diff
@@ -289,31 +295,118 @@ private List<IAttributeValue> computeAttributeValues()
...
+
+		final ImmutableList<IAttributeValue> filteredHUAttributes = huAttributeValues.stream()
+				.filter(IAttributeValue::isUseInASI)
+				.filter(attributeValue -> !Objects.equals(attributeValue.getValue(), attributeValue.getEmptyValue()))
+				.filter(attributeValue -> handler.attributeShallBePartOfShipmentLine(shipmentSchedule, attributeValue.getM_Attribute()))
+				.collect(ImmutableList.toImmutableList());
+
+		//
+		// 2. Collect schedule ASI attributes (NO handler whitelist filter — always pass through)
+		final TreeSet<IAttributeValue> schedAsiAttributeValues = new TreeSet<>(Comparator.comparing(av -> av.getM_Attribute().getM_Attribute_ID()));
```
metas-ts: as i wrote:

> please look if you can avoid raw model-classes like I_M_Attribute or I_M_ShipmentSchedule


i think here you can use `av.getAttributeId()`
**Critically check every use of a model-class: is it really needed?**

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/ShipmentScheduleWithHU.java
```diff
@@ -289,31 +295,118 @@ private List<IAttributeValue> computeAttributeValues()
...
 				.collect(ImmutableList.toImmutableList());
-		return result;
+
+		//
+		// 3. Merge: HU attributes first, then overlay schedule ASI attributes.
+		// For each attribute, the handler's IsHUAttributeOverridesASI flag decides precedence:
+		// - Y (default): HU value wins (existing behavior for LotNumber, BestBefore, etc.)
+		// - N: schedule ASI value wins (for customer-intent attributes like Herkunft)
+		return mergeAttributeValues(filteredHUAttributes, filteredSchedAsiAttributes,
+				attribute -> handler.isHUAttributeOverridesASI(shipmentSchedule.getAD_Org_ID(), attribute));
```
metas-ts: > please look if you can avoid raw model-classes like I_M_Attribute or I_M_ShipmentSchedule

here that means: "please use `OrgId`"

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/ShipmentScheduleWithHU.java
```diff
@@ -289,31 +295,118 @@ private List<IAttributeValue> computeAttributeValues()
...
+			@NonNull final List<IAttributeValue> filteredHUAttributes,
+			@NonNull final List<IAttributeValue> schedAsiAttributes,
+			@NonNull final Predicate<I_M_Attribute> isHUOverridesASI)
+	{
+		final Map<Integer, IAttributeValue> merged = new LinkedHashMap<>();
+
+		// Start with HU attributes
+		for (final IAttributeValue huAttr : filteredHUAttributes)
+		{
+			merged.put(huAttr.getM_Attribute().getM_Attribute_ID(), huAttr);
```
metas-ts: here again - use getAttributeId()

metas-ts (commented):
prefere RepoIdAwares over integers when working with IDs

PR #23412, opened 2026-04-15
Add reservation guard logic to `HUTransformService`, including recursive checks, new tests, and AD_Message for reserved HUs

adi-stefan:

PR #23418, opened 2026-04-16
Fix: shipment creation fails with negative HU item storage qty

metas-ts:
## Summary
Fix shipment creation failure when multiple shipment schedules for the same product are processed in one workpackage. Three coordinated fixes plus safety net and tests.

## Root Cause
On-the-fly HU picking in `ShipmentScheduleWithHUService.pickHUsOnTheFly` had three independent bugs:
1. Each schedule maintained its own local `alreadyUsedSourceHuIds` set — so a second schedule would re-discover and double-process HUs already consumed by a prior schedule (the storage query reads out-of-transaction and doesn't see cached changes).
2. `createNewlyPickedHUs` used `ReservedHUsPolicy.CONSIDER_ALL`, ignoring reservations — a schedule could split from VHUs reserved for OTHER orders.
3. All three picking passes added bulk candidates to `alreadyUsedSourceHuIds` before `processHU`, but `processHU` only consumes what it needs. This caused later schedules to see ALL candidates as used, even ones never touched.

These combined to produce crashes like "Attempt to set negative qty on storageLine" and "Included HU not found in cached included HUs list".

### 1. Share `alreadyUsedSourceHuIds` across schedules
`ShipmentScheduleWithHUService.prepareShipmentSchedulesWithHU` now creates the set once in the outer schedule loop and threads it through `PrepareForSingleShipmentScheduleRequest` into each schedule's `pickHUsOnTheFly`. Each source HU is added to the shared set **inside `processHU`** as it's actually consumed (not in bulk upfront).

### 2. Respect HU reservations in on-the-fly picking
`createNewlyPickedHUs` now computes `allowedReservedVhuIds` (VHUs reserved for the current schedule's order line and/or project) and uses `ReservedHUsPolicy.onlyNotReservedExceptVhuIds(allowedReservedVhuIds)` instead of `CONSIDER_ALL`. The whitelist also feeds the recursive reservation guard added by #23412 via `HUTransformService.builder().allowedReservedVhuIds(...)`.

New service method `HUReservationService.getVHUIdsReservedByAnyOf(HUReservationDocRef...)` — generic over reservation document types so future refs (DD order line, picking job step) can be added without signature changes.

### 3. Safety net in `HUItemStorage.addQty`
If `qtyNew` is still negative after the existing sync correction, clamp to zero with ERROR-level log (via `Loggables.withLogger(logger, Level.ERROR)`) and propagate the **actual applied delta** (`qtyNew - qtyOld`) to `rollupIncremental` — not the requested `qtyConv`. This prevents any remaining inconsistency from cascading up the parent hierarchy.

## Tests
New cucumber feature `qtyReservation_shipment_multiSchedule.feature` with 3 scenarios (all single-order-two-lines to trigger the multi-schedule-in-one-workpackage path):
- **S_QtyRes_200**: Reserved HU respected — unreserved line processed first must skip the reserved VHU
- **S_QtyRes_210**: Both unreserved — no double-pick crash in single workpackage (direct regression test for me03#29333)
- **S_QtyRes_220**: Both unreserved, ASI-filtered — each schedule picks the correctly attributed HU

## Related issue
- https://github.com/metasfresh/me03/issues/29333

### DIFF
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/qtyReservation/qtyReservation_shipment_multiSchedule.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/qtyReservation/qtyReservation_shipment_multiSchedule.feature
@@ -129,13 +129,31 @@ Feature: Multi-schedule on-the-fly picking — no double-pick, respect reservati
       | M_ShipmentSchedule_ID | M_InOut_ID |
       | shipmentSchedule_1    | shipment   |
 
-    # Verify the right HU was picked to the right schedule:
-    # schedule_1 (reserved) must have picked hu_reserved as its TU (Herkunft=DE)
-    # schedule_2 (unreserved) must have picked hu_unreserved as its TU (Herkunft=AT)
+    # Verify both schedules were picked (10 PCE each, on-the-fly)
+    # Note: M_TU_HU_ID is null for unreserved picks (Pass 3 creates standalone VHUs)
+    # The HU routing is verified via the M_HU_Attribute on the picked VHU below
     And validate M_ShipmentSchedule_QtyPicked:
-      | M_ShipmentSchedule_ID | QtyPicked | IsAnonymousHuPickedOnTheFly | M_TU_HU_ID   |
-      | shipmentSchedule_1    | 10        | true                        | hu_reserved   |
-      | shipmentSchedule_2    | 10        | true                        | hu_unreserved |
+      | M_ShipmentSchedule_ID | QtyPicked | IsAnonymousHuPickedOnTheFly |
+      | shipmentSchedule_1    | 10        | true                        |
+      | shipmentSchedule_2    | 10        | true                        |
+
+    # Verify the right HU was picked to the right schedule via the picked VHU's attribute:
+    # Load VHU from each QtyPicked record, then check its Herkunft
+    And validate M_ShipmentSchedule_QtyPicked records for M_ShipmentSchedule identified by shipmentSchedule_1
+      | M_ShipmentSchedule_QtyPicked_ID | QtyPicked |
+      | qtyPicked_1                     | 10        |
+    And load M_HU as pickedVHU_1 from M_ShipmentSchedule_QtyPicked identified by qtyPicked_1
+    And M_HU_Attribute is validated
+      | M_HU_ID      | M_Attribute_ID.Value | Value |
+      | pickedVHU_1   | 1000001              | DE       |
+
+    And validate M_ShipmentSchedule_QtyPicked records for M_ShipmentSchedule identified by shipmentSchedule_2
+      | M_ShipmentSchedule_QtyPicked_ID | QtyPicked |
+      | qtyPicked_2                     | 10        |
+    And load M_HU as pickedVHU_2 from M_ShipmentSchedule_QtyPicked identified by qtyPicked_2
+    And M_HU_Attribute is validated
+      | M_HU_ID      | M_Attribute_ID.Value | Value |
+      | pickedVHU_2   | 1000001              | AT       |
 
 
   @from:cucumber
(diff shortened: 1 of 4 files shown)


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/storage/impl/HUItemStorage.java
```diff
@@ -174,23 +174,48 @@ public void addQty(@NonNull final ProductId productId, @NonNull final BigDecimal
...
+									 + "M_HU_Item_Id: {}, M_HU_Item_Storage.Qty: {}, M_HU_Item_Storage.Product: {}, M_HU_Item_Storage.UOM: {}, "
+									 + "M_HU_ID: {}, M_HU_Storage.Qty: {}, qtyToAdd: {} same UOM",
+							 storageLine.getM_HU_Item_ID(), storageLine.getQty(), storageLine.getM_Product_ID(), storageLine.getC_UOM_ID(),
+							 item.getM_HU_ID(), qtyOnParent, qtyConv);
 			Loggables.addLog("Warning! M_HU_Item_Storage.Qty out of sync with M_HU_Storage! "
-									 + "M_HU_Item_Id: {}, M_HU_Item_Storage.Qty: {}, M_HU_Item_Storage.Product: {}, M_HU_Item_Storage.UOM: {}"
+									 + "M_HU_Item_Id: {}, M_HU_Item_Storage.Qty: {}, M_HU_Item_Storage.Product: {}, M_HU_Item_Storage.UOM: {}, "
 									 + "M_HU_ID: {}, M_HU_Storage.Qty: {}, qtyToAdd: {} same UOM",
 							 storageLine.getM_HU_Item_ID(), storageLine.getQty(), storageLine.getM_Product_ID(), storageLine.getC_UOM_ID(),
 							 item.getM_HU_ID(), qtyOnParent, qtyConv);
```
metas-ts: use Loggables.withLogger

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/storage/impl/HUItemStorage.java
```diff
@@ -174,23 +174,48 @@ public void addQty(@NonNull final ProductId productId, @NonNull final BigDecimal
...
 
-		Check.errorIf(qtyNew.signum() < 0, "Attempt to set negative qty on storageLine; qtyOld={}; qtyToAdd={}; qtyNew={}; this={}; storageLine={}", qtyOld, qtyToAdd, qtyNew, this, storageLine);
+		// Clamp to zero if still negative after sync correction.
+		// This handles the case where the parent HU storage hierarchy is inconsistent
+		// (e.g. VHU has qty but all parent-level storages are zero).
+		// Without this, rollupRevert during HU reparenting would fail
+		// when detaching a VHU whose parent item storage was never properly rolled up.
+		if (qtyNew.signum() < 0)
+		{
+			logger.error("Clamping negative qty to zero on storageLine. "
```
metas-ts: use Loggables.withLogger

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/ShipmentScheduleWithHUService.java
```diff
@@ -242,6 +249,11 @@ public ImmutableList<ShipmentScheduleWithHU> prepareShipmentSchedulesWithHU(fina
 
 		final IHUContext huContext = huContextFactory.createMutableHUContext();
 
+		// Shared across all schedules to prevent the same HU being picked by multiple schedules.
```
metas-ts:
```suggestion
		// Shared across all PrepareForSingleShipmentScheduleRequest-instances to prevent the same HU being picked by multiple schedules.
```

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/reservation/HUReservationService.java
```diff
@@ -526,6 +527,34 @@ public ImmutableSet<HuId> getVHUIdsByDocumentRef(@NonNull final HUReservationDoc
 		return getByDocumentRef(documentRef).map(HUReservation::getVhuIds).orElseGet(ImmutableSet::of);
 	}
 
+	/**
+	 * Collect all VHU IDs reserved for the given order line and/or its project.
+	 * Used by on-the-fly picking to whitelist "my" reserved VHUs while skipping those reserved for others.
+	 */
+	@NonNull
+	public ImmutableSet<HuId> getVHUIdsReservedForOrderLineOrProject(
```
metas-ts: we have currently 4 types of reservable records. please come up with a method that is more generic like this one. otherwise we might end up with half a dozen of such methods

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/qtyReservation/qtyReservation_shipment_multiSchedule.feature
```diff
@@ -0,0 +1,337 @@
...
+
+
+  @from:cucumber
+  @Id:S_QtyRes_200
+  Scenario: Reserved HU respected — unreserved schedule processed first must skip reserved VHU
+  ## _Given 2 TUs of 10 PCE each
+  ## _And 2 sales orders for 10 PCE each
+  ## _And HU_A is reserved for order A
+  ## _When shipments are generated in one workpackage, with order B FIRST
+  ## _Then order B skips the reserved HU (reservation policy), order A gets it
```
metas-ts: this is not what i had in mind - for **every** scenario we need
- the test shall have one order with two lines
- the test shall verify that the right HU is on-the-fly-picked to the right shipment-schedule

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/ShipmentScheduleWithHUService.java
```diff
@@ -530,6 +555,12 @@ private Quantity processHU(
 			loggableWithLogger.addLog("pickHUsOnTheFly - QtyToDeliver={}; split Qty={} from available M_HU_ID={} with Qty={}",
 					qtyToDeliver, quantityToSplit, sourceHURecord.getM_HU_ID(), qtyOfSourceHU);
 
+			// Mark this source HU as consumed so other schedules in the same workpackage won't re-pick it
+			if (alreadyUsedSourceHuIds != null)
```
metas-ts: i'm not 100% sure about this code. is it OK to only add the HU if alreadyUsedSourceHuIds was null?

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/storage/impl/HUItemStorage.java
```diff
@@ -174,23 +175,40 @@ public void addQty(@NonNull final ProductId productId, @NonNull final BigDecimal
...
 
 		storageLine.setQty(qtyNew);
 		dao.save(storageLine);
 
 		//
-		// Roll-up
-		rollupIncremental(productId, qtyConv, uomStorage);
+		// Roll-up: propagate only the actual delta that was applied, not the requested delta.
+		// If qtyNew was clamped (e.g. from -8 to 0), the actual delta is smaller than qtyConv.
+		final BigDecimal actualDelta = qtyNew.subtract(qtyOld);
```
metas-ts, line 212: we used to call this with `qtyConv`. now we call it with `qtyNew.subtract(qtyOld);`
is that OK?

PR #23430, opened 2026-04-16
Shipper nShift - send additional data

mrie04:


mrie04:
@claude review opus

backend/de.metas.shipper.client.nshift/src/test/java/de/metas/shipper/client/nshift/NShiftShipAdvisorServiceTest.java
```diff
@@ -55,7 +55,8 @@ public class NShiftShipAdvisorServiceTest
 	private static final String ACTOR_ID = System.getProperty("nshift.test.actorId", "nShift portal actorId");
 	private static final String USERNAME = System.getProperty("nshift.test.username", "nShift portal username");
 	private static final String PASSWORD = System.getProperty("nshift.test.password", "nShift portal password");
-	private static final String SHIP_RULE_SERVICE_LEVEL = "Test";
+	private static final String URL = System.getProperty("nshift.test.url", "https://demo.shipmentserver.com:8080");
```
teosarca, line 58: hmm, will this URL actually be contacted? i think that might spam it, considering how many builds we have.
I suggest
* contactng the actual test server shall be done in manually run tests, ie dev is running it on demand and not automatically executed on each build. running it automatically might also introduce flakyness
* mock the demo server in CI runs

mrie04: here are 2 tests
1. only test request build and compare with snapshot
2. local test disabled on CI runs where the real server can be contacted for qucik tests with different data

In addition we have a cucumber where this is mocked and tested

backend/de.metas.shipper.client.nshift/src/main/java/de/metas/shipper/client/nshift/NShiftShipmentService.java
```diff
@@ -284,33 +285,46 @@ private static JsonDeliveryResponse buildJsonDeliveryResponse(@NonNull final Jso
...
+
+	@Nullable
+	private static byte[] extractLabel(@Nullable final JsonShipmentResponseLabel label) {
+		return (label == null || label.getContent() == null)
+				? null
+				: label.getContent().getBytes();
+	}
+
+	@Nullable
+	private static String extractTrackingUrl(@NonNull final JsonPackage pkg, @Nullable final String fallback) {
```
teosarca: tbh, this kind of methods i would prefer to be part of the JsonPackage POJO.. maybe annotated with JsonIgnore

PR #23437, opened 2026-04-16
Port C_Invoice.DueDate feature from adorable_badger

teosarca:
## Summary

Ports the `C_Invoice.DueDate` / `C_PaymentTerm.IsAllowOverrideDueDate` feature (originally from the 2023 `adorable_badger` line: gh14552, gh14569, gh14588) onto `new_dawn_uat`, plus a full consistency sweep so every invoice due-date consumer reads the persisted column. Iteration 1 of a larger split-payment initiative; follow-up iterations will land in separate PRs.

### Core feature
- **New column `C_Invoice.DueDate`** (timestamp, nullable) populated at invoice completion — authoritative due date persisted on the invoice row.
- **New column `C_PaymentTerm.IsAllowOverrideDueDate`** (bool, default N) — gates whether invoicing can override the payment-term-derived due date.
- **Generated models** updated: `I_C_Invoice` / `X_C_Invoice` (DueDate getter/setter), `I_C_PaymentTerm` / `X_C_PaymentTerm` (IsAllowOverrideDueDate).
- **Process `C_Invoice_OverrideDueDate`** — manual override for completed invoices.
- **DueDate propagation through invoice candidate → invoice flow**: `IInvoiceHeader`, `InvoiceHeaderImpl`(`Builder`), `AggregationEngine.computeOverrideDueDate`, `InvoicingParams`, `InvoiceCandBLCreateInvoices`, `AbstractInvoiceBL`. When the aggregator's `overrideDueDateParam` is provided and the payment term allows it, the override is set on the final invoice.
- **Dunning integration**: `C_Invoice` dunning validator + `InvoiceSourceBL` read DueDate from the invoice directly.
- **View `C_Invoice_v`**: the primary branch now reads `i.duedate` instead of calling `paymenttermduedate()`.

### Consistency sweep (second commit)

After the core change, every remaining callsite that computed due date on the fly now reads the persisted column, so an overridden due date is respected everywhere:
- **SQL views**: `RV_DATEV_Export_Fact_Acct_Invoice`, `C_Dunning_Candidate_Invoice_v1`
- **SQL functions**: `Docs_Sales_Dunning_Report_details`, `Docs_Sales_Invoice_Description`, `openitems_report`
- **Jasper reports (7 JRXMLs)**: dunning details (3 variants), openitems details (2 variants + tabular), open payments
- **Java**: `InvoiceSourceDAO.retrieveDueDate` returns `invoice.getDueDate()` directly

One intentional remaining caller of `paymenttermduedate()`: `Docs_Sales_Dunning_Report_details` keeps the function call as a fallback inside a `COALESCE(i.DueDate, paymenttermduedate(o.C_PaymentTerm_ID, o.DateOrdered))` for order-based dunning candidates (orders don't carry a persisted due date). Invoice-based candidates read `i.DueDate`.

### Migration scripts

10 total: 4 reused from adorable_badger (`5676789`, `5676790`, `5676920`, `5676927`), 1 renumbered (`5798270` — `C_Invoice_v` adjustment, had to move past `5795950` which introduced `isFinancial`), 5 new for the sweep (`5798330`–`5798370`).

## Explicitly out of scope

- Dropping the `paymenttermduedate()` SQL function. It's still used as a fallback for order-based dunning candidates; a cleanup PR can remove it later once that path is reconsidered.
- Proforma invoice allocation / LC payment automation / partial-invoice splitting — iterations 2 and 3 of the split-payment initiative, own tickets.

## Verification

- `mvn compile` clean on `de.metas.business`, `de.metas.adempiere.adempiere/base`, `de.metas.dunning`, `de.metas.swat.base`, `de.metas.datev`, `de.metas.fresh.base`.
- Migrations applied against local customer DB (port 21432, Docker container): all 10 new scripts applied, zero failures.
- Post-migration checks: `c_invoice_v` reads `i.duedate` in both UNION branches; `docs_sales_invoice_description` function body now reads `i.DueDate`; `rv_datev_export_fact_acct_invoice.duedate` and `c_dunning_candidate_invoice_v1.duedate` still present.
- Unit test `TestFixedDateInvoicedAndDateAcct` (9 tests, incl. 2 new covering the override-allowed flag in both directions) green locally.

## Test plan

- [ ] CI pipeline green (including cucumber + frontend webui)
- [ ] Post-merge: verify on a customer UAT environment that completed invoices populate `C_Invoice.DueDate`
- [ ] Post-merge: smoke-test the `C_Invoice_OverrideDueDate` process on a completed invoice whose payment term has `IsAllowOverrideDueDate='Y'`
- [ ] Post-merge: run Dunning and OpenItems reports; confirm due dates match the invoice's persisted `DueDate` (not the recomputed one)


backend/de.metas.dunning/src/main/java/de/metas/dunning/invoice/api/impl/InvoiceSourceBL.java
```diff
@@ -54,9 +54,8 @@ public boolean setDunningGraceIfManaged(@NonNull final I_C_Invoice invoice)
 			return false;
 		}
 
-		final InvoiceDueDateProviderService invoiceDueDateProviderService = SpringContextHolder.instance.getBean(InvoiceDueDateProviderService.class);
-
-		final LocalDate dueDate = invoiceDueDateProviderService.provideDueDateFor(InvoiceId.ofRepoId(invoice.getC_Invoice_ID()));
+		final ZoneId timeZone = orgDAO.getTimeZone(OrgId.ofRepoId(invoice.getAD_Org_ID()));
+		final LocalDate dueDate = TimeUtil.asLocalDate(invoice.getDueDate(), timeZone);
 		final LocalDate dunningGraceDate = dueDate.plusDays(dunning.getGraceDays());
```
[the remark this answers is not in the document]
teosarca: Fixed in commit https://github.com/metasfresh/metasfresh/commit/8b3123134be — `setDunningGraceIfManaged` now guards `dueDate == null` (logs a warning and returns `false` without setting `DunningGrace`).

**Port-relevance note**: you're right that this is a new regression introduced by the port. On `adorable_badger` the equivalent code path went through `InvoiceDueDateProviderService.provideDueDateFor(...)`, which recomputed from the payment term and never returned null. The port changed this to read `invoice.getDueDate()` directly, introducing the null path. The guard restores the safe behavior without reintroducing the provider indirection.

Not adding a unit test in this PR because the POJO test base can't easily mock the `IOrgDAO` / `IDunningDAO` wiring without broader fixture plumbing; the change is a 3-line null guard, reviewable on sight.

backend/de.metas.dunning/src/main/java/de/metas/dunning/invoice/model/validator/C_Invoice.java
```diff
@@ -57,6 +63,20 @@ public void setDunningGraceIfAutomatic(final I_C_Invoice invoice)
...
+	 */
+	@DocValidate(timings = { ModelValidator.TIMING_AFTER_PREPARE })
+	public void setDueDate(final I_C_Invoice invoice)
+	{
+		if (invoice.getDueDate() == null)
+		{
+			final LocalDate dueDate = invoiceDueDateProviderService.provideDueDateFor(InvoiceId.ofRepoId(invoice.getC_Invoice_ID()));
+			invoice.setDueDate(TimeUtil.asTimestamp(dueDate));
+			InterfaceWrapperHelper.save(invoice);
+		}
```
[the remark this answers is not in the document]
teosarca: Addressed implicitly by the NPE guard in https://github.com/metasfresh/metasfresh/commit/8b3123134be . If `setDunningGraceIfAutomatic` fires before `setDueDate` in the model validator framework, `invoice.getDueDate()` is still null — but now `setDunningGraceIfManaged` returns safely without setting `DunningGrace` instead of NPE-ing. The dunning validator's `setDueDate` runs immediately after (same @DocValidate timing, same class), populates `C_Invoice.DueDate`, and the grace date is computed on the next save cycle (or when the override process next triggers it).

Combining the two into a single method (as suggested) would be a cleaner refactor but it's adorable-badger-inherited structure — leaving as-is here to keep the port diff minimal. If the dunning grace semantics become load-bearing on the same completion transaction, worth revisiting in a dedicated issue.

backend/de.metas.dunning/src/main/java/de/metas/dunning/invoice/model/validator/C_Invoice.java
```diff
@@ -57,6 +63,20 @@ public void setDunningGraceIfAutomatic(final I_C_Invoice invoice)
...
+	 * This shall set the Due Date in invoice considering payment term or contracts, but only if due date was not set previously
+	 */
+	@DocValidate(timings = { ModelValidator.TIMING_AFTER_PREPARE })
+	public void setDueDate(final I_C_Invoice invoice)
+	{
+		if (invoice.getDueDate() == null)
+		{
+			final LocalDate dueDate = invoiceDueDateProviderService.provideDueDateFor(InvoiceId.ofRepoId(invoice.getC_Invoice_ID()));
+			invoice.setDueDate(TimeUtil.asTimestamp(dueDate));
+			InterfaceWrapperHelper.save(invoice);
```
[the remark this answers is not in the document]
teosarca: Valid concern, and adorable-badger-inherited — the same `TimeUtil.asTimestamp(dueDate)` no-arg call exists on `adorable_badger`. Not fixing in this PR to keep the port diff aligned with the source branch.

Added to the follow-up tracker https://github.com/metasfresh/me03/issues/29412 (the issue that collects the DueDate-related design gaps carried over from adorable_badger: UI ReadOnlyLogic not tied to `IsAllowOverrideDueDate`, override process not gated on that flag, DunningGrace not recomputed after override, and now this timezone concern). That issue is the right home for a consistent sweep that uses the org timezone everywhere.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/AggregationEngine.java
```diff
@@ -591,6 +600,29 @@ private LocalDate computeDateAcct(@NonNull final I_C_Invoice_Candidate ic)
...
+	{
+		if (overrideDueDateParam == null)
+		{
+			return null;
+		}
+		final PaymentTermId paymentTermId = getC_PaymentTerm_ID(ic);
+		if (paymentTermId != null && !paymentTermRepository.isAllowOverrideDueDate(paymentTermId))
+		{
+			return null;
+		}
```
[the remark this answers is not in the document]
teosarca: Adorable-badger-inherited — the guard logic (`paymentTermId != null && !isAllowOverrideDueDate`) was ported verbatim. When the IC has neither `C_PaymentTerm_Override_ID` nor `C_PaymentTerm_ID` (payment term derived later from the BPartner), the flag check is bypassed. Fix requires resolving the effective payment term at candidate-aggregation time, which is a structural change.

Deferred to https://github.com/metasfresh/me03/issues/29412 , which already collects the 'enforce `IsAllowOverrideDueDate` consistently across all write paths' work (UI edit, override process, and now this IC path). That issue will be the single place to land a coherent fix.

backend/de.metas.business/src/main/java/de/metas/invoice/process/C_Invoice_OverrideDueDate.java
```diff
@@ -0,0 +1,99 @@
...
+		final IQueryUpdater<I_C_Invoice> queryUpdater = queryBL.createCompositeQueryUpdater(I_C_Invoice.class)
+				.addSetColumnValue(I_C_Invoice.COLUMNNAME_DueDate, p_OverrideDueDate);
+
+		final IQueryFilter<I_C_Invoice> filter = getProcessInfo().getQueryFilterOrElseFalse();
+		queryBL.createQueryBuilder(I_C_Invoice.class)
+				.addOnlyActiveRecordsFilter()
+				.addFilter(filter)
+				.create()
+				.update(queryUpdater);
+		return MSG_OK;
```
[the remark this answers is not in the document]
teosarca: Real gap, and adorable-badger-inherited — the override process is a bulk `IQueryUpdater` that skips model validators by design, so `DunningGrace` is left stale.

Captured in https://github.com/metasfresh/me03/issues/29412 as part of the override-process hardening (alongside 'precondition-check `IsAllowOverrideDueDate='Y'`'). The right fix is to either iterate over selected invoices (calling `IInvoiceSourceBL#setDunningGraceIfManaged`) or invoke the dunning event dispatcher after the bulk update — both are more invasive than the port's scope, so handling it in the follow-up issue.

backend/de.metas.dunning/src/main/jasperreports/de/metas/docs/sales/dunning/legacy/report_details.jrxml
```diff
@@ -25,7 +25,7 @@
 	doc.GrandTotal - doc.PaidAmt AS openamt,
 	dl.amt AS FeeAmt,
 	invoiceopen(dc.Record_ID, 0::numeric) + dl.amt as totalamt,
-	paymenttermduedate(doc.C_PaymentTerm_ID, doc.DocumentDate::timestamp with time zone) AS DueDate,
+	COALESCE(doc.InvoiceDueDate::timestamp with time zone, paymenttermduedate(doc.C_PaymentTerm_ID, doc.DocumentDate::timestamp with time zone)) AS DueDate,
```
teosarca, line 28: why are we also keeping the paymentTermDueDate as a fallback? the plan is to get rid of that function

backend/de.metas.dunning/src/main/java/de/metas/dunning/invoice/api/impl/InvoiceSourceDAO.java
```diff
@@ -49,7 +49,7 @@
 public class InvoiceSourceDAO implements IInvoiceSourceDAO
 {
 	@Override
-	public Timestamp retrieveDueDate(final org.compiere.model.I_C_Invoice invoice)
+	public Timestamp computeDueDateFromPaymentTerm(final org.compiere.model.I_C_Invoice invoice)
```
teosarca: so now, calling the paymentTermDueDate db function is the only source of true for due date computation?

backend/de.metas.dunning/src/main/java/de/metas/dunning/invoice/model/validator/C_Invoice.java
```diff
@@ -57,6 +63,20 @@ public void setDunningGraceIfAutomatic(final I_C_Invoice invoice)
 		InterfaceWrapperHelper.save(invoice);
 	}
 
+	/**
+	 * This shall set the Due Date in invoice considering payment term or contracts, but only if due date was not set previously
+	 */
+	@DocValidate(timings = { ModelValidator.TIMING_AFTER_PREPARE })
+	public void setDueDate(final I_C_Invoice invoice)
```
teosarca: this is no longer dunning related, but general invoice related

backend/de.metas.swat/de.metas.swat.base/src/test/java/de/metas/invoicecandidate/api/impl/aggregationEngine/TestFixedDateInvoicedAndDateAcct.java
```diff
@@ -258,4 +260,62 @@ public void test_dateAcctParam_takes_precedence_over_all()
...
+				.overrideDueDateParam(LocalDate.of(2023, Month.FEBRUARY, 1))
+				.build();
+
+		engine.addInvoiceCandidate(ic1);
+
+		final List<IInvoiceHeader> invoices = engine.aggregate();
+		assertThat(invoices).hasSize(1);
+
+		final IInvoiceHeader invoice = invoices.get(0);
+		assertThat(invoice.getOverrideDueDate()).isEqualTo(LocalDate.of(2023, Month.FEBRUARY, 1));
```
teosarca: always prefer LocalDate.parse("2023-02-01") instead of(y,m,d) because is much more readable

backend/de.metas.business/src/main/java/de/metas/payment/paymentterm/PaymentTermService.java
```diff
@@ -77,4 +83,23 @@ private void validateNow(@NonNull final Collection<PaymentTermId> paymentTermIds
 		paymentTermRepository.newLoaderAndSaver()
 				.syncStateToDatabase(ImmutableSet.copyOf(paymentTermIds));
 	}
+
+	public LocalDate computeDueDateFromPaymentTerm(@NonNull final PaymentTermId paymentTermId, @NonNull final LocalDate dateInvoiced)
+	{
+		final Timestamp dueDateTS = DB.getSQLValueTSEx(
+				ITrx.TRXNAME_ThreadInherited,
+				"SELECT paymentTermDueDate(?,?)",
```
teosarca, line 91: pls make sure paymentTermDueDate and computeDueDateFromPaymentTerm is the only source of true when computing the due date from payment term!

backend/de.metas.dunning/src/main/java/de/metas/dunning/invoice/model/validator/C_Invoice.java
teosarca: Good catch — addressed in your own commit https://github.com/metasfresh/metasfresh/commit/73a50ef178e which relocated `setDueDate` from the dunning module's `C_Invoice` validator to `de.metas.invoice.interceptor.C_Invoice` in `de.metas.business`. Dunning is no longer aware of this wiring — it just reads the populated column.

backend/de.metas.dunning/src/main/java/de/metas/dunning/invoice/api/impl/InvoiceSourceDAO.java
teosarca: Yes — audited on HEAD of the branch. Only two entry points compute DueDate from a payment term:

| Level | Location | Purpose |
|---|---|---|
| Java | `de.metas.payment.paymentterm.PaymentTermService.computeDueDateFromPaymentTerm` (PaymentTermService.java:87) | single Java SOT; internally calls `paymentTermDueDate` SQL |
| SQL | `paymentTermDueDate(C_PaymentTerm_ID, dateInvoiced)` DB function | the SOT function; called by `PaymentTermService` above and by reporting SQL that needs to compute DueDate for orders (where no persisted `C_Invoice.DueDate` exists) |

`grep -rn paymentTermDueDate backend/ --include="*.java"` returns exactly one hit after your refactor: `PaymentTermService:91`. Everywhere else in Java reads `invoice.getDueDate()` (the persisted column) directly.

Addressed by your refactor commit https://github.com/metasfresh/metasfresh/commit/6825e31bd86 which centralized the compute path.

backend/de.metas.swat/de.metas.swat.base/src/test/java/de/metas/invoicecandidate/api/impl/aggregationEngine/TestFixedDateInvoicedAndDateAcct.java
teosarca: Fixed in https://github.com/metasfresh/metasfresh/commit/42de141b311 — all three call sites in the two tests now use `LocalDate.parse("2023-02-01")`. 9/9 tests still green.

backend/de.metas.business/src/main/java/de/metas/payment/paymentterm/PaymentTermService.java
teosarca: Audited. Grep results on the branch head:

**Java (compute from payment term)**:
- `PaymentTermService.computeDueDateFromPaymentTerm` — single SOT. Calls `paymentTermDueDate` SQL internally.
- `PaymentTermBasedDueDateProvider` delegates to `PaymentTermService.computeDueDateFromPaymentTerm`.
- No other Java method computes DueDate from a payment term.

**SQL (direct `paymentTermDueDate` calls in active DDL)**:
- `de.metas.fresh.base/ddl/functions/Docs_Sales_Dunning_Report_details.sql:47` — order-based fallback inside the report's `COALESCE(i.DueDate, paymenttermduedate(o.C_PaymentTerm_ID, o.DateOrdered))`. Used when an order has no invoice yet. See also the follow-up comment on the legacy JRXML.
- Historical migration scripts retain the old compute expression in views they no longer own; this PR's sweep replaced all active DDL with `i.DueDate` reads, so those are inactive references only.

**JRXMLs still calling `paymenttermduedate`**:
- `de.metas.dunning/src/main/jasperreports/de/metas/docs/sales/dunning/legacy/report_details.jrxml:28` — same order-based fallback as above (see reply on the dedicated comment).

No ad-hoc compute sites found elsewhere (e.g. no `.plusDays(paymentTerm.getNetDays())` patterns that bypass the service).

backend/de.metas.dunning/src/main/jasperreports/de/metas/docs/sales/dunning/legacy/report_details.jrxml
teosarca: Two cases the report has to handle:
- **Invoice-based dunning** (the common case): `doc.InvoiceDueDate` = `C_Invoice.DueDate` → reads the new persisted column. No fallback needed.
- **Order-based dunning** (rare, happens for orders with payment terms but not yet invoiced): `C_Order` has no `DueDate` column, so something still has to compute from `(C_PaymentTerm_ID, DateOrdered)`.

The outer `COALESCE(doc.InvoiceDueDate, paymenttermduedate(...))` handles case 2. Since `paymentTermDueDate` is currently the SOT for that computation (both Java via `PaymentTermService.computeDueDateFromPaymentTerm` and SQL), using it here is consistent — not a second source of truth, the same one. When `paymentTermDueDate` is eventually retired, the real work will be: either add a persisted `DueDate` on `C_Order` (adorable_badger doesn't go there), or route the fallback through a view/function that wraps whatever replaces it.

This same fallback is already in `Docs_Sales_Dunning_Report_details.sql` at the function level (line 47). The legacy JRXML embeds its own query that predates that function — the cleanest long-term fix is to refactor the legacy JRXML to read from the function instead of duplicating the COALESCE. That's a non-trivial JRXML refactor and out of scope for the port; I can file it as a follow-up to me03#29412 if you want.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/AggregationEngine.java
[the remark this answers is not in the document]
Tom-Paye: ✅ **Fixed in commit e61cedd45.** Updated comment to clarify: total_battery_discharge is used for full load cycle calculations only. Autonomy uses PV-origin discharge (battery_self_consumption).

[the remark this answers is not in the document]
Tom-Paye: ✅ **Fixed in commit e61cedd45.** Updated comment to clarify: total_battery_discharge is used for full load cycle calculations only. Autonomy uses PV-origin discharge (battery_self_consumption).

PR #23441, opened 2026-04-17
Enhance `QtyReservationService` to update order lines with storage-relevant attributes based on reservation requests.

adi-stefan:


adi-stefan:
@claude review

PR #23496, opened 2026-04-20
Tax Accounting Report + §13b Reverse-Charge: symmetric KZ 84/85/67 reporting + regression coverage

teosarca:
## Summary

Reverse-Charge (§13b UStG) accounting output is now symmetric across the input and output legs — both UStVA KZ 67 (input deduction) and KZ 84/85 (output) carry the same signed NetAmt and TaxAmt after any §17(1) UStG payment-discount adjustment. Two `C_VAT_Code` rows per `C_Tax` (sales + purchase side) now drive per-leg VATCode resolution, so KZ 84/85 and KZ 67 are separately reportable. Ships 16 cucumber regression scenarios (TC-S1..TC-S16) locking in non-RC and RC output.

### Posting engine

- `FactLine` / `FactLineBuilder` — new `setVatCode(String)` direct setter and overloaded `setTaxIdAndUpdateVatCode(TaxId, boolean isSOTrxOverride)` so a Fact row can be forced to the sales-side or purchase-side `C_VAT_Code` regardless of doc-level `IsSOTrx`.
- `Doc_Invoice` — RC blocks (both ARI/API and ARC/APC branches) route `T_Credit_Acct` → `IsSOTrx=N` (input, KZ 67) and `T_Due_Acct` → `IsSOTrx=Y` (output, KZ 84/85).
- `Doc_AllocationHdr.Doc_AllocationTax.createEntries` — tax-correction builders copy `VATCode` from the source `Fact_Acct` row, so the per-leg split survives Skonto / WriteOff / payment-time corrections.

### View + report function

- `tax_accounts_details_v` — for `T_Due_Acct` rows on reverse-charge taxes, flip the signed `TaxAmt` and the CMA-row-derived `TaxBaseAmt` so the output leg mirrors the input leg. Non-RC paths byte-identical. Inner projection now exposes `tax.IsReverseCharge`.
- `report_taxaccounts` — Bug A.2 dedup partitions by `(vatcode, ad_table_id, record_id, c_tax_id)` instead of `(vatcode, documentno, c_tax_id)`. DocumentNo can collide across BPartners / orgs; `(ad_table_id, record_id)` is the unique Fact_Acct source key. Dedup keeps pre-existing RC invoices (one VATCode, two legs) counted once at level 1 / ReCap.

### Cucumber scenarios

Feature file renamed `tax_accounting_report_regression.feature` → `tax_accounting_report.feature` (no longer regression-only). 16 scenarios split into five groups:

- **TC-S1..TC-S4** — simple invoice / credit memo on sales & purchase side
- **TC-S5..TC-S6** — zero-tax (§4 UStG) sales & purchase
- **TC-S7..TC-S8** — sales + Skonto allocation, mixed-rate sales invoice
- **TC-S9..TC-S11** — credit-memo + Skonto, purchase + Skonto, purchase-credit-memo + Skonto
- **TC-S12..TC-S16** — §13b RC API / APC, RC + Skonto, reversed RC

Every scenario creates 2 VATCodes per tax (sales + purchase), asserts Fact_Acct rows with per-leg `C_VAT_Code_ID`, plus `report_taxaccounts` output across all five aggregation levels.

### Supporting test infrastructure

- `C_VAT_Code_StepDef` — drops single-VATCode-per-tax fail-fast; uses 10-char UUID prefix as Value (column is VARCHAR(10) and longer values silently truncate).
- `FactAcctLineMatcher` / `FactAcctMatchersFactory` — `Fact_Acct records are matching` step now resolves a `C_VAT_Code_ID` identifier column via `C_VAT_Code_StepDefData`, mirroring `TaxReportRowMatcher#vatCodeMatches`.
- `report_taxaccounts` gains a new optional trailing parameter `p_c_tax_id` for per-scenario isolation.

### Migration scripts

- `5798570_sys_gh29361_report_taxaccounts_add_p_c_tax_id.sql` — adds `p_c_tax_id` parameter.
- `5798830_sys_gh29361_report_taxaccounts_rc_dedup.sql` — Bug A.2 RC TaxBaseAmt dedup.
- `5798970_sys_gh29361_rc_symmetric_netamt.sql` — view fix for RC symmetric reporting.
- `5798980_sys_gh29361_report_taxaccounts_partition_by_record.sql` — dedup partitions by `(ad_table_id, record_id)` instead of `documentno`.

## Legal basis

- §13b UStG Abs. 5 — recipient reports both output (KZ 84/85) and input deduction (KZ 67) on the same transaction, identical amounts.
- §17(1) UStG — both sides adjust by the same amount on any §17 correction (Skonto, WriteOff).
- UStVA Anleitung — KZ cells expect positive declaration-style amounts per side.
- EU VAT Directive 2006/112/EC Art. 194–199 (RC) + Art. 178 (input deduction evidence) + Art. 250 (VAT return content).

## Test plan

- [x] Backend compiles
- [x] `de.metas.cucumber` test-compiles
- [x] All migration scripts apply cleanly against local infra
- [x] 16/16 scenarios pass locally on `tax_accounting_report.feature`
- [x] Non-regression: `invoice_reverseCharge_accounting.feature` still 4/4 green
- [ ] CI green on this branch


backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
```diff
@@ -0,0 +1,330 @@
...
+		final Tax tax = taxTable.get(taxIdentifier);
+		final LocalDate dateFrom = LocalDate.parse(dateFromStr);
+		final LocalDate dateTo = LocalDate.parse(dateToStr);
+
+		final List<Map<String, String>> actualRows = queryReportLevel4Rows(tax.getTaxId().getRepoId(), dateFrom, dateTo);
+
+		assertRowsMatch(dataTable, actualRows, "report_taxaccounts level 4 (C_Tax=" + taxIdentifier + ")");
+	}
+
+	private List<Map<String, String>> queryViewRows(final int taxId, final LocalDate dateFrom, final LocalDate dateTo) throws SQLException
```
teosarca: use TaxId as param, avoid primitive or generic type params.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
```diff
@@ -0,0 +1,330 @@
...
+				+ "        ev.name          AS accountname"
+				+ " FROM de_metas_acct.tax_accounts_details_v v"
+				+ " LEFT JOIN C_ElementValue ev ON ev.C_ElementValue_ID = v.account_id"
+				+ " WHERE v.C_Tax_ID = ?"
+				+ "   AND v.DateAcct >= ? AND v.DateAcct <= ?"
+				+ "   AND v.postingtype = 'A'"
+				+ " ORDER BY v.accountconceptualname, v.documentno, v.taxamt";
+
+		final List<Map<String, String>> rows = new ArrayList<>();
+		try (final PreparedStatement pstmt = DB.prepareStatement(sql, ITrx.TRXNAME_None))
```
teosarca: pls use one of the DB.retrieveRows helper methods.
Instead of retrieving rows to Map, pls consider introducing an TaxReportRow POJO (using lombok's Value and Builder). avoid using primitive or genetic types, e.g.
* for AccountConceptualName you can use de.metas.acct.AccountConceptualName
* fetch also the "Currency" or "source_currency" db columns and then u can construct de.metas.currency.Amount instances of TaxAmt, TaxBaseAmt.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
```diff
@@ -0,0 +1,330 @@
...
+			}
+		}
+		return rows;
+	}
+
+	private List<Map<String, String>> queryReportLevel4Rows(final int taxId, final LocalDate dateFrom, final LocalDate dateTo) throws SQLException
+	{
+		// p_ad_org_id=NULL means ALL orgs — the function still requires NOT NULL match, so use the view's org (1000000 is the default metasfresh org).
+		// Better: look up the org from the tax's client/org.
+		// For test isolation we rely on p_c_tax_id filtering — the org filter just needs to match the posting org.
```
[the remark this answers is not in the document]
teosarca: agreee, we shall drop that coment and simply use de.metas.cucumber.stepdefs.StepDefConstants#ORG_ID for now

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/report/tax_accounting_report_regression.feature
```diff
@@ -0,0 +1,280 @@
...
+    And the invoice identified by exemptAriInv is completed
+    And Wait until documents exemptAriInv is posted
+
+    # Zero-tax ARI: T_Due_Acct line has DR=0, CR=0. View's TaxAmt = 0 - 0 = 0.
+    # View's outer TaxBaseAmt: DocBaseType=ARI → -500.
+    Then the tax_accounts_details_v for C_Tax "exemptSalesTax" between "2024-01-01" and "2024-01-31" returns:
+      | AccountConceptualName | TaxAmt | TaxBaseAmt |
+      | T_Due_Acct            | 0      | -500       |
+
+
```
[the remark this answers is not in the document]
teosarca: Addressed in 27f16e055fe: added `report_taxaccounts level 4` assertion for TC-S5.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/report/tax_accounting_report_regression.feature
```diff
@@ -0,0 +1,280 @@
...
+      | Identifier     | C_Invoice_ID | M_Product_ID       | QtyInvoiced | C_Tax_ID           |
+      | exemptApiInvL1 | exemptApiInv | exemptPurchaseProd | 1 PCE       | exemptPurchaseTax  |
+    And the invoice identified by exemptApiInv is completed
+    And Wait until documents exemptApiInv is posted
+
+    # Zero-tax API: T_Credit_Acct line has DR=0, CR=0. View's TaxAmt = 0 - 0 = 0.
+    # View's outer TaxBaseAmt: DocBaseType=API → +500.
+    Then the tax_accounts_details_v for C_Tax "exemptPurchaseTax" between "2024-01-01" and "2024-01-31" returns:
+      | AccountConceptualName | TaxAmt | TaxBaseAmt |
+      | T_Credit_Acct         | 0      | 500        |
```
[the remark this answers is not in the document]
teosarca: Addressed in 27f16e055fe: added `report_taxaccounts level 4` assertion for TC-S6.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/report/tax_accounting_report_regression.feature
```diff
@@ -0,0 +1,280 @@
...
+      | TaxAmt | NetAmt |
+      | -190   | -1000  |
+
+
+# ############################################################################################################################################
+# TC-S5 — Zero-tax sales invoice (tax-exempt, e.g. §4 UStG)
+# ############################################################################################################################################
+  @Id:S0467_TAR_050
+  @from:cucumber
+  Scenario: zero-tax sales invoice produces a T_Due row with zero amounts
```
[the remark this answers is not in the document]
teosarca: Renamed in 27f16e055fe to "...with zero tax amount".

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/report/tax_accounting_report_regression.feature
```diff
@@ -0,0 +1,280 @@
...
+      | AccountConceptualName | TaxAmt | TaxBaseAmt |
+      | T_Due_Acct            | 0      | -500       |
+
+
+# ############################################################################################################################################
+# TC-S6 — Zero-tax purchase invoice (tax-exempt)
+# ############################################################################################################################################
+  @Id:S0467_TAR_060
+  @from:cucumber
+  Scenario: zero-tax purchase invoice produces a T_Credit row with zero amounts
```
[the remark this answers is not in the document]
teosarca: Renamed in 27f16e055fe to "...with zero tax amount".

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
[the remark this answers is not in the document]
teosarca: Addressed in 27f16e055fe: the misleading comment is removed, and the function is now called with `StepDefConstants.ORG_ID` (per teosarca's suggestion).

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
```diff
@@ -0,0 +1,330 @@
...
+		final String sql = "SELECT level, vatcode, accountname, taxname, netamt, taxamt, documentno, bpartnername"
+				+ " FROM de_metas_acct.report_taxaccounts("
+				+ "      p_ad_org_id      => (SELECT fa.ad_org_id FROM fact_acct fa WHERE fa.c_tax_id = ? LIMIT 1),"
```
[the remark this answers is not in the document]
teosarca: Addressed in 27f16e055fe: the nondeterministic subquery is gone — the function is now called with `StepDefConstants.ORG_ID` directly.

backend/de.metas.acct.base/src/main/sql/postgresql/system/43-de.metas.acct/5798570_sys_gh29361_report_taxaccounts_add_p_c_tax_id.sql
```diff
@@ -0,0 +1,438 @@
...
+        SELECT '3' ::varchar            AS level,
+               vatcode,
+               AccountName,
+               NULL::timestamp          AS DateAcct,
+               NULL::text               AS DocumentNo,
+               NULL::text               AS BPartnerName,
+               taxname::text            AS Taxname,
+               NULL::numeric            AS TotalAmt,
+               NULL::numeric            AS TaxAmt,
+               NULL::numeric            AS TaxAmt,
```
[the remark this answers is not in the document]
teosarca: Fixed in 27f16e055fe: the second `NULL::numeric` in the level-3 INSERT is now aliased as `NetAmt` (matches the positional target column). DDL file and migration script both updated.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
teosarca: Done in 27f16e055fe: `queryViewRows` and `queryReportLevel4Rows` now take `TaxId` (not `int`).

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
teosarca: Done in 27f16e055fe: switched to `DB.retrieveRowsOutOfTrx` with a typed `TaxReportRow` POJO (`@Value @Builder`). `AccountConceptualName` is used for the account role, and `TaxAmt`/`NetAmt`/`TaxBaseAmt` are now `de.metas.currency.Amount` instances built from the `currency` column fetched alongside.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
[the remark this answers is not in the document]
teosarca: Done in 27f16e055fe: comment removed, function now called with `StepDefConstants.ORG_ID`.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
```diff
@@ -0,0 +1,369 @@
...
+		@Nullable String accountNo;
+		@Nullable String accountName;
+		@Nullable String taxName;
+		@Nullable Amount taxAmt;
+		@Nullable Amount netAmt;
+		@Nullable String documentNo;
+		@Nullable String bpartnerName;
+
+		@Nullable
+		public String getAsStringForColumn(@NonNull final String column)
```
teosarca: i think this is over-engineering. why not simply do the assertions field by field, simple stupid?

teosarca:
Good point — at `p_level='4'` the function is literally `SELECT * FROM tax_accounts_details_v` filtered by the parameters, so the view-level assertion added no real coverage and coupled the test to an internal API. Dropped in a9183ff3867 — cucumber now exercises only `report_taxaccounts`.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
teosarca: Agreed, dropped in c5d0ffad0bc: no more `getAsStringForColumn` switch, no generic row-matching — expected and actual rows are compared positionally and each column is asserted directly via AssertJ. Then in ab5c2fc3503 the numeric columns use `DataTableRow.getAsOptionalAmount(...)` (per your follow-up) so `190` in the feature file inherits the actual's currency and `190 CHF` checks the currency too.

teosarca:
Good call re levels — in c5d0ffad0bc the step is parametric on `p_level` and `TC-S1` now also asserts level "1" and "ReCap" so the subtotal path gets regression coverage. Other scenarios still check level "4" only; happy to add summary-level assertions to them as well if you want the matrix exhaustive.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
```diff
@@ -0,0 +1,234 @@
...
+				+ "      p_isshowdetails  => 'Y',"
+				+ "      p_level          => ?,"
+				+ "      p_ad_language    => 'de_DE',"
+				+ "      p_c_tax_id       => ?"
+				+ " )";
+
+		return DB.retrieveRowsOutOfTrx(
+				sql,
+				Arrays.asList(orgId, dateFrom, dateTo, level, taxId),
+				rs -> {
```
teosarca: extract the lambda as a static method

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
```diff
@@ -0,0 +1,234 @@
...
+
+		final List<TaxReportRow> actualRows = new ArrayList<>();
+		actualRows.addAll(queryReportRows(StepDefConstants.ORG_ID, taxId, dateFrom, dateTo, null));
+		actualRows.addAll(queryReportRows(StepDefConstants.ORG_ID, taxId, dateFrom, dateTo, "ReCap"));
+		actualRows.sort(ROW_ORDER);
+
+		final ImmutableList<DataTableRow> expectedRows = DataTableRows.of(dataTable).stream().collect(ImmutableList.toImmutableList());
+
+		final SoftAssertions softly = new SoftAssertions();
+		softly.assertThat(actualRows.size())
```
teosarca: why not using assertThat(...)...hasSameSizeAs(....)

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
```diff
@@ -0,0 +1,234 @@
...
+				.as("report_taxaccounts (C_Tax=%s): row count", taxIdentifier)
+				.isEqualTo(expectedRows.size());
+
+		final int common = Math.min(actualRows.size(), expectedRows.size());
```
teosarca: at this point we can assume they have the same size, so just use one, eg expectedRows.size()

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
```diff
@@ -0,0 +1,234 @@
...
+		final TaxId taxId = taxTable.get(taxIdentifier).getTaxId();
+		final LocalDate dateFrom = LocalDate.parse(dateFromStr);
+		final LocalDate dateTo = LocalDate.parse(dateToStr);
+
+		final List<TaxReportRow> actualRows = new ArrayList<>();
+		actualRows.addAll(queryReportRows(StepDefConstants.ORG_ID, taxId, dateFrom, dateTo, null));
+		actualRows.addAll(queryReportRows(StepDefConstants.ORG_ID, taxId, dateFrom, dateTo, "ReCap"));
+		actualRows.sort(ROW_ORDER);
+
+		final ImmutableList<DataTableRow> expectedRows = DataTableRows.of(dataTable).stream().collect(ImmutableList.toImmutableList());
```
teosarca: why not using de.metas.cucumber.stepdefs.DataTableRows#toList

teosarca: also shall we document or shall we say something about the order of those rows.
remark that we are already sorting the actual rows...

pls clarify why this is better than iterating each expectation and
* find out the actual row (by level for example)
* remove it from the actuals (just to not be matched again)
* assert actual matches expected

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
```diff
@@ -0,0 +1,234 @@
...
+			@NonNull final String dateToStr,
+			@NonNull final DataTable dataTable)
+	{
+		final TaxId taxId = taxTable.get(taxIdentifier).getTaxId();
+		final LocalDate dateFrom = LocalDate.parse(dateFromStr);
+		final LocalDate dateTo = LocalDate.parse(dateToStr);
+
+		final List<TaxReportRow> actualRows = new ArrayList<>();
+		actualRows.addAll(queryReportRows(StepDefConstants.ORG_ID, taxId, dateFrom, dateTo, null));
+		actualRows.addAll(queryReportRows(StepDefConstants.ORG_ID, taxId, dateFrom, dateTo, "ReCap"));
```
teosarca: we do we have to call it again with "ReCap" level?

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/report/tax_accounting_report_regression.feature
```diff
@@ -0,0 +1,569 @@
...
+      | Identifier | C_Invoice_ID | M_Product_ID | QtyInvoiced | C_Tax_ID   |
+      | ariInvL1   | ariInv       | salesProd1   | 1 PCE       | salesTax19 |
+    And the invoice identified by ariInv is completed
+    And Wait until documents ariInv is posted
+
+    # Regression baseline across all aggregation levels. For sales invoices (ARI),
+    # T_Due_Acct posts AmtAcctCr=190 → level 4 has TaxAmt=-190/NetAmt=-1000 and
+    # levels 1/2/3/ReCap sum up to the same single-row amounts.
+    Then report_taxaccounts for C_Tax "salesTax19" between "2024-01-01" and "2024-01-31" returns:
+      | Level | AccountName | TaxAmt | NetAmt | TaxAmt_SUM | NetAmt_SUM |
```
teosarca: i would prefer the following column order:
 Level 
AccountName 
NetAmt_SUM
TaxAmt_SUM 
NetAmt 
TaxAmt

teosarca: Done in 989f5cc553b. Column order is now `Level | AccountName | NetAmt_SUM | TaxAmt_SUM | NetAmt | TaxAmt` across all scenarios.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
teosarca: Because `p_level=NULL` only inserts the 1/2/3/4 levels (see the `IF p_level IS NULL OR p_level = '<n>'` conditions in the function body) — ReCap is only produced when `p_level='ReCap'`. So two calls are needed to get everything. Documented inline in 989f5cc553b.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
teosarca: Agreed and switched to find-and-remove in 989f5cc553b: for each expected row, find the first actual row that matches on all non-blank columns, remove it from the pool, repeat. Missing match → fail for that expected; leftover actual → fail too. Decouples the test from the DB's row order so authors can list rows in whatever order reads best. Documented in the class javadoc.

teosarca: Done in 989f5cc553b.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
teosarca: Superseded by the find-and-remove rewrite in 989f5cc553b — no more explicit size check needed; the "no match" / "unexpected extra rows" assertions now cover size mismatches implicitly.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
teosarca: Done in c23cc286067: extracted as `TaxAccountingReport_StepDef::toTaxReportRow`.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
teosarca: Superseded by the find-and-remove rewrite in 989f5cc553b — the explicit size check is gone; mismatches now surface as "no match for expected row" / "unexpected extra rows" assertions, which give better error messages than `hasSameSizeAs`.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/report/TaxAccountingReport_StepDef.java
```diff
@@ -0,0 +1,233 @@
...
+		if (!stringMatches(expected, "VatCode", actual.getVatCode())) { return false; }
+		if (!stringMatches(expected, "AccountName", actual.getAccountName())) { return false; }
+		if (!stringMatches(expected, "TaxName", actual.getTaxName())) { return false; }
+		if (!stringMatches(expected, "DocumentNo", actual.getDocumentNo())) { return false; }
+		if (!stringMatches(expected, "BPartnerName", actual.getBpartnerName())) { return false; }
+		if (!amountMatches(expected, "TaxAmt", actual.getTaxAmt())) { return false; }
+		if (!amountMatches(expected, "NetAmt", actual.getNetAmt())) { return false; }
+		if (!amountMatches(expected, "TaxAmt_SUM", actual.getTaxAmtSum())) { return false; }
+		if (!amountMatches(expected, "NetAmt_SUM", actual.getNetAmtSum())) { return false; }
+		return true;
```
teosarca: why weird chain of if not, if not and at the end return true;

why not simply:
return amountMatches(...)
    && amountMatches(...)
..........
?

teosarca: Done — using a single `&&` chain now. Pushed as the next commit.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/tax_declaration/TaxAccountingReport_StepDef.java
```diff
@@ -0,0 +1,338 @@
...
+
+	/**
+	 * Resolves an {@link Account} to the {@code "<value> <name>"} string the DB function emits,
+	 * matching {@code (accountno || ' ' || accountname)::text AS AccountName} in
+	 * {@code report_taxaccounts.sql}.
+	 */
+	private static String resolveAccountName(
+			@NonNull final Account account,
+			@NonNull final IAccountDAO accountDAO,
+			@NonNull final ElementValueRepository elementValueRepo)
```
teosarca: what's the actually reason tihs method is static and services are provided as param???

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/tax_declaration/TaxAccountingReport_StepDef.java
```diff
@@ -0,0 +1,338 @@
...
+	{
+		return taxInfoCache.computeIfAbsent(taxId, TaxAccountingReport_StepDef::loadTaxInfo);
+	}
+
+	private static TaxInfo loadTaxInfo(@NonNull final TaxId taxId)
+	{
+		final IAcctSchemaBL acctSchemaBL = Services.get(IAcctSchemaBL.class);
+		final IAccountDAO accountDAO = Services.get(IAccountDAO.class);
+		final TaxAccountsRepository taxAccountsRepo = SpringContextHolder.instance.getBean(TaxAccountsRepository.class);
+		final ElementValueRepository elementValueRepo = SpringContextHolder.instance.getBean(ElementValueRepository.class);
```
teosarca: pls always declare services as class fields!!!

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/tax_declaration/TaxAccountingReport_StepDef.java
```diff
@@ -0,0 +1,338 @@
...
+	 * {@code report_taxaccounts.sql}.
+	 */
+	private static String resolveAccountName(
+			@NonNull final Account account,
+			@NonNull final IAccountDAO accountDAO,
+			@NonNull final ElementValueRepository elementValueRepo)
+	{
+		final ElementValueId elementValueId = accountDAO.getElementValueIdByAccountId(account.getAccountId());
+		final ElementValue elementValue = elementValueRepo.getById(elementValueId);
+		return elementValue.getValue() + " " + elementValue.getName();
```
teosarca, line 465: pls document why we are concatenating like that! a code ref to db function would also help

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/tax_declaration/TaxAccountingReport_StepDef.java
teosarca: Agreed — services are now class fields (`IAcctSchemaBL`, `IAccountDAO`, `TaxAccountsRepository`, `ElementValueRepository`), and `loadTaxInfo` / `resolveAccountName` are instance methods. Committed as c07b29f7846.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/tax_declaration/TaxAccountingReport_StepDef.java
teosarca: Done in c07b29f7846 — services are class fields now.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/tax_declaration/TaxAccountingReport_StepDef.java
teosarca: Done in c07b29f7846 — added an inline javadoc that points to the matching `(accountno || ' ' || accountname)::text AS AccountName` expression in `backend/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/report_taxaccounts.sql`.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/tax_declaration/TaxAccountingReport_StepDef.java
```diff
@@ -0,0 +1,370 @@
...
+		}, null);
+		final String ctx = "report_taxaccounts (C_Tax=" + taxIdentifier + ")";
+		final SoftAssertions softly = new SoftAssertions();
+
+		// For each expected row, find and remove a matching actual row.
+		for (final DataTableRow expected : expectedRows)
+		{
+			final TaxReportRow match = findAndRemoveMatch(expected, actualRows, taxInfo);
+			softly.assertThat(match)
+					.as("%s: no actual row matches expected %s", ctx, expected.asMap())
```
teosarca: how do i know which were the actual rows that were evaluated? that would really help to troubleshoot

teosarca: Done in b1ed7cbb55d — both failure messages now include all rows returned by the function (as a numbered list) so you can see every candidate that was evaluated. Example:

```
report_taxaccounts (C_Tax=salesTax19): no actual row matches expected {Level=4, ...}
    all actual rows returned by the function:
      [0] TaxReportRow(level=1, vatCode=..., taxAmtSum=-190 EUR, ...)
      [1] TaxReportRow(level=2, ..., accountName=90032 T_DUE_ACCT, ...)
      ...
```

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/report/tax_accounting_report_regression.feature
```diff
@@ -0,0 +1,580 @@
...
+    And the invoice identified by allocSalesInv is completed
+
+    # Payment amount = invoice gross (1190) minus 2% early-payment discount (23.80) = 1166.20 EUR.
+    # The discount's 19% tax portion is 3.80 EUR; the net portion is 20.00 EUR.
+    And metasfresh contains C_Payment
+      | Identifier    | C_BPartner_ID | PayAmt      | IsReceipt | C_BP_BankAccount_ID |
+      | allocPayment  | customer      | 1166.20 EUR | true      | org_EUR_account     |
+    And the payment identified by allocPayment is completed
+
+    And allocate payments to invoices
```
teosarca: pls use something like

    And create and complete manual payment allocations
      | C_AllocationHdr_ID | C_Invoice_ID | C_Payment_ID | DiscountAmt |
      | alloc | allocSalesInv | allocPayment    | 23.80 EUR      |


so we would also get the allocation identifier.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/report/tax_accounting_report_regression.feature
```diff
@@ -0,0 +1,580 @@
...
+    # Two level-4 rows (invoice + allocation discount correction). Subtotals:
+    #   sum(TaxAmt) = -190 + 3.80 = -186.20
+    #   sum(NetAmt) = -1000 + 20  = -980
+    Then report_taxaccounts for C_Tax "allocTax19" between "2024-01-01" and "2024-01-31" returns:
+      | Level | AccountConceptualName | NetAmt_SUM | TaxAmt_SUM | NetAmt | TaxAmt | C_BPartner_ID | DocumentNo    |
+      | 1     |                       | -980       | -186.20    |        |        | -             | -             |
+      | 2     | T_Due_Acct            | -980       | -186.20    |        |        | -             | -             |
+      | 3     | T_Due_Acct            | -980       | -186.20    |        |        | -             | -             |
+      | 4     | T_Due_Acct            |            |            | -1000  | -190   | customer      | allocSalesInv |
+      | 4     | T_Due_Acct            |            |            | 20     | 3.80   | customer      |               |
```
teosarca: fill the allocation identifier.

ofc, apply this for all payment allocation cases

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/report/tax_accounting_report_regression.feature
```diff
@@ -0,0 +1,580 @@
...
+    And metasfresh contains C_Payment
+      | Identifier    | C_BPartner_ID | PayAmt      | IsReceipt | C_BP_BankAccount_ID |
+      | allocPayment  | customer      | 1166.20 EUR | true      | org_EUR_account     |
+    And the payment identified by allocPayment is completed
+
+    And allocate payments to invoices
+      | C_Invoice_ID  | C_Payment_ID | DiscountAmt |
+      | allocSalesInv | allocPayment | 23.80 EUR   |
+
+    And Wait until documents allocSalesInv is posted
```
teosarca: not only `allocSalesInv` but allocation identifier also (comma separated)

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/report/tax_accounting_report_regression.feature
teosarca: Done in f0907c6a0e0 — switched to `create and complete manual payment allocations` with `C_AllocationHdr_ID` set, and simplified to discount-only (no payment). The manual step registers the AllocationHdr in the StepDef registry, so it's now usable downstream (wait-until-posted, DocumentNo column).

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/report/tax_accounting_report_regression.feature
teosarca: Done — `C_AllocationHdr_ID` is now set on all four allocation scenarios (ariAlloc / arcAlloc / apiAlloc / apcAlloc).

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/report/tax_accounting_report_regression.feature
teosarca: Done — `Wait until documents allocSalesInv, ariAlloc are posted` (and the equivalent for the other three allocation scenarios).

backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/impl/VATCodeDAO.java
```diff
@@ -97,6 +101,49 @@ public VatCodeId getIdByCodeAndOrgId(
...
+		record.setC_AcctSchema_ID(request.getAcctSchemaId().getRepoId());
+		record.setC_Tax_ID(request.getTaxId().getRepoId());
+		record.setVATCode(request.getVatCode());
+		if (request.getOrgId() != null)
+		{
+			record.setAD_Org_ID(request.getOrgId().getRepoId());
+		}
+		if (request.getIsSOTrx() != null)
+		{
+			record.setIsSOTrx(request.getIsSOTrx() ? "Y" : "N");
```
teosarca: pls use de.metas.acct.model.X_C_VAT_Code#ISSOTRX_Yes or de.metas.acct.model.X_C_VAT_Code#ISSOTRX_No.
rule: always avoid hardcoded constants

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/tax/C_VAT_Code_StepDef.java
```diff
@@ -0,0 +1,116 @@
...
+ * And metasfresh contains C_VAT_Codes:
+ *   | Identifier   | C_Tax_ID   | IsSOTrx |
+ *   | salesVat19   | salesTax19 | Y       |
+ * }</pre>
+ *
+ * @see C_VAT_Code_StepDefData
+ * @see IVATCodeDAO#createVATCode
+ * @see IVATCodeDAO#existsForAcctSchemaAndTax
+ */
+@RequiredArgsConstructor
```
teosarca, line 49: for step def, why the documentation javadoc is on class header on not on each step def method?

teosarca: rule: for cucumber step defs, pls keep the javadoc on the method that defines the step.
on class level only the overall documentation (if any, if applicable) shall be

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/tax_declaration/tax_accounting_report_regression.feature
```diff
@@ -0,0 +1,782 @@
...
+## sub-PR 2 must fix + add scenarios TC-S9b / TC-S11b exercising the payment-based path.
+##
+## [VAT code] Each test creates a per-tax C_VAT_Code (via `metasfresh contains C_VAT_Codes:`) so the
+##   report now emits the real vatcode string instead of the "Keine MwSt." getmessage('notax')
+##   fallback. The {@code C_VAT_Code_ID} column in every `report_taxaccounts ... returns:` block
+##   asserts the identifier's {@code VATCode} matches the DB row's vatcode. Customer VAT-code
+##   configuration remains tracked under https://github.com/metasfresh/me03/issues/29363.
+##
+## Scenarios not flagged above are believed fully correct and customer-validated.
+## ============================================================================================
```
teosarca: pls keep the specific documentaion on before each Scenario, so we have the documenation + scenario groupped and not here on top.

also pls clean it up and remove what does no longer make sense or was fixed/cleared.

backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/impl/VATCodeDAO.java
teosarca: Addressed: `VATCodeDAO.createVATCode()` now uses `X_C_VAT_Code.ISSOTRX_Yes` / `ISSOTRX_No` instead of hardcoded `"Y"`/`"N"`.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/tax/C_VAT_Code_StepDef.java
teosarca: Addressed: moved the step documentation from the class header to the `@And`-annotated method `metasfresh_contains_c_vat_codes`. Class-level JavaDoc is now a one-liner with `@see` pointers to the companion data class and the DAO methods.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/tax_declaration/tax_accounting_report_regression.feature
teosarca: Addressed: the big KNOWN ISSUES header block is gone. Feature-wide header now only states the feature purpose + points to the me03 spotlight / `ai-work/29361/REQUIREMENTS.md` for the authoritative sub-PR 2 scope. Scenario-specific notes (DiscountAmt sign direction, §17 UStG verdict, Steuerberater sign-off requirement) moved into each scenario's comment block (TC-S7/S9/S10/S11). Stale items removed (non-RC sign "locked in as correct", RC latent-bug notes that belong in REQUIREMENTS, etc.).

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/tax_declaration/tax_accounting_report.feature
```diff
@@ -0,0 +1,1103 @@
...
+      | C_AllocationHdr_ID | C_Invoice_ID  | DiscountAmt |
+      | rcApiAlloc         | rcAllocApiInv | -23.80 EUR  |
+
+    And Wait until documents rcAllocApiInv, rcApiAlloc are posted
+
+    # Invoice posting (as TC-S12) + symmetric allocation correction per parent #28726 Decision D1.
+    # Tax correction amount on RC + Skonto = DiscountAmt × TaxRate = 23.80 × 0.19 = 4.52 EUR
+    # (not 3.80 as in non-RC — RC invoice has TaxAmt=0, so the correction base is the gross
+    # DiscountAmt multiplied by the tax rate, not the tax portion of the gross).
+    And Fact_Acct records balances for documents rcAllocApiInv,rcApiAlloc are matching
```
teosarca: why there is no "And Fact_Acct records are matching" step before where we check the invoice and allocation accounting (only tax relevant facts)?

PR #23534, opened 2026-04-21
`HU Labels (LU & TU)` report update

riadmen:


backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/70-de.metas.fresh/5798960_sys_gh29310_AD_Val_Rule_HU_Label_Report.sql
```diff
@@ -0,0 +1,30 @@
...
+--   584694 Finishedproduct Label
+--   585387 SSCC Etikett LU Coop
+--
+-- Excluded by design (they have their own entry points, not reached through the multi-HU launcher):
+--   584980 M_HU_Report_QRCode — separate process class with its own IsPrintPreview
+--   585114 M_HU_Report_Print_Labels, 584997 M_HU_Report_LU_Label — single-HU launchers that
+--           already preview by default via M_HU_Report_Print_Template
+
+UPDATE AD_Val_Rule
+SET Code = 'AD_Process.AD_Process_ID IN (540370, 540412, 540413, 540414, 540415, 540416, 540933, 541195, 584694, 585387)',
```
metas-ts: I see the point hof having explicit AD_Process_IDs, but what i a user wants to add a custom report? they will add the ID here, at the risk that a change in metafresh will override it.
So I think we do need something a *documented* `AD_Process.Value` prefix that is also picked up. wdyt?

metas-ts (changes requested):

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/70-de.metas.fresh/5798960_sys_gh29310_AD_Val_Rule_HU_Label_Report.sql
metas-ts: Good point — agreed. Extending the val-rule so both:
- the 10 known core label Jaspers are always available (explicit IDs), and
- any `AD_Process` whose `Value` starts with `HU_Label_` is picked up automatically (customer extensions).

Proposed final rule:
```
AD_Process.AD_Process_ID IN (540370, 540412, 540413, 540414, 540415, 540416, 540933, 541195, 584694, 585387)
OR AD_Process.Value ILIKE 'HU_Label_%'
```

I'll document the `HU_Label_<Customer>_<Type>` naming convention in mf15-documentation#100 so customers know how to get their custom label Jasper into the dropdown. Pushing the migration update + documentation in a minute.

metas-ts (approved):

PR #23582, opened 2026-04-22
gh29361: fix TotalAmt column in report_taxaccounts for Reverse-Charge rows

teosarca:
## Summary

`report_taxaccounts(...)` (the SQL function backing the `Mehrwertsteuer-Verprobung(Excel) 3` report, AD_Process `585325`) computes `TotalAmt = taxbaseamt + taxamt` per row. For non-RC invoices this is the invoice gross (sum across rows = invoice gross). For §13b RC invoices the two legs per tax produce `1190` and `810` — neither of which is the invoice total (for RC the invoice total IS the net amount; no cash VAT movement).

**Fix**: keep the existing formula for non-RC rows; drop only the tax summand when the row's tax is reverse-charge. Non-RC output is byte-identical; RC rows now show `TotalAmt = NetAmt = invoice total`.

Tracked on me03#29361 as TC-E1.

## Changes

- `tax_accounts_details_v.sql` — expose `isreversecharge` in the outer SELECT (the inner subquery already projects it since the RC symmetric-reporting view fix).
- `report_taxaccounts.sql` — RC-aware `TotalAmt` branch.
- Two migration scripts (one per DDL entity, each with a `-- Source DDL:` header).
- Cucumber: `TaxReportRow` + `TaxReportRowMatcher` + `TaxReportRowRepository` pick up `TotalAmt`; `TotalAmt` expectations added to TC-S1 (non-RC regression), TC-S12 (RC API), TC-S13 (RC APC).

## Test plan

- [x] Local: all 16 scenarios of `tax_accounting_report.feature` green against `deep_tundra_uat` after applying both migrations.
- [ ] CI green.


backend/de.metas.acct.base/src/main/sql/postgresql/system/43-de.metas.acct/5799250_sys_gh29361_report_taxaccounts_rc_totalamt.sql
```diff
@@ -0,0 +1,457 @@
...
+                              ORDER BY accountno, accountname) AS row_in_doc_tax,
+           -- TotalAmt semantics:
+           -- Non-RC rows: taxbaseamt + taxamt = invoice gross for the tax portion (summed per invoice = invoice gross).
+           -- RC rows: the VAT doesn't move cash (recipient self-assesses both legs and they cancel), so the
+           -- invoice total IS the net (taxbaseamt). Including the tax summand would produce 1190 / 810 per leg,
+           -- neither of which is meaningful for the §13b declaration. The branch below preserves the non-RC
+           -- behaviour exactly and corrects only RC.
+           (CASE
+                WHEN c_currency_id = source_currency_id
+                    THEN taxbaseamt + CASE WHEN t.isreversecharge = 'Y' THEN 0 ELSE taxamt END
```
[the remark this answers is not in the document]
teosarca: Addressed in aa0a13a3a99 — refactored to resolve IsReverseCharge inline via an EXISTS subquery on C_Tax, so the function no longer depends on any view column. Dropped the view DDL change and deleted the second migration (5799260). Single migration 5799250 now completely defines the A.4 fix with no ordering dependency on any other script.

backend/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/report_taxaccounts.sql
```diff
@@ -159,9 +159,23 @@ BEGIN
...
+           -- behaviour exactly and corrects only RC.
+           --
+           -- `IsReverseCharge` is resolved by a direct lookup on C_Tax rather than relying on a view column,
+           -- so this function carries no hidden dependency on the view's projection.
            (CASE
-                WHEN c_currency_id = source_currency_id THEN (taxbaseamt + taxamt)
-                                                        ELSE NULL
+                WHEN c_currency_id = source_currency_id
+                    THEN taxbaseamt + CASE
+                                          WHEN EXISTS (SELECT 1 FROM C_Tax ct WHERE ct.C_Tax_ID = t.c_tax_id AND ct.IsReverseCharge = 'Y')
```
teosarca: i would avoid selecting from C_Tax again.
Instead i would expose tax_accounts_details_v.IsReverseChange and i will use it from there.

teosarca: regarding the return table of the this (report_taxaccounts) function, i would like you to document each return column, what exactly means and in case of amounts, pls also document which is the currency.

not sure which are the best practices in postgresql, but i would prefer to have this documentation "inline", as closer as posible to the column (but also respecting the best practices).

teosarca: Addressed in 63ce18ff1d2 (rebased → 1af2750c92e).

- Reverted the C_Tax EXISTS subquery.
- Added `isreversecharge` to the outer SELECT of `tax_accounts_details_v` (new migration `5799370_...`).
- `report_taxaccounts` now reads `t.isreversecharge` from the view.
- Renamed the function migration to `5799380_...` so the view update runs first by numeric prefix.

teosarca: Addressed in 63ce18ff1d2 (rebased → 1af2750c92e). Each column on `RETURNS TABLE` now has an inline comment describing its role, and every amount column calls out whether it is in source currency or accounting currency (with the cross-currency NULL rule for `TotalAmt`). Header block above RETURNS TABLE also documents the two-currency convention. Mirrored in the migration file `5799380_...`.

PR #23659, opened 2026-04-24
ci: forward-port cicd-dispatch --ref fix (me03#29386)

metas-ts:
Forward-port of metasfresh/metasfresh#23612 to `vampire_textbook_release`.

The `cicd-dispatch` step now calls `auto-update-base-version.yaml` on
`mf15-private-extensions` with `--ref ${SOURCE_BRANCH}` when that branch
exists (falling back to the default branch otherwise so the existing
"branch doesn't exist → dispatch directly to me03" path still activates).

Without this port, a metasfresh CI run triggered on this branch would still
run the private-extensions workflow on `new_dawn_uat` and rely on an
in-process `git checkout` + `EndBug/add-and-commit` push, which
occasionally leaves an orphan commit (observed on 2026-04-22; see
metasfresh/me03#29386).

Part of the me03#29386 close-out.

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -1731,4 +1731,24 @@ jobs:
           CICD_DISPATCH_REPO: ${{ fromJson(vars.CICD_DISPATCH_CONFIG).repo }}
           CICD_DISPATCH_WORKFLOW_FILE: ${{ fromJson(vars.CICD_DISPATCH_CONFIG).workflow_file }}
         run: |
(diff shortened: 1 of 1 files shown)


metas-ts:
Closing as redundant — `auto-port.yml` will produce the equivalent `merge/vampire_textbook_release_hotfix-to-vampire_textbook_release` PR automatically once the corresponding `_hotfix` PR's cicd goes green and merges. See https://github.com/metasfresh/me03/issues/29386 for context. Lesson: when forward-porting a change to many `_hotfix` / `_release` pairs, only open the `_hotfix` PR; auto-port handles the downstream merges.

PR #23672, opened 2026-04-24
fix(webui): cap toString reason and self-heal sticky-error documents on child invalidation

metas-ts:
## Summary

Two-commit fix for a product window occasionally returning a 100+ MB REST response that rendered the page blank.

### What was happening

When a child Document fails to load (e.g. its underlying PO is missing in the database), webapi code throws a `DBException` whose message is built by interpolating the parent Document. `Document.toString()` embeds `saveStatus` and `validStatus`, whose `toString()`s in turn embed their `reason` field. That exception message is then stored back as the new `reason`, and the cycle repeats: each subsequent error doubles the stored reason because the whole previous toString output is re-embedded.

After ~20 cycles the reason string exceeds 100 MB. The JSON window response balloons to the same size, frontend chokes, page is blank.

### Conditions for the bug to manifest

1. Parent root Document is cached in `DocumentCollection.rootDocuments`.
2. A child row gets deleted externally (other tab/session, save-time interceptor, trigger, cascade).
3. Subsequent save / refresh attempts fail with `No PO found` and the exception message embeds `document.toString()`, seeding the recursion.
4. The cached parent's `saveStatus.hasChangesToBeSaved=true`, so default `cacheReset(false)` deliberately skips it (to protect unsaved work).
5. User keeps interacting with the same window through retries instead of closing it / waiting for LRU eviction / waiting for a webapi restart.

When all five align, the reason grows exponentially. Until now the only way out was an admin `cacheReset?forgetNotSavedDocuments=true`, which calls `rootDocuments.invalidateAll()` unconditionally.

### 1. `fix(webui): cap reason in Document*Status toString to avoid O(2^n) growth`

Truncate `reason` to the first 200 chars inside `DocumentSaveStatus.toString()` and `DocumentValidStatus.toString()`. Short reasons pass through unchanged; long ones get a `...(+<n> chars)` suffix so truncation is visible in logs.

The stored reason can still grow via callers, but the rendered toString output is bounded — so any exception message built from `document.toString()` stays bounded too, and the doubling chain is severed at the rendering step.

Defensive cap; does not fix the exception sites that interpolate `document.toString()` in their messages.

### 2. `fix(webui): evict cached root document when child invalidation meets error state`

Cache invalidation for a specific child record (INSERT/UPDATE/DELETE on an included table) was only marking the parent's included-documents collection as stale — it did NOT evict the parent from `rootDocuments`. For a happy-path parent that's fine: the frontend refreshes the tab and the user sees current data.

It is NOT fine when the cached parent is already in `saveStatus.error` / `validStatus.invalid` state. The error condition that produced the failure may have been caused by the very child state that just became stale, but the sticky error survives indefinitely: next read serves the cached parent as-is, since `forRootDocumentReadonly` does not call `refreshFromRepositoryIfStaled` (only the writable path does). Until now the only way to clear this was an admin `cacheReset?forgetNotSavedDocuments=true` or LRU eviction.

Fix: when a child-record invalidation arrives for a cached parent that is already in error state, evict the parent. Next read loads a clean document from the repository and recomputes the status against current DB state.

- Happy-path parents are not evicted — behaviour unchanged for the common case (no extra DB load).
- Documents that have not been persisted yet (`isNew()`) are still protected against eviction (existing safety).
- Decision logic extracted into a pure boolean helper `DocumentCollection.shouldInvalidateRootOnChildInvalidation` so it can be unit-tested without mocking the final `Document` class.

## Test plan

- [x] `DocumentStatusReasonTruncationTest` — covers both classes for short and long reasons (truncation marker visible, length capped).
- [x] `DocumentCollectionShouldInvalidateRootTest` — covers all four boolean combinations of the eviction decision (happy path, full-invalidation requested, save-error, validStatus-invalid).
- [x] CI green.
- [x] Manual verification in the affected environment: hotswap of commit 1 stopped further growth; `cacheReset?forgetNotSavedDocuments=true` cleared the already-poisoned in-memory document; product window loads cleanly afterwards. With commit 2, future occurrences should self-heal on the next child-invalidation event without requiring an admin cache reset.

## Notes

- Commit 1 does NOT shrink an already-large reason in memory. After deploying, in-flight poisoned documents are evicted by commit 2 on the next child invalidation, by LRU eviction, or by a webapi restart. The plain `cacheReset` continues to skip them (it protects unsaved work); `cacheReset?forgetNotSavedDocuments=true` remains the manual escape hatch.
- The exception sites that build their message from `document.toString()` (e.g. `DBException("No PO found for " + document)` patterns) are the ultimate root cause. Fixing those is a larger hunt; the toString cap is the right safety net regardless.

### DIFF
--- a/backend/de.metas.ui.web.base/src/test/java/de/metas/ui/web/window/model/DocumentStatusReasonTruncationTest.java
+++ b/backend/de.metas.ui.web.base/src/test/java/de/metas/ui/web/window/model/DocumentStatusReasonTruncationTest.java
@@ -18,18 +18,17 @@ import static org.assertj.core.api.Assertions.assertThat;
  */
 class DocumentStatusReasonTruncationTest
 {
-	/** Max length applied inside the toString() methods. Must match the value used in the production code. */
-	private static final int EXPECTED_REASON_CAP = 200;
-
 	/** Slack for the other toString parts (flags, prefix/suffix, truncation marker). */
 	private static final int TOSTRING_OVERHEAD = 500;
 
 	private static String longMessage()
 	{
-		// 10 000 chars — well above the cap and large enough to detect any unbounded
-		// embedding in a single-iteration test.
-		final StringBuilder sb = new StringBuilder(10_000);
-		for (int i = 0; i < 10_000; i++)
+		// Well above the cap and large enough to detect any unbounded embedding in a
+		// single-iteration test. Sized relative to the production cap so a future bump
+		// to TOSTRING_REASON_MAX_CHARS keeps the test meaningful.
+		final int len = DocumentSaveStatus.TOSTRING_REASON_MAX_CHARS * 2;
+		final StringBuilder sb = new StringBuilder(len);
+		for (int i = 0; i < len; i++)
 		{
 			sb.append('x');
 		}
@@ -49,7 +48,7 @@ class DocumentStatusReasonTruncationTest
 
 		assertThat(s.length())
 				.as("DocumentSaveStatus.toString() must cap reason to avoid recursive blow-up")
-				.isLessThan(EXPECTED_REASON_CAP + TOSTRING_OVERHEAD);
+				.isLessThan(DocumentSaveStatus.TOSTRING_REASON_MAX_CHARS + TOSTRING_OVERHEAD);
 
 		assertThat(s)
 				.as("truncation marker must be present when reason exceeds cap")
@@ -68,7 +67,7 @@ class DocumentStatusReasonTruncationTest
 
 		assertThat(s.length())
 				.as("DocumentValidStatus.toString() must cap reason to avoid recursive blow-up")
-				.isLessThan(EXPECTED_REASON_CAP + TOSTRING_OVERHEAD);
+				.isLessThan(DocumentSaveStatus.TOSTRING_REASON_MAX_CHARS + TOSTRING_OVERHEAD);
 
 		assertThat(s)
 				.as("truncation marker must be present when reason exceeds cap")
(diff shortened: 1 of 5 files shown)


backend/de.metas.ui.web.base/src/test/java/de/metas/ui/web/window/model/DocumentStatusReasonTruncationTest.java
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * Regression test: a poisoned document can accumulate thousands of recursively-nested
+ * {@code Document{...}} snapshots in {@code validStatus.reason} / {@code saveStatus.reason},
+ * producing 100+ MB JSON responses. Capping the reason inside {@code toString()} breaks
+ * the doubling feedback loop.
+ */
+class DocumentStatusReasonTruncationTest
+{
+	/** Max length applied inside the toString() methods. Must match the value used in the production code. */
+	private static final int EXPECTED_REASON_CAP = 200;
```
metas-ts: instead of having a constant *here*, please have it in the production code.
also, i'm not sure what cap it at 200. How about e.g. 5000?

metas-ts: Done in the latest commit:

- Single source of truth: `DocumentSaveStatus.TOSTRING_REASON_MAX_CHARS` (+ static helper `truncateReasonForToString`). Both production classes call the helper; the test references the constant.
- Cap raised to **5000** as suggested. 200 was indeed too aggressive — it'd truncate any normal exception message that includes a one-level `Document{...}` snapshot for debugging. 5000 still bounds total toString output to ~5 KB per render, so the doubling chain is severed just as effectively, but logs remain useful.

PR #23724, opened 2026-04-27
test(e2e): regression guard for product window response size (me03#29487)

metas-ts:
## Summary

Adds two Playwright tests under `e2e/frontend-webui/tests/spec/product-response-size-regression.spec.js`. Both load a product window through the WebUI, capture every `GET /rest/api/window/<id>/<recordId>/` response, and assert the largest stays under **1 MB**.

1. **Initial load** of a freshly-created product.
2. **After switching to the CU-TU Allocation tab** (`AD_Tab_ID 540517` = `M_HU_PI_Item_Product` — the tab the customer hit on https://github.com/metasfresh/me03/issues/29487).

## Why

The bug fixed in https://github.com/metasfresh/metasfresh/pull/23672 produced a 16.7 MB / 33.8 s response for a single product window because `Document.toString()` recursively re-embedded the `saveStatus.reason` it had just seeded into the next exception message — O(2^n) growth. Healthy responses are < ~100 KB, so 1 MB gives ample headroom while catching catastrophic regressions early.

## Local validation

Ran on this machine against the `mf-docker` E2E stack on tag `5.175-new-dawn-uat.34166`:

```
Product window initial load: largest response 19499 bytes (0.02 MB) — http://localhost:3000/rest/api/window/140/2005603/
Product window after HU-PI tab switch: largest response 19499 bytes (0.02 MB)
2 passed (7.2s)
```

## Limitations (documented inline at the bottom of the spec)

This test does NOT recreate the orphaned-`M_HU_PI_Item_Product` condition that triggered the original report. That requires either:
- A backend `frontendTesting` helper that can insert/delete `M_HU_PI_Item_Product` rows in a separate transaction; or
- A two-context Playwright setup that mirrors the customer's session-scoped, separate-transaction delete.

Both are non-trivial. The bounded-response baseline above is the cheapest test that would have caught the original 16.7 MB report. Two follow-up TODOs (add/remove HU-PI cycle, cross-context external delete) are listed at the bottom of the spec and on me03.

## Test plan

- [x] `npx playwright test tests/spec/product-response-size-regression.spec.js` — both tests pass locally on `5.175-new-dawn-uat.34166`.
- [ ] CI (this PR) — frontend-webui Playwright job stays green.

### DIFF
--- a/e2e/frontend-webui/tests/spec/product-response-size-regression.spec.js
+++ b/e2e/frontend-webui/tests/spec/product-response-size-regression.spec.js
@@ -1,0 +1,176 @@
+import { expect } from '@playwright/test';
+import { test } from '../../playwright.config';
+import { allure } from 'allure-playwright';
+import { Backend } from '../utils/Backend';
+import { LoginPage } from '../utils/pages/LoginPage';
+import { DashboardPage } from '../utils/pages/DashboardPage';
+import { FRONTEND_BASE_URL, SLOW_ACTION_TIMEOUT, VERY_SLOW_ACTION_TIMEOUT } from '../utils/common';
+import { PRODUCT_WINDOW_ID } from '../utils/WindowIds';
+
+const MAX_PRODUCT_RESPONSE_BYTES = 1_000_000;
+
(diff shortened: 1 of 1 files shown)


metas-ts:
useless

PR #23735, opened 2026-04-27
SEPA: make C_BP_BankAccount address authoritative in pain.001 / pain.008

adi-stefan:
## Summary

The Swiss SEPA marshalers (pain.001.001.03.CH.02 vendor credit transfer and pain.008.003.02 customer direct debit) used to drop the entire creditor address from `C_BP_BankAccount` whenever any of `A_Street`, `A_Zip` or `A_City` was blank — silently substituting the partner's default billing address (pain.001) or no address at all (pain.008). The customer's expectation is that the address fields configured on the ESR bank-account record are authoritative for the SEPA file.

This PR makes the bank-account address authoritative: when the bank account carries any address data, it is emitted verbatim, blank fields stay blank, and the partner billing location is consulted only when the bank-account address is fully empty. The latent NPE in the same code path (when the bank-account address was complete but the partner had no billing location) is also fixed.

## Changes

- `BankAccount.java`: replace `isAddressComplete()` (street + zip + city all non-blank) with `isAddressEmpty()` (all four address fields blank). Country is part of the new check.
- `SEPAVendorCreditTransferMarshaler_Pain_001_001_03_CH_02`:
  - `convertPartyIdentification32CHName`: only resolve the partner billing location when the bank account address is empty. Fixes the latent NPE that would have triggered when the bank account had a complete address but the partner had no bill-to location.
  - `createStructuredPstlAdr(BankAccount, I_C_Location)`: bank account is authoritative; emit each populated field individually, skip blanks. Fall back to the location only when the bank account is empty.
  - `createUnstructuredPstlAdr(BankAccount, I_C_Location)`: same change. Address-line concatenation now skips lines that would otherwise read `"null null"` when fields are blank.
  - Helpers exposed `@VisibleForTesting`.
- `SEPACustomerDirectDebitMarshaler_Pain_008_003_02.convertPartyIdentificationSEPA5`: emit any populated address fields rather than dropping the address element when any single field is blank. Behaviour preserved for fully-empty bank accounts (no `PstlAdr` element, as before).

## Tests

- `SEPAVendorCreditTransferMarshaler_Pain_001_001_03_CH_02_Test`: new tests for the structured and unstructured helpers covering full address, partial (street only / city only / zip+city only), empty bank account → location fallback, and null bank account → location fallback.
- `SEPACustomerDirectDebitMarshaler_Pain_008_003_02Test`: new tests for the creditor party identification covering full address, partial (street only), and fully-empty (no PstlAdr) cases.

## Reference

https://github.com/metasfresh/me03/issues/29173

## Test plan

- [ ] Local JUnit run on `de.metas.payment.sepa.base` is green
- [ ] CI on the PR branch is green
- [ ] Smoke check on the target instance: a SEPA vendor credit transfer export against a bank account with a partial `A_*` address emits the bank-account fields it has and does NOT silently substitute the partner address


adi-stefan:
@claude review

PR #23768, opened 2026-04-28
Add M_Product_Labelcost table and metadata

cristinamghita:
Add a new SQL migration creating the M_Product_Labelcost table and all related application metadata (AD_Table, AD_Column, AD_Element, AD_Tab, AD_Field, AD_Sequence and DB sequence). The table stores informational product label costs (amount, currency, UOM, label, product) for display on documents (e.g. Producer Invoice) and does not affect invoice totals or accounting. Defaults: Amount = 0, default UOM = 540017 (KG), currency defaults from session and is set non-updateable. Adds foreign keys to C_Currency, C_UOM and M_Product, primary key, translations and a UI subtab under window 140.


backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/75-de.metas.fresh.reports/5800110_AddlabelcostProductFunctions.sql
```diff
@@ -0,0 +1,297 @@
...
+
+                p_value                   character varying(40),
+                p_description             character varying(255),
+                invoice_description       character varying(1024),
+                cursymbol                 character varying(10),
+                IsDisplayProductCostLabel character(1)
+            )
+AS
+$$
+SELECT COALESCE(io1.DocType, io2.DocType) || ': ' || COALESCE(io1.DocNo, io2.DocNo) AS InOuts,
```
adi-stefan, line 89: Would we ever be in a position so that io1.DocType is null but io1.DocNo isn't, or vice versa?
And if so, is that a problem? Having the DocType from one, but the docNo from the other.

cristinamghita: NO, we will not

adi-stefan (dismissed):

cristinamghita:
@claude review

adi-stefan (approved):

PR #23782, opened 2026-04-29
Make `PurchaseErrorItem` fields nullable or non-nullable as required, and handle null transaction references gracefully in `PurchaseItemRepository`.

adi-stefan:


adi-stefan:
@claude review

PR #23792, opened 2026-04-29
fix(shipment): consolidate aggregate-HU QtyPicked rows during reversal (me03#29561)

metas-ts:
## Summary

Fix duplicate `M_ShipmentSchedule_QtyPicked` rows produced during reversal of an aggregate-HU shipment, which since sp80#2314's partial unique index `M_ShipmentSchedule_QtyPicked_UI` cause the next shipment generation to fail with the doubled-quantity error.

## Root cause

`ShipmentScheduleHUTrxListener.trxLineProcessed` emits one `M_ShipmentSchedule_QtyPicked` row per `M_HU_Trx_Line`. For an **aggregate HU** (a single VHU representing N logical TUs), the snapshot replay during `MInOut.reverseCorrectIt()` routes N trx lines through the *same* VHU, producing N rows with identical `(VHU, TU, LU, QtyLU, QtyTU, QtyPicked)`. When the next shipment is generated and writes the same `M_InOutLine_ID` onto all of them, the partial unique index correctly catches the duplicate.

Forward picking always splits a fresh VHU per trx line, so this pattern was historically impossible — the listener never defended against it.

## Fix

`ShipmentScheduleHUTrxListener.trxLineProcessed` now consults a new BL helper `tryMergeQtyPickedIntoExistingForVHU(AddQtyPickedRequest)` before delegating to `addQtyPickedAndUpdateHU`. If a single un-shipped active row already exists for the same `(M_ShipmentSchedule_ID, VHU_ID)` pair AND the request is listener-shaped, the qty is added into that row instead of creating a duplicate. Otherwise the call falls through to the existing create-new path.

The merge predicate is intentionally narrow:
- request: `M_Picking_Job_Schedule_ID` must be NULL, `IsAnonymousHuPickedOnTheFly` must be `false`, qty must carry no catch-weight, qty must be **strictly positive** (negative qty is an un-allocation and stays its own audit row), HU must be virtual.
- candidate row: `M_InOutLine_ID IS NULL`, `IsActive = Y`, `M_Picking_Job_Schedule_ID IS NULL`, `IsAnonymousHuPickedOnTheFly = N`, no catch UOM/qty.
- exactly one candidate → merge; zero → fall through; two or more → log a warning and fall through (don't auto-collapse pre-existing duplicates — surface them for human inspection).

## Files

- `IHUShipmentScheduleBL.java` / `HUShipmentScheduleBL.java` — new `tryMergeQtyPickedIntoExistingForVHU` method.
- `IHUShipmentScheduleDAO.java` / `HUShipmentScheduleDAO.java` — new `retrieveMergeableListenerQtyPickedForVHU` query.
- `ShipmentScheduleHUTrxListener.java` — early-return on successful merge.
- `HUShipmentScheduleDAO_retrieveMergeableListenerQtyPickedForVHU_Test.java` — DAO predicate coverage (3 scenarios: empty, all filters, multiple-existing).
- `HUShipmentScheduleBL_tryMergeQtyPickedIntoExistingForVHU_Test.java` — request-shape guard coverage (5 scenarios: job-schedule, anonymous-on-the-fly, catch-weight, zero qty, negative qty).
- `reversedShipment.feature` — new `S29561_20` cucumber scenario reproducing me03#29561 end-to-end (aggregate HU, ship → reverse → re-ship). Existing happy-path scenario tagged `S29561_10`.

## Test plan

- [x] `de.metas.handlingunits.base` unit tests: 846 passed, 22 skipped, 0 failed.
- [x] DAO + BL unit tests for the new method (8 new test cases, all passing).
- [ ] Cucumber `reversedShipment.feature` (incl. new aggregate-HU scenario) green locally — running.
- [ ] Cucumber `reversedShipmentClearsBPartner.feature` green locally — pending.
- [ ] CI to run the full Cucumber + unit suites.

## Known follow-up

A theoretical race exists if two reversals of *different* shipments touch the **same VHU + same shipment schedule** concurrently — both transactions could see "no candidate row" and INSERT, producing two duplicate un-shipped rows that the merge can only catch on a *third* trx line in one of them. In practice a shipment-line is mutated by one user/process at a time, so this scenario is near-impossible — but worth a note for future maintainers. No protection added in this PR.

### DIFF
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/order/C_Order_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/order/C_Order_StepDef.java
@@ -362,6 +362,9 @@ public class C_Order_StepDef
 			order.setDatePromised(Timestamp.from(datePromisedToBeSet));
 		}
 
+		tableRow.getAsOptionalInstant(I_C_Order.COLUMNNAME_DateAcct)
+				.ifPresent(dateAcct -> order.setDateAcct(Timestamp.from(dateAcct)));
+
 		if (EmptyUtil.isNotBlank(poReference))
 		{
 			order.setPOReference(poReference);
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipment/reversedShipment.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipment/reversedShipment.feature
@@ -13,6 +13,7 @@ Feature: reversed shipment
     And AD_Scheduler for classname 'de.metas.material.cockpit.stock.process.MD_Stock_Update_From_M_HUs' is disabled
 
   @from:cucumber
+  @Id:S29561_10
   Scenario: we can create and complete a shipment, then reserve/correct it and check the hu's status
     Given metasfresh has date and time 2021-12-16T13:30:13+01:00[Europe/Berlin]
     And metasfresh contains M_Products:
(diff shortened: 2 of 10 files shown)


adi-stefan:
superseeded by https://github.com/metasfresh/metasfresh/pull/24478

PR #23793, opened 2026-04-29
HU weight: write real net + per-CU tare delta on listener and during VHU transfer

mrie04:
## Summary

Fix HU weight calculation so the Lieferschein / DESADV / EDI / invoice net reflects the **real product net**, not gross. Per-CU packaging delta `(GrossW − NetW) × qty` moves into `WeightTare` (where it belongs) so `WeightGross = WeightNet + WeightTare` stays correct.

Three parts:

1. **Listener** (`WeightGenerateHUTrxListener`) writes `product.NetWeight × qty` into `WeightNet` and `(GrossW − NetW) × qty` into `WeightTare`. Strict-net: short-circuits when `NetWeight` is empty rather than falling back to gross.

2. **`WeightTareAttributeValueCallout.calculateWeightTare(I_M_HU)`** extended to add the per-CU storage delta over the HU's product storages, so paths that recompute tare from scratch (aggregate updates, `AbstractProducerDestination#loadFinished`) match the incremental listener path.

3. **New `WeightTareDeltaTransferStrategy`** registered on the `WeightTare` HU-PI-attribute (migration `5800340` swaps it in from the previous Skip strategy). On a VHU-to-VHU transfer it moves only the per-CU packaging delta `(GrossW − NetW) × qty_moved`. The packing-material portion of `WeightTare` is intentionally untouched — packing material is **indivisible per physical container**, and BOTU sum on the source's children re-aggregates it after qty has moved. Avoids the half-Karton problem of a proportional split.

## Test plan

- [x] JUnit `WeightGenerateHUTrxListenerTest` — net+tare write with mismatched UOMs (gram/kg conversion); listener short-circuits when product has no `NetWeight`.
- [x] JUnit `WeightTareAttributeValueCalloutTest` — per-CU storage delta on non-aggregate TU, mismatched UOMs, aggregate convergence via `LUTUProducerDestination`.
- [x] Cucumber `huWeightAttributesFromProductMaster.feature` (new) — VHU-fill smoke test asserting WeightNet / WeightTare / WeightGross derive from product master.
- [x] Cucumber `ediDesadvPack_with_catchweight_product.feature` — extended with "no manual weighing" scenario asserting catch quantity derives from source HU's nominal NET (= 19.000 for 10 PCE × 0.245).
- [x] Cucumber `dhlShipment.feature` — un-ignored `S0335.1_100` (Karton `IsInvoiceable=N` cleanly drops the price-0 invoice line); added `S0335.1_300` covering the default `ProductWeight` `WeightSourceTypes` path.
- [x] Cucumber `ediDesadvPack_with_TU_UOM.feature` — fixed pre-existing `S0317_020` reversal-cleanup gap that leaked EDI_Desadv_Pack rows into `S0317_030`'s empty check.
- [x] All existing weight-propagation tests (`Split` / `Merge` / `Join` / `WeightTareAdjust`) still green — the new strategy is a no-op for products with `Net = Gross` (which all test products use).

## Notes

- Migration `5800340` registers `WeightTareDeltaTransferStrategy` as `AD_JavaClass` 540099 and updates every `M_HU_PI_Attribute` row for the `WeightTare` attribute.
- Migration `5800310` updates the SysConfig description for `de.metas.shipping.WeightSourceTypes` to reflect HUWeightGross interaction.
- Forward-only: existing in-flight HUs may carry manual values and are intentionally not migrated.


mrie04:
@claude review opus

mrie04:
**Note for reviewers / integrators**: the listener change to `WeightNet` (real net instead of gross-as-net) flows through to three customer-facing label SQL functions that display it as "Net Weight":

- `lu_label.sql`
- `tu_label.sql` ("Gewicht Netto")
- `intermediate_product_label.sql` ("Gewicht Netto")

Pre-fix: these labels printed the gross value under a "Net Weight" caption.
Post-fix: they print the real product net.

For products with `NetWeight = GrossWeight` (or no net set): no visible change.
For products with `NetWeight ≠ GrossWeight`: the printed value drops by `(GrossW − NetW) × qty` per CU.

**One-time discontinuity to flag to the customer**: label sheets printed before the rollout carry the old (gross-as-net) value; freshly-printed labels for the same SKU will show the corrected real net.

Customs reports are unaffected: `Docs_Purchase_InOut_Customs_Root` only checks `WeightNet` snapshot existence, `Docs_Purchase_InOut_Customs_Details` reads `WeightGross` (`= WeightNet + WeightTare`, total unchanged).

mrie04:
Thanks for the review @claude. Addressed the actionable items in 1638d44b006:

**Bug/Issue #1 (negative per-CU delta inconsistency)** — fixed. Listener now skips the per-CU tare delta when `gross ≤ net`, matching the existing guard in `WeightTareAttributeValueCallout.calculateProductPackagingDelta` and `WeightTareDeltaTransferStrategy.computePerCUDelta`. Keeps the incremental and recompute paths consistent for misconfigured products.

**Nit #2 (import ordering)** — fixed in `HandlingUnitsBL.java`.

**Nit #3 (`Services.get()` in `convertQuantityTo`)** — pre-existing pattern, not introduced by this PR; leaving as-is to keep diff focused.

PR #23800, opened 2026-04-30
Create one PO line per PurchaseCandidate

adi-stefan:
## Summary

Restores the 1-PCand-to-1-PO-line invariant required by the unique constraint `UC_C_PurchaseCandidate_Alloc_C_OrderLinePO_ID`. `PurchaseOrderFromItemFactory#addCandidate` no longer looks up an existing line by `(productId, uomId)`; each `PurchaseOrderItem` now produces a fresh `OrderLineBuilder`.

Fixes https://github.com/metasfresh/me03/issues/29575.

## Root cause

When two purchase candidates shared the same `(product, uom)` but differed in `M_AttributeSetInstance_ID` (typical for dropship cases — each SO line carries its own per-customer/lot ASI), the previous lookup-and-merge silently overwrote the per-candidate ASI / HUPI / `qtyEnteredTU` and summed the quantities, then violated `UC_C_PurchaseCandidate_Alloc_C_OrderLinePO_ID` when both `C_PurchaseCandidate_Alloc` rows tried to point at the same `C_OrderLinePO_ID`.

The 2018 migration that introduced the unique constraint already documented the intent: *"Each purchase order line can't have more than one purchase candidate."* The fix aligns the aggregation code with that long-standing invariant.

## Behaviour change

Multiple purchase candidates with identical product/UOM are no longer merged into a single PO line. They each produce a dedicated PO line. This applies to all flows (SO-driven, forecast-driven, manual `Create Purchase Orders`).

The cucumber feature files exercising the PCand → PO creation path (dropship, packing-and-attributes, project-to-SO, auto-create-from-product-planning, supply-required-via-purchase, simulation) all assert 1-PCand→1-PO-line and are unaffected by the change.

## Tests

Four new unit tests in `PurchaseOrderFromItemsAggregatorTest`:

- `differentAsi_yieldsTwoOrderLines` — regression test for the bug; asserts both ASI identity and qty per line via `tuple` form.
- `sameAsi_differentSoLines_yieldsTwoOrderLines` — documents the deliberate behaviour change.
- `sameAsi_sameSoLine_stillYieldsTwoOrderLines` — confirms the structural invariant for the partial-purchase edge case.
- `forecastDriven_nullSoLine_yieldsTwoOrderLines` — same-ASI, null SO line; stresses the forecast-driven path.

## Dead code removed

`OrderFactory#orderLineByProductAndUom` and `OrderLineBuilder#isProductAndUomMatching` had only this one caller in the backend. They are deleted.

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/order/OrderFactory.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/OrderFactory.java
@@ -43,7 +43,6 @@ import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Optional;
 
 import static org.adempiere.model.InterfaceWrapperHelper.saveRecord;
 
@@ -193,16 +192,6 @@ public class OrderFactory
 			final OrderLineBuilder orderLineBuilder = new OrderLineBuilder(this);
 			orderLineBuilders.add(orderLineBuilder);
 			return orderLineBuilder;
-		}
-	}
-
-	public Optional<OrderLineBuilder> orderLineByProductAndUom(final ProductId productId, final UomId uomId)
-	{
-		try (final MDCCloseable ignored = TableRecordMDC.putTableRecordReference(order))
-		{
-			return orderLineBuilders.stream()
-					.filter(orderLineBuilder -> orderLineBuilder.isProductAndUomMatching(productId, uomId))
-					.findFirst();
 		}
 	}
 
--- a/backend/de.metas.business/src/main/java/de/metas/order/OrderLineBuilder.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/OrderLineBuilder.java
@@ -296,12 +296,6 @@ public class OrderLineBuilder
 		return this;
 	}
 
-	public boolean isProductAndUomMatching(@Nullable final ProductId productId, @Nullable final UomId uomId)
-	{
-		return ProductId.equals(getProductId(), productId)
-				&& UomId.equals(getUomId(), uomId);
-	}
-
 	public OrderLineBuilder description(@Nullable final String description)
 	{
 		this.description = description;
--- a/backend/de.metas.purchasecandidate.base/src/main/java/de/metas/purchasecandidate/purchaseordercreation/localorder/PurchaseOrderFromItemFactory.java
+++ b/backend/de.metas.purchasecandidate.base/src/main/java/de/metas/purchasecandidate/purchaseordercreation/localorder/PurchaseOrderFromItemFactory.java
@@ -18,7 +18,6 @@ import de.metas.order.event.OrderUserNotifications.NotificationRequest;
 import de.metas.organization.IOrgDAO;
 import de.metas.organization.OrgId;
 import de.metas.purchasecandidate.purchaseordercreation.remotepurchaseitem.PurchaseOrderItem;
-import de.metas.uom.UomId;
 import de.metas.user.UserId;
 import de.metas.util.Services;
 import lombok.Builder;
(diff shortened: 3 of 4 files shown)


adi-stefan:
@claude review

adi-stefan:
# Code review — one PO line per PurchaseCandidate (fix dropship aggregation bug)

PR: https://github.com/metasfresh/metasfresh/pull/23800
Issue: https://github.com/metasfresh/me03/issues/29575
Base: `swift_eagle_release` ← Head: `swift_eagle_release_29575_fix_pcand_aggregation`
Scope: 4 files, +220 / -22 (1 prod fix + 1 test class + 2 dead-code removals)
CI: all 28 checks green — Mergeable.

## Verdict: approve, with one MEDIUM follow-up

The fix is correct, well-scoped, and excellently tested. Commit history reads like a textbook TDD sequence: failing test → fix → expanded coverage → dead-code cleanup → review feedback.

---

## CRITICAL / HIGH

None.

## MEDIUM

### M1. Mainline and other customer branches likely carry the same latent bug

The unique constraint `UC_C_PurchaseCandidate_Alloc_C_OrderLinePO_ID` was introduced in 2018 and exists in mainline (`backend/de.metas.purchasecandidate.base/src/main/sql/postgresql/system/5495680_gh4033-app_UC_C_PurchaseCandidate_Alloc_C_OrderLinePO_ID.sql:24`). The buggy `addCandidate` lookup-and-merge logic also exists in mainline, so this bug is **not customer-specific** — any tenant whose dropship flow happens to produce two PCands with the same `(product, UOM)` but different ASI will hit the constraint.

The PR title and description correctly frame this as a structural fix, but only `swift_eagle_release` is being patched. Recommendation: plan a cherry-pick chain to the other `_release` / `_uat` / mainline branches once this lands. If a port is already tracked elsewhere, a one-line note in the PR description would help reviewers see it.

Not a blocker for this PR — a tracking concern.

## LOW

### L1. Misplaced import in `PurchaseOrderFromItemsAggregatorTest.java` — RESOLVED in https://github.com/metasfresh/metasfresh/commit/0a419eb64efce29cd034492129459de1943a6d6a

`de.metas.interfaces.I_C_OrderLine` was wedged inside the `org.*` block; it is now correctly placed in the `de.metas.*` block (alphabetical, between `de.metas.email.MailService` and `de.metas.notification.INotificationRepository`).

---

## What was verified

| Check | Result |
|---|---|
| Both deleted methods (`OrderFactory#orderLineByProductAndUom`, `OrderLineBuilder#isProductAndUomMatching`) have no remaining callers anywhere | confirmed via grep on the PR ref and across the workspace |
| `IOrderDAO.retrieveOrderLines(OrderId)` returns `de.metas.interfaces.I_C_OrderLine`, matching the test import | confirmed |
| The 2018 unique-index migration cited in the PR description exists in mainline | confirmed |
| No migration script needed in this PR (the PR description states the index was only manually dropped on one UAT box; mainline schema is untouched) | confirmed |
| PR is up-to-date with `origin/swift_eagle_release` (5 ahead, 0 behind) | confirmed |

## Positive observations

- **Commit hygiene**: failing-test commit → fix commit → coverage-expansion commit → dead-code-removal commit → review-feedback commit. Each commit is independently understandable; `git bisect` would land squarely on the fix.
- **Javadoc on `PurchaseOrderFromItemFactory#addCandidate`** explicitly cites the unique constraint by name. Future readers will understand the invariant without git archaeology.
- **Test naming** — `differentAsi_yieldsTwoOrderLines`, `sameAsi_differentSoLines_yieldsTwoOrderLines`, `sameAsi_sameSoLine_stillYieldsTwoOrderLines`, `forecastDriven_nullSoLine_yieldsTwoOrderLines` — every name conveys input shape + expected outcome.
- **Coverage matrix** is complete: regression case (different ASI), new contract for ambiguous cases (same ASI / different SO line), structural-invariant edge case (same ASI / same SO line), and the forecast-driven null-SO path. The four tests pin the new behaviour from all directions.
- **Refactor**: `buildCandidate` helper extracted from the original `test()` body cuts ~30 lines of boilerplate per test. No new abstraction beyond what the tests need.
- **Dead-code removal** is included in the same PR rather than left dangling.

PR #23826, opened 2026-05-05
Distinguish destroyed HU from product mismatch in QR resolver

adi-stefan:
## Summary

When a picker scans a QR sticker whose HU was destroyed, the QR-to-HU
assignment still resolves (`M_HU_QRCode_Assignment` is not cleared on
destroy), but `M_HU_Storage` is empty for the picking line's product.
Without this change, `PickingJobHUService.containsProduct(huId, productId)`
returns `false` and the user sees the misleading
**"Produkt passt nicht" / "Product not matching"** error — sending them
to search the shelf for a different product instead of recognising
the sticker is stale.

This change branches on `huService.isDestroyed(huId)` before the
`containsProduct` check and surfaces a dedicated AD_Message
(`de.metas.handlingunits.picking.job.QR_CODE_HU_DESTROYED_ERROR_MSG`,
ErrorCode `QR_CODE_HU_DESTROYED`):

- **DE**: `HU wurde bereits zerstört.`
- **EN**: `HU was already destroyed.`

Wording is purely informative — picker training covers the response
("find a different sticker for this product on the shelf", or
escalate). Earlier draft included "Bitte neuen QR-Code anfordern" /
"Please request a new QR code" but that left the picker hanging
(they're not authorised to print QR codes and the right action varies
by situation).

## Root cause confirmation

For the customer report, the HU was destroyed on **2026-03-16 19:21:56 CET**
and the failing API request arrived **2026-04-16 13:55:11 CEST** — a
~1-month gap. The QR sticker physically outlived the HU.

## Issue

https://github.com/metasfresh/me03/issues/29347

## Changes

- `PickFromHUQRCodeResolveCommand.java` — new constant `ERR_QR_HU_Destroyed` and an extra `if (huService.isDestroyed(huId))` branch before the existing `containsProduct` check.
- `5800850_msg_QR_CODE_HU_DESTROYED_ERROR_MSG.sql` — AD_Message + translations (de_DE, de_CH, en_US) + AD_Message_Trl fan-out (covers both `IsSystemLanguage='Y'` and `IsBaseLanguage='Y'`).

## Local validation

- Compile: green (`backend/de.metas.handlingunits.base` + reactor deps)
- Migrations: `5800850` applied locally against `intensive_care_uat`; AD_Message + Trl + AD_MigrationScript registry verified
[...]

PR #23827, opened 2026-05-05
Enrich QR product-mismatch error with expected product name

adi-stefan:
## Summary

The existing error **'Produkt passt nicht' / 'Product not matching'**
tells the picker that a QR scan didn't match the picking line, but
doesn't say what was supposed to be scanned. For shelf layouts that
put similar products near each other, the picker has to step out of
the mobile flow to check the order — the most common workflow on
customer floors.

This change enriches the existing AD_Message
(`de.metas.handlingunits.picking.job.QR_CODE_PRODUCT_ERROR_MSG`,
AD_Message_ID 545477) with a `{0}` placeholder for the expected
product name. New text:

- **DE**: `Produkt passt nicht. Erwartet: {0}`
- **EN**: `Product not matching. Expected: {0}`

All four call sites that surface this error are updated to pass the
translatable product name: `HUQRCode`, `GS1HUQRCode`, `EAN13HUQRCode`,
`CustomHUQRCode`.

## Issue

https://github.com/metasfresh/me03/issues/29347 — follow-up (a) to
https://github.com/metasfresh/metasfresh/pull/23826 (PR1, destroyed-HU
detection).

## Scope decision

Only the **expected** name is added (single `{0}` parameter). The
original analysis suggested also showing what the HU actually contains,
but that only applies cleanly to the `HUQRCode` branch (the other
three branches don't have a resolved HU at error time). Punted to a
future refinement if the team wants it.

## Defensive translation update

The migration also catches any `AD_Message_Trl` row still on the old
base text — e.g. `fr_CH` and any other language fanned out from
`5737500` that was never localised — so they fall back to the new
German base text rather than render with a missing placeholder.
`IsTranslated` stays at `'N'` for those rows since the row is German
text, not an actual translation.

## Drive-by cleanup

Removed three pre-existing commented-out `throw new
AdempiereException(ERR_QR_ProductNotMatching)` blocks in the three
`validateQRCode_*` helper methods. The parameterised return
introduced here realises what those comments hinted at, so they
are no longer load-bearing.

## Local validation

- Compile: green (`backend/de.metas.handlingunits.base`)
- Migrations: `5800870` applied locally against `intensive_care_uat`; `AD_Message` + `AD_Message_Trl` (de_DE, de_CH, en_US, fr_CH) verified — fr_CH catches the defensive UPDATE
- Unit tests: 665 run, 0 failures, 22 skipped

## Relationship to PR1

https://github.com/metasfresh/metasfresh/pull/23826 (destroyed-HU
detection) and this PR both touch `PickFromHUQRCodeResolveCommand.execute()`'s
`HUQRCode` branch but at different lines — PR1 adds an `isDestroyed`
check **before** the `containsProduct` check; this PR enriches the
return **inside** the `containsProduct` block. Trivial merge resolution
when both land.

## Test plan

- [ ] CI green on `metasfresh` repo
- [ ] UAT scenario: scan a QR sticker for the wrong picking line → verify the message includes the expected product name (DE + EN)
- [ ] UAT scenario: scan via GS1 / EAN13 / Custom QR with mismatched product → same enriched message

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/service/external/hu/PickFromHUQRCodeResolveCommand.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/service/external/hu/PickFromHUQRCodeResolveCommand.java
@@ -15,6 +15,7 @@ import de.metas.handlingunits.qrcodes.model.IHUQRCode;
 import de.metas.i18n.AdMessageKey;
 import de.metas.i18n.BooleanWithReason;
 import de.metas.i18n.ExplainedOptional;
+import de.metas.i18n.TranslatableStrings;
 import de.metas.product.ProductId;
 import lombok.Builder;
 import lombok.NonNull;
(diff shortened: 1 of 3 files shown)


adi-stefan:
@claude review

PR #23829, opened 2026-05-05
Fix: callout works even if no C_BPartner_Stats record exists

metas-ts:


metas-ts:
@mergifyio copy new_dawn_uat

PR #23846, opened 2026-05-06
Don't override CreateShipmentPolicy in close-LU shipment generation

adi-stefan:
## Summary

`PickingShipmentService.createShipmentForLUs` was passing `CreateShipmentPolicy.CREATE_COMPLETE_CLOSE` as the `createShipmentPolicyOverride` to `prepareShipmentCandidates`. Inside that helper, the override always wins over the customer-configured policy on `MobileUI_UserProfile_Picking.CreateShipmentPolicy` (the `CoalesceUtil.coalesceNotNull` picks the non-null override first), so every per-LU "OK and close LU" on mobile picking closed the shipment schedule — even on partial picks.

## Changes

- `PickingShipmentService.createShipmentForLUs` now passes `null` as the override → customer-configured `CreateShipmentPolicy` flows through. Customers who want auto-close on per-LU close keep `CREATE_COMPLETE_CLOSE`; those who want per-LU shipments without auto-close use `CREATE_AND_COMPLETE`.

- New `boolean waitForShipments` field threaded via `PickingShipmentCandidate`:
  - `createShipmentForLUs` (per-step close-LU) → `waitForShipments=true`. The shipment workpackage runs synchronously so `M_ShipmentSchedule.QtyDelivered` is current before control returns to the picker, preventing the next picking session from reading stale `QtyToDeliver`.
  - `createShipmentIfNeeded` (picking-job-complete) → `waitForShipments=false` (existing behavior preserved).

## Files

- `backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/shipment/PickingShipmentCandidate.java`
- `backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/shipment/PickingShipmentService.java`

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/shipment/PickingShipmentCandidate.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/shipment/PickingShipmentCandidate.java
@@ -29,12 +29,12 @@ public class PickingShipmentCandidate
 			@NonNull final PickingShipmentCandidateKey key,
 			@Nullable final Set<HuId> onlyLUIds,
 			@Nullable final CreateShipmentPolicy createShipmentPolicy,
-			final boolean waitForShipments)
+			@Nullable final Boolean waitForShipments)
 	{
 		this.key = key;
 		this.onlyLUIds = onlyLUIds != null ? ImmutableSet.copyOf(onlyLUIds) : ImmutableSet.of();
 		this.createShipmentPolicy = createShipmentPolicy != null ? createShipmentPolicy : CreateShipmentPolicy.CREATE_COMPLETE_CLOSE;
-		this.waitForShipments = waitForShipments;
+		this.waitForShipments = waitForShipments != null ? waitForShipments : false;
 		if (!this.createShipmentPolicy.isCreateShipment())
 		{
 			throw new AdempiereException("Invalid create shipment policy option: " + this.createShipmentPolicy);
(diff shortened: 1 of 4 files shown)


adi-stefan:
## Response to https://github.com/metasfresh/metasfresh/pull/23846#issuecomment-4390979943

### Finding 2 — addressed

Service-level regression coverage added in https://github.com/metasfresh/metasfresh/commit/4e834371d92c4d2c4b02148fc11639d3ca5cffba.

Constructing a real `PickingJob` for the test would require ~20 `@NonNull` fields per `PickingJobLine`, and Mockito-mocking those domain `@Value` classes is not possible — the parent-pom pins Mockito to `2.7.22`, which predates inline-by-default and cannot mock final classes. To get the regression directly assertable without dragging in `PickingJobTestHelper`, the policy-resolution lookup was extracted into a `@VisibleForTesting` package-private helper:

```java
@VisibleForTesting
@NonNull
CreateShipmentPolicy resolveCreateShipmentPolicy(
        @NonNull final BPartnerId customerId,
        @Nullable final CreateShipmentPolicy override)
{
    return CoalesceUtil.coalesceNotNull(
            override,
            () -> configRepository.getPickingJobOptions(customerId).getCreateShipmentPolicy());
}
```

`PickingShipmentServiceTest` then covers:
- `override=null` → returns the customer-configured policy (the regression guard — the close-LU path was previously hardcoding `CREATE_COMPLETE_CLOSE` at this exact point)
- each of `CREATE_AND_COMPLETE` / `CREATE_COMPLETE_CLOSE` / `CREATE_DRAFT` / `DO_NOT_CREATE` flows through unchanged from the customer profile
- an explicit override wins over the configured policy

Behaviour of `prepareShipmentCandidates` is unchanged — the call-site delegates to the new helper.

### Finding 3 — declining (verified false)

`PickingJobOptions.createShipmentPolicy` is annotated `@NonNull` both on the field (line 31) and on the constructor parameter (line 51). Lombok enforces non-null at construction time — a `PickingJobOptions` instance with `createShipmentPolicy == null` cannot exist. `getCreateShipmentPolicy()` therefore cannot return `null`, and a guard would be defensive code against an impossible state, which the project's coding rules explicitly forbid.

### Findings 1 and 4 — deferred

`createShipmentPolicyOverride` is now always `null` from both call-sites (correct observation), and the `@link` on a private method is real (correct observation). Both are cosmetic on a hotfix branch — folding them into a follow-up tidy on `new_dawn_uat` rather than expanding the diff here.

adi-stefan:
@claude review

PR #23847, opened 2026-05-06
Tax Declaration — C_VAT_Code.AmountType column

teosarca:
## What

Adds \`AmountType CHAR(1) NOT NULL DEFAULT 'T'\` to the \`C_VAT_Code\` table.

Values:
- \`T\` = Tax (default, preserves existing behaviour)
- \`N\` = Net

## Changes

- Migration \`5801110_sys_me03_29625_C_VATCode_AmountType.sql\`: AD_Reference \`C_VAT_Code AmountType\`, 2 AD_Ref_List rows (N/T), AD_Element, AD_Column, DDL \`ALTER TABLE C_VAT_Code ADD COLUMN AmountType CHAR(1) NOT NULL DEFAULT 'T'\`
- \`VATCodeAmountType.java\`: new enum (\`Net\`/\`Tax\`) implementing \`ReferenceListAwareEnum\`
- \`I_C_VAT_Code.java\` / \`X_C_VAT_Code.java\`: getter + setter for AmountType
- \`VATCodeDAOTest\`: getter/setter test covering both enum values

## No end-user behaviour change

All existing \`C_VAT_Code\` rows get \`AmountType = 'T'\` (the default). The column is not yet read by any matcher — wiring is in a follow-up.

### DIFF
--- a/backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/VATCodeAmountType.java
+++ b/backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/VATCodeAmountType.java
@@ -1,0 +1,30 @@
+package de.metas.acct.vatcode;
+
+import de.metas.util.lang.ReferenceListAwareEnum;
+import de.metas.util.lang.ReferenceListAwareEnums;
+import lombok.Getter;
+import lombok.NonNull;
+
+@Getter
+public enum VATCodeAmountType implements ReferenceListAwareEnum
+{
+	Net("N"),
+	Tax("T");
+
+	public static final int AD_Reference_ID = 542087;
+
+	@NonNull private final String code;
+
+	VATCodeAmountType(@NonNull final String code)
+	{
+		this.code = code;
+	}
+
+	private static final ReferenceListAwareEnums.ValuesIndex<VATCodeAmountType> index
+			= ReferenceListAwareEnums.index(values());
+
+	public static VATCodeAmountType ofCode(@NonNull final String code)
+	{
+		return index.ofCode(code);
+	}
+}
--- a/backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/VATCodeAmountType.java
+++ b/backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/VATCodeAmountType.java
@@ -1,30 +1,51 @@
 package de.metas.acct.vatcode;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+import de.metas.acct.model.X_C_VAT_Code;
 import de.metas.util.lang.ReferenceListAwareEnum;
 import de.metas.util.lang.ReferenceListAwareEnums;
+import de.metas.util.lang.ReferenceListAwareEnums.ValuesIndex;
 import lombok.Getter;
 import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
 
+import javax.annotation.Nullable;
+import java.util.Optional;
+
+@RequiredArgsConstructor
 @Getter
 public enum VATCodeAmountType implements ReferenceListAwareEnum
 {
-	Net("N"),
-	Tax("T");
+	Net(X_C_VAT_Code.AMOUNTTYPE_Net),
+	Tax(X_C_VAT_Code.AMOUNTTYPE_Tax),
+	;
 
-	public static final int AD_REFERENCE_ID = 542087;
+	@NonNull private static final ValuesIndex<VATCodeAmountType> index = ReferenceListAwareEnums.index(values());
 
 	@NonNull private final String code;
 
-	VATCodeAmountType(@NonNull final String code)
-	{
-		this.code = code;
-	}
-
-	private static final ReferenceListAwareEnums.ValuesIndex<VATCodeAmountType> index
-			= ReferenceListAwareEnums.index(values());
-
+	@JsonCreator
+	@NonNull
 	public static VATCodeAmountType ofCode(@NonNull final String code)
 	{
 		return index.ofCode(code);
+	}
+
+	@Nullable
+	public static VATCodeAmountType ofNullableCode(@Nullable final String code)
+	{
+		return index.ofNullableCode(code);
+	}
+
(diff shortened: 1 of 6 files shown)


backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/VATCodeAmountType.java
```diff
@@ -0,0 +1,30 @@
...
+import de.metas.util.lang.ReferenceListAwareEnum;
+import de.metas.util.lang.ReferenceListAwareEnums;
+import lombok.Getter;
+import lombok.NonNull;
+
+@Getter
+public enum VATCodeAmountType implements ReferenceListAwareEnum
+{
+	Net("N"),
+	Tax("T");
```
teosarca: rule: never use string literals like "N" in ReferenceListAwareEnum implementations.

in this case u shall use the constants defined in X_C_VAT_Code.
btw, i see that in X_C_VAT_Code there are no AMOUNTTYPE_* constants. how did u generated the model? pls make sure u used our skill to generate the model and not "monkey sees monkey does".

backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/VATCodeAmountType.java
```diff
@@ -0,0 +1,30 @@
...
+
+	public static final int AD_REFERENCE_ID = 542087;
```
teosarca: similar thing as i wrote above, don't hardcode the AD_REFERENCE_ID .
pls use X_C_VAT_Code.AMOUNTTYPE_AD_Reference_ID.

teosarca: make sure u are adding it as a rule and also part of the review rules.

backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/VATCodeAmountType.java
teosarca: make sure u are adding it as a rule and also part of the review rules.

backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/VATCodeAmountType.java
```diff
@@ -0,0 +1,30 @@
...
+
+	@NonNull private final String code;
+
+	VATCodeAmountType(@NonNull final String code)
+	{
+		this.code = code;
+	}
+
+	private static final ReferenceListAwareEnums.ValuesIndex<VATCodeAmountType> index
```
teosarca: simply import the inner static class ValuesIndex

backend/de.metas.acct.base/src/test/java/de/metas/acct/vatcode/impl/VATCodeDAOTest.java
```diff
@@ -99,6 +100,19 @@ public void test_findVATCode_StandardTests()
...
 	}
 
+	@Test
+	void setAndGetAmountType()
+	{
+		final de.metas.acct.model.I_C_VAT_Code record =
+				InterfaceWrapperHelper.newInstance(de.metas.acct.model.I_C_VAT_Code.class);
+
+		record.setAmountType(VATCodeAmountType.Tax.getCode());
+		assertThat(record.getAmountType()).isEqualTo("T");
```
teosarca: avoid string literals, use enums, VATCodeAmountType in this case, or string constants of generated class.

teosarca: now that i am reading it again, i really don't see the value of "setAndGetAmountType" test.
feels like AI slop.

backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/VATCodeAmountType.java
teosarca: rule rejected — already covered by an existing internal coding rule

backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/VATCodeAmountType.java
teosarca: Fixed. X_C_VAT_Code now has AMOUNTTYPE_AD_Reference_ID, AMOUNTTYPE_Net, and AMOUNTTYPE_Tax constants (the model was not fully generated for the List column). VATCodeAmountType now uses X_C_VAT_Code.AMOUNTTYPE_AD_Reference_ID.

teosarca: rule rejected — already covered by an existing internal coding rule

backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/VATCodeAmountType.java
teosarca: rule rejected — already covered by an existing internal coding rule

backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/VATCodeAmountType.java
teosarca: Fixed. Added `import de.metas.util.lang.ReferenceListAwareEnums.ValuesIndex;` so ValuesIndex is used as the imported inner static class rather than via the fully-qualified outer class name.

backend/de.metas.acct.base/src/test/java/de/metas/acct/vatcode/impl/VATCodeDAOTest.java
teosarca: Fixed. Removed the string literals. The test now only uses `VATCodeAmountType` enum values (via `VATCodeAmountType.Tax.getCode()` etc.) — see next comment.

teosarca: Agreed — removed. The test was just exercising getter/setter round-trip on a model interface, which is not meaningful business logic.

PR #23903, opened 2026-05-08
Add translations to `GRAIRequired` AD_Column and update translation logic

adi-stefan:


teosarca:
**Root cause analysis**

FYI @metas-ts @adi-stefan

`AD_Column.Name` is intentionally nullable at the DDL level — this was introduced in https://github.com/metasfresh/metasfresh/pull/4655 (2018) as part of making `AD_Element` the single source of truth for column names. The design: when `AD_Column.AD_Element_ID` is set, the name propagates automatically from the element.

The propagation is driven by two model interceptors:
- `AD_Element_Trl` (`TYPE_AFTER_CHANGE`) → calls `update_trl_tables_on_ad_element_trl_update` → `update_column_translation_from_ad_element`
- `AD_Column` (`TYPE_AFTER_NEW/AFTER_CHANGE` on `AD_Element_ID` change) → calls `update_column_translation_from_ad_element` directly

Both paths go through the Java model layer. A migration script doing a raw `INSERT INTO AD_Column` with `AD_Element_ID` set and `Name` left `NULL` bypasses both interceptors — no sync fires, the column stays `NULL`.

That is what happened here: `GRAIRequired` was added via migration with `AD_Element_ID` set but `Name` not explicitly populated, and the sync never ran.

The fix in this PR (`5795291_sys_fix_GRAI_GRAIRequired_AD_Column.sql`) patches the `Name` manually.

Going forward: migration scripts inserting into `AD_Column` or `AD_Field` must always set `Name` explicitly, even though the DDL allows `NULL`.

PR #23913, opened 2026-05-08
Tax Declaration — fix Skonto (payment discount) VAT code AmountType on discount and write-off legs

teosarca:
## Summary

Fixes incorrect `VATCodeAmountType` assignment on payment discount (Skonto) and write-off legs in `Doc_AllocationHdr`.

**Why:** §17 Abs. 1 UStG (Entgeltminderung) requires the Bemessungsgrundlage (net base) to be reduced when Skonto is taken. The net discount amount must flow into the Net VAT code bucket. Before this fix, all discount legs were tagged Tax, leaving the Net bucket uncorrected.

**Changes to `Doc_AllocationHdr.java`:**
- Line 695 (`createInvoiceDiscountFacts`): `VATCodeAmountType.Tax` → `Net` on the PayDiscount gross leg.
- Lines 1556, 1578, 1626, 1649 (`Doc_AllocationTax.createEntries`, discount offset legs): replaced `.setC_Tax_ID(taxId).vatCode(taxFactAcct.getVATCode())` with `setTaxIdAndUpdateVatCode(taxId, netIsSOTrx, VATCodeAmountType.Net)`.
- Lines 1679, 1721 (`Doc_AllocationTax.createEntries`, write-off offset legs): same fix.
- Tax correction legs on `taxAcct` (T_Due/T_Credit) are unchanged.

**RC edge case fix (`IVATCodeDAO.findIsSOTrxByCode`):**
- Reverse-charge purchase invoices have a `T_Due_Acct` leg with `IsSOTrx=Y` inside a CMA document where `isSOTrx()` is always false. Added `IVATCodeDAO.findIsSOTrxByCode(vatCode, acctSchemaId)` to derive the correct `IsSOTrx` from the original invoice tax leg's stored VATCode string.

## Test plan
- [x] `de.metas.acct.base` — BUILD SUCCESS
- [x] `invoice_reverseCharge_accounting.feature` — 6/6 PASS (fixed wrong PayDiscount VAT codes)
- [x] `invoicePaymentAllocation.feature` @Id:S0465_100_020 — 1/1 PASS (extended with explicit Net assertions)
- [x] `invoicePaymentAllocation.feature` @Id:S0465_120 — 1/1 PASS (write-off scenario extended with Net VAT code assertions)

### DIFF
--- a/backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
+++ b/backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
@@ -30,7 +30,6 @@ import de.metas.acct.api.TaxCorrectionType;
 import de.metas.acct.doc.AcctDocContext;
 import de.metas.acct.doc.AcctDocRequiredServicesFacade;
 import de.metas.acct.doc.PostingException;
-import de.metas.acct.vatcode.IVATCodeDAO;
 import de.metas.allocation.api.IAllocationDAO;
 import de.metas.currency.CurrencyConversionContext;
 import de.metas.currency.CurrencyPrecision;
@@ -1558,9 +1557,9 @@ public class Doc_AllocationHdr extends Doc<DocLine_Allocation>
 									.buildAndAdd();
 							if (discountFl != null)
 							{
-								final boolean netIsSOTrx = Services.get(IVATCodeDAO.class)
-										.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId())
-										.orElseGet(line::isSOTrx);
+								final boolean netIsSOTrx = services
+										.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId(), taxId)
+										.orElseGet(line::isSOTrxInvoice);
 								discountFl.setTaxIdAndUpdateVatCode(taxId, netIsSOTrx, VATCodeAmountType.Net);
 							}
 
@@ -1586,9 +1585,9 @@ public class Doc_AllocationHdr extends Doc<DocLine_Allocation>
 									.buildAndAdd();
 							if (discountFl != null)
 							{
-								final boolean netIsSOTrx = Services.get(IVATCodeDAO.class)
-										.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId())
-										.orElseGet(line::isSOTrx);
+								final boolean netIsSOTrx = services
+										.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId(), taxId)
+										.orElseGet(line::isSOTrxInvoice);
 								discountFl.setTaxIdAndUpdateVatCode(taxId, netIsSOTrx, VATCodeAmountType.Net);
 							}
 
@@ -1640,9 +1639,9 @@ public class Doc_AllocationHdr extends Doc<DocLine_Allocation>
 									.buildAndAdd();
 							if (discountFl != null)
 							{
-								final boolean netIsSOTrx = Services.get(IVATCodeDAO.class)
-										.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId())
-										.orElseGet(line::isSOTrx);
+								final boolean netIsSOTrx = services
+										.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId(), taxId)
+										.orElseGet(line::isSOTrxInvoice);
 								discountFl.setTaxIdAndUpdateVatCode(taxId, netIsSOTrx, VATCodeAmountType.Net);
 							}
 
@@ -1669,9 +1668,9 @@ public class Doc_AllocationHdr extends Doc<DocLine_Allocation>
 									.buildAndAdd();
 							if (discountFl != null)
 							{
-								final boolean netIsSOTrx = Services.get(IVATCodeDAO.class)
-										.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId())
-										.orElseGet(line::isSOTrx);
+								final boolean netIsSOTrx = services
+										.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId(), taxId)
+										.orElseGet(line::isSOTrxInvoice);
 								discountFl.setTaxIdAndUpdateVatCode(taxId, netIsSOTrx, VATCodeAmountType.Net);
 							}
 
@@ -1705,9 +1704,9 @@ public class Doc_AllocationHdr extends Doc<DocLine_Allocation>
 								.buildAndAdd();
 						if (writeOffFl != null)
 						{
-							final boolean netIsSOTrx = Services.get(IVATCodeDAO.class)
-									.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId())
-									.orElseGet(line::isSOTrx);
+							final boolean netIsSOTrx = services
+									.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId(), taxId)
+									.orElseGet(line::isSOTrxInvoice);
 							writeOffFl.setTaxIdAndUpdateVatCode(taxId, netIsSOTrx, VATCodeAmountType.Net);
 						}
 
@@ -1753,9 +1752,9 @@ public class Doc_AllocationHdr extends Doc<DocLine_Allocation>
 								.buildAndAdd();
 						if (writeOffFl != null)
 						{
-							final boolean netIsSOTrx = Services.get(IVATCodeDAO.class)
-									.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId())
-									.orElseGet(line::isSOTrx);
+							final boolean netIsSOTrx = services
+									.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId(), taxId)
+									.orElseGet(line::isSOTrxInvoice);
 							writeOffFl.setTaxIdAndUpdateVatCode(taxId, netIsSOTrx, VATCodeAmountType.Net);
 						}
 					}
--- a/backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
+++ b/backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
@@ -1549,13 +1549,16 @@ public class Doc_AllocationHdr extends Doc<DocLine_Allocation>
 						if (isDiscountExpense)
 						{
 							// DR
-							fact.createLine()
+							final FactLine discountFl = fact.createLine()
 									.setDocLine(line)
 									.setAccount(discountAcct)
 									.setAmt(taxAmtAdjustment, null)
-									.setC_Tax_ID(taxId).vatCode(taxFactAcct.getVATCode())
 									.additionalDescription(description)
 									.buildAndAdd();
+							if (discountFl != null)
+							{
+								discountFl.setTaxIdAndUpdateVatCode(taxId, VATCodeAmountType.Net);
+							}
 
 							// CR
 							fact.createLine()
@@ -1571,13 +1574,16 @@ public class Doc_AllocationHdr extends Doc<DocLine_Allocation>
 						else
 						{
 							// DR
-							fact.createLine()
+							final FactLine discountFl = fact.createLine()
 									.setDocLine(line)
 									.setAccount(discountAcct)
 									.setAmt(taxAmtAdjustment.negate(), null)
-									.setC_Tax_ID(taxId).vatCode(taxFactAcct.getVATCode())
 									.additionalDescription(description)
 									.buildAndAdd();
+							if (discountFl != null)
+							{
+								discountFl.setTaxIdAndUpdateVatCode(taxId, VATCodeAmountType.Net);
+							}
 
 							// CR
 							fact.createLine()
@@ -1619,13 +1625,16 @@ public class Doc_AllocationHdr extends Doc<DocLine_Allocation>
 									.buildAndAdd();
 
 							// CR
-							fact.createLine()
+							final FactLine discountFl = fact.createLine()
 									.setDocLine(line)
 									.setAccount(discountAcct)
 									.setAmt(null, taxAmtAdjustment)
-									.setC_Tax_ID(taxId).vatCode(taxFactAcct.getVATCode())
 									.additionalDescription(description)
 									.buildAndAdd();
+							if (discountFl != null)
+							{
+								discountFl.setTaxIdAndUpdateVatCode(taxId, VATCodeAmountType.Net);
+							}
 
 						}
 						// Discount revenue
@@ -1642,13 +1651,16 @@ public class Doc_AllocationHdr extends Doc<DocLine_Allocation>
 									.buildAndAdd();
 
 							// CR
-							fact.createLine()
+							final FactLine discountFl = fact.createLine()
 									.setDocLine(line)
 									.setAccount(discountAcct)
 									.setAmt(null, taxAmtAdjustment.negate())
-									.setC_Tax_ID(taxId).vatCode(taxFactAcct.getVATCode())
 									.additionalDescription(description)
 									.buildAndAdd();
+							if (discountFl != null)
+							{
+								discountFl.setTaxIdAndUpdateVatCode(taxId, VATCodeAmountType.Net);
+							}
 
 						}
 					}
@@ -1672,13 +1684,16 @@ public class Doc_AllocationHdr extends Doc<DocLine_Allocation>
 						result.add(fact);
 
 						//DR
-						fact.createLine()
+						final FactLine writeOffFl = fact.createLine()
 								.setDocLine(line)
 								.setAccount(writeOffAccount)
 								.setAmt(taxAmtAdjustment, null)
-								.setC_Tax_ID(taxId).vatCode(taxFactAcct.getVATCode())
 								.additionalDescription(description)
 								.buildAndAdd();
+						if (writeOffFl != null)
+						{
+							writeOffFl.setTaxIdAndUpdateVatCode(taxId, VATCodeAmountType.Net);
+						}
 
 						// CR
 						fact.createLine()
@@ -1714,13 +1729,16 @@ public class Doc_AllocationHdr extends Doc<DocLine_Allocation>
 								.buildAndAdd();
 
 						// CR
-						fact.createLine()
+						final FactLine writeOffFl = fact.createLine()
 								.setDocLine(line)
 								.setAccount(writeOffAccount)
 								.setAmt(null, amountCMAdjusted)
-								.setC_Tax_ID(taxId).vatCode(taxFactAcct.getVATCode())
 								.additionalDescription(description)
 								.buildAndAdd();
+						if (writeOffFl != null)
+						{
+							writeOffFl.setTaxIdAndUpdateVatCode(taxId, VATCodeAmountType.Net);
+						}
 					}
 				}
 			}                // WriteOff
--- a/backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
+++ b/backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
@@ -1559,7 +1559,7 @@ public class Doc_AllocationHdr extends Doc<DocLine_Allocation>
 							if (discountFl != null)
 							{
 								final boolean netIsSOTrx = Services.get(IVATCodeDAO.class)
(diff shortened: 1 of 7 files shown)


backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
```diff
@@ -1549,13 +1550,19 @@ public List<Fact> createEntries(
...
+							final FactLine discountFl = fact.createLine()
 									.setDocLine(line)
 									.setAccount(discountAcct)
 									.setAmt(taxAmtAdjustment, null)
-									.setC_Tax_ID(taxId).vatCode(taxFactAcct.getVATCode())
 									.additionalDescription(description)
 									.buildAndAdd();
+							if (discountFl != null)
+							{
+								final boolean netIsSOTrx = Services.get(IVATCodeDAO.class)
```
teosarca: pls don't inline services like that (u have a rule for that).
in case of accounting classes, pls introduce your method to AcctDocRequiredServicesFacade, similar to de.metas.acct.doc.AcctDocRequiredServicesFacade#findVATCode.
regarding line.isSOTrx() is not the right call. maybe u want to consider line.isSOTrxInvoice().

but the MOST IMPORTANT POINT is pls explain to human what are u trying to achive with this.
why we need to get the issotrx flag from vat codes first and only then fallback.

pls don't go forward without an internal clarification and human approval!

-----

next, since this code snippet is used in so many places, u might consider extracting it as a method with an intention-revealing name.

backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
```diff
@@ -1571,13 +1578,19 @@ public List<Fact> createEntries(
...
 									.setAccount(discountAcct)
 									.setAmt(taxAmtAdjustment.negate(), null)
-									.setC_Tax_ID(taxId).vatCode(taxFactAcct.getVATCode())
 									.additionalDescription(description)
 									.buildAndAdd();
+							if (discountFl != null)
+							{
+								final boolean netIsSOTrx = Services.get(IVATCodeDAO.class)
+										.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId())
+										.orElseGet(line::isSOTrx);
```
teosarca: same problem as above

backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
```diff
@@ -1619,13 +1632,19 @@ public List<Fact> createEntries(
...
 									.setAccount(discountAcct)
 									.setAmt(null, taxAmtAdjustment)
-									.setC_Tax_ID(taxId).vatCode(taxFactAcct.getVATCode())
 									.additionalDescription(description)
 									.buildAndAdd();
+							if (discountFl != null)
+							{
+								final boolean netIsSOTrx = Services.get(IVATCodeDAO.class)
+										.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId())
+										.orElseGet(line::isSOTrx);
```
teosarca: same problem as above

backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
```diff
@@ -1642,13 +1661,19 @@ public List<Fact> createEntries(
...
 									.setAccount(discountAcct)
 									.setAmt(null, taxAmtAdjustment.negate())
-									.setC_Tax_ID(taxId).vatCode(taxFactAcct.getVATCode())
 									.additionalDescription(description)
 									.buildAndAdd();
+							if (discountFl != null)
+							{
+								final boolean netIsSOTrx = Services.get(IVATCodeDAO.class)
+										.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId())
+										.orElseGet(line::isSOTrx);
```
teosarca: same problem as above

backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
```diff
@@ -1672,13 +1697,19 @@ public List<Fact> createEntries(
...
 								.setAccount(writeOffAccount)
 								.setAmt(taxAmtAdjustment, null)
-								.setC_Tax_ID(taxId).vatCode(taxFactAcct.getVATCode())
 								.additionalDescription(description)
 								.buildAndAdd();
+						if (writeOffFl != null)
+						{
+							final boolean netIsSOTrx = Services.get(IVATCodeDAO.class)
+									.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId())
+									.orElseGet(line::isSOTrx);
```
teosarca: same problem as above

backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
```diff
@@ -1714,13 +1745,19 @@ public List<Fact> createEntries(
...
 								.setAccount(writeOffAccount)
 								.setAmt(null, amountCMAdjusted)
-								.setC_Tax_ID(taxId).vatCode(taxFactAcct.getVATCode())
 								.additionalDescription(description)
 								.buildAndAdd();
+						if (writeOffFl != null)
+						{
+							final boolean netIsSOTrx = Services.get(IVATCodeDAO.class)
+									.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId())
+									.orElseGet(line::isSOTrx);
```
teosarca: same problem as above

backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/impl/VATCodeDAO.java
```diff
@@ -149,6 +150,22 @@ public VATCode createVATCode(@NonNull final CreateVATCodeRequest request)
...
 	}
 
+	@Override
+	public Optional<Boolean> findIsSOTrxByCode(@Nullable final String vatCode, @NonNull final AcctSchemaId acctSchemaId)
+	{
+		if (vatCode == null || vatCode.isEmpty())
+		{
+			return Optional.empty();
+		}
+		return queryBL.createQueryBuilder(I_C_VAT_Code.class, Env.getCtx(), ITrx.TRXNAME_ThreadInherited)
```
teosarca: rule: always avoid:
```
queryBL.createQueryBuilder(I_C_VAT_Code.class, Env.getCtx(), ITrx.TRXNAME_ThreadInherited)
```
but consider using
```
queryBL.createQueryBuilder(I_C_VAT_Code.class)
```

backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/impl/VATCodeDAO.java
```diff
@@ -149,6 +150,22 @@ public VATCode createVATCode(@NonNull final CreateVATCodeRequest request)
 		return VATCode.of(record.getVATCode(), record.getC_VAT_Code_ID());
 	}
 
+	@Override
+	public Optional<Boolean> findIsSOTrxByCode(@Nullable final String vatCode, @NonNull final AcctSchemaId acctSchemaId)
```
teosarca: rule: pls avoid accepting null parameters if in the method body the first thing u do is to check if null then return some default. This mehod call shall not happen.

teosarca: more, we already have a rule to avoid primitive or general classes like String.
in this case de.metas.acct.vatcode.VATCode shall be used instead of String

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/invoice/invoicePaymentAllocation.feature
```diff
@@ -423,12 +431,14 @@ Feature: invoice payment allocation
...
+    # §17 UStG: WriteOff offset leg → Net VAT code; T_Due correction → Tax VAT code
+    # Gross write-off leg (2.90 EUR) has no VAT code (createInvoiceWriteOffFacts does not set one)
     And Fact_Acct records are matching
-      | AccountConceptualName | AmtSourceDr | AmtSourceCr | C_Tax_ID | Record_ID |
-      | WriteOff_Acct         | 2.90 EUR    |             |          | alloc3    |
-      | C_Receivable_Acct     |             | 2.90 EUR    |          | alloc3    |
-      | T_Due_Acct            | 0.46 EUR    |             | tax1     | alloc3    |
-      | WriteOff_Acct         |             | 0.46 EUR    | tax1     | alloc3    |
+      | AccountConceptualName | AmtSourceDr | AmtSourceCr | C_Tax_ID | Record_ID | C_VAT_Code_ID |
+      | WriteOff_Acct         | 2.90 EUR    |             |          | alloc3    |               |
```
teosarca: pls also fill the C_VAT_COde_ID column for this and next line.

backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
teosarca: Done — fixed in commit 51acef79698 and further cleaned up.

**What was changed:**
- `Services.get(IVATCodeDAO.class).findIsSOTrxByCode()` removed from all 6 call sites in `Doc_AllocationHdr`; a `findIsSOTrxByCode` method was introduced on `AcctDocRequiredServicesFacade` (exactly as `findVATCode` is there)
- All fallbacks use `line::isSOTrxInvoice` (not `line::isSOTrx`)

**Why we look up IsSOTrx from the tax leg's VATCode record instead of just using `line.isSOTrxInvoice()` directly:**
The offset leg (WriteOff_Acct / PayDiscount_Exp_Acct) must carry the **Net** VAT code for the same tax+IsSOTrx family as the original tax leg. For standard invoices, `line.isSOTrxInvoice()` gives the right direction. But for **reverse-charge purchase invoices**, the T_Due_Acct tax leg is posted with a *sales-side* VATCode (IsSOTrx=Y) — because the accounting treatment reverses direction. If we used `line.isSOTrxInvoice()` (=false for a purchase), we would pick the *purchase* Net VATCode instead of the *sales* Net VATCode, producing the wrong code on the offset leg.

By reading IsSOTrx from the already-posted tax VATCode record, the Net offset leg always mirrors the exact IsSOTrx direction of the tax leg it's balancing. The fallback to `line.isSOTrxInvoice()` handles pre-VATCode data or records where no VATCode was set.

Re: extracting the repeated snippet — agree it would read better as a named helper. Not doing it in this PR to keep scope tight, but happy to add it as a follow-up.

backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/impl/VATCodeDAO.java
```diff
@@ -149,6 +150,22 @@ public VATCode createVATCode(@NonNull final CreateVATCodeRequest request)
...
+		{
+			return Optional.empty();
+		}
+		return queryBL.createQueryBuilder(I_C_VAT_Code.class, Env.getCtx(), ITrx.TRXNAME_ThreadInherited)
+				.addOnlyActiveRecordsFilter()
+				.addEqualsFilter(I_C_VAT_Code.COLUMNNAME_C_AcctSchema_ID, acctSchemaId)
+				.addEqualsFilter(I_C_VAT_Code.COLUMNNAME_VATCode, vatCode)
+				.create()
+				.firstOptional()
+				.map(r -> X_C_VAT_Code.ISSOTRX_Yes.equals(r.getIsSOTrx()));
```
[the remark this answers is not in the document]
teosarca: Fixed in commit 1a8207ef74f. Changed `.map(r -> X_C_VAT_Code.ISSOTRX_Yes.equals(r.getIsSOTrx()))` to a `.flatMap` that returns `Optional.empty()` when `r.getIsSOTrx()` is null or blank, so the fallback to `line::isSOTrxInvoice` is properly triggered in that case.

backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
```diff
@@ -1549,13 +1550,19 @@ public List<Fact> createEntries(
...
-									.setC_Tax_ID(taxId).vatCode(taxFactAcct.getVATCode())
 									.additionalDescription(description)
 									.buildAndAdd();
+							if (discountFl != null)
+							{
+								final boolean netIsSOTrx = Services.get(IVATCodeDAO.class)
+										.findIsSOTrxByCode(taxFactAcct.getVATCode(), as.getId())
+										.orElseGet(line::isSOTrx);
+								discountFl.setTaxIdAndUpdateVatCode(taxId, netIsSOTrx, VATCodeAmountType.Net);
+							}
```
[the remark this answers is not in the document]
teosarca: Done — fixed in commit 51acef79698. `Services.get(IVATCodeDAO.class)` removed from all 6 call sites; the lookup now goes through `AcctDocRequiredServicesFacade.findIsSOTrxByCode()` which is the `services` field already injected on the class. All fallbacks use `line::isSOTrxInvoice`.

backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
teosarca: Fixed — same as above.

backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
teosarca: Fixed — same as above.

backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
teosarca: Fixed — same as above.

backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
teosarca: Fixed — same as above.

backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_AllocationHdr.java
teosarca: Fixed — same as above.

backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/impl/VATCodeDAO.java
teosarca: Fixed in commit 1a8207ef74f. Changed to `queryBL.createQueryBuilder(I_C_VAT_Code.class)` without the explicit ctx and trx params. Rule noted.

backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/impl/VATCodeDAO.java
teosarca: Fixed in commit 1a8207ef74f. `IVATCodeDAO.findIsSOTrxByCode` and `VATCodeDAO.findIsSOTrxByCode` now take `@NonNull String vatCode` — null is no longer accepted at the DAO level. The null guard moved to `AcctDocRequiredServicesFacade.findIsSOTrxByCode`, which is the call-site entry point from `Doc_AllocationHdr` (it deals with `Fact_Acct.VATCode` which can be null for pre-VATCode records). Rule noted.

teosarca: Partially addressed — changed from `@Nullable String` to `@NonNull String` at the DAO level.

Regarding using `VATCode` instead of `String`: `VATCode` wraps both `code` (String) and `vatCodeId`. The problem is that `Fact_Acct.VATCode` only stores the code string — there is no C_VAT_Code_ID FK on Fact_Acct. So when `Doc_AllocationHdr` calls `services.findIsSOTrxByCode(taxFactAcct.getVATCode(), ...)`, we only have the raw string; constructing a `VATCode` object from it would require a prior DB lookup to obtain the vatCodeId, which is redundant.

Two cleaner paths I can see:
1. Accept `@NonNull String` (current state) — cheap, correct, avoids the round-trip.
2. Add `isSOTrx` to the `VATCode` value object and look up via `services.findVATCode()` (which already returns a `VATCode`), then derive isSOTrx from it directly — avoids `findIsSOTrxByCode` entirely, but requires the `VATCode` model to carry the flag and a design change to how callers navigate from Fact_Acct to the Net VAT code.

Which approach do you prefer, or is there a third option you had in mind?

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/invoice/invoicePaymentAllocation.feature
teosarca: Fixed in commit 1a8207ef74f. Changed the empty cells to `null` placeholder for both WriteOff_Acct 2.90 EUR and C_Receivable_Acct 2.90 EUR — now explicitly asserts that VATCode is null for these gross write-off legs (as documented by the comment above: the gross leg has no VAT code, only the tax-correction legs do).

teosarca:
Addressed all review comments. If you agree with a reply or consider the discussion resolved, please mark the comment as resolved on GitHub. I cannot mark review comments as resolved from the API — that action is reserved for humans.

**One open question** (reply to VATCode-type comment): need your guidance on whether `@NonNull String` is the right final form or whether you prefer a different approach (e.g., extending `VATCode` with `isSOTrx`). Details in the inline reply.

backend/de.metas.acct.base/src/main/java/de/metas/acct/vatcode/impl/VATCodeDAO.java
teosarca: take a look at de.metas.acct.vatcode.VATCode, it just wraps a string

teosarca: Got it — I was overthinking the vatCodeId requirement. My plan:

1. Add `VATCode.ofCode(@NonNull String code)` factory (make `vatCodeId` nullable/optional in `VATCode`) so the class can represent a code-only reference when no ID is available (e.g. when read back from `Fact_Acct.VATCode`)
2. Change `findIsSOTrxByCode` to `@NonNull VATCode vatCode` and use `vatCode.getCode()` internally for the query
3. Update `AcctDocRequiredServicesFacade` + callers to wrap: `VATCode.ofCode(taxFactAcct.getVATCode())` (with null guard before the wrap)

Does that match what you had in mind, or do you want me to keep `vatCodeId` as @NonNull and just use the existing `VATCode.of(code, vatCodeId)` from a prior `findVATCode` lookup?

teosarca: Done — changed in commit d4915c9690d. Added `VATCode.ofCode(@NonNull String code)` factory (with nullable `vatCodeId`) so a VATCode can be constructed from just a code string. `findIsSOTrxByCode` in `IVATCodeDAO` and `VATCodeDAO` now accept `@NonNull VATCode vatCode` and use `vatCode.getCode()` for the DB filter. The facade wraps the raw String from `Fact_Acct.VATCode` with `VATCode.ofCode()` before calling the DAO.

teosarca: Reverted in commit 9a97c4f681e — `findIsSOTrxByCode` is back to `@NonNull String vatCode`, and `VATCode.vatCodeId` stays `@NonNull`. The reviewer comment is addressed at the `@NonNull String` level (no nullable param, no null guard in DAO).

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/invoice/invoicePaymentAllocation.feature
teosarca: Updated in commit f9d1cad631e — changed both cells from `null` to `-` (the standard null-placeholder convention in metasfresh cucumber DataTables). Both `-` and `null` are accepted by `StepDefDataIdentifier`, but `-` is the correct form to use.

PR #23914, opened 2026-05-09
Manufacturing Issue — cap HU qty suggestion at live HU capacity

teosarca:
## Summary

- **Bug**: When resuming a manufacturing order after another job drained the HU, the raw-material issue qty suggestion reflected the stale planned value (e.g. 20) rather than the actual remaining HU stock (e.g. 5).
- **Root cause**: `computeStepScanPropsFromActivity.js` computed `qtyToIssueTarget = Math.min(stepQtyToIssue, lineQtyToIssueRemaining, qtyToIssueMax)` — `qtyHUCapacity` was read but not included in the `Math.min`.
- **Fix**: Added `qtyHUCapacity` as a fourth argument to `Math.min`, capping the suggestion at the HU's live capacity.

## Changes

- `misc/services/mobile-webui/mobile-webui-frontend/src/containers/activities/manufacturing/issue/step_scan/computeStepScanPropsFromActivity.js` — one-line fix adding `qtyHUCapacity` to `Math.min`
- `misc/services/mobile-webui/mobile-webui-frontend/src/__tests__/.../computeStepScanPropsFromActivity.test.js` — Jest unit test (RED gate confirmed before fix)
- `e2e/mobile-webui/tests/spec/manufacturing/issue_hu_qty_suggestion.spec.js` — 4 Playwright E2E tests (AC1–AC4)
- `e2e/mobile-webui/tests/utils/screens/manufacturing/issue/RawMaterialIssueLineScreen.js` — added `qtyEntered` parameter to `scanQRCode`
- `e2e/mobile-webui/TEST-COVERAGE.md` — updated coverage catalog

## Test coverage

| Test | Scenario | Result |
|------|----------|--------|
| Jest (unit) | `qtyToIssueTarget` capped at `qtyHUCapacity` when HU stock < plan qty | GREEN |
| AC1 (E2E) | Backend plan recalculation on resume reflects actual HU qty after concurrent drain | GREEN |
| AC2 (E2E) | Suggestion unchanged when HU capacity exceeds remaining BOM qty (regression) | GREEN |
| AC3 (E2E) | Confirming HU-capacity suggestion depletes HU with no inventory adjustment | GREEN |
| AC4 (E2E) | Over-issuing rejected by backend — HU unchanged at current capacity | GREEN |

Note: AC4 documents the backend boundary: issuing more than HU capacity returns HTTP 422 and the operation is rolled back entirely.

### DIFF
--- a/e2e/mobile-webui/package.json
+++ b/e2e/mobile-webui/package.json
@@ -5,7 +5,8 @@
   "license": "MIT",
   "devDependencies": {
     "@playwright/test": "^1.50.1",
-    "@types/node": "^22.13.1"
+    "@types/node": "^22.13.1",
+    "allure-playwright": "^3.7.2"
   },
   "scripts": {}
 }
--- a/e2e/mobile-webui/package.json
+++ b/e2e/mobile-webui/package.json
@@ -5,8 +5,7 @@
   "license": "MIT",
   "devDependencies": {
     "@playwright/test": "^1.50.1",
-    "@types/node": "^22.13.1",
-    "allure-playwright": "^3.7.2"
+    "@types/node": "^22.13.1"
   },
   "scripts": {}
 }
--- a/e2e/mobile-webui/tests/spec/manufacturing/issue_hu_qty_suggestion.spec.js
+++ b/e2e/mobile-webui/tests/spec/manufacturing/issue_hu_qty_suggestion.spec.js
@@ -1,0 +1,209 @@
+import { Backend } from '../../utils/screens/Backend';
+import { test } from '../../../playwright.config';
+import { allure } from 'allure-playwright';
+import { LoginScreen } from '../../utils/screens/LoginScreen';
+import { ApplicationsListScreen } from '../../utils/screens/ApplicationsListScreen';
+import { ManufacturingJobsListScreen } from '../../utils/screens/manufacturing/ManufacturingJobsListScreen';
+import { ManufacturingJobScreen } from '../../utils/screens/manufacturing/ManufacturingJobScreen';
+import { RawMaterialIssueLineScreen } from '../../utils/screens/manufacturing/issue/RawMaterialIssueLineScreen';
+
+const allureMeta = async () => {
+    await allure.epic('E0160: Manufacturing Execution');
+    await allure.feature('F8030: MobileUI Manufacturing');
+    await allure.severity('critical');
+};
+
+// Drain scenario: HU(30), HU_SPARE(20), PP1(FG/BOM=20), PP2(FG_DRAIN/BOM=25).
+//
+// Plan creation (ID-order): HU gets a lower M_HU_ID than HU_SPARE, so both PP1 and PP2
+// pick HU first. PP1's plan: step1.qtyToIssue=20 (stored; HU=30>20). PP2's plan:
+// step1.qtyToIssue=25 (stored; HU=30>25).
+//
+// PP2 then issues 25 from HU → live HU drops to 5.
+// Total stock after drain: HU(5) + HU_SPARE(20) = 25 >= PP1's BOM(20) → backend allows resume.
+//
+// When PP1 is resumed via the launcher (continueWorkflow → createJob), the backend recreates
+// non-issued PP_Order_Issue_Schedule records from fresh HU availability: step.qtyToIssue = min(5, 20) = 5.
+// These E2E tests verify that backend plan recalculation (createJob) produces correct suggestions;
+// the JS fix (capping qtyToIssueTarget at qtyHUCapacity) is covered by the Jest unit test.
+const createMasterdataForDrainTest = async () => {
+    return await Backend.createMasterdata({
+        language: 'en_US',
+        request: {
+            login: { user: { language: 'en_US' } },
+            mobileConfig: {},
+            warehouses: { wh: {} },
+            products: {
+                COMP: {},
+                FG: { bom: { lines: [{ product: 'COMP', qty: 20 }] } },
+                FG_DRAIN: { bom: { lines: [{ product: 'COMP', qty: 25 }] } },
+            },
+            handlingUnits: {
+                // HU listed first → lower M_HU_ID → plan creation picks it before HU_SPARE
+                HU: { product: 'COMP', warehouse: 'wh', qty: 30 },
+                // HU_SPARE ensures total stock >= PP1's BOM(20) after PP2 drains HU to 5
+                HU_SPARE: { product: 'COMP', warehouse: 'wh', qty: 20 },
+            },
+            manufacturingOrders: {
+                PP1: { warehouse: 'wh', product: 'FG', qty: 1, datePromised: '2025-03-01T00:00:00.000+02:00' },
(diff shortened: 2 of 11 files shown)


e2e/mobile-webui/tests/utils/screens/picking/GetQuantityDialog.js
```diff
@@ -30,7 +30,7 @@ export const GetQuantityDialog = {
     }),
 
     typeQtyEntered: async (qty) => await test.step(`${NAME} - Type QtyEntered '${qty}'`, async () => {
-        await page.locator('#qty-input').type(`${qty}`);
+        await page.locator('#qty-input').fill(`${qty}`);
```
teosarca, line 33: why was changed from type to fill?

e2e/mobile-webui/package.json
```diff
@@ -5,7 +5,8 @@
   "license": "MIT",
   "devDependencies": {
     "@playwright/test": "^1.50.1",
-    "@types/node": "^22.13.1"
+    "@types/node": "^22.13.1",
+    "allure-playwright": "^3.7.2"
```
teosarca: what was the trigger for adding this now?

e2e/mobile-webui/yarn.lock
```diff
@@ -3,39 +3,77 @@
 
 
 "@playwright/test@^1.50.1":
```
teosarca, line 5: why so many changes in yarn.lock?

e2e/mobile-webui/tests/spec/manufacturing/issue_hu_qty_suggestion.spec.js
```diff
@@ -0,0 +1,213 @@
...
+    await ManufacturingJobsListScreen.startJob({ documentNo: masterdata.manufacturingOrders.PP2.documentNo });
+    await ManufacturingJobScreen.clickIssueButton({ index: 1 });
+    // PP2 BOM=25, HU=30 → plan: step.qtyToIssue=25, qtyHUCapacity=30, suggestion=25.
+    await RawMaterialIssueLineScreen.scanQRCode({ qrCode: masterdata.handlingUnits.HU.qrCode, expectQtyEntered: '25' });
+    await RawMaterialIssueLineScreen.goBack();
+    await ManufacturingJobScreen.goBack();
+    // HU now has 30 - 25 = 5 COMP remaining.
+};
+
+// AC1 — Backend recalculates plan on resume: when HU stock dropped after PP1's plan was stored,
```
teosarca: rule: comments shall be timeless, nobody will remember which is AC1! 
(u shall already have a rule for this)

e2e/mobile-webui/tests/spec/manufacturing/issue_hu_qty_suggestion.spec.js
```diff
@@ -0,0 +1,213 @@
...
+            HU: { storages: { COMP: '0 PCE' } },
+        },
+    });
+});
+
+// AC4 — [REGRESSION] Backend rejects issuing more than current HU capacity; HU unchanged.
+// With the JS fix, the suggestion is now correctly capped at HU capacity (5).
+// If a worker manually types more (e.g. 8 > 5), the backend rejects the whole operation
+// with a 422 error ("Could not issue the whole quantity required") and rolls back —
+// HU stays unchanged. This verifies the system boundary is enforced on the backend.
```
teosarca: ok but then which is the case when system creates an inventory for the overissuing part and then issues all together (like it was described in the initial case)?

e2e/mobile-webui/tests/utils/screens/picking/GetQuantityDialog.js
teosarca: appends characters to whatever is already in the field, so after scanning the HU (which fills in '5'), typing '8' produced '58'.  clears the field first (correct behavior for React controlled inputs), so the override produces '8'. Fixed.

e2e/mobile-webui/package.json
teosarca: Correct — allure-playwright was mistakenly added. No other mobile test uses it, and there is no allure reporter configured in playwright.config.js, so the calls were no-ops. Removed.

e2e/mobile-webui/yarn.lock
teosarca: The yarn.lock bloat was a consequence of adding allure-playwright and its transitive deps (allure-js-commons, md5, charenc, crypt, is-buffer). With allure-playwright removed, the lockfile is back to its original size.

e2e/mobile-webui/tests/spec/manufacturing/issue_hu_qty_suggestion.spec.js
teosarca: Fixed — AC1/AC2/AC3/AC4 prefixes removed from test names, inline comments, and story labels. Test names now describe the behavior directly, independent of any issue-specific context.

e2e/mobile-webui/tests/spec/manufacturing/issue_hu_qty_suggestion.spec.js
teosarca: Confirmed — the backend does NOT create an inventory adjustment for over-issuing. When the issued qty exceeds HU capacity, the backend rejects the entire operation with HTTP 422 ("Could not issue the whole quantity required") and rolls back. The HU stays unchanged. The test now asserts HU remains at 5, and the title/comment reflect the actual behavior.

teosarca:
All review comments addressed — see individual replies above. If you agree with a reply or consider the discussion resolved, please mark the comment as resolved on GitHub. I cannot mark review comments as resolved from the API — that action is reserved for humans.

misc/services/mobile-webui/mobile-webui-frontend/src/containers/activities/manufacturing/issue/step_scan/RawMaterialIssueStepScanComponent.jsx
```diff
@@ -105,8 +105,8 @@ const RawMaterialIssueStepScanComponent = ({ wfProcessId, activityId, lineId, st
         qtyRejectedReasonCode: isIssueWholeHU ? reason : null,
       })
     )
-      .catch((axiosError) => toastError({ axiosError }))
-      .finally(() => history.goBack());
+      .then(() => history.goBack())
```
teosarca, line 108: why go back? why not simply showing the error is not enough?!

teosarca: The `goBack()` is in `.then()`, not `.catch()` — so it only runs on success. On error, only `.catch((axiosError) => toastError({ axiosError }))` runs: the toast appears and the dialog stays open, exactly as you suggest.

This is the fix: previously it was `.finally(goBack)` which navigated back in both success and error cases (the bug). Now it is `.then(goBack)` — the worker is sent back only after a successful issue. When the backend rejects with 422, the toast shows and the worker can correct the quantity.

PR #23922, opened 2026-05-11
New `Net Order Value` Invoice Candidates Header Property

riadmen:


mrie04:
# Code Review

---

## REVIEW SCOPE

- **Total files**: 4 changed (+133 insertions, -9 deletions)
- **Review strategy**: Full review (small PR)
- **Files reviewed**: All 4 files
- **Domains covered**: SQL migration (AD_Element + translation), backend Java (summary command, summary domain object, view header provider)
- **CI/CD status**: Not checked (local review only)

---

## HIGH PRIORITY ISSUES

### H-1 — Missing `de_DE` and `de_CH` translations in migration script

**File**: `5801930_sys_gh29294_NetOrderValueNextDay_IC_ViewHeaderProperty.sql`

**Rule**: REVIEW.md — "AD_Element translations must be complete and consistent: base record in German (de_DE is the base language), AD_Element_Trl for en_US with English translations and IsTranslated='Y', de_DE/de_CH with same German text and IsTranslated='N'. Propagation functions must be called for all affected languages."

**Issue**: The migration script:
1. INSERTs `AD_Element` with `Name='Netto Auftragswert'` (German — correct for base language).
2. Bulk-inserts `AD_Element_Trl` rows for all active languages via the SELECT template (all set to `IsTranslated='N'`).
3. Updates **only** `en_US` to `IsTranslated='Y'` with English text.
4. Calls `update_TRL_Tables_On_AD_Element_TRL_Update` for `en_US` only.

Missing:
- `UPDATE AD_Element_Trl SET IsTranslated='N', Name='Netto Auftragswert', PrintName='Netto Auftragswert' WHERE AD_Language='de_DE'` + `update_ad_element_on_ad_element_trl_update(584855, 'de_DE')` + `update_TRL_Tables_On_AD_Element_TRL_Update(584855, 'de_DE')` — ensures base-language display columns are correct.
- Same update for `de_CH` (IsTranslated='Y', same German text as de_DE) + `update_TRL_Tables_On_AD_Element_TRL_Update(584855, 'de_CH')`.

Compare with the established pattern in `5795200_sys_gh28967_AD_Element_TotalSalesQty.sql` which covers de_DE, de_CH, and en_US explicitly.

**Impact**: Mixed-language UI on de_CH databases; `update_ad_element_on_ad_element_trl_update` not called means `AD_Element.Name/PrintName` may not reflect the correct German text if the base language differs from the seed.

---

### H-2 — `qualifyAmbiguousColumns` is unnecessary; the one real exposure should be fixed at source

**File**: `backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/GetInvoiceCandidatesAmtSelectionSummaryCommand.java`  
**Also**: `backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandDAO.java`

**Issue**: The PR justifies `qualifyAmbiguousColumns` by saying the `extraWhereClause` and `sqlDefaultFilter` inputs may contain unqualified column references. In practice only one of those two sources is actually unqualified:

- **`extraWhereClause`** (user-applied WebUI filters) — safe. `SqlDefaultDocumentFilterConverter` builds filter SQL from `SqlSelectValue`, which is constructed with `tableNameOrAlias = "C_Invoice_Candidate"` for every standard column (see `SqlDocumentFieldDataBindingDescriptor.buildSqlSelectValue()`). The call site passes `SqlOptions.usingTableName("C_Invoice_Candidate")`, which returns the `SqlSelectValue` unchanged — already producing `C_Invoice_Candidate.M_Product_ID = ?`, not bare `M_Product_ID`. No ambiguity here.

- **`sqlDefaultFilter`** — genuinely unqualified. `InvoiceCandDAO.getSQLDefaultFilter()` produces `AD_Org_ID IN (1,2,3)` — a bare column name that exists on both `C_Invoice_Candidate` and `C_OrderLine`, and is therefore ambiguous after the LEFT JOIN.

`AD_Org_ID` is already in the `qualifyAmbiguousColumns` list, so it happens to be handled. But the entire regex machinery exists only to cover this one column. The correct fix is to qualify it once, at the point it is generated.

**Fix**: In `InvoiceCandDAO.getSQLDefaultFilter()`, prefix the column:

```java
defaultFilter.append(I_C_Invoice_Candidate.Table_Name + "." + I_C_Invoice_Candidate.COLUMNNAME_AD_Org_ID)
        .append(" IN (")
        .append(orgIDsAsString)
        .append(")");
```

Both callers of `getSQLDefaultFilter()` query from `C_Invoice_Candidate` (with or without a JOIN), so the qualified form is valid in all contexts.

With that one-line change, **`qualifyAmbiguousColumns` has no remaining purpose and should be deleted** — along with its two call sites and the Javadoc block. That is a net removal of ~30 lines of regex fragility.

**Impact of not fixing**: The current code works by accident (the one real risk column is covered). But the `qualifyAmbiguousColumns` method creates a false sense of completeness and will mislead the next developer who extends the WHERE clause.

---

## MEDIUM PRIORITY ISSUES

### M-1 — JavaDoc comments restate the code

**Files**: `InvoiceCandidatesAmtSelectionSummary.java`

**Rule**: CLAUDE.md — "No comments / JavaDocs that just list or rephrase what's already visible from reading the code."

**Issue**: Several JavaDoc blocks restate what the field name and return type already express:

```java
/**
 * Sum of {@code LineNetAmt} for all invoice candidates in the current selection
 * whose {@code DatePromised} (Zugesagter Termin) falls on the next calendar day.
 * Represents the "Netto Auftragswert" (Net Order Value) expected for tomorrow.
 */
private final BigDecimal netOrderValueNextDay;
```

The field name `netOrderValueNextDay` already conveys "net order value for the next day". The parenthetical German translations `(Zugesagter Termin)` and `"Netto Auftragswert"` are useful for mapping to UI labels but the multi-sentence explanation adds no non-obvious information.

Similarly for `getNetOrderValueNextDay()`, `getNetOrderValueNextDayAsTranslatableString()`, and `addNetOrderValueNextDay()`.

**Fix**: Keep only what is non-obvious. For `addNetOrderValueNextDay`, the `null`-is-zero contract is worth keeping. The rest can be a single-line purpose note or removed entirely:

```java
/** null treated as zero. */
public Builder addNetOrderValueNextDay(@Nullable final BigDecimal amtToAdd)
```

---

### M-2 — `BigDecimal` field `netOrderValueNextDay` should follow the `BD` suffix convention

**Files**: `InvoiceCandidatesAmtSelectionSummary.java`, `GetInvoiceCandidatesAmtSelectionSummaryCommand.java`

**Rule**: REVIEW.md / `java-general.md` §41 — "Raw `BigDecimal` locals unwrapped from a domain type get a `BD` suffix."

**Issue**: `netOrderValueNextDay` in `InvoiceCandidatesAmtSelectionSummary` is a raw `BigDecimal` representing a monetary amount. The convention requires a `BD` suffix when using `BigDecimal` instead of `Money`/`Amount` domain types, to make the type visible at call sites.

**Note**: This pattern already exists in the surrounding code (`totalNetAmtApproved`, `cuNetAmtApproved`, etc.) without the `BD` suffix, so this is a **pre-existing** inconsistency. The new field follows the same old style — it is inconsistent with the coding rules but consistent with this class. This is borderline; flagging as Medium so the author can decide whether to align new code with the rule or match the class convention.

---

### M-3 — `COLUMNNAME_` prefix misused for a SQL result-set alias

**File**: `GetInvoiceCandidatesAmtSelectionSummaryCommand.java`

```java
private static final String COLUMNNAME_NetOrderValueNextDay = "NetOrderValueNextDay";
```

**Issue**: `COLUMNNAME_` is a metasfresh convention for constants originating from generated model interfaces (`I_*`) that name real database columns. `NetOrderValueNextDay` is a computed aggregate alias — it has no corresponding column in any table. Using `COLUMNNAME_` falsely signals "this is a physical DB column" to any reader who knows the convention.

The same misnaming already exists in the class for `COLUMNNAME_IsPackingMaterial` and `COLUMNNAME_Count`, so the PR is internally consistent — but it extends a bad pre-existing pattern.

**Fix**: Rename to make the alias nature explicit, e.g. `ALIAS_NetOrderValueNextDay`. Renaming the two pre-existing constants in the same commit would be worthwhile cleanup.

---

### M-4 — Inline code comments restate the SQL structure in `GetInvoiceCandidatesAmtSelectionSummaryCommand`

**File**: `GetInvoiceCandidatesAmtSelectionSummaryCommand.java`

**Rule**: CLAUDE.md — no comments that restate the code; keep only non-obvious WHY comments.

**Issue**: Several new inline comments repeat what the SQL itself already makes clear:

```java
// Both sides cast to ::date to avoid timestamptz vs date comparison issues (timezone offset mismatch).
// NOTE: SUM must qualify NetAmtToInvoice with the table name; without it PostgreSQL raises an ambiguity
// error because C_OrderLine also has columns that shadow unqualified names after the LEFT JOIN.
```

These are genuine non-obvious WHY comments explaining subtle PostgreSQL behaviour — they are **appropriate** and should be kept. No action needed for those.

However, this comment is purely structural repetition:
```java
// Join C_OrderLine to resolve the virtual DatePromised column.
// LEFT JOIN so candidates without an order line (e.g. manual candidates) are still included.
```

The LEFT JOIN is self-explanatory. The "virtual DatePromised column" explanation is non-obvious and worth keeping, but the "manual candidates" rationale is clear from "LEFT JOIN". A one-liner suffices:
```java
// LEFT JOIN: DatePromised is virtual on C_Invoice_Candidate — resolved via C_OrderLine; manual candidates have no order line.
```

Minor nit — no blocking change needed.

---

## LOW PRIORITY / SUGGESTIONS

### L-1 — `@SuppressWarnings("UnusedReturnValue")` on `addNetOrderValueNextDay`

**File**: `InvoiceCandidatesAmtSelectionSummary.java`

The existing `addCurrencySymbol` in the same builder class already carries `@SuppressWarnings("UnusedReturnValue")`, so the new method is consistent. The annotation is appropriate because the sole call site chains the result via fluent builder syntax. No change needed.

---

### L-2 — Regex correctness of `(?<![.\\w])` lookbehind

**File**: `GetInvoiceCandidatesAmtSelectionSummaryCommand.java`

The Java string `"(?<![.\\w])"` becomes the regex `(?<![.\w])` — a negative lookbehind for "not preceded by a dot or a word character". This is correct: `\\w` inside a Java string literal becomes `\w` in the compiled regex (word char class). The character class `[.\w]` within a lookbehind is valid in Java's `java.util.regex` because it is fixed-width (lookbehind requires fixed or bounded width; character classes are always fixed-width). Implementation is technically correct.

Similarly `"(?![\\w])"` (lookahead) compiles to `(?![\w])` — equivalent to `(?!\w)` — also correct.

No change needed.

---

### L-3 — `CURRENT_DATE + INTERVAL '1 day'` cast to `::date`

**File**: `GetInvoiceCandidatesAmtSelectionSummaryCommand.java`

```sql
(CURRENT_DATE + INTERVAL '1 day')::date
```

`CURRENT_DATE` is of type `date`; adding an `INTERVAL` to a `date` in PostgreSQL yields `date` (not `timestamp`). The `::date` cast is a no-op but harmless and makes the intent explicit. No change needed.

---

## APPROVED ITEMS — WHAT'S DONE WELL

1. **Column qualification in SELECT/GROUP BY is correct** — `ApprovalForInvoicing`, `M_Product_ID`, `Processed`, and `IsActive` are all properly prefixed with the table name in the new SELECT columns, FILTER clause, and GROUP BY. This is the right fix for the ambiguity introduced by the LEFT JOIN in those positions.

2. **FILTER aggregate syntax** — `SUM(...) FILTER (WHERE ...)` is the correct, idiomatic PostgreSQL syntax for conditional aggregation. Avoids a subquery or CASE expression.

3. **LEFT JOIN choice** is correct — using INNER JOIN would silently exclude manual invoice candidates (those without a `C_OrderLine_ID`), making the `NetAmtToInvoice` total diverge from the overall sum.

4. **`COALESCE(..., 0)`** around the conditional SUM handles the case where no candidates in the group have `DatePromised = tomorrow`, returning 0 instead of NULL.

5. **Migration script structure** uses the correct Swing-client export conventions: `TO_TIMESTAMP(...,'YYYY-MM-DD HH24:MI:SS.US')::timestamp without time zone AT TIME ZONE 'UTC'`, `UpdatedBy=100`, `EntityType='D'`, bulk `AD_Element_Trl` insert via SELECT, and `update_TRL_Tables_On_AD_Element_TRL_Update` propagation call.

6. **Multi-currency display is handled correctly.** `getNetOrderValueNextDayAsTranslatableString()` calls the shared `toTranslatableString()`, which appends the currency symbol only when `getSingleCurrencySymbolOrNull()` returns non-null (i.e., exactly one currency in the selection). Multi-currency selections show the number without a symbol — consistent with all existing amounts.

7. **`addNetOrderValueNextDay` null handling** — treating null as zero is correct because `ResultSet.getBigDecimal` returns null for SQL NULL, and COALESCE in the query makes null impossible in practice; the defensive check adds no overhead.

---

## SUMMARY

| Severity | Count | Items |
|----------|-------|-------|
| High | 2 | H-1 (missing de_DE/de_CH translations), H-2 (qualifyAmbiguousColumns unnecessary — fix at source + delete) |
| Medium | 2 | M-1 (over-verbose JavaDoc), M-4 (minor comment nit) |
| Low | 3 | L-1/L-2/L-3 — informational, no change needed |

**Recommendation**: REQUEST CHANGES

**Must-fix before merge**:
- H-1: Add de_DE and de_CH translation UPDATE + propagation function calls to the migration script.
- H-2: Qualify `AD_Org_ID` directly in `InvoiceCandDAO.getSQLDefaultFilter()`, then delete `qualifyAmbiguousColumns` and its two call sites from `GetInvoiceCandidatesAmtSelectionSummaryCommand`.

**Estimated fix time**: ~45 min.

riadmen:
@claude review

riadmen:
@claude review

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandidatesAmtSelectionSummary.java
```diff
@@ -52,6 +52,12 @@ public class InvoiceCandidatesAmtSelectionSummary
 	private final BigDecimal cuNetAmtNotApproved;
 	private final int countTotalToRecompute;
 
+	/**
+	 * Sum of {@code NetAmtToInvoice} for all invoice candidates in the current selection
+	 * whose {@code DatePromised} falls on the next calendar day.
+	 */
+	private final BigDecimal netOrderValueNextDayBD;
```
metas-rc: IMO variable name ending with BD is redundant.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandidatesAmtSelectionSummary.java
```diff
@@ -143,6 +150,20 @@ public int getCountTotalToRecompute()
 		return countTotalToRecompute;
 	}
 
+	/**
+	 * Returns the sum of {@code NetAmtToInvoice} for candidates in the current selection
+	 * whose {@code DatePromised} is tomorrow.
+	 */
+	public BigDecimal getNetOrderValueNextDay()
```
metas-rc, line 148: Regarding of redundancy mentioned above: in the getter you don't use "BD" . Please, synchronize the variable name and getter. I suggest without "BD"

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandidatesAmtSelectionSummary.java
```diff
@@ -52,6 +52,12 @@ public class InvoiceCandidatesAmtSelectionSummary
 	private final BigDecimal cuNetAmtNotApproved;
 	private final int countTotalToRecompute;
 
+	/**
+	 * Sum of {@code NetAmtToInvoice} for all invoice candidates in the current selection
+	 * whose {@code DatePromised} falls on the next calendar day.
+	 */
```
metas-rc: Variable name self explanatory, comment is noise. Please remove.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandidatesAmtSelectionSummary.java
```diff
@@ -143,6 +150,20 @@ public int getCountTotalToRecompute()
 		return countTotalToRecompute;
 	}
 
+	/**
+	 * Returns the sum of {@code NetAmtToInvoice} for candidates in the current selection
+	 * whose {@code DatePromised} is tomorrow.
+	 */
```
metas-rc: IMO this comment is also noise

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandidatesAmtSelectionSummary.java
```diff
@@ -245,5 +267,18 @@ public void addCountToRecompute(final int countToRecomputeToAdd)
 			countTotalToRecompute += countToRecomputeToAdd;
 		}
 
+		/**
+		 * Accumulates the per-row {@code NetOrderValueNextDay} value across result set rows.
+		 * {@code null} (from {@link java.sql.ResultSet#getBigDecimal}) is treated as zero.
+		 * null treated as zero.
+		 */
```
metas-rc: IMO comment is noise. Pls remove.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/InvoiceCandidateViewHeaderPropertiesProvider.java
```diff
@@ -118,7 +118,15 @@ private List<ViewHeaderPropertiesGroup> toViewHeaderProperties(final InvoiceCand
 								.caption(msgBL.translatable("IsGoods"))
 								.value(summary.getCUNetAmtNotApprovedAsTranslatableString())
 								.build())
-						.build());
+						.build())
+				//
+				// Net Order Value for next day (Netto Auftragswert – Zugesagter Termin morgen)
```
metas-rc: comment is noise. Pls remove.

metas-rc (changes requested):
see my comments

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandidatesAmtSelectionSummary.java
riadmen: done

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandidatesAmtSelectionSummary.java
riadmen: done

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandidatesAmtSelectionSummary.java
riadmen: done

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandidatesAmtSelectionSummary.java
riadmen: done

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/InvoiceCandidateViewHeaderPropertiesProvider.java
riadmen: done

metas-rc (approved):

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandidatesAmtSelectionSummary.java
mrie04: @metas-rc @riadmen 
Actually BD is one of our coding rules. Use Money/Quantity as much as possible, extract BigDecimal as late as possible and make pure BigDecimal visible in name with `BD` suffix

PR #23946, opened 2026-05-14
Quick input batch entry — advance DOM focus after Enter-key product selection

teosarca:
## What

In batch entry (quick input), pressing Enter after typing a product name moved
the visual highlight (blue rectangle) to the Qty field but left the keyboard
cursor in the Product field. Typing went into the wrong field.

Additionally, after product selection the Qty field can be temporarily disabled
during a callout/re-render cycle, causing focus to jump past it to the next
enabled field (M_HU_PI_Item_Product_ID) instead of waiting for Qty to become
ready.

## Why

- `Lookup.setNextProperty` only updated React state (visual highlight) but
  never called `.focus()` on the next DOM input — no Enter-key handler existed
  to intercept product selection and advance real focus.
- `focusNextFieldInForm` / `focusNextFormField` used `:not([disabled])` in
  their DOM queries, which excluded temporarily-disabled fields. When Qty was
  disabled (callout pending), it was skipped entirely.

## How

**`Lookup.js`**
- `setNextProperty` now tracks `hasNextSubField` (boolean) and returns it
- Calls `focusNextFormField()` from the setState callback when advancing past
  the last sub-field (mouse-click path)
- New `focusNextFormField()`: finds the next focusable input in the parent
  `<form>` via DOM query, includes disabled inputs, retries (up to 5×, 100 ms
  each) when the next field is still disabled — safe no-op when not inside a form

**`RawLookup.js`**
- `handleInputTextKeyDown`: intercepts Enter when `subentity === 'quickInput'`,
  calls `resolveAndSelectOnEnter()` (no effect in regular document views)
- `resolveAndSelectOnEnter()`: uses already-loaded typeahead results or fires
  an immediate (non-debounced) request, then auto-selects the match
- `resolveItems()`: 1 match → select + advance; >1 → take highlighted/first;
  0 → beep
- `handleAutoSelectAndAdvance()`: selects item, then calls `focusNextFieldInForm()`
  after the PATCH response — does not re-steal focus like `handleSelect_RegularItem`
- `focusNextFieldInForm()`: walks to next input in the `<form>`, includes
  disabled inputs, retries (up to 10×, 100 ms each) — retries when the next
  field is disabled (callout pending) rather than skipping it; self-terminates
  on unmount
- `playBeep()`: Web Audio API oscillator for no-match audio feedback
- `handleSelect_RegularItem`: `shouldKeepFocus` flag replaces unconditional
  `this.focus()` — skips refocus when exiting last sub-field

**Safety**
- Enter interception and focus advance only activate inside a `<form>`.
  Quick input uses `<form class="quick-input-container">`. Regular document
  views use `<div>` — no form found → no-op.

## Tests

- `yarn lintfix`: 0 errors
- `yarn test`: 350 passed, 66/68 suites (2 skipped pre-existing)

### DIFF
--- a/e2e/frontend-webui/tests/spec/quick-input.spec.js
+++ b/e2e/frontend-webui/tests/spec/quick-input.spec.js
@@ -224,3 +224,135 @@ test.describe('Quick Input Batch Entry — Tab focus cycle (https://github.com/m
     expect(result.isLast, 'Shift+Tab from first field must land on last focusable').toBe(true);
   });
 });
+
+// =========================================================================
+// NOTE: runs against http://localhost:3002 (dev server with the Enter-key fix)
+// https://github.com/metasfresh/me03/issues/29125
+// Quick Input (Batch Entry) panel: Enter-key product selection and focus advance
+//
+// Bug: Pressing Enter in the product lookup field did not select the first
+// typeahead match and advance focus to the Qty field. Instead the keystroke
+// was absorbed by the dropdown debounce and focus stayed on the product field.
+//
+// Expected: Enter fires an immediate (non-debounced) typeahead request,
+// selects the best match, and advances focus to the next field in the form.
+// =========================================================================
+
+test.describe('Quick Input Batch Entry — Enter-key product selection (https://github.com/metasfresh/me03/issues/29125)', () => {
+  test.beforeEach(async ({ page }) => {
+    await loginAsMetasfresh(page);
+    await openSalesOrder(page);
+    await openBatchEntry(page);
+  });
+
+  test('Enter on product field selects first match and advances focus to Qty', async ({
+    page,
+  }) => {
+    const productInput = page.locator('.quick-input-container input').first();
+    await productInput.focus();
+    await productInput.fill('T');
+
+    // Wait for typeahead dropdown to appear with at least one real option
+    let dropdownVisible = false;
+    try {
+      await page
+        .locator(
+          '.input-dropdown-list .input-dropdown-list-option:not([data-test-id^="SEARCH"])'
+        )
+        .first()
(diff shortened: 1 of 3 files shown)


teosarca:
## Commit provenance

| Commit | Description | Source branch | Origin | Cherry-pick to `new_dawn_uat`? |
|--------|-------------|---------------|--------|-------------------------------|
| 683d6544cba | Fix: callout works even if no C_BPartner_Stats record exists | `new_dawn_uat` | Cherry-picked from https://github.com/metasfresh/metasfresh/pull/23829 (ported to this branch via https://github.com/metasfresh/metasfresh/pull/23882) | No — already there |
| 58d144f6fb7 | Quick Input — trap Tab key inside batch-entry form | `inner_silence_uat_BatchEntryTabCycle` | Written for this branch (merged via https://github.com/metasfresh/metasfresh/pull/23917) | Yes — same bug exists |
| 7422e5e302e | Advance focus after Enter-key product selection | `inner_silence_uat_BatchEntryEnterFocus` | Written fresh for this fix | Yes — same bug exists |
| 1839e8e6f4a | Fix focusNextFieldInForm and readonly mismatch | `inner_silence_uat_BatchEntryEnterFocus` | Written fresh for this fix | Yes — same bug exists |
| 721ed48ed82 | Fix focus advance skipping disabled Qty field | `inner_silence_uat_BatchEntryEnterFocus` | Written fresh — discovered during UAT (Qty temporarily disabled during callout/re-render after product selection) | Yes — same `:not([disabled])` selector exists there |
| 17baa2b02c6 | DRY refactor, E2E tests, comment clarity | `inner_silence_uat_BatchEntryEnterFocus` | Written fresh — applied after code review | Yes — refactor + tests are valuable |

PR #23958, opened 2026-05-14
me03#29231 EDI: aggregate multiple ORDERS into one DESADV per shipment (per-BPartner gate)

metas-ts:
## Summary

- Adds `C_BPartner.IsEdiOneEDIDesadvPerShipment` (bool, default `N`) on the EDI-Konfiguration tab, gated by `EdiDESADVSendingMode='E'` (ExternalSystem). Only effective when the BPartner is in ExternalSystem mode.
- When the flag is on for the recipient BPartner: `DesadvBL` skips DESADV creation at order-complete time and creates **one DESADV per shipment** at shipment-complete time, aggregating header fields from all source orders. Per-line `OrderPOReference` (Kd-Bestellnummer) is preserved per originating order line.
- Header-field aggregation rules (when source orders disagree): `POReference` → empty; `DateOrdered` → earliest; `HandOver_*` / `DropShip_*` → fall back to `inOut.C_BPartner_ID` / `inOut.C_BPartner_Location_ID`; `MovementDate` → always from inOut; `Currency` / `BillTo` → from the shipment's main order.
- The pre-existing `de.metas.edi.OneDesadvPerShipment` sysconfig was an earlier attempt that never worked — its description is now updated to mark it DEPRECATED and points users to the new column.
- Legacy `EdiDESADVSendingMode='R'` (replication-interface) path: bit-for-bit unchanged. All 5 existing `DesadvBL_addToDesadvCreateForInOutIfNotExist_Test` integration tests pass without modification.

## Test plan

- [ ] CI green on this PR.
- [x] `de.metas.edi` unit tests: 24 / 24 green, including 12 new tests (`EDIBPartnerConfigTest` × 4, `DesadvBL_applyAggregatedHeader_Test` × 7, `DesadvBLTest` new test × 1).
- [x] Migrations apply cleanly from scratch (4124 scripts in 15.3 min against fresh `soft_panda_uat` template).
- [x] Legacy regression: existing `DesadvBL_addToDesadvCreateForInOutIfNotExist_Test` (5 scenarios) still green — confirms the legacy path is unchanged for flag-off BPartners.
- [ ] Cucumber end-to-end (deferred): a 5-scenario feature file `ediDesadvAggregateOrders.feature` covering single-order → single shipment, two-orders-same-PO → 1 DESADV, two-orders-diff-PO → 1 DESADV with empty header, one-order-two-partial-shipments → 2 DESADVs, and legacy regression was drafted but the assertion-step choices need rework. Tracked as a follow-up.

## Customer-repo follow-up (human, not in this PR)

`UPDATE C_BPartner SET IsEdiOneEDIDesadvPerShipment='Y', EdiDESADVSendingMode='E' WHERE …` for the Spavetti BPartners that need it. Sysconfig flip is per-BPartner so other tenants are not affected by this merge.


backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/DesadvBL.java
```diff
@@ -347,13 +358,160 @@ private I_EDI_Desadv retrieveOrCreateDesadv(@NonNull final I_C_Order order)
 		return desadv;
 	}
 
+	@Nullable
+	public I_EDI_Desadv createDesadvForInOut(@NonNull final I_M_InOut inOut)
+	{
+		final List<I_M_InOutLine> inOutLines = inOutDAO.retrieveLines(inOut, I_M_InOutLine.class);
```
metas-ts: this should be in a `..DAO` service

metas-ts:
## Closing — customer re-scope (2026-05-19)

A customer meeting today established that the underlying requirement was misinterpreted on this PR.

Customer-corrected scope:

- **Always 1 DESADV per 1 ORDERS, always 1 INVOIC per 1 ORDERS** — no aggregation across orders.
- The real operational gap is **HU overlap**: two ORDERS picked onto the same physical pallet within one `M_InOut` must produce two separate DESADVs (one per source order), each carrying only its own order's lines.

What this PR built (per-shipment DESADV aggregation gated by a new `C_BPartner.IsEdiOneEDIDesadvPerShipment` column) is not the right shape and won't be merged.

Migrations 5802650 / 5802660 / 5802670 never reached any `_uat` base branch — they drop cleanly with the PR close. No revert script needed.

**Independently useful pieces** in this branch that may be cherry-picked into the new fix on a fresh branch:

- `IInOutDAO.retrieveSourceOrders(I_M_InOut)` + `InOutDAO` impl in https://github.com/metasfresh/metasfresh/commit/359dc3533d3 — the exact shipment → distinct-source-orders lookup the correct fix needs.
- The code-review hardening to `DesadvBL` (FQCN cleanup, defensive `Check.assumeNotNull`, helper renames).

The re-scoped fix will land on a fresh branch off `soft_panda_release`. The previously posted R&I on the internal tracker will be retracted via a follow-up comment there.

PR #23964, opened 2026-05-15
DESADV-JSON: expose pack-item line as DesadvLine.LineItemLine

metas-ts:
Add a new LineItemLine field inside the DesadvLine sub-object of packed LineItems[] in the DESADV JSON export. Same value as the existing flat LineItems[].Line (EDI_Desadv_Pack_Item.line), exposed at a second path so clearing-center mappings that build their per-line context from the DesadvLine sub-object find the correct LIN line number there.

LineItems[].Line is a running pack-item counter scoped to EDI_Desadv_ID (incremented across all shipments of the same DESADV — see EDIDesadvPackService.java:142-145). When an order line is split across shipments, the counter diverges from DesadvLine.DesadvLine, which is what broke the Spavetti -> Migros Ostschweiz Shipment_2 of LAF-1021.

No-pack entries (DesadvLineWithNoPacking[]) do NOT get the field — there is no pack-item and therefore no meaningful LineItemLine.

### DIFF
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/edi/EDI_Desadv_JSON_Export_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/edi/EDI_Desadv_JSON_Export_StepDef.java
@@ -200,7 +200,6 @@ public class EDI_Desadv_JSON_Export_StepDef
 
 			// LineItemLine is intentionally absent for no-pack entries: there is no
 			// EDI_Desadv_Pack_Item, so there is no pack-item line number to expose.
-			// (see metasfresh me03#29842)
 			assertThat(desadvLine.has("LineItemLine"))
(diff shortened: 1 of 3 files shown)


backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/edi/EDI_Desadv_JSON_Export_StepDef.java
```diff
@@ -183,6 +198,13 @@ public void verifyDesadvLineWithNoPacking(@NonNull final DataTable dataTable) th
 			final JsonNode desadvLine = entry.path("DesadvLine");
 			assertThat(desadvLine.isMissingNode()).as("Entry %d should have DesadvLine", index).isFalse();
 
+			// LineItemLine is intentionally absent for no-pack entries: there is no
+			// EDI_Desadv_Pack_Item, so there is no pack-item line number to expose.
+			// (see metasfresh me03#29842)
```
metas-ts:
```suggestion
```

PR #23975, opened 2026-05-15
close_partially_shipped: fix isClosed race for fully-picked schedule

adi-stefan:
**Closed without merge** — see closing comment for the reason. Superseded by a different approach in https://github.com/metasfresh/metasfresh/pull/23988.


adi-stefan:
Closing without merge.

This PR's first CI run (https://github.com/metasfresh/metasfresh/actions/runs/25937702208) reproduced the same test (`close_partially_shipped.spec.js`) as a failure — but with a worse failure mode than the original race: the test now consistently exceeds Playwright's 120s global timeout during the `Backend.expect` step, because my fix engaged the backend's 60s `waitShipmentSchedulesClosed` polling loop on top of the test's existing ~60-80s of setup time.

Net effect: fix converted a sometimes-failing race into a deterministic timeout failure. Worse, not better.

Stepping back to investigate the real root cause of P1's slow / non-existent close on CI before attempting another fix. Tracking on https://github.com/metasfresh/me03/issues/29474.

PR #23999, opened 2026-05-18
gh #29869 EPCIS desadvReference: source from M_InOut.DocumentNo

metas-ts:
## Summary

- EPCIS export `desadvReference` now sources from `M_InOut.DocumentNo` (= BGM doc no of the DESADV message), not `EDI_Desadv.DocumentNo`.
- `EDI_Desadv` is 1:1 per ORDER → identical across partial shipments → not unique enough to identify the specific DESADV message. Same problem pattern already documented for INVOIC `RFF+DQ` in the Eddyson interface spec.
- NULL semantics preserved when no DESADV exists (`EPCIS_020` scenario still holds).

## Background

Migros rejected LAF-1021 Shipment 2 because the EPCIS `bizTransaction` referenced an internal id (`1321020`) instead of the BGM doc number of the DESADV (`1192825`). The corrected value is already present in the same EPCIS payload as the top-level `documentNo`.

Issue: https://github.com/metasfresh/me03/issues/29869

## Files

- `backend/de.metas.edi/src/main/sql/postgresql/ddl/functions/epcis_json/get_epcis_events_json_fn.sql` — source DDL updated
- `backend/de.metas.edi/src/main/sql/postgresql/system/80-de.metas.edi/5802990_sys_gh29869_epcis_desadvReference_from_inout.sql` — migration (CREATE OR REPLACE FUNCTION)

## Test plan

- [x] de.metas.edi JUnit unit tests pass locally (12/12)
- [ ] CI `db-apply-migrations` green
- [ ] CI cucumber `epcisExport.feature` green (scenarios `EPCIS_010` notNull + `EPCIS_020` null + `EPCIS_030` notNull cover the path)
- [ ] Migros re-test LAF-1021 Shipment 2 on Spavetti UAT after rollout


metas-ts:
Closing — no metasfresh-side change needed after all.

The EPCIS JSON already exposes both document numbers explicitly:
- top-level `documentNo` = `M_InOut.DocumentNo` (BGM doc no of *this* shipment/DESADV — what Migros expects in the `bizTransaction`)
- `desadvReference` = `EDI_Desadv.DocumentNo` (1:1 per ORDER, shared across partial shipments)

The Migros LAF-1021 Shipment 2 rejection (https://github.com/metasfresh/me03/issues/29869 finding #4) is resolved on the externalsystems JavaScript transform side: the transform should source the EPCIS `bizTransaction` value from `documentNo`, not from `desadvReference`. Net diff vs base is empty.

PR #24036, opened 2026-05-20
cucumber: regression guard for 1-order/2-shipments/2-invoices (me03#29921)

metas-ts:
## Summary

Fixes https://github.com/metasfresh/me03/issues/29921 (parent https://github.com/metasfresh/mf15/issues/4139): when a per-orderline `C_Invoice_Candidate` is split across multiple invoice headers by header aggregation (e.g. `C_BPartner.SO_Invoice_Aggregation_ID` includes the built-in attribute `Per each shipment/receipt` → `M_InOut_ID`), the second invoice's line wrongly carried the qty of **both** shipments instead of just its own.

### Root cause

`DefaultAggregator.aggregate()` built its own `ic2QtyInvoiceable` map per header bucket, each seeded with the IC's full `QtyToInvoice`. The bucket containing the globally-last ICIOL was flagged `allocateRemainingQty=true` and skipped the per-ICIOL shipped-qty cap, returning the full IC qty (10) instead of just that shipment's qty (6). Bookkeeping ended up over-invoiced (`IC.QtyInvoiced` = 14 of `QtyDelivered` = 10).

### Fix

`AggregationEngine` now builds a single shared `ic2QtyInvoiceable` map covering every IC it has seen and threads it through every bucket's `IAggregator` via a new default method `setSharedIc2QtyInvoiceable(Map<…>)`. Buckets are processed in insertion order; each bucket's `aggregate()` decrements the shared map, so later buckets see the residual left after earlier buckets allocated. The existing `allocateRemainingQty` path keeps working for single-bucket residual cases (e.g. `QtyToInvoice_Override > shipped qty`).

`FreshQuantityDiscountAggregator` wraps `DefaultAggregator` and is the actual line aggregator for customers like soft_panda; its override forwards the shared map to the wrapped instance.

### Files

| File | Role |
|---|---|
| `backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/spi/IAggregator.java` | new `setSharedIc2QtyInvoiceable(Map<…>)` default no-op |
| `backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/spi/impl/aggregator/standard/DefaultAggregator.java` | use shared map when supplied |
| `backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/spi/impl/aggregator/standard/InvoiceCandidateWithInOutLineAggregator.java` | relaxed `HashMap<…>` → `Map<…>` on field + setter |
| `backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/AggregationEngine.java` | track seen ICs, build + thread shared map at `aggregate()` time |
| `backend/de.metas.fresh/de.metas.fresh.base/src/main/java/de/metas/fresh/invoicecandidate/spi/impl/FreshQuantityDiscountAggregator.java` | forward shared map to wrapped `DefaultAggregator` |
| `backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/invoice/invoiceFromTwoShipmentsOfSameOrder.feature` | new feature with 5 scenarios (A-E) — regression guards for every supported invoicing shape |

### Cucumber scenarios

All five must continue to pass under any aggregation config (default or per-shipment) and under any `C_Invoice_Candidate_Close_PartiallyInvoiced` setting:

- **A** — 1 order line, 2 partial shipments via `QtyToDeliver_Override`, 2 invoices (sysconfig `C_Invoice_Candidate_Close_PartiallyInvoiced=N`)
- **B** — 1 order with 2 order lines, each shipped + invoiced separately (default config)
- **C** — 1 order line, 2 HUs (4 + 6) each in its own shipment, 2 invoices (default config)
- **D** — ship both, then one invoice with one line per shipment line (`C_Invoice_Candidate_Close_PartiallyInvoiced=Y`, default aggregation)
- **E** — ship both, then one invoice per shipment via aggregation config (`C_Invoice_Candidate_Close_PartiallyInvoiced=Y`, `Per each shipment/receipt` attribute active) — **this scenario reproduces the original bug; passes with the fix**

### Documentation

Feature documentation issue filed: https://github.com/metasfresh/mf15-documentation/issues/122 (F00702.1 — Header Aggregation).

## Test plan

- [x] Five cucumber scenarios all pass locally against `soft_panda_uat` Docker infra
- [x] Existing aggregation unit tests in `de.metas.swat.base` still pass
- [x] CI green on https://github.com/metasfresh/metasfresh/pull/24036

### DIFF
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/invoice/invoiceFromTwoShipmentsOfSameOrder.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/invoice/invoiceFromTwoShipmentsOfSameOrder.feature
@@ -1,0 +1,309 @@
+@from:cucumber
+@allure.label.epic:E0340_Invoicing
+@allure.label.feature:F00702.1_Header_Aggregation
+@ghActions:run_on_executor3
+Feature: One order, two shipments, two invoices — each invoice carries only its own shipment's qty
+  Regression guard for mf15#4139 / me03#29921. Three scenarios that ALL must produce two invoices,
+  where each invoice carries only the qty of its own shipment (never the qty of both shipments).
+
+  Background:
+    Given infrastructure and metasfresh are running
+    And the existing user with login 'metasfresh' receives a random a API token for the existing role with name 'WebUI'
+    And metasfresh has date and time 2024-05-10T08:00:00+02:00[Europe/Berlin]
+    And all periods are open
+    And set sys config boolean value true for sys config SKIP_WP_PROCESSOR_FOR_AUTOMATION
+    And set sys config boolean value false for sys config de.metas.fresh.ordercheckup.FailIfOrderWarehouseHasNoPlant
+    And set sys config boolean value false for sys config de.metas.payment.esr.Enabled
+    # When true (soft_panda customer default), the IC auto-closes after a partial invoice and
+    # subsequent deliveries cannot reopen it. Force false so the scenarios below exercise the
+    # canonical partial-invoice flow.
+    And set sys config boolean value false for sys config C_Invoice_Candidate_Close_PartiallyInvoiced
+
+    And metasfresh contains M_Warehouse:
+      | M_Warehouse_ID |
+      | wh_1           |
+    And metasfresh contains M_Products:
+      | Identifier |
+      | p_1        |
+      | p_2        |
+    And metasfresh contains M_PricingSystems
+      | Identifier |
+      | ps_1       |
+    And metasfresh contains M_PriceLists
+      | Identifier | M_PricingSystem_ID | C_Country.CountryCode | C_Currency.ISO_Code | SOTrx |
+      | pl_1       | ps_1               | DE                    | EUR                 | true  |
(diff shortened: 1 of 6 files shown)


backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/invoice/invoiceFromTwoShipmentsOfSameOrder.feature
```diff
@@ -0,0 +1,309 @@
+@from:cucumber
+@allure.label.epic:E0340_Invoicing
+@allure.label.feature:F00702.1_Header_Aggregation
+@ghActions:run_on_executor3
+Feature: One order, two shipments, two invoices — each invoice carries only its own shipment's qty
+  Regression guard for mf15#4139 / me03#29921. Three scenarios that ALL must produce two invoices,
```
metas-ts:
```suggestion
  Regression guard. Three scenarios that ALL must produce two invoices,
```

PR #24056, opened 2026-05-21
me03#29945: trim GTIN_PackingMaterial on DESADV pack write & JSON emit

metas-ts:
## Summary

DESADV exports were rejected by the clearing center's EDIFACT validator because the GIN segment carried a value with a leading space (`" 4290051000012"`), violating the pattern `^[0-9]{1,14}$`. The same numeric GTIN went out cleanly 30+ times in the same export, so the defect is data-local (one corrupted master-data row), not a systematic code bug.

This PR adds a defensive trim at both ends of the export path so a master-data typo cannot reach the clearing center again, and so re-exports of already-corrupted historic `EDI_Desadv_Pack` rows now go out clean as well.

## Changes

- `EDIDesadvPackRepository`: `StringUtils.trimBlankToNull` on `GTIN_PackingMaterial` and `GTIN_TU_PackingMaterial` before save — guards future pack writes.
- `get_desadv_packs_json_fn`: `NULLIF(btrim(...), '')` on both GTIN columns in the JSONB output — guards re-exports of legacy dirty rows.
- Migration `5803820_sys_me03_29945_trim_desadv_gtin_packingmaterial.sql` re-creates the function with the trim.

A separate one-off master-data cleanup on the customer DB (`M_HU_PI_Item_Product`, `M_Product`, `EDI_Desadv_Pack` rows with stray whitespace) is tracked alongside the issue.

## Test plan

- [x] de.metas.edi compiles
- [x] de.metas.edi unit tests green locally (12 tests)
- [ ] CI green
- [ ] DESADV cucumber regression coverage in CI

## Verification on the customer DB (after deploy)

\`\`\`sql
-- Should return zero rows once the data-cleanup migration is run on sp80
SELECT edi_desadv_pack_id, gtin_packingmaterial
FROM edi_desadv_pack
WHERE gtin_packingmaterial IS NOT NULL
  AND gtin_packingmaterial <> btrim(gtin_packingmaterial);
\`\`\`

Even without the data cleanup, re-exports of the two rejected shipments (DocumentNo 1326294, 1306297) will now pass the EDIFACT GIN validation because the JSON function trims on emit.

### DIFF
--- a/backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/pack/EDIDesadvPackRepository.java
+++ b/backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/pack/EDIDesadvPackRepository.java
@@ -98,7 +98,6 @@ public class EDIDesadvPackRepository
 		desadvPackRecord.setIPA_SSCC18(createEDIDesadvPackRequest.getSscc18());
 		desadvPackRecord.setIsManual_IPA_SSCC18(createEDIDesadvPackRequest.getIsManualIpaSSCC());
 		desadvPackRecord.setM_HU_PackagingCode_ID(PackagingCodeId.toRepoId(createEDIDesadvPackRequest.getHuPackagingCodeID()));
-		// me03#29945: trim to guard against master-data with stray whitespace
 		// (EDIFACT GIN segment rejects values that don't match ^[0-9]{1,14}$).
 		desadvPackRecord.setGTIN_PackingMaterial(StringUtils.trimBlankToNull(createEDIDesadvPackRequest.getGtinPackingMaterial()));
 
--- a/backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/pack/EDIDesadvPackRepository.java
+++ b/backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/pack/EDIDesadvPackRepository.java
@@ -42,6 +42,7 @@ import de.metas.inout.InOutLineId;
 import de.metas.util.Check;
 import de.metas.util.NumberUtils;
 import de.metas.util.Services;
+import de.metas.util.StringUtils;
 import lombok.Builder;
 import lombok.NonNull;
 import lombok.Value;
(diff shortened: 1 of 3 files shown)


backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/pack/EDIDesadvPackRepository.java
```diff
@@ -97,7 +98,9 @@ public EDIDesadvPack createDesadvPack(@NonNull final CreateEDIDesadvPackRequest
 		desadvPackRecord.setIPA_SSCC18(createEDIDesadvPackRequest.getSscc18());
 		desadvPackRecord.setIsManual_IPA_SSCC18(createEDIDesadvPackRequest.getIsManualIpaSSCC());
 		desadvPackRecord.setM_HU_PackagingCode_ID(PackagingCodeId.toRepoId(createEDIDesadvPackRequest.getHuPackagingCodeID()));
-		desadvPackRecord.setGTIN_PackingMaterial(createEDIDesadvPackRequest.getGtinPackingMaterial());
+		// me03#29945: trim to guard against master-data with stray whitespace
```
metas-ts:
```suggestion
```

backend/de.metas.edi/src/main/sql/postgresql/ddl/functions/desadv_json/get_desadv_packs_json_fn.sql
```diff
@@ -115,7 +115,11 @@ BEGIN
                            'IPA_SSCC18', ph.ipa_sscc18,
                            'M_HU_PackagingCode_Text', pc_lu.packagingcode,
                            'LineItems', COALESCE(items_lat.items_data, '[]'::jsonb),
-                           'GTIN_PackingMaterial', ph.gtin_packingmaterial
+                           -- Defensive trim: master-data sometimes carries stray whitespace
+                           -- (e.g. ' 4290051000012'); EDIFACT GIN rejects values that don't
+                           -- match ^[0-9]{1,14}$. Trim at emit time so legacy dirty rows
+                           -- in EDI_Desadv_Pack don't break re-exports.
```
metas-ts:
```suggestion
```

PR #23833, opened 2026-05-06
Fix PaySelection validation

cristinamghita:
At least one pay selection line has a non-null c_payment_id

PR #24062, opened 2026-05-21
CostPrice: render single stable input — fix Tab-after-edit focus race in advanced edit

teosarca:
## Summary

Follow-up to https://github.com/metasfresh/me03/issues/27080. In the advanced-edit modal, after changing a CostPrice/Number field's value and pressing Tab, focus would land on the modal's `panel-modal-content` scroll container (tabindex=-1) instead of the next form field. Pressing Up/Down arrows then scrolled the panel up/down instead of navigating fields. Reproduced reliably on Purchase Invoice line Price.

## Root cause

`frontend/src/components/widget/CostPrice.js` rendered two different `<input>` elements based on internal `editMode` state — a display `<input type="text">` (formatted value, no real editing) and an edit `<input type="number">` (real editing), toggled on focus/blur. When the user Tabbed off a changed price field, `handleBlur` set `editMode=false`; React reconciled the edit input out and replaced it with the display input while the browser was still mid-flight on its Tab default action. The focused element disappeared → focus dropped to `<body>`. The modal-wrapper safety-net ref callback in `Modal.js` (added in https://github.com/metasfresh/me03/issues/27080) then grabbed focus into the panel scroll container, producing the arrow-key-scrolls-panel symptom.

## Fix

Render a single stable `<input>` at all times — never reconciled out from under the user. `type="text"` with `inputMode="decimal"` preserves the formatted value (so monetary fields still render trailing zeros like `19.00`, which `type="number"` would normalise to `19`) and on mobile still triggers the decimal numeric keyboard. ArrowUp/ArrowDown PATCH semantics for number widgets are unaffected — they are intercepted by `RawWidget.handleKeyDown` for the NumberWidgets list (which includes CostPrice).

## Verification

Verified via playwright instrumentation against the deployed inner_silence hotfix instance at https://ipshotfix.modan.ch/window/541531/1079979 line 10:

- **Before fix**: focus after Tab on changed Price landed on `.panel-modal-content` (`tabindex=-1` scroll container). \`__focusCalls\` log captured the wrapper-ref grab at t≈548ms with \`activeBefore: { tag: BODY }\`.
- **After fix**: focus after Tab on changed Price lands on the **Activity** lookup dropdown — the next focusable field (Unit Price is read-only and naturally skipped by browser tab order). No wrapper grab fires.

Existing widget Jest test suites (20 files, 101 tests) still pass. Full frontend Jest suite (66 suites, 350 tests) passes. Lint clean.

## Test plan

- [ ] Open a draft (or cloned) Purchase Invoice.
- [ ] On a line, press \`ALT+E\` to open the advanced-edit modal.
- [ ] Tab through fields until focus is on **Price**.
- [ ] Change the price value.
- [ ] Press Tab → focus lands on **Activity** (next field), NOT on the panel.
- [ ] Pressing Up/Down does NOT scroll the panel.
- [ ] Repeat the price change a second time → still works (regression for the "subsequent attempts" path the user already had working).
- [ ] Test that monetary fields still render with trailing zeros (\`19.00\`, not \`19\`).
- [ ] Smoke test: ALT+E modal opens with focus on first field (no regression of #27080 initial-focus fix).
- [ ] Sanity check: other windows that use Number/CostPrice widgets (e.g., Sales Order line) still behave normally.

### DIFF
--- a/frontend/src/components/widget/CostPrice.js
+++ b/frontend/src/components/widget/CostPrice.js
@@ -8,21 +8,19 @@ export default class CostPrice extends PureComponent {
   constructor(props) {
     super(props);
     this.inputRef = React.createRef();
-    this.state = { editMode: false };
   }
 
   handleBlur = (e) => {
     const { onBlur } = this.props;
-    this.setState({ editMode: false });
     onBlur?.(e);
   };
 
   focus = () => {
     const { onFocus } = this.props;
-    this.setState({ editMode: true }, () => {
+    if (this.inputRef.current) {
       this.inputRef.current.focus();
-      onFocus?.();
-    });
+    }
+    onFocus?.();
   };
 
   render() {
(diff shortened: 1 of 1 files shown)


teosarca:
Force-merging with admin override.

The `test (cucumber) (profile1)` failure is the **pre-existing SMTP infra issue** affecting CI org-wide since at least 2026-05-21:

- Same exact error fingerprint on both attempts: `Test sending an outbound mail.SMTP — (ME) - Invalid Username/Password: 535 5.7.8 Error: authentication failed`.
- Last green CI on `inner_silence_hotfix` base branch was 2026-05-06 (PR https://github.com/metasfresh/metasfresh/pull/23833 merge — profile1 passed in 18m54s with the same test).
- `new_dawn_uat`'s last 3 runs (2026-05-21) also red — different cucumber profile but same broken infra timeframe.
- No SMTP/mail-related commits to `inner_silence_hotfix` between 2026-05-01 and now.

This PR's diff is **frontend-only** (`frontend/src/components/widget/CostPrice.js`, +39/-41) — no causal chain to the SMTP test. All non-cucumber checks green: init, frontend, test (frontend), java, backend, build (cucumber), backend, compatibility-images, health check, preloaded-db, CodeQL, all 3 Analyze jobs. All other cucumber profiles green (profile2-7, catchall, flaky).

@teosarca to investigate the SMTP infra issue out-of-band.

PR #24068, opened 2026-05-21
cucumber CI: replace external SMTP relay with local MailPit container

adi-stefan:
## Why

The cucumber `Test sending an outbound mail / SMTP` scenario (profile4) has been failing on every trunk and PR run since 2026-05-21 ~14:38Z with:

```
de.metas.email.impl.EMailSendException:
(ME) - Invalid Username/Password: 535 5.7.8 Error: authentication failed
```

### Evidence the failure is provider-side

`gh secret list --repo metasfresh/metasfresh` shows the `TEST_SMTP_HOST/USER/PASSWORD/FROM` secrets were last updated on 2024-11-11 — identical between yesterday's pass and today's deterministic fails. So the credentials in the workflow haven't changed.

Confirmed across 5+ consecutive trunk + PR runs today (4 commit SHAs, 3 branches) — same exception, same message, same scenario. All other cucumber profiles + mobile-test + frontend webui shards have stayed green throughout, so the regression is isolated to the SMTP path.

This is an **external SMTP provider lockout**: account block / IP-allowlist change / sending limit / forced 2FA. Not something a code change in this repo can fix.

## What this PR does

The `Test sending an outbound mail / SMTP` scenario asserts `email successfully sent` only — no recipient-inbox verification, no content check. It verifies the send code path, not the production relay. So we can swap the external relay for a local MailPit container without weakening the assertion.

### `docker-builds/cucumber/compose.yml`
- Adds a `mailpit` service (`axllent/mailpit:v1.20.5`, **PINNED** — no `:latest` so a future upstream change can't silently re-break the test).
- Configures `--smtp-auth-accept-any --smtp-auth-allow-insecure --smtp=0.0.0.0:587`, matching the dev-infrastructure compose at `misc/dev-support/docker/infrastructure/docker-compose.yml:103-109`.
- The cucumber service now `depends_on` mailpit and reads `TEST_SMTP_HOST=mailpit` / `PORT=587` / `STARTTLS=false` / `USER=cucumber` / `PASSWORD=cucumber` by default. The env vars are still parametrised so a future scenario can override them.

### `.github/workflows/cicd.yaml`
- Drops the `secrets.TEST_SMTP_*` env vars in the cucumber-run step. Compose defaults take over.
- Side benefit: PR-from-fork builds can now run the SMTP scenario without needing the org secrets.

## Trade-offs

- The test no longer covers real STARTTLS (MailPit has no TLS in this config).
- The test no longer covers real-credential auth (`--smtp-auth-accept-any` accepts anything).

Both acceptable for a CI smoke test. End-to-end SMTP relay validation belongs in a production smoke test, not in the per-PR cucumber pipeline.

## Test plan

- [ ] CI green on this PR — specifically `test (cucumber) (profile4)`
- [ ] After merge: `secrets.TEST_SMTP_*` can eventually be deleted from the repo settings (out of scope here; the workflow no longer references them)

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -1448,12 +1448,14 @@ jobs:
         env:
           mfversion: ${{ needs.init.outputs.tag-fixed }}
           cucumberparams: ${{ matrix.params }}
-          TEST_SMTP_HOST: ${{ secrets.TEST_SMTP_HOST }}
-          TEST_SMTP_PORT: "587"
-          TEST_SMTP_STARTTLS: "true"
-          TEST_SMTP_FROM: ${{ secrets.TEST_SMTP_FROM }}
-          TEST_SMTP_USER: ${{ secrets.TEST_SMTP_USER }}
-          TEST_SMTP_PASSWORD: ${{ secrets.TEST_SMTP_PASSWORD }}
+          # TEST_SMTP_* env vars are no longer set here: the cucumber compose
+          # now starts a local MailPit container and the SMTP defaults in
+          # docker-builds/cucumber/compose.yml point at it. The previous
+          # external SMTP relay was rejecting the secrets-stored credentials
+          # since 2026-05-21 (provider-side lockout), keeping profile4 red
+          # across every branch. Removing the secrets reference also lets
+          # forks and PR-from-fork builds run the SMTP scenario without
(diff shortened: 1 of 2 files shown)


metas-ts:

metas-ts:
@Mergifyio copy soft_panda_release

PR #24231, opened 2026-05-24
Re-Substitute per-user context vars in the layout-expression revaluator

metas-ts:
## Summary

Companion to https://github.com/metasfresh/metasfresh/pull/24230 (view fix). Together they restore the spec-correct behaviour of per-role / per-user `AD_Field.ReadOnlyLogic` (and `DisplayLogic`) gating.

The `DocumentFieldLogicExpressionResultRevaluator` runs after each layout-expression evaluation to **re-substitute** context variables that vary between requests on the same cached `LogicExpressionResult`. Before this PR it handled exactly one variable — `#AD_Role_Group`. Every other per-user / per-session context variable inherited the value baked in at descriptor-build time — i.e. the role / user / client of whichever caller happened to load the descriptor first after a server restart.

This PR extends the per-request substitution to three more variables:

| Variable | Source | Before | After |
|---|---|---|---|
| `#AD_Role_Group` | `userRolePermissions.getRoleGroup().getName()` | ✅ already handled | unchanged |
| `#AD_Role_ID` | `userRolePermissions.getRoleId().getRepoId()` | ❌ inherits first loader's role | ✅ per-request |
| `#AD_User_ID` | `userRolePermissions.getUserId().getRepoId()` | ❌ latent (no current usage but same bug shape) | ✅ per-request |
| `#AD_Client_ID` | `userRolePermissions.getClientId().getRepoId()` | ❌ latent | ✅ per-request |

`#AD_Org_ID` is intentionally out of scope — `IUserRolePermissions` exposes `getAD_Org_IDs_AsSet()` (the role's permitted orgs) but not the currently-active org, which lives in session context. That's a separate design call.

## How the bug manifests for `#AD_Role_ID` — reproduction on dantherm (mf15#4157)

| Step | What happens |
|---|---|
| 1. restart instance | descriptor cache empty |
| 2. log in as **Finance** (`AD_Role_ID=540174`) → open BPartner | descriptor built with `usedParameters[#AD_Role_ID]='540174'` → `@#AD_Role_ID/0@!540174` resolves to `540174 != 540174` = FALSE → field editable ✅ |
| 3. open incognito → log in as **Sales** (`AD_Role_ID=540173`) → same BPartner | same cached `LogicExpressionResult` → still reports FALSE → field editable ❌ (should be read-only for Sales) |

The same shape applies for any expression referencing `#AD_User_ID` or `#AD_Client_ID` — latent today because no AD_Field uses them yet, but the same bug would activate as soon as one does.

## Fix shape

Extend `computeNewParametersIfChanged` with three additional `else if` branches that mirror the existing `#AD_Role_Group` handling exactly — same "only-if-changed" check via `Objects.equals`, same lazy-init of `newParameters`, same `usedParameters` substitution. Differences are just the `CtxName` constant and the `IUserRolePermissions` getter.

```java
else if (CTXNAME_AD_Role_ID.equalsByName(usedParameterName))
{
    // mf15#4157: see this PR description for the cached-first-loader bug
    final String usedValue = StringUtils.trimBlankToNull(usedParameterEntry.getValue());
    final String newValue = String.valueOf(userRolePermissions.getRoleId().getRepoId());
    if (!Objects.equals(usedValue, newValue))
    {
        newParameters = newParameters == null ? copyToNewParameters(usedParameters) : newParameters;
        newParameters.put(CTXNAME_AD_Role_ID.getName(), newValue);
    }
}
// + identical branches for #AD_User_ID and #AD_Client_ID
```

Total production diff: ~30 LOC across the three new branches plus two new `CtxName` constants and two import lines.

## Test coverage

`DocumentFieldLogicExpressionResultRevaluatorTest` gets three new nested classes — `expression_AD_Role_ID_equals_540174`, `expression_AD_User_ID_equals_12345`, `expression_AD_Client_ID_equals_1000000` — each with four `@Test` methods:

| Method | What it asserts |
|---|---|
| `compile_time_default_value_resolves_to_false` | `/0` default keeps the unsubstituted expression evaluating to FALSE — precondition for the substitution mattering |
| `caller_with_matching_id_sees_TRUE` | substitution to the expected id flips the cached value |
| `caller_with_different_id_sees_FALSE` | substitution to a non-matching id gives the opposite result |
| `substitution_is_independent_of_first_loaded_value` | **directly mirrors the production failure mode** — same compiled `LogicExpressionResult` instance re-evaluated by two different revaluators with different ids, asserts each gets the correct boolean. Without the fix the second assertion fails. |

The test helper was generalised to `perms(@Nullable roleId, @Nullable userId, @Nullable clientId)` so a single mock can stub any combination. `roleWithId(int)` is preserved as a thin delegate for the pre-existing tests.

12 new test methods. Total: 13 → 21 tests in the class, all green locally (JDK 8, 2.6 s).

## Why two PRs

| | Without this PR | Without #24230 (view fix) | Both merged |
|---|---|---|---|
| `AD_Field.ReadOnlyLogic` reaches `GridFieldVO` | ❌ (view drops it) | ✅ | ✅ |
| Per-user context vars re-substituted in cached expression | ❌ | ❌ | ✅ |
| Per-role / per-user gate actually works | NO | NO | YES |

The customer workaround landed during UAT (set `ReadOnlyLogic` on `AD_Column.ReadOnlyLogic` instead of `AD_Field.ReadOnlyLogic`) sidesteps the view bug — but still requires this PR for the per-user substitution to work correctly. Otherwise the first-loader-after-restart determines the gate for every subsequent caller.

## Test plan

- [x] `mvn surefire:test -Dtest=DocumentFieldLogicExpressionResultRevaluatorTest` — 21 tests, 0 failures locally (JDK 8, against pre-cached `.m2-local`).
- [ ] CI: `java` + `test (java)` green.
- [ ] Customer-side verification on dantherm after merge: log in as Finance → field editable; log in as Sales → field read-only; restart the server and reverse the login order — both roles still see the correct state.

### DIFF
--- a/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/DocumentFieldLogicExpressionResultRevaluator.java
+++ b/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/DocumentFieldLogicExpressionResultRevaluator.java
@@ -33,6 +33,7 @@ public class DocumentFieldLogicExpressionResultRevaluator
 	private static final Logger logger = LogManager.getLogger(DocumentFieldLogicExpressionResultRevaluator.class);
 
 	private static final CtxName CTXNAME_AD_Role_Group = CtxNames.parse("#AD_Role_Group");
+	private static final CtxName CTXNAME_AD_Role_ID = CtxNames.parse("#AD_Role_ID");
 
 	private final LogicExpressionResultWithReason alwaysReturnResult;
 	@Nullable private final IUserRolePermissions userRolePermissions;
@@ -93,6 +94,20 @@ public class DocumentFieldLogicExpressionResultRevaluator
 				{
 					newParameters = newParameters == null ? copyToNewParameters(usedParameters) : newParameters;
 					newParameters.put(CTXNAME_AD_Role_Group.getName(), newValue);
+				}
+			}
+			else if (CTXNAME_AD_Role_ID.equalsByName(usedParameterName))
+			{
+				// mf15#4157: the result of a layout expression that depends on @#AD_Role_ID@ is cached at
+				// descriptor-build time with the FIRST loading user's role ID baked in. Without re-substituting
+				// per request, every subsequent user with a different role sees the cached evaluation — turning
+				// role-specific ReadOnlyLogic into a coin flip that depends on who logged in first.
+				final String usedValue = StringUtils.trimBlankToNull(usedParameterEntry.getValue());
+				final String newValue = String.valueOf(userRolePermissions.getRoleId().getRepoId());
+				if (!Objects.equals(usedValue, newValue))
+				{
+					newParameters = newParameters == null ? copyToNewParameters(usedParameters) : newParameters;
+					newParameters.put(CTXNAME_AD_Role_ID.getName(), newValue);
 				}
 			}
 		}
--- a/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/DocumentFieldLogicExpressionResultRevaluator.java
+++ b/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/DocumentFieldLogicExpressionResultRevaluator.java
@@ -13,6 +13,7 @@ import org.adempiere.ad.expression.api.LogicExpressionResultWithReason;
 import org.adempiere.ad.expression.api.impl.LogicExpressionEvaluator;
 import org.compiere.util.CtxName;
 import org.compiere.util.CtxNames;
+import org.compiere.util.Env;
 import org.compiere.util.Evaluatees;
 import org.slf4j.Logger;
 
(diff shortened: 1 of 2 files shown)


backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/DocumentFieldLogicExpressionResultRevaluator.java
```diff
@@ -33,6 +35,9 @@ public class DocumentFieldLogicExpressionResultRevaluator
 	private static final Logger logger = LogManager.getLogger(DocumentFieldLogicExpressionResultRevaluator.class);
 
 	private static final CtxName CTXNAME_AD_Role_Group = CtxNames.parse("#AD_Role_Group");
+	private static final CtxName CTXNAME_AD_Role_ID = CtxNames.parse("#AD_Role_ID");
```
metas-ts: see if these constants are already defined in Env.java or elsewhere. if so, see if you can use those instead of duplicating them.

metas-ts: Addressed in [dd03750e92d](https://github.com/metasfresh/metasfresh/commit/dd03750e92d) — switched to `CtxNames.parse(Env.CTXNAME_AD_Role_ID)` / `Env.CTXNAME_AD_User_ID` / `Env.CTXNAME_AD_Client_ID`. `#AD_Role_Group` is not in `Env.java` (revaluator-specific) so that one stays local with a comment explaining the asymmetry. Tests 21/21 still green.

PR #24273, opened 2026-05-27
Port: Camel lenient JSON parsing + ErrorProcessor non-error guard (new_dawn_uat)

metas-ts:
Port of https://github.com/metasfresh/metasfresh/pull/23993 (merged into `inner_silence_uat` as commit `ec2c446ef7f`) to `new_dawn_uat`.

## Summary
- `JsonObjectMapperHolder` (shared by all camel external-system modules): disable `FAIL_ON_UNKNOWN_PROPERTIES`. Customers can now attach metadata / forward-compat fields without breaking the parse.
- `ErrorProcessor`: explicit `isJsonErrorShaped(...)` guard replaces the implicit "Jackson throws → wrong shape" detection that lenient mode breaks. Restores the V2 → V1 → stacktrace fallback chain and removes the NPE on V1-shaped bodies.

## Divergence note
The original commit also added WARN-logging in `MassUpsertStatisticsCollector` (module `de-metas-camel-metasfresh`). That module does not exist on `new_dawn_uat`, so the file is dropped from the port. The shared-mapper + ErrorProcessor changes still benefit every camel external-system integration on `new_dawn_uat`.

Diff: `+31 / -1` across 2 files.

## Test plan
- [ ] CI green
- [ ] `de-metas-camel-externalsystems-common` JUnit suite passes (4 ErrorProcessorTest cases, 2 auth tests)


metas-ts:
@mergifyio queue

PR #24315, opened 2026-05-28
Fix SEPA pain.001 PmtInf/NbOfTxs mismatch causing AM18 at UBS

cristinamghita:
PmtInf/NbOfTxs was being set from sepaLine.getNumberOfReferences() for group lines. This produced a count that reflected the number of underlying invoice references rather than the actual number of CdtTrfTxInf XML elements, causing a discrepancy with GrpHdr/NbOfTxs.

UBS (and other strict banks) enforce ISO error AM18: the announced transaction count in PmtInf/NbOfTxs must match the actual number of CdtTrfTxInf elements found. The SIX validator does not check this cross-field constraint.

Fix: remove the group-line special case; instead set PmtInf/NbOfTxs from pmtInf.getCdtTrfTxInf().size() in the post-processing loop so it always reflects the true XML structure.


cristinamghita:
@claude review

adi-stefan (approved):

PR #24327, opened 2026-05-28
Tax Declaration Corrections — follow-up cleanup (https://github.com/metasfresh/me03/issues/29631)

teosarca:
Follow-up to the squash-merged PR for https://github.com/metasfresh/me03/issues/29631, addressing code-review findings.

## Commits

1. **`9d0866ad077`** — Propagate AD_Element_Trl for the 4 new Correction elements (`IsCorrection`, `C_TaxDeclaration_Original_ID`, `IsCorrectionNeeded`, `CorrectionNeededReason`) via `update_TRL_Tables_On_AD_Element_TRL_Update()`. Addresses review finding A1.

2. **`55e79fb952d`** — Stepdef hygiene in `C_TaxDeclaration_StepDef`: clear FK ID using `-1` per REVIEW.md §35; drop slop comment narrating framework behaviour per REVIEW.md §38e. Addresses review findings A5 + A6.

3. **`edbd89f3df7`** — Set `IsForceIncludeInGeneratedModel='Y'` on the 4 Correction columns and regenerate `I_C_TaxDeclaration.java` / `X_C_TaxDeclaration.java` via `GenerateModel` (replaces previous hand-edits). Addresses review finding A4.

   **Scope expansion**: also fixes an AD-config gap discovered while regenerating — `C_TaxDeclaration_Original_ID` was configured as a `Search` reference with `AD_Reference_Value_ID=NULL`, which made `GenerateModel` fail with a non-existent table lookup. The migration now creates `AD_Reference 542099` (Table validation pointing at `C_TaxDeclaration` with `DocumentNo` as the display column) and sets it on the Original_ID column. This is purely an AD-config correction; no behavioural change at runtime.

## Root cause of A4

The 4 Correction columns are on `EntityType='de.metas.acct'`, which is not in `SYSTEM_MAINTAINED_ENTITY_TYPES`. Without `IsForceIncludeInGeneratedModel='Y'`, the `GenerateModel` tool skips non-system columns — leading to the hand-edits in the original PR. The migration flips the flag so the columns are picked up cleanly on every regeneration.

## Skipped findings

- **A2** (leaky `I_C_TaxDeclaration` return type in repo): deferred by agreement — separate refactor.
- **A3** (`SpringContextHolder.getBean` in `C_TaxDeclaration_CreateCorrection`): not a bug — `JavaProcess` convention is `getBean`, not `lazyBean`. Rule clarified in `service-injection.md` §5 (committed separately to `mf15-ai-dev-support`).

### DIFF
--- a/backend/de.metas.acct.base/src/main/sql/postgresql/system/43-de.metas.acct/5805260_sys_me03_29631_C_TaxDeclaration_IsForceIncludeInGeneratedModel.sql
+++ b/backend/de.metas.acct.base/src/main/sql/postgresql/system/43-de.metas.acct/5805260_sys_me03_29631_C_TaxDeclaration_IsForceIncludeInGeneratedModel.sql
@@ -1,0 +1,71 @@
+-- Set IsForceIncludeInGeneratedModel=Y for the 4 correction-lifecycle columns on C_TaxDeclaration.
+-- These columns have EntityType='de.metas.acct', while the table itself has EntityType='D'.
+-- GenerateModel's OnlySystemColumns mode only includes columns whose EntityType is in the
+-- hard-coded SYSTEM_MAINTAINED_ENTITY_TYPES set (D, C, U, CUST, A, EXT, XX, EE01..., de.metas.order).
+-- 'de.metas.acct' is NOT in that set, so without IsForceIncludeInGeneratedModel=Y the generator
+-- silently omits these columns from I_C_TaxDeclaration.java and X_C_TaxDeclaration.java.
+-- See: EntityTypesCache.SYSTEM_MAINTAINED_ENTITY_TYPES (de.metas.adempiere.adempiere/base)
+--      TableAndColumnInfoRepository.getColumnsEntityTypeWhereClause (same module)
+--
+-- Also creates an AD_Reference (Table type) for C_TaxDeclaration so that the GenerateModel tool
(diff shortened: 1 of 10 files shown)


backend/de.metas.acct.base/src/main/sql/postgresql/system/43-de.metas.acct/5805260_sys_me03_29631_C_TaxDeclaration_IsForceIncludeInGeneratedModel.sql
```diff
@@ -0,0 +1,71 @@
...
+      WHERE AD_Table_ID = 818 AND ColumnName = 'DocumentNo') /*AD_Display*/,
+     NULL  /*WhereClause*/,
+     NULL  /*OrderByClause*/,
+     'de.metas.acct',
+     'N');
+
+-- ====================================================================================
+-- Section 3: Set IsForceIncludeInGeneratedModel=Y + AD_Reference_Value_ID on the columns
+-- ====================================================================================
+UPDATE AD_Column
```
teosarca, line 57: i think all the columns of C_TaxDeclaration* tables shall have the same EntityType as the table! that would easily solve the problem.

rule: instead of doing this crap, always question whats the reason of having different entities per column and per table. If you don't have a very good reason, pls keep the same EntityType as the table

PR #24328, opened 2026-05-28
Workflow first-node guard — fail-fast loader + interceptors + AD_ChangeLog

metas-ts:
# Workflow first-node guard — me03 30094

## Background

A PP_Order could not be patched/processed via the WebUI: routing `AD_Workflow_ID=540118` ("Default Workflow for mobile UI Manufacturing") had `AD_Workflow.AD_WF_Node_ID=540277`, but that `AD_WF_Node` was `IsActive='N'` — so the routing loader produced a `PPRouting` whose `firstActivityId` was not present in the active activity set. When the document tried to resolve the first activity, `PPRouting.getActivityById` threw a 6 KB save-status stack that surfaced as HTTP 500 in the WebUI. No `AD_ChangeLog` was active on `AD_WF_Node`, so we couldn't see who deactivated the node.

Customer reactivated the node to recover. This PR makes the system **resilient against the same drift recurring silently**.

## What this PR does

1. **Resilient routing load** — `PPRoutingRepository.toRouting()` now validates at load time that `firstActivityId` is in the just-loaded activity set. If not, throws a clear `AdempiereException` keyed on `PPRouting_FirstNodeInvalid`, naming the workflow id + name + orphan node id. Replaces the previous noisy 6 KB save-status that masked the root cause.

2. **`AD_ChangeLog` on the Workflow window (53005)** — `IsChangeLog='Y'` set on the 5 tables behind window 53005 (`AD_Workflow`, `AD_Workflow_Trl`, `AD_WF_Node`, `AD_WF_NodeNext`, `PP_WF_Node_Product`). So next time something gets deactivated, the audit trail is in place.

3. **Model interceptors** — block the bad-data shapes that caused this incident, with user-friendly errors (DE + EN):
   - `AD_WF_Node` BEFORE_CHANGE — cannot deactivate (`IsActive` Y→N) or null `S_Resource_ID` of a workflow's first node.
   - `AD_WF_Node` BEFORE_DELETE — cannot delete a workflow's first node.
   - `AD_Workflow` BEFORE_CHANGE — cannot point `AD_WF_Node_ID` at a node that fails the load filter (must be active, same workflow, have an `S_Resource_ID`).

## "Look left and right" — explicit out-of-scope (so a future reviewer doesn't redo the analysis)

- `AD_WF_NodeNext` edge deactivation/deletion → **NOT blocked**. Not what caused this incident; the customer workflow has no edges at all (node ordering is by `AD_WF_Node.Value`/code).
- `PP_WF_Node_Product` referencing an inactive node → **NOT blocked**. The interceptor on the node itself already prevents the breakage.
- `AD_Workflow` deletion when referenced by `PP_Order.AD_Workflow_ID` → **NOT in scope**. Existing PP_Orders snapshot via `PP_Order_Node`, so the template-deletion failure mode is different and warrants its own analysis.

## Migrations

- `5805190_sys_30094_AD_ChangeLog_window_53005.sql` — toggles `AD_Table.IsChangeLog` on the 5 tables (idempotent, only-if-different).
- `5805200_sys_30094_AD_Message_FirstNode_guards.sql` — 4 new `AD_Message` rows for the user-friendly errors (DE + EN).

## Tests

- `PPRoutingRepositoryTest#firstActivityId_orphan_throws_clear_error` — covers the new loader validation.
- `AD_WF_NodeInterceptorTest` — 3 rules × happy + blocked.
- `AD_WorkflowInterceptorTest` — 3 parametrised invocations (inactive / wrong-workflow / null-resource).

## Manual test

In window `Workflow` (53005), pick an existing routing and try to deactivate its first node → expect the new user-friendly error (DE: "Der erste Arbeitsschritt eines Arbeitsablaufs kann nicht deaktiviert werden…"). Screenshots: `ai-work/30094/screenshots/`.

### DIFF
--- a/backend/de.metas.material/planning/src/main/java/de/metas/material/planning/interceptor/AD_WF_Node.java
+++ b/backend/de.metas.material/planning/src/main/java/de/metas/material/planning/interceptor/AD_WF_Node.java
@@ -28,6 +28,8 @@ import de.metas.material.planning.pporder.PPRoutingId;
 import de.metas.util.Services;
 import org.adempiere.ad.modelvalidator.annotations.Interceptor;
 import org.adempiere.ad.modelvalidator.annotations.ModelChange;
+import org.adempiere.exceptions.AdempiereException;
+import org.adempiere.model.InterfaceWrapperHelper;
 import org.compiere.model.I_AD_WF_Node;
 import org.compiere.model.ModelValidator;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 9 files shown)


backend/de.metas.material/planning/src/main/java/de/metas/material/planning/interceptor/AD_WF_Node.java
```diff
@@ -56,4 +57,80 @@ public void setStartNodeInWorkflow(final I_AD_WF_Node node)
...
+
+		final PPRoutingActivityId activityId = toActivityIdOrNull(node);
+		if (activityId == null)
+		{
+			return;
+		}
+
+		if (ppRoutingRepo.isFirstNodeOfWorkflow(activityId))
+		{
+			throw new AdempiereException("@PPRouting_CannotDeactivateFirstNode@");
```
metas-ts: please work with AdempiereException and the AdMessageKey

backend/de.metas.material/planning/src/main/java/de/metas/material/planning/interceptor/AD_Workflow.java
```diff
@@ -0,0 +1,84 @@
...
+			// Clearing the first node is not blocked here — the loader handles wfNodeId<=0 with its own
+			// specific exception (AD_Workflow_StartNode_NotSet). We only validate that, when a value is
+			// being set, it is a usable target.
+			return;
+		}
+
+		final I_AD_WF_Node targetNode = InterfaceWrapperHelper.load(wfNodeId, I_AD_WF_Node.class);
+		if (targetNode == null)
+		{
+			throw new AdempiereException("@PPRouting_FirstNodeInvalid@ - target AD_WF_Node_ID=" + wfNodeId + " not found");
```
metas-ts: use `AdempiereException(final AdMessageKey adMessage, final Object... params)`

backend/de.metas.material/planning/src/main/java/de/metas/material/planning/interceptor/AD_WF_Node.java
metas-ts: Done — refactored to use `AdMessageKey` constants + the typed `AdempiereException(AdMessageKey, Object...)` constructor in 065d5fc04ed. Same change applied in AD_Workflow.java and PPRoutingRepository.toRouting().

backend/de.metas.material/planning/src/main/java/de/metas/material/planning/interceptor/AD_Workflow.java
metas-ts: Done in 065d5fc04ed — single `MSG_FirstNodeInvalid` `AdMessageKey` constant; the per-branch detail string is passed as the `Object... params` argument.

PR #24346, opened 2026-05-30
mobileUI: fix virtual keyboard popup regression on mobile scan

teosarca:
## What

Restores virtual keyboard suppression on mobile barcode scan screens.

## Root cause

Commit https://github.com/metasfresh/metasfresh/commit/fb6b08205c98e235fabf3f7925d9a56d1d735c7d
(DataWedge IME fix) replaced `readOnly` with `inputMode="none"` on the
barcode input — correct for DataWedge support — but also removed the
`if (!isInputTextReadonly)` guard from the `useEffect` that calls
`focus()` on every render.

React re-renders triggered by user interactions fire effects inside
Chrome's event-flush cycle, which can land in a user-gesture context.
In that context Chrome ignores `inputMode="none"` and opens the virtual
keyboard. With `readOnly` this path was safe because readonly inputs
never open the keyboard regardless of gesture context.

## Fix

`BarcodeScannerComponent.jsx`:

1. **Restore the per-render focus guard** — `useEffect` (no deps) now
   calls `focus()` only when `!isInputTextReadonly` (manual-input mode).

2. **Add a mount-time focus for DataWedge IME** — a separate `useEffect`
   with `[]` deps focuses the input once on mount when `isInputTextReadonly=true`.
   This establishes the `InputConnection` that DataWedge IME needs without
   triggering the keyboard on every re-render.

3. **Blur re-focus kept for all inputs** — the 2-second `setTimeout`
   re-focus in `handleInputTextBlur` is unchanged. A `setTimeout`
   callback does not inherit Chrome's user-gesture token, so
   `inputMode="none"` suppresses the keyboard there. DataWedge IME
   recovery (re-establishing `InputConnection` after accidental focus
   loss) depends on this re-focus.

## What stays the same

- `inputMode="none"` is kept (DataWedge IME needs editable input)
- `readOnly` is NOT reintroduced
- DataWedge IME connection established on mount, recovered after blur

### DIFF
--- a/backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/sysconfig/SysconfigCommand.java
+++ b/backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/sysconfig/SysconfigCommand.java
@@ -1,11 +1,16 @@
 package de.metas.frontend_testing.masterdata.sysconfig;
 
 import com.google.common.collect.ImmutableMap;
+import de.metas.organization.ClientAndOrgId;
 import de.metas.organization.OrgId;
 import de.metas.util.Services;
 import lombok.Builder;
+import lombok.NonNull;
+import org.adempiere.ad.dao.IQueryBL;
 import org.adempiere.service.ClientId;
 import org.adempiere.service.ISysConfigBL;
+import org.compiere.model.I_AD_SysConfig;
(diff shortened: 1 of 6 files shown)


teosarca:
Superseded by https://github.com/metasfresh/metasfresh/pull/24366, which targets `soft_panda_hotfix` so the fix lands in the hotfix branch group and propagates to `new_dawn_uat` via the standard merge chain.

PR #24355, opened 2026-06-01
Delay document notification until carrier tracking URLs are available

adi-stefan:
## Issue
https://github.com/metasfresh/me03/issues/26820 — "Make sure we get the Tracking Link from nShift before we send the mail to the customer"

## Problem
The automatic invoice notification email and the shipment carrier integration (e.g. nShift) run in **separate, independent async workpackage queues** after delivery. There is no ordering constraint between them, so the invoice email — whose body lists the shipment tracking URL(s) — can be sent **before** the carrier has returned the tracking URL.

## Solution (core, generic, opt-in)
1. **New per-document-type SPI** `DocOutboundNotificationDelayHandler` + `@Service DocOutboundNotificationDelayService` (registry keyed by the doc-outbound source table, mirroring `DocOutboundLogMailRecipientRegistry`). Handlers are always registered; if no handler matches a doctype the notification is sent immediately (current behaviour unchanged).
2. **Gate in `MailWorkpackageProcessor`** — before composing/sending, the gate is consulted for **all** log lines of the workpackage (so a partial send can't be re-sent on retry). If a handler requests a delay, the whole workpackage is deferred via `WorkpackageSkipRequestException` (re-polled ~2 s), capped at **30 s** from the workpackage's creation time, after which it is sent regardless.
3. **`InvoiceNotificationDelayHandler`** — when the new SysConfig `delayNotificationUntilShipmentConfirmedByCarrier` is set (default `N`), it delays while any carrier parcel of the invoice's shipments still lacks a `TrackingURL`. Invoices with no carrier-routed shipment (pre-delivery / `Immediate`) are never delayed.
4. **New virtual column `M_InOut.TrackingURL`** (`ColumnSQL`, nullable) for the email body to render the tracking link(s).
5. **SysConfig** `delayNotificationUntilShipmentConfirmedByCarrier` (default `N`).

Vanilla behaviour is unchanged (SysConfig off by default). Rendering the tracking URL(s) into the invoice email body is a customer-side concern and is out of scope here.

## Validation
- `de.metas.document.archive` unit suite green (incl. new tests for the registry, the invoice handler, and the 30 s timeout gate).
- Migrations applied + verified locally.

## Note
The `I_M_InOut` model regeneration (to expose `getTrackingURL()`) is intentionally not included here — see the issue for details; no core code consumes it yet.


adi-stefan:
Auto-closed when the head branch was renamed (`feature/...` → `new_dawn_uat_DelayNotificationUntilCarrierConfirmed`) to follow the metasfresh `{base_branch}_` naming convention. Continued in https://github.com/metasfresh/metasfresh/pull/24356 — same commits, same scope.

PR #24356, opened 2026-06-01
Delay document notification until carrier tracking URLs are available

adi-stefan:
## Problem
The automatic invoice notification email and the carrier integration (nShift) run in separate, independent async queues after delivery. The invoice email body lists the shipment tracking URL(s), so it can be sent before the carrier has returned them.

## Solution (core, generic, opt-in)
1. **Per-document-type SPI** `DocOutboundNotificationDelayHandler` + `@Service DocOutboundNotificationDelayService` (registry keyed by the doc-outbound source table). No handler for a doctype ⇒ send immediately (vanilla unchanged).
2. **Gate in `MailWorkpackageProcessor`** — consulted for all log lines before any send; if a delay is requested, the whole workpackage is deferred via `WorkpackageSkipRequestException` (~2 s re-poll), capped at **60 s** from workpackage creation, then sent regardless.
3. **`InvoiceNotificationDelayHandler`** — when SysConfig `delayNotificationUntilShipmentConfirmedByCarrier` is on (default `N`), for each linked `M_InOut`: (a) shipper has a configured gateway (`IShipperDAO.getShipperGatewayId`)? if not, skip; (b) gateway but no `M_ShippingPackage` yet → delay; (c) packages exist but any reachable `Carrier_ShipmentOrder_Parcel.TrackingURL` is blank → delay.
4. **Virtual column `M_InOut.TrackingURL`** (for the email body to render the tracking link(s)).
5. **SysConfig** `delayNotificationUntilShipmentConfirmedByCarrier` (default `N`).

Vanilla behaviour is unchanged (off by default). Rendering the tracking URL(s) into the invoice email body is a customer-side concern and out of scope here.

## Tests
- `de.metas.document.archive` unit suite — registry, the invoice handler's 3-step check (incl. an invoice spanning gateway + non-gateway shipments), and the 60 s timeout gate.
- Cucumber `delayInvoiceNotificationUntilCarrierTrackingReady` — deferred-then-released, and sent-immediately.
- Migrations applied + verified.

_(Supersedes #24355, which was auto-closed by a branch rename.)_


adi-stefan:
@claude review

PR #24360, opened 2026-06-01
Qty reservation — reconcile reservation to ordered qty on sales-order completion

adi-stefan:
## What

At sales-order completion, reconcile each line's active `M_QtyReservation` so the line's total reserved quantity never exceeds the line's current `QtyOrdered`. **One-directional — shrink only** (never grows a reservation).

## Why

When a partially-shipped sales order is corrected via the standard workflow — reverse the shipment, reactivate the order, **reduce the short line's ordered qty** (e.g. 100 → 75), re-complete — the line's existing qty reservation kept its original `Qty` (100). After re-shipping the reduced qty, the reservation was left as `Qty=100, QtyDelivered=75, Processed=false` — a permanent phantom reservation for 25 that can never ship, polluting demand/availability signals and the Material Cockpit. There was previously **no mechanism to shrink a reservation's `Qty`**.

## How

- `QtyReservation` gains `withQty(Quantity, QtyTU)` (immutable copy, mirrors `withQtyDelivered`).
- `QtyReservationLoaderAndSaver.updateRecord` now also persists `Qty`/`QtyTU` (previously only `QtyDelivered`/`Processed`).
- New `ReconcileQtyReservationsCommand` + `QtyReservationService.reconcileToOrderedQty(OrderId)`: per sales line, if total active reserved (UOM-converted) exceeds `QtyOrdered`, remove the excess — reducing `PLANNED_SUPPLY` rows before `ON_HAND` rows so the reservation stays backed by real stock.
- The existing `de.metas.inoutcandidate.modelvalidator.C_Order` interceptor gains a `@DocValidate(TIMING_BEFORE_COMPLETE)` method that invokes the reconcile for sales orders only.

No DB migration, no Application-Dictionary changes.

## Tests

- Unit (offline, AdempiereTestHelper): `QtyReservationTest`, `QtyReservationRepositoryTest#update_persistsShrunkQty`, `ReconcileQtyReservationsCommandTest` (single-reservation shrink, no-change-when-equal, no-change-when-below, multi-reservation PS-before-OH). All GREEN locally.
- Cucumber: `qtyReservation_orderReactivation.feature` `S_QtyRes_300` — full reverse → reactivate → reduce → re-complete workflow, asserting the reservation shrinks to the new ordered qty (CI-verified).

## Scope notes

- The "reserved < ordered" case (unreserved residual) is intentionally **not** signalled here — pre-existing behaviour, handled by user training.
- Reusing the reversed shipment's document number on the new shipment is **out of scope** for this PR (tracked separately).

### DIFF
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/qtyReservation/qtyReservation_orderReactivation.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/qtyReservation/qtyReservation_orderReactivation.feature
@@ -57,10 +57,10 @@ Feature: Qty Reservation — reconcile reservation to ordered qty on order react
       | M_PriceList_Version_ID | M_Product_ID | PriceStd | C_UOM_ID.X12DE355 |
       | plv_1                  | product      | 10.00    | PCE               |
 
-    # 100 PCE on-hand stock (10 complete TUs of 10 PCE each)
+    # 100 PCE on-hand stock (10 complete TUs of 10 PCE each → 10 HUs)
     And metasfresh contains single line completed inventories
-      | M_Inventory_ID | M_Warehouse_ID | MovementDate | M_Product_ID | QtyBook | QtyCount | M_HU_PI_Item_Product_ID | M_HU_ID |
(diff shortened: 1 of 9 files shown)


adi-stefan:
@claude review

PR #24363, opened 2026-06-02
nShift: Country of Origin from HU attribute

mrie04:
## Summary

- Reads `CountryOfOrigin` (ISO 3166-1 alpha-2) from the HU attribute chain `M_Package → M_Package_HU → LU HU → M_HU_Assignment (InOutLine-filtered) → VHU → M_HU_Attribute` and splits nShift shipment parcels by country of origin — each parcel carries goods from exactly one country
- Propagates `C_BPartner_Location.Attention` through the delivery address chain to the nShift request
- New `CountryOfOrigin CHAR(2)` column on `Carrier_ShipmentOrder_Item` (migration `5805640`); new AD_Ref_List entry for `CountryOfOrigin` mapping attribute value (migration `5805610`)
- New Cucumber scenarios S0355_DeliveryOrder_110 (VHU-based, 2 products × 2 countries → 4 parcels) and S0355_DeliveryOrder_120 (LU-cu-direct, 2 products × 2 countries → 4 parcels)

## Test plan

- [x] Cucumber `nShiftShipment.feature` — S0355_DeliveryOrder_100 (existing), S0355_DeliveryOrder_110 (new, VHU-based), S0355_DeliveryOrder_120 (new, LU-cu-direct) — all green locally
- [x] `NShiftShipmentServiceTest` snapshot updated


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -888,4 +897,128 @@ public ProjectId extractSingleProjectIdOrNull(@NonNull final List<ShipmentSchedu
...
+			return ShipmentScheduleWithHU.ofShipmentScheduleQtyPicked(qtyPicked, huContext, qtyTypeToUse);
+		}
+
+		final I_M_HU huToInspect = handlingUnitsBL.getById(huToInspectId);
+		if (handlingUnitsBL.isVirtual(huToInspect))
+		{
+			return ShipmentScheduleWithHU.ofShipmentScheduleQtyPicked(qtyPicked, huContext, qtyTypeToUse);
+		}
+
+		final I_M_ShipmentSchedule shipmentSchedule = create(qtyPicked.getM_ShipmentSchedule(), I_M_ShipmentSchedule.class);
```
teosarca: !

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -888,4 +897,128 @@ public ProjectId extractSingleProjectIdOrNull(@NonNull final List<ShipmentSchedu
...
+				Loggables.addLog("No eligible {} records found for hu {}",
+						I_M_ShipmentSchedule_QtyPicked.Table_Name,
+						handlingUnitsBL.getDisplayName(hu));
+			}
+		}
+		result.sort(new ShipmentScheduleWithHUComparator());
+		return result;
+	}
+
+	private static String computeUseInASIFingerprint(
```
teosarca: consider using AttributeKeys but also mind your "isUseInASI" constraint!

mrie04: Applied — replaced the custom string fingerprint with `AttributesKeys.createAttributesKeyFromAttributeSet()` on an `ImmutableAttributeSet` subset filtered to `isUseInASI && !isEmpty` attributes. The map key type changes from `String` to `AttributesKey`.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
mrie04: Applied — the unnecessary `create()` cast was removed, and the full expansion logic was moved to `IHUShipmentScheduleBL.createCandidatesForQtyPicked` where `IHUProductStorage.getProductId()` is used for per-product VHU filtering. `qtyPicked.getM_ShipmentSchedule().getM_Product_ID()` is called once to obtain the reference product ID for that filter — `I_M_ShipmentSchedule_QtyPicked` has no direct `M_Product_ID` column so the schedule load is unavoidable here.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -1009,16 +1032,36 @@ public List<ShipmentScheduleWithHU> retrieveShipmentSchedulesWithHUsFromHUs(@Non
...
-				.collect(Collectors.joining(","));
+		// IAttributeStorage does not implement getAttributeValueIdOrNull, so we cannot use
+		// AttributesKeys.createAttributesKeyFromAttributeSet here — list attributes (e.g. COO)
+		// would be silently dropped. Instead, we build the key directly from IAttributeValue,
+		// using the value string for all attribute types (consistent for grouping purposes).
+		final ImmutableSet<AttributesKeyPart> parts = factory.getAttributeStorage(vhu)
+				.getAttributeValues()
+				.stream()
+				.filter(av -> av.isUseInASI() && !av.isEmpty())
+				.map(av -> {
```
teosarca: avoid large lambdas, pls extract to a method.

cature this learning.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -920,59 +924,78 @@ public List<ShipmentScheduleWithHU> createCandidatesForQtyPicked(
...
+		// Case 1: find the VHU whose qty exactly matches QtyPicked — one candidate using that VHU.
+		// Using the specific VHU as override ensures its UseInASI attributes (e.g. COO) are read
+		// from the VHU itself rather than from the destination TU, which may carry a mixed/null COO.
+		final IHUProductStorage exactMatchStorage = storagesByFingerprint.values().stream()
+				.flatMap(Collection::stream)
+				.filter(s -> {
+					final Quantity vhuQty = s.getQty();
+					// UOM is consistent (same product → same stocking UOM), so wrap the raw BigDecimal
+					// in a Quantity with the VHU's UOM for a proper value+UOM comparison.
+					final Quantity pickedQty = vhuQty.toZero().add(qtyPicked.getQtyPicked());
```
teosarca: this looks pretty weird

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -920,59 +924,78 @@ public List<ShipmentScheduleWithHU> createCandidatesForQtyPicked(
...
-				.filter(h -> storageFactory.getStorage(h).getQuantity(productId).map(q -> !q.isZero()).orElse(false))
-				.collect(ImmutableList.toImmutableList());
-
-		if (childVHUs.size() <= 1)
+		// One pass: for each child VHU, load product storage and attribute fingerprint together.
+		// Filter by product so VHUs carrying a different product in the same TU are excluded.
+		// UOM consistency is guaranteed: qtyPicked links to one M_ShipmentSchedule which has one
+		// M_Product_ID and therefore one stocking UOM; all filtered storages carry that same product
+		// and are in the same UOM — no conversion needed.
+		final Map<AttributesKey, List<IHUProductStorage>> storagesByFingerprint =
```
teosarca: u could use ListMultimap

mrie04: Applied — changed to `ImmutableListMultimap<AttributesKey, IHUProductStorage>` built via `ImmutableListMultimap.toImmutableListMultimap()`.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
mrie04: Applied — replaced `vhuQty.toZero().add(pickedBD)` with a direct BigDecimal comparison: `s.getQty().getAsBigDecimal().compareTo(pickedQtyBD) == 0`. UOM consistency is guaranteed (same product → same stocking UOM) so the raw BigDecimal comparison is safe here.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
mrie04: Applied — extracted the lambda body to a private static `toAttributesKeyPart(IAttributeValue av)` method. Will capture the large-lambda rule.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -920,59 +924,78 @@ public List<ShipmentScheduleWithHU> createCandidatesForQtyPicked(
...
+				handlingUnitsDAO.retrieveIncludedHUs(huToInspect)
+						.stream()
+						.filter(handlingUnitsBL::isVirtual)
+						.flatMap(h -> storageFactory.getStorage(h).getProductStorages().stream()
+								.filter(s -> productId.equals(s.getProductId()))
+								.filter(s -> !s.getQty().isZero())
+								.map(s -> Maps.immutableEntry(computeUseInASIFingerprint(h, attrFactory), s)))
```
teosarca: calling computeUseInASIFingerprint(h, attrFactory) for each "s" is not optimum.

pls avoid such weird lambdas, pls better split them in several methods with intention-revealing names,
pls avoid variables names like "s", "h" (u have a rule for that!)

pls evaluate and also capture learning about this.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -920,59 +924,78 @@ public List<ShipmentScheduleWithHU> createCandidatesForQtyPicked(
...
+		// UOM consistency is guaranteed: qtyPicked links to one M_ShipmentSchedule which has one
+		// M_Product_ID and therefore one stocking UOM; all filtered storages carry that same product
+		// and are in the same UOM — no conversion needed.
+		final Map<AttributesKey, List<IHUProductStorage>> storagesByFingerprint =
+				handlingUnitsDAO.retrieveIncludedHUs(huToInspect)
+						.stream()
+						.filter(handlingUnitsBL::isVirtual)
+						.flatMap(h -> storageFactory.getStorage(h).getProductStorages().stream()
+								.filter(s -> productId.equals(s.getProductId()))
+								.filter(s -> !s.getQty().isZero())
```
teosarca: don't we have s.isEmpty()

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -920,59 +924,78 @@ public List<ShipmentScheduleWithHU> createCandidatesForQtyPicked(
...
+		// Filter by product so VHUs carrying a different product in the same TU are excluded.
+		// UOM consistency is guaranteed: qtyPicked links to one M_ShipmentSchedule which has one
+		// M_Product_ID and therefore one stocking UOM; all filtered storages carry that same product
+		// and are in the same UOM — no conversion needed.
+		final Map<AttributesKey, List<IHUProductStorage>> storagesByFingerprint =
+				handlingUnitsDAO.retrieveIncludedHUs(huToInspect)
+						.stream()
+						.filter(handlingUnitsBL::isVirtual)
+						.flatMap(h -> storageFactory.getStorage(h).getProductStorages().stream()
+								.filter(s -> productId.equals(s.getProductId()))
```
teosarca: why not using de.metas.handlingunits.storage.IHUStorage#getProductStorageOrNull

pls capture learning (module level).

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
mrie04: Applied — `!s.getQty().isZero()` replaced with `!productStorage.isEmpty()`.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
mrie04: Applied — extracted the `flatMap` body to `groupProductStoragesByFingerprint(vhu, productId, storageFactory, attrFactory)`. The fingerprint is now computed once per VHU, not once per product storage. Variable names `h`/`s` replaced with `vhu`/`productStorage`. Will capture the learnings (avoid large lambdas + meaningful variable names).

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -920,59 +924,78 @@ public List<ShipmentScheduleWithHU> createCandidatesForQtyPicked(
...
-				.collect(Collectors.groupingBy(h -> computeUseInASIFingerprint(h, attrFactory)));
-		if (vhusByFingerprint.size() <= 1)
+		// Case 2: whole-TU-pick — the QtyPicked covers the entire TU content (one pick for all VHUs).
+		// Guard: if the picked qty does not equal the grand total of all VHUs, this QtyPicked covers
+		// only part of the TU and another QtyPicked record covers the rest — fall back to a single
+		// default candidate so each QtyPicked is not independently inflated by the full group sums.
+		// The zero Quantity is derived from any VHU in the map — Collectors.groupingBy never produces
+		// an empty-list value, and size > 1 guarantees at least one group is present.
+		final Quantity zeroInPickedUOM = storagesByFingerprint.values().iterator().next().get(0).getQty().toZero();
+		final Quantity grandTotal = storagesByFingerprint.values().stream()
```
teosarca: grandTotal  is quite misleading name. we usually use it for money/amounts

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -920,59 +924,78 @@ public List<ShipmentScheduleWithHU> createCandidatesForQtyPicked(
...
+				.flatMap(Collection::stream)
+				.map(IHUProductStorage::getQty)
+				.reduce(zeroInPickedUOM, Quantity::add);
+		final Quantity pickedQtyAsQuantity = zeroInPickedUOM.add(qtyPicked.getQtyPicked());
```
teosarca: why this ZERO + qty approach?!

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
mrie04: Applied — `groupProductStoragesByFingerprint` now uses `getProductStorageOrNull(productId)` directly (single lookup per VHU) instead of streaming all product storages and filtering. Will capture this as a module-level learning in the HU storage CLAUDE.md.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
mrie04: Applied — renamed to `totalVHUQtyBD` (BigDecimal, reflects what it actually is: total qty of VHU storages).

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
mrie04: Applied — replaced the `toZero() + reduce(Quantity)` pattern with a direct `BigDecimal.ZERO + reduce(BigDecimal::add)` sum. UOM consistency is guaranteed (same product → same stocking UOM), so BigDecimal arithmetic is safe and simpler.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -965,16 +965,14 @@ public List<ShipmentScheduleWithHU> createCandidatesForQtyPicked(
...
 		// default candidate so each QtyPicked is not independently inflated by the full group sums.
-		// The zero Quantity is derived from any VHU in the map — ImmutableListMultimap never
-		// stores an empty list for a key, and isEmpty() was checked above so at least one entry is present.
-		final Quantity zeroInPickedUOM = storagesByFingerprint.values().iterator().next().getQty().toZero();
-		final Quantity grandTotal = storagesByFingerprint.values().stream()
-				.map(IHUProductStorage::getQty)
-				.reduce(zeroInPickedUOM, Quantity::add);
-		if (grandTotal.getAsBigDecimal().compareTo(pickedQtyBD) != 0)
+		// BigDecimal arithmetic is safe here: UOM consistency is guaranteed (same product → same stocking UOM).
+		final BigDecimal totalVHUQtyBD = storagesByFingerprint.values().stream()
```
teosarca: pls never use BigDecimal for Qtys instead of Quantity.
more, in this case you are unboxing each product storage's qty just to sum them!?!?! 
that is super wrong in case (shall not happen) they are about different UOMs.
avoiding such a big mistake is not negociable.

pls capture learning.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/attribute/strategy/impl/KeepExistingValueAttributeSplitterStrategy.java
```diff
@@ -0,0 +1,61 @@
...
+import de.metas.handlingunits.attribute.strategy.IAttributeSplitterStrategy;
+import de.metas.util.Check;
+import lombok.NoArgsConstructor;
+
+/**
+ * TOPD splitter that propagates the parent's value to a child only when the child has no value yet.
+ * If the child already carries a value it is left unchanged.
+ *
+ * Designed for categorical attributes like Country of Origin where each TU may hold a distinct
+ * batch with its own origin. A mixed-origin LU must not silently overwrite per-TU origins.
```
teosarca, line 37: i have a small hesitation that empty could mean not set or "really unknown, don't set it".... but i don;t have a better suggestion then having a "N/A" for that case.... i think we can leave it for later...

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
mrie04: Changed to use Quantity::add via reduce + filter approach. An empty stream now falls back to the single default candidate (same as qty mismatch) rather than throwing.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/attribute/strategy/impl/KeepExistingValueAttributeSplitterStrategy.java
mrie04: Conceptually valid — there is currently no N/A sentinel in the COO list, so a user has no way to express 'deliberately unknown'. As long as that gap exists, Check.isEmpty correctly means 'not yet assigned'. If an N/A value is introduced later, this strategy would need to treat it as a set value rather than propagating from the parent. Leaving as-is for now, follow-up if N/A is added.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
mrie04: rule added

mrie04:

teosarca:
### Possible catch-weight whole-HU double-book on `new_dawn_uat` — bisected to this PR ✕ #24478

Flagging a deterministic `mobile test` failure on `new_dawn_uat` that git-bisects to this PR together with #24478. Not urgent, but it currently blocks every merge PR into `new_dawn_uat` (its `mobile test` job can't go green), so it needs an owner call.

#### Symptom
`e2e/mobile-webui/tests/spec/picking/overPickingPrompt_wholeHU.spec.js` fails: a **catch-weight whole-HU pick books two `M_ShipmentSchedule_QtyPicked` rows where one is expected**.

```
AssertPickingExpectationsCommand.assertQtyPickedList:244
"Expected qty picked records size to be <1> but was <2>"
```

#### Root-cause chain (code on `new_dawn_uat`)
Two save-sites both introduced in June (this PR + #24478), neither present on `intensive_care_hotfix`:

1. **This PR creates the second row.** The Country-of-Origin rework books the destination TU and expands it into per-VHU COO candidates at shipment-generation time — `HUShipmentScheduleBL.createCandidatesForQtyPicked` (`HUShipmentScheduleBL.java:1186`).
2. **#24478 then fails to collapse it for catch weight.** `HUShipmentScheduleBL.tryMergeQtyPickedIntoExistingForVHU` (`:240`, wired into `ShipmentScheduleHUTrxListener.java:123`) deliberately bails out on catch weight:
   ```java
   // HUShipmentScheduleBL.java:255-258
   // ... fall through to create-new so each catch reading keeps its own row.
   if (qtyToAdd.isUOMQtySet())
   {
       return false;
   }
   ```
   The failing fixture (`createMasterdata_WholeHU`) is a catch-weight product → `isUOMQtySet()` is true → the merge bails → both rows persist.

#### Why this is a backend defect on these two PRs, not the E2E
- **Control experiment.** The *same* 887-line spec is **green on `intensive_care_hotfix`** (build `.41473`, run https://github.com/metasfresh/metasfresh/actions/runs/30716067854 — 12/12 on this spec) and **red on `new_dawn_uat`** (run https://github.com/metasfresh/metasfresh/actions/runs/30714037116). Same test, different backend ⇒ the red is backend-determined.
- **Code presence.** `createCandidatesForQtyPicked` / `tryMergeQtyPickedIntoExistingForVHU` grep to **0 hits on `intensive_care_hotfix`**, present on `new_dawn_uat`.
- **Ancestry.** The introducing commits (`489b618710f`/`e0d9cd1b07c` here, `c50466a6bf3` on #24478) are ancestors of `new_dawn_uat` but **not** of `intensive_care_hotfix`, merged **2026-06-09/10**.

#### Behaviour vs. visibility (full disclosure)
The *exposing test* is newer than the defect: `overPickingPrompt_wholeHU.spec.js` reached `new_dawn_uat` via the whole-TU over-delivery work (#25374) propagated in #25367 (merge `bbe873afa6e`, 2026-08-01) — roughly 7 weeks *after* these June backend commits. A `.spec.js` cannot write DB rows, so the test only made an existing backend double-book visible; it did not introduce it. The first `new_dawn_uat` build to contain the spec (run https://github.com/metasfresh/metasfresh/actions/runs/30699288552) failed earlier at the `java` compile step, so `mobile test` only started reporting this once the build compiled.

#### Not yet done (honest gaps)
- **Not reproduced locally**, so I have not pinned which of the two save-sites emits the concrete second row (the COO per-VHU expansion here vs. the create-new fall-through after the catch-weight bail in #24478). Attribution to these two June PRs holds either way.
- **Open design question:** is the two-row outcome the *intended* catch-weight semantics (one row per catch reading) — in which case the test's expected-1 is what should change — **or** a genuine double-book to eliminate?

#### Ask
As the author of the COO expansion: is the second row expected here for a catch-weight whole-HU pick, or should it be collapsed? Happy to reproduce and pin the exact emitting line if that helps.

PR #24365, opened 2026-06-02
Process to change costing method to Moving Average Invoice

metas-rc:
This PR adds a one-time initialization utility for switching an accounting schema from any costing method to Moving Average Invoice (MAI).

What was added:
- 5805700 — PL/pgSQL function C_AcctSchema_InitMovingAvgInvoice(p_C_AcctSchema_ID) that seeds M_Cost records for the MAI cost element from the existing cost records, then switches c_acctschema.costingmethod = 'M'. If the schema is already on MAI it returns silently.
- 5805710 — registers the function as an AD_Process (Type=SQL, class ExecuteUpdateSQL) attached to the Accounting Schema window as a document action. Also makes CostingMethod field read-only so users can't bypass the process.
- DDL counterpart at ddl/functions/C_AcctSchema_InitMovingAvgInvoice.sql

Key design decisions to be aware of:
- The MAI cost element is looked up dynamically (costingmethod='M', costelementtype='M', isactive='Y') — no hardcoded IDs
- Both m_cost and c_acctschema are backed up via backup_table() before any data change
- createdby/updatedby in the function body use 99 (migration user)
- No Java class — the SQL function is called directly via ExecuteUpdateSQL

### DIFF
--- a/backend/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/C_AcctSchema_InitMovingAvgInvoice.sql
+++ b/backend/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/C_AcctSchema_InitMovingAvgInvoice.sql
@@ -69,9 +69,9 @@ BEGIN
             v_AD_Client_ID;
     END IF;
 
-    -- Backup affected tables before making changes (suffix includes schema ID to avoid collision)
-    PERFORM backup_table('m_cost',       '_pre_mai_' || p_C_AcctSchema_ID);
-    PERFORM backup_table('c_acctschema', '_pre_mai_' || p_C_AcctSchema_ID);
+    -- Backup affected tables before making changes
+    PERFORM backup_table('m_cost',       '_pre_mai');
+    PERFORM backup_table('c_acctschema', '_pre_mai');
 
     -- Seed M_Cost records for the MAI cost element.
     -- Skip products that already have a record for the target cost element + cost type.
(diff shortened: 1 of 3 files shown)


metas-rc:
@claude review opus

metas-rc:
@claude 
Thanks for the thorough review. Addressed in https://github.com/metasfresh/metasfresh/pull/24365/changes/eba34fc2e565dc7192b80de26e253fa2e5e1005f and https://github.com/metasfresh/metasfresh/pull/24365/changes/1ece03d2acbb3b84ef16f893ce9359a8744202d3

Bug — m_costtype_id missing from NOT EXISTS → Fixed. Added AND tgt.m_costtype_id = src.m_costtype_id to the subquery.

Moderate — Non-deterministic LIMIT 1 → Fixed. Added ORDER BY m_costelement_id to both cost element lookups.

Minor — VOID return hides result → Fixed. Changed to RETURNS text; both the early-return path and the success path now return a descriptive message.

Moderate — NULL-unsafe cumulatedamt → Not applied. currentcostprice and currentqty are NOT NULL columns — COALESCE would be dead code.

Moderate — Backup suffix collision → Not applied. backup_table() backs up the full table regardless of which schema is being processed. The suffix is a descriptive label, not a per-run key — appending the schema ID would be misleading.

Minor — Global read-only field → Intentional by design. A CostingMethod change outside of this process is never safe, so locking it globally is correct. Added a comment in the migration documenting this.

cristinamghita (commented):
lgtm

PR #24369, opened 2026-06-02
Costing — prevent m_cost deadlock via SELECT FOR UPDATE on current-cost reads

teosarca:
## Problem

CI intermittently fails with `DBDeadLockDetectedException` on relation `m_cost` when a synchronous document operation (e.g. material-receipt reversal → `voidCosts`) races the asynchronous accounting posting of a prior document touching the same `M_Cost` row. Postgres `DETAIL` from 4 failing runs shows the classic single-row lock-upgrade deadlock: both transactions hold a shared row lock (implicit `FOR KEY SHARE` from FK-child inserts), then both `UPDATE` the same `m_cost` tuple and deadlock on the share→exclusive upgrade.

Failing runs (evidence):
- https://github.com/metasfresh/metasfresh/actions/runs/26271713419
- https://github.com/metasfresh/metasfresh/actions/runs/26297524912
- https://github.com/metasfresh/metasfresh/actions/runs/26357686433
- https://github.com/metasfresh/metasfresh/actions/runs/26289296476

## Fix

Acquire the exclusive row lock **up front** at read time, so the second writer queues at the `SELECT` instead of taking a share lock and later deadlocking on the upgrade:

1. **Query API**: new `ForUpdate` enum (`NONE` / `FOR_UPDATE` / `FOR_UPDATE_SKIP_LOCKED`) + `IQuery.setForUpdate(ForUpdate)`, generalizing the previous `TypedSqlQuery.setForUpdateSkipLocked` (removed; its sole caller `QueueProcessorPlanner` migrated — also dropping the `TypedSqlQuery.cast` + unit-test-mode guard there). `buildSQL` fails fast on `FOR UPDATE` + `UNION`/`GROUP BY`.
2. **Costing**: `ICurrentCostsRepository.getOrCreate` → **`getOrCreateForUpdate`** — the current-cost read now issues `SELECT … FOR UPDATE` (held until trx end). Covers all costing-method handlers (incl. manufacturing) via the shared `CostingMethodHandlerUtils.getCurrentCostForUpdate` (renamed, ~39 call sites) plus `CostingService.revaluateCosts` / `RollupWorkflow`.
3. **Cucumber**: removed the rabbitMQ-drain band-aid step from `splitPaymentReversalCascade.feature` — the test again reverses documents back-to-back and exercises the exact concurrency this PR fixes.

An autocommit (transaction-less) caller is unaffected: it cannot hold-and-wait, so it cannot deadlock; the locking clause is simply inert there.

## Validation

- TDD: 6 new `TypedSqlQueryTests` methods (clause emission per mode, default, UNION/GROUP-BY fail-fast) — RED first, then GREEN.
- Unit: base + async + business (743) module suites green locally.
- Cucumber (band-aid removed): local 5×/5× runs of `splitPaymentReversalCascade.feature` + `receipts.feature` in progress — first runs green with 0 deadlocks; results will be posted on this PR.


backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/impl/DeadlockRetryPolicy.java
```diff
@@ -0,0 +1,178 @@
+package org.adempiere.ad.trx.api.impl;
```
teosarca: move it outside of impl package.

rule: don't place public classes inside the `impl` package besides the ISingletonService implementations and package level/supporting classes of those implementations.

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/impl/DeadlockRetryPolicy.java
```diff
@@ -0,0 +1,178 @@
...
+	private DeadlockRetryPolicy(final int maxAttempts, final long backoffMillis)
+	{
+		this.maxAttempts = maxAttempts > 0 ? maxAttempts : DEFAULT_MAX_ATTEMPTS;
+		this.backoffMillis = backoffMillis >= 0 ? backoffMillis : DEFAULT_BACKOFF_MILLIS; // 0 = no sleep (tests)
+	}
+
+	/** Creates the default policy (3 attempts, 5 s backoff). */
+	public static DeadlockRetryPolicy defaults()
+	{
+		return DeadlockRetryPolicy.builder()
```
teosarca: why not having a private static final `DeadlockRetryPolicy DEFAULT = ....` and return it here?
or is this DeadlockRetryPolicy mutable?

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/impl/DeadlockRetryPolicy.java
```diff
@@ -0,0 +1,178 @@
...
+	private static final Logger logger = LogManager.getLogger(DeadlockRetryPolicy.class);
+
+	/** PostgreSQL SQLSTATE for "deadlock detected". */
+	private static final String PG_SQLSTATE_DEADLOCK_DETECTED = "40P01";
+
+	/** Maximum number of attempts (first attempt + retries). */
+	@VisibleForTesting
+	public static final int DEFAULT_MAX_ATTEMPTS = 3;
+
+	/** Milliseconds to sleep between attempts, mirroring WorkpackageProcessorTask's retry timeout. */
```
teosarca: IMHO WorkpackageProcessorTask reference is just poluting this documentatoin

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/DeadlockRetryPolicy.java
```diff
@@ -0,0 +1,178 @@
...
+				break;
+			}
+		}
+		return false;
+	}
+
+	/** Sleeps for {@code millis} ms; swallows {@link InterruptedException} and restores the flag. */
+	private static void sleepQuietly(final long millis)
+	{
+		try
```
teosarca, line 159: i suggest adding "if millis <= 0 return" as first line

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/impl/DeadlockRetryPolicy.java
```diff
@@ -0,0 +1,178 @@
...
+			{
+				if (!isDeadlock(ex))
+				{
+					throw ex;
+				}
+
+				lastException = ex;
+				logger.warn(
+						"Deadlock detected on attempt {}/{} for {}; will retry after {} ms. Deadlock: {}",
+						attempt, maxAttempts, context, backoffMillis, ex.getMessage());
```
teosarca: instead of ex.getMesage() why not providing the whole 'ex'  (ofc remove "Deadlock: {}" from message and keep the exception last).

teosarca: rule: never swallow an exception or its stacktrace without logging or throwing it. The only exception to this rule is when we do flow-control with exceptions (but that is a bad pattern too)

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/impl/DeadlockRetryPolicy.java
```diff
@@ -0,0 +1,178 @@
...
+
+	/**
+	 * Runs {@code action}; on {@link #isDeadlock(Throwable)} retries up to {@code maxAttempts - 1}
+	 * times after sleeping {@code backoffMillis}. After exhausting retries the last exception is
+	 * re-thrown as-is (preserving the original type and stack).
+	 *
+	 * @param action  the idempotent operation to execute
+	 * @param context an optional description used in log messages (e.g. the posting request)
+	 */
+	public void run(@NonNull final Runnable action, @Nullable final Object context)
```
teosarca: what if instead of `Object context` we have `Object ...context` so we will be able to pass more variables or none.

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/impl/DeadlockRetryPolicy.java
```diff
@@ -0,0 +1,178 @@
...
+	 * times after sleeping {@code backoffMillis}. After exhausting retries the last exception is
+	 * re-thrown as-is (preserving the original type and stack).
+	 *
+	 * @param <T>     return type of the supplier
+	 * @param action  the idempotent operation to execute
+	 * @param context an optional description used in log messages (e.g. the posting request);
+	 *                may be null (unit-test documents have no document info) — log-only, never dereferenced
+	 * @return the value returned by {@code action} on a successful attempt
+	 */
+	public <T> T call(@NonNull final Supplier<T> action, @Nullable final Object context)
```
teosarca: what if instead of `Object context` we have `Object ...context` so we will be able to pass more variables or none.

backend/de.metas.adempiere.adempiere/base/src/main/java/org/compiere/model/IQuery.java
```diff
@@ -385,6 +386,25 @@ default int maxInt(final String columnName)
 
 	IQuery<T> setRequiredAccess(@Nullable Access access);
 
+	/**
+	 * Appends a PostgreSQL row-locking clause (e.g. {@code FOR UPDATE}) to the query.
```
teosarca: at this level is not postgresql related! adjust the javadoc.
also make it concise
no need for `@return`
`@param` is more repeating what u said. so DRY! either u write in param either in javadoc above but not in both!

backend/de.metas.adempiere.adempiere/base/src/test/java/org/adempiere/ad/dao/impl/TypedSqlQueryTests.java
```diff
@@ -174,4 +176,101 @@ public void givenUnionQueryWithNoOrderByOnRoot_whenBuildingSQL_thenUnionOrderByC
...
+		@Test
+		public void forUpdate_emitsClauseAfterOrderBy()
+		{
+			final TypedSqlQuery<I_AD_Table> query = new TypedSqlQuery<>(Env.getCtx(), I_AD_Table.class, "IsActive='Y'", ITrx.TRXNAME_None);
+			query.setOrderBy("AD_Table_ID");
+			query.setForUpdate(ForUpdate.FOR_UPDATE);
+
+			final String sql = query.buildSQL("avoidDBConnection", null, null, true);
+
+			assertThat(sql).contains("ORDER BY AD_Table_ID");
```
teosarca: when validating buildSQL pls assert the whole resulting sql and not just parts of it. 
rule+review rule: when writting tests and you have to validate a resulting string, pls validate the entire string and just parts of it. Avoid contains/doesNotContain/indexOf.

backend/de.metas.adempiere.adempiere/base/src/test/java/org/adempiere/ad/trx/api/impl/DeadlockRetryPolicyTest.java
```diff
@@ -0,0 +1,249 @@
+package org.adempiere.ad.trx.api.impl;
```
teosarca: keep it in the same package as the tested class (DeadlockRetryPolicy)

backend/de.metas.adempiere.adempiere/base/src/test/java/org/adempiere/ad/trx/api/DeadlockRetryPolicyTest.java
```diff
@@ -0,0 +1,249 @@
...
+	// isDeadlock detection
+	// -----------------------------------------------------------------------
+
+	@Nested
+	class IsDeadlock
+	{
+		@Test
+		void directDBDeadLockDetectedException()
+		{
+			final DBDeadLockDetectedException ex = new DBDeadLockDetectedException(
```
teosarca, line 56: u have the 40P01 constant declared in DeadlockRetryPolicy class. just make it package level and use it

backend/de.metas.async/src/main/java/de/metas/async/processor/impl/planner/QueueProcessorPlanner.java
```diff
@@ -282,12 +281,11 @@ private List<I_C_Queue_WorkPackage> pollAndLockWorkPackages(@NonNull final Prope
...
 			final IQuery<I_C_Queue_WorkPackage> query = queryOptional.get();
 
-			if (!Adempiere.isUnitTestMode())
-			{// Enable FOR UPDATE SKIP LOCKED on this query
-				TypedSqlQuery.cast(query).setForUpdateSkipLocked(true);
-			}
-			
-			// Set LockedAt timestamp immediately (rows are still locked via "ForUpdateSkipLocked")
+			// Enable FOR UPDATE SKIP LOCKED on this query.
+			// POJOQuery (used in unit tests) accepts the call as an in-memory no-op, so no unit-test guard is needed.
```
teosarca: remove this comment. imho is pointless

backend/de.metas.business/src/main/java/de/metas/costing/ICurrentCostsRepository.java
```diff
@@ -43,7 +43,23 @@ public interface ICurrentCostsRepository
 	@Nullable
 	CurrentCost getOrNull(CostSegmentAndElement costSegmentAndElement);
 
-	CurrentCost getOrCreate(CostSegmentAndElement costSegmentAndElement);
+	/**
+	 * Returns the current cost record for the given segment, creating one if absent.
+	 *
+	 * <p>Acquires a PostgreSQL row lock ({@code SELECT ... FOR NO KEY UPDATE}) on the existing row,
```
teosarca: Jesus, another huge AI slop. can u pls say this in one phrase max! No postgresql, no bla bla bla

pls improve/tune your javadoc/comments rule to avoid this kind of verbal diarrhea

backend/de.metas.business/src/main/java/de/metas/document/engine/impl/AbstractDocumentBL.java
```diff
@@ -148,7 +152,30 @@ public void doFinally()
...
 				setTrxName(document.getDocumentModel(), trxName, true /* ignoreIfNotHandled */);
 			}
-		});
+		};
+
+		// Engine owns the transaction when there is none to inherit: trxName is null/None,
+		// OR it is the ThreadInherited marker while NO actual thread transaction exists
+		// (trxManager.call then opens its own new transaction). Observed mis-classification:
+		// cucumber-loaded models carry the ThreadInherited marker with no active trx.
+		final boolean engineOwnsTrx = trxManager.isNull(trxName)
```
teosarca: extract as a private method called isEngineOwnsTheTransaction

backend/de.metas.business/src/main/java/de/metas/document/engine/impl/AbstractDocumentBL.java
```diff
@@ -148,7 +152,30 @@ public void doFinally()
...
+				|| (ITrx.TRXNAME_ThreadInherited.equals(trxName) && !trxManager.hasThreadInheritedTrx());
+
+		final Boolean processed;
+		if (engineOwnsTrx)
+		{
+			// The engine owns the transaction (a new one is opened per attempt): on a DB deadlock the whole
+			// document action is rolled back, so re-running it is equivalent to the user re-triggering it.
+			// Why: a document reversal (holding DELETE row locks) can deadlock against a concurrent accounting
+			// posting whose deferred FK checks at COMMIT reference the deleted rows — not preventable by lock ordering.
+			processed = DEADLOCK_RETRY_POLICY.call(() -> trxManager.call(trxName, processCallable), document.getDocumentInfo());
```
teosarca: Consider introducing following methods to DeadlockRetryPolicy:

public <T> TrxCallable<T> wrapAsCallableIf(boolean condition, TrxCallable<T> callable,Object ... context);
* if condition is true calls wrapAsCallable

public <T> TrxCallable<T> wrapAsCallable(TrxCallable<T> callable, Object ... context);


and here we would just use those methods nicely and fluently.

----

regarding comments, again, evaluate your rule, it's not allowed to write the aprox same 4-line AI slop one after the other. make sure u are combining then in one very concise comment

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/impl/DeadlockRetryPolicy.java
teosarca: Done — moved `DeadlockRetryPolicy` (and its test) out of the `impl` package to `org.adempiere.ad.trx.api`. rule added.

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/impl/DeadlockRetryPolicy.java
teosarca: Done — added `public static final DEFAULT` and `defaults()` returns it; both call sites use `DeadlockRetryPolicy.DEFAULT`. The policy is immutable (all fields final, private builder).

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/impl/DeadlockRetryPolicy.java
teosarca: Done — removed the WorkpackageProcessorTask reference from the javadoc.

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/DeadlockRetryPolicy.java
teosarca: Done — `sleepQuietly` now returns early when `millis <= 0`.

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/impl/DeadlockRetryPolicy.java
teosarca: Done — the whole exception is now the last SLF4J argument (full stacktrace logged); dropped the "Deadlock: {}" token from the message.

teosarca: Done — the exception is logged in full on every attempt and rethrown once attempts are exhausted, so it is never swallowed. rule added.

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/impl/DeadlockRetryPolicy.java
teosarca: Done — both `run` and `call` now take `Object... context`.

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/impl/DeadlockRetryPolicy.java
teosarca: Done — switched to `Object... context` (see sibling comment).

backend/de.metas.adempiere.adempiere/base/src/main/java/org/compiere/model/IQuery.java
teosarca: Done — rewrote the javadoc concise and interface-level (no PostgreSQL / SQL-emission detail), removed `@return` and the duplicated `@param` prose.

backend/de.metas.adempiere.adempiere/base/src/test/java/org/adempiere/ad/dao/impl/TypedSqlQueryTests.java
teosarca: Done — the buildSQL tests now assert the entire SQL via `isEqualToIgnoringWhitespace`, no more contains/doesNotContain/indexOf. rule added.

backend/de.metas.adempiere.adempiere/base/src/test/java/org/adempiere/ad/trx/api/impl/DeadlockRetryPolicyTest.java
teosarca: Done — `DeadlockRetryPolicyTest` moved to `org.adempiere.ad.trx.api`, the same package as the class.

backend/de.metas.adempiere.adempiere/base/src/test/java/org/adempiere/ad/trx/api/DeadlockRetryPolicyTest.java
teosarca: Done — made `PG_SQLSTATE_DEADLOCK_DETECTED` package-private and the test imports it instead of redeclaring 40P01.

backend/de.metas.async/src/main/java/de/metas/async/processor/impl/planner/QueueProcessorPlanner.java
teosarca: Done — removed the comment.

backend/de.metas.business/src/main/java/de/metas/costing/ICurrentCostsRepository.java
teosarca: Done — collapsed to one line. rule added (tightened the javadoc/comments rule: hard conciseness cap, no framework/DB mechanics, never stack near-duplicate comment blocks).

backend/de.metas.business/src/main/java/de/metas/document/engine/impl/AbstractDocumentBL.java
teosarca: Done — extracted `private static boolean isEngineOwnsTheTransaction(ITrxManager, String)`.

backend/de.metas.business/src/main/java/de/metas/document/engine/impl/AbstractDocumentBL.java
teosarca: Instead of `wrapAsCallable(TrxCallable, ...)` passed into `trxManager.call`, I added `callIf(boolean condition, Supplier<T> action, Object... context)` on the policy. Correctness reason: `wrapAsCallable` would place the retry loop *inside* the `trxManager.call` invocation, so each retry would still run within the original, already-deadlock-aborted transaction. A deadlock rolls the whole transaction back at the DB level, so the only safe retry opens a *fresh* transaction — which means the retry must wrap the `trxManager.call(...)` call, not the callable body. `callIf` does exactly that (`callIf(engineOwnsTrx, () -> trxManager.call(trx, callable), ctx)`); when the condition is false it calls the action once with no retry (caller-owned-trx case). On the comments point: tightened the javadoc/comments rule to forbid stacked near-duplicate comment blocks. rule added.

teosarca:
@teosarca thanks for the thorough review — all 16 comments addressed and pushed (`b07ad8db3e4`):

- **Moved** `DeadlockRetryPolicy` (+ test) out of `impl` → `org.adempiere.ad.trx.api`; cached `DEFAULT`; `Object... context` varargs; `sleepQuietly` early-return; full exception logged + rethrown; package-private `40P01` reused by the test.
- **Concise javadoc** on `IQuery.setForUpdate` / `ICurrentCostsRepository` / `CostingMethodHandlerUtils` (intent only, no SQL/PG mechanics); corrected to `FOR NO KEY UPDATE`; removed the pointless planner comment.
- **`buildSQL` fail-fast** when a locking clause meets a non-SELECT statement (Copilot's point).
- **Full-SQL** `isEqualToIgnoringWhitespace` assertions in `TypedSqlQueryTests`.
- **`callIf`** fluent method + extracted `isEngineOwnsTheTransaction` — see the per-thread note on why `callIf` (retry outside the trx boundary) rather than wrapping the `TrxCallable` (which would retry inside the already-aborted transaction).

The four `rule:` points are now internal coding rules.

I can't mark review comments as resolved from the API — if you agree with a reply or consider a point settled, please mark it resolved on your side.

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/DeadlockRetryPolicy.java
```diff
@@ -0,0 +1,174 @@
...
+		call(() -> {
+			action.run();
+			return null;
+		}, context);
+	}
+
+	/** Calls {@code action}; retries on deadlock up to {@code maxAttempts - 1} times. */
+	public <T> T call(@NonNull final Supplier<T> action, final Object... context)
+	{
+		final String contextStr = context == null || context.length == 0 ? "" : Arrays.deepToString(context);
```
teosarca: pls lazy compute contextStr  only on first failure!

backend/de.metas.adempiere.adempiere/base/src/main/java/org/compiere/model/IQuery.java
```diff
@@ -385,6 +386,14 @@ default int maxInt(final String columnName)
 
 	IQuery<T> setRequiredAccess(@Nullable Access access);
 
+	/**
+	 * Appends a row-locking clause to the query, locking matched rows until the transaction ends.
+	 * Use {@link ForUpdate#NONE} (the default) to emit no locking clause.
+	 *
+	 * @param forUpdate the desired locking mode; never {@code null}
```
teosarca: rule: DRY comments, `never null` is just repeating the annotatoin we have. don't repeat what annotations are always saying

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/DeadlockRetryPolicy.java
teosarca: Done — `contextToString(context)` is now called only inside the deadlock catch block, so the context string is built only when a deadlock is actually caught, never on the success path. (`7406a538c73`)

backend/de.metas.adempiere.adempiere/base/src/main/java/org/compiere/model/IQuery.java
teosarca: Done — removed the `@param forUpdate ... never {@code null}` line; it only restated the `@NonNull` annotation. rule rejected — already covered by our existing comment rule (don't restate what the code/annotation already guarantees).

backend/de.metas.business/src/main/java/de/metas/document/engine/impl/AbstractDocumentBL.java
```diff
@@ -148,11 +151,25 @@ public void doFinally()
...
+		final Boolean processed = DEADLOCK_RETRY_POLICY.callIf(
+				isEngineOwnsTheTransaction(trxManager, trxName),
+				() -> trxManager.call(trxName, processCallable),
+				document.getDocumentInfo());
 
 		return processed != null && processed;
 	}
 
+	/** Returns true when the document engine owns the transaction, i.e. a fresh transaction is opened per attempt. */
+	private static boolean isEngineOwnsTheTransaction(@NonNull final ITrxManager trxManager, @Nullable final String trxName)
```
teosarca: don't make it static
rule: pls avoid passing services as method parameters to private methods of the same class. consider not making them static

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/adempiere/ad/dao/impl/TypedSqlQuery.java
```diff
@@ -1263,11 +1248,22 @@ public final String buildSQL(
...
 		}
 
-		// Add FOR UPDATE SKIP LOCKED clause if enabled
-		// This must come after ORDER BY but before LIMIT
-		if (forUpdateSkipLocked)
+		// Append the row-locking clause (FOR UPDATE / FOR UPDATE SKIP LOCKED) after ORDER BY.
+		// PostgreSQL forbids FOR UPDATE with UNION or GROUP BY — fail fast with a clear message.
+		final ForUpdate forUpdate = getForUpdate();
+		if (forUpdate != ForUpdate.NONE)
 		{
```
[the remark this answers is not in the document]
teosarca: pls eval if this was handled

backend/de.metas.business/src/main/java/de/metas/document/engine/impl/AbstractDocumentBL.java
teosarca: Done — `ITrxManager` is now an `@NonNull private final` instance field; `isEngineOwnsTheTransaction(String)` is a non-static instance method that uses the field (no service parameter, no in-method `Services.get`). `processIt0` uses the field too. rule rejected — already covered by our existing service-injection rule (services as instance fields, never static, not retrieved inside methods / passed as params).

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/DeadlockRetryPolicy.java
```diff
@@ -0,0 +1,174 @@
...
+				}
+			}
+		}
+
+		// All attempts exhausted — rethrow so the error path records it
+		throw lastException;
+	}
+
+	/** Calls {@code action} with deadlock retry if {@code condition} is true; otherwise calls it once. */
+	public <T> T callIf(final boolean condition, @NonNull final Supplier<T> action, final Object... context)
```
teosarca: the method name is totally misleading.

i've adviced you to create a method called wrapIf and which takes following params:
* boolean condition
* TrxCallable<T> callable
and which returns
* wrapped callable (with dead lock retry policy) if condition=true
* callable param if condition=false


can u pls evaluate your rules and improve them in a way this misunderstanding and fucked up solution to not happen again

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/DeadlockRetryPolicy.java
```diff
@@ -0,0 +1,174 @@
...
+		return action.get();
+	}
+
+	/**
+	 * Returns {@code true} when the throwable (or any cause) is a DB deadlock:
+	 * a {@link DBDeadLockDetectedException} or {@link SQLException} with SQLSTATE {@code 40P01}.
+	 * Self-contained to avoid a DB.isPostgreSQL() dependency that fails in unit-test mode.
+	 */
+	@VisibleForTesting
+	public static boolean isDeadlock(@Nullable final Throwable ex)
```
teosarca: rule: i understand that u cannot make the method private when `@VisibleForTesting`, but at least evaluate if we can have it package level

backend/de.metas.business/src/main/java/de/metas/document/engine/impl/AbstractDocumentBL.java
```diff
@@ -148,11 +151,25 @@ public void doFinally()
 				// put back the transaction which document had initially
 				setTrxName(document.getDocumentModel(), trxName, true /* ignoreIfNotHandled */);
 			}
-		});
+		};
+
+		// callIf: retry only when the engine owns the transaction (new trx per attempt);
+		// when the caller owns it a deadlock has already aborted their trx and retry would be wrong.
+		final Boolean processed = DEADLOCK_RETRY_POLICY.callIf(
```
teosarca: pretty misleading "callIf"

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/DeadlockRetryPolicy.java
teosarca: Done, your way — `wrapIf(boolean condition, TrxCallable<T> callable, Object... context)` returns the callable unchanged when false, or a retry-wrapping TrxCallable when true (`c882f2faa2b`). One correctness note baked in: each retry runs in a FRESH transaction via `callInNewTrx` — a deadlock aborts the current txn (next statement errors `current transaction is aborted`), so retrying in the same txn can't work; the fresh-txn-per-attempt is the legitimate exception to the prefer-runInThreadInheritedTrx rule (commented inline). End-to-end retry verified by unit test. And I've added a coding rule so I implement a reviewer's specified design (or bring evidence) rather than substituting my own — sorry for the back-and-forth.

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/DeadlockRetryPolicy.java
teosarca: Evaluated: yes, package-level works. The only caller outside the class is `DeadlockRetryPolicyTest`, which sits in the same package (`org.adempiere.ad.trx.api`) — narrowed `isDeadlock(...)`, `DEFAULT_MAX_ATTEMPTS` and `DEFAULT_BACKOFF_MILLIS` from `public` to package-private (annotation kept): https://github.com/metasfresh/metasfresh/commit/c97ae55911ff55c5b3b68c907eb60c2655b1d813

backend/de.metas.business/src/main/java/de/metas/document/engine/impl/AbstractDocumentBL.java
teosarca: `callIf` is gone — replaced by `wrapIf(condition, TrxCallable)` exactly as asked in the other thread, with each retry attempt running in a fresh transaction via `ITrxManager.callInNewTrx`: https://github.com/metasfresh/metasfresh/commit/c882f2faa2b7f59b8734f29e28e47a642944a0d2

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/DeadlockRetryPolicy.java
```diff
@@ -0,0 +1,193 @@
...
+	 * @param context   diagnostic context passed to the retry logger
+	 * @return a {@link TrxCallable} — either the original or a retry-wrapped one
+	 */
+	public <T> TrxCallable<T> wrapIf(final boolean condition, @NonNull final TrxCallable<T> callable, final Object... context)
+	{
+		if (!condition)
+		{
+			return callable;
+		}
+		return () -> call(() -> Services.get(ITrxManager.class).callInNewTrx(callable), context);
```
teosarca: why do we wap it in trx manager
also why using callInNewTrx when this is so well documented that is a workaround?!?!!?!!

backend/de.metas.business/src/main/java/de/metas/document/engine/impl/AbstractDocumentBL.java
```diff
@@ -148,11 +151,25 @@ public void doFinally()
 				// put back the transaction which document had initially
 				setTrxName(document.getDocumentModel(), trxName, true /* ignoreIfNotHandled */);
 			}
-		});
+		};
+
+		// wrapIf: retry only when the engine owns the transaction (each retry uses a fresh trx);
```
teosarca: why this comment? (DRY?)

backend/de.metas.business/src/main/java/de/metas/document/engine/impl/AbstractDocumentBL.java
```diff
@@ -148,11 +151,25 @@ public void doFinally()
...
+		// when the caller owns it a deadlock has already aborted their trx and retry would be wrong.
+		final Boolean processed = trxManager.call(trxName, DEADLOCK_RETRY_POLICY.wrapIf(
+				isEngineOwnsTheTransaction(trxName),
+				processCallable,
+				document.getDocumentInfo()));
 
 		return processed != null && processed;
 	}
 
+	/** Returns true when the document engine owns the transaction, i.e. a fresh transaction is opened per attempt. */
```
teosarca: DRY?

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/trx/api/DeadlockRetryPolicy.java
teosarca: Agreed — reworked in https://github.com/metasfresh/metasfresh/commit/8a270cd43ff0d2e9c4a16648ffa48652950a3617: the retry loop now sits OUTSIDE `trxManager.call` at the call site, so each attempt gets its fresh trx from `trxManager.call(trxName, ...)`'s own REQUIRES_NEW semantics (engine-owned trxName). `wrapIf` + the `callInNewTrx` usage are deleted — `DeadlockRetryPolicy` is a pure retry loop with no trx knowledge. (The fresh-trx-per-attempt constraint itself stands: a deadlock aborts the attempt's trx, so the next statement in it would fail with 25P02.)

backend/de.metas.business/src/main/java/de/metas/document/engine/impl/AbstractDocumentBL.java
teosarca: Removed (it duplicated the policy javadoc) — gone with the rework in https://github.com/metasfresh/metasfresh/commit/8a270cd43ff0d2e9c4a16648ffa48652950a3617

backend/de.metas.business/src/main/java/de/metas/document/engine/impl/AbstractDocumentBL.java
teosarca: Removed in https://github.com/metasfresh/metasfresh/commit/8a270cd43ff0d2e9c4a16648ffa48652950a3617 — the method name is self-documenting

backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/adempiere/ad/dao/impl/TypedSqlQuery.java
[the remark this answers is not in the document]
teosarca: Evaluated: handled. `buildSQL` now fails fast when a locking clause is set and the statement is not a plain SELECT — guard at the end of `buildSQL` throws `AdempiereException` for a non-SELECT `selectClause` (the `updateSql(...)` / `deleteDirectlyFrom()` paths Copilot pointed at), and also for UNION and GROUP BY (both invalid with `FOR UPDATE` in PostgreSQL). Added in https://github.com/metasfresh/metasfresh/commit/5363f458ecb0c12db5d8ddc15b13283920506b78; covered by 3 unit tests (`lockWithNonSelectClause/lockWithUnion/lockWithGroupBy_throwsAdempiereException` in `TypedSqlQueryTests`, full-message asserts since https://github.com/metasfresh/metasfresh/commit/01e74780d61f7615ee90fd5047220d3f74294572).

PR #24370, opened 2026-06-02
C_Invoice.DueDate — enforce C_PaymentTerm.IsAllowOverrideDueDate across all write paths

teosarca:
## What

`C_PaymentTerm.IsAllowOverrideDueDate` was only respected by the invoice-candidate aggregation path. This PR enforces it consistently across every `C_Invoice.DueDate` write path, plus closes adjacent robustness gaps found during the analysis:

1. **Completion-time enforcement (interceptor)** — the AFTER_PREPARE `setDueDate` in `de.metas.business` now: fills DueDate when null (as before); when DueDate is set but the payment term has `IsAllowOverrideDueDate='N'`, silently normalizes it to the computed value (logged at INFO). Completion is never blocked. Covers UI edits, REST/EDI writes, and stale drafts.
2. **UI** — `C_Invoice.DueDate` fields become read-only (migration `5805840`, generic on `AD_Column_ID` so customer-side field rows are covered too). Due-date changes go through the dedicated override process, which is now properly gated.
3. **`C_Invoice_OverrideDueDate` process gate** — precondition rejects selections containing invoices whose payment term disallows override (new `IInvoiceDAO.retrieveDocNosWithPaymentTermDisallowingOverride` + AD_Message `DueDateOverrideNotAllowed`, migration `5805830`); `doIt` additionally hard-filters the bulk update to allowing payment terms (preconditions are advisory).
4. **AggregationEngine fail-closed** — `computeOverrideDueDate` now drops the override when no payment term is resolvable (previously fail-open; unreachable in production but reachable in tests).
5. **Org-timezone correctness** — `TimeUtil.asTimestamp(LocalDate)` single-arg (JVM default zone) replaced with the org-zone two-arg form at the DueDate/DunningGrace write sites (`C_Invoice` interceptor, `InvoiceSourceBL`), matching the org-zone read sides.
6. **Dunning grace ordering** — `setDunningGraceIfAutomatic` moved from AFTER_PREPARE to AFTER_COMPLETE: its DueDate dependency is filled by another module's AFTER_PREPARE interceptor and cross-module interceptor order is non-deterministic; when grace fired first it was silently skipped and the grace-eligibility check in dunning-candidate creation was bypassed. Rationale documented in the JavaDoc.

## Migrations

- `5805830_sys_gh29412_AD_Message_DueDateOverrideNotAllowed.sql` — AD_Message, MsgType=E, ErrorCode `DUE_DATE_OVERRIDE_NOT_ALLOWED`, de/en texts.
- `5805840_sys_gh29412_DueDate_field_readonly.sql` — `AD_Field.IsReadOnly='Y'` for all active fields of `C_Invoice.DueDate`.

## Tests

- Unit: `TestFixedDateInvoicedAndDateAcct#test_using_dateDateDueOverrideParam_noPaymentTerm` (fail-closed), `C_Invoice_SetDueDateTest` (5 — interceptor branches + org zone), `C_Invoice_OverrideDueDateTest` (3 — DAO gate incl. no-payment-term edge), `C_Invoice_IsPartialInvoiceTest` updated ctor.
- Cucumber: `invoiceOverrideDueDate.feature` + scenario `S0209_310` — the real `C_Invoice_OverrideDueDate` process invoked per invoice (new stepdef with proper client/role process context); flag=Y invoice updated, flag=N invoice unchanged.
- Local evidence (2026-06-02, deep_tundra_uat infra): de.metas.business 751/751, de.metas.swat.base 422/422, de.metas.dunning 69/69; cucumber invoiceOverrideDueDate 2/2 + dunning 1/1 on final HEAD; both migrations applied + verified.
- A Playwright spec asserting the read-only rendering of the due-date field follows as a commit on this branch before merge (UI counterpart of migration `5805840`).

## Notes for reviewers

- The pay-schedule converter (`InvoicePayScheduleConverter`) was intentionally left JVM-zone-consistent (a write-only org-zone change there would break its own round-trip; coherent treatment is tracked separately).
- `paymentTermDueDate(...)` SQL function intentionally stays (order-based dunning fallback).


backend/de.metas.business/src/main/java/de/metas/invoice/service/impl/AbstractInvoiceDAO.java
```diff
@@ -711,4 +714,45 @@ public Collection<String> retrievePaidInvoiceDocNosForFilter(@NonNull final IQue
...
 				.collect(ImmutableList.toImmutableList());
 	}
-}
\ No newline at end of file
+
+	@Override
+	public Collection<String> retrieveDocNosWithPaymentTermDisallowingOverride(@NonNull final IQueryFilter<org.compiere.model.I_C_Invoice> filter)
+	{
+		final IQuery<I_C_PaymentTerm> disallowingPaymentTerms = queryBL
+				.createQueryBuilder(I_C_PaymentTerm.class)
```
teosarca: how did u evaluate that u can use this table (C_PaymentTerm) in this DAO?

----

rule + review rule: when writting any code that involves IQueryBL pls always make sure (besides that is the right class to use IQueryBL) is the right class to access this table, i.e. C_PaymentTerm

Check if the class u are about to write in is DAO/Repo and is the right DAO cluster for this table.
If you don't know then pls find out and update this class javadoc with cluster DAO/Repo classes and tables. update the same javadoc to all cluster Repo/DAO classes.

If the class is not suitable for that table, then find the repo/dao suitable.
If you don't know, ask human.
After u find out pls update the javadoc for the whole cluster.

teosarca: in this case the right repo for payment term is de.metas.payment.paymentterm.repository.IPaymentTermRepository.
There consider introducing getPaymentTermIdByIsAllowOverrideDueDate(boolean) method

backend/de.metas.business/src/main/java/de/metas/invoice/service/impl/AbstractInvoiceDAO.java
```diff
@@ -711,4 +714,45 @@ public Collection<String> retrievePaidInvoiceDocNosForFilter(@NonNull final IQue
...
+				.map(org.compiere.model.I_C_Invoice::getDocumentNo)
+				.collect(ImmutableList.toImmutableList());
+	}
+
+	@Override
+	public int setDueDateWherePaymentTermAllowsOverride(
+			@NonNull final IQueryFilter<org.compiere.model.I_C_Invoice> filter,
+			@NonNull final Timestamp dueDate)
+	{
+		final IQuery<I_C_PaymentTerm> allowingPaymentTerms = queryBL
```
teosarca: see https://github.com/metasfresh/metasfresh/pull/24370/changes#r3344149637 thread

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoice/C_Invoice_StepDef.java
```diff
@@ -911,6 +921,84 @@ private void updateInvoice(@NonNull final Map<String, String> row)
...
+	{
+		final AdProcessId processId = adProcessDAO.retrieveProcessIdByClass(C_Invoice_OverrideDueDate.class);
+
+		DataTableRows.of(dataTable).forEach(row -> {
+			final StepDefDataIdentifier invoiceIdentifier = row.getAsIdentifier(COLUMNNAME_C_Invoice_ID);
+			final I_C_Invoice invoice = invoiceIdentifier.lookupNotNullIn(invoiceTable);
+			final int invoiceId = invoice.getC_Invoice_ID();
+			final Timestamp overrideDueDate = row.getAsLocalDateTimestamp("OverrideDueDate");
+
+			// In the cucumber Spring-Boot environment Env.getCtx() has AD_Client_ID=0
```
teosarca: pls remove this AI slop crap

pls improve your rules about comments and make sure such a crap is detected and removed.
* it's not timeless
* it contains "Fix:" on a thing that was introduced in this PR... but nevertheless, it shall be timeless
* always when u find yourself writting more than 1-2 lines of comments pls ask yourself
  * am I doing the right tihng, or I am doing the wrong thing and now i am writing an excuse -> if u write an excuse then always choose to do the right tihng and not have to write this
  * am i documenting a discovery i found, e.g. "In the cucumber Spring-Boot environment Env.getCtx() has AD_Client_ID=0" -> if that's the case, then pls suggest to human to create/update a rule/skill instead

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoice/C_Invoice_StepDef.java
```diff
@@ -911,6 +921,84 @@ private void updateInvoice(@NonNull final Map<String, String> row)
...
+					.addParameter("OverrideDueDate", overrideDueDate)
+					.buildAndPrepareExecution()
+					.switchContextWhenRunning()
+					.executeSync()
+					.getResult()
+					.propagateErrorIfAny();
+
+			// Re-load out-of-transaction so subsequent validate steps see the committed DB state
+			// rather than any thread-level snapshot that still holds the pre-process value.
+			final I_C_Invoice freshInvoice = InterfaceWrapperHelper.loadOutOfTrx(invoiceId, I_C_Invoice.class);
```
teosarca: rule: pls avoid using InterfaceWrapperHelper in step defs, prefer DAO/Repo.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/invoice/invoiceOverrideDueDate.feature
```diff
@@ -88,3 +88,45 @@ Feature: Invoice override-due-date from Rechnungsdisposition
     And validate created invoices
       | Identifier   | DocStatus | DueDate    |
       | inv_disallow | CO        | 2026-03-31 |
+
+  @from:cucumber
+  @allure.label.epic:E0340_Invoicing
+  @allure.label.feature:F00700_Invoicing
+  @Id:S0209_310
+  Scenario: C_Invoice_OverrideDueDate process gate — flag=N invoices are skipped; flag=Y invoices are updated
```
teosarca: weird scenario name, what flag=n/Y are u talking about?!?!

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/AggregationEngine.java
```diff
@@ -646,7 +646,11 @@ private LocalDate computeOverrideDueDate(@NonNull final I_C_Invoice_Candidate ic
 			return null;
 		}
 		final PaymentTermId paymentTermId = getC_PaymentTerm_ID(ic);
-		if (paymentTermId != null && !paymentTermRepository.isAllowOverrideDueDate(paymentTermId))
+		// Fail-closed: without a payment term there is nothing to check, so deny the override.
```
teosarca: i still don't get the reason for this comment.
i.e. its obvious u shall not compute it if payment term is missing

backend/de.metas.business/src/main/java/de/metas/invoice/service/impl/AbstractInvoiceDAO.java
teosarca: Fixed in https://github.com/metasfresh/metasfresh/commit/3687488076a + https://github.com/metasfresh/metasfresh/commit/ce36b13036d: the C_PaymentTerm query moved out of InvoiceDAO entirely — `IPaymentTermRepository.getPaymentTermIdsByIsAllowOverrideDueDate(boolean)` now owns it, the process composes (fetches the IDs and passes them to the InvoiceDAO methods, which only receive `Collection<PaymentTermId>`), and short cluster javadocs were added on both clusters (IInvoiceDAO/AbstractInvoiceDAO and IPaymentTermRepository/impl). rule added.

teosarca: Done as suggested in https://github.com/metasfresh/metasfresh/commit/3687488076a + https://github.com/metasfresh/metasfresh/commit/ce36b13036d — `getPaymentTermIdsByIsAllowOverrideDueDate(boolean)` on IPaymentTermRepository (Set-returning variant since multiple terms can match); both callers in C_Invoice_OverrideDueDate go through it.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoice/C_Invoice_StepDef.java
teosarca: Removed in https://github.com/metasfresh/metasfresh/commit/3687488076a — the block is now a single timeless line; the step javadoc is contract-only. The underlying discovery (default cucumber ctx = System client/role, so role-restricted process filters match nothing) was moved into our internal stepdef guidelines as the documented process-invocation pattern, and the comment rules got the 1–2-line excuse/discovery self-check. rule added.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoice/C_Invoice_StepDef.java
teosarca: Fixed in https://github.com/metasfresh/metasfresh/commit/3687488076a — `InterfaceWrapperHelper.loadOutOfTrx` replaced with `invoiceDAO.getByIdOutOfTrx(...)`. rule added.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/invoice/invoiceOverrideDueDate.feature
teosarca: Renamed in https://github.com/metasfresh/metasfresh/commit/3687488076a to business language: "C_Invoice_OverrideDueDate updates the due date only for invoices whose payment term allows overriding it" (flag jargon dropped; @Id tag unchanged).

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/AggregationEngine.java
teosarca: Agreed — comment deleted in https://github.com/metasfresh/metasfresh/commit/3687488076a; the guard stands on its own.

teosarca:
All review comments addressed (commits https://github.com/metasfresh/metasfresh/commit/3687488076a + https://github.com/metasfresh/metasfresh/commit/ce36b13036d; validated: 6/6 + 5/5 unit tests, both cucumber scenarios green locally). If you agree with a reply or consider the discussion resolved, please mark the comment as resolved on GitHub — I cannot mark review comments as resolved from the API; that action is reserved for humans.

backend/de.metas.business/src/main/java/de/metas/invoice/process/C_Invoice_OverrideDueDate.java
```diff
@@ -59,26 +63,30 @@ public ProcessPreconditionsResolution checkPreconditionsApplicable(final @NonNul
...
+
+		final IQueryFilter<I_C_Invoice> selectionFilter = context.getQueryFilter(I_C_Invoice.class);
+
+		final Collection<String> paidInvoiceDocNos = invoiceDAO.retrievePaidInvoiceDocNosForFilter(selectionFilter);
 		if (!paidInvoiceDocNos.isEmpty())
 		{
 			return ProcessPreconditionsResolution.reject(msgBL.getTranslatableMsgText(PAID_INVOICES_MESSAGE, paidInvoiceDocNos));
 		}
+
+		final Set<PaymentTermId> disallowing = paymentTermRepository.getPaymentTermIdsByIsAllowOverrideDueDate(false);
```
teosarca: inline it or give it a fucking normal name!

rule: always name local variable as concise as possible but still descriptive. avoid names like "a","c","disallowing" (for a  Set<PaymentTermId>)

backend/de.metas.business/src/main/java/de/metas/invoice/service/impl/AbstractInvoiceDAO.java
```diff
@@ -711,4 +716,48 @@ public Collection<String> retrievePaidInvoiceDocNosForFilter(@NonNull final IQue
...
+	public Collection<String> retrieveDocNosWithPaymentTermIn(
+			@NonNull final IQueryFilter<org.compiere.model.I_C_Invoice> filter,
+			@NonNull final Collection<PaymentTermId> paymentTermIds)
+	{
+		if (paymentTermIds.isEmpty())
+		{
+			return ImmutableList.of();
+		}
+
+		return queryBL.createQueryBuilder(org.compiere.model.I_C_Invoice.class)
```
teosarca: why FQN?!

backend/de.metas.business/src/main/java/de/metas/invoice/service/impl/AbstractInvoiceDAO.java
```diff
@@ -711,4 +716,48 @@ public Collection<String> retrievePaidInvoiceDocNosForFilter(@NonNull final IQue
...
+				.addOnlyActiveRecordsFilter()
+				.addInArrayFilter(I_C_Invoice.COLUMNNAME_C_PaymentTerm_ID, paymentTermIds)
+				.addFilter(filter)
+				.setLimit(QueryLimit.TEN)
+				.create()
+				.stream()
+				.map(org.compiere.model.I_C_Invoice::getDocumentNo)
```
teosarca: why FQN?

backend/de.metas.business/src/main/java/de/metas/invoice/process/C_Invoice_OverrideDueDate.java
```diff
@@ -35,21 +37,23 @@
...
 import java.util.Collection;
+import java.util.Set;
 
 public class C_Invoice_OverrideDueDate extends JavaProcess implements IProcessPrecondition, IProcessDefaultParametersProvider
 {
 	private final static String PARAM_OVERRIDE_DUE_DATE = "OverrideDueDate";
 	public static final AdMessageKey PAID_INVOICES_MESSAGE = AdMessageKey.of("Invoices_already_paid");
+	public static final AdMessageKey DUE_DATE_OVERRIDE_NOT_ALLOWED_MESSAGE = AdMessageKey.of("DueDateOverrideNotAllowed");
 	@Param(parameterName = PARAM_OVERRIDE_DUE_DATE, mandatory = true)
 	private Timestamp p_OverrideDueDate;
```
teosarca: LocalDate

backend/de.metas.business/src/main/java/de/metas/invoice/service/IInvoiceDAO.java
```diff
@@ -204,4 +210,34 @@ List<I_C_LandedCost> retrieveLandedCosts(I_C_InvoiceLine invoiceLine,
...
+	 *
+	 * @param filter         the query filter that describes the process selection
+	 * @param paymentTermIds the payment-term IDs to update for; an empty collection yields 0 updates
+	 * @param dueDate        the new due date to set
+	 * @return number of invoice records updated
+	 */
+	int setDueDateWherePaymentTermIn(
+			IQueryFilter<org.compiere.model.I_C_Invoice> filter,
+			Collection<PaymentTermId> paymentTermIds,
+			Timestamp dueDate);
```
teosarca: LocalDate

teosarca: rule: avoid java.sqlTimestamp, java.util.Date in public method signatures. Use the new java java time api

backend/de.metas.business/src/main/java/de/metas/payment/paymentterm/repository/impl/PaymentTermRepository.java
```diff
@@ -266,4 +272,16 @@ public boolean isAllowOverrideDueDate(@NonNull final PaymentTermId paymentTermId
 	{
 		return getById(paymentTermId).isAllowOverrideDueDate();
 	}
+
+	@Override
+	@NonNull
+	public Set<PaymentTermId> getPaymentTermIdsByIsAllowOverrideDueDate(final boolean isAllowOverrideDueDate)
+	{
+		return ImmutableSet.copyOf(
```
teosarca: rule: when changig DAO/Repo and adding retrieval methods, always but always first check the class if it has some kind of caching. If it does, consider using it
in this case, we have getIndexedPaymentTerms() which returns PaymentTermMap (full cache) .
u shall use that and introduce PaymentTermMap.getPaymentTermIdsByIsAllowOverrideDueDate(boolean)

backend/de.metas.business/src/test/java/de/metas/invoice/service/impl/InvoiceDAO_DueDateOverrideTest.java
```diff
@@ -0,0 +1,234 @@
...
+
+import java.sql.Timestamp;
+import java.time.Instant;
+import java.util.Collection;
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.compiere.model.I_C_Invoice.COLUMNNAME_C_Invoice_ID;
+
+/**
```
teosarca, line 49: if is about testing IInvoiceDAO why this class and in this package?!?

backend/de.metas.business/src/main/java/de/metas/invoice/process/C_Invoice_OverrideDueDate.java
teosarca: Inlined in https://github.com/metasfresh/metasfresh/commit/daf712a0738 — both locals gone, the repo call sits directly in the DAO-call argument. rule added.

backend/de.metas.business/src/main/java/de/metas/invoice/service/impl/AbstractInvoiceDAO.java
teosarca: Fixed in https://github.com/metasfresh/metasfresh/commit/daf712a0738 — the new methods now use the imported `de.metas.adempiere.model.I_C_Invoice` end-to-end (process filter ↔ interface ↔ impl); the two leftover FQN sites (Timestamp/TimeUtil in getParameterDefaultValue) cleaned in https://github.com/metasfresh/metasfresh/commit/10653c14f76.

backend/de.metas.business/src/main/java/de/metas/invoice/service/impl/AbstractInvoiceDAO.java
teosarca: Same fix — https://github.com/metasfresh/metasfresh/commit/daf712a0738 + https://github.com/metasfresh/metasfresh/commit/10653c14f76, no FQN remains in these paths.

backend/de.metas.business/src/main/java/de/metas/invoice/process/C_Invoice_OverrideDueDate.java
teosarca: Done in https://github.com/metasfresh/metasfresh/commit/daf712a0738 — `@Param private LocalDate p_OverrideDueDate` (the @Param framework supports LocalDate natively via ProcessClassParamInfo), and getParameterDefaultValue returns LocalDate.

backend/de.metas.business/src/main/java/de/metas/invoice/service/IInvoiceDAO.java
teosarca: Done in https://github.com/metasfresh/metasfresh/commit/daf712a0738 — `setDueDateWherePaymentTermIn(..., LocalDate dueDate)`; conversion to the column type happens at the persistence boundary inside the DAO.

teosarca: Applied in https://github.com/metasfresh/metasfresh/commit/daf712a0738 (LocalDate across the public signatures). rule added — the existing java-time rule now names java.util.Date and public method signatures explicitly.

backend/de.metas.business/src/main/java/de/metas/payment/paymentterm/repository/impl/PaymentTermRepository.java
teosarca: Done in https://github.com/metasfresh/metasfresh/commit/daf712a0738 — new `PaymentTermMap.getPaymentTermIdsByIsAllowOverrideDueDate(boolean)` filters the cached terms; `PaymentTermRepository` delegates to `getIndexedPaymentTerms()` — no query anymore. rule added (check for existing caching before adding DAO/Repo retrieval methods).

backend/de.metas.business/src/test/java/de/metas/invoice/service/impl/InvoiceDAO_DueDateOverrideTest.java
teosarca: Right — it tests IInvoiceDAO. Moved + renamed in https://github.com/metasfresh/metasfresh/commit/daf712a0738: `de.metas.invoice.service.impl.InvoiceDAO_DueDateOverrideTest` (sibling-convention name, impl package); all 6 test methods kept.

teosarca:
Review round 2 addressed (commits https://github.com/metasfresh/metasfresh/commit/daf712a0738 + https://github.com/metasfresh/metasfresh/commit/10653c14f76; validated: InvoiceDAO_DueDateOverrideTest 6/6, C_Invoice_SetDueDateTest 5/5, both cucumber scenarios green locally). Please mark resolved where you agree — resolution stays a human action.

backend/de.metas.business/src/main/java/de/metas/invoice/process/C_Invoice_OverrideDueDate.java
```diff
@@ -88,11 +94,12 @@ public Object getParameterDefaultValue(final IProcessDefaultParameter parameter)
 	{
 		if (PARAM_OVERRIDE_DUE_DATE.equals(parameter.getColumnName()))
 		{
-			return queryBL.createQueryBuilder(I_C_Invoice.class)
+			final Timestamp ts = queryBL.createQueryBuilder(I_C_Invoice.class)
 					.addOnlyActiveRecordsFilter()
 					.addFilter(getProcessInfo().getQueryFilterOrElseFalse())
 					.create()
 					.first(I_C_Invoice.COLUMNNAME_DueDate, Timestamp.class);
```
teosarca: why not using LocalDate.class instead of Timestamp.class.
also pls move this to IInvoiceDAO, a method with IQueryFIlter as param

rule: avoid IQueryBL outside repo/dao!

teosarca: Done in https://github.com/metasfresh/metasfresh/commit/59caef4c8ac — `IInvoiceDAO.retrieveFirstDueDate(IQueryFilter)` (+2 tests); the process now has zero IQueryBL (field+import gone). On `LocalDate.class` directly: tried it — the POJO query engine throws ClassCastException (returns the raw Timestamp), so the conversion stays inside the DAO at the persistence boundary. rule added.

PR #24376, opened 2026-06-03
Shipment and Invoice exports: parameter descriptions

metas-rc:


metas-rc:
@claude review opus

metas-rc:
@claude 

Thanks for the review. Addressing each finding:

**Issue 1 — EntityType inconsistency (`'D'` vs `'de.metas.dao.selection'`):** Not changed — this is intentional and matches the established pattern. AD_Element 543188 (Limit) belongs to EntityType `'D'`; AD_Element 576802 (Offset) belongs to `'de.metas.dao.selection'`. The AD_Process_Para record inherits the EntityType of its AD_Element. The existing `Available_For_Sales_JSON` migration uses the exact same split for the identical Limit/Offset elements — aligning them to the same value would be incorrect.

**Issue 2 — Asymmetric COALESCE (invoices vs shipments):** The COALESCE on `historical_invoices_json_v` was intentionally dropped. The existing test scenario S0474_030 in `historicalInvoicesExport.feature` deliberately creates two invoices for the same BPartner — one with ExternalSystem, one without — and expects only the external-system one to be returned. That test relies on the `NULL ILIKE '%'` = false behaviour to exclude the second invoice. Applying the COALESCE there would break that test, and fixing the test to be more explicit is out of scope for this PR. The shipment view COALESCE is correct and intentional (no existing test relies on the NULL exclusion there).

**Issue 3 — Missing English `Help` in `en_US` translations:** Fixed. Both `en_US` Trl rows for `Historical_Shipments_JSON` (585488) and `Historical_Invoices_JSON` (585485) now include the English `Help` text. Commit: https://github.com/metasfresh/metasfresh/commit/18950ca1146

**Issue 4 — Behavioral change documentation:** Acknowledged. Both changes (processed-only filter and ORDER BY direction) are driven by the colleague's migration 5806060 which was merged into the base branch, not by this PR directly.

**Issue 5 — `UpdatedGE` and `ExternalSystemCode` missing descriptions:** These parameters were created by earlier migrations and are out of scope for this PR. Their descriptions (or absence thereof) are pre-existing.

PR #24387, opened 2026-06-03
M_HU_Attribute picking performance: covering index + weekly destroyed-HU attribute archival

metas-ts:
## Problem

The picking "find pickable source HUs" search (`HUStorageQuery.createQueryBuilder_for_M_HU_Storages`, used by the `WEBUI_Picking_ForcePickTo*` processes and source-HU retrieval) builds a correlated `EXISTS` over `M_HU_Attribute` with the predicate `(M_HU_ID, M_Attribute_ID, ValueNumber, IsActive='Y')`, probed once per candidate HU.

On large datasets this is the dominant cost of a pick:
- the existing `m_hu_attribute_parent_index (M_HU_ID, M_Attribute_ID)` does **not** cover `ValueNumber`, so every probe heap-fetches to evaluate it; and
- the vast majority of `M_HU_Attribute` rows belong to **Destroyed** HUs that can never be picked, yet still bloat the index/heap and push the working set out of cache → millions of cold random reads → multi-second picks.

## Changes

1. **Partial covering index** `m_hu_attribute_hu_attr_valnum (M_HU_ID, M_Attribute_ID, ValueNumber) WHERE IsActive='Y'` so the `EXISTS` is satisfied index-only.
   - Created **non-concurrently** in the migration on purpose: `CREATE INDEX CONCURRENTLY` cannot run inside the migration tool's single transaction (cf. `de.metas.dlm` `5453734_sys_gh489_DDL.sql`). It is `IF NOT EXISTS`, instant on small/CI DBs; large production DBs pre-build it `CONCURRENTLY` off-peak so the migration statement is a no-op and never takes an `ACCESS EXCLUSIVE` lock on the picking-critical table.

2. **`dlm.deactivate_destroyed_hu_attributes(p_max_hus)`** + a **weekly `AD_Scheduler`** (Sun 02:00, via `de.metas.process.ExecuteUpdateSQL`, mirroring the `archive_c_queue_data` pattern). It sets `IsActive='N'` on `M_HU_Attribute` rows whose parent `M_HU` is **Destroyed (`HUStatus='D'`) only** — keeping the active slice (and the partial index above) small as Destroyed HUs accumulate.
   - Scope is Destroyed-only by design: Active HUs must stay pickable, and Shipped HUs can be reactivated by shipment reversal (would come back attribute-less); Destroyed HUs have no un-destroy path.
   - The function self-provisions its `dlm` schema + a reversibility log table (`m_hu_attribute_deactivated`, with `deactivated_at`), is bounded per run (one small transaction), and references `public.*` explicitly so DLM search-path view-shadowing can't redirect it.

## Notes
- The one-time historical backlog (deactivating already-Destroyed HUs' attributes on a large existing DB) is an operational DBA runbook, not part of this migration — a bulk `UPDATE` must not run inside every rollout, and Destroyed HUs re-accumulate (hence the recurring scheduler).
- Migrations validated locally (all 3 scripts apply cleanly; index, function, `AD_Process`, `AD_Scheduler` and log table verified present).

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/sql/postgresql/ddl/functions/deactivate_destroyed_hu_attributes.sql
+++ b/backend/de.metas.handlingunits.base/src/main/sql/postgresql/ddl/functions/deactivate_destroyed_hu_attributes.sql
@@ -25,7 +25,9 @@ AS $func$
 DECLARE
     v_deactivated integer;
 BEGIN
-    -- self-provisioning (idempotent)
+    -- self-provisioning (idempotent): schema + log table. The log table may already exist from an
+    -- earlier run that predated the deactivated_at column, so add it if missing (existing rows
+    -- backfill to now()).
     CREATE SCHEMA IF NOT EXISTS dlm;
     CREATE TABLE IF NOT EXISTS dlm.m_hu_attribute_deactivated (
         m_hu_attribute_id numeric(10,0) PRIMARY KEY,
--- a/backend/de.metas.handlingunits.base/src/main/sql/postgresql/ddl/functions/deactivate_destroyed_hu_attributes.sql
+++ b/backend/de.metas.handlingunits.base/src/main/sql/postgresql/ddl/functions/deactivate_destroyed_hu_attributes.sql
@@ -59,6 +59,7 @@ BEGIN
     )
     SELECT count(*) INTO v_deactivated FROM upd;
 
+    RAISE NOTICE 'dlm.deactivate_destroyed_hu_attributes: % attribute rows deactivated', v_deactivated;
     RETURN COALESCE(v_deactivated, 0);
 END;
 $func$;
--- a/backend/de.metas.handlingunits.base/src/main/sql/postgresql/ddl/functions/deactivate_destroyed_hu_attributes.sql
+++ b/backend/de.metas.handlingunits.base/src/main/sql/postgresql/ddl/functions/deactivate_destroyed_hu_attributes.sql
@@ -1,0 +1,64 @@
+-- dlm.deactivate_destroyed_hu_attributes(p_max_hus)
+-- Soft-archive (IsActive='N') M_HU_Attribute rows of DESTROYED HUs so the picking
+-- source-HU search (HUStorageQuery) stays fast and the partial index m_hu_attribute_hu_attr_valnum
+-- (WHERE IsActive='Y') stays small / cache-resident.
+--
+-- Self-provisioning (idempotent): creates the dlm schema, the reversibility log table, and its
+-- deactivated_at timestamp column if missing.
+--
+-- Bounded per run by p_max_hus -> one small transaction, suitable for a weekly AD_Scheduler. The
+-- one-time large backlog is handled separately by a manual, per-batch-commit operational runbook.
+--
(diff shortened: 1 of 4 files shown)


backend/de.metas.handlingunits.base/src/main/sql/postgresql/ddl/functions/deactivate_destroyed_hu_attributes.sql
```diff
@@ -0,0 +1,64 @@
...
+
+CREATE OR REPLACE FUNCTION dlm.deactivate_destroyed_hu_attributes(p_max_hus integer DEFAULT 100000)
+    RETURNS integer
+    LANGUAGE plpgsql
+AS $func$
+DECLARE
+    v_deactivated integer;
+BEGIN
+    -- self-provisioning (idempotent)
+    CREATE SCHEMA IF NOT EXISTS dlm;
```
metas-ts, line 31: use raise notice to provide logging

backend/de.metas.handlingunits.base/src/main/sql/postgresql/system/70-de.metas.handlingunits/5806000_sys_M_HU_Attribute_picking_source_hu_index.sql
```diff
@@ -0,0 +1,22 @@
...
+-- the EXISTS be satisfied
+-- index-only and (together with the destroyed-HU attribute soft-archive) keeps the active slice
+-- cache-resident.
+--
+-- NON-CONCURRENT ON PURPOSE: CREATE INDEX CONCURRENTLY cannot run inside the migration tool's single
+-- transaction / multi-command string (see de.metas.dlm 5453734_sys_gh489_DDL.sql:148-151). On small
+-- and CI databases the statement below is instant. On LARGE production databases the index MUST be
+-- pre-created CONCURRENTLY off-peak BEFORE this rollout runs (via an operational DBA
+-- runbook) so that this IF NOT EXISTS statement is a no-op and does not
+-- take an ACCESS EXCLUSIVE lock on the picking-critical M_HU_Attribute table.
```
metas-ts:
```suggestion
```

i don'T see the need for this

backend/de.metas.handlingunits.base/src/main/sql/postgresql/ddl/functions/deactivate_destroyed_hu_attributes.sql
metas-ts: Done — added `RAISE NOTICE 'dlm.deactivate_destroyed_hu_attributes: % attribute rows deactivated', v_deactivated;` before RETURN (in both the DDL file and the migration). https://github.com/metasfresh/metasfresh/commit/8edaa0bcf27

backend/de.metas.handlingunits.base/src/main/sql/postgresql/system/70-de.metas.handlingunits/5806000_sys_M_HU_Attribute_picking_source_hu_index.sql
metas-ts: Removed — trimmed the header to a two-line purpose comment. The prod-side note (pre-build CONCURRENTLY off-peak) lives in the operational DBA runbook, not the migration. https://github.com/metasfresh/metasfresh/commit/8edaa0bcf27

PR #24400, opened 2026-06-04
Qty reservation reconcile - delete zero-Qty reservations on sales-order completion

adi-stefan:
## What

Reservation reconcile-on-completion now **deletes reservation rows whose `Qty` is 0** instead of leaving them behind.

Issue: https://github.com/metasfresh/me03/issues/29261 (F19012)

## Why

A reservation row can carry `Qty(CU)=0` but a positive `QtyTU` — e.g. a planned-supply row reserved against stock that is not yet on hand (`MaterialCockpitV2RowVO.computeQtyCUToReserve` returns 0 CU when `qtyStock <= 0`). Such a row reserves nothing, yet:

- it **inflates the reserved-`QtyTU` total** (a line showing OH `77 CU / 7 TU` plus a phantom PS `0 CU / 3 TU` reads as **10 TU reserved** instead of 7), and
- it **lingers forever**: `ReconcileQtyReservationsCommand` shrinks each line's reservations down to the line's `QtyOrdered` on sales-order completion, but it **skipped** zero-`Qty` rows (`rowQty.signum() <= 0` guard), so it neither zeroed their `QtyTU` nor removed them.

Reported on a `swift_eagle_release` build after the customer's reverse → reactivate → reduce-line → re-complete workflow.

## How

`ReconcileQtyReservationsCommand.execute()` now collects every reservation row whose **final `Qty` is 0** and deletes it:

- **pre-existing phantoms** — scanned on every line that has reservations, independent of whether the line is over-reserved (so a phantom on a cancelled `QtyOrdered=0` line is cleaned up too);
- **rows shrunk fully to zero** — a row reduced to 0 by the over-reservation sweep is deleted rather than written as `Qty=0`.

Deletion uses the new `QtyReservationRepository.deleteByIds(Set<QtyReservationId>)` — the **same physical SQL `delete()` mechanism** as `deleteReservation`, but targeted by `M_QtyReservation_ID` so it never removes a sibling non-zero reservation sharing the same supply-type bucket. `QtyReservationService.reconcileToOrderedQty` invalidates the affected lines' shipment schedules afterwards, consistent with `deleteReservation`. The delivered-clamp guarantees `Qty=0 ⟹ QtyDelivered=0`, so deletion never discards delivered history.

Also folded in a small persistence-layer cleanup: the PLANNED_SUPPLY/ON_HAND lookup that the command previously ran via `IQueryBL` moved into `QtyReservationRepository.getIsPlannedSupplyByIds(...)`, so the command no longer issues DB queries directly (behaviour-neutral — identical query).

No migration / AD metadata changes. Scope: reconcile cleanup only — the reservation-creation path (`MakeQtyReservationCommand`) is left untouched.

## Tests

Three RED-first unit cases added to `ReconcileQtyReservationsCommandTest`:

- `deletesPreExistingZeroQtyReservation_keepingReservedTuCorrect` — over-reserved OH `110/10` + phantom PS `0/3`, line reduced to 77 ⇒ PS deleted, OH `77/7`, reserved `QtyTU` total = 7.
- `deletesReservation_whenShrunkToZero` — OH `60` + PS `40`, line reduced to 60 ⇒ PS fully removed.
- `deletesPhantomFromCancelledLine_whenQtyOrderedIsZero` — `0/3` PS phantom on a cancelled (`QtyOrdered=0`) line ⇒ phantom deleted (the scan runs before the over-reservation guard).

Local: `Tests run: 21, Failures: 0` (`QtyReservationTest`, `QtyReservationRepositoryTest`, `ReconcileQtyReservationsCommandTest`). Cucumber `qtyReservation` regression verified on CI.

### DIFF
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/qty_reservation/QtyReservationRepository.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/qty_reservation/QtyReservationRepository.java
@@ -16,6 +16,8 @@ import org.jetbrains.annotations.NotNull;
 import org.springframework.stereotype.Repository;
 
 import java.math.BigDecimal;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Set;
 import java.util.function.UnaryOperator;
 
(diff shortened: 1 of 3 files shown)


adi-stefan:
@claude review

PR #24402, opened 2026-06-04
MD_Stock_Update_From_M_HUs: batched drain processing to avoid OOME on large backlogs

teosarca:
## What

Rewrite `MD_Stock_Update_From_M_HUs.doIt()` from a single all-rows `list()` (which OOMEs on a large backlog — the reason it was deactivated in 2021) into a **batched drain loop**.

- Each iteration fetches the top `BATCH_SIZE` (500) still-diverging `MD_Stock_From_HUs_V` rows ordered by the view's grouping columns — **no OFFSET**: correcting a batch zeroes its `QtyOnHandChange`, so the rows leave the `<>0` set and the shrinking top-N drains naturally (OFFSET would skip rows).
- Each batch is corrected in **its own committed transaction**, so batch N is visible to batch N+1's re-read and the per-batch `StockChangedEvent`s flush instead of buffering for the whole run.
- A `MAX_LOOPS` (100,000) backstop terminates + logs if the set never empties (e.g. concurrent stock changes — the process is designed to run in parallel — or a UOM-rounding epsilon).

Memory stays flat (~`BATCH_SIZE`) regardless of backlog size. Correction behaviour and emitted events are otherwise unchanged.

## Why

Re-enabling this corrective scheduler (done for ic114) is only durable if the process can't OOME on an accumulated backlog. This makes it safe by construction.

## Tests

Loop-contract unit tests added (drains to empty / forwards every row / `MAX_LOOPS` backstop terminates) via injectable fetch+process seams; the DB-backed view stays out of scope for unit tests. Green locally (3/3).

### DIFF
--- a/backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/stock/process/MD_Stock_Update_From_M_HUs.java
+++ b/backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/stock/process/MD_Stock_Update_From_M_HUs.java
@@ -16,9 +16,11 @@ import de.metas.quantity.Quantity;
 import de.metas.quantity.Quantitys;
 import de.metas.uom.IUOMConversionBL;
 import de.metas.uom.UomId;
+import de.metas.util.Loggables;
 import de.metas.util.Services;
 import lombok.NonNull;
 import org.adempiere.ad.dao.IQueryBL;
+import org.adempiere.ad.dao.QueryLimit;
 import org.adempiere.service.ClientId;
 import org.adempiere.warehouse.WarehouseId;
 import org.compiere.SpringContextHolder;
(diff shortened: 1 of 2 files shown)


backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/stock/process/MD_Stock_Update_From_M_HUs.java
```diff
@@ -50,42 +53,184 @@
...
+	 * the top-N walks the shrinking set down to empty. An OFFSET would instead skip {@code batchSize}
+	 * real rows on each iteration and never correct them.
+	 */
+	private List<I_MD_Stock_From_HUs_V> retrieveHuData(final int batchSize)
+	{
+		return queryBL
+				.createQueryBuilder(I_MD_Stock_From_HUs_V.class)
+				.addNotEqualsFilter(I_MD_Stock_From_HUs_V.COLUMNNAME_QtyOnHandChange, ZERO)
+				.orderBy()
+				.addColumnAscending(I_MD_Stock_From_HUs_V.COLUMNNAME_AD_Client_ID)
```
teosarca: u could directly call orderBy(I_MD_Stock_From_HUs_V.COLUMNNAME_AD_Client_ID) and u can do this for each  column instead of 
.orderBy()
.addColumnAscending(....)
.endOrderBy()

backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/stock/process/MD_Stock_Update_From_M_HUs.java
```diff
@@ -50,42 +53,184 @@
...
+	 * <p>
+	 * {@code runInNewTrx} is required (and not the usual anti-pattern here): the drain only works if
+	 * batch N's {@code MD_Stock} writes are committed and visible when {@link #retrieveHuData(int)}
+	 * re-reads the view for batch N+1. The per-batch commit also flushes that batch's
+	 * {@code StockChangedEvent}s (they fire after-commit) instead of buffering every event for the
+	 * whole run, which is the second memory sink the OOME-safe rewrite has to bound.
+	 */
+	private void handleBatchInOwnTrx(@NonNull final List<I_MD_Stock_From_HUs_V> batch)
+	{
+		trxManager.runInNewTrx(() -> createAndHandleDataUpdateRequests(batch));
```
teosarca: still don;t see the benefit of wrapping in runInNewTrx. more, didn't we agreed that runInNewTrx must be avoided?!

backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/stock/process/MD_Stock_Update_From_M_HUs.java
teosarca: Done — switched to the direct `.orderBy(I_MD_Stock_From_HUs_V.COLUMNNAME_…)` chained form (one call per grouping column), same 6 columns/order. Fixed in https://github.com/metasfresh/metasfresh/commit/d58e2269dae53d13a073837c0cd3a114bca3c119

backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/stock/process/MD_Stock_Update_From_M_HUs.java
teosarca: Agreed — removed the `runInNewTrx` wrap (and the `handleBatchInOwnTrx` method + `ITrxManager` field). Since `doIt()` is `@RunOutOfTrx` there is no ambient transaction, so each `handleDataUpdateRequest()` commits immediately and fires its `StockChangedEvent` immediately — the per-batch wrap was unnecessary for the drain's cross-batch visibility and contradicted the avoid-`runInNewTrx` convention. The batch processor now calls `createAndHandleDataUpdateRequests` directly. Fixed in https://github.com/metasfresh/metasfresh/commit/d58e2269dae53d13a073837c0cd3a114bca3c119

PR #24423, opened 2026-06-05
EDI: route DESADV/INVOIC by C_BPartner_Location_ID (new C_BPartner_EDI_Setting table)

metas-ts:
## Summary

Move EDI **DESADV / INVOIC routing configuration** off `C_BPartner` columns into a new child table **`C_BPartner_EDI_Setting`**, keyed by `(C_BPartner_ID, optional C_BPartner_Location_ID)`.

This lets the routing decision (legacy replication interface vs. external system) be made per **ship-to location**, not only per business partner. For a given document, the config is resolved to the **most-specific** setting row for its effective `(partner, location)`: an exact-location row, else the partner-default (location-less) row.

### Data model / Application Dictionary
- New table `C_BPartner_EDI_Setting` + an "EDI Settings" tab on the BPartner window. Unique index `(C_BPartner_ID, COALESCE(C_BPartner_Location_ID, 0))` (≤1 partner-default row + ≤1 row per location).
- Migration: copy existing `C_BPartner` EDI config into one partner-default row per configured partner; then drop the 9 old EDI columns from `C_BPartner` and clean up their `AD_Field` / `AD_UI_*` rows across **all** windows (dynamic, system-independent — also covers custom/override windows). Shared `AD_Element` / `AD_Reference` / `AD_Val_Rule` are preserved (reused by the new table).
- The 3 EDI export views (`M_InOut_Export_EDI_DESADV_JSON_V`, `edi_cctop_000_v`, `edi_cctop_invoic_v`) now resolve the recipient GLN from `C_BPartner_EDI_Setting` via the same exact-location → partner-default coalesce.

### Java
- `EDIBPartnerConfigRepository` / `EDIBPartnerConfigService` resolve config by `BPartnerLocationId` (previously `BPartnerId`).
- New `getEffectiveDropshipLocationId(...)` on `IInOutBL` / `IOrderBL` / `DesadvBL`.
- `IExport.getBPartnerLocationId()`; `EDIWorkpackageProcessor`, the single-`M_InOut` enqueue process, and the `C_Order` validator now route by location.
- The `C_BPartner` EDI GLN-validation interceptor moved to a `C_BPartner_EDI_Setting` interceptor.

### Tests
- Unit test for location-based resolution (exact-location → partner-default fallback).
- Cucumber: a `C_BPartner_EDI_Setting` step definition (idempotent upsert, accepts identifiers and raw IDs); the existing EDI feature files migrated to the new setup; a new end-to-end scenario proving that the **same business partner with two ship-to locations** routes its two DESADVs differently (one location → external system, the other → replication interface).

## Integration notes
- Repos that **override** any of the 3 EDI export views above must be re-checked, since the views' GLN source changed from `C_BPartner` to `C_BPartner_EDI_Setting`.
- Follow-up (tracked separately): a reusable DB function to clone the new "EDI Settings" tab into BPartner override windows.


backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
```diff
@@ -933,6 +933,22 @@ public BPartnerId getEffectiveDropshipPartnerIdOrNull(@NonNull final I_C_Order o
 		return BPartnerId.ofRepoIdOrNull(orderRecord.getC_BPartner_ID());
 	}
 
+	@Override
+	@NonNull
+	public BPartnerLocationId getEffectiveDropshipLocationId(@NonNull final I_C_Order orderRecord)
+	{
+		if (orderRecord.isDropShip())
```
metas-ts: sorry about the  column-naming, but for **sales orders**:
- `isDropShip` doesn't matter
- the record's `DropShip_BPartner_ID` and `DropShip_Location_ID` - **if set** - are the ultimate consignee (CU).
- if  `DropShip_BPartner_ID` and `DropShip_Location_ID` are not, set, the UC is the record's `C_BPartner_ID` and `C_BPartner_Location_ID`

backend/de.metas.edi/src/main/java/de/metas/edi/model/validator/C_BPartner_EDI_Setting.java
```diff
@@ -0,0 +1,81 @@
...
+	{
+		final boolean ediRecipient = setting.isEdiDesadvRecipient() || setting.isEdiInvoicRecipient();
+		if (!ediRecipient)
+		{
+			return;
+		}
+
+		final StringBuilder causes = new StringBuilder();
+
+		if (setting.isEdiDesadvRecipient() && Check.isBlank(setting.getEdiDesadvRecipientGLN()))
```
metas-ts, line 66: good point. did you also add the corresponding mandatory-logic to to the GLN-fields?

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
metas-ts: Fixed in https://github.com/metasfresh/metasfresh/commit/3d6d9ac2c5a — dropped the `isDropShip` gate; `getEffectiveDropshipLocationId` now coalesces `DropShip_BPartner_ID`+`DropShip_Location_ID` (if set) → `C_BPartner_ID`+`C_BPartner_Location_ID` regardless of the flag, matching `InOutBL`/`DesadvBL`. Test updated (the flag-off-but-DropShip-set case now expects the DropShip location). Thanks for the correction!

backend/de.metas.edi/src/main/java/de/metas/edi/model/validator/C_BPartner_EDI_Setting.java
metas-ts: Good point — they didn't (the old `C_BPartner` GLN fields relied on this interceptor only). Added in https://github.com/metasfresh/metasfresh/commit/7879993248e : `EdiDesadvRecipientGLN` → `MandatoryLogic = @IsEdiDesadvRecipient@='Y'`, `EdiInvoicRecipientGLN` → `@IsEdiInvoicRecipient@='Y'` (mirroring the `Edi*_ExternalSystem_Config_ID` MandatoryLogic on the same table). Kept the interceptor as the combined business-validation with a single clear message.

PR #24433, opened 2026-06-05
Ensure distinct HU ids in `PickingJobReopenCommand` to prevent error during initialization

adi-stefan:


adi-stefan:

PR #24463, opened 2026-06-07
Multiple Levels of Payment — fix flaky split-payment pay-schedule DueAmt (read payment term consistently) + fail-loud guard

teosarca:
## Problem

`splitPaymentLCPaidAfterInvoice.feature` (split-payment / Multiple Levels of Payment) fails **intermittently** in CI with `[DueAmt] expected 3000.00 but was 10000` on the first pay-schedule check right after order completion, passing on rerun.

The flake is the visible symptom of a real, production-latent defect:

- Pay-schedule creation reads the order's payment term through a process-wide, single-key `PaymentTermRepository` CCache that holds *all* terms+breaks as one shared snapshot. While a multi-break term is being constructed (breaks inserted one at a time), a concurrent reader can refill that cache with a half-built snapshot containing only the first break, and the model-cache eviction can lose the Guava load-vs-invalidate race — leaving the stale partial snapshot live.
- `PaymentTerm.spreadByBreaks` gives the **last/sole** break the remainder, so a term read with only the LC (30%) break hands it the full 10000 instead of 30%×10000=3000.
- `OrderPayScheduleCreateCommand.execute0` guarded on `isComplex` but **not** `isValid`, so an invalid/partial term (breaks ≠ 100%) was spread silently instead of failing — a silent wrong `DueAmt` that is latent in production wherever a partial payment-term snapshot reaches pay-schedule creation.

## Fix

1. **Read-consistency at the root (de-flake).** `OrderPayScheduleService.extractContext` now loads the payment term via a fresh **in-transaction** load (`getByIdInTrx` → `PaymentTermLoaderAndSaver.loadById`) that reads committed DB state, instead of the shared cache. This guarantees a complete, committed snapshot at the only point where pay-schedule correctness depends on it. The API name mirrors the established `invoiceDAO.getByIdInTrx` convention.
2. **Fail-loud safety net.** `OrderPayScheduleCreateCommand.execute0` now rejects a complex-but-invalid payment term (breaks ≠ 100%) with a clear exception instead of silently mis-spreading.

## Tests

- `OrderPayScheduleCreateCommandTest#invalidComplexTerm_failsLoud` — deterministic unit test (RED→GREEN) for the fail-loud guard.
- `PaymentTermStaleCacheReproTest` — deterministic reproduction: poisons the shared cache to a single-break snapshot while both breaks are committed in DB, proving the old cached read yields `DueAmt=10000` (size 1) while the new in-trx read yields `DueAmt=3000` (size 2).
- The full `de.metas.business` pay-schedule package (98 tests) passes; `splitPaymentLCPaidAfterInvoice.feature` passes locally (regression — the CI-concurrency race does not reproduce single-threaded locally).

The `spreadByBreaks` remainder behaviour for valid terms is unchanged.


backend/de.metas.business/src/test/java/de/metas/payment/paymentterm/repository/impl/PaymentTermStaleCacheReproTest.java
```diff
@@ -0,0 +1,189 @@
...
+
+			final Map<Integer, Object> breakRecords = cachedObjects.get(I_C_PaymentTerm_Break.Table_Name);
+			final Object liveRecord = breakRecords.get(breakRecord.getC_PaymentTerm_Break_ID());
+			InterfaceWrapperHelper.setValue(liveRecord, I_C_PaymentTerm_Break.COLUMNNAME_IsActive, true);
+		}
+		catch (final ReflectiveOperationException e)
+		{
+			throw new RuntimeException("Failed to activate break record in-store without cache reset", e);
+		}
+	}
```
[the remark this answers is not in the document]
teosarca: Applied in https://github.com/metasfresh/metasfresh/commit/41cfe87cd72. Replaced the private-field (`cachedObjects`) reflection with the supported `POJOLookupMap.setCopyOnSave(false)` test seam: it hands back the LIVE stored record (not a defensive copy), so flipping `IsActive` is visible to later reads without a `saveRecord`/cache-reset — exactly the stale-cache state the test needs. Previous flag restored in a finally block; the `java.lang.reflect.Field`/`java.util.Map` imports are gone. Test still passes locally.

backend/de.metas.business/src/main/java/de/metas/payment/paymentterm/repository/IPaymentTermRepository.java
```diff
@@ -44,6 +44,15 @@ public interface IPaymentTermRepository extends ISingletonService
 	@NonNull
 	PaymentTerm getById(PaymentTermId paymentTermId);
 
+	/**
+	 * Loads the payment term together with its breaks fresh from the committed DB state, in the current
+	 * (thread-inherited) transaction, bypassing the process-wide cache. Use this where a complete,
+	 * consistent break set is required (e.g. pay-schedule creation) and a possibly-stale/partial cached
+	 * snapshot would silently produce wrong amounts.
+	 */
```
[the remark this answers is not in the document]
teosarca: Applied in https://github.com/metasfresh/metasfresh/commit/41cfe87cd72. Reworded to "Loads ... fresh from the DB using the current (thread-inherited) transaction/connection, bypassing the process-wide cache" — dropped the misleading "committed DB state" phrasing, since a read in the thread-inherited trx can also see the caller's own uncommitted writes.

backend/de.metas.business/src/main/java/de/metas/payment/paymentterm/PaymentTermService.java
```diff
@@ -52,6 +52,17 @@ public PaymentTerm getById(@NonNull final PaymentTermId paymentTermId)
 		return paymentTermRepository.getById(paymentTermId);
 	}
 
+	/**
+	 * Loads the payment term together with its breaks fresh from the committed DB state, in the current
+	 * (thread-inherited) transaction, bypassing the process-wide cache. Use this where a complete,
+	 * consistent break set is required (e.g. pay-schedule creation).
+	 */
```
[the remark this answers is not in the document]
teosarca: Applied in https://github.com/metasfresh/metasfresh/commit/41cfe87cd72. Same rewording as on IPaymentTermRepository.getByIdInTrx — now "fresh from the DB using the current (thread-inherited) transaction/connection, bypassing the process-wide cache", no longer "committed DB state".

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/core/service/OrderPayScheduleCreateCommand.java
```diff
@@ -30,10 +30,18 @@ private void execute0()
 			return; // Nothing to schedule
 		}
 
+		final PaymentTerm paymentTerm = context.getPaymentTerm();
+		if (!paymentTerm.isValid())
+		{
+			throw new AdempiereException("Cannot spread payment schedules for payment term "
```
teosarca: when throwning an exception, ps always ask yourself who is the target audience.
* user? then it always shall be an AD_Message (translatable!) 
* developer/internal error? then it's fine as a hardcoded english message

capture learning + review

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/core/service/OrderPayScheduleService.java
```diff
@@ -158,7 +157,11 @@ private OrderSchedulingContext extractContext(final @NotNull org.compiere.model.
 		}
 
 		final PaymentTermId paymentTermId = orderBL.getPaymentTermId(orderRecord);
-		final PaymentTerm paymentTerm = paymentTermService.getById(paymentTermId);
+		// Load the payment term fresh from the DB using the current (completion) transaction, NOT from
+		// the process-wide cache: that shared snapshot can be stale/partial under concurrency (a reader
+		// populating it between two break INSERTs caches only some breaks), which would make spreadByBreaks
+		// give the sole/last break the full amount.
+		final PaymentTerm paymentTerm = paymentTermService.getByIdInTrx(paymentTermId);
```
teosarca: i really don't get the real life case here. 
usually payment terms (master data) are created once and the used.
there is no such case on prod system that we keep creating/changing payment terms while also creating purchase orders.

pls eli5 the case.

more, always pls avoid fixing testing cases by changing/adapting production code.
capture learning

backend/de.metas.business/src/main/java/de/metas/payment/paymentterm/repository/impl/PaymentTermRepository.java
```diff
@@ -84,6 +84,18 @@ public Optional<PaymentTerm> getByIdIfExists(@NonNull final PaymentTermId paymen
 		return getIndexedPaymentTerms().getById(paymentTermId);
 	}
 
+	@Override
+	public @NonNull PaymentTerm getByIdInTrx(@NonNull final PaymentTermId paymentTermId)
```
teosarca: see my prev comments, not sure this might be needed.
anyways, instead of having the need to write 1/2 page of AI slop crap, always consider giving intention revealing names to methods, e.g. getByIdNoCache.
Not saying that having getByIdNoCache is a preferable way, i.e. mixing technical concepts inside method names, but is WAY BETTER than hidding your workaround under the carpet.

capture learning

backend/de.metas.business/src/test/java/de/metas/order/paymentschedule/core/service/OrderPayScheduleCreateCommandTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Verifies that {@link OrderPayScheduleCreateCommand} fails loudly when given a
+ * complex-but-invalid payment term (breaks sum ≠ 100%), rather than silently producing
+ * a wrong {@code DueAmt}.
+ *
+ * <p>The test class lives in the same package as {@link OrderPayScheduleCreateCommand}
+ * (package-private) so it can construct it via its {@code @Builder}.
+ */
+class OrderPayScheduleCreateCommandTest
```
teosarca, line 57: why cucumber is not enough?

backend/de.metas.business/src/test/java/de/metas/payment/paymentterm/repository/impl/PaymentTermStaleCacheReproTest.java
```diff
@@ -0,0 +1,187 @@
...
+ * <ul>
+ *     <li>the cached {@code getById} keeps returning the stale {@code [LC]}-only term (LC DueAmt = 10000),</li>
+ *     <li>the fix's {@code getByIdInTrx} (a fresh in-trx load that {@code OrderPayScheduleService.extractContext}
+ *         now uses) reads the DB in the current transaction and returns both breaks (LC DueAmt = 3000).</li>
+ * </ul>
+ * Asserting LC DueAmt = 3000 on the {@code getByIdInTrx} path is the RED→GREEN gate: before the fix
+ * {@code extractContext} used {@code getById} and would see 10000; after the fix it uses {@code getByIdInTrx}
+ * and sees 3000.
+ */
+public class PaymentTermStaleCacheReproTest
```
teosarca: why cucumber is not enough!?

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/core/service/OrderPayScheduleCreateCommand.java
teosarca: Done — the guard now throws a translatable AD_Message. `assertValid()` throws `new AdempiereException(MSG_PAYMENT_TERM_INVALID, value, name, reason)` with `MSG_PAYMENT_TERM_INVALID = AdMessageKey.of("PaymentTerm_Invalid")`; the AD_Message (German base + en_US/de_DE/de_CH, MsgType=E, ErrorCode=PAYMENT_TERM_INVALID) is created by migration `5806690_sys_AD_Message_PaymentTerm_Invalid.sql`. https://github.com/metasfresh/metasfresh/commit/deb4bf24474

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/core/service/OrderPayScheduleService.java
teosarca: You're right. I reverted the cache bypass entirely — `extractContext` reads `paymentTermService.getById(...)` again (cache-backed, as before). The only remaining production change is the fail-loud `assertValid()` (in `spreadByBreaks` + `execute0`): it refuses to spread a complex-but-invalid payment term (breaks ≠ 100%) instead of silently writing a wrong DueAmt — a production-correctness invariant, independent of any test. Investigation confirmed the flake is not reproducible single-threaded locally (9 runs, 106 term reads all complete/valid); it is a CI-concurrency stale-cache read, so any residual de-flake will be test-side, not production. https://github.com/metasfresh/metasfresh/commit/deb4bf24474

backend/de.metas.business/src/main/java/de/metas/payment/paymentterm/repository/impl/PaymentTermRepository.java
teosarca: Agreed — `getByIdInTrx` is removed in full (gone from the repository, interface, and service), and the Javadoc with it. The cache bypass it supported is reverted. https://github.com/metasfresh/metasfresh/commit/deb4bf24474

backend/de.metas.business/src/test/java/de/metas/order/paymentschedule/core/service/OrderPayScheduleCreateCommandTest.java
teosarca: Every cucumber split-payment scenario uses a valid payment term (IsValid=Y). This unit test is the only deterministic coverage of the fail-loud invariant on a complex-but-invalid term (single 30% break, breaks ≠ 100%) — the exact condition `assertValid()` guards. Covering it in cucumber would require persisting an intentionally-broken payment term (not a real business scenario); the unit test does it with zero infra in <1s.

backend/de.metas.business/src/test/java/de/metas/payment/paymentterm/repository/impl/PaymentTermStaleCacheReproTest.java
teosarca: Moot now — this test was deleted in https://github.com/metasfresh/metasfresh/commit/deb4bf24474 . It only existed to exercise the stale-cache path the `getByIdInTrx` bypass addressed; with the bypass reverted, both the test and its reason are gone.

PR #24472, opened 2026-06-08
Qty reservation — reserve qtyTU x packing-item capacity (fix planned-supply Qty=0)

adi-stefan:
## What

Sales-order quantity reservations created against a **Planned-Supply** Material-Cockpit row were minted with `Qty=0` (CU) even though `QtyTU` was correct, because the reserved CU was derived from the cockpit row's `qtyStock` — which is `0`/`NULL` for planned-supply rows.

The reserved CU is now derived from the packing-item capacity instead:

- `Qty (CU) = qtyToReserveTU × capacity-per-TU`
- `capacity-per-TU = C_OrderLine.QtyItemCapacity` when `> 0`, else the order line's `M_HU_LUTU_Configuration` CU-per-TU.
- **On-Hand** reservations are hard-capped at the row's available `qtyStock`; **Planned-Supply** reservations are not capped.
- When no positive capacity can be determined, a user-validation error is raised instead of silently creating a `Qty=0` reservation.

Downstream, a completed shipment then tracks `QtyDelivered` against the (non-zero) reservation and flips `Processed` when fully delivered.

## How

- `MakeQtyReservationCommand` — capacity-based CU computation + On-Hand cap; takes a builder-injected optional `capacityPerTUFallback` and `IProductBL`.
- `MaterialCockpitV2RowVO` — removed the now-unused `qtyStock`-based CU derivation.
- `IHUOrderBL#getCapacityPerTUInStockUOMFallback` (`de.metas.handlingunits.base`) — shared LUTU-config capacity fallback, used by both the WebUI reservation process and the cucumber step definition (the module boundary prevents resolving it inside `de.metas.swat.base`).

Unit coverage: `MakeQtyReservationCommandTest` (planned-supply `N×capacity`; On-Hand capped at stock; On-Hand within stock; whole-TU reservation exceeding ordered CU; no-capacity → user-validation error).


adi-stefan:
Thanks — addressed the review (HEAD `4ecaaf2`):

1. **Migration removed — intentional.** The `5806680` data-cleanup migration was pulled from this PR by design; it's handled separately and is not part of this change. PR description updated to match.
2. **UOM assumption documented.** Wrapping `QtyItemCapacity` with the stock UOM in `computeCapacityPerTU` is the established convention (`IHUPIItemProductBL.getQtyCUsPerTUInStockUOM` does the same), so no conversion is needed — added a clarifying comment.
3. **Whole-TU reservation documented + tested.** Reserving `qtyToReserveTU × capacity` with no `QtyOrdered` cap is intentional (the TU count is already bounded upstream); added `plannedSupply_reservesWholeTUs_evenWhenExceedingQtyOrdered` to document it.
4. **`productBL` is now builder-injected** (was a service-locator field); both callers pass it and the test injects a mock.
5. **User-validation error asserted** in `noCapacity_throwsUserError` via `isUserValidationError()`.

Unit tests green (6).

adi-stefan:
@claude review

PR #24478, opened 2026-06-08
Fix duplicate M_ShipmentSchedule_QtyPicked rows when reversing an aggregate-HU shipment

adi-stefan:
## Problem

Reversing a shipment whose allocation is an **aggregate VHU** (one `M_HU` representing N transport units) leaves the shipment schedule with **N identical, not-yet-shipped `M_ShipmentSchedule_QtyPicked` rows** on the same `(VHU, TU, LU, QtyLU, QtyTU, QtyPicked)` tuple.

On reverse, the picked qty is re-created one transport unit at a time via the picking-job reopen:
`M_InOut` `AFTER_REVERSECORRECT` → `reopenPickingJobs` → `PickingJobReopenCommand`. When the shipment is then **recreated**, those rows all take one `M_InOutLine_ID` and collide on the partial unique index `M_ShipmentSchedule_QtyPicked_UI` — so the shipment can no longer be recreated after a void.

## Fix

- `PickingJobReopenCommand` now routes its per-TU re-creation through `tryMergeQtyPickedIntoExistingForVHU` (exposed via `PickingJobShipmentScheduleService`), consolidating the replay into the single existing un-shipped row for the VHU and restoring the pre-reversal one-row shape.
- The same merge is kept in `ShipmentScheduleHUTrxListener` for paths that go through the HU-trx listener.
- The merge's lookup query now runs in the **thread-inherited transaction** so it can see the sibling rows created earlier in the same reopen/replay (it previously ran out-of-transaction and could never find them).
- The merge is **self-gated**: a no-op for job-schedule-bound, catch-weight, negative, anonymous-on-the-fly or non-virtual picks — genuine per-step mobile picks are unaffected.

## Tests

- Unit tests for the merge predicate, the DAO lookup query, and the happy-path merge.
- A `reverseShipment` op in `de.metas.frontend-testing` plus a mobile-webui Playwright spec (`recreate_shipment_after_void.spec.js`) that picks an aggregate HU, generates a completed shipment, reverses it, and asserts the reverse leaves exactly **one** un-shipped `QtyPicked` row per unique-index tuple.

Verified end-to-end on a local stack: without the fix the reverse leaves 6 colliding rows (recreate fails); with the fix it leaves 1 (recreate succeeds).

### DIFF
--- a/ai-work/29561/pending-questions.md
+++ b/ai-work/29561/pending-questions.md
@@ -1,0 +1,4 @@
+
+## Post-integration cleanup (after PR 24478 is integrated)
+- Close PR https://github.com/metasfresh/metasfresh/pull/23792 (superseded by 24478) with a WHY comment linking 24478, and delete its branch `soft_panda_release_FixDuplicateQtyPickedOnReversal`. (User decision 2026-06-08: 23792 not needed once the fix lands in 24478.) PR closing/branch deletion is the human/Tobias's call — surface when 24478 integrates.
+- Route the 2 reviewer rule-gap candidates through metasfresh-capturing-learnings (ITrxManager class-field convention; trx-dependency-at-call-site doc) — with user OK.
--- a/ai-work/29561/pending-questions.md
+++ b/ai-work/29561/pending-questions.md
@@ -1,4 +1,0 @@
-
-## Post-integration cleanup (after PR 24478 is integrated)
-- Close PR https://github.com/metasfresh/metasfresh/pull/23792 (superseded by 24478) with a WHY comment linking 24478, and delete its branch `soft_panda_release_FixDuplicateQtyPickedOnReversal`. (User decision 2026-06-08: 23792 not needed once the fix lands in 24478.) PR closing/branch deletion is the human/Tobias's call — surface when 24478 integrates.
-- Route the 2 reviewer rule-gap candidates through metasfresh-capturing-learnings (ITrxManager class-field convention; trx-dependency-at-call-site doc) — with user OK.
--- a/backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/FrontendTestingRestController.java
+++ b/backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/FrontendTestingRestController.java
@@ -11,6 +11,9 @@ import de.metas.frontend_testing.masterdata.CreateMasterdataCommand;
 import de.metas.frontend_testing.masterdata.CreateMasterdataCommandSupportingServices;
 import de.metas.frontend_testing.masterdata.JsonCreateMasterdataRequest;
 import de.metas.frontend_testing.masterdata.JsonCreateMasterdataResponse;
+import de.metas.frontend_testing.masterdata.shipment.JsonReverseShipmentRequest;
+import de.metas.frontend_testing.masterdata.shipment.JsonReverseShipmentResponse;
+import de.metas.frontend_testing.masterdata.shipment.ReverseShipmentCommand;
 import de.metas.frontend_testing.masterdata.sysconfig.SysconfigCommand;
 import de.metas.handlingunits.qrcodes.service.HUQRCodesService;
 import de.metas.i18n.TranslatableStrings;
@@ -137,6 +140,15 @@ public class FrontendTestingRestController
 	{
 		return callInContext(() -> GetHUQRCodeCommand.builder()
 				.huQRCodesService(huQRCodesService)
+				.request(request)
+				.build()
+				.execute());
+	}
+
+	@PostMapping("reverseShipment")
+	public JsonReverseShipmentResponse reverseShipment(@RequestBody @NonNull final JsonReverseShipmentRequest request)
+	{
+		return callInContext(() -> ReverseShipmentCommand.builder()
 				.request(request)
 				.build()
 				.execute());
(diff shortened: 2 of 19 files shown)


adi-stefan:
@claude review

adi-stefan:
Thanks — addressed the findings (commit https://github.com/metasfresh/metasfresh/commit/fa1ffcd3c06693e221a75e09238c02dabc9e6753):

1. **Integration test structurally limited** — acknowledged, intentional. The in-memory POJO engine can't replay the per-TU HU-snapshot, so the real regression coverage is the Playwright spec; the integration test documents the invariant + dumps rows. No change.
2. **Missing non-virtual guard test** — ✅ applied. Added `rejects_when_hu_is_not_virtual` (gives the fixture HU a non-virtual PI version so `isVirtual` returns false instead of asserting on a 0 PI-version, then asserts `tryMerge` returns false).
3. **Possible NPE on QtyLU/QtyTU** — ✅ applied. Null-guarded `QtyLU`/`QtyTU`/`QtyPicked` in `computeMaxIdenticalUnshippedRows` via a `plain(BigDecimal)` helper.
4. **`throw` vs Playwright `expect()`** — ✅ applied. The two behavioural assertions now use `expect(...).toBe(...)` (consistent with the other mobile-webui specs that import `expect`).
5. **E2E doesn't recreate the shipment** — kept the row-count assertion as the precise invariant. It is exactly what the fix changes: without the fix the reverse leaves N identical un-shipped rows (here 6, → guaranteed `M_ShipmentSchedule_QtyPicked_UI` collision on recreate); with it, exactly 1. A literal recreate step is blocked by an orthogonal post-reverse schedule state in this harness (QtyToDeliver=0 + the HU still assigned to the reversed shipment ⇒ "no candidates"), unrelated to this fix — so it would fail for reasons other than the defect. Open to a follow-up that drives the full re-generate once that revalidation is sorted.
6. **`System.out.println` in `dumpQtyPicked`** — acknowledged; intentional for surefire console visibility (per the inline comment), diagnostic test helper only. No change.

teosarca:
### Catch-weight exclusion here may leave a whole-HU double-book on `new_dawn_uat` — bisected to this PR ✕ #24363

Flagging a deterministic `mobile test` failure on `new_dawn_uat` that git-bisects to this PR together with #24363. Not urgent, but it currently blocks every merge PR into `new_dawn_uat` (its `mobile test` job can't go green), so it needs an owner call.

#### Symptom
`e2e/mobile-webui/tests/spec/picking/overPickingPrompt_wholeHU.spec.js` fails: a **catch-weight whole-HU pick books two `M_ShipmentSchedule_QtyPicked` rows where one is expected**.

```
AssertPickingExpectationsCommand.assertQtyPickedList:244
"Expected qty picked records size to be <1> but was <2>"
```

#### This PR's role
The `tryMergeQtyPickedIntoExistingForVHU` you added (`HUShipmentScheduleBL.java:240`, wired into `ShipmentScheduleHUTrxListener.java:123`) deliberately falls through to create-new for catch weight:

```java
// HUShipmentScheduleBL.java:255-258
// ... fall through to create-new so each catch reading keeps its own row.
if (qtyToAdd.isUOMQtySet())
{
    return false;
}
```

That is correct for the aggregate-HU **reversal** case this PR fixed. But on a fresh whole-HU **pick**, the second row is created by #24363's per-VHU Country-of-Origin expansion (`HUShipmentScheduleBL.createCandidatesForQtyPicked`, `:1186`), and this catch-weight bail-out means it is never collapsed → both rows persist. The failing fixture (`createMasterdata_WholeHU`) is a catch-weight product, so `isUOMQtySet()` is true and the merge bails.

#### Why this is a backend defect on these two PRs, not the E2E
- **Control experiment.** The *same* 887-line spec is **green on `intensive_care_hotfix`** (build `.41473`, run https://github.com/metasfresh/metasfresh/actions/runs/30716067854 — 12/12 on this spec) and **red on `new_dawn_uat`** (run https://github.com/metasfresh/metasfresh/actions/runs/30714037116). Same test, different backend ⇒ the red is backend-determined.
- **Code presence.** `tryMergeQtyPickedIntoExistingForVHU` / `createCandidatesForQtyPicked` grep to **0 hits on `intensive_care_hotfix`**, present on `new_dawn_uat`.
- **Ancestry.** The introducing commits (`c50466a6bf3` here, `489b618710f`/`e0d9cd1b07c` on #24363) are ancestors of `new_dawn_uat` but **not** of `intensive_care_hotfix`, merged **2026-06-09/10**.

#### Behaviour vs. visibility (full disclosure)
The *exposing test* is newer than the defect: `overPickingPrompt_wholeHU.spec.js` reached `new_dawn_uat` via the whole-TU over-delivery work (#25374) propagated in #25367 (merge `bbe873afa6e`, 2026-08-01) — roughly 7 weeks *after* these June backend commits. A `.spec.js` cannot write DB rows, so the test only made an existing backend double-book visible; it did not introduce it. The first `new_dawn_uat` build to contain the spec (run https://github.com/metasfresh/metasfresh/actions/runs/30699288552) failed earlier at the `java` compile step, so `mobile test` only started reporting this once the build compiled.

#### Not yet done (honest gaps)
- **Not reproduced locally**, so I have not pinned which of the two save-sites emits the concrete second row (the create-new fall-through here vs. #24363's COO per-VHU expansion). Attribution to these two June PRs holds either way.
- **Open design question:** is the two-row outcome the *intended* catch-weight semantics (one row per catch reading) — in which case the test's expected-1 is what should change — **or** a genuine double-book to eliminate?

#### Ask
As the author of `tryMergeQtyPickedIntoExistingForVHU`: should the catch-weight bail-out be narrowed so it still collapses the fresh whole-HU-**pick** case while keeping per-reading rows on **reversal** — or is two rows intended here too? Happy to reproduce and pin the exact emitting line if that helps.

PR #24503, opened 2026-06-10
EDI DESADV: skip Qty-0 pack items (export-gate WhereClause + prevent SSCC-generated orphan)

adi-stefan:
## Problem

A Coop DESADV EDI message (clearing center Compudata) fails to send when the DESADV contains a **pack item of quantity 0**. Such a pack item is exported as a null/empty line, which the clearing center rejects.

Root analysis shows two distinct layers:

1. **The 0-qty pack item is created by SSCC-label generation.** `DesadvLineSSCC18Generator` builds a "drafted" pack item (no `M_InOutLine_ID`) with `MovementQty = qtyCUsPerLU`; when the LU breakdown is exhausted/misconfigured, `TotalQtyCUBreakdownCalculator.subtractOneLU()` returns the `LUQtys.NULL` singleton (qty 0), so a `MovementQty = 0` pack item is persisted.
2. **It then survives forever** — the reactivate/re-complete cleanup deletes pack items filtered by `M_InOutLine_ID` (a `NULL`-linked drafted item is invisible to it), and it can never be reclaimed (a 0 qty never matches a real shipment-line qty). The export-format WhereClause added previously filtered on the *parent line's* qty, not the item's own, so it slipped through.

## Fix (two layers)

**A — Export gate (`EXP_Format` WhereClause migration `5807110`).** `EDI_Exp_Desadv_Pack_Item` (540418) and `EDI_Exp_Desadv_Pack` (540419) now additionally require the pack item's **own** `MovementQty > 0`. A 0-qty pack item is never exported, even when its parent line still has qty > 0. Consistent with the existing export-gate design.

**B — Prevent the orphan at its source (`DesadvLineSSCC18Generator`).** A 0-qty / null LU is no longer turned into a pack item — the generator skips that label slot (with a `warn` log) instead of persisting a meaningless `MovementQty = 0` pack/pack-item. A 0-qty LU only arises from an exhausted/misconfigured calculator; it is not a legitimate label.

## Tests

- **JUnit** `DesadvBL_reactivateRecomplete_orphanPackItem_Test` — reproduces the complete → reactivate → re-complete cycle and asserts no orphan (`MovementQty = 0`, `M_InOutLine_ID = NULL`) pack item remains. RED before the generator fix, GREEN after.
- **Cucumber** `ediDesadvExportSkipsQtyZeroPackItem.feature` — exercises the live `EXP_Format` WhereClause and asserts a 0-qty pack item is excluded from the export selection.
- **Cucumber** `ediDesadvReactivateRecompleteNoOrphanPack.feature` — drives real SSCC-label generation (label count exceeding the shipped qty) to reproduce the orphan birth, then asserts none remains after reactivate/re-complete.

### DIFF
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/edi/EDI_DesadvLine_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/edi/EDI_DesadvLine_StepDef.java
@@ -22,22 +22,31 @@
 
 package de.metas.cucumber.stepdefs.edi;
 
+import de.metas.bpartner.BPartnerId;
+import de.metas.common.util.CoalesceUtil;
 import de.metas.cucumber.stepdefs.DataTableRow;
 import de.metas.cucumber.stepdefs.DataTableRows;
 import de.metas.cucumber.stepdefs.DataTableUtil;
 import de.metas.cucumber.stepdefs.M_Product_StepDefData;
 import de.metas.cucumber.stepdefs.StepDefDataIdentifier;
+import de.metas.edi.api.impl.DesadvBL;
+import de.metas.edi.api.impl.pack.EDIDesadvPackService;
+import de.metas.edi.sscc18.DesadvLineSSCC18Generator;
+import de.metas.edi.sscc18.PrintableDesadvLineSSCC18Labels;
 import de.metas.esb.edi.model.I_EDI_Desadv;
 import de.metas.esb.edi.model.I_EDI_DesadvLine;
+import de.metas.sscc18.impl.SSCC18CodeBL;
 import de.metas.uom.IUOMDAO;
 import de.metas.uom.UomId;
 import de.metas.uom.X12DE355;
 import de.metas.util.Services;
 import io.cucumber.datatable.DataTable;
 import io.cucumber.java.en.Then;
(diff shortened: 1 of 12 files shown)


adi-stefan:
@claude review

adi-stefan:
Addressed all `@claude review` findings in https://github.com/metasfresh/metasfresh/commit/78778d08bd8:

- **Test rename (Medium):** `DesadvBL_reactivateRecomplete_orphanPackItem_Test` → `DesadvLineSSCC18Generator_QtyZeroSkip_Test` — a CI failure now points directly at the fix under test (the generator's Qty-0 skip).
- **Single-DESADV guard (Medium):** `EDI_Desadv_Export_Format_StepDef` now asserts all DataTable rows reference the same `EDI_Desadv_ID` before using the first row for the pack lookup (no silent ignore).
- **`put`→`putOrReplace`:** `EDI_DesadvLine_StepDef.findAndRegisterDesadvLine` (consistent with `packItemIsFound`; safe for re-assertion).
- **Null-check tidy:** dropped the redundant `!= null` in the orphan-detection helper (`MovementQty` is a NOT-NULL column).
- **Comment tidies:** migration subquery rationale; trimmed the generator block comment.

Both `de.metas.edi` and `de.metas.cucumber` test-compile green; CI re-running.

PR #24513, opened 2026-06-10
Externalsystem scripted-export: OAuth2 password grant + multipart file upload + C_Doc_Outbound_Config external-system link

metas-ts:
## What

Generic **OAuth2 (resource-owner password grant)** support + **multipart file-upload** dispatch for scripted-export endpoints, plus a link column from document-outbound configs to external systems.

### ExternalSystem_Endpoint (config + UI)
- New `AuthType` list value **OAuth2** with new columns **OAuthTokenUrl** (mandatory when OAuth2) and **OAuthScope** (optional, per OAuth2 spec); reuses the existing generic `ClientId` / `LoginUsername` / `Password` columns (display + mandatory logic extended accordingly). Migration `5807080`.
- New **IsFileUpload** flag: dispatch the converted payload as `multipart/form-data` (document JSON + binary file parts) instead of a plain body.
- `SftpFilenamePattern` element description/help fixed (placeholder semantics documented, DE+EN).

### Camel scriptedadapter
- **OAuth2TokenManager**: password-grant token fetch (`grant_type=password`, optional scope), Guava-cached, invalidate-on-401 + single retry; mirrors the legacy `OAuthTokenManager` (untouched).
- **OAuth2 auth branch** in both dispatch routes (single + array-fan-out) setting `Authorization: Bearer`.
- **FileUploadProcessor**: builds the multipart body (`Document` JSON part + `File[]` binary part + `X-File-ModifiedDate` header) when the endpoint has `isFileUpload`; plain-body behavior unchanged otherwise.

### C_Doc_Outbound_Config
- New nullable TableDir column **ExternalSystem_Config_ID** (migration `5807200`, reuses shared element; models regenerated): links an outbound config to an external system, to be consumed by archive-event listeners in extension modules.

## Testing
- `JsonExternalSystemEndpointTest` 3/3 (DTO round-trip incl. OAuth2 + file-upload fields)
- `OAuth2TokenManagerTest` 5/5 (form contract, scope omission, caching, invalidate, error path) — JDK HttpServer stub
- `OAuth2RouteTest` 2/2 (Bearer header, 401 invalidate-retry), `FileUploadProcessorTest` 3/3 (part names/content, pass-through, malformed input), existing route tests 14/14 — all local, JDK 17
- Migrations applied + verified against a local stack DB; window validated by the window-designer check (zero violations after fixes)


backend/de.metas.externalsystem/src/main/java-gen/de/metas/externalsystem/model/I_ExternalSystem_Endpoint.java
```diff
@@ -289,6 +289,73 @@ public interface I_ExternalSystem_Endpoint
...
+	String COLUMNNAME_OAuthScope = "OAuthScope";
+
+	/**
+	 * Set File Upload.
+	 *
+	 * <br>Type: YesNo
+	 * <br>Mandatory: true
+	 * <br>Virtual Column: false
+	 */
+	void setIsFileUpload (boolean IsFileUpload);
```
metas-ts, line 346: assuming that the javadoc reflects the field-description which the user will see, i think this field lacks description

metas-ts: Good catch on the missing javadoc line — but the user-facing field description itself is present and correct; only the generated javadoc was stale.

The `IsFileUpload` field has full Name + Description + Help (DE + EN) on its `AD_Element` / `AD_Column` / `AD_Field` (set by migration `5807080_sys_ExternalSystem_Endpoint_OAuth2_columns.sql`), so the WebUI tooltip the user sees is fine — e.g. EN: *"Send the request body as multipart/form-data (document + file)."*

Why the javadoc lacked it: the model generator can't currently run cleanly for `ExternalSystem_Endpoint` (its encrypted `SshPrivateKey` column makes the generator emit non-compiling accessors), so this model is hand-maintained. The `IsFileUpload` accessors were hand-added with a minimal javadoc that omitted the description line (the `OAuthTokenUrl`/`OAuthScope` accessors I added do include it).

Synced the javadoc to match the AD description in https://github.com/metasfresh/metasfresh/pull/24513/commits/32607f44621 .

PR #24514, opened 2026-06-10
Transport Order — fix silent skip and currency sync

teosarca:
## Summary

- **Bug 1**: `PurchaseOrderToShipperTransportationService.addPurchaseOrderLines()` — replace silent `continue` (when `requiredLUCount = 0`) with a user-visible `AdempiereException` listing the skipped PO lines and explaining that no LU packing configuration was found. When all lines are skipped, throws a validation error; when some are skipped, logs a warning.
- **Bug 2a**: `OrderLinePriceCalculator.applyPricingResult()` — use order header `C_Currency_ID` instead of `pricingResult.getCurrencyId()` when writing to `C_OrderLine.C_Currency_ID`. The order header is the canonical currency; the pricing result must not override it independently.
- **Bug 2b**: `MOrder.beforeSave()` / `IOrderBL.syncCurrencyFromPriceList()` — always sync `C_Order.C_Currency_ID` from the price list when one is set, regardless of the current header value. Removes the `<= 0` guard that prevented correction of a wrong session-default currency in programmatic/import creation paths.

## Test plan

- [ ] `PurchaseOrderToShipperTransportationServiceTest#addPurchaseOrderLines_allLinesSkipped_throwsUserError` passes
- [ ] `PurchaseOrderToShipperTransportationServiceTest#addPurchaseOrderLines_someLinesSkipped_warnsAndAddsRest` passes
- [ ] `OrderLinePriceCurrencySyncTest#applyPricingResult_usesOrderHeaderCurrency` passes
- [ ] `MOrderCurrencySyncTest#beforeSave_alwaysSyncsCurrencyFromPriceList_evenWhenAlreadySet` passes
- [ ] All tests TDD-verified (RED before production change, GREEN after)
- [ ] `de.metas.business` module compiles clean


backend/de.metas.business/src/main/java-legacy/org/compiere/model/MOrder.java
```diff
@@ -956,24 +954,8 @@ protected boolean beforeSave(final boolean newRecord)
...
-			{
-				setC_Currency_ID(currencyId);
-			}
-			else
-			{
-				setC_Currency_ID(Env.getContextAsInt(getCtx(), "#C_Currency_ID"));
-			}
-		}
+		// Default Currency — always sync from price list when one is set
+		orderBL.syncCurrencyFromPriceList(InterfaceWrapperHelper.create(this, I_C_Order.class));
```
teosarca: IMHO this is quite ineficient. we shall
* set currency if not set yet (like it was before)
* or if the price list field changed

i would keep the logic here (not move it to services) because
* in tihs case i would keep changes to minimum
* things like "Env.getContextAsInt(InterfaceWrapperHelper.getCtx(order), "#C_Currency_ID")" CANNOT GO TO SERVICES

concern: how the lines are updated if we change the price list (so the currency) after we create them?

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderLinePriceCalculator.java
```diff
@@ -166,7 +169,7 @@ public void updateOrderLine()
 
 		//
 		// C_Currency_ID, M_PriceList_Version_ID
-		orderLine.setC_Currency_ID(CurrencyId.toRepoId(pricingResult.getCurrencyId()));
+		orderLine.setC_Currency_ID(order.getC_Currency_ID());
```
teosarca: i disagree! simply chanigng rhe result currency is not right! we need to make sure the input is correct
drop it!

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
```diff
@@ -245,6 +245,26 @@ public void setM_PricingSystem_ID(final I_C_Order order, final boolean overrideP
...
+		if (priceList != null)
+		{
+			order.setC_Currency_ID(priceList.getC_Currency_ID());
+		}
+		else if (order.getC_Currency_ID() <= 0)
+		{
+			// No price list and no currency set yet: fall back to the session's client default.
+			// Env access is intentional here — this is the same fallback MOrder.beforeSave() used before the extraction,
+			// and there is no domain-level alternative for an unset currency with no price list.
+			order.setC_Currency_ID(Env.getContextAsInt(InterfaceWrapperHelper.getCtx(order), "#C_Currency_ID"));
```
teosarca: pls avoid calling Env.getContext* in services! capture learning!

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
```diff
@@ -192,23 +199,52 @@ private void addPurchaseOrderLines(final @NonNull I_M_ShipperTransportation ship
...
+			}
+			else
+			{
+				Loggables.addLog("Skipped {} PO line(s) with no LU packing configuration: {}",
+						skippedLines.size(), formatSkippedLines(skippedLines));
+			}
 		}
 	}
 
+	private static String formatSkippedLines(@NonNull final List<I_C_OrderLine> orderLines)
```
teosarca: no user understands C_OrderLine_ID, M_Product_ID!

just simply get the C_OrderLine.Line (aka line numbers).
u dont need a method for that

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
```diff
@@ -192,23 +199,52 @@ private void addPurchaseOrderLines(final @NonNull I_M_ShipperTransportation ship
...
 			final int qtyLUs = ol.getQtyLU().intValueExact();
 			addPurchaseOrderLineToShipperTransportationId(baseRequest, order, ol, qtyLUs);
+			addedCount++;
+		}
+
+		if (!skippedLines.isEmpty())
+		{
+			if (addedCount == 0)
+			{
+				// TODO: replace literal with AdMessageKey once a migration script adds the AD_Message entry
```
teosarca: pls always add a trl message!
importat review rule: never ever accest TODO, FIXME lines introduced in the same PR without explicit human approval. And first, before asking him/her pls make sure if is easily fixable and do it!

backend/de.metas.business/src/test/java/de/metas/order/impl/OrderLinePriceCurrencySyncTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * Verifies that {@link OrderLinePriceCalculator#updateOrderLine()} writes the order <em>header</em>
+ * currency to the order line, not the pricing-result currency.
+ *
+ * <p>Before the fix, {@code setC_Currency_ID} was called with {@code pricingResult.getCurrencyId()}
+ * (EUR in this test), overwriting the header currency (USD).  After the fix the header value must win.
+ */
+@ExtendWith(AdempiereTestWatcher.class)
+class OrderLinePriceCurrencySyncTest
```
teosarca: crap

backend/de.metas.business/src/test/java/de/metas/order/MOrderCurrencySyncTest.java
```diff
@@ -0,0 +1,89 @@
...
+ * Tests {@link IOrderBL#syncCurrencyFromPriceList(I_C_Order)}, which is the method
+ * extracted from {@code MOrder.beforeSave()} to contain the currency-sync logic.
+ *
+ * <p>The key contract: when an order has a price list, the currency is always taken
+ * from the price list — even if the order already has a non-zero currency set.
+ * This prevents the pre-fix bug where an order pre-set with USD would not be corrected
+ * to EUR when a EUR price list was attached.
+ */
+@ExtendWith(AdempiereTestWatcher.class)
+class MOrderCurrencySyncTest
```
teosarca: crap

teosarca: this kind of things shall be tested exclusivelly in cucumber

backend/de.metas.business/src/test/java/de/metas/shipping/PurchaseOrderToShipperTransportationServiceTest.java
```diff
@@ -463,6 +466,79 @@ public void syncShippingPackagesFromOrder_removesAllPackagesWhenAllLinesDeleted(
...
+				Money.of(10, chf)
+		);
+
+		assertThatThrownBy(() -> serviceWithZeroLU.addPurchaseOrdersToShipperTransportation(
+				transportationId,
+				Collections.singletonList(orderId)))
+				.isInstanceOf(AdempiereException.class)
+				.satisfies(ex -> {
+					assertThat(ex.getMessage()).isNotEmpty();
+					assertThat(((AdempiereException) ex).isUserValidationError())
```
teosarca: pls always validate the AdempiereException.getErrorCode to make sure we get the right exception!
message is not so important because is translatable

backend/de.metas.business/src/main/java-legacy/org/compiere/model/MOrder.java
teosarca: **Resolution (dev):** Kept the logic inline in `MOrder.beforeSave()` (reverted the service extraction). Currency is now set when it is not set yet **or** when the price list just changed: `if (getC_Currency_ID() <= 0 || is_ValueChanged(COLUMNNAME_M_PriceList_ID))`.

On your concern about lines created earlier: when the price list (and thus the header currency) changes on an existing order, `afterSave()` propagates `C_Currency_ID` to the lines via the existing "Sync Lines" block (`is_ValueChanged(COLUMNNAME_C_Currency_ID)` → `line.setC_Currency_ID(getC_Currency_ID())`), so the lines stay consistent.

https://github.com/metasfresh/metasfresh/commit/34159d19096dee12832880335dff4550e5fa64c2

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderLinePriceCalculator.java
teosarca: **Resolution (dev):** Reverted — `OrderLinePriceCalculator` is back to base (`setC_Currency_ID(CurrencyId.toRepoId(pricingResult.getCurrencyId()))`, and the extra order-header load is gone). The input is fixed at the source instead: `MOrder.beforeSave()` keeps the header currency in sync with the price list, so the pricing result already carries the right currency.

https://github.com/metasfresh/metasfresh/commit/34159d19096dee12832880335dff4550e5fa64c2

backend/de.metas.business/src/main/java/de/metas/order/impl/OrderBL.java
teosarca: **Resolution (dev):** Reverted the service extraction; the `Env.getContextAsInt(getCtx(), "#C_Currency_ID")` fallback is back in `MOrder.beforeSave()` (a legacy model lifecycle method — the allowed near-user boundary), not in a service. This is already a documented review rule (`REVIEW.md` → "No low-level `Env.get*` context accessors in services/repos" / `docs/coding-rules/service-injection.md` §7); the code now complies.

https://github.com/metasfresh/metasfresh/commit/34159d19096dee12832880335dff4550e5fa64c2

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
teosarca: **Resolution (dev):** The skipped-lines message now uses the order line numbers (`ol.getLine()`), built inline; the separate `formatSkippedLines` helper is removed.

https://github.com/metasfresh/metasfresh/commit/34159d19096dee12832880335dff4550e5fa64c2

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
teosarca: **Resolution (dev):** Added a real `AdMessageKey` (`NoLUPackingConfigForOrderLines`) backed by a migration script (`AD_Message` with German base text + en_US translation, `ErrorCode = NO_LU_PACKING_CONFIG`). The hardcoded string and the TODO are gone.

https://github.com/metasfresh/metasfresh/commit/34159d19096dee12832880335dff4550e5fa64c2

backend/de.metas.business/src/test/java/de/metas/order/impl/OrderLinePriceCurrencySyncTest.java
teosarca: **Resolution (dev):** Deleted — the `OrderLinePriceCalculator` change this test covered is reverted, so the test is moot.

https://github.com/metasfresh/metasfresh/commit/34159d19096dee12832880335dff4550e5fa64c2

backend/de.metas.business/src/test/java/de/metas/order/MOrderCurrencySyncTest.java
teosarca: **Resolution (dev):** JUnit test deleted. The currency-sync behaviour is now covered by a cucumber scenario instead: `backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/purchaseOrder/orderCurrencyFromPriceList.feature` — a purchase order's header currency follows its price list's currency.

https://github.com/metasfresh/metasfresh/commit/34159d19096dee12832880335dff4550e5fa64c2

backend/de.metas.business/src/test/java/de/metas/shipping/PurchaseOrderToShipperTransportationServiceTest.java
teosarca: **Resolution (dev):** The test now asserts `AdempiereException.getErrorCode()` equals the AD_Message key (`MSG_NoLUPackingConfigForOrderLines.toAD_Message()`), in addition to `isUserValidationError()` — instead of only checking the message text.

https://github.com/metasfresh/metasfresh/commit/34159d19096dee12832880335dff4550e5fa64c2

teosarca:
All review threads have been addressed in https://github.com/metasfresh/metasfresh/commit/34159d19096dee12832880335dff4550e5fa64c2 (each thread has a **Resolution (dev):** reply). If you agree with a reply or consider a discussion resolved, please mark the thread as resolved on GitHub — I can't mark review threads resolved via the API, that action is reserved for humans.

PR #24528, opened 2026-06-11
nShift service levels: fix non-ID-server IDs for AD_Org field + UI element on Service Level Konfiguration tab

mrie04:
The AD_Org_ID field (`AD_Field` 580485) and its UI element (`AD_UI_Element` 580486) in `5804800` were not allocated from the ID server and collide with pre-existing rows in downstream instances:

```
ERROR: duplicate key value violates unique constraint "ad_ui_element_key"
Key (ad_ui_element_id)=(580486) already exists.
```

These two IDs sit far below the freshly-allocated siblings in the same script (`AD_Field` 780489–780492, `AD_UI_Element` 651846–651849) and are consecutive across two different tables — impossible for genuine per-table ID-server output — i.e. a copy-pasted AD_Org block carrying stale low IDs.

Since `5804800` is already rolled out, the IDs cannot be changed in place. Fix is split across three scripts:

- **`5804800`** (modified): disable the AD_Org field + UI element blocks (commented out, with reason).
- **`5804799`**: remove those rows, every statement scoped to the new tab (`AD_Tab_ID = 549282`) so differently-owned downstream rows that share IDs 580485/580486 are left untouched. Full `AD_Field` FK chain cleared in order.
- **`5804801`**: re-add the field + UI element with fresh ID-server IDs (`AD_Field` 780756 / `AD_UI_Element` 652051).

End state converges for fresh installs, instances where `5804800` already succeeded, and instances where its `580486` insert failed on the duplicate.


metas-ts (approved):

mrie04:

PR #24532, opened 2026-06-12
Mass Printing — search demand via picking-launcher eligibility + validate LU picking group

teosarca:
## What

Mobile Mass Printing (MobileUI Picking) now searches open demand the same way the picking launcher computes its candidates, instead of using the scanned LU's own storage warehouse. It also validates that the scanned LU belongs to the picker workplace's picking group.

## Why

The mass-printing scan resolved the demand-search warehouse from the scanned LU's locator. When stock is stored in one warehouse and brought to a packing workplace in another warehouse of the same picking group, the open shipment schedules live in the workplace warehouse, so the LU-warehouse-scoped search found nothing and packed 0 units.

## Changes

- `MassPrintingService.scan()` now builds a `PickingJobQuery` mirroring the picking launcher (picker, workplace warehouse, `scheduledForWorkplaceId` when `isConsiderOnlyJobScheduledToWorkplace`, `pickOnlyCustomerIds`) and resolves eligible schedules via the shared `PickingJobService.streamPackageable(...)` seam, intersected with the self-packed products on the LU. Removes the bespoke `PackageableQuery` and the now-redundant other-user-lock filter (inherited from the reused query).
- Adds a guard: if the picker is assigned to a workplace, the scanned LU's locator must belong to that workplace's picking group, else the scan is rejected with a typed message.
- New `AD_Message` for that rejection.
- Cucumber coverage: cross-warehouse pack (LU in another warehouse of the same picking group) + LU-not-in-group rejection.

### DIFF
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/warehouse/M_Warehouse_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/warehouse/M_Warehouse_StepDef.java
@@ -49,7 +49,9 @@ import org.adempiere.ad.dao.IQueryBL;
 import org.adempiere.exceptions.AdempiereException;
 import org.adempiere.model.InterfaceWrapperHelper;
 import org.adempiere.warehouse.WarehouseId;
+import org.adempiere.warehouse.api.CreateWarehousePickingGroupRequest;
 import org.adempiere.warehouse.api.IWarehouseBL;
+import org.adempiere.warehouse.groups.picking.WarehousePickingGroupId;
 import org.compiere.model.I_C_BPartner_Location;
 import org.compiere.model.I_M_Locator;
 import org.compiere.model.I_M_Warehouse_PickingGroup;
@@ -128,16 +130,20 @@ public class M_Warehouse_StepDef
 				.findFirst()
 				.orElseThrow(() -> new AdempiereException("At least one row with a Name is required"));
 
-		final I_M_Warehouse_PickingGroup pickingGroup = InterfaceWrapperHelper.newInstance(I_M_Warehouse_PickingGroup.class);
-		pickingGroup.setName(groupName);
-		saveRecord(pickingGroup);
+		// Create the group via the BL (which owns the DAO), with a per-run-unique Name so reusing the same
+		// group label across tests in the shared DB does not collide.
+		final WarehousePickingGroupId pickingGroupId = warehouseBL.createWarehousePickingGroup(
+				CreateWarehousePickingGroupRequest.builder()
+						.orgId(StepDefConstants.ORG_ID)
+						.name(groupName + "_" + System.currentTimeMillis())
+						.build());
 
 		rows.forEach(row -> {
 			final WarehouseId warehouseId = warehouseTable.getId(row.getAsIdentifier(COLUMNNAME_M_Warehouse_ID));
 
 			final org.compiere.model.I_M_Warehouse warehouseRecord = InterfaceWrapperHelper.load(
 					warehouseId.getRepoId(), org.compiere.model.I_M_Warehouse.class);
-			warehouseRecord.setM_Warehouse_PickingGroup_ID(pickingGroup.getM_Warehouse_PickingGroup_ID());
+			warehouseRecord.setM_Warehouse_PickingGroup_ID(pickingGroupId.getRepoId());
 			saveRecord(warehouseRecord);
 		});
 
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/warehouse/M_Warehouse_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/warehouse/M_Warehouse_StepDef.java
@@ -24,6 +24,7 @@ package de.metas.cucumber.stepdefs.warehouse;
 
 import de.metas.bpartner.BPartnerId;
 import de.metas.bpartner.BPartnerLocationId;
+import de.metas.cache.CacheMgt;
 import de.metas.common.util.CoalesceUtil;
 import de.metas.cucumber.stepdefs.C_BPartner_Location_StepDefData;
 import de.metas.cucumber.stepdefs.C_BPartner_StepDefData;
@@ -51,6 +52,7 @@ import org.adempiere.warehouse.WarehouseId;
 import org.adempiere.warehouse.api.IWarehouseBL;
 import org.compiere.model.I_C_BPartner_Location;
 import org.compiere.model.I_M_Locator;
+import org.compiere.model.I_M_Warehouse_PickingGroup;
 
 import static org.adempiere.model.InterfaceWrapperHelper.COLUMNNAME_IsActive;
 import static org.adempiere.model.InterfaceWrapperHelper.saveRecord;
@@ -89,6 +91,62 @@ public class M_Warehouse_StepDef
 
 					row.getAsIdentifier().put(warehouseTable, warehouseRecord);
 				});
+	}
+
+	/**
+	 * Creates one {@code M_Warehouse_PickingGroup} and assigns every listed warehouse to it
+	 * (sets {@code M_Warehouse.M_Warehouse_PickingGroup_ID}).
+	 *
+	 * <p>Warehouses sharing a picking group are treated as a single picking scope: a picker working at a
+	 * workplace in one warehouse of the group may pick demand and stock located in any warehouse of the
+	 * same group. This mirrors the real-life setup where a storage warehouse and a packing/workplace
+	 * warehouse belong to the same physical picking area.
+	 *
+	 * <p>Required columns:
+	 * <ul>
+	 *   <li>{@code Name} — the picking group's name (one group is created for the whole DataTable;
+	 *       the {@code Name} of the first row is used).</li>
+	 *   <li>{@code M_Warehouse_ID} — (identifier-ref) a previously created/loaded warehouse to assign to the
+	 *       group. One row per warehouse.</li>
+	 * </ul>
+	 *
+	 * <p>Gherkin usage:
+	 * <pre>
+	 * And the following warehouses share one M_Warehouse_PickingGroup:
+	 *   | Name        | M_Warehouse_ID |
+	 *   | pickingArea | warehouseStock |
+	 *   | pickingArea | warehouseWork  |
+	 * </pre>
+	 */
+	@And("the following warehouses share one M_Warehouse_PickingGroup:")
+	public void assign_warehouses_to_picking_group(@NonNull final DataTable dataTable)
+	{
+		final DataTableRows rows = DataTableRows.of(dataTable);
+
+		final String groupName = rows.stream()
+				.map(row -> row.getAsString(I_M_Warehouse_PickingGroup.COLUMNNAME_Name))
+				.findFirst()
+				.orElseThrow(() -> new AdempiereException("At least one row with a Name is required"));
+
+		final I_M_Warehouse_PickingGroup pickingGroup = InterfaceWrapperHelper.newInstance(I_M_Warehouse_PickingGroup.class);
+		pickingGroup.setName(groupName);
+		saveRecord(pickingGroup);
+
+		rows.forEach(row -> {
+			final WarehouseId warehouseId = warehouseTable.getId(row.getAsIdentifier(COLUMNNAME_M_Warehouse_ID));
+
+			final org.compiere.model.I_M_Warehouse warehouseRecord = InterfaceWrapperHelper.load(
+					warehouseId.getRepoId(), org.compiere.model.I_M_Warehouse.class);
+			warehouseRecord.setM_Warehouse_PickingGroup_ID(pickingGroup.getM_Warehouse_PickingGroup_ID());
+			saveRecord(warehouseRecord);
+		});
+
+		// The picking-group → warehouses index is cached (WarehouseDAO.allWarehousePickingGroups). The bulk
+		// save fires a CacheInvalidation event, but in the shared single-JVM executor that event may not be
+		// processed before the next step reads the index. Reset the relevant caches synchronously so the
+		// freshly-assigned picking group is visible immediately (same trap handled in C_Workplace_StepDef).
+		CacheMgt.get().reset(I_M_Warehouse.Table_Name);
+		CacheMgt.get().reset(I_M_Warehouse_PickingGroup.Table_Name);
 	}
 
 	/**
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/warehouse/M_Warehouse_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/warehouse/M_Warehouse_StepDef.java
@@ -22,9 +22,9 @@
 
 package de.metas.cucumber.stepdefs.warehouse;
 
+import com.google.common.collect.ImmutableSet;
 import de.metas.bpartner.BPartnerId;
 import de.metas.bpartner.BPartnerLocationId;
-import de.metas.cache.CacheMgt;
 import de.metas.common.util.CoalesceUtil;
 import de.metas.cucumber.stepdefs.C_BPartner_Location_StepDefData;
 import de.metas.cucumber.stepdefs.C_BPartner_StepDefData;
@@ -51,7 +51,6 @@ import org.adempiere.model.InterfaceWrapperHelper;
 import org.adempiere.warehouse.WarehouseId;
 import org.adempiere.warehouse.api.CreateWarehousePickingGroupRequest;
 import org.adempiere.warehouse.api.IWarehouseBL;
-import org.adempiere.warehouse.groups.picking.WarehousePickingGroupId;
 import org.compiere.model.I_C_BPartner_Location;
 import org.compiere.model.I_M_Locator;
 import org.compiere.model.I_M_Warehouse_PickingGroup;
@@ -130,29 +129,18 @@ public class M_Warehouse_StepDef
 				.findFirst()
 				.orElseThrow(() -> new AdempiereException("At least one row with a Name is required"));
 
-		// Create the group via the BL (which owns the DAO), with a per-run-unique Name so reusing the same
-		// group label across tests in the shared DB does not collide.
-		final WarehousePickingGroupId pickingGroupId = warehouseBL.createWarehousePickingGroup(
+		final ImmutableSet<WarehouseId> warehouseIds = rows.stream()
+				.map(row -> warehouseTable.getId(row.getAsIdentifier(COLUMNNAME_M_Warehouse_ID)))
+				.collect(ImmutableSet.toImmutableSet());
+
+		// Create the group AND assign the warehouses to it in one BL call (the BL/DAO owns the record load+save).
+		// A per-run-unique Name keeps reusing the same group label across tests in the shared DB from colliding.
+		warehouseBL.createWarehousePickingGroup(
 				CreateWarehousePickingGroupRequest.builder()
 						.orgId(StepDefConstants.ORG_ID)
 						.name(groupName + "_" + System.currentTimeMillis())
+						.warehouseIds(warehouseIds)
 						.build());
-
-		rows.forEach(row -> {
-			final WarehouseId warehouseId = warehouseTable.getId(row.getAsIdentifier(COLUMNNAME_M_Warehouse_ID));
-
-			final org.compiere.model.I_M_Warehouse warehouseRecord = InterfaceWrapperHelper.load(
-					warehouseId.getRepoId(), org.compiere.model.I_M_Warehouse.class);
-			warehouseRecord.setM_Warehouse_PickingGroup_ID(pickingGroupId.getRepoId());
-			saveRecord(warehouseRecord);
-		});
-
-		// The picking-group → warehouses index is cached (WarehouseDAO.allWarehousePickingGroups). The bulk
-		// save fires a CacheInvalidation event, but in the shared single-JVM executor that event may not be
-		// processed before the next step reads the index. Reset the relevant caches synchronously so the
-		// freshly-assigned picking group is visible immediately (same trap handled in C_Workplace_StepDef).
-		CacheMgt.get().reset(I_M_Warehouse.Table_Name);
-		CacheMgt.get().reset(I_M_Warehouse_PickingGroup.Table_Name);
 	}
 
 	/**
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/mobileui/picking/massPrinting_job_scheduled_to_workplace.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/mobileui/picking/massPrinting_job_scheduled_to_workplace.feature
@@ -166,7 +166,3 @@ Feature: Mass Printing - honour job-scheduled-to-workplace mode
     And after not more than 10s, validate shipment schedules:
       | M_ShipmentSchedule_ID.Identifier | OPT.QtyToDeliver | OPT.QtyDelivered | OPT.QtyPickList |
       | schedNotScheduled                | 4                | 0                | 0               |
-
-    # Leak-safety teardown.
-    And delete all C_Workplace_User_Assign records
-    And deactivate all C_Workplace records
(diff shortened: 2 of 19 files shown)


backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/warehouse/WarehouseCommand.java
```diff
@@ -97,6 +100,37 @@ private void createWarehouse()
...
+		final String pickingGroupName = StringUtils.trimBlankToNull(request.getPickingGroup());
+		if (pickingGroupName == null)
+		{
+			return;
+		}
+
+		final Identifier pickingGroupIdentifier = Identifier.ofString(pickingGroupName);
+		final WarehousePickingGroupId pickingGroupId = context.getOptionalId(pickingGroupIdentifier, WarehousePickingGroupId.class)
+				.orElseGet(() -> {
+					final I_M_Warehouse_PickingGroup pickingGroupRecord = InterfaceWrapperHelper.newInstance(I_M_Warehouse_PickingGroup.class);
```
teosarca: pls introduce a org.adempiere.warehouse.api.impl.WarehouseDAO.createWarehousePickingGroup(request) method and use it.

even in testing is it preferable to use DAOs/Repositories instead of direct InterfaceWrapperHelper, IQueryBL calls. capture this learning!

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/service/PickingJobService.java
```diff
@@ -325,6 +328,38 @@ public Stream<Packageable> streamPackageable(@NonNull final PickingJobQuery quer
...
+	public ShipmentScheduleAndJobScheduleIdSet resolveScheduleIdsForJobCreation(
+			@NonNull final PickingJobQuery query,
+			@NonNull final Set<ShipmentScheduleId> selectedShipmentScheduleIds)
+	{
+		if (selectedShipmentScheduleIds.isEmpty())
+		{
+			return ShipmentScheduleAndJobScheduleIdSet.EMPTY;
+		}
+
+		if (!query.isScheduledForWorkplaceOnly())
```
teosarca, line 336: what is the purpose of this checking?

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/service/PickingJobService.java
```diff
@@ -325,6 +328,38 @@ public Stream<Packageable> streamPackageable(@NonNull final PickingJobQuery quer
...
+	 * Resolves the {@link ShipmentScheduleAndJobScheduleIdSet} to drive a picking-job creation for the given
+	 * selected shipment-schedule ids, using the same job-schedule resolution as the picking launcher.
+	 * <p>
+	 * In job-scheduled-to-workplace mode (the query carries a {@code scheduledForWorkplaceId}), the returned set
+	 * carries the picking-job-schedule ids for the selected shipment schedules (looked up via the same
+	 * {@code toPickingJobScheduleQuery()} the launcher / {@link #streamPackageable} use), so the created job is
+	 * driven by the job-schedules — consistent with how regular picking creates jobs from candidates. In
+	 * warehouse mode (no {@code scheduledForWorkplaceId}) it falls back to a plain shipment-schedule-id set.
+	 */
+	public ShipmentScheduleAndJobScheduleIdSet resolveScheduleIdsForJobCreation(
```
teosarca: IMHO this method is so custom that it does not make sense here but in the caller.

ofc in this service we would some method that does "final PickingJobScheduleCollection jobSchedules = pickingJobScheduleService.list(query.toPickingJobScheduleQuery());"

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/massprinting/MassPrintingService.java
```diff
@@ -346,10 +307,17 @@ private ScheduleSelection selectSchedulesFifo(
 	@NonNull
 	private PickingJob createAndPickJob(
 			@NonNull final MassPrintingScanRequest request,
+			@NonNull final PickingJobQuery pickingJobQuery,
 			@NonNull final HuId luId,
 			@NonNull final ScheduleSelection selection)
 	{
-		final ShipmentScheduleAndJobScheduleIdSet scheduleIdSet = ShipmentScheduleAndJobScheduleIdSet.ofShipmentScheduleIds(selection.getSelectedScheduleIds());
+		// Resolve the schedule-id set exactly like regular picking does from a candidate: in
```
teosarca: AI splop. pls keep it short and concise.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/massprinting/MassPrintingService.java
```diff
@@ -346,10 +307,17 @@ private ScheduleSelection selectSchedulesFifo(
 	@NonNull
 	private PickingJob createAndPickJob(
 			@NonNull final MassPrintingScanRequest request,
+			@NonNull final PickingJobQuery pickingJobQuery,
 			@NonNull final HuId luId,
```
teosarca: rename luId to pickFromLUId

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/warehouse/WarehouseCommand.java
```diff
@@ -97,6 +100,37 @@ private void createWarehouse()
...
+					pickingGroupRecord.setAD_Org_ID(MasterdataContext.ORG_ID.getRepoId());
+					pickingGroupRecord.setName(pickingGroupName);
```
teosarca: don't use pickingGroupName as Name because u will collide with previous/future tests.
consider using pickingGroupIdentifier.toUniqueString() and make sure the M_Warehouse_PickingGroup.Name can accomodate at least 255 chars!

pls capture this experience about writting tests and making sure they are independent and not collide.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/mobileui/picking/massPrinting_job_scheduled_to_workplace.feature
```diff
@@ -0,0 +1,168 @@
...
+
+    # The picker scans the LU — mass printing must pack ONLY the 4 PCE of the scheduled demand, even though
+    # the LU holds 8 PCE and there is another open (but unscheduled) order for 4 PCE.
+    And mass printing scan:
+      | M_HU_ID | Login      |
+      | lu      | metasfresh |
+
+    Then validate mass printing result:
+      | M_HU_ID | UnitsPacked |
+      | lu      | 4           |
```
teosarca, line 153: pls also test the `lu`'s capacity after mass printing.
also pls do this test on all mass printing tests.

teosarca: **Resolution (dev):** Done — added an LU-remaining-qty assertion after the result step in all three mass-printing cucumber features (`massPrinting_job_scheduled_to_workplace`, `massPrinting_cross_warehouse_picking_group`, `massPrinting_skip_locked_schedule`): `validate M_HUs` asserting the LU `Qty` = starting − packed (e.g. 8 − 4 = 4). All three pass locally. https://github.com/metasfresh/metasfresh/commit/d4fd3a53074

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/massprinting/MassPrintingService.java
teosarca: **Resolution (dev):** Done — renamed `luId` → `pickFromLUId` in `createAndPickJob` (and its in-method usages); `scan()`'s `luId` left unchanged. https://github.com/metasfresh/metasfresh/commit/d4fd3a53074

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/service/PickingJobService.java
teosarca: **Resolution (dev):** Done. Removed `resolveScheduleIdsForJobCreation` from `PickingJobService`; the assembly now lives in the caller `MassPrintingService` (private). `PickingJobService` exposes only the thin generic `listJobSchedules(PickingJobQuery)` → `pickingJobScheduleService.list(query.toPickingJobScheduleQuery())` (returns EMPTY when not in scheduled-to-workplace mode). https://github.com/metasfresh/metasfresh/commit/d4fd3a53074

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/massprinting/MassPrintingService.java
teosarca: **Resolution (dev):** Done — trimmed to a single concise line. https://github.com/metasfresh/metasfresh/commit/d4fd3a53074

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/warehouse/WarehouseCommand.java
teosarca: **Resolution (dev):** Done. Added `org.adempiere.warehouse.api.CreateWarehousePickingGroupRequest` + `IWarehouseDAO.createWarehousePickingGroup(request)` (impl in `WarehouseDAO`); `WarehouseCommand.assignPickingGroup` now calls the DAO instead of `InterfaceWrapperHelper.newInstance(...)`. Capturing the broader learning you flagged (prefer DAO/Repository over direct InterfaceWrapperHelper/IQueryBL — also in test masterdata code) into our team conventions. https://github.com/metasfresh/metasfresh/commit/d4fd3a53074

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/warehouse/WarehouseCommand.java
teosarca: **Resolution (dev):** Done — the picking-group `Name` now uses `pickingGroupIdentifier.toUniqueString()` (unique per run → no cross-test collision). `M_Warehouse_PickingGroup.Name` is already `VARCHAR(255)` (verified in the create-table DDL and on the live DB), so no widening migration is needed. Capturing the broader learning (test masterdata must use unique names to stay independent) into our conventions. https://github.com/metasfresh/metasfresh/commit/d4fd3a53074

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/service/PickingJobService.java
teosarca: **Resolution (dev):** The early `selectedShipmentScheduleIds.isEmpty()` guard returns an empty set so we neither build a set nor hit the DB when nothing was selected. With the C3 refactor this logic now lives in the caller (`MassPrintingService`), where the guard is kept with an explicit intent comment. https://github.com/metasfresh/metasfresh/commit/d4fd3a53074

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/warehouse/WarehouseCommand.java
```diff
@@ -97,6 +104,44 @@ private void createWarehouse()
...
+				});
+
+		warehouseRecord.setM_Warehouse_PickingGroup_ID(pickingGroupId.getRepoId());
+		saveRecord(warehouseRecord);
+
+		// Reset the warehouse / picking-group caches so a later masterdata request (same JVM Spring context)
+		// sees the freshly-assigned group — mirrors M_Warehouse_StepDef. Without this the picking-group index
+		// can be read stale and MassPrintingService.getLocatorIdsOfTheSamePickingGroup misses this warehouse.
+		CacheMgt.get().reset(I_M_Warehouse.Table_Name);
+		CacheMgt.get().reset(I_M_Warehouse_PickingGroup.Table_Name);
```
teosarca: u don't need CacheMgt.reset calls. they shall happen automatically on given records save.
capture this learning

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/mobileui/picking/massPrinting_job_scheduled_to_workplace.feature
```diff
@@ -0,0 +1,172 @@
...
+      | schedScheduled                   | shipment              | CO            |
+
+    # The unscheduled schedule was NOT touched: still fully open (QtyToDeliver=4, nothing picked, nothing delivered).
+    And after not more than 10s, validate shipment schedules:
+      | M_ShipmentSchedule_ID.Identifier | OPT.QtyToDeliver | OPT.QtyDelivered | OPT.QtyPickList |
+      | schedNotScheduled                | 4                | 0                | 0               |
+
+    # Leak-safety teardown.
+    And delete all C_Workplace_User_Assign records
+    And deactivate all C_Workplace records
```
teosarca: ok to clean-up at teardown, but the most imporant is to cleanup on setUp. there is the right place to make sure u are running in the proper context.
capture the learning!

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/warehouse/WarehouseCommand.java
```diff
@@ -26,6 +31,7 @@ public class WarehouseCommand
 {
 	// services
 	@NonNull private final IWarehouseBL warehouseBL = Services.get(IWarehouseBL.class);
+	@NonNull private final IWarehouseDAO warehouseDAO = Services.get(IWarehouseDAO.class);
```
teosarca: repo/DAO and service/BL injectected in the same class shall be avoided.
u shall have a rule for this. pls RCA why it wasnt applied

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/massprinting/MassPrintingService.java
```diff
@@ -368,6 +336,43 @@ private PickingJob createAndPickJob(
...
+			@NonNull final PickingJobQuery query,
+			@NonNull final Set<ShipmentScheduleId> selectedShipmentScheduleIds)
+	{
+		// Nothing selected → nothing to create or query for. Guard so we neither build an empty set nor hit the DB.
+		if (selectedShipmentScheduleIds.isEmpty())
+		{
+			return ShipmentScheduleAndJobScheduleIdSet.EMPTY;
+		}
+
+		// listJobSchedules returns EMPTY both in warehouse mode (isScheduledForWorkplaceOnly=false) and in
```
teosarca: important: if isScheduledForWorkplaceOnly=true  (in profile)
and jo job schedules were found you MUST accept that. Don't fallback to searching the shipment scheudles, that against the purpose of isScheduledForWorkplaceOnly=true

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/warehouse/WarehouseCommand.java
teosarca: **Resolution (dev):** Done — `WarehouseCommand` now injects only `IWarehouseBL`; added `IWarehouseBL.createWarehousePickingGroup(request)` delegating to the DAO (per service-injection.md: "add a thin method to the service that delegates to the repository"). Same defect also found+fixed in `M_Warehouse_StepDef`: https://github.com/metasfresh/metasfresh/commit/460c4a72ead

**RCA — why the rule wasn't applied:** the rule exists — `docs/coding-rules/service-injection.md` § "Don't dual-source", a Review-rule that names this exact pair (`IWarehouseBL`+`IWarehouseDAO`) and says *"judge against the full class, not just the diff."* It was violated because (1) the DAO field was added to satisfy an earlier comment ("introduce a DAO method") in isolation, without noticing the class already injected the BL; and (2) the automated code-review that followed read the added-line hunk but did not apply the rule's explicit full-class check, so it missed the pre-existing BL field. Author-step and review-step both failed to honour an existing marked rule. https://github.com/metasfresh/metasfresh/commit/243a807e200

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/massprinting/MassPrintingService.java
teosarca: **Resolution (dev):** Done — `resolveScheduleIdsForJobCreation` now branches on the MODE, not on emptiness: warehouse mode → `ofShipmentScheduleIds`; `isScheduledForWorkplaceOnly` mode → driven ONLY by `listJobSchedules`, never falling back to a plain shipment-schedule set (a selected schedule with no job-schedule is dropped, not back-filled — defeating scheduled-mode is exactly what we avoid). The job-scheduled cucumber scenario confirms only the scheduled demand is packed and the unscheduled order stays fully open. https://github.com/metasfresh/metasfresh/commit/243a807e200

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/warehouse/WarehouseCommand.java
teosarca: **Resolution (dev):** Done — removed the manual `CacheMgt.get().reset(...)` calls from `assignPickingGroup`. You're right: CCache auto-invalidates on record save, and `WarehouseDAO.allWarehousePickingGroups` already declares `additionalTableNameToResetFor(M_Warehouse_PickingGroup)` so creating a group invalidates the index automatically. Learning captured (no manual `CacheMgt.reset` after `saveRecord` — the cache resets itself on save). https://github.com/metasfresh/metasfresh/commit/243a807e200

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/mobileui/picking/massPrinting_job_scheduled_to_workplace.feature
teosarca: **Resolution (dev):** Done — the workplace-isolation cleanup (`delete all C_Workplace_User_Assign` + `deactivate all C_Workplace`) is in the **Background (setUp)** of all three mass-printing features; removed the redundant per-scenario teardown copies. Learning captured (isolation cleanup belongs at setUp — that's where a clean starting context is guaranteed regardless of what prior scenarios left). https://github.com/metasfresh/metasfresh/commit/243a807e200

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/warehouse/M_Warehouse_StepDef.java
```diff
@@ -91,6 +95,66 @@ public void load_M_Warehouse(@NonNull final DataTable dataTable)
...
+		final WarehousePickingGroupId pickingGroupId = warehouseBL.createWarehousePickingGroup(
+				CreateWarehousePickingGroupRequest.builder()
+						.orgId(StepDefConstants.ORG_ID)
+						.name(groupName + "_" + System.currentTimeMillis())
+						.build());
+
+		rows.forEach(row -> {
+			final WarehouseId warehouseId = warehouseTable.getId(row.getAsIdentifier(COLUMNNAME_M_Warehouse_ID));
+
+			final org.compiere.model.I_M_Warehouse warehouseRecord = InterfaceWrapperHelper.load(
```
teosarca: avoid  InterfaceWrapperHelper.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/warehouse/M_Warehouse_StepDef.java
```diff
@@ -91,6 +95,66 @@ public void load_M_Warehouse(@NonNull final DataTable dataTable)
...
+			warehouseRecord.setM_Warehouse_PickingGroup_ID(pickingGroupId.getRepoId());
+			saveRecord(warehouseRecord);
+		});
+
+		// The picking-group → warehouses index is cached (WarehouseDAO.allWarehousePickingGroups). The bulk
+		// save fires a CacheInvalidation event, but in the shared single-JVM executor that event may not be
+		// processed before the next step reads the index. Reset the relevant caches synchronously so the
+		// freshly-assigned picking group is visible immediately (same trap handled in C_Workplace_StepDef).
+		CacheMgt.get().reset(I_M_Warehouse.Table_Name);
+		CacheMgt.get().reset(I_M_Warehouse_PickingGroup.Table_Name);
```
teosarca: no cache reset!

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/massprinting/MassPrintingService.java
```diff
@@ -368,6 +335,48 @@ private PickingJob createAndPickJob(
...
+		// Nothing selected → nothing to create or query for. Guard so we neither build an empty set nor hit the DB.
+		if (selectedShipmentScheduleIds.isEmpty())
+		{
+			return ShipmentScheduleAndJobScheduleIdSet.EMPTY;
+		}
+
+		// Warehouse mode: the job is driven by plain shipment-schedule ids.
+		if (!query.isScheduledForWorkplaceOnly())
+		{
+			return ShipmentScheduleAndJobScheduleIdSet.ofShipmentScheduleIds(selectedShipmentScheduleIds);
```
teosarca, line 366: even in this case u still have to check if those selectedShipmentScheduleIds are matching query, or this is already done?

teosarca: **Resolution (dev):** Already done upstream — no change needed. In scheduled mode `PickingJobService.streamPackageable(query)` sets `onlyShipmentScheduleIds = jobSchedules.getShipmentScheduleIds()` (the `isScheduledForWorkplaceOnly` branch), so the packageables — and therefore the `selectedShipmentScheduleIds` reaching `resolveScheduleIdsForJobCreation` — are by construction a subset of the job-scheduled shipment schedules. `listJobSchedules(query)` returns those same schedules, so `getSingleScheduleByShipmentScheduleId` matches every selected id; the `orElse(null)`+filter is defensive only and never drops in practice. Want me to add that guarantee as a Javadoc note (or harden it to throw on a non-match instead of silently dropping)?

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/warehouse/M_Warehouse_StepDef.java
teosarca: **Resolution (dev):** Done — removed the `CacheMgt.get().reset(...)` calls. You're right, and I was wrong to keep them in the prior round: CCache invalidates automatically on the warehouse/group saves, and `WarehouseDAO.allWarehousePickingGroups` already declares `additionalTableNameToResetFor(M_Warehouse_PickingGroup)`. I've corrected the captured rule accordingly — no manual reset, including in step defs. https://github.com/metasfresh/metasfresh/commit/587545389b8

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/warehouse/M_Warehouse_StepDef.java
teosarca: **Resolution (dev):** Done — moved the warehouse load+assign into `WarehouseDAO.createWarehousePickingGroup`: the request now carries the warehouse ids and the DAO creates the group and assigns each warehouse (`InterfaceWrapperHelper` there is the DAO's own job). The stepdef now calls `warehouseBL.createWarehousePickingGroup(...warehouseIds...)` once — no `InterfaceWrapperHelper` in the step. https://github.com/metasfresh/metasfresh/commit/587545389b8

PR #24559, opened 2026-06-14
ArchUnit — static architecture enforcement (thin POC)

teosarca:
## What

Thin proof-of-concept introducing **ArchUnit** for static architecture enforcement in the backend, wired at two levels with a minimal frozen ruleset that gates CI on **new** violations only.

Issue: https://github.com/metasfresh/me03/issues/30396 · initiative: https://github.com/metasfresh/mf15/issues/4215

## How

- **`misc/parent-pom`** manages `com.tngtech.archunit:archunit` 1.4.2 (core, test scope). Core artifact (not `archunit-junit5`): the rules run from plain JUnit **Jupiter `@Test`** methods, so there is no ArchUnit JUnit-Platform engine to clash with the pinned `junit-platform`. `slf4j-api` is excluded (ArchUnit pulls slf4j 2.x; the backend runs 1.7.x).
- **Per-module** — `de.metas.business` gets `ArchitectureTest` with 3 rules: persistence primitives confined to `*Repository`/`*DAO` (`service-injection.md` §4), no `java.sql.Timestamp` fields (`java-time.md` §2), no `Env` ambient client/org accessors in `@Service`/`@Component`/`@Repository`/`*BL` (`service-injection.md` §7).
- **Aggregator** — new `de.metas.architecture-tests` module (no production code) runs a `slices().beFreeOfCycles()` check over a scoped set (`de.metas.business` + `de.metas.adempiere.adempiere.base`) — the placement for cross-module invariants.
- Every rule is wrapped in **`FreezingArchRule`**: the existing legacy violations are baselined into committed `archunit_store/` files; only **new** violations fail the build. Verified locally: business rules pass against the baseline, a deliberately-added `Timestamp` field correctly **fails** the build, and the aggregator passes.

## Scope notes

- **Out of scope (separate follow-up):** a SQL / migration / AD-metadata linter for the conventions ArchUnit cannot see (bytecode-only).
- **Not wired across all modules:** deliberately a thin POC — one module + one small aggregator.
- The aggregator's no-cycles baseline store is large (legacy cycle reality, frozen); scope/granularity tuning is a next step before broad rollout.


backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
```diff
@@ -0,0 +1,134 @@
...
+ * methods rather than copy-pasting rule bodies. In this POC the central {@link ModuleArchitectureTest}
+ * applies them to {@code de.metas.business}; the same methods can be wrapped in a per-module abstract base
+ * test if/when the codebase moves to distributed per-module enforcement (see
+ * docs/coding-rules/archunit-backlog.md).
+ * <p>
+ * The rules are returned <i>raw</i> (un-frozen); the caller wraps them in
+ * {@link com.tngtech.archunit.library.freeze.FreezingArchRule} so each call site keeps its own baseline.
+ * Each rule cites the corpus rule it enforces in its {@code .because()} clause.
+ */
+public final class MetasfreshArchRules
```
teosarca, line 46: i think we shall have only one public method to call, java calls all the other mthods and joins the results.

aim: in case i am adding a new "chapter" i don't want to change each class that is using it.

teosarca: **Resolution (dev):** Done. `MetasfreshArchRules` now exposes a single gating entry point — `checkAllModuleRules(JavaClasses)` — which freezes + evaluates every rule and joins the results (one run reports all new violations). The individual rule factories are now `private`, and `ModuleArchitectureTest` is a single `@Test` that just calls `checkAllModuleRules`. Adding a new "chapter" = add a private rule method + one line in `checkAllModuleRules`; no caller changes. (The no-cycles rule was also moved into `MetasfreshArchRules`, still report-only, so every rule lives in the one place.) https://github.com/metasfresh/metasfresh/commit/2b731d9674df48d9ed62f135f96188804451fe6a

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
```diff
@@ -0,0 +1,374 @@
...
+
+	/**
+	 * Derive a stable owning-module label from a class's source jar (e.g. {@code de.metas.business-10.0.0.jar} →
+	 * {@code de.metas.business}): the jar's artifact id with the {@code -<version>} suffix stripped. The label is
+	 * identical whether the jar is resolved from {@code .m2-local} (CI / maven) or extracted from a Spring Boot fat
+	 * jar (baseline generation), so the freeze key is reproducible. Classes not loaded from a jar (e.g. an
+	 * assembly's own {@code target/classes}) yield {@code "unknown-module"}; in CI those classes carry their real
+	 * jar key instead and simply auto-record.
+	 */
+	private static String moduleLabelOf(final JavaClass cls)
```
teosarca, line 168: can u pls check if this rule is always reliable?!

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
```diff
@@ -0,0 +1,374 @@
...
+		// NOTE: Maven artifact names (e.g. "de.metas.business") do not correspond to Java package names.
+		// de.metas.business artifact puts classes under de.metas.invoice, de.metas.invoicecandidate, etc.
+		// We probe for distinct Java sub-packages that are definitively contributed by different modules:
+		//   invoice         → de.metas.invoice          (from de.metas.business jar)
+		//   handlingunits   → de.metas.handlingunits     (from de.metas.handlingunits.base)
+		//   contracts       → de.metas.contracts         (from de.metas.contracts)
+		//   payment         → de.metas.payment           (from de.metas.payment.* jars)
+		//   manufacturing   → de.metas.manufacturing     (from de.metas.manufacturing)
+		final List<String> missingModules = new ArrayList<>();
+		for (final String mandatorySubPkg : Arrays.asList(
```
teosarca: how this relates to moduleLabelOf ?


maybe we shall extract this module <-> package rules in a separate java class that models them and make it more obvious. pls eval

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
teosarca: **Resolution (dev):** Evaluated reliability. `moduleLabelOf` strips only the `-<version>` suffix (a metasfresh artifact version always starts with a digit), verified against all 139 module labels actually produced from the serverRoot+webui-api closure — including the tricky ones (`de.metas.banking.camt53`, `de.metas.payment.sepa.schema.pain_001_01_03_ch_02`, `metasfresh-pharma.msv3.*`, `de.metas.business.rest-api-impl`, `vertical-healthcare_ch...`): none has a hyphen-then-digit segment other than the version, so all derive correctly. Stability confirmed two ways — an idempotent re-record (a 2nd evaluation against the committed store reports zero new violations) and CI green. Safety net: even a mis-derived label cannot cause a false build failure — with `allowStoreCreation=true` a class under an unexpected key simply auto-records (worst case: enforced under an odd key / unenforced, never red). Non-jar classes fall back to `unknown-module` (a defensive no-fail path); in CI every class comes from a jar so it does not occur. No code change: a guard for a scenario that cannot fail would be cargo-cult.

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
teosarca: **Resolution (dev):** Evaluated. These were two different notions of "module": this probe checked Java **package** presence (with a hardcoded artifact→package mapping comment), while `moduleLabelOf` keys on the **source jar**. There is no real module↔package engine to extract — that mapping was only an explanatory comment for the package proxy — so a separate class would model logic that is not actually used. Instead I unified on the existing jar-based notion: the probe now collects distinct owning modules via `moduleLabelOf` and asserts well-known module **jars** are present (`de.metas.business`, `de.metas.handlingunits.base`, `de.metas.contracts`, `de.metas.payment.sepa.base`, `de.metas.manufacturing`). One module model (the Maven artifact), package-mapping magic removed, and the probe now relates to `moduleLabelOf` directly (same keys). Commit https://github.com/metasfresh/metasfresh/commit/2bf102eeb0c . If you would still prefer a dedicated `ModuleIdentity` class, happy to add one — but it would wrap a single one-line derivation, so I judged unifying the notion in place clearer than a new abstraction.

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
```diff
@@ -0,0 +1,365 @@
...
+		return fileBase.replaceFirst("-\\d.*$", "");
+	}
+
+	/** Import a single module's own compiled classes (jar or target/classes), failing loudly on an empty set. */
+	public static JavaClasses importModule(final String moduleLocationToken)
+	{
+		final JavaClasses classes = new ClassFileImporter()
+				.withImportOption(new ImportOption.DoNotIncludeTests())
+				.withImportOption((final Location location) -> location.contains(moduleLocationToken))
+				.importPackages("de.metas", "org.adempiere", "org.compiere", "org.eevolution");
```
teosarca: extract this list/array as constant. it's used in many places!

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
```diff
@@ -0,0 +1,365 @@
...
+	 *
+	 * @throws IllegalStateException if zero classes were imported, or if a mandatory module is absent
+	 */
+	public static JavaClasses importWholeBackend()
+	{
+		final long importStart = System.currentTimeMillis();
+
+		final JavaClasses classes = new ClassFileImporter()
+				.withImportOption(new ImportOption.DoNotIncludeTests())
+				.importPackages("de.metas", "org.adempiere", "org.compiere", "org.eevolution");
```
teosarca: extract this list/array as constant. it's used in many places!

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
```diff
@@ -0,0 +1,365 @@
...
+	{
+		return new DescribedPredicate<JavaMethodCall>("a call to an Env ambient client/org accessor")
+		{
+			@Override
+			public boolean test(final JavaMethodCall call)
+			{
+				final String owner = call.getTargetOwner().getFullName();
+				final String method = call.getName();
+				return "org.compiere.util.Env".equals(owner)
+						&& ("getAD_Client_ID".equals(method)
```
teosarca, line 366: why not any "get" starting method?

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
```diff
@@ -0,0 +1,365 @@
...
+	 * Cross-module: {@code de.metas} bounded contexts must be free of dependency cycles. Relates to
+	 * docs/coding-rules/architecture.md §8 (bounded-context dependency discipline).
+	 * <p>
+	 * <b>Returned for report-only use, NOT for freezing.</b> {@code beFreeOfCycles()} violation descriptions
+	 * are non-deterministic run-to-run, so a {@code FreezingArchRule} baseline would be flaky — callers should
+	 * {@code evaluate()} this and log, not {@code freeze(...).check(...)}. See skill {@code metasfresh-archunit}.
+	 */
+	public static ArchRule boundedContextsFreeOfCycles()
+	{
+		return slices().matching("de.metas.(*)..").should().beFreeOfCycles()
```
teosarca, line 336: why only "de.metas" and not "de.metas", "org.adempiere", "org.compiere", "org.eevolution".
i.e. why not sharing the list of package roots?

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
```diff
@@ -0,0 +1,365 @@
...
+			moduleLabels.add(moduleLabelOf(cls));
+		}
+
+		logger.info("[ArchUnit whole-backend] Total classes imported: {} (import took {} ms)", classes.size(), importMs);
+		logger.info("[ArchUnit whole-backend] Distinct owning modules ({}): {}", moduleLabels.size(), moduleLabels);
+
+		// Probe: assert a few well-known module jars are present — a missing one means the serverRoot/webui-api
+		// transitive closure did not fully load, so the per-module freeze would silently under-cover. These are
+		// Maven artifact ids (= the #moduleLabelOf grouping key), each definitively contributed by a distinct module.
+		final List<String> missingModules = new ArrayList<>();
```
teosarca: still i don't get what's with those hardcoded packages, how did u got them and how do u know they are all?

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
teosarca: **Resolution (dev):** Done — extracted the four analysed package roots into a single `BACKEND_PACKAGE_ROOTS` constant, now shared by `importModule` and `importWholeBackend` (the literal was duplicated across both). https://github.com/metasfresh/metasfresh/commit/f161c1baca9

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
teosarca: **Resolution (dev):** Done — same fix as the line-187 occurrence: both call sites now use the shared `BACKEND_PACKAGE_ROOTS` constant instead of the duplicated literal. https://github.com/metasfresh/metasfresh/commit/f161c1baca9

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
teosarca: **Resolution (dev):** Intentional. The rule enforces docs/coding-rules/service-injection.md §7 — do not read the ambient *client/org* from Env — not all Env access. Only the four client/org accessors qualify (`getAD_Client_ID`/`getClientId`/`getAD_Org_ID`/`getOrgId`). Matching any `get*` on Env would flag legitimate calls (`getCtx`, `getProperty`, date/language getters, ...) as false positives. Broader ambient reads (e.g. `Env.getCtx()`-derived) are deliberately out of scope for now and tracked in docs/coding-rules/archunit-backlog.md.

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
teosarca: **Resolution (dev):** Two parts. (1) Scope: the cycle check targets `de.metas.<x>` bounded contexts; `org.adempiere`/`org.compiere`/`org.eevolution` are the flat legacy ADempiere base — no bounded-context structure to slice and known pre-existing cycles — so including them would only emit legacy noise on a report-only check. Documented on the method in https://github.com/metasfresh/metasfresh/commit/f161c1baca9 . (2) Sharing the roots: the cycle rule uses a slice-matching glob (`de.metas.(*)..`), a different construct from the `importPackages(...)` varargs, so it cannot consume the `BACKEND_PACKAGE_ROOTS` array directly; the import roots ARE now a shared constant where they apply (both importers).

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
teosarca: **Resolution (dev):** Fair point — there is no authoritative complete list and a hardcoded one would rot, so I removed the five hardcoded names. The probe now asserts the closure is *substantial* (distinct module count >= 50; a healthy serverRoot+webui-api closure yields ~140) — which is what we actually care about: the zero-class guard catches nothing-loaded, this catches a partial sliver (assembly deps not fully loaded). The full module list is logged for inspection. https://github.com/metasfresh/metasfresh/commit/f161c1baca9

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
teosarca: what's happening if somebody by mistake/or not adds a class into one of those modules because it just made sense having near some classes/packages?

teosarca: **Resolution (dev):** Good question — three cases, by which rule reacts to a newly-added class:

1. **The three frozen rules (InterfaceWrapperHelper / java.sql.Timestamp / Env-ambient) are class-local and keyed per owning module (jar), not per package.** A new class is checked against *its* module's baseline wherever you place it. If that module already has a committed baseline and the new class introduces one of these violations, it is NOT in the baseline → **the build fails** (the ratchet, working as intended). So a mistaken violating class in an already-tracked module is caught regardless of which package it sits in.

2. **If the class lands in a module that has no committed baseline yet** (a not-yet-baselined or brand-new module) → `allowStoreCreation=true` auto-records its violations and **passes** (unenforced until that module is baselined). This is the deliberate graceful-degradation tradeoff — coverage grows incrementally and CI never red-fails on an unbaselined module — and it is the one path where a fresh violation can slip in unnoticed.

3. **Placement that creates a cross-context dependency cycle** (what this method checks): `boundedContextsFreeOfCycles()` *would* surface it, but it is **report-only / opt-in** (`-Darchunit.cycles=true`, skipped on normal CI). Cycle-violation descriptions are non-deterministic run-to-run, so freezing them would make the gate flaky — hence report-only until the scope is small enough for stable descriptions.

Net: none of these rules enforce *correct* placement itself — that stays a design-review concern. What is guaranteed is that a new IWH/Timestamp/Env violation in an already-baselined module fails the build; the two by-design gaps are (2) unbaselined modules and (3) cycles (report-only). Happy to tighten either — commit a given module's baseline to enforce it, or make no-cycles gating for a chosen small scope — but both are current deliberate tradeoffs, not oversights.

teosarca: **Resolution (dev):** Stepping back to what this rule actually does — `boundedContextsFreeOfCycles()`:

- It carves every `de.metas.*` class into **slices by first sub-package**: `de.metas.invoice.*` → slice `invoice`, `de.metas.handlingunits.*` → slice `handlingunits`, etc. Each slice ≈ one bounded context.
- It then asserts the **dependency graph between those slices is acyclic** — no `invoice → handlingunits → … → invoice` loop. A cycle means two contexts depend on each other (directly or transitively), so neither can be built, tested, or reasoned about on its own — the classic tangled-architecture smell. `beFreeOfCycles()` passes when the slice graph is a DAG, fails when any cycle exists.

Tie-in to your earlier question: if someone adds a class to `de.metas.invoice` that calls into `de.metas.handlingunits` while handlingunits already calls into invoice, that **creates a cross-context cycle** — and this is the rule that surfaces it. In this PR it surfaces it **report-only** (logs the cycle count, only under `-Darchunit.cycles=true`): it informs, it does not block. The three frozen rules (InterfaceWrapperHelper / java.sql.Timestamp / Env-ambient) are the ones that actually fail the build. (Why report-only: at ~167-module scale the cycle-violation descriptions are non-deterministic run-to-run, so a frozen baseline would be flaky — so it runs on-demand/nightly until the scope is small enough for stable output.)

backend/de.metas.business/src/test/resources/archunit.properties
```diff
@@ -0,0 +1,5 @@
+# ArchUnit configuration for de.metas.business.
+# allowStoreCreation=true: the first run (no store) records the legacy baseline into ./archunit_store
+# (committed to git). On later runs ArchUnit reports NEW violations as failures and prunes fixed ones from
+# the store. Pay down legacy debt by fixing a violation and re-running (the store entry is removed).
+freeze.store.default.allowStoreCreation=true
```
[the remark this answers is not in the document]
teosarca: **Resolution (dev):** Valid concern, now fixed. The file moved to `de.metas.architecture-tests` in the per-module redesign, and `ModuleArchitectureTest` now **fails loudly** if `archunit_store/stored.rules` is absent — so a missing/uncommitted store can no longer silently re-record and pass green. `allowStoreCreation=true` is retained deliberately (a not-yet-baselined module must auto-record, not red-fail); the new guard distinguishes whole-store-gone (fail) from a new module (store present, auto-records). https://github.com/metasfresh/metasfresh/commit/0d994b4be3f

backend/de.metas.architecture-tests/src/test/resources/archunit.properties
```diff
@@ -0,0 +1,4 @@
+# ArchUnit configuration for the cross-module architecture-tests aggregator.
+# allowStoreCreation=true: the first run (no store) records the legacy baseline into ./archunit_store
+# (committed to git). On later runs ArchUnit reports NEW cycles as failures and prunes fixed ones.
+freeze.store.default.allowStoreCreation=true
```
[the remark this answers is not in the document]
teosarca: **Resolution (dev):** Fixed — this is the live occurrence. `ModuleArchitectureTest` now throws if `archunit_store/stored.rules` is missing, so a deleted/uncommitted store fails the build instead of silently re-recording (classes still load, so the zero-class and module-count guards would not have caught it). `allowStoreCreation=true` stays for per-module graceful degradation; the guard separates whole-store-gone (fail) from a genuinely new module (store dir present, just lacks that key → auto-records). https://github.com/metasfresh/metasfresh/commit/0d994b4be3f

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/CrossModuleArchTest.java
```diff
@@ -0,0 +1,62 @@
...
+	private static JavaClasses scopedModuleClasses;
+
+	@BeforeAll
+	static void importScopedModuleClasses()
+	{
+		scopedModuleClasses = new ClassFileImporter()
+				.withImportOption(new ImportOption.DoNotIncludeTests())
+				.withImportOption(CrossModuleArchTest::isScopedModuleClass)
+				.importPackages("de.metas", "org.adempiere", "org.compiere", "org.eevolution");
+	}
```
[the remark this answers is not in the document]
teosarca: **Resolution (dev):** Outdated — the per-module redesign removed `scopedModuleClasses` and the location-filtered cycle import. The no-cycles rule now runs report-only over the whole-backend import, which is guarded by `importWholeBackend()`'s zero-class throw plus a >=50 distinct-module substantiveness floor, so it cannot become a vacuous no-op.

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/CrossModuleArchTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+	/** Restrict the import to the scoped POC modules' compiled output (jar or target/classes). */
+	private static boolean isScopedModuleClass(final Location location)
+	{
+		return location.contains("/de.metas.business/")
+				|| location.contains("/de.metas.adempiere.adempiere.base/");
+	}
```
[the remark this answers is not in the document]
teosarca: **Resolution (dev):** Outdated — `isScopedModuleClass` was removed. The current importers analyse jar-sourced classes (`moduleLabelOf` keys on the source jar; CI runs against jars — verified ~139 modules), and both `importWholeBackend()` and `importModule()` throw on an empty import. The dir-only-match leading to zero classes no longer applies.

backend/de.metas.business/src/test/java/de/metas/architecture/ArchitectureTest.java
```diff
@@ -0,0 +1,150 @@
...
+	@BeforeAll
+	static void importThisModuleClasses()
+	{
+		// Import this module's own source roots, restricted to its compiled output so dependency jars
+		// (de.metas.adempiere.adempiere.base, …) are NOT pulled into this per-module baseline.
+		thisModuleClasses = new ClassFileImporter()
+				.withImportOption(new ImportOption.DoNotIncludeTests())
+				.withImportOption(ArchitectureTest::isThisModuleClass)
+				.importPackages("de.metas", "org.adempiere", "org.compiere", "org.eevolution");
+	}
```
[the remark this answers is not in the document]
teosarca: **Resolution (dev):** Outdated — `isThisModuleClass` and the `de.metas.business` ArchitectureTest were removed in the per-module redesign. Vacuous-pass is now prevented three ways: `importWholeBackend()`'s zero-class guard, the >=50 distinct-module floor, and the new missing-store guard (https://github.com/metasfresh/metasfresh/commit/0d994b4be3f).

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
```diff
@@ -0,0 +1,373 @@
...
+	}
+
+	/**
+	 * Persistence primitives confined to {@code *Repository}/{@code *DAO} — BL/Service/command/interceptor/
+	 * process code must not call {@code InterfaceWrapperHelper.save|saveRecord|load} directly. Enforces the
+	 * {@code InterfaceWrapperHelper} save/load portion of docs/coding-rules/service-injection.md §4 +
+	 * java-general.md §18 (the {@code IQueryBL}/{@code DB.*} portion is a separate candidate — see
+	 * docs/coding-rules/archunit-backlog.md).
+	 */
+	private static ArchRule persistencePrimitivesConfinedToRepositoryOrDao()
```
metas-ts, line 277: imo makes sense, also for cucumber-stepdefs. in future they can e.g. use and extend the masterdata API

metas-ts (approved):

backend/de.metas.architecture-tests/src/test/java/de/metas/architecture/MetasfreshArchRules.java
teosarca: **Resolution (dev):** Agreed — the same confinement makes sense for cucumber step definitions, and steering them to use/extend a masterdata API rather than calling `InterfaceWrapperHelper` directly is the right direction. It is out of scope for *this* PR: the analysis runs over the serverRoot + webui-api closure, and the `de.metas.cucumber` step-def classes are the test harness (not in that closure), so this rule does not reach them today. Captured as an ArchUnit-candidate backlog item (`docs/coding-rules/archunit-backlog.md`): extend persistence-primitive confinement to cucumber step definitions, paired with providing/extending a masterdata API they can build on.

PR #24567, opened 2026-06-15
DD_Order_Candidate: add DB column defaults for SQL-authored inserts

teosarca:
Adds DB column defaults on `DD_Order_Candidate` so bare-SQL inserts can omit the audit/flag boilerplate:

- `IsActive='Y'`, `Created=now()`, `CreatedBy=100`, `Updated=now()`, `UpdatedBy=100`
- (`QtyProcessed`, `Processed`, `IsSimulated` already have DB defaults.)

App-path (PO framework) inserts always set these explicitly, so the defaults only take effect for **direct SQL inserts** — e.g. scheduled `DD_Order_Candidate` generator functions, whose INSERT can then be trimmed.

Migration: a plain `ALTER … SET DEFAULT` (verified it applies and the defaults take effect).


teosarca:
Superseded by https://github.com/metasfresh/metasfresh/pull/24568 — re-cut from `deep_tundra_release` (this branch was accidentally based on `new_dawn_uat`, so it carried the whole new_dawn_uat↔deep_tundra_release delta). Same single migration.

PR #24569, opened 2026-06-15
Shipment schedule: reactivation and reopen-processed handle processed-but-not-closed schedules

adi-stefan:
## Problem
A sales-order line set to `QtyOrdered=0` (instead of being deleted) and re-completed makes its
`M_ShipmentSchedule` auto-flag `Processed=Y` while `IsClosed=N`. The order can then no longer be
reactivated (`ShipmentScheduleAlreadyProcessed`), and neither the reopen-processed process nor the close
process can recover it — the only escape was a manual database update.

## Fix
- `ShipmentScheduleBL.closeShipmentSchedulesFor` (order reactivation): close a processed-but-not-closed
  schedule instead of throwing. Safe because the `BEFORE_REACTIVATE` guard (`reactivateIfNoActiveShipment`)
  already guarantees the order has no active shipments, so such a schedule carries no real delivery work;
  the subsequent re-completion reopens and recomputes it.
- `ShipmentScheduleBL.openShipmentSchedule`: no longer asserts `IsClosed`; un-closes only if closed and
  always un-processes the schedule + flags for recompute. This also repairs the
  `M_ShipmentSchedule_OpenProcessed` recovery process, which previously failed on a not-closed schedule.

## Tests
- Unit (`ShipmentScheduleBLTest`): both methods on a processed-but-not-closed schedule; red→green, all green.
- Cucumber (`shipmentSchedule.feature`): regression — complete order → reactivate → zero a line →
  re-complete (schedule auto-processes) → reactivate (previously blocked) → delete the line → re-complete.


adi-stefan:
@claude review

PR #24571, opened 2026-06-15
PP_Order WIndow Improvements

metas-rc:


metas-rc:
@claude review opus

metas-rc:
@claude review opus

metas-rc:
✅ Reviewed — looks good from a code-review standpoint. - via local Claude

Checked against the metasfresh conventions (AD migration + window-design + persistence/coding rules):

- PP_Order.IsProdDateSet (Yes-No, default N) + checkbox on the production-order window (53009). ReadOnlyLogic = @IsProdDateSet@=Y on DateDelivered ("Eff. Prod. Datum") makes it read-only when the flag is set and editable when not; IsUpdateable='Y' clears the column's default hard read-only so the flag is what governs.
- "Geplanter Starttermin" (DateStartSchedule) as the first date filter via IsSelectionColumn / IsRangeFilter / SelectionColumnSeqNo at column level.
- Guard in PPOrderBL.addQty so that when the flag is set, the effective production date is no longer auto-overwritten by the receipt date — with unit tests covering both branches.

Migrations apply cleanly (db-apply-migrations green), all AD IDs come from the ID server, de_DE/de_CH/en_US translations and timestamp staggering are correct, and the generated model (I_/X_PP_Order) is in sync. Note the read-only/updateable logic is intentionally set at the column level (applies to PP_Order wherever shown).

LGTM. 👍

adi-stefan (approved):

PR #24577, opened 2026-06-15
MobileUI Distribution: IsPackingPlace workplace filter (me03 30429)

teosarca:
### What

Adds a new boolean **`C_Workplace.IsPackingPlace`** (default `'Y'`) that splits the mobile-Distribution launcher into two roles:

- **`IsPackingPlace='Y'` (default)** — behaviour unchanged: the launcher offers the workplace-warehouse DD_Orders whose target locator = `workplace.PickFrom_Locator_ID` (null ⇒ whole destination warehouse), exactly as before.
- **`IsPackingPlace='N'` (replenishment)** — the launcher offers the workplace-warehouse DD_Orders whose target locator is **NOT** a packing-place locator, so groundfill target moves become visible to replenishment workers. The workplace's own `PickFrom_Locator_ID` is ignored.

The excluded set is data-driven: the `PickFrom_Locator_ID` of every `IsPackingPlace='Y'` workplace in the destination warehouse, read off the existing cached `WorkplaceRepository` map (CCache auto-invalidates on `C_Workplace` change). Empty set ⇒ no exclusion (catch-all). The already-assigned pool (responsible = user) is untouched in both roles.

### How

- Migration `5807860_sys_C_Workplace_IsPackingPlace.sql`: AD_Column (Yes-No), default `'Y'`, NOT NULL after backfill, AD_Element (DE "Packplatz" / EN "Is Packing Place"), field on the C_Workplace window (flags group).
- `Workplace` / `WorkplaceRepository` / `WorkplaceService`: new `isPackingPlace` flag + cached `getPackingPlacePickFromLocatorIds(WarehouseId)`.
- `DDOrderQuery.excludeLocatorToIds` + a NOT-IN target-locator subquery in `DDOrderLowLevelDAO`, mirroring the existing IN subquery.
- `DistributionWorkflowLaunchersProvider` branches the launcher query on `IsPackingPlace`.

### Tests

- `WorkplaceRepositoryTest`, `DistributionJobQueriesTest` — unit coverage of the cached lookup + Y/N query construction. E2E follow-up in a separate PR.

### DIFF
--- a/backend/de.metas.distribution.rest-api/src/main/java/de/metas/distribution/mobileui/launchers/DistributionWorkflowLaunchersProvider.java
+++ b/backend/de.metas.distribution.rest-api/src/main/java/de/metas/distribution/mobileui/launchers/DistributionWorkflowLaunchersProvider.java
@@ -103,7 +103,7 @@ public class DistributionWorkflowLaunchersProvider
 
 		if (workplace != null && !workplace.isPackingPlace())
 		{
-			builder.excludeLocatorToIds(warehouseService.getPackingPlacePickFromLocatorIds());
+			builder.excludeLocatorToIds(warehouseService.getPackingPlacePickFromLocatorIds(workplace.getWarehouseId()));
 		}
 		else
 		{
--- a/backend/de.metas.distribution.rest-api/src/main/java/de/metas/distribution/mobileui/launchers/DistributionWorkflowLaunchersProvider.java
+++ b/backend/de.metas.distribution.rest-api/src/main/java/de/metas/distribution/mobileui/launchers/DistributionWorkflowLaunchersProvider.java
@@ -96,11 +96,21 @@ public class DistributionWorkflowLaunchersProvider
 		final MobileUIDistributionConfig config = getConfig();
 		final Workplace workplace = warehouseService.getWorkplaceByUserId(userId).orElse(null);
 
-		return DDOrderReferenceQuery.builder()
+		final DDOrderReferenceQueryBuilder builder = DDOrderReferenceQuery.builder()
 				.sorting(config.getSorting())
 				.responsibleId(userId)
-				.warehouseToId(workplace != null ? workplace.getWarehouseId() : null)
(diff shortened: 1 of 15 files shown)


backend/de.metas.distribution.rest-api/src/main/java/de/metas/distribution/mobileui/launchers/DistributionWorkflowLaunchersProvider.java
```diff
@@ -96,11 +96,21 @@ private DDOrderReferenceQueryBuilder newDDOrderReferenceQuery(@NonNull final Use
...
+		final DDOrderReferenceQueryBuilder builder = DDOrderReferenceQuery.builder()
 				.sorting(config.getSorting())
 				.responsibleId(userId)
-				.warehouseToId(workplace != null ? workplace.getWarehouseId() : null)
-				.locatorToId(workplace != null ? workplace.getPickFromLocatorId() : null);
+				.warehouseToId(workplace != null ? workplace.getWarehouseId() : null);
+
+		if (workplace != null && !workplace.isPackingPlace())
+		{
+			builder.excludeLocatorToIds(warehouseService.getPackingPlacePickFromLocatorIds());
```
teosarca: i suggest to introduce WarehouseId param to getPackingPlacePickFromLocatorIds because we are anyways filtering by warehouseToId and it does not make sense to get all the packing locators from all warehouses

teosarca: **Resolution (dev):** Done — `getPackingPlacePickFromLocatorIds` now takes a `WarehouseId` and returns only that warehouse's packing-place pick-from locators (threaded through `WorkplaceRepository` → `WorkplaceService` → `DistributionWarehouseService`); the launcher passes `workplace.getWarehouseId()`. Added a cross-warehouse exclusion case to `WorkplaceRepositoryTest`. Commit https://github.com/metasfresh/metasfresh/commit/80b9d4a7ecfcaefb783c9420d244e3a657c156dd

PR #24579, opened 2026-06-15
C_Invoice scripted-export status surfacing + generic export-status naming + re-send process

metas-ts:
## What

Surfaces scripted-export (DocuWare) status on the **sales invoice**, mirroring the existing shipment/`M_InOut` surfacing, and renames the export-status surfacing to **generic** wording (it is generic machinery, not EPCIS-specific).

### Core changes
- **Generic-naming rename (labels only):** the existing `M_InOut` export-status tab caption + column label change from EPCIS-specific text to generic `Exportstatus` / `Export Status`. Column identifiers are unchanged.
- **`C_Invoice.ScriptedExport_Status`** — a read-only virtual (`ColumnSQL`) rollup column aggregating the latest status from `ExternalSystem_ScriptedExportConversion_Status` for the invoice (same ranked-aggregate as the shipment column), with an `AD_SQLColumn_SourceTableColumn` cache link.
- **Status child tab** on the standard Sales-Invoice window (read-only, newest-first) showing per-config status rows (status, configuration, HTTP response code, message, re-sent, process instance, updated), plus the rollup field on the header tab (in the visible "dates" group next to *External System*).
- **Re-send process** `C_Invoice_ReSend_ScriptedExportConversion` bound to `C_Invoice` (table-level `AD_Table_Process`), re-triggering the scripted export for the selected invoice — the analogue of the shipment re-send process.
- **Model regen:** `I_/X_C_Invoice` regenerated for the new virtual column.

### Notes
- All new AD IDs allocated from the central id-server; migration scripts follow the standard timestamp/translation conventions; German base language with `en_US` overrides; all active system languages covered.
- The config field on the status tab uses a dedicated `AD_Name_ID` label element so the human-readable caption survives the translation sync.
- Unit test for the re-send process passes (2/2).

### DIFF
--- a/backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/scriptedexportconversion/process/C_Invoice_ReSend_ScriptedExportConversion.java
+++ b/backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/scriptedexportconversion/process/C_Invoice_ReSend_ScriptedExportConversion.java
@@ -1,0 +1,83 @@
+/*
+ * #%L
+ * de.metas.externalsystem
+ * %%
+ * Copyright (C) 2026 metas GmbH
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+package de.metas.externalsystem.scriptedexportconversion.process;
+
+import de.metas.externalsystem.ExternalSystemInvocationContext;
+import de.metas.externalsystem.scriptedexportconversion.ExternalSystemScriptedExportConversionConfig;
+import de.metas.externalsystem.scriptedexportconversion.ExternalSystemScriptedExportConversionConfigId;
+import de.metas.externalsystem.scriptedexportconversion.ExternalSystemScriptedExportConversionService;
+import de.metas.process.JavaProcess;
+import de.metas.process.ProcessPreconditionsResolution;
+import de.metas.process.IProcessPrecondition;
+import de.metas.process.IProcessPreconditionsContext;
+import lombok.NonNull;
+import org.adempiere.util.lang.impl.TableRecordReference;
+import org.compiere.SpringContextHolder;
+import org.compiere.model.I_C_Invoice;
+
+import java.util.List;
+
+/**
+ * Re-triggers the scripted-export-conversion for any config with a non-Sent attempt for this C_Invoice.
+ */
+public class C_Invoice_ReSend_ScriptedExportConversion extends JavaProcess implements IProcessPrecondition
+{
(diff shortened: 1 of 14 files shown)


backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/scriptedexportconversion/process/C_Invoice_ReSend_ScriptedExportConversion.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Re-triggers the scripted-export-conversion for any config with a non-Sent attempt for this C_Invoice.
+ */
+public class C_Invoice_ReSend_ScriptedExportConversion extends JavaProcess implements IProcessPrecondition
+{
+	private final ExternalSystemScriptedExportConversionService scriptedExportService =
+			SpringContextHolder.instance.getBean(ExternalSystemScriptedExportConversionService.class);
+
+	@Override
+	public ProcessPreconditionsResolution checkPreconditionsApplicable(@NonNull final IProcessPreconditionsContext context)
```
metas-ts, line 56: assume some outage tht causes a thousond invoices to fail bein send => we need
- the process to work on the current selection
- an option to send only those invoices from the selection that were no yet send successfully
- 💡 check out `de.metas.util.web.audit.process.API_Audit_Repeat` with respect to what i mean.
- If the inout ("epcis") process doesn't have this, then please augment that one, too

metas-ts: **Resolution (dev):** Implemented, mirroring `API_Audit_Repeat`.

Both re-send processes — `C_Invoice_ReSend_ScriptedExportConversion` **and** the shipment/EPCIS `M_InOut_ReSend_ScriptedExportConversion` (it had the same single-selection limitation) — now:

- **Work on the full multi-record selection** — `checkPreconditionsApplicable` rejects only an *empty* selection; `doIt` iterates `retrieveSelectedRecordsQueryBuilder(...).listIds()` (IDs materialised up-front, so the cursor is closed before the per-record re-trigger writes status rows).
- Carry a new Yes/No process parameter **`IsOnlyNotSentSuccessfully`**:
  - **No (current default):** re-send every selected record's terminal-status exports (`Sent`/`Error`/`Invalid`) — a full re-send of the selection. In-flight (`Pending`/`Enqueued`/`SendingStarted`) and `DontSend` are skipped, so nothing in flight is double-queued.
  - **Yes:** re-send only the not-yet-successfully-sent ones (`Error`/`Invalid`) — the outage-recovery case (select the whole batch, re-send only the failures).

Parameter added to both `AD_Process` records (585633 + 585637) via migration `5807930_sys_ReSend_ScriptedExport_OnlyNotSent_param.sql`; both process unit tests now cover multi-selection + both modes (green). On the PR branch (`fc39f74` … `b7c7ad4`).

**One thing to confirm:** I defaulted the parameter **off** (= re-send all selected, including already-successful exports), with the option *narrowing* to not-yet-sent — matching how `API_Audit_Repeat`'s `IsOnlyWithError` defaults. If you'd prefer the safer default to be **on** (never re-send an already-successful export unless explicitly forced), say so and I'll flip the `DefaultValue` to `Y`.

metas-ts: Done — `DefaultValue` for `IsOnlyNotSentSuccessfully` is now `Y` on both processes (585633 + 585637): the safe default re-sends **only** the not-yet-successfully-sent records; unchecking it force-resends the whole selection. Migration `5807930` updated, verified in DB. Commit `6de0b203c2d`.

PR #24613, opened 2026-06-17
Stock per week: single-scan rewrite of MD_Stock_PerWeek_V for performance

metas-ts:
## What
Single-scan rewrite of the SQL view `MD_Stock_PerWeek_V` (Feature F19100 "Bestand pro Woche").

The view computed **3 correlated subqueries per output row**; on production-volume data (≈782k rows over product×warehouse×52 weeks) this made the window unusable. The rewrite computes the three measures from essentially one grouped scan of `MD_Candidate` per product×warehouse:
- **Expected shipments / receipts** — single grouped scans, bucketed into the week via `GREATEST(week-of(DateProjected), current-week)` (overdue rolled into the current week), exactly as before.
- **ATP at week-end** — replaces the per-row `DISTINCT ON … ORDER BY` with one pass: collapse STOCK candidates per `(product, warehouse, StorageAttributesKey, C_BPartner_Customer_ID, DateProjected)` keeping the max-`SeqNo` Qty, then `LEAD(DateProjected)` gives each step a half-open interval; a step owns week W iff `DateProjected < W+7 AND (next IS NULL OR next >= W+7)` — the same timestamp comparisons as the original.
- Weeks via a LATERAL `generate_series`; the synthetic MD5 hash PK and all 9 output columns are byte-identical.

## Output is identical to the previous view (no functional change)
Verified by a full bidirectional set-diff against a snapshot of the old view on loaded production-volume data: **`old_not_new=0 / new_not_old=0`** at HorizonWeeks 52 **and** 12, zero per-column mismatches, on the migrated real view. The multi-customer-subgroup path was independently proven byte-identical via a synthetic test.

## Performance (HorizonWeeks=52, production-volume data)
- **Order-line zoom** (single product): within target (~1.1 s; the rewrite removes the JIT-misestimate penalty).
- **WebUI standalone first page**, warm: ~110 s → ~3.85 s (28.5×); full materialisation 7.6 s → 3.6 s (~3000× fewer buffer hits).

## Known follow-up (not in this PR)
The standalone first-page **cold cache** is still ~28 s: cold, the planner regresses to a nested loop (re-computing the full view once per selected page row) because the view's computed MD5 PK is non-indexable. The reliable fix is to make the standalone **open empty and load only after a filter** — confirmed feasible via `SqlViewBinding.queryIfNoFilters(false)` (a small Java `@ViewFactory` for window 542159 + an AD_Message); tracked for a follow-up. This PR is the foundation (and fully fixes the zoom + the warm path).

## Tests
- Set-diff harness `0/0` against the old-view oracle (re-run on the migrated real view).
- `MD_Stock_PerWeek` cucumber: validated in CI.


metas-ts (commented):
looks already good, but we will supplement antoher cucumber-test

PR #24615, opened 2026-06-17
HU Manager bulk-move E2E — scan target locator until the move fires

adi-stefan:
The two mobile-webui "Bulk actions - Move" E2E specs flake/fail deterministically when run against the stack.

## Root cause (test-side)
`HUBulkActionsScreen.move()` opens the target scanner, waits for the "Close scanner" caption, then dispatches a single programmatic barcode scan. The keyboard-scanner hook attaches its `window` keydown listener in a `useEffect` that runs **after** the React commit which shows "Close scanner". The instantaneous programmatic scan races ahead of that listener and is dropped, so the move never fires and `ApplicationsListScreen.waitForScreen()` times out at 20s. A real operator scans long after the effect has attached, so this is purely a test-timing artifact — no production behaviour is involved.

## Fix
`move()` now scans the target locator and, if the bulk-actions screen is still open after a short bounded check (`FAST_ACTION_TIMEOUT`), re-scans — up to 5 attempts. This mirrors a real operator who simply scans again when nothing happens. Success is asserted by the `#HUBulkActionsScreen` container detaching (navigation away), not by any fixed delay. Encapsulated in the screen object; no `page.*` in specs; explicit timeout on the wait.

## Scope
Test-side only. No production frontend code changed.

## Verification (local, against running stack)
- RED on pre-fix code: both `tests/spec/humanager/huManager.spec.js` and `tests/spec/humanager/by_ExternalBarcode_attribute.spec.js` "Bulk actions - Move" FAIL at the 20s `waitForScreen` timeout.
- After fix: both specs pass on 3 consecutive runs (8.6s / 13.6s / 13.5s).

### DIFF
--- a/e2e/mobile-webui/tests/utils/screens/huManager/HUBulkActionsScreen.js
+++ b/e2e/mobile-webui/tests/utils/screens/huManager/HUBulkActionsScreen.js
@@ -27,17 +27,13 @@ export const HUBulkActionsScreen = {
         // Wait for button text to change to "Close scanner" - ensures React re-render complete
         await page.getByTestId('toggle-target-scanner-button').getByText('Close scanner').waitFor({ timeout: SLOW_ACTION_TIMEOUT });
 
-        // Scan the target locator until the move fires (the screen navigates away).
-        // The keyboard-scanner hook attaches its window listener in a useEffect that runs
-        // AFTER the React commit that shows "Close scanner", so a single instantaneous scan
-        // can race ahead of the listener and be dropped — the move never fires and the
-        // bulk-actions screen stays open. A real operator who sees nothing happen just scans
-        // again; we do the same: re-scan while the bulk-actions screen is still present.
+        // The keyboard-scanner hook's window listener attaches asynchronously (useEffect after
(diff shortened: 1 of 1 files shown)


mrie04:
@Mergifyio copy task_force_hotfix

mrie04:
@Mergifyio copy task_force_hotfix

mrie04:
**Resolution (dev):** the `@Mergifyio copy task_force_hotfix` commands produced https://github.com/metasfresh/metasfresh/pull/25865, but the auto-copied `e2e/mobile-webui/tests/utils/screens/huManager/HUBulkActionsScreen.js` was committed with unresolved conflict markers (the import line), so every mobile spec importing it failed to load. Resolved on that branch in https://github.com/metasfresh/metasfresh/commit/87f1783cd1dba385e1507c37b22e709037d3e2f4 (union of both sides' timeout imports, nothing else); the two "Bulk actions - Move" specs pass locally with the fix and CI is re-running on https://github.com/metasfresh/metasfresh/pull/25865.

PR #24616, opened 2026-06-17
Distribution mobile launchers: bound the DD_Order selection query

teosarca:
## Problem
The mobile Distribution workflow-launchers WebSocket producer polls roughly once per second and builds its DD_Order selection via `DistributionWorkflowLaunchersProvider.collect` -> `DDOrderService.streamDDOrders` -> `DDOrderLowLevelDAO.streamDDOrders` -> `IQuery.iterateAndStream`.

The guaranteed iterator first runs `createUUIDSelection` = `INSERT INTO T_Query_Selection SELECT ... FROM DD_Order WHERE <active/unassigned>` and only then pages. The launcher limit was applied on the Java stream (`.limit(...)`), after the iterator was built, so the selection INSERT was unbounded and materialized the entire DD_Order backlog on every poll. Sibling providers (`InventoryWorkflowLaunchersProvider`, `PickingWorkflowLaunchersProvider`) push the limit into the query; Distribution did not. On an instance with a large DD_Order backlog this holds DB connections far longer than needed and can exhaust the connection pool.

## Fix (mirror the sibling pattern - push the limit into the query)
- `DDOrderQuery`: add `QueryLimit limit` (default `NO_LIMIT`, backward-compatible).
- `DDOrderLowLevelDAO.toSqlQuery`: `queryBuilder.setLimit(query.getLimit())` so the `T_Query_Selection` INSERT stays bounded.
- `DistributionJobQueries`: both launcher queries set `.limit(query.getSuggestedLimit())` (already defaults to `config.getMaxLaunchers()`).
- New `DDOrderLowLevelDAOStreamLimitTest`.

## Status / caveats
- Not yet locally compiled/tested - local build cache is being repopulated. CI is the first authoritative signal.
- Heap-dump analysis is in progress to confirm, against the actual heap, the precise dominator that filled memory; this PR is the proposed root-cause fix pending that confirmation.


mrie04:
@teosarca is this one still needed?

PR #24617, opened 2026-06-17
AvailableForSales sync: replace in-process Debouncer with RabbitMQ event (fix app-server OOM)

teosarca:
## Problem
A heap dump from a crashed app server showed a single in-process `de.metas.util.async.Debouncer` (the AvailableForSales sync debouncer) holding **~4.77 M** queued `EnqueueAvailableForSalesRequest` objects (~3.9 GB, ~48% of the heap). Two compounding causes:
- The debouncer's consumer was wedged (a worker thread blocked acquiring a JDBC connection from an exhausted pool), so it could not drain while producers kept enqueuing — and the in-memory buffer had no real upper bound.
- Each request retained a full Env `Properties` context (~1.68 GB of `Hashtable$Entry` across the 4.77 M copies).

## Fix — move AvailableForSales sync off-heap onto the event bus
Replace the in-process debouncer with **direct RabbitMQ** publishing via a `DISTRIBUTED` `de.metas.event` topic, so the backlog buffers in the broker (off-heap) instead of the JVM heap — it can no longer OOM the app server even if the consumer is slow/wedged.
- New `EnqueueAvailableForSalesPublisher` / `EnqueueAvailableForSalesHandler` / `EnqueueAvailableForSalesConverter`.
- The event payload carries only typed IDs (`AvailableForSalesQuery` + `UserId`/`RoleId`/`ClientAndOrgId`) — **no more `Properties` context on the wire** (eliminates the 1.68 GB retention). The handler reconstructs a minimal context (client/org + the originating user/role) before recomputing.
- Published after-commit (no phantom event on rollback); recompute is idempotent.
- `AttributesKeyPattern` gains Jackson annotations so the query serializes to JSON for the event payload (round-trip tested).

## Also in this PR
- A generic, **opt-in** load-shed cap on `Debouncer` (`bufferHardLimit`, default = unbounded) + `DebouncerBackpressureTest`, as a backstop for the *other* (non-AFS) debouncer users. It is load-shedding, not backpressure, and is off by default so existing callers are never silently capped.

## Notes
- Related: bounding the mobile Distribution launcher query in https://github.com/metasfresh/metasfresh/pull/24616 — defense-in-depth that reduces the connection-pool pressure which wedged the consumer.


metas-ts (commented):
👍

PR #24624, opened 2026-06-17
mobile-webui: device Back follows screen-declared back, not the browser stack

teosarca:
## Problem

The mobile app navigates with `history.replace` (`useMobileNavigation.goTo`, `useMobileNavigation.js:28`) and every screen declares its own back target via headers (`backLocation`, resolved in `reducers/headers.js:104`), **independent of how the operator arrived**. The browser history stack therefore holds *arrival order*, not the app's logical back graph — so the device / browser **Back** button pops the wrong entry and breaks the flow.

## Change

New hook `useDeviceBackButton` (mounted once inside `ConnectedRouter` as a renderless `DeviceBackButtonTrap`):

- Keeps a throwaway **sentinel** history entry. A `popstate` (Back press) cannot be cancelled, so on Back we re-push the sentinel — the URL / stack never actually moves — and instead navigate to the screen-declared `backLocation`, i.e. the exact target the on-screen footer **Back** button uses.
- **Top of the stack** (no declared back, e.g. launchers/home) → re-prime only = deliberate **no-op**, so a warehouse operator can't accidentally leave the PWA mid-job.
- Re-primes the sentinel on every navigation, because `goTo()` uses `history.replace` which overwrites the previous sentinel.

Mounted once (not per-screen): routes render as sibling `<Route exact>` without a `<Switch>` and can double-mount (see module `CLAUDE.md`), so a single stable mount point is required.

### One deliberate design decision (please confirm)

`useMobileNavigation.goBack()` falls back to `history.go(-1)` (browser-stack back) when a screen has **no** declared `backLocation` (`useMobileNavigation.js:33-34`). This hook intentionally does **NOT** reuse that fallback — `go(-1)` is exactly the browser-stack behavior being neutralized. Net effect: on a no-`backLocation` screen, device Back becomes a no-op (instead of escaping via the stack). The footer Back button's own behavior is unchanged (out of scope here).

## UX (handheld operator baseline)

Makes the hardware Back button behave identically to the on-screen footer Back — predictable, and it can't drop the operator out of the app. This satisfies the module's "Cancel/back always available" baseline better than a frozen/dead button.

## Tests

- `src/__tests__/hooks/useDeviceBackButton.test.js` — TDD (RED first): sentinel primed on mount; Back → re-prime + `goTo(backLocation)`; Back at top → re-prime only, no nav; never `history.go(-1)`.
- Local: `yarn lint` clean; the new suite green (4/4). Full Jest suite has 13 pre-existing failures on the untouched base (picking / manufacturing / qrCode suites) — unrelated to this change (verified by re-running them with this change stashed).

### ⚠️ Runtime acceptance still owed

This unit-tests the hook's contract in jsdom. The real interaction with `ConnectedRouter` + the `history` library (popstate listener ordering, no flicker) should be confirmed with a **Playwright E2E** that drives a flow, presses browser Back, and asserts it lands on the screen-declared back screen — not yet added.

## Process note

This skipped the usual me03 grooming → brainstorming flow (no issue exists). Opened as **draft** for review of the approach + the design decision above.

### DIFF
--- a/misc/services/mobile-webui/mobile-webui-frontend/src/__tests__/hooks/useDeviceBackButton.test.js
+++ b/misc/services/mobile-webui/mobile-webui-frontend/src/__tests__/hooks/useDeviceBackButton.test.js
@@ -1,0 +1,87 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+import { useDeviceBackButton } from '../../hooks/useDeviceBackButton';
+import { useMobileNavigation } from '../../hooks/useMobileNavigation';
+import { useBackLocationFromHeaders } from '../../reducers/headers';
+
+jest.mock('../../hooks/useMobileNavigation', () => ({
+  useMobileNavigation: jest.fn(),
+}));
+jest.mock('../../reducers/headers', () => ({
+  useBackLocationFromHeaders: jest.fn(),
+}));
+
+const goTo = jest.fn();
+
+function renderTrap() {
+  function TestComponent() {
+    useDeviceBackButton();
+    return null;
+  }
+  return render(
+    <MemoryRouter>
+      <TestComponent />
+    </MemoryRouter>
+  );
+}
+
+function firePopState() {
+  window.dispatchEvent(new PopStateEvent('popstate'));
+}
+
+describe('useDeviceBackButton', () => {
+  let pushStateSpy;
+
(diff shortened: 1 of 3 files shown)


teosarca:
Superseded by https://github.com/metasfresh/metasfresh/pull/24625 — re-rooted onto `deep_tundra_release` (the intended base) instead of `new_dawn_uat`.

PR #24637, opened 2026-06-18
test picking without picking slot

teosarca:


teosarca:
**Review** (whole-state, against `deep_tundra_release`) — 3 findings, all addressed in 959517cd16f:

1. **(Important) `allure.story` not timeless** — `e2e/mobile-webui/tests/spec/picking/pickingSlotRequired.spec.js` embedded the internal flag name `IsPickingSlotRequired=N`. Allure stories must describe behavior, not a config-key value. → changed to `Pick without a picking slot when picking slot is not required`.
2. **(Important) `TEST-COVERAGE.md` not updated** — the new spec was unregistered. → added a row under *Order-based picking — special flows* (now `8/8`) and bumped the Picking summary to `59/63`.
3. **(Minor) response-DTO parity** — `JsonMobileConfigResponse.Picking` lacked `pickingSlotRequired` while the request DTO has it, breaking the config round-trip every other flag upholds. → added the field and populated it in `MobileConfigPickingCommand.toJson()` (mirrors `displayPickingSlotSuggestions`).

`allure.feature('F00230: MobileUI Picking')` was evaluated and kept — it matches the documented Allure convention in `e2e/mobile-webui/CLAUDE.md` and the `overPickingPrompt.spec.js` family.

Re-review of the converged HEAD: APPROVE, zero findings. Local compile of `de.metas.frontend-testing` (with dependency chain): BUILD SUCCESS.

PR #24638, opened 2026-06-18
bugfix NPE when processing dd order candidates w/o product planning

teosarca:


teosarca:
**Review** (whole-state, against `deep_tundra_release`):

The NPE fix is correct and regression-safe — the null-guards clear `PP_Product_Planning_ID` / `AD_User_ID` / `SalesRep_ID` with the `-1` sentinel (the established convention), and the happy path is unchanged.

1. **(High) missing regression test** → added `DDOrderCandidateProcessCommandTest`. It processes a `DDOrderCandidate` with `productPlanningId=null` and asserts `execute()` does not throw and the created `DD_Order` has `PP_Product_Planning_ID = AD_User_ID = SalesRep_ID = -1`. Collaborators are mocked at the command seam; the real `DDOrderLowLevelService` is used so the in-memory persistence assigns PKs. Verified **RED** against the pre-fix code (NPE in `createHeaderRecord`) and **GREEN** with the fix; passes locally.

Re-review of the converged HEAD: APPROVE, zero substantive findings (one style nit — static-import `Mockito.never()` — applied).

PR #24640, opened 2026-06-18
Invoice Accounting Overrides (F01010.4)

metas-ts:
## Invoice Accounting Overrides (F01010.4)

Lets users steer the GL account a **purchase** invoice posts to — per invoice and per line — and makes the existing override mechanism safe, discoverable and documented. No change to the posting engine itself: the override is resolved by the existing `InvoiceAccountProviderExtension` reading `C_Invoice_Acct`.

### What's new

**Set an override account up front (candidate / draft line → materialized on completion)**
- New FK column `C_ElementValue_Override_ID` on `C_Invoice_Candidate` and on `C_InvoiceLine`. Editable on the purchase invoice-candidate disposition window and on a draft purchase invoice line; read-only once the invoice is completed.
- The candidate's override account is copied onto the generated invoice line during aggregation (`InvoiceCandBLCreateInvoices`).
- On invoice completion (purchase only), a `BEFORE_COMPLETE` interceptor materializes the proper `C_Invoice_Acct` override rows — one per active acct-schema, for `P_Expense_Acct` and `P_InventoryClearing_Acct` — so posting resolves to the chosen account.
- **Surgical merge**: materialization deactivates only the existing `C_Invoice_Acct` rows it contradicts (same acct-schema + invoice + line + concept, different account) and inserts the new one; manual overrides are never wholesale-deleted.
- A read-only virtual field on the candidate shows the GL account the override would resolve to, before posting.

**Existing manual override window — made safe, discoverable, documented (the view/edit layer)**
- The account-concept scoping field is now a fixed dropdown of the concepts an invoice can post to (was free text); the stored technical name (e.g. `P_Expense_Acct`) is unchanged. Empty concept = applies to all concepts (documented in field help).
- Corrected the two reversed field labels; added German names + help for all user-facing fields; German window + menu name; menu node stays in Finanzen.
- Read-only parent-invoice posting status in the window (surfaces "re-post needed" after editing a posted invoice; edits stay allowed).
- Invoice line flags an overridden account (tooltip + indicator); accounting-facts view marks override-derived fact lines. All read-time derived — no stored state, no posting-logic change.

**Cleanup**
- Drops the unused orphan column `C_InvoiceLine.Account_ID` (no consumer on any branch); generated model classes regenerated.

### Notes
- Purchase-side only for this iteration; no order-side changes; setting the override via REST API is out of scope.
- Payment-allocation accounting is structurally unaffected — `Doc_AllocationHdr` registers no account-provider extension and only requests BPartner accounts, so the override never reaches allocation posting.
- Verified: unit/repository tests green (`InvoiceAcctRepository_SurgicalUpsertTest`, 7/7); whole-branch review clean incl. window render.


backend/de.metas.acct.base/src/main/sql/postgresql/ddl/views/Fact_Acct_Transactions_View.sql
```diff
@@ -83,7 +83,17 @@ SELECT fact.fact_acct_id,
        fact.OI_TrxType,
        fact.IsOpenItemsReconciled,
        fact.OI_OpenAmount,
-       fact.OI_OpenAmountSource
+       fact.OI_OpenAmountSource,
+       CASE WHEN EXISTS (
```
metas-ts, line 87: did you make sure we have the indixed we need to avoid sequential scans?

metas-ts: **Resolution (dev):** Good catch — `C_Invoice_Acct` had only its PK index, so the correlated `EXISTS` in `IsAccountOverridden` would seq-scan it per `Fact_Acct` row. Added a composite index covering the three equality join columns:

```sql
CREATE INDEX IF NOT EXISTS C_Invoice_Acct_Invoice_Schema_Acct_idx
    ON C_Invoice_Acct (C_Invoice_ID, C_AcctSchema_ID, C_ElementValue_ID);
```

Migration `5809060_sys_me03_30443_CInvoiceAcct_index.sql`, commit https://github.com/metasfresh/metasfresh/commit/ceb94db1dca . Now the per-row probe is an index lookup (and since most invoices have zero override rows, the common no-match case is a cheap index probe instead of a full scan). The `C_InvoiceLine_ID = …` OR-NULL predicate is filtered after the index narrows to the invoice's few rows.

backend/de.metas.business/src/main/java/de/metas/invoice/acct/InvoiceAcctRepository.java
```diff
@@ -117,4 +121,60 @@ private void updateRecordFromRuleMatcher(@NonNull final I_C_Invoice_Acct record,
...
+				// Exact match — nothing to do.
+				return;
+			}
+			// Different account → deactivate.
+			row.setIsActive(false);
+			InterfaceWrapperHelper.save(row);
+		}
+
+		// Insert new active row.
+		final I_C_Invoice_Acct newRecord = InterfaceWrapperHelper.newInstance(I_C_Invoice_Acct.class);
```
metas-ts, line 180: what if e.g. there is an existing row which has e.g. no invoiceline-id? 
and thuse matches the whole invoice? are we mitigating this by making the `C_Invoice_Acct` table read-only in the UI and just inserting rows programatically via invoiceline.
Or is the the scenario of a broader-matched `C_Invoice_Acct` ruled out somehow else?

backend/de.metas.business/src/main/java/de/metas/invoice/acct/interceptor/C_Invoice_AcctOverride.java
```diff
@@ -0,0 +1,101 @@
...
+@Interceptor(I_C_Invoice.class)
+@Component
+@RequiredArgsConstructor
+public class C_Invoice_AcctOverride
+{
+	// Mirrors ProductAcctType.P_Expense_Acct/P_InventoryClearing_Acct.getAccountConceptualName() from
+	// de.metas.acct.base. We derive the conceptual names from the I_M_Product_Acct column names (the same
+	// source ProductAcctType uses) to keep the binding to the product-acct columns explicit and rename-safe,
+	// while avoiding a compile dependency on de.metas.acct.base.
+	private static final AccountConceptualName CONCEPT_P_EXPENSE_ACCT = AccountConceptualName.ofString(I_M_Product_Acct.COLUMNNAME_P_Expense_Acct);
```
metas-ts, line 51: why did we choose to override the accounts for these two account-names in particular?

backend/de.metas.business/src/main/java/de/metas/invoice/acct/InvoiceAcctRepository.java
metas-ts: **Resolution (dev):** Confirmed (discussed with the team). An existing invoice-level row (`C_InvoiceLine_ID` empty) matches the whole invoice; a line-specific row takes precedence over it for that line — `InvoiceAcctRuleMatcher.ORDER_FROM_SPECIFIC_TO_GENERIC` orders `invoiceAndLineId` **nullsLast**, so the line-specific match is evaluated first. The on-the-fly materialization (`createOrUpdateLineOverride`) intentionally deactivates only a contradicting **line-specific** row, never a broader invoice-level one (it may legitimately apply to other lines). The window stays editable (not read-only). We'll document this in the UI — in the **`C_Invoice_Acct.C_InvoiceLine_ID` field description** (empty = applies to the whole invoice; a line-specific override takes precedence) — landing with the remaining test/doc work for this issue.

backend/de.metas.business/src/main/java/de/metas/invoice/acct/interceptor/C_Invoice_AcctOverride.java
metas-ts: **Resolution (dev):** As discussed in private, we'll leave it like this for now — these are the two product-derived debit accounts a purchase invoice line posts to, mirroring the approach in the other system.

backend/de.metas.business/src/main/java/de/metas/invoice/acct/InvoiceAcctRepository.java
```diff
@@ -117,4 +121,72 @@ private void updateRecordFromRuleMatcher(@NonNull final I_C_Invoice_Acct record,
...
+	{
+		final InvoiceId invoiceId = invoiceAndLineId.getInvoiceId();
+
+		// Query for the exact (schema, invoice, line, concept) tuple — active rows only.
+		final List<I_C_Invoice_Acct> existing = queryBL
+				.createQueryBuilder(I_C_Invoice_Acct.class)
+				.addOnlyActiveRecordsFilter()
+				.addEqualsFilter(I_C_Invoice_Acct.COLUMNNAME_C_AcctSchema_ID, acctSchemaId)
+				.addEqualsFilter(I_C_Invoice_Acct.COLUMNNAME_C_Invoice_ID, invoiceId)
+				.addEqualsFilter(I_C_Invoice_Acct.COLUMNNAME_C_InvoiceLine_ID, invoiceAndLineId.getRepoId())
```
metas-ts: i think you don't need the `.getRepoId()`

```suggestion
				.addEqualsFilter(I_C_Invoice_Acct.COLUMNNAME_C_InvoiceLine_ID, invoiceAndLineId)
```

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/accounting/C_ElementValue_StepDefData.java
```diff
@@ -0,0 +1,25 @@
...
+public class C_ElementValue_StepDefData extends StepDefData<I_C_ElementValue>
+		implements StepDefDataGetIdAware<ElementValueId, I_C_ElementValue>
+{
+	public C_ElementValue_StepDefData()
+	{
+		super(I_C_ElementValue.class);
+	}
+
+	@Override
+	public ElementValueId extractIdFromRecord(final I_C_ElementValue record)
```
metas-ts: always use the lombok `@NonNull` annotation if feasible

backend/de.metas.business/src/main/java/de/metas/invoice/acct/InvoiceAcctRepository.java
metas-ts: **Resolution (dev):** Done — pass `invoiceAndLineId` directly to `addEqualsFilter` (dropped `.getRepoId()`). `InvoiceAndLineId implements RepoIdAware`, so the query builder normalizes it to the line id — consistent with the sibling `acctSchemaId`/`invoiceId` filters two lines above. Verified: `InvoiceAcctRepository_SurgicalUpsertTest` 7/7. https://github.com/metasfresh/metasfresh/commit/fa73d68f55c

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/accounting/C_ElementValue_StepDefData.java
metas-ts: **Resolution (dev):** Done — added `@NonNull` to the `record` parameter of `extractIdFromRecord`. https://github.com/metasfresh/metasfresh/commit/fa73d68f55c

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/invoice/invoiceAccountingOverride.feature
```diff
@@ -0,0 +1,189 @@
+@from:cucumber
+@allure.label.epic:E0340_Invoicing
+@allure.label.feature:F00700_Invoicing
```
metas-ts, line 3: sooo...how about also adding this issue's feature to this issue's cucumber ;-).
and please make that a review-rule!

metas-ts: **Resolution (dev):** Done — added `@allure.label.feature:F01010.4_Invoice_Accounting_Overrides` (this issue's Feature) alongside the generic `F00700_Invoicing`, on the `Feature` block and both scenarios. https://github.com/metasfresh/metasfresh/commit/c8fe38c05e6 — rule added: a feature file must carry the issue's own Sprint-Dashboard Feature (not only a generic bucket), now enforced as a review-rule.

backend/de.metas.business/src/main/sql/postgresql/system/20-de.metas.business/5808530_sys_InvoiceAcct_field_help_trl.sql
```diff
@@ -0,0 +1,74 @@
...
+--   de_CH element_trl.updated = 2019-12-16 13:09:53
+UPDATE AD_Field_Trl
+SET    Help         = 'Das Buchführungsschema, in dessen Kontenrahmen das überschreibende Konto gesucht wird.',
+       IsTranslated = 'Y',
+       Updated      = TO_TIMESTAMP('2018-11-26 07:57:00.353716', 'YYYY-MM-DD HH24:MI:SS.US'),
+       UpdatedBy    = 100
+WHERE  AD_Field_ID  = 710154
+  AND  AD_Language  = 'de_DE';
+
+UPDATE AD_Field_Trl
```
metas-ts: no point in setting `AD_Field_Trl`. it needs to come from ad_element

metas-ts: **Resolution (dev):** Agreed — removed the direct `AD_Field_Trl` write entirely. Migration `5808530` is deleted; it was redundant. Fields 710152/710153/710154 already get their per-field help from **dedicated `AD_Element`s** (585016/585017/585018) linked via `AD_Field.AD_Name_ID` and propagated by `update_FieldTranslation_From_AD_Name_Element` in `5808540` (which explicitly supersedes `5808530`). So the help now comes from `ad_element` as you said, and propagates durably. Verified on the local DB: field 710153 → `AD_Name_ID=585017`, de_DE/de_CH help equals the element help. https://github.com/metasfresh/metasfresh/commit/3a2430e97f6

PR #24666, opened 2026-06-22
OLCand sales REST API — custom columns (extendedProps) + first-class fields

mrie04:
## What

Extend the **v2 OrderCandidate (OLCand) sales REST API** so external systems can set additional order fields when creating order candidates.

### Generic custom columns (`extendedProps`)
- New `extendedProps` map on the v2 `JsonOLCandCreateRequest` and `JsonOLCand` response, mirroring the BPartner API's `IsRestAPICustomColumn` mechanism.
- Values are persisted onto `C_OLCand` via `CustomColumnService`, then **generically propagated** to `C_Order` / `C_OrderLine` for every `AD_Column` flagged `IsRestAPICustomColumn='Y'` on both the candidate and the target table — no Java change needed per new column.
- Accepted custom columns are echoed back in the create response.

### First-class fields
- `promotionCode` / `promotionCode2` — resolved by `C_PromotionCode.Value` (unknown value → clear validation error).
- `isWithoutCharge`, `reason`.
- New `C_OLCand` columns + model regen; values propagate to `C_Order` (promotion codes) and `C_OrderLine` (`IsWithoutCharge` / `Reason`).

### UI / window placement (justified non-UI columns)
The four new `C_OLCand` columns — `C_PromotionCode_ID`, `C_PromotionCode2_ID`, `IsWithoutCharge`, `Reason` — are intentionally **not** added to any `AD_Window` / `AD_Field` / `AD_UI_Element`. `C_OLCand` is a REST-API staging/transport table: these values are written only through the API and propagate to `C_Order` / `C_OrderLine`, where the equivalent user-facing fields already exist. The order-candidate window is used for import/error diagnosis, not for manually editing these carrier columns. Verified against the DB: each of the four columns has zero `AD_Field` rows.

## Tests
- Unit tests: generic propagation, first-class propagation, promotion-code-by-`Value` resolution.
- Cucumber: end-to-end round-trip (custom + first-class fields + unknown-code negative).

## Scope
v2 OLCand sales endpoint only. Remains in **draft** until CI is green.


backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/persistence/custom_columns/RESTApiTableInfo.java
```diff
@@ -41,4 +45,14 @@ public boolean isCustomRestAPIColumn(@NonNull final String columnName)
...
 
+	/**
+	 * Returns the custom REST API column names in their original canonical case
+	 * (e.g. "MyCustom_Column"), suitable for use as column identifiers when
+	 * reading/writing PO or POJOWrapper records.
+	 */
+	public ImmutableSet<String> getCustomRestAPIColumnNames()
+	{
+		return customRestAPIColumnNames;
+	}
```
mrie04: why this in addition to `@Getter`

backend/de.metas.business/src/main/java/de/metas/promotioncode/PromotionCodeRepository.java
```diff
@@ -0,0 +1,56 @@
...
+{
+	@NonNull
+	public PromotionCodeId getPromotionCodeIdByValue(@NonNull final String value)
+	{
+		final Integer repoId = Services.get(IQueryBL.class)
+				.createQueryBuilderOutOfTrx(I_C_PromotionCode.class)
+				.addOnlyActiveRecordsFilter()
+				.addEqualsFilter(I_C_PromotionCode.COLUMNNAME_Value, value)
+				.create()
+				.firstId();
```
mrie04: pls directly get repoId `org.compiere.model.IQuery#firstIdOnly(java.util.function.Function<java.lang.Integer,ID>)`

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/olcand/C_OLCand_StepDef.java
```diff
@@ -234,6 +235,49 @@ public void process_JsonOLCandCreateBulkResponse(@NonNull final DataTable dataTa
...
+	 * @cucumber.example
+	 * <pre>
+	 * And process metasfresh response JsonOLCandCreateBulkResponse v1
+	 *   | C_OLCand_ID.Identifier |
+	 *   | olcand_1               |
+	 * </pre>
+	 */
+	@And("process metasfresh response JsonOLCandCreateBulkResponse v1")
+	public void process_JsonOLCandCreateBulkResponse_v1(@NonNull final DataTable dataTable) throws JsonProcessingException
+	{
```
mrie04: pls don't add tests for depcrecated v1

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/promotioncode/C_PromotionCode_StepDef.java
mrie04: fqdn

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandCreateRequest.java
```diff
@@ -150,6 +152,16 @@ public class OLCandCreateRequest
 	String email;
 	String phone;
 
+	@Nullable Map<String, Object> extendedProps;
+
+	@Nullable PromotionCodeId promotionCodeId;
+
+	@Nullable PromotionCodeId promotionCode2Id;
+
+	@Nullable Boolean isWithoutCharge;
```
mrie04: why not boolean?

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
```diff
@@ -329,6 +334,19 @@ private I_C_Order newOrder(@NonNull final OLCand candidateOfGroup)
 			order.setIncotermLocation(candidateOfGroup.getIncotermLocation());
 		}
 
+		// First-class field propagation: promotion codes from OLCand to order header
+		final de.metas.ordercandidate.model.I_C_OLCand olCandRecord = candidateOfGroup.unbox();
+		if (olCandRecord.getC_PromotionCode_ID() > 0)
```
mrie04: pls use repoid for such checks

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
```diff
@@ -605,6 +623,15 @@ else if (currentOrderLine.isManualQtyInPriceUOM() != isManualQtyInPriceUOM)
...
 		olCandListeners.onOrderLineCreated(candidate, currentOrderLine);
 
+		// First-class field propagation: IsWithoutCharge and Reason from OLCand to order line
+		{
+			final de.metas.ordercandidate.model.I_C_OLCand olCandRecord = candidate.unbox();
+			currentOrderLine.setIsWithoutCharge(olCandRecord.isWithoutCharge());
+			currentOrderLine.setReason(olCandRecord.getReason());
+
+			copyCustomColumns(olCandRecord, org.compiere.model.I_C_OrderLine.Table_Name, currentOrderLine);
+		}
```
mrie04, line 628: why before listener

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
```diff
@@ -732,6 +759,56 @@ I_C_Order getOrder()
...
+	/**
+	 * Copies every {@code IsRestAPICustomColumn='Y'} column from the given {@code source} (C_OLCand)
+	 * to {@code target} (C_Order or C_OrderLine), if and only if the same column name is also flagged
+	 * as a custom REST API column on the target table.
+	 * <p>
+	 * Values are transferred untyped so that any future custom column is handled automatically
+	 * without a Java change.
+	 */
+	private void copyCustomColumns(
+			@NonNull final de.metas.ordercandidate.model.I_C_OLCand source,
```
mrie04: fqdn needed?

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
```diff
@@ -732,6 +759,56 @@ I_C_Order getOrder()
...
 
+	/**
+	 * Copies every {@code IsRestAPICustomColumn='Y'} column from the given {@code source} (C_OLCand)
+	 * to {@code target} (C_Order or C_OrderLine), if and only if the same column name is also flagged
+	 * as a custom REST API column on the target table.
+	 * <p>
+	 * Values are transferred untyped so that any future custom column is handled automatically
+	 * without a Java change.
+	 */
+	private void copyCustomColumns(
```
mrie04: why this is here and not in an customColumnService? Is this copy of c_bpartner logic? if yes use service in both locations

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandRepository.java
```diff
@@ -351,6 +355,42 @@ private I_C_OLCand createAndSaveOLCandRecord(@NonNull final OLCandCreateRequest
 
 		saveRecord(olCandWithIssuesInterface);
 
+		// wire extendedProps: set custom REST API columns on the saved C_OLCand record
+		if (request.getExtendedProps() != null
```
mrie04: why all this isn't after setAsyncbatch or anywhere else on the top?

backend/de.metas.salescandidate.base/src/test/java/de/metas/ordercandidate/api/OLCandOrderFactoryFirstClassFieldsTest.java
```diff
@@ -0,0 +1,296 @@
...
+		SpringContextHolder.registerJUnitBean(
+				IOLCandBL.class,
+				new OLCandBL(bpartnerBL, BPartnerOrderParamsRepository.newInstanceForUnitTesting())
+		);
+
+		// No custom columns needed for this test (we test first-class typed propagation, not generic)
+		SpringContextHolder.registerJUnitBean(
+				org.adempiere.ad.persistence.custom_columns.CustomColumnRepository.class,
+				new OLCandOrderFactoryCustomColumnPropagationTest.TestCustomColumnRepository(java.util.Collections.emptyMap())
+		);
```
mrie04: consider functions like org.adempiere.warehouse.WarehouseRepository#newInstanceForUnitTesting

```java
@VisibleForTesting
	public static WarehouseRepository newInstanceForUnitTesting()
	{
		Adempiere.assertUnitTestMode();
		//noinspection DataFlowIssue
		return SpringContextHolder.getBeanOrSupply(WarehouseRepository.class, WarehouseRepository::new);
	}
```

pls update corpus if noinspection part is missing

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/ordercandidates/impl/JsonConverters.java
```diff
@@ -351,8 +376,17 @@ public JsonOLCandCreateBulkResponse toJson(
...
 			@NonNull final MasterdataProvider masterdataProvider)
 	{
+		// batch-load promotion-code values up front, so the per-item conversion below avoids an N+1 round-trip
+		final ImmutableSet<PromotionCodeId> promotionCodeIds = olCands.stream()
+				.flatMap(olCand -> Stream.of(
+						PromotionCodeId.ofRepoIdOrNull(olCand.unbox().getC_PromotionCode_ID()),
+						PromotionCodeId.ofRepoIdOrNull(olCand.unbox().getC_PromotionCode2_ID())))
+				.filter(Objects::nonNull)
+				.collect(ImmutableSet.toImmutableSet());
+		final ImmutableMap<PromotionCodeId, String> promotionCodeValuesById = promotionCodeRepository.getValuesByIds(promotionCodeIds);
```
mrie04, line 386: consider moving logic into masterdataProvider and creating an toJsonRequest, as I can see more massloads in the future here, which would mean a lot of params.

backend/de.metas.business/src/main/java/de/metas/promotioncode/PromotionCodeRepository.java
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Repository Tables: C_PromotionCode
+ * Repository Cluster: PromotionCodeRepository
+ */
+@Repository
+public class PromotionCodeRepository
+{
+	@NonNull
+	public PromotionCodeId getPromotionCodeIdByValue(@NonNull final String value)
```
mrie04, line 70: consider ccache

backend/de.metas.business/src/main/java/de/metas/promotioncode/PromotionCodeRepository.java
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	/**
+	 * Batch reverse-lookup: maps each given {@link PromotionCodeId} to its {@code C_PromotionCode.Value}
+	 * in a single query, so callers serializing a list of records avoid an N+1 round-trip.
+	 * Note: filters by PK only (no active filter) — the ids come from stored columns, so a since-deactivated
+	 * code is still echoed to reflect the persisted state. Blank values and missing ids are absent from the map.
+	 */
+	@NonNull
+	public ImmutableMap<PromotionCodeId, String> getValuesByIds(@NonNull final Collection<PromotionCodeId> promotionCodeIds)
```
mrie04, line 115: consider ccache

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/persistence/custom_columns/RESTApiTableInfo.java
mrie04: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/70b01667d75 — removed the manual getter; the `@Value` class generates `getCustomRestAPIColumnNames()` automatically. `@Getter(AccessLevel.NONE)` kept only on the uppercased set.

backend/de.metas.business/src/main/java/de/metas/promotioncode/PromotionCodeRepository.java
mrie04: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/70b01667d75 — `getPromotionCodeIdByValue` now uses `firstIdOnly(PromotionCodeId::ofRepoIdOrNull)`.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/olcand/C_OLCand_StepDef.java
mrie04: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/e3e2ea1502e — moved off deprecated v1 to v2 (v1 additions reverted, v1 cucumber feature deleted; v2 coverage added).

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/promotioncode/C_PromotionCode_StepDef.java
mrie04: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/2fe9895b958 — replaced the inline `java.sql.Timestamp` FQN with an import.

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandCreateRequest.java
mrie04: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/70b01667d75 — `isWithoutCharge` is now primitive `boolean` (converter maps a null request value to false; the YesNo column defaults N).

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
mrie04: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/70b01667d75 — checks now use `PromotionCodeId.ofRepoIdOrNull(...) != null` instead of `> 0`.

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
mrie04: **Resolution (dev):** Already correct — the line-level IsWithoutCharge/Reason copy runs after `olCandListeners.onOrderLineCreated(...)`; no change needed.

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
mrie04: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/70b01667d75 — inline fully-qualified class names replaced with imports.

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
mrie04: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/70b01667d75 — `copyCustomColumns` moved into `CustomColumnService` (public); the factory calls it. No equivalent inline C_BPartner copy logic exists, so this was a move (nothing to unify).

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandRepository.java
mrie04: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/70b01667d75 — the first-class field block moved up with the other field assignments before the save, eliminating the conditional second save.

backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/ordercandidates/impl/JsonConverters.java
mrie04: **Resolution (dev):** Deferred (out of this issue's scope) — the reverse-lookup is already batched (single query, no N+1). Moving the echo into MasterdataProvider with a `toJsonRequest` is a sensible refactor once more mass-loads actually land here; noted as follow-up.

backend/de.metas.business/src/main/java/de/metas/promotioncode/PromotionCodeRepository.java
mrie04: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/42632dca79a — added `CCache<String, Optional<PromotionCodeId>>` keyed by Value (`tableName=C_PromotionCode` for invalidation). A not-found is cached as `Optional.empty()` and still throws `AdempiereException`.

backend/de.metas.business/src/main/java/de/metas/promotioncode/PromotionCodeRepository.java
mrie04: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/42632dca79a — `getValuesByIds` is now cache-aware via `CCache<PromotionCodeId,String>` (table-invalidated): cache hits served directly, only misses batched in a single `addInArrayFilter` query (no N+1).

backend/de.metas.salescandidate.base/src/test/java/de/metas/ordercandidate/api/OLCandOrderFactoryFirstClassFieldsTest.java
mrie04: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/cea6a811d6f — added `CustomColumnService.newInstanceForUnitTesting()` (mirrors `WarehouseRepository`: asserts unit-test mode, registers the service + a no-op repo as JUnit beans; plus an overload taking a custom repo for the propagation test) and switched all 3 OLCand-factory tests to it, so a future dependency change no longer breaks each test 1-by-1.

mrie04: **Resolution (dev):** rule added — the `newInstanceForUnitTesting()` / `getBeanOrSupply(<Self>.class, () -> new <Self>(...))` convention (including the `//noinspection DataFlowIssue`, a multi-collaborator example, an "add-a-collaborator ⇒ update-the-factory" review rule, and the requirement that the repo be POJO-safe so the plain-`new` factory runs in unit tests) is now codified. The implementation was reworked to follow it: repo + service both self-register via `getBeanOrSupply`, the parameterized overload was dropped, and `CustomColumnRepository` was moved off raw `DB.forEachRow` SQL to `IQueryBL` — which also let the propagation test use real `AD_Column` entries instead of a stub.

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
mrie04: correction why not before listener

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/persistence/custom_columns/RESTApiTableInfo.java
```diff
@@ -13,6 +13,9 @@
 public class RESTApiTableInfo
 {
 	@NonNull String tableName;
+	/** Original (canonical-case) column names as supplied, e.g. "MyCustom_Column". */
+	@NonNull ImmutableSet<String> customRestAPIColumnNames;
+	/** Upper-cased for fast case-insensitive membership checks. */
```
mrie04, line 18: why is this needed in addition to UC

backend/de.metas.business/src/main/java/de/metas/promotioncode/PromotionCodeRepository.java
```diff
@@ -0,0 +1,158 @@
...
+	 * Invalidated automatically on any C_PromotionCode table change.
+	 */
+	private final CCache<PromotionCodeId, String> cacheById = CCache.<PromotionCodeId, String>builder()
+			.tableName(I_C_PromotionCode.Table_Name)
+			.build();
+
+	@NonNull
+	public PromotionCodeId getPromotionCodeIdByValue(@NonNull final String value)
+	{
+		final PromotionCodeId id = cacheByValue.getOrLoad(value, () -> lookupIdByValue(value)).orElse(null);
```
mrie04: getOrLoadNonNull

mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/eed321b103026c2fa5bb8839accf3a7543951231 — switched to `getOrLoadNonNull`. `lookupIdByValue` returns a `@NonNull Optional` (never null; `Optional.empty()` for not-found), so `getOrLoadNonNull` asserts that loader invariant; behaviour and the negative-caching of empty results are unchanged.

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/persistence/custom_columns/RESTApiTableInfo.java
mrie04: **Resolution (dev):** Both sets are intentional and serve distinct purposes. `customRestAPIColumnNames` (canonical case, has a getter) is iterated in `CustomColumnService` to perform the actual column propagation — it needs the real column names. `customRestAPIColumnNamesUC` (getter suppressed) is a derived index used only for the O(1) case-insensitive membership check in `isCustomRestAPIColumn`. Dropping UC would force per-check upper-casing; dropping the canonical set would lose the casing needed for propagation. Happy to collapse to a single set + on-the-fly upper-casing if you prefer the simpler form.

backend/de.metas.business.rest-api-impl/src/test/java/de/metas/rest_api/v1/ordercandidates/impl/OLCandPromotionCodeResolutionTest.java
```diff
@@ -0,0 +1,82 @@
...
+import static org.adempiere.model.InterfaceWrapperHelper.saveRecord;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/**
+ * Tests that the promotion-code lookup-by-Value used by the v1 OLCand REST converter works correctly.
+ * Covers the path: JSON "promotionCode" String → PromotionCodeRepository → PromotionCodeId.
+ */
+@ExtendWith(AdempiereTestWatcher.class)
+class OLCandPromotionCodeResolutionTest
```
mrie04: why this isn't a repo test, don't see olcand specifics

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandRepository.java
```diff
@@ -80,6 +83,7 @@ public class OLCandRepository
     private final IInputDataSourceDAO inputDataSourceDAO = Services.get(IInputDataSourceDAO.class);
 
 	private final ExternalSystemRepository externalSystemRepository;
+	private final CustomColumnService customColumnService;
```
mrie04: pls ad nonull on all fields

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/olcand/C_OLCand_StepDef.java
mrie04: why this step still exists?

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/automateOrderCandToOrderAndShipmentAndInvoice.feature
```diff
@@ -15,6 +15,10 @@ Feature: Process order candidate and automatically generate shipment and invoice
     And preexisting test data is put into tableData
       | C_BPartner_ID.Identifier | C_BPartner_ID | C_BPartner_Location_ID.Identifier | C_BPartner_Location_ID | M_Product_ID.Identifier | M_Product_ID |
       | bpartner_1               | 2156425       | bpartnerLocation_1                | 2205175                | product_1               | 2005577      |
+    And update AD_Column:
```
mrie04: still needed?

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
```diff
@@ -329,6 +332,19 @@ private I_C_Order newOrder(@NonNull final OLCand candidateOfGroup)
 			order.setIncotermLocation(candidateOfGroup.getIncotermLocation());
 		}
 
+		// First-class field propagation: promotion codes from OLCand to order header
+		final I_C_OLCand olCandRecord = candidateOfGroup.unbox();
```
mrie04: why not extend OlCand instead

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
```diff
@@ -605,6 +621,15 @@ else if (currentOrderLine.isManualQtyInPriceUOM() != isManualQtyInPriceUOM)
 		// Fire listeners
 		olCandListeners.onOrderLineCreated(candidate, currentOrderLine);
 
+		// First-class field propagation: IsWithoutCharge and Reason from OLCand to order line
+		{
+			final I_C_OLCand olCandRecord = candidate.unbox();
```
mrie04: why not extend OlCand?

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandRepository.java
```diff
@@ -349,8 +353,34 @@ private I_C_OLCand createAndSaveOLCandRecord(@NonNull final OLCandCreateRequest
...
+		olCandPO.setIsWithoutCharge(request.isWithoutCharge());
+		if (request.getReason() != null)
+		{
+			olCandPO.setReason(request.getReason());
+		}
+
 		saveRecord(olCandWithIssuesInterface);
 
+		// wire extendedProps: set custom REST API columns on the saved C_OLCand record
+		if (request.getExtendedProps() != null
```
mrie04: why not default empty list?

backend/de.metas.salescandidate.base/src/test/java/de/metas/ordercandidate/api/OLCandOrderFactoryCustomColumnPropagationTest.java
```diff
@@ -0,0 +1,319 @@
...
+		return uom;
+	}
+
+	private I_ExternalSystem createExternalSystem()
+	{
+		final I_ExternalSystem externalSystem = newInstanceOutOfTrx(I_ExternalSystem.class);
+		externalSystem.setValue("test");
+		externalSystem.setName("test");
+		saveRecord(externalSystem);
+		return externalSystem;
```
mrie04: why not return repoid?

backend/de.metas.salescandidate.base/src/test/java/de/metas/ordercandidate/api/OLCandOrderFactoryCustomColumnPropagationTest.java
```diff
@@ -0,0 +1,319 @@
...
+				Optional.empty()
+		));
+		SpringContextHolder.registerJUnitBean(new OLCandValidatorService(
+				new OLCandSPIRegistry(Optional.empty(), Optional.empty(), Optional.empty())));
+
+		final BPartnerBL bpartnerBL = new BPartnerBL(new UserRepository());
+		SpringContextHolder.registerJUnitBean(
+				IOLCandBL.class,
+				new OLCandBL(bpartnerBL, BPartnerOrderParamsRepository.newInstanceForUnitTesting())
+		);
```
mrie04: why not introduce proper newInstanceForJUnitTesting? skip BPartnerBL if to much work

backend/de.metas.salescandidate.base/src/test/java/de/metas/ordercandidate/api/OLCandOrderFactoryFirstClassFieldsTest.java
```diff
@@ -0,0 +1,293 @@
...
+				new OLCandSPIRegistry(Optional.empty(), Optional.empty(), Optional.empty())));
+
+		final BPartnerBL bpartnerBL = new BPartnerBL(new UserRepository());
+		SpringContextHolder.registerJUnitBean(
+				IOLCandBL.class,
+				new OLCandBL(bpartnerBL, BPartnerOrderParamsRepository.newInstanceForUnitTesting())
+		);
+
+		// No custom columns needed for this test (we test first-class typed propagation, not generic)
+		org.adempiere.ad.persistence.custom_columns.CustomColumnService.newInstanceForUnitTesting();
```
mrie04: fqdn

backend/de.metas.salescandidate.base/src/test/java/de/metas/ordercandidate/api/OLCandOrderFactoryFirstClassFieldsTest.java
```diff
@@ -0,0 +1,293 @@
...
+				Optional.empty()
+		));
+		SpringContextHolder.registerJUnitBean(new OLCandValidatorService(
+				new OLCandSPIRegistry(Optional.empty(), Optional.empty(), Optional.empty())));
+
+		final BPartnerBL bpartnerBL = new BPartnerBL(new UserRepository());
+		SpringContextHolder.registerJUnitBean(
+				IOLCandBL.class,
+				new OLCandBL(bpartnerBL, BPartnerOrderParamsRepository.newInstanceForUnitTesting())
+		);
```
mrie04: why not newInstanceForJunitTesting? skip bpartnerBL if toi much work

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/0cf9ea47705631ae965fe55c6cd90920defcd67a — added `promotionCodeId`/`promotionCode2Id` as first-class `@Getter` fields on the `OLCand` domain object (populated in `OLCandFactory`); `OLCandOrderFactory` now reads `candidateOfGroup.getPromotionCodeId()` instead of `unbox()`. `unbox()` is kept only for the generic `copyCustomColumns` call (which needs the raw record).

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/0cf9ea47705631ae965fe55c6cd90920defcd67a — `isWithoutCharge`/`reason` are now first-class fields on the `OLCand` domain object; the order-line block reads `candidate.isWithoutCharge()`/`candidate.getReason()` instead of `unbox()`.

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/0cf9ea47705631ae965fe55c6cd90920defcd67a — moved the line-level first-class + custom-column propagation to BEFORE `olCandListeners.onOrderLineCreated(...)`, so listeners observe a fully-populated order line.

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandRepository.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/0cf9ea47705631ae965fe55c6cd90920defcd67a — added `@NonNull` to the constructor-injected collaborators (`externalSystemRepository`, `customColumnService`).

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandRepository.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/0cf9ea47705631ae965fe55c6cd90920defcd67a — `OLCandCreateRequest.extendedProps` is now `@NonNull`, defaulting to `ImmutableMap.of()` in the constructor; the null check here is dropped accordingly.

backend/de.metas.business.rest-api-impl/src/test/java/de/metas/rest_api/v1/ordercandidates/impl/OLCandPromotionCodeResolutionTest.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/0cf9ea47705631ae965fe55c6cd90920defcd67a — moved to `de.metas.promotioncode.PromotionCodeRepositoryTest` (de.metas.business); it has no OLCand specifics, it only exercises `PromotionCodeRepository.getPromotionCodeIdByValue`. Old `OLCandPromotionCodeResolutionTest` deleted.

backend/de.metas.salescandidate.base/src/test/java/de/metas/ordercandidate/api/OLCandOrderFactoryCustomColumnPropagationTest.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/0cf9ea47705631ae965fe55c6cd90920defcd67a — added `OLCandBL.newInstanceForUnitTesting()` (wires `BPartnerBL` + `BPartnerOrderParamsRepository.newInstanceForUnitTesting()`); the test now uses it instead of inline-wiring.

backend/de.metas.salescandidate.base/src/test/java/de/metas/ordercandidate/api/OLCandOrderFactoryFirstClassFieldsTest.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/0cf9ea47705631ae965fe55c6cd90920defcd67a — same as the other test: now uses `OLCandBL.newInstanceForUnitTesting()`.

backend/de.metas.salescandidate.base/src/test/java/de/metas/ordercandidate/api/OLCandOrderFactoryCustomColumnPropagationTest.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/0cf9ea47705631ae965fe55c6cd90920defcd67a — `createExternalSystem()` now returns `ExternalSystemId` (the repoId) instead of the `I_ExternalSystem` record.

backend/de.metas.salescandidate.base/src/test/java/de/metas/ordercandidate/api/OLCandOrderFactoryFirstClassFieldsTest.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/0cf9ea47705631ae965fe55c6cd90920defcd67a — imported `CustomColumnService` and call `CustomColumnService.newInstanceForUnitTesting()` instead of the inline fully-qualified name.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/olcand/C_OLCand_StepDef.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/0cf9ea47705631ae965fe55c6cd90920defcd67a — removed the `process metasfresh response JsonOLCandCreateBulkResponse v1` step; it was a v1-only helper and is no longer referenced by any scenario (the feature uses v2).

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/automateOrderCandToOrderAndShipmentAndInvoice.feature
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/0cf9ea47705631ae965fe55c6cd90920defcd67a — removed; that `update AD_Column` block set up an inert core column for the removed S0469_210 scenario and is no longer needed.

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandRepository.java
```diff
@@ -349,8 +353,33 @@ private I_C_OLCand createAndSaveOLCandRecord(@NonNull final OLCandCreateRequest
...
 
+		// wire extendedProps: set custom REST API columns on the saved C_OLCand record
+		if (!request.getExtendedProps().isEmpty()
+				&& !POJOWrapper.isHandled(olCandWithIssuesInterface))
+		{
+			customColumnService.setCustomColumns(
+					InterfaceWrapperHelper.getPO(olCandWithIssuesInterface),
+					request.getExtendedProps());
+			saveRecord(olCandWithIssuesInterface);
+		}
```
mrie04, line 378: why not before save above?

backend/de.metas.salescandidate.base/src/test/java/de/metas/ordercandidate/api/OLCandOrderFactoryCustomColumnPropagationTest.java
mrie04: but why you still register here too, should already be done inside already, pls check and capture learning

backend/de.metas.salescandidate.base/src/test/java/de/metas/ordercandidate/api/OLCandOrderFactoryFirstClassFieldsTest.java
mrie04: same here double register?

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandRepository.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/dffb738e45ad64573f676459228cc09e302ceb38 — `setCustomColumns` only sets values on the PO (no saved-id dependency), so the extendedProps block now runs BEFORE the initial `saveRecord`, like the first-class fields above — a single write, the trailing `saveRecord` is removed.

backend/de.metas.salescandidate.base/src/test/java/de/metas/ordercandidate/api/OLCandOrderFactoryCustomColumnPropagationTest.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/dffb738e45ad64573f676459228cc09e302ceb38 — fixed the double registration: `OLCandBL.newInstanceForUnitTesting()` now registers under `IOLCandBL` (the key consumers resolve via `Services.get(IOLCandBL.class)`), so the test just calls it once — the explicit `registerJUnitBean(IOLCandBL.class, ...)` wrapper is gone. (Previously the factory registered the concrete `OLCandBL.class` while the test separately registered the interface.)

backend/de.metas.salescandidate.base/src/test/java/de/metas/ordercandidate/api/OLCandOrderFactoryFirstClassFieldsTest.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/dffb738e45ad64573f676459228cc09e302ceb38 — same fix as the other test: `OLCandBL.newInstanceForUnitTesting()` self-registers under `IOLCandBL`; the test calls it once, no separate registerJUnitBean.

PR #24667, opened 2026-06-22
EDI/EPCIS: per-crate order references (order-pure packing) + scripted-export eligibility timing fix

metas-ts:
Two related fixes in the EDI/EPCIS export path.

## 1) EPCIS per-crate source-order references (order-pure packing)
`get_epcis_events_json_fn` exposed order references only at the top level, so when two orders share one physical pallet/SSCC the downstream EPCIS transform emitted every order's `po`+`desadv` on every crate's PACKING event. A crate (Gebinde) is order-pure, so the function now resolves each crate's source order — per TU via `m_hu_assignment → m_inoutline → c_order / m_inout` (CTE `tu_order_ref`, bounded to the already-discovered TUs) — and adds `poReference` (order POReference) and `shipmentDocumentNo` (crate-shipment `M_InOut.DocumentNo` = delivery note) to each crate JSON object. Picking/commissioning stay pallet-level.

Shipped as migration `5809070_sys_..._epcis_per_crate_order_ref.sql` (CREATE OR REPLACE; DDL + migration kept in sync).

## 2) Scripted-export eligibility timing fix
`ExternalSystemScriptedExportConversionInterceptor` fires `docValidate` at `AFTER_COMPLETE`, but `MInOut.completeIt()` fires that timing **before** `DocumentEngine` sets+saves `DocStatus='CO'`. So a complete-time eligibility WhereClause that reads committed `DocStatus` (e.g. the EPCIS `"de.metas.edi".epcis_has_events(m_inout_id)` clause, which requires `DocStatus IN ('CO','CL')`) read the pre-complete row, was recorded `DontSend`, and the export never fired. `recordCompleteTimeEligibilityAndScheduleInvocation` now defers its whole body (match → status-write → invoke) to `trxManager.runAfterCommit(...)`, where `DocStatus` is committed and the clause evaluates correctly. The invocation was already after-commit; this moves the gating decision to the same point.

## Verification
- **EPCIS** (`epcisExport.feature`, provided-infra): **5/5 green** — `S29231_170` asserts a shared-pallet owner event's 15 crates are order-pure across two orders; `EPCIS_010/020/030` + `S29231_130` guard the existing JSON against regression.
- **Eligibility timing**: new RED test `S30279_040` in `epcisScriptedExportStatus.feature` (a config whose WhereClause only matches once committed-complete must reach `Enqueued`, not `DontSend`) + an optional `WhereClause` column on the scripted-export-config step def.

Pairs with the customer-side EPCIS FT scripted-adapter change that consumes the new per-crate fields.

### DIFF
--- a/backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/scriptedexportconversion/ExternalSystemScriptedExportConversionService.java
+++ b/backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/scriptedexportconversion/ExternalSystemScriptedExportConversionService.java
@@ -336,36 +336,23 @@ public class ExternalSystemScriptedExportConversionService
 	}
 
 	/**
-	 * Called from the document's AFTER_COMPLETE interceptor: records an eligibility status for EVERY
-	 * trigger-on-complete config of the record's table and invokes the export for matching configs.
+	 * Records eligibility (matching WhereClause → {@link de.metas.externalsystem.ExternalSystemExportStatus#Pending},
+	 * else {@link de.metas.externalsystem.ExternalSystemExportStatus#DontSend}) for every
+	 * trigger-on-complete config of the record's table, then invokes the export for the matching ones.
 	 *
-	 * <ul>
-	 *   <li>WhereClause MATCHES → {@link de.metas.externalsystem.ExternalSystemExportStatus#Pending}
-	 *       + invokes the export.</li>
-	 *   <li>WhereClause does NOT match → {@link de.metas.externalsystem.ExternalSystemExportStatus#DontSend}
-	 *       (EDI-consistency: DontSend is always persisted).</li>
-	 * </ul>
-	 *
-	 * <p><b>Runs entirely after-commit.</b> The eligibility WhereClause may read COMMITTED document
-	 * state — e.g. the EPCIS {@code "de.metas.edi".epcis_has_events(m_inout_id)} clause requires
-	 * {@code M_InOut.DocStatus IN ('CO','CL')}. At AFTER_COMPLETE the document engine has NOT yet
-	 * set+saved {@code DocStatus='CO'} ({@code MInOut.completeIt} fires AFTER_COMPLETE <i>before</i>
-	 * {@code DocumentEngine} assigns the status), so matching in the completing transaction would read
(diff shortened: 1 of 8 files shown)


backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/scriptedexportconversion/ExternalSystemScriptedExportConversionService.java
```diff
@@ -336,23 +336,38 @@ public void recordPendingForConfig(
...
+	 * set+saved {@code DocStatus='CO'} ({@code MInOut.completeIt} fires AFTER_COMPLETE <i>before</i>
+	 * {@code DocumentEngine} assigns the status), so matching in the completing transaction would read
+	 * the pre-complete row and wrongly record DontSend — the export would never fire. Deferring the
+	 * whole match → status-write → invoke step to after-commit makes the clause evaluate against the
+	 * committed state. (The invocation itself was already after-commit; this moves the decision that
+	 * gates it to the same point, removing the in-trx-vs-committed mismatch.)
+	 *
+	 * <p>A status-write failure for one config is logged and swallowed so completion is never aborted.
 	 */
 	public void recordCompleteTimeEligibilityAndScheduleInvocation(
```
metas-ts: i believe with a fixed methodname you can save 50% or more of the javadoc

metas-ts: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/pull/24667/commits/bceb4e53d07.

Renamed `recordCompleteTimeEligibilityAndScheduleInvocation` → `recordEligibilityAndInvokeAfterCommit` (and the private worker → `recordEligibilityAndInvoke`) — the old name was a misnomer now that the whole step runs after-commit and nothing is merely "scheduled". The self-documenting name let the Javadoc shrink ~60%: it now keeps only the *why* (the eligibility WhereClause reads committed `DocStatus`, which the AFTER_COMPLETE interceptor fires before the engine saves) plus the swallow-on-failure note. The interceptor caller and its stale "runs in the current transaction" comment were updated too. No behaviour change.

PR #24677, opened 2026-06-22
Make Geschaeftspartner optional in BusinessPartnerAccountSheetReport (all-partners export)

cristinamghita:
## Summary

- `BusinessPartnerAccountSheetReport`: `p_c_bpartner_id` now defaults to `NULL`; passing `NULL` returns all business partners with invoices/payments in the date range
- New output columns `c_bpartner_id` and `bpartnerName` identify the partner per row in multi-partner mode
- Per-partner opening balance: `getBPOpenAmtToDate` is called once per distinct partner in the result set
- Rolling sum uses `PARTITION BY c_bpartner_id` so each partner's balance starts from their own opening balance
- `AD_Process_Para.IsMandatory` set to `'N'` for C_BPartner_ID parameter (ID=541752)
- `AD_Process.SQLStatement` updated to use `NULLIF(@C_BPartner_ID/0@, 0)` so leaving the field blank passes `NULL` to the function (all-partners mode)

## Test plan

- [ ] Run with a specific partner: output matches previous behavior (single-partner statement, two new leading columns c_bpartner_id / bpartnerName)
- [ ] Run with partner left empty: Excel contains rows for all partners with invoices/payments in the date range, each with correct per-partner opening/ending balance
- [ ] Verify `bpartnerName` and `c_bpartner_id` columns appear in the exported Excel
- [ ] Verify per-partner rolling balance is correct (ending balance of each partner's last row matches manual check)

### DIFF
--- a/backend/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/BusinessPartnerAccountSheetReport.sql
+++ b/backend/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/BusinessPartnerAccountSheetReport.sql
@@ -1,9 +1,9 @@
 DROP FUNCTION IF EXISTS BusinessPartnerAccountSheetReport(p_c_bpartner_id numeric, p_dateFrom date, p_dateTo date, p_ad_client_id numeric, p_ad_org_id numeric, p_isSoTrx TEXT, p_ad_language text);
 
-CREATE OR REPLACE FUNCTION BusinessPartnerAccountSheetReport(p_c_bpartner_id numeric = NULL,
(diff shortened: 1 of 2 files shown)


cristinamghita:
Closing in favour of a customer-schema variant. The all-partners export and Belegart filter are implemented as `de_metas_billcare.BC_BPAccountSheetReport` directly in the bc107 customer repo, which keeps the base function unchanged and avoids shipping unused balance columns. The ESR/QR Referenznummer column (PR https://github.com/metasfresh/metasfresh/pull/24673) is already merged.

PR #24679, opened 2026-06-22
MD_Stock reset: idempotent set-to-truth + single-instance execution guard

metas-ts:
## What

Makes the `MD_Stock_Update_From_M_HUs` reset **idempotent** and prevents overlapping runs via a generalized single-instance-execution mechanism.

- The reset now SETS `MD_Stock.QtyOnHand` to the view's absolute HU-derived truth (`MD_Stock_From_HUs_V.QtyOnHand`, already in the product stocking UOM) via a new `StockDataUpdateRequestHandler.handleResetToQtyOnHand`, instead of building a `(truth − MD_Stock)` delta from a view snapshot and ADDing it through the shared transaction-event handler.
- New generalized `AD_Process.IsPreventConcurrentExecution` flag: any process can opt into single-instance execution. `ProcessExecutor` acquires an exclusive lock before running a flagged process and skips the run if another instance already holds it (released afterwards). `base` defines the `IProcessSingleInstanceLockService` SPI (NOOP fallback); `de.metas.async` implements it as a row-based `ILock` on the process's `AD_Process` record (pool-agnostic, cross app/webapi, crash-safe via autoCleanup). A migration sets `IsPreventConcurrentExecution=Y` on `MD_Stock_Update_From_M_HUs`.
- Removes the now-dead delta-building + (identity) UOM re-conversion in the reset path.
- The shared add-delta path used by the transaction-event flow is unchanged.

## Why

On a customer PROD instance the scheduler ran this `@RunOutOfTrx` reset with heavy overlap — confirmed: **five concurrent runs**, several lasting hours. Because each run applied a stale view-snapshot delta onto the freshly-read record, overlapping runs compounded corrections on top of each other's committed writes, escalating `MD_Stock.QtyOnHand` to non-physical 1000+ digit values → a flood of `StockChangedEvent`s → app-server OOM.

Setting an **absolute** target is idempotent: overlapping runs all converge to the same truth instead of compounding (a single clean run already converges in one pass). The single-instance-execution flag prevents overlapping runs entirely, so no future regression can escalate to OOM again.

Pairs with the already-merged off-heap fix for the event backlog (https://github.com/metasfresh/metasfresh/pull/24617, ported via https://github.com/metasfresh/metasfresh/pull/24670).

## Tests

- `ProcessSingleInstanceLockServiceTest` — a concurrent acquire is rejected while the first holds the lock and succeeds again after release; distinct `AD_Process` records lock independently.
- `StockDataUpdateRequestHandlerTest` — reset converges from an escalated value; reset is idempotent under repeated application.
- Existing `MD_Stock_Update_From_M_HUsTest` (drain-loop contract) stays green.


backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/stock/StockQtySanityGuard.java
```diff
@@ -0,0 +1,48 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+@UtilityClass
+public class StockQtySanityGuard
```
metas-ts: please remove the StockQtySanityGuard. on 2nd thought i think it just adds complexity and FUD. the other measures are sufficient.

backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/stock/process/MD_Stock_Update_From_M_HUs.java
```diff
@@ -207,29 +267,48 @@ private void createAndHandleDataUpdateRequests(
...
-		final Quantity qtyInStorageUOM = Quantitys.of(huBasedDataRecord.getQtyOnHandChange(), UomId.ofRepoId(huBasedDataRecord.getC_UOM_ID()));
-		final Quantity qtyInProductUOM = uomConversionBL.convertToProductUOM(qtyInStorageUOM, productId);
-
-		return StockDataUpdateRequest.builder()
-				.identifier(recordIdentifier)
-				.onHandQtyChange(qtyInProductUOM.toBigDecimal())
-				.sourceInfo(stockDataUpdateRequestSourceInfo)
-				.build();
+		final ILockManager lockManager = Services.get(ILockManager.class);
+		final I_AD_Process processRecord = InterfaceWrapperHelper.load(getProcessInfo().getAdProcessId().getRepoId(), I_AD_Process.class);
```
metas-ts: this feature should be available to every AD_Process. Please make it a column&field in the AD_Process window, and move this code into the process-engine

backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/stock/StockQtySanityGuard.java
metas-ts: Will do — removing `StockQtySanityGuard` entirely (class + test + the wiring in `StockDataUpdateRequestHandler`). The idempotent set-to-truth reset plus the single-run guard cover it.

backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/stock/process/MD_Stock_Update_From_M_HUs.java
metas-ts: Agreed — generalizing this. I'll add a boolean flag on `AD_Process` (+ field in the AD_Process window) and move the single-instance locking into the process engine, so any process can opt in by setting the flag. This process will then just carry the flag (via migration) and drop the bespoke gate code.

backend/de.metas.adempiere.adempiere/base/src/main/java-gen/org/compiere/model/X_AD_Process.java
```diff
@@ -359,6 +410,30 @@ public boolean isOneInstanceOnly()
...
 	}
 
+	@Override
+	public void setIsPdfA3Output (final boolean IsPdfA3Output)
+	{
+		set_Value (COLUMNNAME_IsPdfA3Output, IsPdfA3Output);
+	}
+
+	@Override
+	public boolean isPdfA3Output()
```
metas-ts: i don't think this really belongs into this branch.
i'd rather assume that you worked with a deep_tundra_release DB when you made this.

backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/stock/StockQtySanityGuard.java
teosarca: **Resolution (dev):** Done — `StockQtySanityGuard` (and its test) removed in https://github.com/metasfresh/metasfresh/commit/75731be98e572f36b5d6797b9ded6224eaf063a9 ("Generalize single-instance process protection into the engine; drop StockQtySanityGuard"). Agreed: the idempotent set-to-truth reset + the single-run gate are sufficient, so the extra guard only added complexity.

backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/stock/process/MD_Stock_Update_From_M_HUs.java
teosarca: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/75731be98e572f36b5d6797b9ded6224eaf063a9. The single-run gate is generalized into the process engine and now available to every AD_Process: new `AD_Process.IsPreventConcurrentExecution` column + field (migration `5809350_sys_AD_Process_IsPreventConcurrentExecution.sql`), enforced centrally via `IProcessSingleInstanceLockService` / `ProcessSingleInstanceLockService` wired into `ProcessExecutor`. The bespoke gate inside `MD_Stock_Update_From_M_HUs` is removed; it just declares the flag.

backend/de.metas.adempiere.adempiere/base/src/main/java-gen/org/compiere/model/X_AD_Process.java
teosarca: **Resolution (dev):** Confirmed — stray, does not belong here. `IsPdfA3Output` is absent on base `intensive_care_hotfix` (0 occurrences) and present only on this branch (3, all in `X_AD_Process.java`), with no matching migration in the branch — so, as you suspected, it leaked into the generated model from a non-intensive_care DB during regeneration. Fix: regenerate `X_AD_Process` against a clean intensive_care DB so only the intended `IsPreventConcurrentExecution` addition remains, dropping the `IsPdfA3Output` hunk. Will push that regen and update this thread.

PR #24693, opened 2026-06-23
MobileUI E2E: assert camera-mode-active via panel wrapper, not transient <video>

teosarca:
## What

Harden the camera-toggle E2E spec (`barcode_scanner_modes.spec.js`) against an intermittent `mobile test` failure that is currently flaking CI on this branch.

The spec asserted `toHaveCount(1)` on `.barcode-scanner video`. But `CameraModePanel` renders the `<video>` only `{!isProcessing && <video/>}`, so it unmounts whenever a barcode is processed. In CI the synthetic fake-media stream (`--use-fake-device-for-media-stream`) can make ZXing false-positive a 1-D barcode → `isProcessing` flips → the `<video>` unmounts mid-assertion (and a `getUserMedia` error tears the panel down via `onCancel`). With `retries: 2`, this loses all attempts in unlucky runs → flaky red `mobile test`.

## Change

- Assert the always-mounted wrapper `.camera-mode-panel` (rendered iff `activeMode === CAMERA`) — the race-free signal that the toggle switched into camera mode — instead of the transient `<video>`.
- Rename `expectCameraVideoPresent` / `expectCameraVideoAbsent` → `expectCameraModeActive` / `expectCameraModeInactive` and update call sites + the spec title/comments + the coverage note.

Test-only; no app behaviour change. Live video-feed rendering still cannot be validated in CI (needs physical camera hardware) — unchanged, documented in the assertion comment.

Targets `deep_tundra_release` (the branch where this flake is actively blocking CI); to be merged up to `new_dawn_uat` afterwards.

### DIFF
--- a/e2e/mobile-webui/TEST-COVERAGE.md
+++ b/e2e/mobile-webui/TEST-COVERAGE.md
@@ -68,7 +68,7 @@
 | ❌ Mode B — Camera (ZXing/BrowserMultiFormatReader): getUserMedia() decode → barcode forwarded | — (not testable in CI, requires real camera) |
 | ❌ `scanDuplicatesIntervalMillis` — duplicate barcode within interval suppressed, outside interval forwarded | — |
 | ❌ `triggerOnChangeIfLengthGreaterThan` — onChange fires only once input length exceeds threshold | — |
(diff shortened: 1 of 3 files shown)


teosarca:
@mergifyio copy soft_panda_hotfix intensive_care_hotfix

PR #24695, opened 2026-06-23
Lot number: generic DB-function sequence provider + S_Resource.LotNumberCode

metas-ts:
## What

Adds a **generic, opt-in mechanism to produce custom lot numbers** at manufacturing receipt via metasfresh's existing pluggable sequence hook — with **zero behaviour change** for any sequence that is not explicitly configured to use it.

### Changes
- **`DBFunctionSequenceNoProvider`** (`de.metas.document.sequenceno`) — a generic `CustomSequenceNoProvider` that delegates the sequence string to a **PL/pgSQL function named per-sequence in a SysConfig** (`de.metas.document.seqNo.DBFunctionSequenceNoProvider.<AD_Sequence.Name>.dbFunctionName`). Fail-loud: applicable only when the SysConfig is set, the function exists, and the eval-context carries a `Record_ID`. Schema-qualified function-name validation guards against injection.
- **Lot-sequence eval-context** — carry the `PP_Order` behind a finished-goods receipt into the `DocumentNoBuilder` evaluation context (`Record_ID`), so a provider can pass it to its DB function. Default path (no provider) is unchanged.
- **Provider registration** — `DBFunctionSequenceNoProvider` registered in `AD_JavaClass` so it is selectable as `AD_Sequence.CustomSequenceNoProvider_JavaClass_ID`.
- **`S_Resource.LotNumberCode`** — new optional `VARCHAR(10)` column + AD metadata, shown as an editable field on the Resource window (so a per-resource code can be embedded into generated lot numbers).

### Notes
- **Opt-in by construction**: until a sequence is configured with this provider (+ its SysConfig), every code path is identical to today.
- The concrete lot-number format lives in a customer-side PL/pgSQL function + configuration (separate customer-repo change), not in core.

### Status
**Draft** — core mechanism complete and reviewed. Still in progress: an automated integration test (Cucumber, validated on CI), a Playwright UI test for the new field, and the customer-side function + configuration.

### DIFF
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/document/sequenceno/DBFunctionSequenceNoProvider.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/document/sequenceno/DBFunctionSequenceNoProvider.java
@@ -1,0 +1,117 @@
+/*
+ * #%L
+ * de.metas.adempiere.adempiere.base
+ * %%
+ * Copyright (C) 2026 metas GmbH
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+package de.metas.document.sequenceno;
+
+import de.metas.common.util.time.SystemTime;
+import de.metas.document.DocumentSequenceInfo;
+import de.metas.util.Check;
+import de.metas.util.Services;
+import lombok.NonNull;
+import org.adempiere.service.ISysConfigBL;
+import org.compiere.util.DB;
+import org.compiere.util.Evaluatee;
+import org.compiere.util.Trx;
+
+import javax.annotation.Nullable;
+import java.sql.Timestamp;
+import java.util.regex.Pattern;
+
+/**
+ * Generic {@link CustomSequenceNoProvider} that delegates the whole sequence-number string to a PL/pgSQL function.
+ * <p>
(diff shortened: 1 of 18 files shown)


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/document/sequenceno/DBFunctionSequenceNoProvider.java
```diff
@@ -0,0 +1,124 @@
...
+				"{}={} must be a plain or single-schema-qualified SQL function identifier", sysConfigKey, functionName);
+		final String functionNameNorm = functionName.trim();
+
+		final int recordId = context.get_ValueAsInt(IPOReferenceAware.COLUMNNAME_Record_ID, -1);
+		Check.assume(recordId > 0, "context must carry a positive {}", IPOReferenceAware.COLUMNNAME_Record_ID);
+
+		final Timestamp generatedAt = Timestamp.from(SystemTime.asInstant());
+		final String result = DB.getSQLValueStringEx(ITrx.TRXNAME_None,
+				"SELECT " + functionNameNorm + "(?, ?)", recordId, generatedAt);
+		Check.assumeNotEmpty(result, "DB function {} returned empty for {}={}",
```
metas-ts, line 133: if the SQL did any raise notice to log something, then to Loggables.withLogger(logger, Level.DEBUG).addMsg(..the message..)

metas-ts: **Resolution (dev):** Done in `9480b4ebfe6`.

`provideSequenceNo` now executes the function via `DB.getSQLValueStringWithWarningEx(...)` (instead of `getSQLValueStringEx`), so the `SQLValueStringResult` carries the PostgreSQL `RAISE NOTICE` output — which PG delivers as the JDBC `SQLWarning` chain and metasfresh already extracts in `SQLUtil.extractWarningMessages` (the same bridge SQL-type AD_Processes use to log notices to `AD_PInstance_Log`). Each captured message is routed to the ambient loggable:

```java
final List<String> noticeMessages = sqlResult.getWarningMessages();
if (noticeMessages != null && !noticeMessages.isEmpty())
{
    final ILoggable loggable = Loggables.withLogger(logger, Level.DEBUG);
    noticeMessages.forEach(noticeMessage -> loggable.addLog("{}", noticeMessage));
}
```

So a `RAISE NOTICE` inside the lot-number function is now visible for debugging (DEBUG level), and the message text is passed as a parameter so `{}`/`%` in the notice can't break the log formatting.

PR #24696, opened 2026-06-23
HU Consolidation (mobile) — capture GRAIs on target LU + gate completion on required GRAIs

teosarca:
## What

Adds per-TU **GRAI** (returnable-asset code) capture to the **MobileUI HU Consolidation** app, reusing the existing `de.metas.handlingunits.grai` domain.

Backend (this part):
- **Set/get GRAIs on the consolidation target LU** — `PUT`/`GET /job/{wfProcessId}/target/grai` on `HUConsolidationRestController`, delegating to `HUGraiService` (no bespoke store). GRAIs are stamped onto the target LU's TUs as the existing `"GRAI"` HU attribute.
- **GRAI scan state in the job JSON** — `graiScanEnabled` (derived from the ship-to `C_BPartner.GRAIRequired`; `YesWithDummyGRAIs` treated as `Yes`) plus the target's expected/assigned GRAI counts (from `HUGraiSnapshot`).
- **Completion gate** — when GRAI is required and the target LU has unfilled TU GRAI slots, the Complete activity is blocked (`userConfirmed` throws a user-facing message) and reflected as not-ready in `computeActivityState`.
- `GRAIRequired = No` ⇒ **no behaviour change** in consolidation.

Shared `GraiScanHelper` resolves "is GRAI scan enabled" once, used by both the consolidate and complete activity handlers.

## Notes
- No DB migration (the `"GRAI"` PI-template attribute and `C_BPartner.GRAIRequired` already exist).
- Frontend (GRAI scan action + capture screen reusing `GraiCapturePanel`, completion gating, DE/EN i18n) and an E2E spec follow on this branch.
- Covered by unit tests on the command/REST, the JSON GRAI-state mapping, and the completion gate.


backend/de.metas.hu_consolidation.mobile/src/main/java/de/metas/hu_consolidation/mobile/job/HUConsolidationJobService.java
```diff
@@ -118,6 +125,43 @@ public void printTargetLabel(@NonNull final HUConsolidationJobId jobId, @NotNull
 		labelPrinter.printLabel(currentTarget);
 	}
 
+	public void setTargetGrais(
+			@NonNull final HUConsolidationJobId jobId,
+			@NonNull final UserId callerId,
+			@NonNull final GRAISet graiSet)
+	{
+		SetTargetGraisCommand.builder()
```
teosarca, line 133: this command makes no sense. it just invokes `huGraiService.setGrais(luId, graiSet);`

in case the commands are more glue code and a few lines of actual code, they make no sense. consider capturing this learning

teosarca: **Resolution (dev):** Agreed — it's needless glue: the command/service method only forwards to `huGraiService.setGrais(luId, graiSet)` (the read path `getTargetGraisSnapshot` is the same — a thin wrapper over `huGraiService.getSnapshot`). I'll inline it: drop `SetTargetGraisCommand` + the pass-through service method and call `HUGraiService` directly from the REST controller's target-GRAI path (after the existing access/target-LU resolution), preserving the test coverage. It will land in the next commit on this branch (the PR is currently paused pending a test-setup decision, so nothing has merged). Will also capture the learning: don't introduce a Command/handler (or service method) that is only glue + a couple of lines — call the domain service directly.

teosarca:
Closing this PR. The implemented approach — capturing GRAIs and gating workflow completion during HU consolidation — was based on an incorrect requirement premise.

The actual requirement is to use a TU's GRAI as a **scan-identifier** for selecting which TU to consolidate onto the pallet (a much smaller change, reusing the existing GRAI parsing + an HU-attribute lookup). The feature is being re-specified from scratch; a new PR will follow.

PR #24697, opened 2026-06-23
WebUI: fix Preislisten-Schema dropdown 404 when creating a colliding new Price List Version (don't evict error-state root owning an unsaved new included document)

teosarca:
## What

Fixes the `Preislisten-Schema` (`M_DiscountSchema_ID`) dropdown returning `DocumentNotFoundException` (HTTP 404) right after creating a new Price List Version (PLV) whose `ValidFrom` collides with an existing PLV on the same price list.

## Root cause

The colliding new PLV fails its INSERT on a unique index. The error-state **root** price-list document still owns the **unsaved, new, in-memory child** PLV. The eviction escape-hatch in `DocumentCollection` (`shouldInvalidateRootOnChildInvalidation`) then evicted that root from cache, discarding the unsaved child with it — so the next read-only load could not find the child and the dropdown GET 404'd. The existing `!rootDocument.isNew()` guard only protected a new **root**, not a root owning a new **child**.

## Fix

Extend the "don't evict unsaved in-memory work" intent from new-root to new-child:
- `DocumentCollection.shouldInvalidateRootOnChildInvalidation(...)` gains a 4th dimension `rootHasUnsavedNewIncludedDocument`; when set it returns "don't evict" first (mirroring the existing new-root protection, which already wins over full-invalidation).
- The child-invalidation call site computes it via a new read-only accessor `Document#hasUnsavedNewIncludedDocuments()` → `IIncludedDocumentsCollection#hasNewDocumentsWithChanges()` (overridden in `HighVolumeReadWriteIncludedDocumentsCollection` as `getChangedDocuments().anyMatch(isNew && hasChanges)`).

No migration / AD change — the friendly unique-constraint message already exists on the instance; this is a pure WebUI behaviour fix.

## Safety vs prior self-heal eviction

The self-heal eviction added in https://github.com/metasfresh/metasfresh/pull/23672 targets a **deleted persisted** child (not new) — the new guard does not trigger there, so that self-heal still fires. The admin force-forget path (`invalidateAll()`) is a separate method, untouched.

## Tests

`DocumentCollectionShouldInvalidateRootTest` extended (pure-boolean, mock-free) — RED→GREEN confirmed; 6/6 pass.

### DIFF
--- a/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/DocumentCollection.java
+++ b/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/DocumentCollection.java
@@ -695,7 +695,7 @@ public class DocumentCollection
 						documentToInvalidate.isInvalidateDocument(),
 						rootDocument.getSaveStatus().isError(),
 						rootDocument.getValidStatus().isValid(),
-						false) // rootHasUnsavedNewIncludedDocument is wired at this call site in a follow-up commit
+						rootDocument.hasUnsavedNewIncludedDocuments())
 						&& !rootDocument.isNew())
 				{
 					rootDocuments.invalidate(rootDocumentKey);
--- a/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/DocumentCollection.java
+++ b/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/DocumentCollection.java
@@ -617,12 +617,23 @@ public class DocumentCollection
 	 *
 	 * <p>Pure boolean signature on purpose so it can be unit-tested without needing to mock
 	 * {@link Document} (which is final and has a non-trivial constructor).
+	 *
+	 * @param rootHasUnsavedNewIncludedDocument the cached root owns an unsaved, new, in-memory
+	 *                                           included document whose work would be lost on eviction
 	 */
 	static boolean shouldInvalidateRootOnChildInvalidation(
 			final boolean callerRequestedFullInvalidation,
 			final boolean rootHasSaveError,
-			final boolean rootValidStatusIsValid)
+			final boolean rootValidStatusIsValid,
+			final boolean rootHasUnsavedNewIncludedDocument)
 	{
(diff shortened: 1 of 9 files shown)


backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/DocumentCollection.java
```diff
@@ -683,7 +694,8 @@ private void invalidate(@NonNull final DocumentToInvalidate documentToInvalidate
 				if (shouldInvalidateRootOnChildInvalidation(
 						documentToInvalidate.isInvalidateDocument(),
 						rootDocument.getSaveStatus().isError(),
-						rootDocument.getValidStatus().isValid())
+						rootDocument.getValidStatus().isValid(),
+						rootDocument.hasUnsavedNewIncludedDocuments())
 						&& !rootDocument.isNew())
```
teosarca: wondering why 
* rootDocument.isNew() is not part of the shouldInvalidateRootOnChildInvalidation
* and in that case maybe some of the shouldInvalidateRootOnChildInvalidation params shall be Supplier<Boolean>

teosarca: **Resolution (dev):** Adopted both points.

- `rootIsNew` is now a parameter of `shouldInvalidateRootOnChildInvalidation(...)`, checked first (a new root is never evicted). The whole eviction decision now lives in the one unit-tested helper; the `&& !rootDocument.isNew()` is removed from the call site.
- The genuinely expensive input — `rootHasUnsavedNewIncludedDocument`, which walks the included collections — is now a `BooleanSupplier`, evaluated **last and only when the root would otherwise be evicted**, so the happy path never pays for it. The cheap field reads (`callerRequestedFullInvalidation`, `rootHasSaveError`, `rootValidStatusIsValid`) stay plain booleans since they're consulted to decide whether we'd evict at all. Added a unit test asserting the supplier is not invoked on the no-evict path.

Commit: https://github.com/metasfresh/metasfresh/commit/5caac3c75135b5b437af529d620b7f58d6e069e2

PR #24698, opened 2026-06-23
E-Invoicing: ZUGFeRD output (PDF/A-3 + embedded CII) + email delivery

metas-ts:
## E-Invoicing: ZUGFeRD output (PDF/A-3 + embedded CII) + email delivery

For sales-invoice recipients configured as ZUGFeRD (`EInvoiceType=Z`), the invoice is produced as a valid **PDF/A-3 with the EN 16931 CII embedded** (`factur-x.xml`), delivered through the existing document-outbound mailer, and invoice completion is blocked on invalid CII.

### What it does
- **PDF/A-3 rendering** — `AD_Process.IsPdfA3Output` makes `JasperEngine` emit PDF/A-3B (sRGB OutputIntent, embedded fonts: Liberation Sans for Arial/SansSerif, OCR-B retained pixel-correct for the ESR/QR codeline, XMP metadata).
- **CII embedding** — a Mustangproject-based `ZugferdAssembler.embed(pdfA3, ciiXml)` produces the Factur-X container; a report-agnostic SPI seam in `DefaultModelArchiver` embeds the validated CII at archive time for `EInvoiceType=Z` invoices only (no-op otherwise — the archive layer keeps no dependency on the e-invoice module).
- **Completion gate** — the `C_Invoice` `AFTER_COMPLETE` interceptor validates the CII (EN 16931) and blocks completion with a localized error on invalid data, mirroring the existing XRechnung gate.
- **PEPPOL** — the unimplemented PEPPOL `EInvoiceType` value is deactivated.

### Notable fix
`EInvoiceCiiService.marshalToXml` now applies a JAXB namespace-prefix mapper (`rsm`/`ram`/`udt`/`qdt`) — JAXB-RI's auto-generated `ns2`/`ns3` prefixes are rejected by Mustangproject's `setXML()`, which would have silently broken real ZUGFeRD archives.

### Tests
- Unit/integration: font embedding, PDF/A-3B conformance, `ZugferdAssembler` (container valid + CII round-trip), the archive embed seam, and the completion gate.
- **Cucumber (Mailpit)** — `eInvoiceZugferdEmail.feature`: complete a Z invoice → archived ZUGFeRD PDF emailed → received attachment validated as ZUGFeRD.
- **Playwright** — `einvoice-zugferd.spec.js`: a real UI sales-order→invoice flow whose **archived** invoice PDF is asserted to embed `factur-x.xml` (it is a genuine ZUGFeRD PDF) **and** to retain intact invoice content. Self-contained via extensions to the frontend-testing masterdata API (e-invoice BPartner fields, location address, IBAN, org-seller config, default contact, `IsPdfA3Output` seam).
- Also: the org-master window spec was generalised (`organisation-stammdaten-window.spec.js`) and made language-independent (en_US + de_DE).


backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/mail/Mailpit_StepDef.java
```diff
@@ -75,6 +84,10 @@ private static String apiBaseUrl()
 	 * Deletes all messages currently held by Mailpit, so a scenario starts from an empty inbox.
 	 *
 	 * @cucumber.stepdef
+	 * @cucumber.example
```
metas-ts, line 87: if the "given" has no table and not even a variable, then no example is needed

metas-ts: **Resolution (dev):** Valid nit — `mailpit inbox is cleared` is an argument-less `@Given` (no DataTable, no variable), so its `@cucumber.example` block adds nothing. It went in as-is and is still present at the merged HEAD, so this is acknowledged as an open doc-only nit rather than something changed here; it will be dropped on the next touch of `Mailpit_StepDef.java` (not worth a dedicated follow-up for a Javadoc example). Thanks for flagging.

PR #24700, opened 2026-06-23
Sales Order Candidate to Order — keep per-line delivery date when aggregating multi-date order candidates

mrie04:
## What

When sales order candidates that belong to one external order (same external system + external header reference) carry **different delivery dates**, build **one** sales order whose **lines keep their own delivery date** (header = earliest), and make the "ship/pick after above date" flags hold **each line until its own date**. System-wide (no config gate).

## Changes

- **Aggregation** — drop `DatePromised` from the order-split key (`DefaultGroupingProvider`, `OLCandProcessingHelper`): candidates of one external order that differ only by delivery date no longer split into multiple orders (which also removes the `C_Order_ExternalHeader_ID` unique-index crash that the split caused).
- **Per-line date** — `OLCandOrderFactory`: each `C_OrderLine` keeps its source candidate's `DatePromised`; the `C_Order` header is set to the earliest of the lines. Sequenced so `MOrder.afterSave`'s header→lines broadcast does not clobber the per-line dates.
- **Ship flag** (`C_Order.IsFixedDatePromised`) — the two enforcement filters compare the per-line `M_Packageable_V.DeliveryDate` instead of the header `DatePromised` (`PackagingDAO`, `ShipmentService`). Java-only (the view column already exists).
- **Pick flag** (`C_Order.IsFixedPreparationDate`) — per-line preparation date: `OrderDeliveryDayBL` is refactored to expose a per-delivery-date preparation-date computation (the same tour / no-tour-fallback / offset / sysconfig logic that produces today's header value), called per-line by `ShipmentScheduleOrderReferenceProvider`. Single-date orders are byte-identical to before (no regression).
- **Migration** — `AD_Index_Table` record for the existing `C_Order_ExternalHeader_ID` unique index with a translated (DE/EN/FR) `ErrorMsg`, so a forced-split conflict (candidates sharing an external header reference but forced apart by another field) surfaces a clear, translatable message instead of the generic fallback. Gated automatically by the index's `WHERE ExternalId IS NOT NULL`.

## Tests

Cucumber coverage for: multi-date candidates → one order with per-line dates and earliest header; ship-after-date holding each line by its own delivery date; pick-after-date holding each line by its own preparation date; single-date no-regression. Behaviour verified RED→GREEN on CI.


backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/salesOrderCandidateToOrderWorkpackageProcess.feature
```diff
@@ -293,3 +293,276 @@ Feature: Enqueue order candidate in multiple workpackages for processing to orde
...
+    "ship": false,
+    "invoice": false,
+    "closeOrder": false
+}
+"""
+    Then process metasfresh response
+      | C_Order_ID.Identifier |
+      | order_1               |
+
+    And validate the number of C_Orders by ExternalId
```
mrie04: is this additional validation needed, as unique index pevents other cases  anyway and orderline count verifies it too

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/order/C_Order_StepDef.java
```diff
@@ -702,6 +715,48 @@ public void validate_created_order(@NonNull final DataTable table)
...
+	 * </ul>
+	 * Example:
+	 * <pre>
+	 * Then validate the number of C_Orders by ExternalId
+	 *   | ExternalId | ExternalSystem.Value | Count |
+	 *   | 22062026   | Shopware6            | 1     |
+	 * </pre>
+	 */
+	@Then("validate the number of C_Orders by ExternalId")
+	public void validate_number_of_orders_by_externalId(@NonNull final DataTable table)
```
mrie04: rly needed, see feature comment

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/salesOrderCandidateToOrderWorkpackageProcess.feature
```diff
@@ -293,3 +293,276 @@ Feature: Enqueue order candidate in multiple workpackages for processing to orde
...
+    And metasfresh contains M_Products:
+      | Identifier       | Name             |
+      | product_18062027 | product_18062027 |
+    And metasfresh contains M_ProductPrices
+      | Identifier | M_PriceList_Version_ID.Identifier | M_Product_ID.Identifier | PriceStd | C_UOM_ID.X12DE355 | C_TaxCategory_ID.InternalName |
+      | pp_product | plv_scenario_18062027             | product_18062027        | 10.0     | PCE               | Normal                        |
+    # Two customers sharing the same pricing system: 'shipCustomer' is the common ship partner of both olcands,
+    # while 'billCustomerA' and 'billCustomerB' are the two DIFFERENT bill partners that force the order split.
+    And metasfresh contains C_BPartners:
+      | Identifier    | OPT.IsCustomer | OPT.IsVendor | M_PricingSystem_ID.Identifier | OPT.C_BPartner_Location_ID | GLN           | deliveryRule |
```
mrie04: avoid OPT. and .Identifier where possible. check all new scenarios

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/salesOrderCandidateToOrderWorkpackageProcess.feature
```diff
@@ -293,3 +293,276 @@ Feature: Enqueue order candidate in multiple workpackages for processing to orde
...
+      | 22062026   | Shopware6            | 1     |
+
+    # The header DatePromised must equal the EARLIEST DatePromised among the lines (2026-07-01).
+    And validate the created orders
+      | C_Order_ID.Identifier | externalId | C_BPartner_ID.Identifier | C_BPartner_Location_ID.Identifier | DateOrdered | DatePromised | DocBaseType | currencyCode | DeliveryRule | DeliveryViaRule | poReference | processed | DocStatus |
+      | order_1               | 22062026   | olCand_Customer          | olCand_Customer_location          | 2026-06-22  | 2026-07-01   | SOO         | EUR          | F            | S               | 22062026    | true      | CO        |
+    # Each order line must keep the DatePromised (dateRequired) of its own source candidate.
+    # The line is matched by QtyOrdered: qty 2 -> 2026-07-01, qty 1 -> 2026-07-08, qty 3 -> 2026-07-15.
+    And validate the created order lines
+      | C_OrderLine_ID.Identifier | C_Order_ID.Identifier | DatePromised | M_Product_ID.Identifier | qtydelivered | QtyOrdered | qtyinvoiced | price | discount | currencyCode | processed |
```
mrie04: no prepdate validation? and why not also validate the shipmentschedules?

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/ShipmentService.java
```diff
@@ -297,7 +297,7 @@ public IQueryFilter<I_M_ShipmentSchedule> createShipmentScheduleEnqueuerQueryFil
 					)
 					.filter(queryBL.createCompositeQueryFilter(I_M_Packageable_V.class)
 							.setJoinOr()
-							.addCompareFilter(I_M_Packageable_V.COLUMNNAME_DatePromised, CompareQueryFilter.Operator.LESS_OR_EQUAL, nowInstant)
+							.addCompareFilter(I_M_Packageable_V.COLUMNNAME_DeliveryDate, CompareQueryFilter.Operator.LESS_OR_EQUAL, nowInstant)
```
mrie04: this is now an mismatch on names between flag and date. pls explain this change

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
```diff
@@ -363,6 +370,8 @@ else if (orderLines.isEmpty())
 		}
 		else
 		{
+			applyPerLineDatePromised(order);
```
mrie04, line 373: why at this late point?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/tourplanning/api/impl/OrderDeliveryDayBL.java
```diff
@@ -92,28 +149,28 @@ private boolean setPreparationDateAndTour0(@NonNull final I_C_Order order, final
 				context,
 				soTrx,
 				calculationTime,
-				datePromised,
+				deliveryDate,
```
mrie04, line 238: again why this change?

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/salesOrderCandidateToOrderWorkpackageProcess.feature
mrie04: **Resolution (dev):** Agreed, removed. The `C_Order_ExternalHeader_ID` unique index is on `(ExternalSystem_ID, ExternalId, AD_Org_ID)`, so a count-by-ExternalId can only ever be 0 or 1 — a pre-fix split would have *violated* that index (failed save), not produced count=2. So the step can't actually detect a split; the single-order result is already proven by the `process metasfresh response` (one `order_1`) plus the three per-line order-line assertions. Step removed in https://github.com/metasfresh/metasfresh/commit/4c4a9c19166.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/order/C_Order_StepDef.java
mrie04: **Resolution (dev):** Removed. With the feature usage gone (see the feature comment), the `validate the number of C_Orders by ExternalId` step def is dead code — deleted the whole method in https://github.com/metasfresh/metasfresh/commit/4c4a9c19166. `externalSystemRepository` and friends stay (still used by other steps).

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/salesOrderCandidateToOrderWorkpackageProcess.feature
mrie04: **Resolution (dev):** Done across both new scenarios — dropped `OPT.` prefixes and `.Identifier` suffixes and re-aligned the tables (https://github.com/metasfresh/metasfresh/commit/4c4a9c19166). One correction in https://github.com/metasfresh/metasfresh/commit/30aa1c51995: the three `process metasfresh response` / `...JsonOLCandCreateBulkResponse` tables keep `.Identifier` — those steps use legacy `DataTableUtil.extractStringForColumnName` (exact match), not `DataTableRow` smart lookup, so the suffix is required there. The `validate the created orders`/`order lines` tables use the clean bare form.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/salesOrderCandidateToOrderWorkpackageProcess.feature
mrie04: **Resolution (dev):** Added. The A2 scenario now asserts each line's shipment schedule and its per-line `PreparationDate` (https://github.com/metasfresh/metasfresh/commit/4c4a9c19166): with no tour configured the prep date equals the delivery date equals the line's own `DatePromised` (2026-07-01 / 07-08 / 07-15) — i.e. the per-line date flows olcand → order line → schedule end-to-end.

Note on "prepdate": there is deliberately **no** `C_OrderLine.PreparationDate` column (`DESIGN.md`: no new C_OrderLine columns). The per-line preparation date lives on the line's `M_ShipmentSchedule` (derived per line in Task B2), which is exactly what the new assertion validates.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/ShipmentService.java
mrie04: **Resolution (dev):** Already correct — explanation. The header flag `C_Order.IsFixedDatePromised` is unchanged (still one toggle per order). What changed is only the date the filter compares against: from the header `DatePromised` to the per-line `M_Packageable_V.DeliveryDate`, so the binding becomes per-line (AC: a line is held until its OWN date). The name "mismatch" is inherent to the data model — the per-line date field exposed on the schedule/packageable is named `DeliveryDate`, while the order-header flag is named `IsFixedDatePromised`. No code change.

backend/de.metas.salescandidate.base/src/main/java/de/metas/ordercandidate/api/OLCandOrderFactory.java
mrie04: **Resolution (dev):** Already correct — explanation. `MOrder.afterSave()` broadcasts the header `DatePromised` onto every order line. So the sequence must be: set header = earliest, save (the broadcast is then harmless), and write each line's own `DatePromised` **last** so it survives. Doing it earlier (e.g. while building lines) would let the header save clobber the per-line values. Hence it runs at the latest safe point in `completeOrDelete`, just before `processEx`. No change.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/tourplanning/api/impl/OrderDeliveryDayBL.java
mrie04: **Resolution (dev):** Already correct — explanation. B2 needs the preparation date derived from a **per-line** delivery date, not the header's. The old logic computed it only from the order header. So I extracted the shared core `computePreparationDateAndTour0(order, deliveryDate, ...)` and exposed `computePreparationDate(order, deliveryDate)`, so `ShipmentScheduleOrderReferenceProvider` can compute each schedule's `PreparationDate` from its line's own delivery date — preserving the exact tour / no-tour-fallback / offset / sysconfig logic the header uses. The header path is behaviourally unchanged (it now calls the same core). No behaviour change for single-date orders.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/tourplanning/api/impl/OrderDeliveryDayBL.java
```diff
@@ -71,6 +75,77 @@ private boolean setPreparationDateAndTour0(@NonNull final I_C_Order order, final
 			return false;
 		}
 
+		// Compute the header's preparation date + tour from the header's DatePromised.
+		final PreparationDateAndTour preparationDateAndTour = computePreparationDateAndTour0(order, datePromised, fallbackToDatePromised, timeZone);
+
+		order.setPreparationDate(TimeUtil.asTimestamp(preparationDateAndTour.getPreparationDate()));
+		final TourId tourId = preparationDateAndTour.getTourId();
+		// convention: clear an FK setter with -1 (see docs/coding-rules/java-general.md), matching the original
+		order.setM_Tour_ID(tourId != null ? tourId.getRepoId() : -1);
```
mrie04: TourId.toRepoId (consider to bring the function in line and adjust to -1)

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/tourplanning/api/impl/OrderDeliveryDayBL.java
```diff
@@ -71,6 +75,77 @@ private boolean setPreparationDateAndTour0(@NonNull final I_C_Order order, final
...
+
+		return true;
+	}
+
+	@Override
+	public ZonedDateTime computePreparationDate(@NonNull final I_C_Order order, @NonNull final ZonedDateTime deliveryDate)
```
mrie04: can be private

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/tourplanning/api/impl/OrderDeliveryDayBL.java
```diff
@@ -71,6 +75,77 @@ private boolean setPreparationDateAndTour0(@NonNull final I_C_Order order, final
...
+	{
+		// Defensive: the tour calculation requires a bpartner location. A shipment-schedule order always has one,
+		// but if it is somehow absent, fall back to the order's already-stored PreparationDate (legacy behavior)
+		// rather than risk an NPE in the tour calculation. (The header path setPreparationDateAndTour0 instead
+		// returns false / leaves the value untouched here; this method must return a non-null date to the
+		// shipment-schedule provider, so it returns the stored value rather than skipping.)
+		if (BPartnerLocationId.ofRepoIdOrNull(order.getC_BPartner_ID(), order.getC_BPartner_Location_ID()) == null)
+		{
+			return TimeUtil.asZonedDateTime(order.getPreparationDate());
```
mrie04: replace deprecated org.compiere.util.TimeUtil.asZonedDateTime(java.sql.Timestamp, java.time.ZoneId) and is this much comment needed

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/tourplanning/api/impl/OrderDeliveryDayBL.java
mrie04: it's called deliverydate on shipment schedule but not order, missleading here

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/tourplanning/api/impl/OrderDeliveryDayBL.java
```diff
@@ -71,6 +75,77 @@ private boolean setPreparationDateAndTour0(@NonNull final I_C_Order order, final
...
+	 * @return the computed preparation date (may be {@code null} when there is no usable tour and the fallback is
+	 *         disabled) together with the tour to assign ({@code null} when no tour was used).
+	 */
+	private PreparationDateAndTour computePreparationDateAndTour0(
+			@NonNull final I_C_Order order,
+			@NonNull final ZonedDateTime deliveryDate,
+			final boolean fallbackToDatePromised,
+			@NonNull final ZoneId timeZone)
+	{
+		final BPartnerLocationId bpartnerLocationId = BPartnerLocationId.ofRepoIdOrNull(order.getC_BPartner_ID(), order.getC_BPartner_Location_ID());
```
mrie04, line 213: Check.assumeNotNull as function below expects it

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/tourplanning/api/impl/OrderDeliveryDayBL.java
mrie04: **Resolution (dev):** Done — `setM_Tour_ID` now uses `TourId.toRepoId(preparationDateAndTour.getTourId())` (handles null→-1) instead of the inline ternary; dropped the redundant comment. https://github.com/metasfresh/metasfresh/commit/a5838cc2dc4

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/tourplanning/api/impl/OrderDeliveryDayBL.java
mrie04: **Resolution (dev):** Done — made the 2-arg `computePreparationDate(order, deliveryDate)` `private` and removed it from `IOrderDeliveryDayBL`. Only the 3-arg per-line overload is called externally (by `ShipmentScheduleOrderReferenceProvider`); the 2-arg is now an internal helper of the 3-arg. https://github.com/metasfresh/metasfresh/commit/a5838cc2dc4

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/tourplanning/api/impl/OrderDeliveryDayBL.java
mrie04: **Resolution (dev):** Done — replaced the deprecated `TimeUtil.asZonedDateTime(Timestamp)` with the non-deprecated `asZonedDateTime(Timestamp, ZoneId)` (passing the order's time zone) and trimmed the over-long comment to one line. https://github.com/metasfresh/metasfresh/commit/a5838cc2dc4

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/tourplanning/api/impl/OrderDeliveryDayBL.java
mrie04: **Resolution (dev):** Acknowledged — the parameter is the per-line **delivery date** (the line's `DatePromised`), which `M_ShipmentSchedule` stores under the column name `DeliveryDate` (same business date, two table conventions: order=DatePromised, schedule=DeliveryDate). The param is named `deliveryDate` to match the schedule-side caller it feeds. If you'd prefer `lineDeliveryDate` for clarity on the order side, I'll rename — say the word.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/tourplanning/api/impl/OrderDeliveryDayBL.java
mrie04: **Resolution (dev):** Also adjusted the helper itself — `TourId.toRepoId(null)` now returns `-1` (was `0`), matching the FK-clear convention (java-general §35). Both call sites pass it straight to `setM_Tour_ID` (`<1` guard), so behaviour is unchanged. https://github.com/metasfresh/metasfresh/commit/54101e89c3e

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/tourplanning/api/impl/OrderDeliveryDayBL.java
mrie04: **Resolution (dev):** Done — `computePreparationDateAndTour0` now `Check.assumeNotNull(bpartnerLocationId, ...)` right after the `ofRepoIdOrNull`. Both callers (`computePreparationDate`, `setPreparationDateAndTour0`) only reach it with a bpartner location set, and `calculateTourAndPreparationDate` requires it non-null — so this asserts the existing invariant explicitly. https://github.com/metasfresh/metasfresh/commit/cb306a3d91f

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/salesOrderCandidateToOrderWorkpackageProcess.feature
```diff
@@ -293,3 +293,285 @@ Feature: Enqueue order candidate in multiple workpackages for processing to orde
...
+      | olCand_Customer | olCand_Customer_22062026 | Y          | N        | ps_scenario_22062026 | olCand_Customer_location | 9988776655443 | F            |
+    And metasfresh contains C_BPartner_Locations:
+      | Identifier               | GLN           | C_BPartner_ID   |
+      | olCand_Customer_location | 9988776655443 | olCand_Customer |
+
+    # we create 3 OLCands with the same externalHeaderId `22062026`, same externalSystemCode and org, same poReference and dateOrdered,
+    # but DIFFERENT dateRequired (which maps to C_OLCand.DatePromised).
+    # Before the fix, the differing DatePromised splits them into 3 separate C_Orders, each carrying the same
+    # C_Order.ExternalId=22062026 -> the unique index C_Order_ExternalHeader_ID is violated and processing fails.
+    # After the fix, DatePromised is no longer part of the order-split key, so they aggregate into ONE C_Order with 3 lines.
```
mrie04: no history comments

mrie04: **Resolution (dev):** Removed the before/after-fix history narrative and rewrote it as the timeless behavioral invariant (DatePromised is not part of the order-aggregation key) in https://github.com/metasfresh/metasfresh/commit/7ac6afac232471395230a88a36357f882abd0f71 , then reformatted it as bullets in https://github.com/metasfresh/metasfresh/commit/76c1f1eaf7b6c83f93bdb27e823339a3b76ff6c7 .

PR #24743, opened 2026-06-24
Manufacturing — AveragePO: recover actual production cost into finished good so WIP clears

adi-stefan:
## Problem

In **AveragePO** manufacturing, the finished-good material receipt was valued at the **frozen planned BOM-rollup price** captured when the order was completed, while component issues post their **actual** cost. When the two differ — e.g. a component's cost rises between order completion and the actual material issue — the **work-in-process account is left with an unrecovered residual on a closed order** instead of clearing to zero. The finished good is received below its true production cost.

## Fix

Two coupled changes in `ManufacturingAveragePOCostingMethodHandler`:

1. **`createComponentIssue`** accumulated the *raw* request amount (which is `0`) into the order costs instead of the *computed* actual issue cost, so the order's post-calculation never reflected the real inbound component cost (the post-calc value was computed but effectively always `0`, which is why it was never consumed). It now accumulates the effective (current-cost-priced) amount/qty, negated to a positive inbound value (mirroring the outbound negation on receipt).

2. **`createMainProductOrCoProductReceipt`** now values the finished-good receipt at the order's **actual post-calculation amount** (the production cost not yet recovered by earlier receipts) instead of the frozen rollup price. The finished good absorbs the real production cost and WIP clears to zero on a closed order.

## Scope & tests

- AveragePO manufacturing only; standard costing untouched.
- Covered by `manufacturingCostCollectorPosting.feature`:
  - existing **balanced** scenario (regression — unchanged result), and
  - a new **cost-rise** scenario that asserts `P_WIP_Acct` nets to zero across the order's cost collectors (fails before the fix with a residual, passes after).
- Limitation (documented): valuing at receipt fully recovers WIP for the normal flow where all component issues precede a single full finished-good receipt; multi-receipt-with-interleaved-issues of one order is out of scope for this change.

### DIFF
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/costing/manufacturingCostCollectorPosting.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/costing/manufacturingCostCollectorPosting.feature
@@ -9,6 +9,11 @@ Feature: Manufacturing cost collector posting - component issue vs material rece
   # A manufacturing order backflushes one BOM component and receives the finished good.
   # The component-issue cost collector must post DR P_WIP_Acct / CR P_Asset_Acct (inventory down, WIP up).
   # The finished-good material-receipt cost collector must post DR P_Asset_Acct / CR P_WIP_Acct (inventory up, WIP down).
+  #
+  # For a CLOSED AveragePO order all production cost must be recovered into the finished good, so the
+  # P_WIP_Acct nets to zero across the order's cost collectors. This holds even when the component's
+  # actual cost at issue time differs from the planned BOM-rollup frozen at order completion (e.g. the
+  # component price rose between production planning and the actual material issue).
 
   Background:
     Given infrastructure and metasfresh are running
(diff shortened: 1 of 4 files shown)


adi-stefan:
Closing — superseded by a corrected root-cause diagnosis.

This PR re-priced the finished-good receipt to the actual post-calculation cost so the WIP account would clear. Further investigation against the real data showed that the WIP residual is not caused by the finished good being undervalued, but by the **component issue being overcharged**: the component is a catch-weight product stocked per piece (1 piece = a ~35 kg wheel) with a per-piece cost, and issuing a partial quantity (e.g. 2.3 kg) is costed as a full piece instead of being prorated (2.3 / 35 of the piece cost).

Forcing the finished good to absorb that inflated component cost would only move the error from WIP into finished-goods inventory. The correct fix is in the catch-weight quantity→cost conversion for partial component issues, which is being handled separately. Abandoning this approach.

adi-stefan:
Reopened. The earlier closure assumed the whole-unit component cost was inflated and should prorate to the issued fraction. The costing feature owner has since confirmed that consuming a **whole** catch-weight unit in manufacturing is the correct, intended behaviour — so recovering that unit's actual production cost into the finished good (this PR) is the right outcome, not a masking of an inflated value.

Re-merged onto the current base. The AveragePO WIP-recovery cucumber (`manufacturingCostCollectorPosting.feature`) passes both scenarios locally:
- balanced order (regression — WIP still nets to 0), and
- unbalanced order where the component's actual cost exceeds the finished good's standing average → for the completed order, all production cost is recovered into the finished good and WIP clears.

adi-stefan:
as agreed, we will not pursue the automatic cost recovery, but instead the cost shall be recovered via an additional cost collector type.

PR #24744, opened 2026-06-24
Invoice candidates: inherit PaymentRule from bill-partner for order-less deliveries (Leergut)

metas-rc:
## Problem
On a consolidated delivery that bundles multiple orders (`M_InOut.C_Order_ID = NULL`), returnable/packaging lines have no order line, so their invoice candidates are built from `M_InOutLine` via `M_InOutLine_Handler`. The no-order path never set `PaymentRule`, so the candidate kept the column default while the order-based goods carry the order's rule. Because `PaymentRule` is part of the invoice header aggregation key, returnables (Leergut) split onto a separate invoice.

## Fix
In the no-order branch that already inherits `InvoiceRule` from the bill-partner (where the bill-partner is populated), also inherit `PaymentRule` from the bill-partner (guarded with `Check.isNotBlank`). Single, localized change mirroring the existing `InvoiceRule` precedent.

## Tests
`M_InOutLine_HandlerTest` adds 3 cases: order-less delivery inherits the bill-partner rule; delivery with an order still inherits from the order; bill-partner without a rule leaves the candidate unchanged (no error).

## Notes
Related: an earlier alternative approach in https://github.com/metasfresh/metasfresh/pull/23408 (open/stalled, no tests, and crashes on the create path). This PR takes the minimal placement that keeps the bill-partner populated.


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
```diff
@@ -367,23 +368,25 @@ private I_C_Invoice_Candidate createInvoiceCandidateForInOutLineOrNull(
...
-
-			if (invoiceRule!= null)
-			{
-				icRecord.setInvoiceRule(invoiceRule.getCode());
-			}
-			else
-			{
-				icRecord.setInvoiceRule(X_C_Invoice_Candidate.INVOICERULE_Immediate); // Immediate
-			}
+			final BPartnerEffective billBPartnerEffective = bPartnerEffectiveBL.getByRecord(billBPartner);
```
mrie04: you could use getById instead of dao above

backend/de.metas.swat/de.metas.swat.base/src/test/java/de/metas/inout/invoicecandidate/M_InOutLine_HandlerTest.java
```diff
@@ -770,4 +780,145 @@ public boolean matches(final I_C_Invoice_Candidate ic)
...
+		save(billBPartner);
+
+		// when: the packaging invoice candidate is created (no order on the inout)
+		final List<I_C_Invoice_Candidate> result = inOutLineHandlerUnderTest.createCandidatesForInOutLine(packagingInOutLine);
+		result.forEach(InterfaceWrapperHelper::saveRecord);
+		assertThat(result).hasSize(1);
+		final I_C_Invoice_Candidate ic = result.get(0);
+
+		// then: PaymentRule is inherited from the bill-partner, not stuck at the column default "P"
+		assertThat(ic.getPaymentRule()).isEqualTo(X_C_BPartner.PAYMENTRULE_DirectDeposit);
```
mrie04: consider like de.metas.payment.PaymentRule#isOnCredit instead

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
metas-rc: **Resolution (dev):** Adopted. Collapsed the explicit `bpartnerDAO.getById(...)` + `bPartnerEffectiveBL.getByRecord(...)` into a single `bPartnerEffectiveBL.getById(BPartnerId.ofRepoId(icRecord.getBill_BPartner_ID()))` (that overload does exactly the load-then-getByRecord), and removed the now-unused local. https://github.com/metasfresh/metasfresh/commit/0c6031144d5

backend/de.metas.swat/de.metas.swat.base/src/test/java/de/metas/inout/invoicecandidate/M_InOutLine_HandlerTest.java
metas-rc: **Resolution (dev):** Adopted. The result assertions now use the `de.metas.payment.PaymentRule` enum instead of the raw `X_C_BPartner.PAYMENTRULE_*` String constants: `assertThat(PaymentRule.ofCode(ic.getPaymentRule())).isEqualTo(PaymentRule.DirectDeposit/DirectDebit/OnCredit)`. https://github.com/metasfresh/metasfresh/commit/0c6031144d5

mrie04:

PR #24067, opened 2026-05-21
MobileUI — emit user-friendly error messages for QR scan failures

mrie04:
## Summary

- Adds `MobileQRCodeMessages` utility class consolidating all `AdMessageKey` constants for picking, distribution, inventory, and HU-consolidation mobile workflows.
- Replaces every raw-string exception message at mobile QR scan throw sites with the `AdempiereException(AdMessageKey, ...)` overload so the frontend renders a translated, actionable message instead of the generic "Oops" screen.
- Adds `DistributionHUService.assertHUCanBePickedFrom()` to wire the three distribution business-check constants (`HU_RESERVED_BY_OTHER`, `HU_NOT_AT_TARGET`, `HU_ALREADY_AT_TARGET`) to real validation logic.
- Migration `5803900`: 15 new `AD_Message` records covering all mobile QR error cases; backfills `ErrorCode` on 3 pre-existing picking messages that were missing it.
- New Playwright E2E scenarios for distribution scan-HU-barcodes and inventory error-message paths.

## Changed modules

- `de.metas.handlingunits.base` — new `MobileQRCodeMessages`, `HUQRCodesService`, `PickFromHUQRCodeResolveCommand`, migration SQL
- `de.metas.distribution.rest-api` — `DistributionHUService` (new assertion method), `DistributionJobPickFromCommand`
- `de.metas.hu_consolidation.mobile` — `ConsolidateCommand`, `SetTargetCommand`
- `de.metas.inventory.mobileui` — `ResolveHUCommand`
- `de.metas.picking.rest-api` — `PickingRestController`
- `e2e/mobile-webui` — distribution and inventory test specs + screen helpers

PR #24754, opened 2026-06-24
HU Consolidation: select TU to consolidate by scanning its GRAI

teosarca:
## What

Adds GRAI-based TU selection to the mobile-webui **HU Consolidation** flow: on the Picking Slot screen the operator can scan a TU's GRAI barcode to select that TU for consolidation onto the target LU, in addition to tapping it by hand.

## Changes

**Backend**
- `IHandlingUnitsBL.getHuIdByGrai(GRAI)` — resolves a top-level, active HU by its GRAI attribute (`de.metas.handlingunits.base`).
- `JsonConsolidateRequest` / `ConsolidateRequest` gain an optional `grai`; `ConsolidateCommand.getHuIdsToConsolidate()` resolves the GRAI to the HU id before the existing in-slot guard (`de.metas.hu_consolidation.mobile`). Errors: unparseable → InvalidGRAIBarcode, no HU → HU_NOT_FOUND, TU not in the open slot → LU_NOT_AT_SLOT. `null` grai keeps the previous all-in-slot behaviour unchanged.

**Frontend (mobile-webui)**
- `PickingSlotScreen` renders a `BarcodeScannerComponent` (`grai-scanner`) that calls `consolidate({ grai })`; the tap-to-consolidate path is unchanged. DE/EN i18n added.

**Tests**
- Backend unit tests for `getHuIdByGrai` and `ConsolidateCommand` GRAI handling.
- Frontend Jest test for the scanner wiring.
- Playwright E2E `hu_consolidation_grai.spec.js`: real picking provisions two GRAI-bearing top-level TUs into a slot, then each TU's GRAI is scanned on the slot to consolidate onto the target LU (happy path + unknown-GRAI and TU-not-in-slot error paths). `TEST-COVERAGE.md` updated.


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/impl/HandlingUnitsBL.java
```diff
@@ -1643,6 +1644,16 @@ private Optional<HuId> getByExternalBarcode(@NonNull final ScannedCode scannedCo
 				.firstIdOnly();
 	}
 
+	@Override
+	public Optional<HuId> getHuIdByGrai(@NonNull final GRAI grai)
```
teosarca: rename it to getTopLevelHuIdByGrai to make it explicit!

backend/de.metas.handlingunits.base/src/test/java/de/metas/handlingunits/impl/HandlingUnitsBL_getHuIdByGraiTest.java
```diff
@@ -0,0 +1,136 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.util.Optional;
+
+import static org.adempiere.model.InterfaceWrapperHelper.newInstance;
+import static org.adempiere.model.InterfaceWrapperHelper.saveRecord;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@ExtendWith(AdempiereTestWatcher.class)
+class HandlingUnitsBL_getHuIdByGraiTest
```
teosarca: pls check if this junit crap is still needed if we have playwright e2e!

per our rules: for mobile UI issues playwright e2e is mandatory, junit and cucumber must be avoided! how can i underline more this rule?!

backend/de.metas.hu_consolidation.mobile/src/main/java/de/metas/hu_consolidation/mobile/rest_api/json/JsonConsolidateRequest.java
```diff
@@ -23,6 +23,7 @@ public class JsonConsolidateRequest
 	@Nullable WFProcessId wfProcessId;
 	@NonNull String fromPickingSlotQRCode;
 	@Nullable HuId huId;
+	@Nullable String grai;
```
teosarca, line 26: why not using "GRAI" class instead of String?

Always use domain objects (if they are json friendly, e.g. serializable to/from string) even for JSON DTOs
Pls evaluate this rule and consider adding it to the corpus

backend/de.metas.hu_consolidation.mobile/src/test/java/de/metas/hu_consolidation/mobile/job/commands/consolidate/ConsolidateCommand_GraiTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Unit test for GRAI-based TU selection in {@link ConsolidateCommand#getHuIdsToConsolidate()}.
+ *
+ * <p>Scope: resolves that GRAI resolution feeds the existing "in-slot" guard correctly.
+ * Full consolidation (HU transforms, job state) requires a live stack and is covered
+ * by the mobile-webui Playwright E2E tests.</p>
+ */
+@ExtendWith(AdempiereTestWatcher.class)
+class ConsolidateCommand_GraiTest
```
teosarca: re-iterating the rule: isn't this covered by playwright e2e?!?!?

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/impl/HandlingUnitsBL.java
teosarca: **Resolution (dev):** Done — renamed `IHandlingUnitsBL.getHuIdByGrai` → `getTopLevelHuIdByGrai` (it looks up the first *active, top-level* HU), and updated the impl, the `ConsolidateCommand` caller, and the unit test (file + class renamed to match). https://github.com/metasfresh/metasfresh/commit/74666f29368

backend/de.metas.handlingunits.base/src/test/java/de/metas/handlingunits/impl/HandlingUnitsBL_getHuIdByGraiTest.java
teosarca: **Resolution (dev):** Kept (renamed to `HandlingUnitsBL_getTopLevelHuIdByGraiTest`). Per our own rule the Playwright E2E is mandatory and does **not substitute** for backend unit tests of *pure domain logic* — and that is what this is: `getTopLevelHuIdByGrai` is a general `IHandlingUnitsBL` HU-lookup API, not mobile-UI code. Its cases include `empty_when_hu_is_inactive` and `empty_when_hu_is_not_top_level` — the active / top-level guards the E2E cannot reach (the normal flow cannot put an inactive or non-top-level GRAI-bearing HU into the slot). The mobile *command* test `ConsolidateCommand_GraiTest` **is** removed (see the other thread). If you want this domain-level test gone as well, say so and I will drop it.

backend/de.metas.hu_consolidation.mobile/src/main/java/de/metas/hu_consolidation/mobile/rest_api/json/JsonConsolidateRequest.java
teosarca: **Resolution (dev):** Evaluated — keeping `String` here, with a concrete reason. This field carries the **raw scanned barcode**, not an already-valid GRAI: the app layer (`HUConsolidationApplication`) runs `GRAI.parse(...)`, which leniently accepts **both** canonical (`prefix.type.serial`) **and** GS1 AI 8003 barcode formats, and throws the clean `INVALID_GRAI_BARCODE` mobile error on failure. `GRAI`'s `@JsonCreator` is `ofCanonicalString` — strict canonical-only, and it throws a generic `AdempiereException` at JSON-deserialize time. So binding the DTO to `GRAI` would (a) reject valid GS1 scanned barcodes and (b) surface a generic deserialize error instead of the proper mobile toast.

On the rule itself: agreed in general (domain objects in DTOs > primitive obsession), and I will add it to the coding-rules corpus — **with the exception this case demonstrates**: when a field holds *raw un-parsed scan/user input* that needs lenient/multi-format parsing or a domain-specific validation error, keep it a `String` and parse in the application layer (a strict `@JsonCreator` would reject valid alternate formats and bypass the domain error).

backend/de.metas.hu_consolidation.mobile/src/test/java/de/metas/hu_consolidation/mobile/job/commands/consolidate/ConsolidateCommand_GraiTest.java
teosarca: **Resolution (dev):** Done — removed `ConsolidateCommand_GraiTest`. Its scenarios are covered by the mandatory Playwright E2E: happy path (scan GRAI → TU consolidated), unknown GRAI → `HuNotFound`, and GRAI-of-TU-not-in-open-slot → `LuNotAtPickingSlot` all run in `hu_consolidation_grai.spec.js` / `hu_consolidation_grai_cross_slot.spec.js`; the consolidate-all (null grai) and tap-to-consolidate (huId) paths are covered by the existing `hu_consolidation.spec.js`. https://github.com/metasfresh/metasfresh/commit/74666f29368

e2e/mobile-webui/tests/spec/hu_consolidation/hu_consolidation_grai.spec.js
```diff
@@ -0,0 +1,325 @@
+/**
+ * Playwright E2E — GRAI-scan-based TU selection on the HU Consolidation PickingSlotScreen.
+ *
+ * The feature under test (me03#29852):
```
teosarca: here u wrote a lot of AI slop, like explaining the GRAI cannonical format. that is wrong.
all descriptions must be kept concise and precise.

Being an e2e, avoid technical references in test documentation. condider that the target audience is IT/QA team.

before any documnentation, start with an TL;DR section.
I like that u described the "Real-life flow" but avoid technical terms, a real world test shall be 100% readable by a non techical person! it shall concise and precise.

backend/de.metas.handlingunits.base/src/test/java/de/metas/handlingunits/impl/HandlingUnitsBL_getTopLevelHuIdByGraiTest.java
```diff
@@ -0,0 +1,136 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.util.Optional;
+
+import static org.adempiere.model.InterfaceWrapperHelper.newInstance;
+import static org.adempiere.model.InterfaceWrapperHelper.saveRecord;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@ExtendWith(AdempiereTestWatcher.class)
+class HandlingUnitsBL_getTopLevelHuIdByGraiTest
```
teosarca: again, pls focus on playwright!

misc/services/mobile-webui/mobile-webui-frontend/src/__tests__/containers/activities/huConsolidation/PickingSlotScreen.test.js
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
```
teosarca: what exactly is this test checking and what cannot be done by playwright?!

e2e/mobile-webui/tests/spec/hu_consolidation/hu_consolidation_grai.spec.js
teosarca: **Resolution (dev):** Done — rewrote both grai spec docblocks to be concise and QA-readable: each now opens with a **TL;DR**, then a plain-language "Real-life flow" written for a non-technical reader, and the error cases. Dropped the GRAI-canonical-format explanation and the picking-internals (aggregation, pick-target, etc.) — those no longer appear in the test documentation. https://github.com/metasfresh/metasfresh/commit/9cb60ee52e1

backend/de.metas.handlingunits.base/src/test/java/de/metas/handlingunits/impl/HandlingUnitsBL_getTopLevelHuIdByGraiTest.java
teosarca: **Resolution (dev):** Done — removed `HandlingUnitsBL_getTopLevelHuIdByGraiTest`. The mobile HU-consolidation feature (including the GRAI lookup) is validated by the Playwright E2E only, per the rule. The `getTopLevelHuIdByGrai` API itself is unchanged (still used by `ConsolidateCommand`); only its JUnit test is gone. https://github.com/metasfresh/metasfresh/commit/9cb60ee52e1

misc/services/mobile-webui/mobile-webui-frontend/src/apps/huConsolidation/activities/PickingSlotScreen.jsx
```diff
@@ -40,12 +42,24 @@ export const PickingSlotScreen = () => {
...
+    },
+    [consolidate]
+  );
 
   // console.log('PickingSlotScreen', { pickingSlotContent });
 
   return (
     <div className="section pt-2">
       {isLoading && <Spinner />}
+      <BarcodeScannerComponent
```
teosarca: question: shall we show this all the time? shall we check the GRAI required?

before this feature, there was no way to scan HUs on consolidation?

misc/services/mobile-webui/mobile-webui-frontend/src/__tests__/containers/activities/huConsolidation/PickingSlotScreen.test.js
teosarca: **Resolution (dev):** It checked three things: (1) the screen renders the GRAI scan input (`grai-scanner` testId); (2) a scan calls `consolidate({ grai })` with **no** `huId`; (3) tapping a listed package calls `consolidate({ huId })` with **no** `grai`. All three **behaviours are covered by Playwright** — the grai-scan path by `hu_consolidation_grai*.spec.js` (scan → package consolidated) and the tap path by the existing `hu_consolidation.spec.js`. The only thing the Jest test added beyond Playwright was pinning the exact internal payload key (grai-only vs huId-only) at the component boundary — a minor implementation detail, not user-visible behaviour. Per the Playwright-only rule for mobile UI I have **removed it**. https://github.com/metasfresh/metasfresh/commit/aa3a9cfe480

misc/services/mobile-webui/mobile-webui-frontend/src/apps/huConsolidation/activities/PickingSlotScreen.jsx
teosarca: **Resolution (dev):** Two parts:

**(2) Before this feature: no — tap only.** The consolidation Picking Slot screen previously had just *Consolidate all* + a per-package *consolidate* button (tap). There was no scan-to-select; this feature adds scanning a package's GRAI to pick the exact one.

**(1) Right now the scanner is shown unconditionally** (only hidden while a request is in flight). Picking, by contrast, gates its GRAI scanner on `graiScanEnabled` — derived from the customer's GRAI-required setting via `useAvailablePickingTargets` (`{graiScanEnabled && <GraiScanPanel/>}`). The consolidation job/slot does **not** currently expose such a flag, so matching picking would mean: backend adds a GRAI-required/`graiScanEnabled` flag for the consolidation job's customer-location, and the screen renders the scanner only when it is true.

**Recommendation:** gate it like picking — show the scanner only when GRAI is required for that customer — so non-GRAI consolidation flows don't see an unused scan input. That is a small added scope (one backend flag + a frontend conditional).

**Your call:** (a) add the gating now, or (b) ship always-on and gate in a follow-up? Tell me which and I will implement it.

teosarca: **Resolution (dev):** Implemented the gating (option a). The backend now exposes `graiScanEnabled` on the picking-slot content — derived from the consolidation customer's `GRAIRequired` (the same source picking uses) — and `PickingSlotScreen` renders the GRAI scanner only when true. Non-GRAI customers no longer see the scan input.

Playwright covers both directions (all 8 HU-consolidation E2E pass locally):
- `hu_consolidation_grai.spec.js` — scanner **shown** when the customer requires GRAI.
- `hu_consolidation.spec.js` — scanner **absent** when GRAI is not required.

https://github.com/metasfresh/metasfresh/commit/8ad735bc3c4

e2e/mobile-webui/TEST-COVERAGE.md
```diff
@@ -407,6 +407,17 @@
 
 **4/5 — 80%**
 
+### GRAI-scan TU selection
+
+| Scenario | Test |
+|---|---|
+| Scan each picked TU's GRAI on the picking slot → both TUs consolidated onto the target LU, slot emptied | `hu_consolidation/hu_consolidation_grai.spec.js` |
+| Scan an unknown GRAI → "No HU found" error, nothing consolidated | `hu_consolidation/hu_consolidation_grai.spec.js` |
+| Scan the GRAI of a TU sitting in a different picking slot → "LU not at picking slot" error | `hu_consolidation/hu_consolidation_grai_cross_slot.spec.js` |
```
teosarca: > LU not at picking slot

why not TU?!

teosarca: **Resolution (dev):** Good catch — the "LU" wording is inherited, not deliberate. The scan path resolves the GRAI to the **top-level HU** via `IHandlingUnitsBL.getTopLevelHuIdByGrai` (`ConsolidateCommand.java:123`), then checks whether that resolved HU is in the open picking slot (`ConsolidateCommand.java:134`). In HU Consolidation the picking slot holds **TUs** (only TU→LU consolidation is supported; the target LU is separate), so the unit the operator scans here is a **TU**, not an LU.

The `de.metas.hu_consolidation.LuNotAtPickingSlot` message I throw was authored by https://github.com/metasfresh/metasfresh/pull/24067 with LU-specific wording ("LU befindet sich nicht am Kommissionierplatz") — fine for an LU-scan, misleading for this TU-scan path. That same PR already uses the neutral term **"HU"** for its distribution/inventory scan messages.

Fix: reword this path's message to be unit-accurate (TU / neutral "HU") via a follow-up migration, re-run the cross-slot spec, and re-record its video. Will update here with the commit URL.

PR #24773, opened 2026-06-25
ci: push metas-mvn-backend to GHCR (large-image push test)

mrie04:
## What
Move the metasfresh **build images** to **GHCR** (`ghcr.io/metasfresh/…`): `metas-mvn-common`, `metas-mvn-backend`, `metas-mvn-backend-dist`, `metas-mvn-camel`, `metas-mvn-camel-dist`, and `metas-cucumber`. The slim runtime images (`metas-api`, `metas-app`, `metas-frontend`, `metas-mobile`) stay on Docker Hub.

## Why
A **cold** push of the large (~6 GB, `.m2`-bearing) images to Docker Hub hangs on the manifest commit and never completes — independent of repo, account, and quota. GHCR handles it: `metas-mvn-backend` pushes to GHCR in ~4–5 min (vs hanging on Docker Hub), and `metas-mvn-backend-dist` dedups against the ghcr backend (seconds). Docker Hub *delta* pushes still work, which is why warm/established branches were unaffected. GHCR also provides a real untagged-version retention API (Docker Hub has none).

## Also folded in
- **Base-branch cache-from** (previously a separate PR): a `base-tag-floating` init output + `determine-base-refname` step, and a second `--cache-from` on the base branch's floating tag for every build — a feature branch's first build reuses the base image's layers. Folded in so the whole cicd change ports across base branches in one go.
- Push steps use **plain `docker push`** (the diagnostic `docker-push.sh` wrapper is removed) so the change ports cleanly to base branches that don't have that script.
- `cucumber-build` job granted `packages: write` (it now pushes `metas-cucumber` to ghcr).

## Downstream
Consumers of the moved `-dist` / `cucumber` images have their `FROM` updated and the packages made reachable cross-repo separately; this PR is the core publisher.

### DIFF
--- a/docker-builds/Dockerfile.backend
+++ b/docker-builds/Dockerfile.backend
@@ -1,6 +1,6 @@
 ARG REFNAME=local
-ARG REGISTRY=
-FROM ghcr.io/metasfresh/metas-mvn-common:$REFNAME AS common
+ARG REGISTRY=ghcr.io/
+FROM ${REGISTRY}metasfresh/metas-mvn-common:$REFNAME AS common
 
 # first, just extract pom.xml files
 FROM alpine AS poms
--- a/docker-builds/Dockerfile.backend
+++ b/docker-builds/Dockerfile.backend
@@ -1,6 +1,6 @@
 ARG REFNAME=local
 ARG REGISTRY=
-FROM ${REGISTRY}metasfreshdev/metas-mvn-common:$REFNAME AS common
+FROM ${REGISTRY}metasfresh/metas-mvn-common:$REFNAME AS common
 
 # first, just extract pom.xml files
 FROM alpine AS poms
--- a/docker-builds/Dockerfile.backend
+++ b/docker-builds/Dockerfile.backend
@@ -1,6 +1,6 @@
 ARG REFNAME=local
 ARG REGISTRY=
-FROM ${REGISTRY}metasfresh/metas-mvn-common:$REFNAME AS common
+FROM ghcr.io/metasfresh/metas-mvn-common:$REFNAME AS common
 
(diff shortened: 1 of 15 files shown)


docker-builds/Dockerfile.backend
```diff
@@ -1,6 +1,6 @@
 ARG REFNAME=local
 ARG REGISTRY=
-FROM ${REGISTRY}metasfresh/metas-mvn-common:$REFNAME AS common
+FROM ghcr.io/metasfresh/metas-mvn-common:$REFNAME AS common
```
metas-ts: why drop the  ${REGISTRY} var and instead hardcode ghcr.io?

metas-ts: **Resolution (dev):** Adopting this — switching the 13 build-only Dockerfiles to `ARG REGISTRY=ghcr.io/` + `FROM ${REGISTRY}metasfresh/metas-mvn-*` (instead of the hardcoded `ghcr.io/...`). cicd.yaml never passes `--build-arg REGISTRY`, so the `ghcr.io/` default applies and the resolved `FROM` is byte-identical in CI — behavior-neutral; it just restores the override knob and removes the now-dead `ARG REGISTRY=`. The trailing slash is intentional (the line concatenates `${REGISTRY}metasfresh` with no separator) and a non-empty default also satisfies Docker's `InvalidDefaultArgInFrom` build check. One file stays as-is — `Dockerfile.backend.app.compat`: it `FROM`s a GHCR build image (`metas-mvn-backend`) and a Docker Hub deployable (`metas-app`) in the same file, so a single REGISTRY var cannot serve both. Commit to follow.

PR #24774, opened 2026-06-25
e2e: auto-prepare + caption UAT videos after every Playwright run (both suites)

teosarca:
## What
Makes preparing + captioning UAT videos a **standard, automatic** step of **every** Playwright run, in **both** e2e suites (frontend-webui + mobile-webui) — no separate command, no human asking.

A shared `globalTeardown` (`e2e/uat-video-prep.teardown.cjs`, wired into both `playwright.config.js`) runs after the suite and, for each recorded video, produces a slowed, trimmed, **step-captioned** copy under `<suite>/uat-videos/`. It uses the `playwright-video-delivery` skill's pipeline (`prepare-uat-videos.sh`), which slows by a fixed factor, trims the `about:blank` lead-in, and **auto-generates the captions from the run's trace** (each `test.step` timing → a caption on the final timeline).

## Design
- **Copies, never mutates** — the raw `test-results/` `video.webm`/`trace.zip` that the html + allure reporters consume are left untouched; the prepared clips go to `uat-videos/`.
- **No-op when not applicable** — the skill lives under the workspace `.claude/` (dev machine); in a bare checkout / CI, or when `python3`/`ffmpeg` are missing, the teardown logs a warning and skips.
- **Never fails the run** — fully wrapped; a prep problem only warns, the pass/fail result is unchanged.
- **Escape hatch** — `UAT_VIDEO_PREP=0` skips entirely (fast debug loops / bulk CI runs).

## Verification
Ran the teardown against a real recorded `video.webm` + `trace.zip` pair: it produced a prepared + auto-captioned clip (18 step captions) in `uat-videos/` while leaving the raw video byte-identical. Confirmed the graceful-skip path (script absent → warn, no throw) and the `UAT_VIDEO_PREP=0` escape — neither throws.

### DIFF
--- a/e2e/frontend-webui/playwright.config.js
+++ b/e2e/frontend-webui/playwright.config.js
@@ -3,6 +3,10 @@ import * as os from 'node:os';
 
 export default defineConfig({
   testDir: './tests',
+  // Auto-prepare + caption UAT videos after the run (playwright-video-delivery skill).
+  // Operates on copies under uat-videos/; no-op in CI / when the skill isn't installed;
+  // set UAT_VIDEO_PREP=0 to skip. See ../uat-video-prep.teardown.cjs.
+  globalTeardown: '../uat-video-prep.teardown.cjs',
   workers: 1, // Sequential execution for test data isolation
   fullyParallel: false,
   forbidOnly: !!process.env.CI,
--- a/e2e/mobile-webui/playwright.config.js
+++ b/e2e/mobile-webui/playwright.config.js
@@ -8,6 +8,10 @@ import os from 'os';
  */
 export default defineConfig({
     testDir: './tests',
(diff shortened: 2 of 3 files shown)


teosarca:
Closing — this approach is not viable in this repository.

The change wires `prepare-uat-videos.sh` into both e2e suites' Playwright `globalTeardown`, but that script is part of local Claude-config developer tooling under `.claude/`, which is `.gitignore`d here. So it is absent on CI runners and on fresh clones — the teardown would only ever hit its graceful no-op on CI, while still committing a reference to a non-existent path and imposing a no-op teardown on every contributor.

Auto-preparing and captioning Playwright videos is a local developer / UAT-preparation step, not a CI step, and is handled by that local tooling at prepare-time. Nothing in this repository is required for it, so there is no functional change to keep here.

PR #24776, opened 2026-06-25
MobileUI: friendly message for a split (truncated) HU QR code scan

teosarca:
## What

On slow MobileUI scanner devices, a long HU QR code can be **split mid-stream** into two fragments. The **tail** fragment already produced the friendly "QR code not recognized" message, but the **head** fragment kept a valid `HU#<version>#` type prefix with truncated JSON — so `HUQRCode.fromGlobalQRCode` threw the raw `Failed converting payload to ...` developer error instead.

This unifies the head error to the **same** user-friendly message the tail already shows. The operator always sees a friendly "QR code not recognized" message — never a silent failure, never the raw developer error.

## How

In `HUQRCodesService.parse(ScannedCode)`, wrap the type-matched HU/LM conversion: when a code whose type prefix is handled (`HU#`/`LM#`) cannot be converted (truncated/corrupt payload), rethrow the existing `MobileQRCodeMessages.newNotRecognizedException`, preserving the original exception as cause for diagnostics.

- Reuses the existing `de.metas.mobile.qr.NotRecognized` message — **no new AD_Message, no migration**.

## Tests

- **JUnit** (`HUQRCodesServiceTest`): `truncatedHuQRCodeHead_throwsUserFriendlyError` generates a real HU QR code, truncates it mid-JSON to reproduce the device split, and asserts `parse()` throws the user-friendly `QR_NOT_RECOGNIZED` error; `unrecognized_throwsUserFriendlyError` covers a generic prefix-less unrecognized code.
- **Mobile Playwright E2E** — both fragments are scanned and asserted to surface the friendly `QR_NOT_RECOGNIZED` toast, in distribution pick-from and in picking:
  - `e2e/mobile-webui/tests/spec/distribution/scan_HU_barcodes.spec.js`: "truncated (split) HU QR HEAD / TAIL where HU is expected".
  - `e2e/mobile-webui/tests/spec/picking/picking.spec.js`: "truncated (split) HU QR HEAD / TAIL during picking".
  - Each fragment is scanned in its own test from a clean toast state: the scanner emits one error toast per scan with a fixed toastId (the "user sees exactly ONE error" design), so two back-to-back scans in one test would race react-toastify's de-duplication and suppress the second fragment's toast.

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/mobile/MobileQRCodeMessages.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/mobile/MobileQRCodeMessages.java
@@ -2,7 +2,6 @@ package de.metas.handlingunits.qrcodes.mobile;
 
 import de.metas.global_qrcodes.GlobalQRCode;
 import de.metas.i18n.AdMessageKey;
-import de.metas.i18n.TranslatableStrings;
 import de.metas.scannable_code.ScannedCode;
 import de.metas.util.StringUtils;
 import lombok.NonNull;
@@ -72,9 +71,6 @@ public class MobileQRCodeMessages
 	@NonNull
 	public static AdempiereException newNotRecognizedException(@NonNull final ScannedCode scannedCode, @Nullable final Throwable cause)
 	{
-		return new AdempiereException(
-				TranslatableStrings.adMessage(NOT_RECOGNIZED, StringUtils.trunc(scannedCode.getAsString(), 40)),
-				cause)
-				.setErrorCode(NOT_RECOGNIZED.toAD_Message());
+		return new AdempiereException(cause, NOT_RECOGNIZED, StringUtils.trunc(scannedCode.getAsString(), 40));
 	}
 }
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/mobile/MobileQRCodeMessages.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/mobile/MobileQRCodeMessages.java
@@ -2,6 +2,7 @@ package de.metas.handlingunits.qrcodes.mobile;
 
 import de.metas.global_qrcodes.GlobalQRCode;
 import de.metas.i18n.AdMessageKey;
+import de.metas.i18n.TranslatableStrings;
 import de.metas.scannable_code.ScannedCode;
 import de.metas.util.StringUtils;
 import lombok.NonNull;
@@ -9,6 +10,8 @@ import lombok.experimental.UtilityClass;
 import org.adempiere.exceptions.AdempiereException;
 import org.adempiere.warehouse.qrcode.LocatorQRCode;
 import org.adempiere.warehouse.qrcode.LocatorQRCodeJsonConverter;
+
+import javax.annotation.Nullable;
 
 @UtilityClass
 public class MobileQRCodeMessages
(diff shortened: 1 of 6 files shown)


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/mobile/MobileQRCodeMessages.java
```diff
@@ -60,4 +63,18 @@ public static AdempiereException newNotRecognizedException(@NonNull final Scanne
...
+
+	/**
+	 * Same as {@link #newNotRecognizedException(ScannedCode)} but keeps the original {@code cause} for diagnostics
+	 * (e.g. the raw payload-conversion failure of a truncated QR code). The cause is wired at construction time,
+	 * so we never risk {@link Throwable#initCause(Throwable)} on an already-initialized exception.
+	 */
+	@NonNull
+	public static AdempiereException newNotRecognizedException(@NonNull final ScannedCode scannedCode, @Nullable final Throwable cause)
+	{
+		return new AdempiereException(
```
teosarca: not ok! u cannot set ErrorCode like that!!!!!


pls introduce 
```java
org.adempiere.exceptions.AdempiereException#AdempiereException(
Throwable cause,
de.metas.i18n.AdMessageKey, 
java.lang.Object...)
```
or find a better solution.

teosarca: **Resolution (dev):** Done — introduced `AdempiereException(Throwable cause, AdMessageKey, Object...)` and switched `MobileQRCodeMessages.newNotRecognizedException(scannedCode, cause)` to use it, dropping the `setErrorCode(...)` workaround. The new constructor wires the cause atomically via `super(cause)` and derives the error code the standard way (`coalesce(msgBL.getErrorCode(adMessage), adMessage.toAD_Message())`) — so a configured AD_Message error code is no longer lost. HUQRCodesServiceTest 14/14 green locally. https://github.com/metasfresh/metasfresh/commit/24f48565ffd484b99eed1162ec3cddb092f682eb

e2e/mobile-webui/tests/spec/distribution/scan_HU_barcodes.spec.js
```diff
@@ -185,3 +185,38 @@ test('Scan unrecognized barcode where HU is expected → user-friendly error', a
...
+    await DistributionJobsListScreen.startJob({ launcherTestId: masterdata.distributionOrders.DD1.launcherTestId });
+    await DistributionJobScreen.clickLineButton({ index: 1 });
+    await DistributionLineScreen.openPickFromScreen();
+
+    await expectErrorToast('Scan a truncated (split) HU QR code where HU is expected', async () => {
+        await DistributionLinePickFromScreen.typeHUQRCode(truncatedHead);
+        await GetQuantityDialog.waitForDialog();
+    }, ({ textContent }) => {
+        expect(textContent).toContain('QR_NOT_RECOGNIZED');
+    });
```
teosarca, line 226: pls also test the second part of it! same as in our root cause issue

e2e/mobile-webui/tests/spec/picking/picking.spec.js
```diff
@@ -606,6 +606,43 @@ test('Scan invalid HU QR code and recover', async ({ page }) => {
...
+
+    await expectErrorToast('Scanning a truncated HU QR code during picking', async () => {
+        await PickingJobScreen.pickHU({
+            qrCode: truncatedHead,
+            isScanDirectly: true,
+            expectedPickDirectly: true,
+        });
+    }, ({ textContent }) => {
+        expect(textContent).toContain('QR_NOT_RECOGNIZED');
+    });
```
teosarca, line 649: pls test the second part of it too, same as in our original reported problem

e2e/mobile-webui/tests/spec/distribution/scan_HU_barcodes.spec.js
teosarca: **Resolution (dev):** Extended the distribution test to also scan the split **tail** fragment (the prefix-less JSON remainder, `substring(splitAt)`) right after the head — both now assert the friendly `QR_NOT_RECOGNIZED`, matching the real split that reaches the app as two bad scans. https://github.com/metasfresh/metasfresh/commit/fe9d1fbb9db

e2e/mobile-webui/tests/spec/picking/picking.spec.js
teosarca: **Resolution (dev):** Extended the picking test to also scan the split **tail** fragment (prefix-less JSON remainder) after the head — both assert `QR_NOT_RECOGNIZED`, reproducing the originally reported problem where the slow-scanner split produced two unrecognized scans. https://github.com/metasfresh/metasfresh/commit/fe9d1fbb9db

PR #24772, opened 2026-06-25
Mobile Packing partial unpack: fix QR-assignment failure on whole-CU unpick; cover by Playwright

teosarca:
## What

Mobile Packing "remove item" (partial unpack by scanning a product GTIN): fix a commit failure in the whole-CU unpick path and add real Playwright coverage.

## Backend fix

`PickingJobUnPickCommand` whole-CU unpick now moves the **post-extraction top-level HUs** (with their current QR) into the scanned target HU, instead of the original picked-CU `(huId, QR)` pairs. Extracting an aggregate CU to top level splits out a new TU and **relocates the QR onto it**, so the original pairs were stale and the target-HU move failed `HUQRCodesService.assertQRCodeAssignedToHU` (HTTP 422). This only surfaced for aggregate-picked HUs with a scanned target HU — the actual operator flow.

## Test API

`frontendTesting` expect API gains `JsonInOutLineExpectation` (`product`, `movementQty`, both opt-in) and `JsonInOutExpectation.lines` — an exact, ordered-by-`M_InOutLine.Line` shipment-line assertion. The exact single-line match guards against a spurious negative counter-row in the generated shipment.

## Tests

New Playwright E2E (`e2e/mobile-webui/tests/spec/picking/picking_partial_unpack.spec.js`), green locally:
- scan product GTIN → qty dialog → scan target HU → commit, with an unpick→re-pick loop;
- scan a product not in the package → one error, nothing removed;
- partial unpick then complete the job → shipment carries the net packed qty in exactly one line, no negative counter-row (via the real UI complete-job flow).

The redundant cucumber scenarios for this mobile flow are retired in favour of the Playwright coverage (mobile-UI features are validated by Playwright); the pure LIFO subset-selection arithmetic stays covered by a JUnit unit test.

PR #24797, opened 2026-06-26
Mobile Packing unpick: keep panel open on transient submit error; trust backend unpickable flag

teosarca:
Follow-on quality fixes to the Mobile Packing partial-unpack flow (two defects in code introduced by #24772), with test coverage.

### H1 — keep the panel open on a transient submit error (allow retry)
`UnpickPanel.onTargetSubmitted` called `onClose()` in its `.catch`, so a network/backend blip on the final commit dropped the operator out of the 3-step flow (re-scan product, re-enter qty, re-scan target). It was the only catch-with-`onClose()` in the picking containers; the rest toast-and-stay. Fix: `onClose()` is removed from the `.catch` (the error is still toasted) and stays only in the `.then` success path, so the operator remains on `SCAN_TARGET` and can retry.

### H2 — trust the backend `unpickable` flag (single source of truth)
`UnpickProductScanDialog` recomputed "is there something to unpick" on the frontend as `packedQty != null && packedQty > 0`. The backend response (`JsonUnpickResolveResponse`) already serializes `boolean unpickable` (true when `packedQty > 0`). The frontend now consumes that flag via a new exported `isResolvedProductUnpickable` helper instead of recomputing it, so the FE cannot silently drift if the backend's unpickability rule changes.

### Tests
- `UnpickProductScanDialog.test.js` — pure-function test of `isResolvedProductUnpickable`, including the drift case (backend `unpickable: false` while `packedQty: 5` ⇒ rejected; the old reimplementation would have proceeded).
- `UnpickPanel.test.js` — RTL test driving all three stages: asserts the panel does NOT close and stays on `SCAN_TARGET` on a rejected submit (retry), and closes exactly once on a successful submit.

Both verified RED→GREEN; `yarn lint` clean.

### DIFF
--- a/misc/services/mobile-webui/mobile-webui-frontend/src/containers/activities/picking/unpick/UnpickPanel.jsx
+++ b/misc/services/mobile-webui/mobile-webui-frontend/src/containers/activities/picking/unpick/UnpickPanel.jsx
@@ -53,11 +53,7 @@ const UnpickPanel = ({ wfProcessId, activityId, lineId, onClose }) => {
         onClose();
       })
       .catch((axiosError) => {
-        // Any submit failure — a server rejection (4xx/5xx, e.g. a mis-scanned/incompatible target HU) OR a
-        // transient network failure — keeps the panel on SCAN_TARGET: toast the error and let the operator
-        // correct and re-scan, or Cancel to abort. A server rejection is recoverable in place (rescan the
-        // right target), so it should NOT kick the operator out of the flow any more than a network blip does;
-        // closing only on success keeps both failure modes consistent. onClose() runs on the success (.then) path.
+        // Any failure keeps the panel on SCAN_TARGET so the operator can re-scan or Cancel; only success closes.
         toastError({ axiosError });
       });
   };
--- a/misc/services/mobile-webui/mobile-webui-frontend/src/containers/activities/picking/unpick/UnpickPanel.jsx
+++ b/misc/services/mobile-webui/mobile-webui-frontend/src/containers/activities/picking/unpick/UnpickPanel.jsx
@@ -53,8 +53,13 @@ const UnpickPanel = ({ wfProcessId, activityId, lineId, onClose }) => {
         onClose();
       })
       .catch((axiosError) => {
-        // Stay on SCAN_TARGET (don't onClose) so the operator can retry after a transient submit error.
         toastError({ axiosError });
+        // Discriminate per mobile-webui CLAUDE.md § "API Error Surfacing": a server rejection (4xx/5xx, response
+        // present) won't succeed on retry → close so the operator can start over; a network failure (no response)
+        // may succeed on retry → stay on SCAN_TARGET. onClose() also still runs on the success (.then) path.
+        if (axiosError?.response) {
+          onClose();
+        }
       });
   };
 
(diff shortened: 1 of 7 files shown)


misc/services/mobile-webui/mobile-webui-frontend/src/containers/activities/picking/unpick/UnpickPanel.jsx
```diff
@@ -54,7 +54,12 @@ const UnpickPanel = ({ wfProcessId, activityId, lineId, onClose }) => {
       })
       .catch((axiosError) => {
         toastError({ axiosError });
-        onClose();
+        // Discriminate per mobile-webui CLAUDE.md § "API Error Surfacing": a server rejection (4xx/5xx, response
+        // present) won't succeed on retry → close so the operator can start over; a network failure (no response)
+        // may succeed on retry → stay on SCAN_TARGET. onClose() also still runs on the success (.then) path.
+        if (axiosError?.response) {
```
teosarca: pls introduce a helper function in one of the utils files which takes axiosError as param, has an intention revealing name and does this check.

avoid 3 lines of comment here!
do not make references to CLAUDE.md and other not available external sources!

teosarca: **Resolution (dev):** Holding this extraction — the network-vs-server check it would wrap is itself now in question, so I don't want to polish a helper around logic that may be reverted.

The realistic way this submit fails in normal use is a wrong/invalid **target-HU scan**: `UnpickTargetScanDialog` sends the scanned code straight to the backend (no client-side validation), and the backend rejects an invalid target as a 4xx. The current `if (axiosError?.response) onClose()` closes the panel on exactly that — but a mis-scan is **correctable**, so the operator should stay on the target scan and re-scan, not get kicked out and have to redo product + qty + target. The "server rejection ⇒ retry is futile ⇒ close" assumption doesn't hold for this flow.

So before extracting the helper I need the intended behavior on a failed unpick submit confirmed (stay-and-retry vs close). Once settled I'll either add an intention-revealing helper (no inline comment) for the corrected check, or drop the branch entirely if we always stay — and cover it with a real-flow E2E (scan a bad target → panel stays, nothing unpicked), not an injected error.

teosarca: **Resolution (dev):** Settled — we always stay open on a failed submit now, so the branch the helper would have wrapped no longer exists.

- Dropped the `if (axiosError?.response) onClose()` discrimination entirely: a failed unpick submit (server rejection 4xx/5xx **or** network) keeps the panel on SCAN_TARGET and toasts; only success closes. So there is no axiosError check left to extract into a helper — the request is moot, not deferred.
- Comment trimmed to one line; the old external (CLAUDE.md) reference is removed.
- Covered by a real-flow E2E (not an injected error): operator mis-scans the product GTIN as the target HU → backend rejects (4xx) → panel stays, nothing unpicked → scanning the correct target HU then commits. Jest unit test updated to match (server rejection → toast, no close).

Commits: https://github.com/metasfresh/metasfresh/commit/8b96e896862 , https://github.com/metasfresh/metasfresh/commit/05015ee1636

PR #24811, opened 2026-06-26
EPCIS: one merged order-pure SSCC event for two orders sharing one LU (close-gate + closer-emits)

metas-ts:
## What

EPCIS export for two sales orders sharing **one physical LU (one SSCC)** now emits exactly **one** complete, order-pure event for that SSCC — on the completion that **closes the pallet** — instead of a premature/partial/duplicate/mis-attributed event depending on the order in which the two shipments complete.

## Why

The per-SSCC EPCIS export evaluated ownership + order references as a `DocStatus IN ('CO','CL')` snapshot at whichever shipment completion triggered it. With two orders on one pallet, any sharing shipment still open — or not yet created — at that moment corrupted the event. Surfaced on the LAF1010-3 "2 Bestellungen auf einer Palette" retest.

## How

SQL-only change to `"de.metas.edi".get_epcis_events_json_fn`, replacing the fixed lowest-`M_InOut_ID` "owner" election with:

1. **Close-gate** — the function returns `{}` while any child TU physically on an LU this shipment touches is **not** covered by a CO/CL shipment (covered = an active `M_HU_Assignment` from the TU to an inoutline of a CO/CL `M_InOut`, `isactive='Y'`). So every non-closing completion defers.
2. **Closer-emits** — `pallet_list` keeps all touched LUs and `desadv_agg` unions all CO/CL shipments sharing a touched LU (the `lu_owner` CTE is removed). Export eligibility is evaluated once per completion (the after-commit timing already in place), so the pallet-closing completion is the sole emitter; post-closure either sharer returns the full merged pallet.

Ships as the DDL file + a byte-identical migration (`5809860`).

## Tests

- `epcisExport.feature`:
  - `S30558_010` / `S30558_020` — two orders on one shared LU, both shipments created as drafts, completed in each order (A-first / B-first): the first completion's function call returns `{}` (gate), the closing completion returns the merged order-pure pallet. (Picking uses `DO_NOT_CREATE` + `AllowConsolidateInOut=N`, matching the EDI-recipient production config, so the two shipments are separate `M_InOut`s each holding only their own TUs.)
  - `S29231_170` reworked to the close-driven semantics (the old owner/sibling-`{}` assertions removed; both sharers return the merged pallet post-closure).
- `epcisScriptedExportStatus.feature`:
  - `S30558_040` — e2e proving the **closing** completion is the one enqueued for scripted export (`U`), while the first completion is `DontSend` (`N`).
- `C_BPartner_StepDef` gains optional `AllowConsolidateInOut` column support.

## Out of scope

The **sequential** flow (create+complete order 1, *then* create+complete order 2 on a shared pallet) currently fails at order-1 completion (the shipment's HU-assignment sweeps the whole LU → DESADV `validateMovementQtySum`). That is a separate, larger shipment-generation/HU-assignment fix, tracked separately. The in-scope contract here matches REQUIREMENTS AC1: emit one complete event only once the whole physical pallet is covered by completed shipments.

### DIFF
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/C_BPartner_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/C_BPartner_StepDef.java
@@ -107,6 +107,7 @@ import static org.compiere.model.I_C_BPartner.COLUMNNAME_M_PricingSystem_ID;
 import static org.compiere.model.I_C_BPartner.COLUMNNAME_PO_DiscountSchema_ID;
 import static org.compiere.model.I_C_BPartner.COLUMNNAME_PO_InvoiceRule;
 import static org.compiere.model.I_C_BPartner.COLUMNNAME_PO_PricingSystem_ID;
+import static org.compiere.model.I_C_BPartner.COLUMNNAME_AllowConsolidateInOut;
 import static org.compiere.model.I_C_BPartner.COLUMNNAME_PaymentRule;
 import static org.compiere.model.I_C_BPartner.COLUMNNAME_PaymentRulePO;
 import static org.compiere.model.I_C_BPartner.COLUMNNAME_Value;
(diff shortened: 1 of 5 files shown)


metas-ts:
Latest builkd https://github.com/metasfresh/metasfresh/actions/runs/28260927752 is green

PR #24830, opened 2026-06-28
Implement traffic manager order board

cristinamghita:


cristinamghita:
## Order Board window — how it works

The window is backed by the read-only view `M_Picking_OrderBoard_v`, which aggregates `m_picking_job_schedule_view` rows by product / UOM / status / delivery date / ship-to country / client / org.

### Three status tabs

| Tab | Label | Filter condition | Meaning |
|-----|-------|-----------------|---------|
| W | Wartend | `isassigned = 'N'` | Schedules waiting to be assigned to a picking job |
| K | In Kommissionierung | `isassigned = 'Y' AND processed = 'N'` | Schedules assigned to an active picking job |
| P | Packen | `isassigned = 'Y' AND processed = 'Y'` | Picking done; schedules ready for packing / shipment |

Each row in any tab shows:
- **Product** (value + name), **UOM**, **Delivery date**, **Ship-to country**
- **QtyTotal** — sum of `qtytopick` (or `qtytoscheduleforpicking` as fallback) across all grouped schedules
- **Order line count** — distinct `M_ShipmentSchedule_ID` count

Only rows where the schedule is already assigned (`isassigned = 'Y'`) **or** there is stock on hand (`qtyonhand > 0`) appear; schedules with neither are hidden.

### Synthetic PK

Because the view groups rows, there is no natural single-column key. The PK is a deterministic integer derived from the eight grouping columns via MD5 hash, with the sign bit masked off (`& x'7fffffff'::bit(32)`) to guarantee the value is always ≥ 0 — required by `TableRecordReference.of()`.

### Menu placement

Added under the **Picking** menu node (parent `AD_Menu_ID = 541856`) at `SeqNo = 3`, directly after the existing Picking entries.

PR #24845, opened 2026-06-29
Delete linked M_ReceiptSchedule records when purchase order line is deleted (#30516)

riadmen:
When a C_OrderLine is deleted, also delete any associated M_ReceiptSchedule records to prevent orphaned data and potential FK-constraint errors.

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/reservation/interceptor/C_OrderLine.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/reservation/interceptor/C_OrderLine.java
@@ -5,8 +5,10 @@ import de.metas.handlingunits.HUPIItemProductId;
 import de.metas.handlingunits.IHUPIItemProductDAO;
 import de.metas.handlingunits.model.I_C_Order;
 import de.metas.handlingunits.model.I_M_HU_PI_Item_Product;
+import de.metas.handlingunits.model.I_M_ReceiptSchedule;
 import de.metas.product.ProductId;
 import de.metas.util.Services;
+import org.adempiere.ad.dao.IQueryBL;
 import lombok.NonNull;
 import org.adempiere.ad.callout.annotations.Callout;
 import org.adempiere.ad.callout.annotations.CalloutMethod;
@@ -51,6 +53,7 @@ import java.util.Properties;
 public class C_OrderLine
 {
 	private final IHUPIItemProductDAO hupiItemProductDAO = Services.get(IHUPIItemProductDAO.class);
+	private final IQueryBL queryBL = Services.get(IQueryBL.class);
 
 	@Init
 	public void registerCallouts()
@@ -61,7 +64,11 @@ public class C_OrderLine
 	@ModelChange(timings = ModelValidator.TYPE_BEFORE_DELETE)
 	public void deleteReservation(@NonNull final I_C_OrderLine orderLineRecord)
 	{
-		// TODO
+		queryBL
+				.createQueryBuilder(I_M_ReceiptSchedule.class)
+				.addEqualsFilter(I_M_ReceiptSchedule.COLUMNNAME_C_OrderLine_ID, orderLineRecord.getC_OrderLine_ID())
+				.create()
+				.delete();
 	}
 
 	@ModelChange( //
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/reservation/interceptor/C_OrderLine.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/reservation/interceptor/C_OrderLine.java
@@ -13,6 +13,7 @@ import lombok.NonNull;
 import org.adempiere.ad.callout.annotations.Callout;
 import org.adempiere.ad.callout.annotations.CalloutMethod;
 import org.adempiere.ad.callout.spi.IProgramaticCalloutProvider;
+import org.adempiere.ad.dao.IQueryUpdater;
 import org.adempiere.ad.modelvalidator.annotations.Init;
 import org.adempiere.ad.modelvalidator.annotations.Interceptor;
 import org.adempiere.ad.modelvalidator.annotations.ModelChange;
@@ -64,6 +65,15 @@ public class C_OrderLine
 	@ModelChange(timings = ModelValidator.TYPE_BEFORE_DELETE)
 	public void deleteReservation(@NonNull final I_C_OrderLine orderLineRecord)
 	{
+		queryBL
+				.createQueryBuilder(I_M_ReceiptSchedule.class)
+				.addEqualsFilter(I_M_ReceiptSchedule.COLUMNNAME_C_OrderLine_ID, orderLineRecord.getC_OrderLine_ID())
+				.create()
+				.update(rs -> {
+					rs.setIsActive(false);
+					return IQueryUpdater.MODEL_UPDATED;
+				});
+
 		queryBL
 				.createQueryBuilder(I_M_ReceiptSchedule.class)
 				.addEqualsFilter(I_M_ReceiptSchedule.COLUMNNAME_C_OrderLine_ID, orderLineRecord.getC_OrderLine_ID())
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/reservation/interceptor/C_OrderLine.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/reservation/interceptor/C_OrderLine.java
@@ -6,6 +6,8 @@ import de.metas.handlingunits.IHUPIItemProductDAO;
 import de.metas.handlingunits.model.I_C_Order;
 import de.metas.handlingunits.model.I_M_HU_PI_Item_Product;
 import de.metas.inoutcandidate.api.IReceiptScheduleDAO;
+import de.metas.order.IOrderBL;
+import de.metas.order.OrderId;
 import de.metas.order.OrderLineId;
 import de.metas.product.ProductId;
 import de.metas.util.Services;
@@ -53,6 +55,7 @@ import java.util.Properties;
 public class C_OrderLine
 {
 	private final IHUPIItemProductDAO hupiItemProductDAO = Services.get(IHUPIItemProductDAO.class);
+	private final IOrderBL orderBL = Services.get(IOrderBL.class);
 	private final IReceiptScheduleDAO receiptScheduleDAO = Services.get(IReceiptScheduleDAO.class);
 
 	@Init
@@ -69,12 +72,8 @@ public class C_OrderLine
 	@ModelChange(timings = ModelValidator.TYPE_BEFORE_DELETE)
 	public void beforeOrderLineDeleted(@NonNull final I_C_OrderLine orderLineRecord)
 	{
-		// Receipt schedules only exist for purchase order lines.
-		// Guard against null: getC_Order() returns null when C_Order_ID <= 0 (malformed data).
-		// Treat that as a no-op to avoid NPE; deleteByOrderLineId will be a no-op anyway since
-		// no receipt schedules can be linked to an order-less order line.
-		final org.compiere.model.I_C_Order order = orderLineRecord.getC_Order();
-		if (order == null || order.isSOTrx())
+		// Receipt schedules only exist for purchase order lines
+		if (orderBL.isSalesOrder(OrderId.ofRepoId(orderLineRecord.getC_Order_ID())))
 		{
 			return;
 		}
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/reservation/interceptor/C_OrderLine.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/reservation/interceptor/C_OrderLine.java
@@ -69,7 +69,8 @@ public class C_OrderLine
 				.createQueryBuilder(I_M_ReceiptSchedule.class)
 				.addEqualsFilter(I_M_ReceiptSchedule.COLUMNNAME_C_OrderLine_ID, orderLineRecord.getC_OrderLine_ID())
 				.create()
-				.update(rs -> {
+				.updateDirectly(rs -> {
+					rs.setProcessed(false);
 					rs.setIsActive(false);
 					return IQueryUpdater.MODEL_UPDATED;
 				});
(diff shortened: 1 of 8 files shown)


riadmen:
@claude review

riadmen:
@claude review

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleDAO.java
```diff
@@ -387,6 +390,41 @@ public Optional<ReceiptScheduleId> getIdByQuery(@NonNull final ReceiptScheduleQu
...
+		// Deleting such schedules would corrupt receipt and invoice-candidate history.
+		final boolean hasReceivedAllocations = queryBL
+				.createQueryBuilder(I_M_ReceiptSchedule_Alloc.class)
+				.addInSubQueryFilter(I_M_ReceiptSchedule_Alloc.COLUMNNAME_M_ReceiptSchedule_ID, I_M_ReceiptSchedule.COLUMNNAME_M_ReceiptSchedule_ID, receiptSchedulesForOrderLine)
+				.addNotNull(I_M_ReceiptSchedule_Alloc.COLUMNNAME_M_InOutLine_ID)
+				.create()
+				.anyMatch();
+		if (hasReceivedAllocations)
+		{
+			throw new AdempiereException("Cannot delete order line: a receipt already exists for the linked receipt schedule.")
```
mrie04: Userfriendly AdMsg pls

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleDAO.java
```diff
@@ -387,6 +390,41 @@ public Optional<ReceiptScheduleId> getIdByQuery(@NonNull final ReceiptScheduleQu
...
+					.appendParametersToMessage()
+					.setParameter("orderLineId", orderLineId);
+		}
+
+		// updateDirectly bypasses model interceptors and goes straight to SQL.
+		// We unset Processed and IsActive here first so that the subsequent .delete() call is not blocked by that guard.
+		receiptSchedulesForOrderLine
+				.updateDirectly(rs -> {
```
mrie04: why not org.compiere.model.IQuery#deleteDirectly ?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleDAO.java
```diff
@@ -387,6 +390,41 @@ public Optional<ReceiptScheduleId> getIdByQuery(@NonNull final ReceiptScheduleQu
...
+				.createQueryBuilder(I_M_ReceiptSchedule.class)
+				.addEqualsFilter(I_M_ReceiptSchedule.COLUMNNAME_C_OrderLine_ID, orderLineId)
+				.create();
+
+		// Guard: refuse to delete if any receipt schedule has already been received (has M_ReceiptSchedule_Alloc with M_InOutLine_ID set).
+		// Deleting such schedules would corrupt receipt and invoice-candidate history.
+		final boolean hasReceivedAllocations = queryBL
+				.createQueryBuilder(I_M_ReceiptSchedule_Alloc.class)
+				.addInSubQueryFilter(I_M_ReceiptSchedule_Alloc.COLUMNNAME_M_ReceiptSchedule_ID, I_M_ReceiptSchedule.COLUMNNAME_M_ReceiptSchedule_ID, receiptSchedulesForOrderLine)
+				.addNotNull(I_M_ReceiptSchedule_Alloc.COLUMNNAME_M_InOutLine_ID)
```
mrie04, line 427: if you hit this case, you will have foreign key violation on delete becuase of I_M_ReceiptSchedule_Alloc.COLUMNNAME_M_ReceiptSchedule_ID most likely. so what if inout is reversed. Are the allocs still there and what state do the HUs have?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleDAO.java
riadmen: message added

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleDAO.java
riadmen: done

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleDAO.java
riadmen: `.addOnlyActiveRecordsFilter()` added excludes deactivated allocs, so reversed receipts don't block deletion.

riadmen:
@claude review

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleDAO.java
mrie04: agree, good reason to not use here https://github.com/metasfresh/metasfresh/pull/24845#issuecomment-4835368137 , but update of processed could also trigger material events. Needs to be checked
For sure don't add new mrp logic inside delete. Delete should be done in a way so mrp handeles it correctly

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleDAO.java
mrie04: alloc has 
```
m_receiptschedule_id       numeric(10)                      NOT NULL
        CONSTRAINT mreceiptschedule_mreceiptsched
            REFERENCES m_receiptschedule
            DEFERRABLE INITIALLY DEFERRED
```
so you can't delete without cleaning up linked allocs first. And there you need to be sure they can be removed

riadmen:
@claude review

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleDAO.java
riadmen: Now the new comment makes two things explicit:                                                                                                                                                                                                                                                                          
  1. updateDirectly suppresses the spurious ReceiptScheduleUpdatedEvent that would otherwise fire from the Processed field change.                                                                                                                                        
  2. MRP is correctly notified: the model-based delete() that follows fires M_ReceiptSchedule_PostMaterialEvent on TYPE_BEFORE_DELETE, posting the correct ReceiptScheduleDeletedEvent.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleDAO.java
riadmen: 1. The FK constraint exists: allocs are deleted with deleteDirectly() before the parent M_ReceiptSchedule rows.
2. Alloc removal must be safe: 

- the guard checks for active allocs addOnlyActiveRecordsFilter() with M_InOutLine_ID NOT NULL. This means only allocs tied to an actual open receipt are deleted 
- reversed-receipt allocs (IsActive=false) and plain unreceived allocs (M_InOutLine_ID=null) are confirmed safe to delete, as the comment at lines 420–424 explains.

riadmen:
@claude review

riadmen:
@claude review

riadmen:
@claude review

riadmen:
-  Finding: NPE on getC_Order() — Added a null guard: final I_C_Order order = orderLineRecord.getC_Order(); if (order == null || order.isSOTrx()). If C_Order_ID <= 0, getC_Order() returns null and the method now safely returns as a no-op.                                                                                                                                
                                                                                                                                             
-   Finding #1 (double ReceiptScheduleDeletedEvent) — Added a comment explaining why this is harmless: inactivateReceiptSchedules() always immediately physically deletes after setting IsActive=false, so the steady state of "IsActive=false, not yet deleted" doesn't exist in practice. If it did, QtyToMove would be 0, making the second event a MRP no-op.                                                                                         
                                                                                                                                                     
-   Finding #3 (getQtyOrdered sensitivity) — Added a comment confirming getQtyOrdered() reads the QtyOrdered column directly and is not affected by the Processed=false set via updateDirectly().                                                                                                         
                                                                                                                                                  
-   Finding #4 (error message recovery hint) — Updated both EN and DE messages to add actionable guidance: "Please reverse the associated receipt before deleting this order line."                                                                                                                  
                                                                                                                                                   
-   Finding #2 (alloc cleanup test) and Finding #5 (isSOTrx guard test) — These are integration-level gaps already acknowledged in the test comments;  

-   no code change needed now.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/reservation/interceptor/C_OrderLine.java
```diff
@@ -51,17 +53,32 @@
...
-		// TODO
+		// Receipt schedules only exist for purchase order lines.
+		// Guard against null: getC_Order() returns null when C_Order_ID <= 0 (malformed data).
+		// Treat that as a no-op to avoid NPE; deleteByOrderLineId will be a no-op anyway since
+		// no receipt schedules can be linked to an order-less order line.
+		final org.compiere.model.I_C_Order order = orderLineRecord.getC_Order();
+		if (order == null || order.isSOTrx())
+		{
+			return;
+		}
```
mrie04: 1. why fqdn?
2. use DAO/Repo never getC_Order()
3. how should order be null?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleDAO.java
mrie04: where are the allocs deleted first to prevent FK-violation?

backend/de.metas.swat/de.metas.swat.base/src/test/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleTestBase.java
```diff
@@ -395,6 +397,39 @@ protected I_C_OrderLine createOrderLine(final I_C_Order order, final I_M_Product
...
+	}
+
+	/**
+	 * Creates an {@link I_M_ReceiptSchedule_Alloc} for the given receipt schedule.
+	 *
+	 * @param inOutLineId set to a positive value to simulate a received alloc (goods were received); 0 means not yet received
+	 */
+	protected I_M_ReceiptSchedule_Alloc createReceiptScheduleAlloc(
+			final I_M_ReceiptSchedule receiptSchedule,
+			final int inOutLineId)
```
mrie04: repoid

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleDAO.java
mrie04: so I checked `de.metas.purchasecandidate.material.interceptor.M_ReceiptSchedule_PostMaterialEvent#createReceiptScheduleEvent` myself. It also reduces qtys on close, which should be done on order reactivate before orderLine delete, so is the delete interceptor rly needed in addition?

backend/de.metas.swat/de.metas.swat.base/src/test/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleTestBase.java
riadmen: now using `@Nullable InOutLineId inOutLineId` instead of `int inOutLineId`

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/reservation/interceptor/C_OrderLine.java
riadmen: 1. Why FQDN? — Removed entirely. The local variable that required org.compiere.model.I_C_Order is gone.                                                                                                                                                                                                             
2. Use DAO/Repo, never getC_Order() — Replaced with orderBL.isSalesOrder(OrderId.ofRepoId(orderLineRecord.getC_Order_ID())). IOrderBL is now injected as a field alongside receiptScheduleDAO.                                                                                                                      
3. How should order be null? — The null guard and its comment are gone. C_OrderLine.C_Order_ID is NOT NULL in the database, so null is not possible in production. OrderId.ofRepoId() would throw a proper exception for ID ≤ 0 if it ever occurred, which is the correct behaviour.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ReceiptScheduleDAO.java
riadmen: - Check `de.metas.handlingunits.receiptschedule.model.validator.M_ReceiptSchedule#onReceiptScheduleDelete`:

```Java
// Finally, delete all the allocations ((hu or not hu)
		for (final I_M_ReceiptSchedule_Alloc alloc : allocations)
		{
			InterfaceWrapperHelper.delete(alloc);
		}
```

- This runs inside onReceiptScheduleDelete which is a @ModelChange(timings = TYPE_BEFORE_DELETE) — so it fires before the actual DELETE FROM M_ReceiptSchedule SQL executes. The allocs are therefore gone by the time the parent row is removed. The FK constraint (DEFERRABLE INITIALLY DEFERRED) covers the gap within the transaction.

mrie04 (approved):
lgtm

PR #24898, opened 2026-06-30
MTree: log orphan-node recovery at DEBUG, not ERROR

teosarca:
## What

In `MTree.java`, downgrade the log statement for the orphan-node recovery path from `log.error(...)` to `log.debug(...)`:

```
- log.error("Nodes w/o parent - adding to root - {}", buffer);
+ log.debug("Nodes w/o parent - adding to root - {}", buffer);
```

The sibling `log.error("Still nodes in Buffer - {}", buffer);` is intentionally left at ERROR — that path is a genuine failure (nodes that could not be re-attached).

## Why

`MetasfreshIssueAppender` creates an `AD_Issue` row for every log event at `Level.ERROR` (`backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/logging/MetasfreshIssueAppender.java:47,156-161`).

The "Nodes w/o parent" path is **benign and self-recovering**: immediately after the log line, the `for` loop calls `root.add(node)` to re-attach the orphan nodes to the tree root. So each occurrence was minting a noise `AD_Issue` for a non-error.

This path is also **not real log spam**: on dantherme2e it produced ~139 events over ~10 months (peak 49 in one month, ~0.5/day overall). The only actual noise was the `AD_Issue`-per-ERROR. Per review on this PR (teosarca), `debug` is preferred over `warn`: it keeps the message out of the prod INFO log entirely while remaining available when debugging tree loading.

## Scope

- **No DB migration**
- **No API change**
- **No behaviour change** — pure log-level change; the orphan-node re-attachment logic is untouched.


backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/model/MTree.java
```diff
@@ -290,7 +290,7 @@ private MTreeNode retrieveRootWithChildren(@Nullable final UserId userId)
 		// Nodes w/o parent
 		if (!buffer.isEmpty())
 		{
-			log.error("Nodes w/o parent - adding to root - {}", buffer);
+			log.warn("Nodes w/o parent - adding to root - {}", buffer);
```
teosarca: IMHO u can even use "debug" instead!

teosarca: **Resolution (dev):** Kept `warn` over `debug` deliberately. The root logger runs at `INFO`, so `debug` would make this event **invisible in production logs** — we'd lose all trace that orphaned tree nodes occurred. `warn` is the sweet spot: it stops the `AD_Issue` noise (the appender only mints on `ERROR`) while keeping a visible breadcrumb for a genuine structural anomaly that we'd still want to notice if it spikes. Happy to switch to `debug` if you'd rather fully silence it — your call.

teosarca: Done — switched to `debug` in c58520ce69a.

For the record: this path fired ~139 times over ~10 months on dantherme2e (peak 49 in one month, ~0.5/day overall), so it was never real log spam — the AD_Issue-per-ERROR was the only actual noise. `debug` keeps it out of the prod INFO log entirely while staying available when debugging tree loading.

teosarca: **Resolution (dev):** Adopted — switched `log.warn` → `log.debug` in commit https://github.com/metasfresh/metasfresh/commit/c58520ce69a. (Evidence it was never real spam: ~139 events over ~10 months on dantherme2e, peak 49/month; `debug` keeps it out of the prod INFO log entirely while staying available when debugging tree loading.)

PR #24907, opened 2026-07-01
HU Label Config: desktop picking label consignee parity (#30763 AC8)

teosarca:
## What

Desktop-picking label parity for me03 #30763 **AC8**: https://github.com/metasfresh/me03/issues/30763

Stamps the picking consignee (BPartner + delivery location) onto the desktop WebUI picking HU **before** the label print, so the unchanged `M_HU_Label_Config` matching path selects the correct per-BPartner config for a partner-less picking HU.

Two call sites in `de.metas.ui.web.base`:
- `WEBUI_Picking_TU_Label.printPickingLabel` — stamp before `huLabelService.print(...)`.
- `WEBUI_Picking_PickQtyToNewHU.printPickingLabelIfAutoPrint` — stamp before the auto-print.

The consignee is resolved from the current shipment schedule via `IShipmentScheduleEffectiveBL.getBPartnerLocationId(...)` and stamped by reusing the **same** guarded facade method as the merged mobile fix: `PickingJobHUService.setBPartnerAndLocationIfNotSet` (only-if-unset). The `M_HU` `updateChildren` interceptor cascades the bpartner + location to child TUs/CUs.

## Why

The desktop picking label processes key the label-config lookup on the HU's **own** bpartner (via `HUToReportWrapper`). Picking HUs in the non-pack-for-shipping flow are materialised **without** a partner, so the per-BPartner `M_HU_Label_Config` rows never match — the same root cause as the mobile bug fixed in https://github.com/metasfresh/metasfresh/pull/24904 (AC1-AC5, commit 3ec948c1fae).

Concretely on the desktop flow:
- `WEBUI_Picking_TU_Label` prints with `failOnMissingLabelConfig=true`, so a partner-less HU **throws** rather than printing the consignee's label.
- `WEBUI_Picking_PickQtyToNewHU` creates its pack-to TU via `createTU` with no bpartner, so its auto-print is silently skipped.

This PR brings the desktop flow to parity with the mobile fix.

## Scope / safety

- The matching path is **untouched** (`HULabelPrintRequest` / `HULabelPrintCommand` / `HULabelConfigRoute` / `HUToReportWrapper`).
- Guard is only-if-unset (`C_BPartner_ID > 0` short-circuit inside the reused method), so pack-for-shipping HUs that already carry a partner are not disturbed.
- No AD / migration / schema change.

## Testing

- Offline reactor compile of `de.metas.ui.web.base` + all upstream dependencies: **success** (JDK8, `mvn -o install -pl de.metas.ui.web.base -am`).
- No unit-test harness exists for these `ViewBasedProcessTemplate` process classes (they require live view/selection state that only the WebUI produces); the stamp logic is a thin guarded delegate. A faithful desktop-picking Playwright/manual proof is deferred (same note as the approach memo).

### DIFF
--- a/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/picking/pickingslot/process/PickingSlotViewBasedProcess.java
+++ b/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/picking/pickingslot/process/PickingSlotViewBasedProcess.java
@@ -1,8 +1,12 @@
 package de.metas.ui.web.picking.pickingslot.process;
 
+import de.metas.bpartner.BPartnerLocationId;
+import de.metas.handlingunits.HuId;
 import de.metas.handlingunits.model.I_M_ShipmentSchedule;
+import de.metas.handlingunits.picking.job.service.external.hu.PickingJobHUService;
 import de.metas.inout.ShipmentScheduleId;
 import de.metas.inoutcandidate.api.IShipmentScheduleBL;
+import de.metas.inoutcandidate.api.IShipmentScheduleEffectiveBL;
 import de.metas.inoutcandidate.api.IShipmentSchedulePA;
 import de.metas.process.IProcessPrecondition;
 import de.metas.ui.web.picking.packageable.PackageableView;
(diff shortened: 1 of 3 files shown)


teosarca:
wrong requiremnt

PR #24911, opened 2026-07-01
HU hotfix: deactivate surplus HU QR-code assignments when aggregate TU count decreases

teosarca:
## Root cause

**Hypothesis (mechanism verified by code, runtime symptom not re-reproduced in this code-only PR — a faithful reproduction test lands in the follow-up test PR):**

When TUs are split/picked out of an **aggregate HU**, `AggregateHUTrxListener.updateItemQtyAndSplitIfNeeded` reduces the aggregate's `M_HU_Item.Qty` (its TU count) but **never deactivated the now-excess `M_HU_QRCode_Assignment` rows** (there is an explicit `TODO` in that method acknowledging the missing cleanup).

QR-code generation only ever **adds** codes (one per TU) and only removes them on HU-destroy. As a result an aggregate HU accumulates more active QR-code assignments than its current TU count. Mobile picking then throws `Erwartet {X} QR-Codes, aber nur {Y} erhalten` — `PickingJobPickCommand` checks `getOrCreateQRCodesByHuId(hu).size() == tu.getQtyTU()`.

Verified (mechanism): `AggregateHUTrxListener.updateItemQtyAndSplitIfNeeded` calls `item.setQty(newTuQty)` then has no QR-assignment cleanup (explicit `TODO`); `PickingJobPickCommand` asserts `#QRCodes == getQtyTU()`. No path deactivates surplus assignments except HU-destroy (`deactivateAssignmentsByHuId`).

## Fix

After the aggregate's TU qty is finalized, keep the **oldest** N active assignments (N = current TU count) and soft-delete the surplus (`IsActive='N'`), so the active assignment count stays in sync with the TU count. Hooked on the proven trx-listener path (not a model interceptor).

- **`HUQRCodesRepository.deactivateSurplusAssignments(HuId, keepCount)`** — query the HU's active assignments ordered by `Created` then `M_HU_QRCode_Assignment_ID`, keep the first `keepCount`, deactivate the rest (deactivate all if `keepCount <= 0`). Mirrors the existing `deactivateAssignmentsByHuId` soft-delete pattern.
- **`HUQRCodesService.deactivateSurplusAssignments(HuId, keepCount)`** — delegates to the repository.
- **`AggregateHUTrxListener.updateItemQtyAndSplitIfNeeded`** — calls the service on the aggregate VHU after qty is finalized, guarded on a non-empty included-HUs list.

No DB column / migration is needed.

## Tests

None in this PR by design — this is the **code fix only, part of a multi-part delivery**. Automated tests are added in a **separate follow-up PR**.

## Verification

Module `de.metas.handlingunits.base` compiles clean (`mvn install -pl de.metas.handlingunits.base -am`, BUILD SUCCESS).

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/service/HUQRCodesRepository.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/service/HUQRCodesRepository.java
@@ -173,6 +173,41 @@ public class HUQRCodesRepository
 				});
 	}
 
+	/**
+	 * Soft-delete the SURPLUS active {@code M_HU_QRCode_Assignment} rows of the given HU: keep the oldest
+	 * {@code keepCount} active assignments (ordered ascending by {@code Created}, then
+	 * {@code M_HU_QRCode_Assignment_ID}) and set {@code IsActive='N'} on every row beyond that.
+	 * <p>
+	 * Needed because splitting/picking TUs out of an aggregate HU reduces its TU count (its
+	 * {@code M_HU_Item.Qty}) but leaves the previously-generated QR-code assignments active, so the aggregate
+	 * ends up with more active assignments than TUs. Mobile picking then fails its
+	 * {@code #QRCodes == getQtyTU()} check. Keeping the oldest ones deactivates the newer surplus.
+	 * <p>
+	 * When {@code keepCount <= 0}, all active assignments are deactivated.
+	 */
+	public void deactivateSurplusAssignments(@NonNull final HuId huId, final int keepCount)
+	{
+		final List<I_M_HU_QRCode_Assignment> activeAssignments = queryBL.createQueryBuilder(I_M_HU_QRCode_Assignment.class)
+				.addOnlyActiveRecordsFilter()
+				.addEqualsFilter(I_M_HU_QRCode_Assignment.COLUMNNAME_M_HU_ID, huId)
+				.orderBy(I_M_HU_QRCode_Assignment.COLUMNNAME_Created)
+				.orderBy(I_M_HU_QRCode_Assignment.COLUMNNAME_M_HU_QRCode_Assignment_ID)
+				.create()
+				.list();
+
+		final int keep = Math.max(keepCount, 0);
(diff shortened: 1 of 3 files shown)


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/service/HUQRCodesRepository.java
```diff
@@ -173,6 +173,41 @@ public void deactivateAssignmentsByHuId(@NonNull final HuId huId)
...
+	 * ends up with more active assignments than TUs. Mobile picking then fails its
+	 * {@code #QRCodes == getQtyTU()} check. Keeping the oldest ones deactivates the newer surplus.
+	 * <p>
+	 * When {@code keepCount <= 0}, all active assignments are deactivated.
+	 */
+	public void deactivateSurplusAssignments(@NonNull final HuId huId, final int keepCount)
+	{
+		final List<I_M_HU_QRCode_Assignment> activeAssignments = queryBL.createQueryBuilder(I_M_HU_QRCode_Assignment.class)
+				.addOnlyActiveRecordsFilter()
+				.addEqualsFilter(I_M_HU_QRCode_Assignment.COLUMNNAME_M_HU_ID, huId)
```
teosarca, line 192: make sure we have a db index on M_HU_ID

teosarca: Index already exists (core schema) — no new index needed.

`idx_unq_m_hu_qrcode_assignmnet` is a partial UNIQUE index on `M_HU_QRCode_Assignment (M_HU_ID, M_HU_QRCode_ID) WHERE isActive='Y'`, added in core migration `5717250_sys_gh17351_add_hu_assignment_unq_idx.sql` (gh17351). `M_HU_ID` is the **leading column** and the partial `WHERE isActive='Y'` matches the `addOnlyActiveRecordsFilter` here, so this query (`WHERE m_hu_id=? AND isactive='Y' ORDER BY created, m_hu_qrcode_assignment_id`) is fully index-served. Verified present on the soft_panda schema.

teosarca:
wrong approach

PR #24930, opened 2026-07-02
`Change Date Promised Override` process update

riadmen:


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/IReceiptScheduleBL.java
```diff
@@ -239,7 +239,7 @@ public interface IReceiptScheduleBL extends ISingletonService
 
 	List<ReceiptScheduleId> retainLUQtySchedules(List<ReceiptScheduleId> receiptSchedules);
 
-	int updateDatePromisedOverrideAndPOReference(@NonNull PInstanceId pinstanceId, @Nullable LocalDateTime datePromisedOverride, @Nullable String poReference);
+	int updateDatePromisedOverrideAndPOReference(@NonNull PInstanceId pinstanceId, @Nullable LocalDateTime datePromisedOverride, @Nullable String poReference, @Nullable Boolean isConfirmedBySupplier);
```
metas-rc: I think the method name no longer fits. Please, come up with a more inclusive name or use a new setter for the flag

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/70-de.metas.fresh/5811220_sys_gh30568_M_ReceiptSchedule_ChangeDatePromised_OverrideAndPOReference_Process_Update.sql
```diff
@@ -0,0 +1,15 @@
...
+;
+
+-- 2026-07-02T08:50:29.953Z
+INSERT INTO AD_Process_Para_Trl (AD_Language,AD_Process_Para_ID, Description,Help,Name, IsTranslated,AD_Client_ID,AD_Org_ID,Created,Createdby,Updated,UpdatedBy,IsActive) SELECT l.AD_Language, t.AD_Process_Para_ID, t.Description,t.Help,t.Name, 'N',t.AD_Client_ID,t.AD_Org_ID,t.Created,t.Createdby,t.Updated,t.UpdatedBy,'Y' FROM AD_Language l, AD_Process_Para t WHERE l.IsActive='Y'AND (l.IsSystemLanguage='Y' OR l.IsBaseLanguage='Y') AND t.AD_Process_Para_ID=543260 AND NOT EXISTS (SELECT 1 FROM AD_Process_Para_Trl tt WHERE tt.AD_Language=l.AD_Language AND tt.AD_Process_Para_ID=t.AD_Process_Para_ID)
+;
+
+-- 2026-07-02T08:50:30.046Z
+/* DDL */  select update_Process_Para_Translation_From_AD_Element(584306)
+;
```
metas-rc: I would also make sure the flag is displayed , grid, filter etc. in metasfresh since this functionality is for everybody

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/IReceiptScheduleBL.java
riadmen: done: renamed to to updateDatePromisedOverridePOReferenceAndConfirmedBySupplier

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/70-de.metas.fresh/5811220_sys_gh30568_M_ReceiptSchedule_ChangeDatePromised_OverrideAndPOReference_Process_Update.sql
riadmen: done

riadmen:
@claude review

riadmen:
@claude review

metas-rc:
@claude @riadmen 
Regarding ``updateDatePromisedOverridePOReferenceAndConfirmedBySupplier`` : what will we do if we get 3 extra things to set? Will we append them all in the name on the method?

riadmen:
@claude review

riadmen:
Addressed both issues from the review above (https://github.com/metasfresh/metasfresh/pull/24930#issuecomment-4866305916):

**1. Stale validation message** — fixed in `5811300_sys_gh30568_ReceiptSchedule_ChangeDatePromised_ParamsValidationError_Message_Update.sql`. The `paramsValidationError` message (EN + de_CH/de_DE) now mentions "Confirmed by Supplier" as a third valid option.

**2. `IsConfirmedBySupplier` possibly never resolving to `null`** — instead of a defensive code change, fixed at the dictionary level: the process parameter was switched from a YesNo checkbox (`AD_Reference_ID=20`) to a List reference (`AD_Reference_ID=17`, `AD_Reference_Value_ID=540528`) with values `Y`/`N` that can be left unselected (`5811220_sys_gh30568_...sql`).

Verified this closes the gap by tracing the resolution path end to end:
- `ProcessClassParamInfo.java:211-214` resolves `Boolean`-typed `@Param` fields via `DisplayType.toBoolean(source.getParameterAsString(parameterName), null)`.
- `ProcessParams.getParameterAsString()` looks up the parameter in a map built from the actual `AD_PInstance_Para` rows for the run. If the List parameter is left unselected, either no row exists (`getParameterAsString` → `null`) or a row with an empty string exists — and `StringUtils.toBoolean("", null)` also falls through to `null` since `""` matches neither `Y/true` nor `N/false`.
- Either way, an untouched `IsConfirmedBySupplier` now reliably resolves to `null`, not `Boolean.FALSE`. It only becomes `Boolean.FALSE` if the user explicitly picks `N`.

So no Java changes were needed — `Boolean isConfirmedBySupplier` and the existing null-checks in `UpdateReceiptScheduleOverridesRequest`/`ReceiptScheduleBL.updateReceiptScheduleOverrides` already handle this correctly once the parameter itself can genuinely be "unset".

metas-rc:
@riadmen @claude 
The finding ``2. YesNo parameter may never be null — medium`` is legit

Please, make sure the process can be ran with a null value in the parameter and when ran like so, the flag is not changed in the selected entries

Please, also address finding ``1. Stale validation error message — medium`` .

metas-rc (approved):

PR #24955, opened 2026-07-02
MobileUI Distribution E2E: assert filtered launchers order-independently (flaky filter_by_packingplace)

teosarca:
## What

Fix the intermittent CI failure of `filter_by_packingplace.spec.js` → `fork_lift_1 (IsPackingPlace=N) is offered the intra-storage groundfill orders DD3, DD4` (mobile-webui). One of a catalogued set of flaky CI tests (`gh30784`).

## Root cause (why it's intermittent)

The test asserted launcher **slot order** via `expectJobButtons([DD3, DD4])`. But this test is a **filtering** assertion — its point is *which* launchers `fork_lift` is offered (DD3 and DD4, not DD1/DD2), not their order.

The rendered launcher order is **not guaranteed** to match DocumentNo:
- SQL is deterministic (`PriorityRule → DatePromised → DD_Order_ID` tiebreaker, `DDOrderLowLevelDAO`), but
- the JSON layer re-sorts as `AlreadyStartedFirst → caption(orderByFields) → completeCaption(translated string)` (`JsonWorkflowLaunchersList.java:74`) with **no id tiebreaker** — full ties fall back to `Stream.sorted` stability + encounter order, and `AlreadyStartedFirst` is the *primary* key.

So two launchers that are both present can swap slots run-to-run → `Expected 1000025 / Received 1000023`.

## Fix (test-only, no production change)

Add `DistributionJobsListScreen.expectJobButtonsInAnyOrder(...)`:
- waits each expected launcher **visible**,
- checks each expectation **by testId** (not by slot),
- keeps the **exact-count** check (no unexpected extras).

Membership + count + per-button props are still asserted — **only slot order is relaxed**. The `fork_lift` filtering test now uses it.

A genuine launcher-ordering guarantee, if wanted, belongs in its own dedicated ordering test — not piggybacked on a filtering test — so no production sort code is touched here (deliberate scope decision).

## Verification

- Parse-verified locally (`playwright test … --list` lists all 3 tests).
- Functional proof: the mobile-webui CI job on this PR (the `.webm` of the green run is the evidence).

## Scope

Item 1 of a flaky-test catalog (`gh30784`); one PR per item. Sibling distribution specs use the same positional `expectJobButtons` and could adopt the any-order variant if they ever flake the same way — not changed here.


e2e/mobile-webui/tests/spec/distribution/filter_by_packingplace.spec.js
```diff
@@ -128,7 +128,10 @@ test('fork_lift_1 (IsPackingPlace=N) is offered the intra-storage groundfill ord
     await openDistributionLaunchers(masterdata.login.user);
 
     await DistributionJobsListScreen.expectHeaderProperty({ caption: 'Workplace', value: masterdata.workplaces.wpForkLift.name });
-    await DistributionJobsListScreen.expectJobButtons([
+    // This is a *filtering* assertion (fork_lift is offered both intra-storage groundfill orders),
+    // not an ordering assertion — DD3/DD4 slot order is not part of what this test proves. Assert the
+    // set order-independently so the non-deterministic launcher render order can't flake it (me03 #30784).
```
teosarca: Need a full RCA on why this AI slop is also
* contains forbidden references
* contains non timeless comments

was not catched by your review!

teosarca: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/9eb29a1b011 — both comments now carry no issue reference and read timelessly. A scan of the added diff lines for issue-ref patterns is clean.

Full RCA, on both halves of the question — why the slop existed, and why review didn't catch it.

**What was wrong.** Two of my added source comments embedded a private-tracker issue reference: (1) a **forbidden reference** — a private-tracker id inside a *public* metasfresh source file; (2) a **non-timeless comment** — permanent code pointing at a transient issue that becomes noise the moment it closes.

**Why it was authored.** I applied "link the issue as `gh<n>`" reasoning — correct for a PR body / commit message — to *source code*, where the correct answer is **no issue reference at all**: the issue↔code linkage lives in the PR/commit, never in permanent code. I conflated PR-metadata convention with source-comment convention.

**Why my review missed it (the real question).** I *did* run the private-reference etiquette scan — but scoped it to **PR metadata only** (title/body/commit), where I caught and fixed the private URL. I never ran the same scan over the **code content of the diff**, so the identical leak class sailed straight through in the comments. Compounding it: the code-reviewer brief I wrote listed the mobile-E2E and coverage rules but **omitted "flag forbidden/private/issue references and non-timeless comments in the changed code"** — a reviewer only checks the criteria it's given, so none of the passes looked for it. The mechanical guards (co-author, etiquette hooks) inspect commit trailers / PR body, not source-comment text — no automated backstop existed for a reference inside code.

**Prevention (validated).**
1. Done — stripped both refs, timeless wording (commit above).
2. The forbidden-reference + timeless-comment scan now runs on the **diff's added lines**, not just PR metadata — folded into my pre-PR self-check and into every code-reviewer brief.
3. Capturing this as a Review-rule + a mechanical guard (grep added diff lines for issue-ref patterns) so the class is caught deterministically, not by review vigilance.

**Self-check:** grepping the added lines for issue-ref patterns — which I've now done and it reports clean post-fix — would have flagged both lines at authoring time; adding the criterion to the reviewer brief would have surfaced them in the review.

PR #24960, opened 2026-07-02
C_UOM_Conversion: DB guards against identity/duplicate conversions crashing product API

teosarca:
## Problem

A `C_UOM_Conversion` row whose `C_UOM_ID = C_UOM_To_ID` (an "identity" conversion), or an active duplicate UOM pair for the same product, makes the per-product UOM-conversion map build throw (Guava `Maps.uniqueIndex` duplicate key). As a result `GET /api/v2/material/products` fails (bulk export → 400, single fetch → 422).

An application interceptor already rejects identity rows on ORM save, so such rows can only enter via ORM-bypassing paths (raw SQL, DB functions, or data predating the interceptor).

## Change

Migration `5812070` makes the table self-guarding (idempotent + re-run-safe):

1. Backs up the table, then deletes the unambiguous junk — non-catch identity rows (`c_uom_id = c_uom_to_id`).
2. **Fails loud** on ambiguous cases needing a human decision: catch-weight identity rows, and active duplicate pairs (product or generic) whose rates may differ.
3. Replaces the product unique index `c_uom_conversion_product` with `c_uom_conversion_product_uom_uq` — same uniqueness, but `m_product_id` leading gives a true index seek for the by-product conversion load.
4. Adds `c_uom_conversion_generic_uq` — closes the generic-duplicate gap the product index misses (Postgres treats a NULL `m_product_id` as distinct in a unique index); also serves the generic-conversion load.
5. Adds `CHECK (c_uom_id <> c_uom_to_id)` — blocks identity rows at the DB level, including ORM-bypassing paths.

Audited every DB writer of `c_uom_conversion` (weight-upsert function, catch-UOM trigger, one-time migrations): none can produce an identity row, so no writer guard is needed.

## Test

- DDL validated on a live stack: idempotent across repeated runs; the CHECK rejects identity inserts; product/generic uniqueness enforced; fail-loud fires on injected duplicates.
- CI `db-apply-migrations` exercises the script on a clean seed.

### DIFF
--- a/backend/de.metas.business/src/main/sql/postgresql/system/43-de.metas.product/5812070_sys_c_uom_conversion_no_self_conversion_and_generic_uq.sql
+++ b/backend/de.metas.business/src/main/sql/postgresql/system/43-de.metas.product/5812070_sys_c_uom_conversion_no_self_conversion_and_generic_uq.sql
@@ -1,0 +1,75 @@
+-- Make C_UOM_Conversion self-guarding against rows that crash the product REST API.
+--
+-- Problem: a row with C_UOM_ID = C_UOM_To_ID (an "identity" conversion) or an
+-- active duplicate UOM pair makes the per-product UOM-conversion map build throw
+-- (Guava Maps.uniqueIndex duplicate key), so GET /api/v2/material/products fails.
+-- An application interceptor already rejects identity rows on ORM save; this adds
+-- DB-level guards so ORM-bypassing paths (raw SQL, DB functions) cannot introduce
+-- them either, and closes the generic-duplicate gap the existing product index
+-- misses (Postgres treats NULL M_Product_ID as DISTINCT in a unique index).
+--
+-- Idempotent + re-run-safe (also runnable as a manual hotfix on a target instance).
+
+-- Backup before the destructive cleanup (non-AD data table).
+SELECT backup_table('c_uom_conversion', '_uom_identity_cleanup');
+
+-- STEP 1 — remove the unambiguous junk: non-catch identity rows (active + inactive).
+-- An identity conversion (from == to) is always redundant; the code serves X->X
+-- synthetically. Catch-weight rows are excluded and handled fail-loud in STEP 2.
+DELETE FROM c_uom_conversion
+WHERE c_uom_id = c_uom_to_id
+  AND iscatchuomforproduct = 'N';
+
+-- STEP 2 — FAIL LOUD on the ambiguous cases (data decisions, not mechanical):
+--   (a) catch-weight identity rows (should never exist; deleting one has
+--       shipment-schedule side effects -> needs manual review);
+--   (b) active duplicate UOM pairs (product OR generic) with from <> to — the two
+--       rows may carry different rates, so which to keep is a master-data decision.
+-- Raising aborts the (transactional) migration cleanly for manual resolution.
+DO $$
+DECLARE
+    catch_identity int;
+    dup_groups     text;
+BEGIN
+    SELECT count(*) INTO catch_identity
+    FROM c_uom_conversion
+    WHERE c_uom_id = c_uom_to_id AND iscatchuomforproduct = 'Y';
+    IF catch_identity > 0 THEN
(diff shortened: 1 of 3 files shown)


backend/de.metas.business/src/main/sql/postgresql/system/43-de.metas.product/5812070_sys_c_uom_conversion_no_self_conversion_and_generic_uq.sql
```diff
@@ -0,0 +1,75 @@
...
+--
+-- Idempotent + re-run-safe (also runnable as a manual hotfix on a target instance).
+
+-- Backup before the destructive cleanup (non-AD data table).
+SELECT backup_table('c_uom_conversion', '_uom_identity_cleanup');
+
+-- STEP 1 — remove the unambiguous junk: non-catch identity rows (active + inactive).
+-- An identity conversion (from == to) is always redundant; the code serves X->X
+-- synthetically. Catch-weight rows are excluded and handled fail-loud in STEP 2.
+DELETE FROM c_uom_conversion
```
teosarca: pls avoid deleting here. that is manual job.
here u just assume everthing is fine and fail nice in case of nok.

teosarca: **Resolution (dev):** Done — removed the DELETE from this migration. It now installs only the guards (the two partial unique indexes + the CHECK) and, if identity rows or active duplicate pairs still exist, it fails loud with a clear message telling the operator to run the manual cleanup first and re-apply. The row cleanup is now the separate manual per-instance script (not shipped in the PR). Fix: https://github.com/metasfresh/metasfresh/commit/3a92d30001f08fadaab0d0eccf3238ecb02b0fbc

backend/de.metas.business/src/main/sql/postgresql/system/43-de.metas.product/5812070_sys_c_uom_conversion_no_self_conversion_and_generic_uq.sql
```diff
@@ -0,0 +1,60 @@
...
+    ) d;
+    IF dup_groups IS NOT NULL THEN
+        RAISE EXCEPTION 'C_UOM_Conversion: active duplicate conversion pair(s) exist; resolve manually (rates may differ), then re-apply: %', dup_groups;
+    END IF;
+END $$;
+
+-- STEP 2 — PRODUCT-scope unique index. Replaces gh18349's c_uom_conversion_product.
+-- Same uniqueness (identical column set), but M_Product_ID leading gives a true index
+-- seek for the by-product load (WHERE M_Product_ID=? AND IsActive='Y').
+DROP INDEX IF EXISTS c_uom_conversion_product;
```
teosarca: why are u dropping "DROP INDEX IF EXISTS c_uom_conversion_product;"??????

teosarca: **Resolution (dev):** You're right — it should not be dropped. Removed the drop+recreate entirely. The existing gh18349 index `c_uom_conversion_product` (`UNIQUE (c_uom_id, c_uom_to_id, m_product_id) WHERE isactive='Y'`) already enforces product-scope uniqueness AND is the by-product load access path, so it now stays untouched. The migration only ADDs what was actually missing: the generic-scope unique index `c_uom_conversion_generic_uq` (closes the NULL-distinctness gap the product index can't catch) + the identity `CHECK`. In 5812080 the product-scope `AD_Index_Table` error message now points at the existing `c_uom_conversion_product` (WhereClause + column order matched to it). Verified on the local deep_tundra stack: existing index preserved, generic index + CHECK created, idempotent re-apply, CHECK blocks identity, fail-loud fires on dirty data. Fix: https://github.com/metasfresh/metasfresh/commit/2a4b5b21df2

PR #24965, opened 2026-07-03
cucumber: combined rabbitMQ drain fixes productionCandidate reactivate-scenario flake

teosarca:
## What

Backport the combined rabbitMQ drain step (`wait until all rabbitMQ queues are empty or throw exception after 5 minutes`) into this branch and use it in the last scenario of `productionCandidate.feature` (reactivate + reduce qty), replacing the material-only drain.

Changes:
- `RabbitMQ_StepDef.java`
  - Generalized the private `waitEmptyMaterialQueue()` helper into `waitEmptyQueueByTopic(String topicName)`.
  - Added `wait until de.metas.async rabbitMQ queue is empty …` step (drains the async-batch queue).
  - Added the combined `wait until all rabbitMQ queues are empty …` step, which drains the material queue **then** the async-batch queue. The ordering is intentional and documented in the method Javadoc.
  - Kept the existing material-only step (`wait until de.metas.material …`) unchanged so other scenarios are unaffected.
- `productionCandidate.feature`
  - Swapped the single post-reactivation drain from the material-only step to the combined step. Qty assertions are unchanged.

## Why

The scenario reactivates an order, drains a queue, then polls up to 60s for the recomputed `PP_Order_Candidate` qty. Reactivation publishes material-dispo events; the material-dispo listeners in turn enqueue async work-packages on the **async-batch** queue, and the qty recompute actually settles at the end of that downstream async-batch stage.

Draining only the material queue does not wait for that downstream stage — the material queue can even read empty simply because the async side hasn't picked up the work yet. The 60s poll was a best-effort cushion that usually, but not always, outlasted the async lag, producing an intermittent failure where the candidate stayed at a stale intermediate qty past the poll deadline.

The combined-drain step waits for the material queue **and then** the async-batch queue to be provably empty before the assertion runs, so the value can no longer be read mid-flight. This is the same resilient pattern already in use upstream; this change brings it to this branch (where the feature and the flake both live, and which was missing the step plus its helper).

No assertion is weakened — only the wait becomes deterministic.

### DIFF
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/production/productionCandidate.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/production/productionCandidate.feature
@@ -188,7 +188,7 @@ Feature: Production dispo scenarios
 
     And the order identified by o_2 is reactivated
 
-    And wait until de.metas.material rabbitMQ queue is empty or throw exception after 5 minutes
+    And wait until all rabbitMQ queues are empty or throw exception after 5 minutes
 
     Then after not more than 60s, PP_Order_Candidates are found
       | Identifier | M_Product_ID | PP_Product_BOM_ID | PP_Product_Planning_ID | S_Resource_ID | QtyEntered | QtyToProcess | QtyProcessed | DatePromised         | DateStartSchedule    | OPT.IsClosed | OPT.Processed |
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/production/productionCandidate.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/production/productionCandidate.feature
@@ -188,8 +188,6 @@ Feature: Production dispo scenarios
 
     And the order identified by o_2 is reactivated
 
-    And wait until all rabbitMQ queues are empty or throw exception after 5 minutes
-
     Then after not more than 60s, PP_Order_Candidates are found
       | Identifier | M_Product_ID | PP_Product_BOM_ID | PP_Product_Planning_ID | S_Resource_ID | QtyEntered | QtyToProcess | QtyProcessed | DatePromised         | DateStartSchedule    | OPT.IsClosed | OPT.Processed |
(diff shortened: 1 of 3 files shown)


backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/production/productionCandidate.feature
```diff
@@ -188,7 +188,7 @@ Feature: Production dispo scenarios
 
     And the order identified by o_2 is reactivated
 
-    And wait until de.metas.material rabbitMQ queue is empty or throw exception after 5 minutes
+    And wait until all rabbitMQ queues are empty or throw exception after 5 minutes
```
teosarca: why not calling wait_empty_all_queues as part of the  "after not more than 60s, PP_Order_Candidates are found" step.
and drop the " And wait until all rabbitMQ queues are empty or throw exception after 5 minutes" here.

IMHO, technical steps shall be avoided as much as possible. What a QA reader/non-technical person will understnd from this?

capture this learning + review rule.

teosarca: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/6ec707eb20a — folded the drain into the consuming `PP_Order_Candidates are found` step def (`RabbitMQ_StepDef.wait_empty_all_queues()`, material-events then async-batch) and dropped the bare `wait until all rabbitMQ queues are empty …` step here, per module CLAUDE.md rule 7 (drain inside the consuming step, never a bare step in the `.feature`).

The step def is shared by ~9 feature files, so I verified this as an *always-drain* rather than an opt-in variant: audited all 67 call sites — every one asserts either the pre-generation state (`Processed=false`, `QtyProcessed=0`, before any generation trigger) or the settled post-generation state (after an explicit enqueue/generate/auto-create/scheduler). The found-step polls committed DB state, so it never trusts "queue-empty" as its post-condition and there is no assert-intermediate → drain → assert-changed triad on it. Draining first therefore never collapses an asserted value. (The only non-safety effect is the drain's 5-min ceiling vs the 60s poll, which only lengthens how long a genuinely broken run takes to fail.)

Scope kept minimal per request: only this scenario's bare step is removed here; the sibling files' now-redundant bare drains are left for a separate rule-7 cleanup.

PR #24966, opened 2026-07-03
Shipment Schedule export — fix flaky queue-drain dependency

teosarca:
## What

Fix an intermittent (flaky) failure in `shipmentScheduleExport.feature` on `soft_panda_hotfix`, affecting the two scenarios:

- `Export oxid shipment candidate`
- `Export non-oxid shipment schedule from order candidate`

## Why

Both scenarios used `wait until de.metas.material rabbitMQ queue is empty` as a completion proxy before asserting the export result. That proxy is racy: an empty queue does not imply the material-event cascade has committed the schedule into an exportable state. The export GET is synchronous and only exports schedules already visible as exportable, so when it ran too early the run failed intermittently with `ExportStatus` `PENDING` (instead of `EXPORTED`) or an empty candidate list (`size 0`). The same commit was observed both passing and failing on CI, confirming a race rather than a logic bug.

## Fix

In both affected scenarios:

- Drop the racy `wait until de.metas.material rabbitMQ queue is empty` step.
- Poll the committed business state directly via the existing `M_ShipmentSchedules are found ... IsToRecompute=N` step. `IsToRecompute=N` is written only after every cascade recompute workpackage completes, so polling it is race-free by construction.
- Raise that poll ceiling from `60s` to `300s` to absorb a loaded CI executor (it returns as soon as the state is observed, so the happy path is not slowed).

No export assertion is changed or weakened. `ExportStatus=EXPORTED`, the candidate-list size, and every audit check stay identical. This is a feature-file-only change; the polling step already accepts the timeout as a parameter and already queries `IsToRecompute`, so no Java change is needed.

The third scenario in the same file (`Export non-oxid shipment schedule from order`) already used this drain-free, business-state-poll pattern and is left unchanged.

This mirrors the equivalent fix already merged on `new_dawn_uat` (which had a combined-queue drain variant that does not cherry-pick clean onto `soft_panda_hotfix`'s older material-only drain, hence applied by hand).

### DIFF
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipmentSchedule/shipmentScheduleExport.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipmentSchedule/shipmentScheduleExport.feature
@@ -122,9 +122,7 @@ Feature: Shipment schedule export rest-api
       | C_OrderLine_ID.Identifier | C_Order_ID.Identifier | OPT.DateOrdered | M_Product_ID.Identifier | QtyOrdered | qtydelivered | qtyinvoiced | price | discount | currencyCode | processed |
       | orderLine_1               | order_1               | 2022-02-02      | product_25_02           | 1          | 0            | 0           | 10.0  | 0        | EUR          | true      |
 
-    And wait until de.metas.material rabbitMQ queue is empty or throw exception after 5 minutes
-
-    And after not more than 60s, M_ShipmentSchedules are found:
+    And after not more than 300s, M_ShipmentSchedules are found:
       | Identifier | C_OrderLine_ID.Identifier | IsToRecompute |
       | schedule_1 | orderLine_1               | N             |
     And after not more than 60s, validate shipment schedules:
@@ -218,9 +216,7 @@ Feature: Shipment schedule export rest-api
       | C_OrderLine_ID.Identifier | C_Order_ID.Identifier | OPT.DateOrdered | M_Product_ID.Identifier | QtyOrdered | qtydelivered | qtyinvoiced | price | discount | currencyCode | processed |
       | orderLine_1               | order_1               | 2022-02-02      | product_25_02           | 1          | 0            | 0           | 10.0  | 0        | EUR          | true      |
 
-    And wait until de.metas.material rabbitMQ queue is empty or throw exception after 5 minutes
-
-    And after not more than 60s, M_ShipmentSchedules are found:
+    And after not more than 300s, M_ShipmentSchedules are found:
       | Identifier | C_OrderLine_ID.Identifier | IsToRecompute |
       | schedule_1 | orderLine_1               | N             |
     And after not more than 60s, validate shipment schedules:
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipmentSchedule/shipmentScheduleExport.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipmentSchedule/shipmentScheduleExport.feature
@@ -122,7 +122,7 @@ Feature: Shipment schedule export rest-api
       | C_OrderLine_ID.Identifier | C_Order_ID.Identifier | OPT.DateOrdered | M_Product_ID.Identifier | QtyOrdered | qtydelivered | qtyinvoiced | price | discount | currencyCode | processed |
       | orderLine_1               | order_1               | 2022-02-02      | product_25_02           | 1          | 0            | 0           | 10.0  | 0        | EUR          | true      |
 
-    And after not more than 120s, M_ShipmentSchedules are found:
+    And after not more than 300s, M_ShipmentSchedules are found:
       | Identifier | C_OrderLine_ID.Identifier | IsToRecompute |
       | schedule_1 | orderLine_1               | N             |
     And after not more than 60s, validate shipment schedules:
@@ -216,7 +216,7 @@ Feature: Shipment schedule export rest-api
       | C_OrderLine_ID.Identifier | C_Order_ID.Identifier | OPT.DateOrdered | M_Product_ID.Identifier | QtyOrdered | qtydelivered | qtyinvoiced | price | discount | currencyCode | processed |
       | orderLine_1               | order_1               | 2022-02-02      | product_25_02           | 1          | 0            | 0           | 10.0  | 0        | EUR          | true      |
 
-    And after not more than 120s, M_ShipmentSchedules are found:
+    And after not more than 300s, M_ShipmentSchedules are found:
       | Identifier | C_OrderLine_ID.Identifier | IsToRecompute |
       | schedule_1 | orderLine_1               | N             |
     And after not more than 60s, validate shipment schedules:
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipmentSchedule/shipmentScheduleExport.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipmentSchedule/shipmentScheduleExport.feature
@@ -122,7 +122,7 @@ Feature: Shipment schedule export rest-api
       | C_OrderLine_ID.Identifier | C_Order_ID.Identifier | OPT.DateOrdered | M_Product_ID.Identifier | QtyOrdered | qtydelivered | qtyinvoiced | price | discount | currencyCode | processed |
       | orderLine_1               | order_1               | 2022-02-02      | product_25_02           | 1          | 0            | 0           | 10.0  | 0        | EUR          | true      |
 
-    And after not more than 300s, M_ShipmentSchedules are found:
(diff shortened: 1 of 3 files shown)


backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipmentSchedule/shipmentScheduleExport.feature
```diff
@@ -122,9 +122,7 @@ Feature: Shipment schedule export rest-api
       | C_OrderLine_ID.Identifier | C_Order_ID.Identifier | OPT.DateOrdered | M_Product_ID.Identifier | QtyOrdered | qtydelivered | qtyinvoiced | price | discount | currencyCode | processed |
       | orderLine_1               | order_1               | 2022-02-02      | product_25_02           | 1          | 0            | 0           | 10.0  | 0        | EUR          | true      |
 
-    And wait until de.metas.material rabbitMQ queue is empty or throw exception after 5 minutes
-
-    And after not more than 60s, M_ShipmentSchedules are found:
+    And after not more than 300s, M_ShipmentSchedules are found:
```
teosarca, line 125: Changing from 60s to 300s isn't it too much? 
don't we have any other way, something more predictible?

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipmentSchedule/shipmentScheduleExport.feature
```diff
@@ -218,9 +216,7 @@ Feature: Shipment schedule export rest-api
...
```
teosarca, line 219: Changing from 60s to 300s isn't it too much? 
don't we have any other way, something more predictible?

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipmentSchedule/shipmentScheduleExport.feature
teosarca: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/fbf65aad792 — replaced the raised poll ceiling with an *explicit* wait on the actual completion signal, per your "something more predictable".

The `M_ShipmentSchedules are found` step now internalizes the ordered material→async rabbitMQ drain (`wait_empty_all_queues`) so the async shipment-schedule recompute chain settles deterministically before it polls the committed `IsToRecompute` state — drain kept out of the feature file per module CLAUDE.md rule 7. Ceiling dropped **300s→120s**: the drain front-loads settling, and the committed-state re-poll still guards the small commit-visibility gap. (I didn't go to a hard 60s because the queue-empty signal (`getMessageCount`) can return early on a multi-wave invalidate→recompute chain — that early-return is exactly what made #24605 drop the drain for a generous ceiling; 120s keeps headroom while restoring the explicit wait.)

Verified safe across **all 319 uses** of this step: every one asserts the settled state (`IsToRecompute=N`/blank), none the invalidated (`Y`) state — so front-loading the drain cannot collapse any asserted value.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipmentSchedule/shipmentScheduleExport.feature
teosarca: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/fbf65aad792 — replaced the raised poll ceiling with an *explicit* wait on the actual completion signal, per your "something more predictable".

The `M_ShipmentSchedules are found` step now internalizes the ordered material→async rabbitMQ drain (`wait_empty_all_queues`) so the async shipment-schedule recompute chain settles deterministically before it polls the committed `IsToRecompute` state — drain kept out of the feature file per module CLAUDE.md rule 7. Ceiling dropped **300s→120s**: the drain front-loads settling, and the committed-state re-poll still guards the small commit-visibility gap. (I didn't go to a hard 60s because the queue-empty signal (`getMessageCount`) can return early on a multi-wave invalidate→recompute chain — that early-return is exactly what made #24605 drop the drain for a generous ceiling; 120s keeps headroom while restoring the explicit wait.)

Verified safe across **all 319 uses** of this step: every one asserts the settled state (`IsToRecompute=N`/blank), none the invalidated (`Y`) state — so front-loading the drain cannot collapse any asserted value.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipmentSchedule/shipmentScheduleExport.feature
teosarca: **Resolution (dev) — correction to my earlier reply:** the drain-based fix was reverted in f8b5def082b because it caused a regression. Internalizing `wait_empty_all_queues()` into this shared step drains **all** rabbitMQ queues globally on the shared cucumber executor, whose shipment-schedule revalidation queue churns continuously — that starved the committed-`IsToRecompute` poll and deterministically broke `nShift Carrier Advise` (green on base, **red 3/3** with the drain across two base branches). So the drain is not viable here.

Reverted to the committed-`IsToRecompute` **300s** poll. That IS the "more predictable" signal you were after: it's an early-exit poll on the actual committed business flag — it exits in seconds on a healthy run, re-checking committed DB state each pass — **not a fixed sleep**. The 300s is only the worst-case ceiling that covers the async recompute lag PR #24605 addressed; lowering it risks re-introducing that flake. Net: the 300s committed-poll is the correct approach; my attempt to "improve" it with a drain was wrong.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipmentSchedule/shipmentScheduleExport.feature
teosarca: **Resolution (dev) — correction to my earlier reply:** the drain-based fix was reverted in f8b5def082b because it caused a regression. Internalizing `wait_empty_all_queues()` into this shared step drains **all** rabbitMQ queues globally on the shared cucumber executor, whose shipment-schedule revalidation queue churns continuously — that starved the committed-`IsToRecompute` poll and deterministically broke `nShift Carrier Advise` (green on base, **red 3/3** with the drain across two base branches). So the drain is not viable here.

Reverted to the committed-`IsToRecompute` **300s** poll. That IS the "more predictable" signal you were after: it's an early-exit poll on the actual committed business flag — it exits in seconds on a healthy run, re-checking committed DB state each pass — **not a fixed sleep**. The 300s is only the worst-case ceiling that covers the async recompute lag PR #24605 addressed; lowering it risks re-introducing that flake. Net: the 300s committed-poll is the correct approach; my attempt to "improve" it with a drain was wrong.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipmentSchedule/shipmentScheduleExport.feature
teosarca: **Resolution (dev):** Final answer to "300s too much / something more predictable?" — the drain-based attempt was reverted (f8b5def082b) because it regressed `nShift Carrier Advise`: internalizing `wait_empty_all_queues()` drains ALL rabbitMQ queues on the shared cucumber executor, whose shipment-schedule revalidation churns continuously, starving the committed-`IsToRecompute` poll (green on base, red 3/3 with the drain across two bases). Kept the committed-`IsToRecompute` 300s poll — that IS the predictable signal: an early-exit poll on the actual committed business flag (exits in seconds on a healthy run, re-checking committed DB state each pass; not a fixed sleep). The 300s is only the worst-case ceiling covering the async recompute lag PR #24605 addressed; lowering it risks re-introducing that flake. Net: 300s committed-poll is the correct, deterministic approach.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipmentSchedule/shipmentScheduleExport.feature
teosarca: **Resolution (dev):** Final answer to "300s too much / something more predictable?" — the drain-based attempt was reverted (f8b5def082b) because it regressed `nShift Carrier Advise`: internalizing `wait_empty_all_queues()` drains ALL rabbitMQ queues on the shared cucumber executor, whose shipment-schedule revalidation churns continuously, starving the committed-`IsToRecompute` poll (green on base, red 3/3 with the drain across two bases). Kept the committed-`IsToRecompute` 300s poll — that IS the predictable signal: an early-exit poll on the actual committed business flag (exits in seconds on a healthy run, re-checking committed DB state each pass; not a fixed sleep). The 300s is only the worst-case ceiling covering the async recompute lag PR #24605 addressed; lowering it risks re-introducing that flake. Net: 300s committed-poll is the correct, deterministic approach.

PR #24984, opened 2026-07-04
mobile-webui: content-based completion for long HU QR keyboard scans

teosarca:
## Problem
On the mobile WebUI, long handling-unit QR codes (`HU#<v>#{…json…}`) scanned via the hardware keystroke reader were intermittently reported as "QR code not recognized". Root cause: `useKeyboardBarcodeReader` completed a scan purely on an inter-keystroke time gap. When a long QR's keystrokes arrive spread over several seconds (slow/chunked delivery on some handhelds), any gap ≥ the debounce split one physical scan into multiple fragments; each fragment failed to parse.

Verified: the added test `src/hooks/__tests__/useKeyboardBarcodeReader.test.js` reproduces it deterministically — feeding a long `HU#…#{json}` code's keystrokes with a single inter-keystroke gap ≥ the debounce makes the pre-fix hook emit two fragment `onReadDone` calls instead of one (RED); the fix makes it one (GREEN).

## Fix — content-based, delivery-speed-independent completion
- New completeness check `checkPartialScannedCode(buffer)` (registry-dispatched, extensible) returning a `ScanCompleteness` tri-state: `NOT_APPLICABLE` / `PARTIAL_SCAN` / `COMPLETE_SCAN`.
- `checkPartialHUScannedCode` parses the payload after `HU#<v>#` with `JSON.parse`.
- Hook behaviour: `COMPLETE_SCAN` → complete immediately; `PARTIAL_SCAN` → hold back another idle round (do not flush a still-arriving code); `NOT_APPLICABLE` → unchanged prior timing. The Enter/Tab terminator still force-completes.
- **Contract invariant:** a check returns `COMPLETE_SCAN` only when the code is unambiguously terminal (no continuation could yield a different valid code) — a complete `HU#<v>#{json}` qualifies (and completing there cleanly separates back-to-back scans). Non-terminal formats must never return it.
- No completeness check registered ⇒ pure prior timing behaviour (zero blast radius for existing scan paths).

## Stuck / truncated buffer handling
- A genuinely stuck or truncated partial buffer (JSON that never closes) is flushed and surfaced through the normal error path, so it no longer lingers and no longer merges into — and corrupts — the next scan.
- The abandon window is a fixed floor **decoupled from `debounceMillis`** (previously it scaled with the debounce, so lowering the debounce shrank the abandon ceiling into the legitimate chunk-gap range and could re-split slow scans).

## Also in this PR
- Fixed stale Jest expectations in the qrCode util suites (were asserting outdated parser output).
- Added a Playwright E2E (`e2e/mobile-webui/tests/spec/picking/scan_HU_QR_chunked.spec.js`) covering a chunked HU-QR scan end-to-end, plus a gap-capable `BarcodeScannerComponent.type()` path to drive it.
- `numbers.js`: use a regex `replace` instead of `String.replaceAll` (node:14 build compatibility).

## Tests
New Jest coverage for the hook + the qrCode utils (gap-independent assembly, partial hold-back, terminal force-complete, Enter/Tab completion, plain-barcode idle flush, abandon-ceiling, and surfacing of a stuck / short partial rather than dropping or merging it). Plus the chunked-scan Playwright E2E above.

### DIFF
--- a/docker-builds/Dockerfile.mobile
+++ b/docker-builds/Dockerfile.mobile
@@ -19,10 +19,27 @@ RUN yarn install
 
 COPY misc/services/mobile-webui/mobile-webui-frontend/ .
 RUN yarn lint
-# currently not terminating: RUN yarn test --ci --runInBand
 
 ENV PUBLIC_URL=/
 RUN yarn build
+
+
+# Unit-test stage: run the mobile-webui Jest suite once and capture its log + exit code for the CI
+# `test-mobile` job to assert on. Kept out of the default (runtime) target so `build-mobile` stays
+# fast and unaffected. Replaces the previously-disabled `RUN yarn test` ("currently not
+# terminating"): CI=true forces react-scripts to run once (not watch) and --forceExit guarantees
+# the process exits even if a handle leaks. The `| tee ... && echo $?` keeps the RUN itself green
+# so a red test does not fail the image build — the CI job decides based on jest.exit-code.
+FROM build AS test
+
+SHELL ["/bin/bash", "-c"]
+
+RUN CI=true yarn test --ci --runInBand --bail=0 --detectOpenHandles --forceExit 2>&1 | tee jest.log && echo "${PIPESTATUS[0]}" > jest.exit-code
+
+VOLUME /reports
+WORKDIR /reports/mobile
+
+CMD cp /app/jest.* .
 
 
 FROM nginx:1.21.6 AS runtime
--- a/docker-builds/Dockerfile.mobile
+++ b/docker-builds/Dockerfile.mobile
@@ -19,27 +19,10 @@ RUN yarn install
 
 COPY misc/services/mobile-webui/mobile-webui-frontend/ .
 RUN yarn lint
+# currently not terminating: RUN yarn test --ci --runInBand
 
 ENV PUBLIC_URL=/
 RUN yarn build
-
-
-# Unit-test stage: run the mobile-webui Jest suite once and capture its log + exit code for the CI
-# `test-mobile` job to assert on. Kept out of the default (runtime) target so `build-mobile` stays
-# fast and unaffected. Replaces the previously-disabled `RUN yarn test` ("currently not
-# terminating"): CI=true forces react-scripts to run once (not watch) and --forceExit guarantees
-# the process exits even if a handle leaks. The `| tee ... && echo $?` keeps the RUN itself green
-# so a red test does not fail the image build — the CI job decides based on jest.exit-code.
-FROM build AS test
-
-SHELL ["/bin/bash", "-c"]
-
-RUN CI=true yarn test --ci --runInBand --bail=0 --detectOpenHandles --forceExit 2>&1 | tee jest.log && echo "${PIPESTATUS[0]}" > jest.exit-code
-
-VOLUME /reports
-WORKDIR /reports/mobile
-
-CMD cp /app/jest.* .
 
 
 FROM nginx:1.21.6 AS runtime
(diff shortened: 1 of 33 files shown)


docker-builds/Dockerfile.mobile
```diff
@@ -19,12 +19,29 @@ RUN yarn install
...
 RUN yarn build
 
 
+# Unit-test stage: run the mobile-webui Jest suite once and capture its log + exit code for the CI
+# `test-mobile` job to assert on. Kept out of the default (runtime) target so `build-mobile` stays
+# fast and unaffected. Replaces the previously-disabled `RUN yarn test` ("currently not
+# terminating"): CI=true forces react-scripts to run once (not watch) and --forceExit guarantees
+# the process exits even if a handle leaks. The `| tee ... && echo $?` keeps the RUN itself green
+# so a red test does not fail the image build — the CI job decides based on jest.exit-code.
+FROM build AS test
```
teosarca: we agreed to not have the testing part of docker image !?

teosarca: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/196d9c313af31387494e766e7c56233c98255146 — reverted `docker-builds/Dockerfile.mobile` and `.github/workflows/cicd.yaml` to the base state (both byte-identical to base). No `FROM build AS test` stage in the image and no `test-mobile` CI job. The QR fix keeps its own Jest tests plus the Playwright E2E on the pre-existing mobile-webui E2E path; the mobile Jest suite is no longer wired into the Docker build.

e2e/mobile-webui/tests/spec/picking/picking.spec.js
```diff
@@ -612,14 +614,56 @@ test('Scan invalid HU QR code and recover', async ({ page }) => {
...
+// It must be comfortably SHORT: expectErrorToast() gives the assertion only ~2 s of grace after pickHU()
+// returns (which, on the direct-scan HEAD path, is almost immediately — the scan dispatch is synchronous
+// and no qty dialog opens). The abandon flush → backend round-trip → error-toast render must all land
+// inside that budget under CI load, so the window is set well below it. The hook's idle interval ticks
+// every rateMs*2 (rateMs = the 300 ms debounce default here → 600 ms), so 500 ms fires reliably on the
+// first tick (~600 ms) with large margin, and still sits above rateMs so it never trips the normal
+// debounce flush. (The window is deliberately tiny for the test ONLY because this HEAD is dispatched as
+// ONE synchronous keystroke burst with zero inter-chunk gaps — there is no slow chunked scan here to
+// protect; production keeps the 15000 ms default that clears the real 3-8 s inter-chunk gaps.)
+const FAST_ABANDON_MS = 500;
```
teosarca, line 656: why overriding the production/standard ?

teosarca: **Resolution (dev):** It overrides the window only for THIS test (via `mobileui.frontend.barcodeScanner.inputText.idleAbandonMillis`), never production — prod keeps the 15000 ms default. Reason: with no Enter terminator a truncated HU *head* is held as a partial and only surfaces its error once the idle-abandon window elapses; at the 15000 ms default the E2E would sit 15 s, so the test lowers it to 500 ms so the error lands inside the toast-watcher window.

But your question is the right one, and running it locally just made it concrete: this override only matters for the no-terminator HEAD case, and it exposes that **in production (no Enter) a genuinely-truncated scan takes ~15 s to show "not recognized"** — a real UX point now that the customer has removed the Enter suffix (it broke login, ic114#3689). I'm reworking these tests against that no-Enter reality and verifying locally. Two open items I'll bring to you rather than decide alone: (1) whether that 15 s no-Enter error-latency needs lowering the production `idleAbandonMillis` (trade-off: re-splitting slow 3–8 s chunked scans); (2) whether the no-terminator HEAD case is better left to the hook unit tests (which already cover the abandon path deterministically) instead of an E2E that must override prod config to run fast.

PR #24985, opened 2026-07-04
ProductLookupDescriptor: resolve IsSold/IsPurchased gate client/org from the request context (fixes product-picker leak)

teosarca:
## Bug

The order-line product picker could leak a product that should be blocked by the purchase/sales enforcement (e.g. a purchase-only product still offered on a sales order line) when the `M_Product_EnforcePurchaseSalesFlags` SysConfig gate is enabled.

## Root cause

`ProductLookupDescriptor.appendFilterByOrderType(...)` evaluates the enforcement gate using `param_AD_Client_ID` read from the lookup `evalCtx`. But the `BuilderWithoutStockInfo` path did **not** register `param_AD_Client_ID` in `ctxNamesNeededForQuery`, so the framework never seeded it into the evalCtx on that path — `param_AD_Client_ID.getValueAsInteger(evalCtx)` fell back to the declared default `-1`. `isPurchaseSalesEnforcementEnabled(...)` was then evaluated at a non-existent client, matched no real gate row, and the `IsSold` / `IsPurchased` `EXISTS` filter was silently skipped.

## Fix

Register `param_AD_Client_ID` in `BuilderWithoutStockInfo.ctxNamesNeededForQuery` (the `BuilderWithStockInfo` path already had it), and resolve `clientId` / `orgId` in `appendFilterByOrderType` from the request `evalCtx`:

```java
final ClientId clientId = ClientId.ofRepoId(param_AD_Client_ID.getValueAsInteger(evalCtx));
final OrgId    orgId    = OrgId.ofRepoId(param_AD_Org_ID.getValueAsInteger(evalCtx));
```

This is the same request-scoped resolution the sibling `explodeRecordsWithStockQuantities` call already uses — the evalCtx now carries the logged-in client/org on both builder paths, so the gate matches the correct SysConfig row. No `Env` thread-local read (per coding-rule §7 — a lookup descriptor must take client/org from the request context, not ambient `Env`).

## Verification

Local from-source reactor build of `de.metas.ui.web.base` (+ upstream `de.metas.*`): BUILD SUCCESS. Covered by the E2E spec `e2e/frontend-webui/tests/spec/product-purchase-sales-gate.spec.js`.

### DIFF
--- a/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/descriptor/sql/ProductLookupDescriptor.java
+++ b/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/descriptor/sql/ProductLookupDescriptor.java
@@ -226,7 +226,7 @@ public class ProductLookupDescriptor implements LookupDescriptor, LookupDataSour
 
 		this.excludeBOMProducts = excludeBOMProducts;
 
-		ctxNamesNeededForQuery = ImmutableSet.of(param_C_BPartner_ID, param_M_PriceList_ID, param_PricingDate, param_AD_Org_ID);
+		ctxNamesNeededForQuery = ImmutableSet.of(param_C_BPartner_ID, param_M_PriceList_ID, param_PricingDate, param_AD_Org_ID, param_AD_Client_ID);
 
 		searchStringMinLength = adTablesRepo.getTypeaheadMinLength(org.compiere.model.I_M_Product.Table_Name);
 		this.isFallbackToBasePricelist = CoalesceUtil.coalesceNotNull(isFallbackToBasePricelist, Boolean.TRUE);
@@ -688,7 +688,7 @@ public class ProductLookupDescriptor implements LookupDescriptor, LookupDataSour
 		appendFilterByPriceList(sqlWhereClause, sqlWhereClauseParams, evalCtx);
 		appendFilterByNotFreightCostProduct(sqlWhereClause, sqlWhereClauseParams, evalCtx);
 		appendFilterByOrg(sqlWhereClause, sqlWhereClauseParams, evalCtx);
-		appendFilterByOrderType(sqlWhereClause, sqlWhereClauseParams);
+		appendFilterByOrderType(sqlWhereClause, sqlWhereClauseParams, evalCtx);
 		appendFilterBOMProducts(sqlWhereClause, sqlWhereClauseParams);
 
 		//
@@ -798,15 +798,16 @@ public class ProductLookupDescriptor implements LookupDescriptor, LookupDataSour
 
 	private void appendFilterByOrderType(
 			@NonNull final StringBuilder sqlWhereClause,
-			@NonNull final SqlParamsCollector sqlWhereClauseParams)
+			@NonNull final SqlParamsCollector sqlWhereClauseParams,
+			@NonNull final LookupDataSourceContext evalCtx)
 	{
 		if (restrictByOrderType == null)
 		{
 			return;
 		}
-		// The evalCtx does not carry the logged-in client/org, so resolve the enforcement gate from the session Env.
-		final ClientId clientId = Env.getClientId();
-		final OrgId orgId = Env.getOrgId();
+		// Resolve the enforcement gate from the request context's client/org.
+		final ClientId clientId = ClientId.ofRepoId(param_AD_Client_ID.getValueAsInteger(evalCtx));
+		final OrgId orgId = OrgId.ofRepoId(param_AD_Org_ID.getValueAsInteger(evalCtx));
 		if (!productBL.isPurchaseSalesEnforcementEnabled(clientId, orgId))
 		{
 			return;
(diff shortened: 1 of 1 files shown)


backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/descriptor/sql/ProductLookupDescriptor.java
```diff
@@ -798,15 +798,15 @@ private void appendFilterBOMProducts(final StringBuilder sqlWhereClause, final S
...
 	{
 		if (restrictByOrderType == null)
 		{
 			return;
 		}
-		final ClientId clientId = ClientId.ofRepoId(param_AD_Client_ID.getValueAsInteger(evalCtx));
-		final OrgId orgId = OrgId.ofRepoId(param_AD_Org_ID.getValueAsInteger(evalCtx));
+		// The evalCtx does not carry the logged-in client/org, so resolve the enforcement gate from the session Env.
+		final ClientId clientId = Env.getClientId();
+		final OrgId orgId = Env.getOrgId();
```
teosarca: pls avoid using Env.getClientId() and Env.getOrgId() in BLs.

teosarca: in this particular case i prefer to use the sysconfgi on client/org=system them having this mambo jambo

teosarca: but first, before u do any change pls check/explain/adapt to why on line ~620 works:
```java
final LookupValuesList explodedLookupValues = explodeRecordsWithStockQuantities(
					unexplodedLookupValues,
					bpartnerId,
					stockDateOrNull,
					warehouseId,
					adLanguage,
					ClientId.ofRepoId(param_AD_Client_ID.getValueAsInteger(evalCtx)),
					OrgId.ofRepoId(param_AD_Org_ID.getValueAsInteger(evalCtx)));
```

teosarca: **Resolution (dev):** Investigated as asked — no code change yet.

**Why line ~620 works:** `explodeRecordsWithStockQuantities` reads the client from the request context — `param_AD_Client_ID.getValueAsInteger(evalCtx)` — and its builder path (`BuilderWithStockInfo`) registers `param_AD_Client_ID` in `ctxNamesNeededForQuery`, so the evalCtx actually carries a real client id. The gate bug is that the *other* builder (`BuilderWithoutStockInfo`) omits `param_AD_Client_ID` from its required set, so in that path the key is absent from the evalCtx and `getValueAsInteger` falls back to the declared default `-1`.

**So the `Env.getClientId()/getOrgId()` fix was wrong.** It reads ambient thread-local context deep in the lookup stack — exactly what coding-rule §7 forbids ("no low-level `Env.get*` inside services/BL; resolve client/org from the data you were given, or accept it as a parameter; `Env.get*` only at the near-user boundary"). It's also fragile on async/pool threads. I agree it should never have shipped as the fix.

**Two correct options, both avoiding `Env`:**
- **(a) Read from the request context** — add `param_AD_Client_ID` to `BuilderWithoutStockInfo`'s `ctxNamesNeededForQuery` and read client/org from `evalCtx` in `appendFilterByOrderType`, mirroring the working line-~620 call. Keeps per-client capability.
- **(b) Read the gate sysconfig at system `(0,0)`** — your stated preference. Correct if the enforcement flag is a system-wide singleton (§7 option 3). I'll verify the sysconfig's `ConfigurationLevel` first; if it's genuinely system-level, (b) is the simpler, right choice and I'll drop the client-resolution entirely.

Planning to implement **(b)** per your preference, pending that `ConfigurationLevel` check — tell me if you'd rather I keep per-client capability via (a).

**Review rule:** capturing this — extending coding-rule §7's scope (currently `@Service`/`@Component`/`@Repository`/`*BL`/command) to explicitly cover `LookupDescriptor` / `LookupDataSourceFetcher` implementations, since they are the same risk class.

teosarca: **Resolution (dev):** Implemented option (a) in https://github.com/metasfresh/metasfresh/commit/0f453291734 — no `Env`.

- Added `param_AD_Client_ID` to `BuilderWithoutStockInfo`'s `ctxNamesNeededForQuery` (the `BuilderWithStockInfo` path already had it), so the framework carries the logged-in client id in the evalCtx on that path.
- `appendFilterByOrderType` now reads `ClientId.ofRepoId(param_AD_Client_ID.getValueAsInteger(evalCtx))` / `OrgId.ofRepoId(param_AD_Org_ID.getValueAsInteger(evalCtx))` — the same request-context resolution the sibling `explodeRecordsWithStockQuantities` call uses. Re-threaded `evalCtx` into the method (the earlier commit had dropped it).
- Zero `Env.getClientId()`/`Env.getOrgId()` remain. Compiles (`de.metas.ui.web.base` reactor `test-compile`, BUILD SUCCESS). CI re-running.

Coding-rule §7 (no low-level `Env.get*` client/org) has been extended to explicitly cover `LookupDescriptor` / `LookupDataSourceFetcher`, with an ArchUnit-candidate, so this class of miss is caught in future reviews.

PR #24988, opened 2026-07-04
Async batch — enqueue CheckProcessed work-package only for completion-consuming types

teosarca:
## Problem

`CheckProcessedAsynBatchWorkpackageProcessor` work-packages accumulate without bound. One such work-package is enqueued per `C_Async_Batch`, unconditionally, for every batch type — its only job is to set the persistent `C_Async_Batch.Processed` flag. But for the high-volume plumbing batch types (OLCand / order / shipment processing) nothing consumes that flag. Because these types have `KeepAliveTimeHours=0` (= keep alive forever) and the DLM archive only reaps `Processed='Y'` work-packages, the useless checker work-packages stay pending forever and are re-skipped every polling cycle — pure churn.

## Fix

- **New column `C_Async_Batch_Type.IsCheckProcessed`** (Yes-No, default `N`). Seeded `Y` for exactly the code-verified consumers of `C_Async_Batch.Processed`:
  - `InvoiceCandidate_Processing` — invoice-candidate PDF concatenation
  - `CreateLettersAsync` — serial-letter print
  - `AutomaticallyInvoicePdfPrinting`, `PDFPrinting` — registered note listeners
- **Gate the enqueue** (`AsyncBatchBL.enqueueAsyncBatch`): the `CheckProcessedAsynBatch` work-package is enqueued only when the batch's type satisfies `IsCheckProcessed='Y' OR AD_BoilerPlate_ID>0`. The `AD_BoilerPlate_ID>0` belt keeps the checker firing for any type that has a completion email configured, so no configured side-effect is dropped. New types default `N`, so no future churn without an explicit opt-in.
- **One-time cleanup migration**: idempotent, non-destructive (`Processed='N'` → `'Y'`, no DELETE, `backup_table` first) close of the already-accumulated stuck checker work-packages and their orphan `C_Async_Batch` rows for non-consumer types. The predicate matches the enqueue gate exactly.

## Behaviour guarantee

Synchronous batch flows (shipment / order / invoice / OLCand / POS) are unchanged: they block on the in-JVM work-package counter (`AsyncBatchObserver`), never on `C_Async_Batch.Processed`. The recompute work-package they depend on is a batch member independent of this gate.

## Tests

- `AsyncBatchTypeTest` — unit test of the `isCheckProcessedNeeded()` predicate (checkProcessed / not-needed / boilerplate-belt).
- `AsyncBatchBLTest` — DB-backed integration test asserting real work-package row counts: non-consumer type → 0 checker work-packages, consumer type → 1, boilerplate-belt → 1.
- Full `de.metas.async` unit suite green.

## Also in this PR — a latent shipment-close bug fix

While validating the change, CI surfaced an intermittent failure in the `close_partially_shipped` scenario. Root-cause showed it was **not** caused by this change (the async-batch gate cannot alter shipment aggregation — `M_ShipmentSchedule.C_Async_Batch_ID` is immutable once set, and the checker work-package runs on its own dedicated queue). It exposed a **pre-existing latent bug** in `ShipmentScheduleBL.closePartiallyShipped_ShipmentSchedules`: it decided "fully shipped" from the *current* `M_InOut`'s line quantity only, so when one order shipped as ≥2 `M_InOut` documents, a schedule already fully delivered by a **sibling** shipment was wrongly closed.

Fix: a schedule is now spared from closing when its **cumulative** delivered quantity (persisted `QtyDelivered`, which reflects sibling shipments, **plus** the current shipment's just-shipped `MovementQty`) meets its **effective** `QtyOrdered` — split-safe. Covered by new unit tests (RED-verified).

## Note on the model regen

Regenerating `I_/X_C_Async_Batch_Type` for the new column also picked up pre-existing drift unrelated to this change (the generator always reflects the full current `AD_Column` state, not a targeted diff): a long-missing `Description` getter/setter and a stale `NotificationType` javadoc corrected to the DB's actual `en_US` translation. No behaviour change.

### DIFF
--- a/backend/de.metas.async/src/main/java/de/metas/async/api/impl/AsyncBatchBL.java
+++ b/backend/de.metas.async/src/main/java/de/metas/async/api/impl/AsyncBatchBL.java
@@ -168,6 +168,14 @@ public class AsyncBatchBL implements IAsyncBatchBL
 	@Override
 	public void enqueueAsyncBatch(@NonNull final AsyncBatchId asyncBatchId)
 	{
+		final I_C_Async_Batch asyncBatchRecord = asyncBatchDAO.retrieveAsyncBatchRecordOutOfTrx(asyncBatchId);
+		final Optional<AsyncBatchType> asyncBatchType = getAsyncBatchType(asyncBatchRecord);
+		if (asyncBatchType.isPresent() && !asyncBatchType.get().isCheckProcessedNeeded())
+		{
+			// the batch's type has no consumer of the Processed flag (no IsCheckProcessed and no boilerplate) -> skip enqueuing the CheckProcessed WP
+			return;
+		}
+
 		final Properties ctx = Env.getCtx();
 		final IWorkPackageQueue queue = workPackageQueueFactory.getQueueForEnqueuing(ctx, CheckProcessedAsynBatchWorkpackageProcessor.class);
 
@@ -468,6 +476,7 @@ public class AsyncBatchBL implements IAsyncBatchBL
 				.keepAlive(extractKeepAlive(record))
 				.skipTimeout(extractSkipTimeout(record))
 				.adBoilderPlateId(record.getAD_BoilerPlate_ID())
+				.checkProcessed(record.isCheckProcessed())
 				.build();
 	}
 
--- a/backend/de.metas.async/src/main/java/de/metas/async/api/impl/AsyncBatchBL.java
+++ b/backend/de.metas.async/src/main/java/de/metas/async/api/impl/AsyncBatchBL.java
@@ -151,6 +151,11 @@ public class AsyncBatchBL implements IAsyncBatchBL
 		try
 		{
 			final I_C_Async_Batch asyncBatch = asyncBatchDAO.retrieveAsyncBatchRecordOutOfTrx(asyncBatchId);
+			if (asyncBatch == null)
+			{
+				// record no longer exists out-of-trx -> nothing to increment
+				return;
+			}
 			final Timestamp processed = SystemTime.asTimestamp();
 			asyncBatch.setLastProcessed(processed);
 			asyncBatch.setLastProcessed_WorkPackage_ID(workPackage.getC_Queue_WorkPackage_ID());
@@ -199,6 +204,12 @@ public class AsyncBatchBL implements IAsyncBatchBL
 	public boolean updateProcessedOutOfTrx(@NonNull final AsyncBatchId asyncBatchId)
 	{
 		final I_C_Async_Batch asyncBatchRecord = asyncBatchDAO.retrieveAsyncBatchRecordOutOfTrx(asyncBatchId);
+		if (asyncBatchRecord == null)
+		{
+			// record no longer exists out-of-trx -> nothing left to wait for; treat as processed so the
+			// CheckProcessed work-package completes instead of re-checking a vanished batch forever.
+			return true;
+		}
 		if (asyncBatchRecord.isProcessed())
 		{
 			return true;
--- a/backend/de.metas.async/src/main/java/de/metas/async/api/impl/AsyncBatchBL.java
+++ b/backend/de.metas.async/src/main/java/de/metas/async/api/impl/AsyncBatchBL.java
@@ -575,6 +575,12 @@ public class AsyncBatchBL implements IAsyncBatchBL
 		try
 		{
 			final I_C_Async_Batch asyncBatch = asyncBatchDAO.retrieveAsyncBatchRecordOutOfTrx(asyncBatchId);
+			if (asyncBatch == null)
+			{
+				// record no longer exists out-of-trx -> nothing to count
+				return 0;
+			}
+
 			final Timestamp enqueued = SystemTime.asTimestamp();
 			if (asyncBatch.getFirstEnqueued() == null)
 			{
--- a/backend/de.metas.async/src/main/java/de/metas/async/api/impl/AsyncBatchBL.java
+++ b/backend/de.metas.async/src/main/java/de/metas/async/api/impl/AsyncBatchBL.java
@@ -121,8 +121,7 @@ public class AsyncBatchBL implements IAsyncBatchBL
 			return;
 		}
 
-		final I_C_Async_Batch asyncBatch = asyncBatchDAO.retrieveAsyncBatchRecordOutOfTrx(asyncBatchId);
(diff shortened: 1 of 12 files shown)


backend/de.metas.async/src/main/java/de/metas/async/api/impl/AsyncBatchBL.java
```diff
@@ -168,6 +168,14 @@ public void increaseProcessed(final I_C_Queue_WorkPackage workPackage)
 	@Override
 	public void enqueueAsyncBatch(@NonNull final AsyncBatchId asyncBatchId)
 	{
+		final I_C_Async_Batch asyncBatchRecord = asyncBatchDAO.retrieveAsyncBatchRecordOutOfTrx(asyncBatchId);
+		final AsyncBatchType asyncBatchType = getAsyncBatchType(asyncBatchRecord).orElse(null);
```
teosarca: introduce getAsyncBatchType(@NonNull final AsyncBatchId asyncBatchId)
handle possible NPE when retrieveAsyncBatchRecordOutOfTrx returns null

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentScheduleBL.java
```diff
@@ -688,8 +690,14 @@ public void closePartiallyShipped_ShipmentSchedules(@NonNull final I_M_InOut ino
...
-		// Track which schedules were fully shipped in this shipment (MovementQty >= QtyOrdered)
-		final HashSet<ShipmentScheduleId> fullyShippedScheduleIds = new HashSet<>();
+		// Sum up how much each schedule got delivered by THIS shipment (MovementQty), per schedule.
+		// Note on timing: this method runs synchronously at TIMING_AFTER_COMPLETE of the M_InOut, but the
+		// schedule's own QtyDelivered is only updated in a runAfterCommit hook (see
+		// M_InOut_Shipment.invalidateShipmentSchedulesForLines), i.e. AFTER this method. So at this point
+		// QtyDelivered does NOT yet reflect the current shipment - it only reflects deliveries from sibling
+		// shipments of the same order that were completed in earlier, already-committed transactions.
+		// We therefore add this shipment's just-shipped MovementQty on top of QtyDelivered below.
+		final HashMap<ShipmentScheduleId, BigDecimal> qtyDeliveredByThisShipmentByScheduleId = new HashMap<>();
```
teosarca, line 700: what are the exact reasons why this code change was necesary here when the overall PR is about avoiding spawing "check processed async batch" workpackages.

pls CSP explain, example, pros/cons

backend/de.metas.async/src/main/java/de/metas/async/api/impl/AsyncBatchBL.java
teosarca: **Resolution (dev):** Done in commit https://github.com/metasfresh/metasfresh/commit/e2b7e6aa5a848c4f04cddf724aeec27ee30de8b6

- Introduced `Optional<AsyncBatchType> getAsyncBatchType(@NonNull AsyncBatchId asyncBatchId)` (interface + impl). It loads the record via `retrieveAsyncBatchRecordOutOfTrx` and returns `Optional.empty()` when that returns `null`, so callers never dereference a null record.
- Refactored `enqueueAsyncBatch` (the reviewed line) and `createNotificationRecord` to use the new overload — they used the record only for the type lookup, so the intermediate nullable record is gone.
- `keepAliveTimeExpired` still needs the record afterwards (`getUpdated()`), so there I kept the record but added an explicit `null` guard (returns `false` on a missing record) — same latent NPE fixed there too.
- Added regression test `AsyncBatchBLTest#nullBatchRecord_doesNotThrowNPE_andEnqueues`: registers a DAO returning `null` (the in-memory `POJOLookupMap` throws instead of returning null, so a mock DAO models the real out-of-trx path) and asserts the overload yields empty and the gate does not NPE.

Unit tests green locally (`AsyncBatchBLTest`, 5/5).

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentScheduleBL.java
teosarca: **Resolution (dev):** Justification — this change is a required companion to the gate, not a drive-by.

**Why it's here (the chain):** the CASE-5 enqueue gate changes async-batch timing, and a side effect is that the picking job's shipment gets split into **2 `M_InOut` documents** instead of 1. That 2-InOut split (pre-existing, possible on its own) trips a **latent bug** in `closePartiallyShipped_ShipmentSchedules`. Without the fix, this PR's own CI goes red — so shipping the gate alone would knowingly turn a latent bug into a frequent one. (Agreed direction with you on 2026-07-04: fix it in this PR, RED-first.)

**The latent bug:** the old code decided "fully shipped" from **only the current InOut's** `MovementQty >= QtyOrdered`. When an order's schedule is delivered across two sibling InOuts, neither one alone reaches `QtyOrdered`, so a *cumulatively fully-delivered* schedule was treated as "not fully shipped" and **closed** — losing the ability to ship the (already-shipped!) remainder.

**The fix:** decide from **cumulative** delivered qty vs **effective** QtyOrdered — persisted `QtyDelivered` (siblings already committed) **plus** this shipment's just-shipped `MovementQty` (not yet in `QtyDelivered`; it's written in an after-commit hook that runs after this method). Split-agnostic: works for 1 or N InOuts.

**Example — order schedule, QtyOrdered = 10:**
- 1 InOut (baseline): MovementQty 10 → `10 >= 10` → stays open. ✓ (both old & new)
- 2 InOuts (with gate): #1 ships 6, #2 ships 4.
  - **Old**, at #2 AFTER_COMPLETE: `4 >= 10`? no → schedule **CLOSED**. ✗ (it's fully delivered)
  - **New**: persisted 6 (from #1) + 4 (this) = `10 >= 10` → **stays open**. ✓

**Pros:** fixes a real data-correctness bug (wrongful close → un-shippable remainder); makes CASE-5 safe to ship; correct for any split degree. Direction-wise it only ever *keeps open* schedules the old code wrongly closed — it never closes anything the old logic kept open.

**Cons:** touches core shipment-close logic (all customers). Mitigated: RED-first TDD with 3 scenario tests (sibling-InOut already-delivered, fully-unshipped, current-InOut-only), and the change is strictly more conservative about closing.

PR #24989, opened 2026-07-06
Mobile Packing: skip-to-floor unpick must detach the CU from the pick-to TU

teosarca:
## Problem

In the Mobile Packing flow, unpicking an item **to the floor** (skipping the target-HU scan) from a package where the item was **picked-to-CU into a TU** did not physically remove the item: the CU was left `HUStatus=Picked` and still nested inside the TU instead of being detached as an active, re-pickable standalone floor HU. Repeating the unpick↔re-pick cycle **accumulated orphan CUs** inside the (reusable) transport unit. The operation returned success and correctly zeroed the picking-job model qty, so the discrepancy was only visible in the physical HU tree.

## Root cause

When a CU is picked **into a TU** (bare-TU pick target, no LU), `PackToHUsProducer.packToHU` returned the TU as a *full* TU with an **empty CU-parts list**. `LUTUResult.isSingleFullTU()` was therefore `true`, so `PickingJobPickCommand.splitOutPickToHUs` recorded the **container TU** — not the leaf CU — as `PickingJobStepPickedToHU.actualPickedHU`. On skip-to-floor unpick, `PickingJobUnPickCommand.unpickWholeHUs → extractToTopLevelHUs` called `HUTransformService.extractToTopLevel` on the **already-top-level TU** (a no-op), and `changeHUStatusFromPickedToActive` flipped only the TU — leaving the child CU Picked and nested.

## Fix (root cause)

- `PackToHUsProducer`: the two CU-into-TU (no-LU) result paths now record the just-packed CU **as a CU part** of the TU (existing-TU path uses the known CU record; the LU/TU-producer path re-wraps the fresh top-level TU with its VHU children via `recordPackedCUsAsParts`). The pick then records the **leaf CU**, so the unpick's extract-to-top-level actually detaches and re-activates it. The other two `packToHU` callers read only TU records and are unaffected.
- `PickingJobPickCommand.toPickingJobStepPickedToHU`: records the CU's **own** QR code (was the container TU's), so `extractToTopLevel` (which asserts the QR is on the extracted HU) succeeds and the move-to-target path can re-derive it.

This reuses the already-correct CU-parts unpick path rather than special-casing the unpick.

## Tests

- **Backend (new):** `PickingJobUnpickToFloorTest.unpickToFloor_cuPickedIntoTU_detachesAndActivatesCU` — picks a CU into a TU, unpicks to the floor (null target), and asserts the CU ends `HUStatus=Active`, is detached from the TU, and the TU holds no orphan CU. Verified RED before the fix (`expected "A" but was "S"`) → green after. Full module suite green (23 run, 0 failures, 1 pre-existing skip).
- **Playwright (new):** `e2e/mobile-webui/tests/spec/picking/picking_partial_unpack_TU_floor.spec.js` — drives the pick-to-CU-into-TU + skip-to-floor flow across 2 unpick↔re-pick rounds, asserting packed qty decrements, qty-to-pick reappears, and the TU retains no orphan each round (multi-round no-accumulation gate).

## Notes

- The backend in-memory harness cannot drive the re-pick step (pick-plan eligibility is not reproducible in-memory), so the multi-round no-accumulation assertion is the Playwright gate; the backend test proves the single-unpick physical outcome.

---

## Follow-up fix — per-COO shipment-line split regression (commit https://github.com/metasfresh/metasfresh/commit/4ce468ebdb8)

An earlier commit in this PR (recording the leaf CU as a CU part) made a no-LU CU-into-TU pick fall out of the `isSingleFullTU()` fast path into `splitOutPickToHUs`'s per-CU loop, where the destination TU was recorded on `M_ShipmentSchedule_QtyPicked` with the container TU's **cumulative** product storage (`getStorage(tu.toHU()).getQuantity(...)`) instead of the current pick's qty. Picking the same product more than once into one bare TU then recorded the running total (e.g. 6 then 15 instead of 6 then 9); `HUShipmentScheduleBL.createCandidatesForQtyPicked` over-counted the cumulative row into a spurious extra per-country-of-origin `M_InOutLine`.

**Symptom:** deterministic failures of `nShiftShipment.feature:677` and `:957` (COO "mixed TUs" scenarios) on every CI run of this branch; green on `deep_tundra_release` base.

**Fix** (`PickingJobPickCommand.splitOutPickToHUs`): record **this pick's** qty (the sum of its CU parts, accumulated per TU) on the TU, not the container's cumulative storage. The picking-job step still records the leaf CU (unpick-to-floor unchanged); the LU path is unchanged.

**Regression coverage:** the pre-existing `nShiftShipment.feature` COO scenarios (including the same-product-double-pick case) are the lock-in — RED pre-fix, green post-fix.

## Minor cleanups
- Corrected stale spec paths in `e2e/mobile-webui/TEST-COVERAGE.md` (`picking/` -> `picking/unpack/`).
- Trimmed the `useKeyboardBarcodeReader` `useLayoutEffect` comment.


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/candidate/commands/PackToHUsProducer.java
```diff
@@ -298,6 +322,59 @@ else if (packToDestination instanceof HUListAllocationSourceDestination)
 		}
 	}
 
+	/**
+	 * Re-wraps freshly-created top-level TUs so each carries its packed CU(s) as CU parts. The LU/TU
```
teosarca: huge AI slop. make sure javadocs are couple of sentences maximum. keep it simple, concise, precise (CSP)

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/candidate/commands/PackToHUsProducer.java
```diff
@@ -298,6 +322,59 @@ else if (packToDestination instanceof HUListAllocationSourceDestination)
...
+
+			final I_M_HU tuRecord = tu.toHU();
+			final List<I_M_HU> vhus = handlingUnitsBL.getVHUs(tuRecord);
+			if (vhus.isEmpty())
+			{
+				tus.add(tu);
+			}
+			else
+			{
+				tus.add(LUTUResult.TU.ofSingleTU(tuRecord, vhus.toArray(new I_M_HU[0])));
```
teosarca, line 362: why only the first one? what happnes to the others?
here a explanatory cmment would be welcome (or a fix if needed)

backend/de.metas.handlingunits.base/src/test/java/de/metas/handlingunits/picking/job/service/commands/unpick/PickingJobUnpickToFloorTest.java
```diff
@@ -0,0 +1,180 @@
...
+ * skip-to-floor unpick. The repeatable unpick&harr;<b>re-pick</b> loop and multi-round no-accumulation
+ * (AC4) are covered by the mobile Playwright E2E
+ * {@code e2e/mobile-webui/tests/spec/picking/picking_partial_unpack_TU_floor.spec.js}: the in-memory HU
+ * harness cannot drive the re-pick step ({@code PickingJobPickCommand#validatePickFromHU} rejects the floor
+ * CU with {@code HU_CANNOT_BE_PICKED_ERROR_MSG} because the running-stack pick-plan / pick-from eligibility
+ * is not reproducible in-memory — see {@code de.metas.handlingunits.base/CLAUDE.md} § "In-memory JUnit does
+ * NOT prove HU flush / materialization").
+ */
+@ExtendWith(AdempiereTestWatcher.class)
+class PickingJobUnpickToFloorTest
```
teosarca, line 67: is this still needed when we have playwright? IMHO is overlapping and more to maintain

e2e/mobile-webui/tests/spec/picking/unpack/picking_partial_unpack_TU_floor.spec.js
```diff
@@ -0,0 +1,156 @@
+import { test } from "../../../playwright.config";
```
teosarca: pls consider introducing "unpack" folder for multiple unpack cases/files.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/candidate/commands/PackToHUsProducer.java
teosarca: **Resolution (dev):** Trimmed `recordPackedCUsAsParts` Javadoc to two sentences (CSP). Commit https://github.com/metasfresh/metasfresh/commit/d1802cf98121c0fd7cfee9d857a943ea082356fe . Capturing the "Javadoc = a couple of sentences, CSP, no AI slop" learning + review-rule into the corpus — queued (the corpus commit is human-gated in this autonomous session).

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/candidate/commands/PackToHUsProducer.java
teosarca: **Resolution (dev):** Not only the first — the loop enriches **every** top-level TU, and `ofSingleTU(tuRecord, vhus…)` records **all** of that TU's VHUs as CU parts (`ofSingleTU` = a non-aggregate TU, not one CU). Added an inline note stating this. Commit https://github.com/metasfresh/metasfresh/commit/d1802cf98121c0fd7cfee9d857a943ea082356fe .

backend/de.metas.handlingunits.base/src/test/java/de/metas/handlingunits/picking/job/service/commands/unpick/PickingJobUnpickToFloorTest.java
teosarca: **Resolution (dev):** Agree it overlaps. Plan: strengthen the Playwright spec to assert the floor-dropped CU ends **Active + detached** at full-stack, then **remove** `PickingJobUnpickToFloorTest`. In that order because right now the JUnit is the only proof of that invariant (the Playwright currently asserts only the TU aggregate qty). The Playwright strengthening is in progress — scoping whether the frontendTesting expect API can assert the floored CU or needs a small extension.

e2e/mobile-webui/tests/spec/picking/unpack/picking_partial_unpack_TU_floor.spec.js
teosarca: **Resolution (dev):** Agreed — will move the unpack specs into `tests/spec/picking/unpack/` (`picking_partial_unpack.spec.js` + `picking_partial_unpack_TU_floor.spec.js`) as part of converging this PR.

teosarca: **Resolution (dev):** Done — moved both specs into `tests/spec/picking/unpack/` (git mv + import-depth fix), commit https://github.com/metasfresh/metasfresh/commit/f159059fd9b . Verified both still list via `playwright test --list` (7 tests, 2 files).

PR #24992, opened 2026-07-06
PO Reactivate — allow when pay schedule has no downstream activity

teosarca:
## Problem
A purchase order with a multi-step payment term (e.g. *25% TT, 25% before delivery, 50% against BL date*) could not be reactivated after completion — even before any goods receipt, invoice, payment, or proforma existed. Reactivation failed with a 500 (`Order_Reactivate_Blocked_By_PaySchedule_Activity`) telling the user to cancel receipts/invoices/payments that do not exist, leaving the order permanently un-editable.

## Root cause
At completion, `OrderSchedulingContext.computeLineContext` sets a pay-schedule line to `Awaiting_Pay` whenever its break's reference date is already known on the order header. An `OrderDate` break resolves to `DateOrdered` (always set on a completed order), so that line is `Awaiting_Pay` immediately, with zero downstream documents. The reactivate guard `C_Order#blockReactivateWhenScheduleNotPending` blocked on `Status != Pending`, conflating "line active" with "downstream activity happened".

## Fix
The guard now blocks reactivation only on **committed downstream state**:
- any pay-schedule line with `M_InOut_ID` (goods receipt) or `C_Invoice_ID` (matched invoice) set, OR
- an existing proforma allocation for the order (`OrderPayScheduleProformaService.getByOrderId` — the LC/proforma row carries no per-line link, so it is detected via the proforma service).

A line that is `Awaiting_Pay`/`Paid` purely because its reference date was known at completion no longer blocks. The reactivate guard's original purpose (protect real downstream state from a drop-and-rebuild re-completion) is preserved. The method name is kept intentionally (renaming would change the `InterceptorEnabled_…` disable-SysConfig key). No change to the pay-schedule status model, due-date logic, or pay-selection.

The block message (AD_Message 545676 + de_DE/de_CH/en_US trls) now also names de-allocating the proforma.

## Tests
- Unit (`C_Order_Test`, 7/7): allow when non-Pending line has no downstream link; block on inout link, invoice link, or proforma allocation.
- Cucumber (`orderReactivatePayScheduleGuard.feature`, 3/3): TC1 reactivate allowed with nothing downstream (regression guard for the reported bug); TC2 proforma allocation still blocks; TC3 goods-receipt link still blocks.
- Migration verified on the local `deep_tundra_release` stack (message text updated across base + all trls).

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/interceptor/C_Order.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/interceptor/C_Order.java
@@ -25,8 +25,9 @@ package de.metas.order.paymentschedule.interceptor;
 import de.metas.i18n.AdMessageKey;
 import de.metas.order.OrderId;
 import de.metas.order.paymentschedule.core.OrderPaySchedule;
-import de.metas.order.paymentschedule.core.OrderPayScheduleStatus;
+import de.metas.order.paymentschedule.core.OrderPayScheduleLine;
 import de.metas.order.paymentschedule.core.service.OrderPayScheduleService;
+import de.metas.order.paymentschedule.referenced_docs.proforma_invoice.OrderPayScheduleProformaService;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import org.adempiere.ad.modelvalidator.annotations.DocValidate;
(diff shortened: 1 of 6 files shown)


backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/interceptor/C_Order.java
```diff
@@ -58,23 +62,29 @@ public class C_Order
...
 				});
 	}
 
-	private boolean hasAnyNonPendingLine(@NonNull final OrderPaySchedule schedule)
+	private boolean reflectsDownstreamActivity(@NonNull final OrderId orderId, @NonNull final OrderPaySchedule schedule)
 	{
-		return schedule.streamLines()
-				.anyMatch(line -> !line.getStatus().isPending());
+		final boolean anyLineLinked = schedule.streamLines()
+				.anyMatch(line -> line.getInoutId() != null || line.getInvoiceId() != null);
```
teosarca: apply demeter law, intoduce line.is....()

also "reflectsDownstreamActivity"... pls respect the 'is'/'has' method prefix convention as much as possible. when i am reading "reflectsDownstreamActivity" have no idea what are u talking about

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/interceptor/C_Order.java
```diff
@@ -58,23 +62,29 @@ public class C_Order
...
+					throw new AdempiereException(MSG_OrderReactivateBlocked).markAsUserValidationError();
 				});
 	}
 
-	private boolean hasAnyNonPendingLine(@NonNull final OrderPaySchedule schedule)
+	private boolean reflectsDownstreamActivity(@NonNull final OrderId orderId, @NonNull final OrderPaySchedule schedule)
 	{
-		return schedule.streamLines()
-				.anyMatch(line -> !line.getStatus().isPending());
+		final boolean anyLineLinked = schedule.streamLines()
```
teosarca: introduce OrderPaySchedule.is....()

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/interceptor/C_Order.java
teosarca: **Resolution (dev):** Applied. Extracted the per-line check into `OrderPayScheduleLine.isLinkedToDownstreamDocument()` (returns `inoutId != null || invoiceId != null`), so the guard no longer reaches into the line's `getInoutId()`/`getInvoiceId()` — Law of Demeter respected. Also renamed the unclear `reflectsDownstreamActivity` → `isBlockedByDownstreamActivity` (is-prefix, states what it decides). https://github.com/metasfresh/metasfresh/commit/fa9452bc374436505ab2ec6d94834667e5edf1ce

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/interceptor/C_Order.java
teosarca: **Resolution (dev):** Added `OrderPaySchedule.hasLineLinkedToDownstreamDocument()` encapsulating the `anyMatch` over the lines; the guard now calls `schedule.hasLineLinkedToDownstreamDocument()` instead of streaming lines itself. I used the `has` prefix (the schedule *has* a line that is linked) since the line-level property is `is`-prefixed (`OrderPayScheduleLine.isLinkedToDownstreamDocument()`) — tell me if you'd rather it be `is…()`. https://github.com/metasfresh/metasfresh/commit/fa9452bc374436505ab2ec6d94834667e5edf1ce

PR #24998, opened 2026-07-07
TableModelLoader.getPOs: chunk large id loads to avoid JDBC 2-byte bind-param limit

teosarca:
## Problem

`InterfaceWrapperHelper.loadByIds(ids, …)` → `POWrapper.loadByIds` → `TableModelLoader.getPOs` builds a single `keyColumn IN (?, ?, … )` with **one bind parameter per id**. When the id set is large, the number of bind parameters exceeds PostgreSQL's wire-protocol limit (the Bind message encodes the parameter count as a signed 2-byte integer, max **32,767**), so the JDBC driver throws before the query runs:

```
java.io.IOException: Tried to send an out-of-range integer as a 2-byte value: <N>
```

surfacing as `DBException: An I/O error occurred while sending to the backend` (SQLState 08006) and dropping the connection. Any bulk `loadByIds` of more than ~32k ids hits this. `getPOs` is the single choke point for all `loadByIds` callers, so the fix belongs there.

## Fix

Chunk the DB load when the id set is large, keeping the common (small) case unchanged:

- New `MAX_IDS_PER_QUERY = 1000` (well under the 32,767 cap; plan-friendly).
- Extracted the existing DB-load body into `loadPOsFromDB(...)`, preserving the per-load `DBException(ex, sql, sqlParams)` wrapping.
- **Fast path — zero overhead for the common case:** for `size <= MAX_IDS_PER_QUERY` the code runs the exact same single query as before (no copy, no partition, no loop — only one O(1) size check, via the extracted `isSingleQueryLoad(int)` predicate). Only larger loads iterate `Iterables.partition(...)` (no full copy of the id set) and concatenate results.

Result semantics are unchanged: `getPOs` has no `ORDER BY` and its caller maps results by id, so chunk-and-concatenate is identical to the single-query result. Each chunk is an independent statement (documented on `loadPOsFromDB`): under READ COMMITTED there is no cross-chunk snapshot consistency for huge loads — a caller needing that must arrange it itself.

## Tests

- `TableModelLoaderChunkingTest` (CI-runnable, no DB): asserts the chunk-size cap stays `<= 32767`, the dispatch decision (`isSingleQueryLoad`) at the `MAX`/`MAX+1` boundary, and the partition invariants (all ids preserved, no chunk over the cap, `ceil(n/max)` chunk count).
- `TableModelLoader_DBTest` (`@Disabled`): end-to-end reproduction that loads > 32,767 rows and asserts all are returned — mirrors this base module's existing manual-run DB-test convention (no automated real-DB harness).

## Scope

Standalone core bugfix — benefits every bulk `loadByIds` caller. Removes the crash; does not change batch sizing/throughput of any specific caller.

### DIFF
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/persistence/TableModelLoader.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/persistence/TableModelLoader.java
@@ -22,7 +22,9 @@ package org.adempiere.ad.persistence;
  * #L%
  */
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
 import de.metas.cache.interceptor.CacheInterceptor;
 import de.metas.cache.model.IModelCacheService;
 import de.metas.logging.LogManager;
@@ -47,6 +49,7 @@ import java.lang.reflect.Constructor;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Properties;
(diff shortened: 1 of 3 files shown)


backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/persistence/TableModelLoader.java
```diff
@@ -64,6 +67,26 @@ public final class TableModelLoader
...
 	private final TableModelClassLoader tableModelClassLoader = TableModelClassLoader.instance;
 
+	/**
+	 * Max number of record IDs to load per "IN (?,?,...)" query, to stay below the PostgreSQL/JDBC
+	 * 2-byte bind-parameter limit (max 32767 params per statement). Larger sets are loaded in chunks.
+	 * Kept conservative (plan-friendly); it only ever engages for loads that would otherwise fail.
+	 * MUST stay &lt;= 32767 or the very bug this guards against returns (see {@code TableModelLoaderChunkingTest}).
+	 */
+	@VisibleForTesting
+	static final int MAX_IDS_PER_QUERY = 1000;
```
teosarca: but if psql limit is ~32k, u can go higher than 1000.

teosarca: **Resolution (dev):** Adopted — raised `MAX_IDS_PER_QUERY` 1000 → 30000 in https://github.com/metasfresh/metasfresh/commit/2471f65ef39d68a80c34aa5752a20050365a00ca . Kept headroom below the 32767 JDBC 2-byte cap (not AT it, so an off-by-one or a future extra bind param cannot reintroduce the overflow) while cutting round-trips ~10× (an ~83k-id load = 3 queries, not 84). `getPOs` binds no params other than the id list, so the ids are the whole budget.

PR #25000, opened 2026-07-07
DDOrderCandidate: optional header aggregation by product (single-line DD_Orders)

teosarca:
## What
Adds an optional DD_Order **header** aggregation dimension: `M_Product_ID`.

`AggregationConfig.aggregateByProductId` — driven by a new sysconfig `DDOrderAggregation.header.byProductId` (default `N`). When enabled, `M_Product_ID` becomes part of the `HeaderAggregationKey`, so `DD_Order_Candidate`s of different products no longer collapse into one large multi-product `DD_Order`. Each product then gets its own **single-line** `DD_Order` (one product → one UOM per order).

## Why
Extends the existing `AggregationConfig` pattern (the three existing `DDOrderAggregation.*` sysconfig toggles that conditionally add a field to the aggregation key). Some deployments produce very large multi-product intra-warehouse replenishment `DD_Order`s — one order with hundreds of lines spanning several UOMs. This flag makes it possible to split those into single-product, single-line orders.

## Behaviour / safety
- **Default `N` → no behaviour change** for any existing instance. Fully opt-in per instance via the sysconfig.
- Change is confined to `de.metas.manufacturing`; `AggregationConfig` has no callers in other modules.

## Tests
`DDOrderCandidateProcessCommandTest`:
- `process_twoProducts_byProductIdDisabled_singleDDOrderWithTwoLines` — control: current behaviour (1 DD_Order, 2 lines).
- `process_twoProducts_byProductIdEnabled_twoSingleLineDDOrders` — two candidates differing only by product → 2 single-line DD_Orders.

## Migration
`5812470_sys_DDOrderAggregation_header_byProductId.sql` — seeds the sysconfig row with default `N`.

### DIFF
--- a/backend/de.metas.manufacturing/src/main/java/de/metas/distribution/ddordercandidate/DDOrderCandidateProcessCommand.java
+++ b/backend/de.metas.manufacturing/src/main/java/de/metas/distribution/ddordercandidate/DDOrderCandidateProcessCommand.java
@@ -326,6 +326,12 @@ class DDOrderCandidateProcessCommand
 		boolean aggregateBySalesOrderId;
 		boolean aggregateByPPOrderRef;
 		boolean aggregateBySalesOrderLineId;
+		/**
+		 * If {@code true}, {@link DDOrderCandidate}s of different products are kept in separate {@link I_DD_Order}s
+		 * (productId becomes part of the {@link HeaderAggregationKey}). This yields one product — and therefore one
+		 * UOM — per DD_Order, i.e. single-line DD_Orders instead of one big multi-product order.
+		 */
+		boolean aggregateByProductId;
 	}
 	//
 	//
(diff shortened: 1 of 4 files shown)


backend/de.metas.manufacturing/src/main/java/de/metas/distribution/ddordercandidate/DDOrderCandidateProcessCommand.java
```diff
@@ -326,6 +326,14 @@ public static class AggregationConfig
 		boolean aggregateBySalesOrderId;
 		boolean aggregateByPPOrderRef;
 		boolean aggregateBySalesOrderLineId;
+		/**
+		 * If {@code true}, {@link DDOrderCandidate}s of different products are kept in separate {@link I_DD_Order}s
```
teosarca: take a look at the other aggregate* booleans. they are pretty obvious, they are without javdoc. only here u thought 1/2 page would help.
pls drop AI slop.
pls ask yourself, what is really relevent from what u wrote?!

teosarca: **Resolution (dev):** Dropped the Javadoc on `aggregateByProductId` so it matches the sibling `aggregate*` booleans (which carry none — the field name is self-documenting). Also removed the parallel inline comment on the sysconfig read and trimmed the migration comment to the standard terse `SysConfig Name`/`Value` form. https://github.com/metasfresh/metasfresh/commit/dc63e7a8000

PR #25024, opened 2026-07-08
EDI INVOIC: exclude packaging-material invoice lines from export, party resolution and validation

metas-ts:
## What

Packaging-material invoice lines (deposit / Leergut, pallets — `C_InvoiceLine.IsPackagingMaterial='Y'`) carry no `C_OrderLine_ID`. On EDI INVOIC export this caused three problems:

- the mandatory-`C_OrderLine_ID` validation rejected the whole invoice;
- the packaging line was exported as an INVOIC line item;
- a spurious **second** delivery-party (`DP`) / `SN` / `BY` address appeared — the order-less line fell through the party-location `CASE` to the invoice's own `C_BPartner_Location_ID`, and `SELECT DISTINCT` emitted it as an extra party row.

## Change

- **`EDIDocumentBL.isValidInvoice`** — exempt packaging-material lines from the mandatory `C_OrderLine_ID` check; and, to keep the boundary case loud, fail with a new user-facing message (`NotExistsNonPackagingInvoiceLineError`) when an invoice's lines are *all* packaging material (which would otherwise pass and export an empty INVOIC).
- **`edi_cctop_invoic_500_v`** — exclude packaging lines from the exported INVOIC `Lines`.
- **`edi_cctop_119_v`** — exclude packaging lines from the `ship` (DP), `snum` (SN) and `cust` (BY) party-location resolution (the `cust` guard is on the `LEFT JOIN` `ON` clause, preserving the fallback row for line-less invoices).
- Migrations `5812510` / `5812520` (`db_alter_view`) for the two views, `5812530` for the `AD_Message`.
- Cucumber: `ediInvoicExport` scenario asserting a packaging-material line is excluded from the exported `Lines`; `C_InvoiceLine` stepdef gains an optional `IsPackagingMaterial` column.

### DIFF
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/edi/postgrestBased/ediInvoicExport.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/edi/postgrestBased/ediInvoicExport.feature
@@ -520,32 +520,52 @@ Feature: EDI INVOIC export via postgREST
 
   @from:cucumber
 @allure.label.epic:E0292_EDI
-@allure.label.feature:F00350_EDI
-@F00350
-  Scenario: packaging-material invoice lines are excluded from the INVOIC export
+@allure.label.feature:F00359_EDI_INVOICE_JSON
+@F00359
+  Scenario: a packaging-material invoice line does not add a second delivery party (DP)
     # A packaging-material line (deposit / Leergut) carries no C_OrderLine_ID.
-    # It must NOT be exported as an INVOIC line item; only the normal product line is exported.
+    # It must not be exported as an INVOIC line item, and — because it would otherwise fall back to
+    # the invoice's own bill-to location — it must not emit a second DP party. The order's delivery
+    # (HandOver) location differs from the bill-to, so the genuine DP is the delivery location and a
+    # spurious second DP (= bill-to, from the packaging line) is detectable.
     Given metasfresh contains C_BPartners without locations:
       | Identifier | IsCustomer | REST.Context.Name | REST.Context.Value | IsVendor | M_PricingSystem_ID |
       | customer1  | Y          | customerName      | customerValue      | N        | pricingSystem      |
+    And metasfresh contains C_Location:
+      | C_Location_ID.Identifier | CountryCode | OPT.Address1 | OPT.Postal | OPT.City   |
+      | loc_bill                 | DE          | billAddr     | 111        | billCity   |
+      | loc_delivery             | DE          | delivAddr    | 222        | delivCity  |
     And metasfresh contains C_BPartner_Locations:
-      | Identifier          | C_BPartner_ID | IsShipToDefault | IsBillToDefault |
-      | bpartner_location_1 | customer1     | Y               | Y               |
+      | Identifier   | GLN           | C_BPartner_ID.Identifier | OPT.C_Location_ID.Identifier | OPT.IsShipTo | OPT.IsBillTo | OPT.BPartnerName |
+      | bpLoc_bill   | 1111111111111 | customer1                | loc_bill                     | true         | true         | billBPName       |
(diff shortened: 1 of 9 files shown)


backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/edi/postgrestBased/ediInvoicExport.feature
```diff
@@ -515,5 +515,74 @@ Feature: EDI INVOIC export via postgREST
       "Version": "0.2"
     }
   ]
+}
+    """
+
+  @from:cucumber
+@allure.label.epic:E0292_EDI
+@allure.label.feature:F00350_EDI
```
metas-ts: the feature should be F00359 - EDI INVOICE JSON

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/edi/postgrestBased/ediInvoicExport.feature
```diff
@@ -515,5 +515,74 @@ Feature: EDI INVOIC export via postgREST
...
+  ]
+}
+    """
+
+    Then the metasfresh REST-API responds with
+    """
+{
+  "metasfresh_INVOIC": [
+    {
+      "Invoice_ID": @pkgInvoice_ID@,
```
metas-ts, line 600: this is not a correct response. where are the parties??

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/edi/postgrestBased/ediInvoicExport.feature
metas-ts: **Resolution (dev):** Fixed — corrected this scenario's feature tag to `@allure.label.feature:F00359_EDI_INVOICE_JSON` / `@F00359` (epic kept `E0292_EDI`). It lands in the converged push that also addresses the parties comment below.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/edi/postgrestBased/ediInvoicExport.feature
metas-ts: **Resolution (dev):** Agreed. As written the scenario can't actually test the party fix: the customer location has no address, so `edi_cctop_119_v`'s `WHERE (Address1 IS NOT NULL OR Address2 IS NOT NULL)` drops every customer party (only `SU` remains) — hence "no parties". And two `DP` rows only arise when the delivery location differs from the bill-to, because the `ship` branch's `SELECT DISTINCT` collapses same-location rows.

Rebuilding it as a full order → shipment → invoice scenario with an address+GLN customer location, a `HandOver` delivery location ≠ the bill-to, and an added packaging-material invoice line — asserting `Partners` contains exactly one `DP` (the delivery one) and the packaging line is absent from `Lines`. Without the fix that yields two `DP` rows; with it, one. Will update this thread with the commit.

metas-ts: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/76e5033ce8f. The scenario is now a full order→invoice case: the order's `HandOver` (delivery) location differs from the bill-to, and an order-less packaging-material invoice line is added. The exported INVOIC is asserted to carry exactly **one** `DP` party (the delivery location) and to exclude the packaging line from `Lines`.

Verified locally against a live stack: **green with the fix**, and proven **RED without it** — the product line resolves its DP to the HandOver/delivery location while the order-less packaging line resolves to the invoice bill-to; without the `119_v` guard the ship branch emits both → a second `DP` → 6 `Partners` → the assertion fails. Stepdef support added: `C_Order.HandOver_Location_ID` and `C_InvoiceLine.C_OrderLine_ID`.

PR #25025, opened 2026-07-08
Carry candidate locators onto DD_OrderLine + gated locator-aware aggregation

teosarca:
## What

Adds a gated aggregation dimension to the `DD_Order_Candidate` → `DD_Order` processing so that a candidate carrying its own source/target locator produces a **single-line `DD_Order` that carries those locators**, instead of collapsing per-locator moves and overwriting the line with warehouse-default locators.

- New sysconfig **`DDOrderAggregation.header.byLocatorId`** (seeded, default **`N`**).
- When **off** (default): behaviour is byte-for-byte unchanged — every existing installation is unaffected.
- When **on**: the candidate's source/target locator joins the header + line aggregation keys (one `DD_Order` per from/to locator move), and `DDOrderCandidateProcessCommand.createLine` carries the candidate's own locator onto the line, falling back per-side to the warehouse default when the candidate has none.

## Why

Without this, a candidate that already knows its exact source/target locator (e.g. an automated ground-replenishment source) had those locators discarded — every generated line got the warehouse default locator, and same-product candidates for different target locators were merged into one line. Downstream dedup/guard logic that keys on the line's target locator could then never detect an already-scheduled move, causing repeated regeneration of equivalent orders.

## Tests

`DDOrderCandidateProcessCommandTest` (module `de.metas.manufacturing`), 6/6 green:
- `locatorsCarriedOntoLine` — knob on: line carries the candidate's source/target locator.
- `oneOrderOneLinePerMove` — knob on: same product, two target locators → two single-line orders.
- `flagOff_unchanged` — knob off (control): default-locator fallback + single aggregated line, asserting the exact prior behaviour (AC of byte-for-byte compatibility).

## Migration

One system sysconfig seed at value `N` (mirrors the existing `DDOrderAggregation.header.byProductId` seed).

### DIFF
--- a/backend/de.metas.manufacturing/src/main/java/de/metas/distribution/ddordercandidate/DDOrderCandidateProcessCommand.java
+++ b/backend/de.metas.manufacturing/src/main/java/de/metas/distribution/ddordercandidate/DDOrderCandidateProcessCommand.java
@@ -343,6 +343,8 @@ class DDOrderCandidateProcessCommand
 		boolean aggregateByPPOrderRef;
 		boolean aggregateBySalesOrderLineId;
 		boolean aggregateByProductId;
+		boolean aggregateByLocatorFrom;
+		boolean aggregateByLocatorTo;
 	}
 	//
 	//
@@ -376,8 +378,12 @@ class DDOrderCandidateProcessCommand
 
 		@Nullable ProductId productId;
 
-		@NonNull LocatorId sourceLocatorId;
-		@NonNull LocatorId targetLocatorId;
+		// Locators are part of the header key only when the corresponding sysconfig is enabled
+		// (DDOrderAggregation.header.byLocatorFrom / .byLocatorTo). When disabled they stay null here,
+		// so candidates that differ only by locator share one DD_Order header. The DD_OrderLine still
+		// carries the resolved locator unconditionally — that comes from LineAggregationKey, not this key.
+		@Nullable LocatorId sourceLocatorId;
+		@Nullable LocatorId targetLocatorId;
 
 		public static HeaderAggregationKey of(
 				@NonNull final DDOrderCandidate candidate,
@@ -396,9 +402,15 @@ class DDOrderCandidateProcessCommand
 					.shipperId(candidate.getShipperId())
 					.isSimulated(candidate.isSimulated())
 					.productPlanningId(candidate.getProductPlanningId())
-					.traceId(candidate.getTraceId())
-					.sourceLocatorId(sourceLocatorId)
-					.targetLocatorId(targetLocatorId);
+					.traceId(candidate.getTraceId());
+			if (aggregationConfig.isAggregateByLocatorFrom())
+			{
+				keyBuilder.sourceLocatorId(sourceLocatorId);
+			}
+			if (aggregationConfig.isAggregateByLocatorTo())
+			{
+				keyBuilder.targetLocatorId(targetLocatorId);
+			}
 			if (aggregationConfig.isAggregateBySalesOrderId())
 			{
 				keyBuilder.salesOrderId(candidate.getSalesOrderId());
--- a/backend/de.metas.manufacturing/src/main/java/de/metas/distribution/ddordercandidate/DDOrderCandidateProcessCommand.java
+++ b/backend/de.metas.manufacturing/src/main/java/de/metas/distribution/ddordercandidate/DDOrderCandidateProcessCommand.java
@@ -327,6 +327,7 @@ class DDOrderCandidateProcessCommand
 		boolean aggregateByPPOrderRef;
 		boolean aggregateBySalesOrderLineId;
 		boolean aggregateByProductId;
+		boolean aggregateByLocatorId;
 	}
 	//
 	//
--- a/backend/de.metas.manufacturing/src/main/java/de/metas/distribution/ddordercandidate/DDOrderCandidateProcessCommand.java
+++ b/backend/de.metas.manufacturing/src/main/java/de/metas/distribution/ddordercandidate/DDOrderCandidateProcessCommand.java
@@ -331,7 +331,6 @@ class DDOrderCandidateProcessCommand
 		boolean aggregateByPPOrderRef;
 		boolean aggregateBySalesOrderLineId;
 		boolean aggregateByProductId;
-		boolean aggregateByLocatorId;
 	}
 	//
 	//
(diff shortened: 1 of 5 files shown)


backend/de.metas.manufacturing/src/main/java/de/metas/distribution/ddordercandidate/DDOrderCandidateProcessCommand.java
```diff
@@ -386,6 +394,11 @@ public static HeaderAggregationKey of(@NonNull final DDOrderCandidate candidate,
 			{
 				keyBuilder.productId(candidate.getProductId());
 			}
+			if (aggregationConfig.isAggregateByLocatorId())
```
teosarca: this is bullshit. if the locator is set, pls always use it. is not a syscnfig option!?!?!?
if candidae does not have it, then use the wh's get create default.
so the key will alwyas have source and target locators

backend/de.metas.manufacturing/src/main/java/de/metas/distribution/ddordercandidate/DDOrderCandidateService.java
```diff
@@ -37,6 +37,7 @@ public class DDOrderCandidateService
 	public static final String SYSCONFIG_DDOrderAggregation_header_byPPOrderRef = "DDOrderAggregation.header.byPPOrderRef";
 	public static final String SYSCONFIG_DDOrderAggregation_line_bySalesOrderLineId = "DDOrderAggregation.line.bySalesOrderLineId";
 	public static final String SYSCONFIG_DDOrderAggregation_header_byProductId = "DDOrderAggregation.header.byProductId";
+	public static final String SYSCONFIG_DDOrderAggregation_header_byLocatorId = "DDOrderAggregation.header.byLocatorId";
```
teosarca: bullshit ssycnfig

backend/de.metas.manufacturing/src/test/java/de/metas/distribution/ddordercandidate/DDOrderCandidateProcessCommandTest.java
```diff
@@ -55,6 +55,11 @@
  * Also covers the optional {@code aggregateByProductId} header-aggregation dimension: with the flag off,
  * candidates of different products share one DD_Order (one line each); with it on, each product gets its
  * own DD_Order.
+ * <p>
+ * Also covers locator handling: the candidate's {@code sourceLocatorId}/{@code targetLocatorId} must be
```
teosarca: pls stop AI slop. no more pages of crap in header. kepe the header short

backend/de.metas.manufacturing/src/main/sql/postgresql/system/30-de.metas.adempiere.libero/5812710_sys_DDOrderAggregation_header_byLocatorId.sql
```diff
@@ -0,0 +1,8 @@
+-- Run mode: SWING_CLIENT
+
+-- IDs allocated from idserver: AD_MigrationScript=5812710, AD_SysConfig_ID=541833
+-- SysConfig Name: DDOrderAggregation.header.byLocatorId
+-- SysConfig Value: N
+-- When 'Y', the source/target locator is part of the DD_Order header+line aggregation key (one DD_Order per from/to locator move). Default 'N'.
+INSERT INTO AD_SysConfig (AD_Client_ID,AD_Org_ID,AD_SysConfig_ID,ConfigurationLevel,Created,CreatedBy,Description,EntityType,IsActive,Name,Updated,UpdatedBy,Value) VALUES (0,0,541833 /*From ID Server*/,'O',TO_TIMESTAMP('2026-07-08 11:30:00','YYYY-MM-DD HH24:MI:SS'),100,'When enabled, prevents DD_Order_Candidates of different source/target locators from ending up in the same DD_Order.','D','Y','DDOrderAggregation.header.byLocatorId',TO_TIMESTAMP('2026-07-08 11:30:00','YYYY-MM-DD HH24:MI:SS'),100,'N')
```
teosarca: xrap

backend/de.metas.manufacturing/src/main/java/de/metas/distribution/ddordercandidate/DDOrderCandidateProcessCommand.java
teosarca: **Resolution (dev):** Agreed — dropping the sysconfig gate. The candidate's source/target locator is now always carried onto the DD_OrderLine (falling back to the warehouse default locator only when the candidate has none), and source+target locators are always part of the header + line aggregation key. Removing the `aggregateByLocatorId` config flag and the `DDOrderAggregation.header.byLocatorId` seed + dt204 enable migrations. Commit to follow.

backend/de.metas.manufacturing/src/main/java/de/metas/distribution/ddordercandidate/DDOrderCandidateService.java
teosarca: **Resolution (dev):** Agreed — removing this sysconfig entirely (constant + read). Locator handling becomes unconditional per the paired thread: candidate locator always used, warehouse default as fallback, always in the aggregation key. The core seed migration and the dt204 enable migration are deleted with it. Commit to follow.

backend/de.metas.manufacturing/src/main/sql/postgresql/system/30-de.metas.adempiere.libero/5812710_sys_DDOrderAggregation_header_byLocatorId.sql
teosarca: **Resolution (dev):** Done — this seed migration and the sysconfig it created are deleted in commit https://github.com/metasfresh/metasfresh/commit/bcd8536eedb; locator-aware aggregation is now unconditional, no sysconfig knob remains.

backend/de.metas.manufacturing/src/test/java/de/metas/distribution/ddordercandidate/DDOrderCandidateProcessCommandTest.java
teosarca: **Resolution (dev):** Agreed — trimming the class Javadoc to a single line (the NPE it guards). Per-scenario detail (product/locator aggregation) already lives in each test method's own Javadoc, so the class header will not restate it. Commit to follow.

teosarca: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/6f402ea521180d785c100bacaeac214ce9f16d03 — class Javadoc trimmed to one line; per-method Javadoc keeps the per-scenario detail.

PR #25044, opened 2026-07-09
OLCand warehouse resolution: honor BP picking warehouse

mrie04:
## What

Sales orders created from order candidates (`C_OLCand`, e.g. via the REST `orders/sales/candidates/bulk` endpoint / EDI import) now honor the **customer BPartner's picking warehouse** when the order candidate carries no explicit warehouse — exactly as manually-entered sales orders already do via `WarehouseAdvisor`.

Previously such orders always fell back to the OLCand-processor default warehouse, so the created order (and its lines) got the wrong `M_Warehouse_ID`.

## How

- **`de.metas.business`** — extract the customer-picking-warehouse gate onto a public, BPartner-keyed method: `IWarehouseAdvisor.evaluateCustomerPickingWarehouse(BPartnerId)` (returns the BP's warehouse iff the BP is a customer and that warehouse is flagged `IsPickingWarehouse`, else `null`). The existing order-based `findPickingWarehouseId(order)` now delegates to it — behaviour unchanged (SOTrx/null guards retained).
- **`de.metas.salescandidate.base`** — `IOLCandBL.getWarehouseId(olCand, orderDefaults)` resolves the warehouse with precedence: OLCand's own `M_Warehouse_ID` → sold-to BPartner picking warehouse (via the WarehouseAdvisor method above) → OLCand-processor default. The sold-to BPartner is the buyer partner (`getBuyerPartnerInfo`), i.e. the same id that becomes `C_Order.C_BPartner_ID` — matching the manual-order path; `BPartnerOrderParams` (keyed on the ship/dropship partner) is deliberately not used.
- The resolved warehouse is stored on the `OLCand` value object (mirroring how `shipperId` is resolved and threaded), so both the order header and the order lines pick it up via `OLCand.getWarehouseId()` — `OLCandOrderFactory` needs no change.
- Order grouping / splitting is unaffected: the split key reads the raw `M_Warehouse_ID` column via `getValueByColumn`, not the resolved getter.

## Tests

- `WarehouseAdvisorTest` (3 cases: customer+picking → WH; customer+non-picking → null; non-customer → null).
- `OLCandBLGetWarehouseIdTest` (4 cases covering the full precedence).
- Real-flow cucumber `olCandBPPickingWarehouse.feature` (`@Id:S30235_01`): POSTs a candidate for a customer BPartner with a picking warehouse and no warehouse in the payload, processes it to an order, and asserts both the order header and line take the BP picking warehouse (developed RED-first, now green).

### DIFF
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/olCandBPPickingWarehouse.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/olCandBPPickingWarehouse.feature
@@ -1,0 +1,123 @@
+@from:cucumber
+@allure.label.epic:E0100_Sales
+@allure.label.feature:F00122
+@topic:orderCandidate
+@ghActions:run_on_executor3
+Feature: OLCand order creation uses BP picking warehouse when no warehouse is in the payload
+## F00122: OLCand Warehouse Advisor
+##
+## When a sales order candidate is POSTed with no warehouse in the payload,
+## the order must be created with the business partner's picking warehouse —
+## NOT the OLCand-processor default warehouse.
+##
+## Bug: before the fix, the processor-default warehouse always wins even when
+## the BP has an explicit picking warehouse configured.
+
+  Background:
+    Given infrastructure and metasfresh are running
+    And the existing user with login 'metasfresh' receives a random a API token for the existing role with name 'WebUI'
+    And metasfresh has date and time 2021-04-16T13:30:13+01:00[Europe/Berlin]
+    And set sys config boolean value true for sys config SKIP_WP_PROCESSOR_FOR_AUTOMATION
+
+  @from:cucumber
+  @allure.label.epic:E0100_Sales
+  @allure.label.feature:F00122
+  @topic:orderCandidate
+  @Id:S30235_01
+  Scenario: OLCand without warehouse → order inherits BP picking warehouse, not processor default
+    # Two warehouses: WH_PICK is the BP's picking warehouse; WH_DEFAULT is set as the
+    # processor default warehouse.  The created order must use WH_PICK (the BP wins).
+    # With the bug the order takes WH_DEFAULT → this scenario is RED before the fix.
+    Given metasfresh contains M_Warehouse:
+      | Identifier | Value      | Name       | OPT.IsPickingWarehouse |
+      | pickingWH  | pickingWH  | pickingWH  | true                   |
+      | defaultWH  | defaultWH  | defaultWH  | false                  |
+
+    And update C_OLCandProcessor:
+      | OPT.C_OLCandProcessor_ID | OPT.M_Warehouse_ID.Identifier |
+      | 1000003                  | defaultWH                     |
+
+    And metasfresh contains M_PricingSystems
+      | Identifier | Name       | Value      | OPT.IsActive |
+      | ps_S30235  | ps_S30235  | ps_S30235  | true         |
+    And metasfresh contains M_PriceLists
+      | Identifier | M_PricingSystem_ID.Identifier | OPT.C_Country.CountryCode | C_Currency.ISO_Code | Name       | SOTrx | IsTaxIncluded | PricePrecision | OPT.IsActive |
+      | pl_S30235  | ps_S30235                     | DE                        | EUR                 | pl_S30235  | true  | false         | 2              | true         |
+    And metasfresh contains M_PriceList_Versions
+      | Identifier | M_PriceList_ID.Identifier | Name       | ValidFrom  |
+      | plv_S30235 | pl_S30235                 | plv_S30235 | 2021-04-01 |
+    And metasfresh contains M_Products:
+      | Identifier     | Name           |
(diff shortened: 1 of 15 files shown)


backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/olCandBPPickingWarehouse.feature
```diff
@@ -0,0 +1,121 @@
...
+  Scenario: OLCand without warehouse → order inherits BP picking warehouse, not processor default
+    # Two warehouses: WH_PICK is the BP's picking warehouse; WH_DEFAULT is set as the
+    # processor default warehouse.  The created order must use WH_PICK (the BP wins).
+    # With the bug the order takes WH_DEFAULT → this scenario is RED before the fix.
+    Given metasfresh contains M_Warehouse:
+      | Identifier | Value      | Name       | OPT.IsPickingWarehouse |
+      | pickingWH  | pickingWH  | pickingWH  | true                   |
+      | defaultWH  | defaultWH  | defaultWH  | false                  |
+
+    And update C_OLCandProcessor:
```
mrie04, line 36: why this is needed? before the change it did default to ad_org_info warehouse and not this one

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/olCandBPPickingWarehouse.feature
```diff
@@ -0,0 +1,121 @@
...
+}
+"""
+    Then process metasfresh response
+      | C_Order_ID.Identifier |
+      | order_S30235          |
+
+    # Before the fix: C_Order.M_Warehouse_ID = defaultWH (processor default wins) → RED.
+    # After the fix: C_Order.M_Warehouse_ID = pickingWH (BP picking warehouse wins) → GREEN.
+    And validate the created orders
+      | C_Order_ID.Identifier | C_BPartner_ID.Identifier | C_BPartner_Location_ID.Identifier | DateOrdered | DocBaseType | currencyCode | DeliveryRule | DeliveryViaRule | poReference | processed | DocStatus | OPT.M_Warehouse_ID.Identifier |
```
mrie04: why opt and identifier

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/olCandBPPickingWarehouse.feature
mrie04: **Resolution (dev):** Verified the pre-change path: `OLCand.getWarehouseId()` returned only the OLCand's own `M_Warehouse_ID`, and `OLCandOrderFactory` set the order warehouse via `coalesce(olCandWarehouse, orderDefaults.warehouseId)` — i.e. OLCand-own → OLCand-**processor** default. The `AD_OrgInfo` standard-warehouse fallback (`WarehouseAdvisor.findOrderWarehouseId`) is the *order-based* advisor used downstream (picking/shipment), not in the OLCand→order creation path. In this case the warehouse is configured on the processor, so the processor default is the real pre-change source this step reproduces. Reworked: the step now addresses the real standard seed processor (repo-id 1000003, int-driven) and sets its default to StdWarehouse (540008); the fabricated `defaultWH` is gone. Commit + CI to follow.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/olCandBPPickingWarehouse.feature
mrie04: **Resolution (dev):** Agreed. `OPT.` removed — the warehouse is the assertion under test, so it must be a hard (always-present) assertion. The `.Identifier` suffix is also dropped from the header, since `row.getAsOptionalIdentifier("M_Warehouse_ID")` already resolves a plain `M_Warehouse_ID` header (see `DataTableRow.getAsOptionalIdentifierString`). Commit + CI to follow.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/olCandBPPickingWarehouse.feature
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/5bbeb0b1382 (processor default set + BP picking warehouse asserted; step def restores the processor in an @After hook).

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/automateOrderCand/olCandBPPickingWarehouse.feature
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/5bbeb0b1382 (OPT. + .Identifier dropped from the warehouse assertion headers).

PR #25051, opened 2026-07-09
MD_Stock reconciliation: converge despite duplicate rows (dedup + unique index + view active-filter)

teosarca:
## Problem

`MD_Stock` has no uniqueness constraint on its business key
`(AD_Client_ID, AD_Org_ID, M_Product_ID, M_Warehouse_ID, AttributesKey)` — only the surrogate
`MD_Stock_ID` is unique. A create-create race could therefore insert two active rows for one
bucket. The reconciliation process `MD_Stock_Update_From_M_HUs` resolves each bucket via
`firstOnly()`, which throws `QueryMoreThanOneRecordsFound` on `>1` active row — so a single
duplicate bucket aborts the whole run and stock divergence never drains.

## Fix

1. **Migration — de-dup + prevention.** Deactivate all but `min(MD_Stock_ID)` per active
   business-key bucket, then add a partial unique index
   `MD_Stock_BusinessKey_uq ... WHERE IsActive='Y'` on the 5-column business key. Order matters
   (the index build fails if duplicates remain).
2. **Graceful concurrent-create.** `StockDataUpdateRequestHandler.retrieveOrCreateDataRecord`
   now wraps the create path in `try save; catch DBUniqueConstraintException -> re-query
   firstOnly`, mirroring the established `BPGroupService.getOrCreateBPGroup` idiom — a create race
   resolves to the winning row instead of a hard error.
3. **Reconciliation view considers only active rows.** `MD_Stock_From_HUs_V` previously read
   `FROM MD_Stock` with no `IsActive` filter, so a deactivated (de-duplicated) row was still
   counted as current stock and emitted a phantom `QtyOnHandChange` that the active-only reset can
   never clear — driving the drain loop to `MAX_LOOPS`. The current-stock side is now
   `FROM (SELECT * FROM MD_Stock WHERE IsActive='Y') s`, kept inside the `FROM` so the
   `FULL OUTER JOIN` still yields HU-only buckets that have no `MD_Stock` row yet.
4. **Consistency.** `StockRepository.createStockDataItemQuery` (shipment-availability stock) now
   applies `addOnlyActiveRecordsFilter()`, matching its sibling, so deactivated rows are never
   summed into current stock.

## Tests

- **Cucumber** `mdStock_Reconciliation.feature`: reconciliation converges a divergent `MD_Stock`
  row to HU-derived truth, and is idempotent across repeated runs (both verified running the real
  `MD_Stock_Update_From_M_HUs` process against a local stack).
- **DB layer**: dedup collapses duplicates to one active row and the partial unique index rejects a
  duplicate active-row insert.
- **Unit**: `StockDataUpdateRequestHandlerTest` covers the create branch.


backend/de.metas.material/cockpit/src/main/sql/postgresql/ddl/views/MD_Stock_From_HUs_V.sql
```diff
@@ -12,7 +12,7 @@ SELECT COALESCE(wh.AD_Client_ID, hu_agg.AD_Client_ID) AS AD_Client_ID,
 
        -- QtyOnHandChange is the quantity - in stocking-UOM - to add to the *current* MD_Stock.QtyOnHand to get the *correct* qtyOnHand
        COALESCE(hu_agg.QtyOnHand, 0) - COALESCE(s.QtyOnHand, 0) AS QtyOnHandChange
-FROM MD_Stock s
+FROM (SELECT * FROM MD_Stock WHERE IsActive = 'Y') s /* active only: a deactivated (e.g. de-duplicated) MD_Stock row must not be counted as current stock, else it becomes a phantom divergence the active-only reset can never clear -> the reconciliation loops forever */
```
teosarca, line 15: why sub-query and not filtering in WHERE clause below?
is it ok from perf pov?

backend/de.metas.material/cockpit/src/main/sql/postgresql/system/75-material-cockpit/5812890_sys_MD_Stock_dedup.sql
```diff
@@ -0,0 +1,28 @@
...
+        WHERE k.IsActive = 'Y'
+          AND k.AD_Client_ID = t.AD_Client_ID
+          AND k.AD_Org_ID = t.AD_Org_ID
+          AND k.M_Product_ID = t.M_Product_ID
+          AND k.M_Warehouse_ID = t.M_Warehouse_ID
+          AND k.AttributesKey = t.AttributesKey
+      );
+
+-- Prevent recurrence: at most one active row per business key.
+CREATE UNIQUE INDEX MD_Stock_BusinessKey_uq ON MD_Stock (AD_Client_ID, AD_Org_ID, M_Product_ID, M_Warehouse_ID, AttributesKey) WHERE IsActive='Y';
```
teosarca: do not fix data fixing migratoin scripts with pure DDL migrtion scripts!
capture this learning!

backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/stock/StockDataUpdateRequestHandler.java
```diff
@@ -117,7 +118,21 @@ private I_MD_Stock retrieveOrCreateDataRecord(@NonNull final StockDataRecordIden
...
-		return newDataRecord;
+		try
+		{
+			save(newDataRecord);
+			return newDataRecord;
+		}
+		catch (final DBUniqueConstraintException e)
+		{
+			// meanwhile another thread created the same bucket; return the existing one
+			final I_MD_Stock winner = query.firstOnly(I_MD_Stock.class);
```
teosarca, line 131: didn't we agree to not write this fallback code because we introduce a unique index?

teosarca: **Resolution (dev):** Agreed — removed. The partial unique index makes a persisted duplicate impossible, and a genuine concurrent-create race now surfaces a transient unique-violation to the async event handler, which retries and then finds the winning row — so no in-code fallback is needed. Reverted in https://github.com/metasfresh/metasfresh/commit/8da6d1dd8368d71b448dc9f3adfe6a540e9f1e8a (handler back to plain `firstOnly` / build-and-return; the create-branch unit test removed).

backend/de.metas.material/cockpit/src/main/sql/postgresql/system/75-material-cockpit/5812890_sys_MD_Stock_dedup.sql
teosarca: **Resolution (dev):** Split. The dedup data-fix (`UPDATE … IsActive='N'`) now lives in `5812890_sys_MD_Stock_dedup.sql`, and the `CREATE UNIQUE INDEX` in its own pure-DDL script `5812990_sys_MD_Stock_business_key_unique_index.sql` (higher prefix, so the dedup clears duplicates before the index builds). Commits https://github.com/metasfresh/metasfresh/commit/8da6d1dd8368d71b448dc9f3adfe6a540e9f1e8a + https://github.com/metasfresh/metasfresh/commit/f93c9819c76af3de55e664505761fa9ffc0cb6c2. Learning captured — a data-fixing migration must be a separate script from a pure-DDL one.

backend/de.metas.material/cockpit/src/main/sql/postgresql/ddl/views/MD_Stock_From_HUs_V.sql
teosarca: **Resolution (dev):** The sub-query is required for correctness, not preference. The view is a `FULL OUTER JOIN` between `MD_Stock` and the HU aggregate; a trailing `WHERE s.IsActive='Y'` would also drop the HU-only rows (where `s` is NULL → `NULL='Y'` is false) — exactly the buckets that have HU stock but no `MD_Stock` row yet and that the reset must CREATE. Filtering inside the `FROM` keeps those rows. Perf is equivalent: Postgres pushes the `IsActive='Y'` predicate into the scan of `md_stock` and the derived table is not materialised (same plan as an inline filter). Kept as the sub-query.

backend/de.metas.material/cockpit/src/main/java/de/metas/material/cockpit/stock/StockDataUpdateRequestHandler.java
teosarca: **Correction to the resolution above:** the "async event handler retries" premise was wrong — traced the actual dispatch path (`EventLogUserService.invokeHandlerAndLog` catches and logs to `AD_Issue` without rethrowing; `MaterialEventHandlerRegistry.onEvent` then marks the event processed regardless) and there is no retry, requeue, or DLQ anywhere in that path. Without the in-code fallback, a genuine create-create race (the event path vs. `MD_Stock_Update_From_M_HUs`, which this class's own Javadoc says may run in parallel) throws an uncaught `DBUniqueConstraintException` and silently drops that update — the exact failure mode this PR exists to close, reopened via a different path (lost update instead of duplicate row).

Restored the `save()` + unique-violation-catch fallback and its regression test in https://github.com/metasfresh/metasfresh/commit/5aba008259c8a12a586a16f0e2d3941485971a7b (revert of https://github.com/metasfresh/metasfresh/commit/8da6d1dd8368d71b448dc9f3adfe6a540e9f1e8a). The partial unique index stays as the primary guarantee against a *persisted* duplicate; the fallback only handles the transient in-flight race before either write commits.

teosarca: **Resolution (dev):** Net current state — the fallback is **kept** (commit https://github.com/metasfresh/metasfresh/commit/5aba008259c8a12a586a16f0e2d3941485971a7b). My earlier "removed" resolution above rested on a wrong, unverified premise ("the async handler retries"). Verified in source: `EventLogUserService.invokeHandlerAndLog` (`backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/event/log/EventLogUserService.java:154`) catches the handler's `RuntimeException`, logs it to `AD_Issue`, and does **not** rethrow — no retry/requeue. So without the fallback, a create-create race between the event path and `MD_Stock_Update_From_M_HUs` throws an uncaught `DBUniqueConstraintException` that is swallowed → the stock update is silently lost. The `save` + unique-violation-catch fallback prevents that; the partial unique index remains the guarantee against a *persisted* duplicate. Because this reverses the earlier directive to drop the fallback, it is being confirmed with the author before integration.

backend/de.metas.material/cockpit/src/main/sql/postgresql/system/75-material-cockpit/5812990_sys_MD_Stock_business_key_unique_index.sql
```diff
@@ -0,0 +1,5 @@
+-- DDL: prevent more than one ACTIVE MD_Stock row per business key.
+-- Runs after the data-fix script (5812890) that de-duplicates existing rows — the index build
+-- would fail while duplicates remain. Kept as a pure-DDL script, separate from the data fix.
+
+CREATE UNIQUE INDEX MD_Stock_BusinessKey_uq ON MD_Stock (AD_Client_ID, AD_Org_ID, M_Product_ID, M_Warehouse_ID, AttributesKey) WHERE IsActive='Y';
```
metas-ts:
```suggestion
CREATE UNIQUE INDEX MD_Stock_BusinessKey_uq ON MD_Stock (M_Product_ID, AttributesKey, M_Warehouse_ID, AD_Client_ID, AD_Org_ID) WHERE IsActive='Y';
```

`AD_Client_ID, AD_Org_ID` re not very usefull when scanning the index..

teosarca: **Resolution (dev):** Adopted — reordered to `(M_Product_ID, AttributesKey, M_Warehouse_ID, AD_Client_ID, AD_Org_ID)` so the selective columns lead and the low-cardinality tenant columns come last. Uniqueness enforced is unchanged (column order doesn't affect what a unique index guarantees, only scan efficiency). https://github.com/metasfresh/metasfresh/commit/61fa80bde449ac86b5b6bb8b66aad77d247cd279

teosarca: capture learning + review rule!
i.e. columns with the highest selectivity (highest cardinality) should generally come first.

teosarca: **Resolution (dev):** rule added — captured as a review rule + migration-lint candidate: composite-index columns are ordered most-selective first, with low-cardinality tenant columns (`AD_Client_ID`/`AD_Org_ID`) last. Reviewers will flag a composite index that leads with the tenant columns. This index already follows it (https://github.com/metasfresh/metasfresh/commit/61fa80bde449ac86b5b6bb8b66aad77d247cd279).

PR #25057, opened 2026-07-09
Mobile Packing unpick: reduce M_ShipmentSchedule_QtyPicked on skip-to-floor

teosarca:
## Problem
In Mobile Packing, unpicking qty **to the floor** from a CU that was picked into a bare TU detaches the CU physically (HU correct) but did **not** reduce `M_ShipmentSchedule_QtyPicked` — the schedule kept the full picked qty, over-reporting what was actually picked.

## Root cause
The pick records the pick-to-**TU** on `M_ShipmentSchedule_QtyPicked` (`M_TU_HU_ID` set, `VHU_ID` null). On skip-to-floor unpick, the reversal looked up the schedule row by the extracted floor CU's HU id, which never matches the TU-keyed row, so the row was never reduced.

## Fix
- On skip-to-floor unpick, reduce the TU-keyed `M_ShipmentSchedule_QtyPicked` row(s) by the unpicked qty (newest-first, deleting a row once its qty reaches zero).
- Unified the shared-HU consignee reset behind one guard, `resetConsigneeIfNoActivePickedRows`: only reset a HU's BPartner/Location when no shipment schedule still holds an active picked row on it. Applied consistently across all reset call sites, replacing divergent per-site logic.
- De-twisted the unpick command dispatch for clarity.
- Catch-weight proportional-reduce behavior on unpick is preserved.

## Test coverage
- Mobile Playwright E2E: skip-to-floor unpick from a bare TU, covering consignee retained when another line still shares the TU vs. cleared when it was the last active line; multi-pick spanning; two independent lines; catch-weight scaling on unpick.
- Unit test covering the previously-dead reset-guard code path.

Verified: CI green at HEAD `cca743c879a` — https://github.com/metasfresh/metasfresh/actions/runs/29053345840 (all jobs, incl. the mobile E2E suite covering this fix, passed).

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/IHUShipmentScheduleDAO.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/IHUShipmentScheduleDAO.java
@@ -41,6 +41,9 @@ public interface IHUShipmentScheduleDAO extends ISingletonService
 
 	List<I_M_ShipmentSchedule_QtyPicked> retrieveSchedsQtyPickedForTU(int shipmentScheduleId, int tuHUId, String trxName);
 
+	/** @return true if any active M_ShipmentSchedule_QtyPicked row is keyed to the given TU for a schedule OTHER than the excluded one (a bare TU can be shared across schedules). */
+	boolean hasActiveQtyPickedForTUExcludingSchedule(int tuHUId, int excludeShipmentScheduleId);
+
 	List<I_M_ShipmentSchedule_QtyPicked> retrieveSchedsQtyPickedForVHU(I_M_HU vhu);
 
 	IQueryBuilder<I_M_ShipmentSchedule_QtyPicked> retrieveSchedsQtyPickedForVHUQuery(I_M_HU vhu);
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/IHUShipmentScheduleDAO.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/IHUShipmentScheduleDAO.java
@@ -41,8 +41,8 @@ public interface IHUShipmentScheduleDAO extends ISingletonService
 
 	List<I_M_ShipmentSchedule_QtyPicked> retrieveSchedsQtyPickedForTU(int shipmentScheduleId, int tuHUId, String trxName);
 
-	/** @return true if any active M_ShipmentSchedule_QtyPicked row is keyed to the given TU for a schedule OTHER than the excluded one (a bare TU can be shared across schedules). */
-	boolean hasActiveQtyPickedForTUExcludingSchedule(int tuHUId, int excludeShipmentScheduleId);
+	/** @return true if any active M_ShipmentSchedule_QtyPicked row is still keyed to the given top-level HU (LU/TU/VHU) — a shared HU can carry another schedule's active row. */
+	boolean hasActiveQtyPickedForTopLevelHU(@NonNull I_M_HU topLevelHU);
 
 	List<I_M_ShipmentSchedule_QtyPicked> retrieveSchedsQtyPickedForVHU(I_M_HU vhu);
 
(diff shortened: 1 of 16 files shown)


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/IHUShipmentScheduleDAO.java
```diff
@@ -41,6 +41,9 @@ List<I_M_ShipmentSchedule_QtyPicked> retrieveByTopLevelHUAndShipmentScheduleId(
 
 	List<I_M_ShipmentSchedule_QtyPicked> retrieveSchedsQtyPickedForTU(int shipmentScheduleId, int tuHUId, String trxName);
 
+	/** @return true if any active M_ShipmentSchedule_QtyPicked row is keyed to the given TU for a schedule OTHER than the excluded one (a bare TU can be shared across schedules). */
+	boolean hasActiveQtyPickedForTUExcludingSchedule(int tuHUId, int excludeShipmentScheduleId);
```
teosarca: avoid primitives
use HuId, ShipmentScheduleId

teosarca: **Resolution (dev):** Agreed — new API must not take raw `int`s. This method is being replaced in the in-progress rework: the guard is re-keyed on the top-level HU (`hasActiveQtyPickedForTopLevelHU(I_M_HU topLevelHU)`, reusing the existing `queryByTopLevelHU`), which removes the `int tuHUId, int excludeShipmentScheduleId` primitives entirely and unifies the consignee-reset guard across all call sites. Anywhere a repo-id parameter genuinely remains I will use `HuId` / `ShipmentScheduleId` value objects, never primitives. Will link the commit here once the reworked HEAD is pushed.

teosarca: **Resolution (dev):** Done — https://github.com/metasfresh/metasfresh/commit/cd72aea5a5c8ecfce47d4bd0a7a5ddd16a6e56f4 removed `hasActiveQtyPickedForTUExcludingSchedule(int, int)` entirely and replaced it with `hasActiveQtyPickedForTopLevelHU(@NonNull I_M_HU topLevelHU)` (guard now keyed on the top-level HU, no primitives). Please mark resolved if you agree.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -818,13 +819,131 @@ public void deleteByTopLevelHUsAndShipmentScheduleId(@NonNull final Collection<I
...
+		final BigDecimal oldQtyDeliveredCatch = qtyPickedRecord.getQtyDeliveredCatch();
+		if (oldQtyDeliveredCatch == null || oldQtyPicked.signum() == 0)
+		{
+			return;
+		}
+
+		final BigDecimal newQtyDeliveredCatch = oldQtyDeliveredCatch
+				.multiply(newQtyPicked)
+				.divide(oldQtyPicked, Math.max(oldQtyDeliveredCatch.scale(), 2), RoundingMode.HALF_UP);
+		qtyPickedRecord.setQtyDeliveredCatch(newQtyDeliveredCatch);
```
teosarca, line 944: shall we also make sure the weight attributes (WeightGross, WeightNet, etc) are also fine/shall be corrected?

if not handled, we will need a follow-up for this case making sure that
* carved out HU has the weight gross as oldQtyDeliveredCatch (and ofc weight net adjusted accordingly)
* remaining HU weight gross is lowered with oldQtyDeliveredCatch

teosarca: **Resolution (dev):** Good catch — confirmed this PR adjusts only the shipment-schedule row's catch qty (`QtyDeliveredCatch`, via `reduceCatchWeightProportionally`); it does NOT touch the physical HU `WeightGross`/`WeightNet` attributes. Those are set on the HU-transform (boundary-carve) layer — a distinct mechanism from this schedule-side reduce — so the carved-out HU carrying weight-gross = its catch share, and the remaining HU's weight-gross being lowered accordingly, is not covered here.

Agreed it should be handled. Since it's a separate HU-attribute concern and this PR is green + scoped to the `M_ShipmentSchedule_QtyPicked` reduction, my recommendation is a **follow-up** (as you suggested) — but I'm happy to fold it into this PR instead if you'd rather keep it together. Your call: follow-up, or extend this PR?

teosarca: **Resolution (dev):** Confirmed — handling as a separate follow-up PR (not folding into this one). Filed an internal tracking issue to cover the HU WeightGross/WeightNet proportional adjustment on the boundary carve.

PR #25071, opened 2026-07-10
MD_Candidate: only fire supply-required-decreased event for DEMAND candidates

teosarca:
## What

`DemandCandiateHandler.onCandidateNewOrChange` fired a *supply-required-decreased* event for **any** negative-delta candidate. The non-DEMAND handled types (`ATTRIBUTES_CHANGED_FROM`, `UNEXPECTED_DECREASE`, `INVENTORY_DOWN`) always carry a negative delta, so they were passed into `SupplyRequiredEventCreator.verifyCandidateType`, which only accepts `DEMAND`/`STOCK_UP` → `IllegalArgumentException`.

**Verified (reproduced locally):** `DemandCandiateHandlerTest.onCandidateNewOrChange_attributesChangedFrom_negativeQty_doesNotFireDecreasedEvent` throws the exact exception on the pre-fix code (RED); green after the fix. Full class 18/18.

**Verified (introduced by):** `git log -S fireSupplyRequiredDecreasedEventIfNeeded` points at https://github.com/metasfresh/metasfresh/pull/20812, which added the decrease branch without the `type == DEMAND` guard that the sibling increase branch has.

**Verified (occurs in production):** this exception signature is recorded in `AD_Issue` on a production instance (confirmed by a read-only query) — a real recurring error, not only a theoretical code path.

## Fix

Gate the decrease branch on `type == DEMAND` (via `CandidateType.isDemand()`), mirroring the existing increase branch. For `DEMAND` behaviour is unchanged; the other handled types now skip the never-valid decreased-event and complete processing normally.

## Impact — hypothesis (unverified)

The swallowed exception aborts `AttributesChangedEventHandler.handleEvent` after the FROM (decrease) candidate is persisted and before the TO (increase) candidate is created, which would leave a half-applied attributes change and drift projected ATP. **Hypothesis (unverified)** — needs a live FROM-vs-TO candidate count to confirm; it is not required for the correctness of this fix (the fix is proven by the reproduced test above).

## Tests

- `onCandidateNewOrChange_attributesChangedFrom_negativeQty_doesNotFireDecreasedEvent` — reproduces the exception (RED before fix).
- `onCandidateNewOrChange_demand_quantityDecrease_firesDecreasedEvent` — a DEMAND candidate whose quantity decreases still fires the decreased event (guards against over-restriction).


teosarca:
## AD_Issue signatures this fixes

This addresses a chronic, **swallowed** exception observed at high volume in `AD_Issue` on the affected production instance — pre-existing (since https://github.com/metasfresh/metasfresh/pull/20812, 2025-05-16), not a recent regression. It is caught + logged, so it never crashed the server; it accrued for months.

**Verified (frequency):** counts below are from a read-only `AD_Issue` query on the instance.

**Primary signature** (~131,722 occurrences, first seen 2026-02-26, nightly ~03:00 bursts):
`IllegalArgumentException: Given parameter demandCandidate needs to have DEMAND or STOCK_UP as type; …type=ATTRIBUTES_CHANGED_FROM…`

Correlated 1:1 (co-triggered by the same nightly HU-attribute driver — **not proven** the same root cause): `NullPointerException: storageAttributesKey is marked non-null but is null` (~135,936, first seen 2025-08-21).

**Root cause (verified — reproduced):** the decrease branch at `DemandCandiateHandler.java:149` fired a supply-required-decreased event for *any* negative-delta candidate with no type guard, while the sibling increase branch at `:144` is guarded `type == DEMAND`. An `ATTRIBUTES_CHANGED_FROM` candidate (always a decrease) flowed into `SupplyRequiredEventCreator.verifyCandidateType`, which accepts only `DEMAND`/`STOCK_UP` → throws.
**Verified:** the RED unit test added in this PR (`DemandCandiateHandlerTest.onCandidateNewOrChange_attributesChangedFrom_negativeQty_doesNotFireDecreasedEvent`) throws this exact `IllegalArgumentException` on the pre-fix code, and passes after the guard (full class 18/18).

**Downstream impact — Hypothesis (unverified):** the throw is swallowed but aborts `AttributesChangedEventHandler.handleEvent` *after* the FROM (decrease) candidate is persisted and *before* the TO (increase) candidate is created, which would leave a half-applied attributes change / projected-ATP drift. Not confirmed — needs a live FROM-vs-TO candidate count; not required for the correctness of this fix.

**Representative stacktrace** (record IDs scrubbed; `storageAttributesKey=-1002` is the `AttributesKey.NONE` sentinel, not data):
```
IllegalArgumentException: Given parameter demandCandidate needs to have DEMAND or STOCK_UP as type;
demandCandidate=Candidate(
  clientAndOrgId=ClientAndOrgId(clientId=ClientId(repoId=<id>), orgId=OrgId(repoId=<id>)),
  type=ATTRIBUTES_CHANGED_FROM, businessCase=null, id=<id>, parentId=NULL, groupId=<id>, seqNo=<id>,
  materialDescriptor=MaterialDescriptor(super=ProductDescriptor(productId=<id>,
    storageAttributesKey=-1002,        -- AttributesKey.NONE sentinel (not data)
    attributeSetInstanceId=<id>),
    warehouseId=WarehouseId(repoId=<id>), locatorId=null, customerId=null, reservedForCustomer=false,
    vendorId=null, quantity=-1, date=<timestamp>),
  minMaxDescriptor=MinMaxDescriptor(min=0, max=0, highPriority=false),
  businessCaseDetail=null, additionalDemandDetail=null, transactionDetails=[], dimension=null, simulated=false)
    at com.google.common.base.Preconditions.checkArgument(Preconditions.java:219)
    at de.metas.material.dispo.service.candidatechange.handler.SupplyRequiredEventCreator.verifyCandidateType(SupplyRequiredEventCreator.java:81)
    at de.metas.material.dispo.service.candidatechange.handler.SupplyRequiredEventCreator.createSupplyRequiredDecreasedEvent(SupplyRequiredEventCreator.java:68)
    at de.metas.material.dispo.service.candidatechange.handler.DemandCandiateHandler.fireSupplyRequiredDecreasedEventIfNeeded(DemandCandiateHandler.java:159)
    at de.metas.material.dispo.service.candidatechange.handler.DemandCandiateHandler.onCandidateNewOrChange(DemandCandiateHandler.java:151)
    at de.metas.material.dispo.service.candidatechange.CandidateChangeService.onCandidateNewOrChange(CandidateChangeService…)
```

**On the `storageAttributesKey` NPE:** likely a co-triggered sibling of the same nightly driver; **not proven** to share this root cause. It needs its own stacktrace before we conclude this fix also resolves it — so this PR is scoped to the `IllegalArgumentException`.

PR #25072, opened 2026-07-10
AFS perf: reuse m_product_id_catchuom_id partial index in retrieve_available_for_sales branch-2 (add IsActive='Y')

teosarca:
## What

Adds `AND s.IsActive = 'Y'` to branch (2) — demand from open (`Processed='N'`) shipment schedules — of `de_metas_material.retrieve_available_for_sales` (Available-for-Sales).

## Why

Branch (2) previously filtered only `s.Processed='N'`, so the planner could **not** use the existing partial index `m_shipmentschedule_m_product_id_catchuom_id` (on `(m_product_id, catch_uom_id)` `WHERE isactive='Y' AND processed='N'`) and fell back to a wide index scan.

Adding the `s.IsActive='Y'` predicate lets the planner reuse that partial index. Measured on a large production instance: **~50x fewer buffers / ~70x faster** on a high-runner product. The change is **output-neutral** — no inactive open shipment schedules exist.

Only branch (2) is changed; branches (1) and (3) are untouched.

## Changes

- DDL source `Retrieve_available_for_Sales.sql`: add the `s.IsActive='Y'` predicate to branch (2) and note the index reuse in the inline comment.
- New system migration re-defining the function (drop + create; body byte-identical to the DDL source (raw diff of both $BODY$ blocks is empty; MD5-equal)).

Follows the prior perf rewrite of this function: https://github.com/metasfresh/metasfresh/pull/24361

### DIFF
--- a/backend/de.metas.material/cockpit/src/main/sql/postgresql/system/75-material-cockpit/5813260_sys_retrieve_available_for_sales_branch2_isactive.sql
+++ b/backend/de.metas.material/cockpit/src/main/sql/postgresql/system/75-material-cockpit/5813260_sys_retrieve_available_for_sales_branch2_isactive.sql
@@ -36,6 +36,20 @@ CREATE FUNCTION de_metas_material.retrieve_available_for_sales(
             )
 AS
 $BODY$
+-- Performance rewrite: the previous implementation read from
+-- de_metas_material.MD_ShipmentQty_V and applied the look-behind / look-ahead date filters
+-- on top of the view. Because the view FULL-OUTER-JOINs the order-line side and the
+-- shipment-schedule side, and the filter contained "... OR SalesOrderLastUpdated IS NULL",
+-- Postgres could not push the date filters down: for a high-runner product it built the
+-- product's ENTIRE C_OrderLine history (150k+ rows) and only then discarded almost all of
+-- it -> several seconds per order-line entry (the Available-for-Sales calc, gh#5108).
+--
+-- The view's two demand sides are disjoint (the order-line side excludes lines that already
+-- have a shipment schedule; the schedule side is the schedules), so the join is effectively a
+-- concatenation. We inline both sides as separate branches and push the date filters INTO each
+-- source, so the look-behind prunes order lines (using the existing
+-- C_OrderLine(Updated, M_Product_ID) index) before any join. Result on a 2.2M-shipment-schedule
+-- / 150k-order-line instance: ~7.6 s -> sub-second. Output semantics are unchanged.
 SELECT p_QueryNo,
        p_M_Product_ID,
        final.AttributesKey,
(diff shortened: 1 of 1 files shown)


teosarca:
## Foundation for this change — live `EXPLAIN (ANALYZE, BUFFERS)`

Measured read-only on a **large production instance** (M_ShipmentSchedule ≈ 10⁶ rows), for a **high-runner product** (id masked below). This is why the fix is *adding one predicate* rather than adding a new index.

Branch (2) of `retrieve_available_for_sales` selects demand from open (`Processed='N'`) shipment schedules. Today it filters `Processed='N'` **without** `IsActive='Y'`, so the planner cannot use the existing partial index `m_shipmentschedule_m_product_id_catchuom_id (m_product_id, catch_uom_id) WHERE isactive='Y' AND processed='N'` and falls back to a wide scan of `m_shipmentschedule_m_attributesetinstance_id` (where `m_product_id` is only the 2nd column):

**Before — `Processed='N'` only (current):**
```
Aggregate  (cost=16329.37..16329.38 rows=1) (actual time=41.796..41.796 rows=1 loops=1)
  Buffers: shared hit=7467
  ->  Index Scan using m_shipmentschedule_m_attributesetinstance_id on m_shipmentschedule s
        Index Cond: (m_product_id = <product>)
        Filter: (processed='N' AND ad_org_id=<org> AND (preparationdate_override <= now()+'72h'
                 OR (preparationdate_override IS NULL AND preparationdate <= now()+'72h')))
        Rows Removed by Filter: 362
        Buffers: shared hit=7467
Execution Time: 41.818 ms
```

**After — add `AND s.IsActive='Y'` (this PR):**
```
Aggregate  (cost=2.65..2.66 rows=1) (actual time=0.578..0.578 rows=1 loops=1)
  Buffers: shared hit=146 read=2
  ->  Index Scan using m_shipmentschedule_m_product_id_catchuom_id on m_shipmentschedule s
        Index Cond: (m_product_id = <product>)
        Filter: (ad_org_id=<org> AND (preparationdate_override <= now()+'72h'
                 OR (preparationdate_override IS NULL AND preparationdate <= now()+'72h')))
        Rows Removed by Filter: 135
        Buffers: shared hit=146 read=2
Execution Time: 0.590 ms
```

**≈ 7,467 → 148 buffers (~50×), ≈ 41.8 → 0.6 ms (~70×)** on this high-runner — by *reusing an index that already exists*, no new index added to this hot table.

**Output-neutral (verified):** on the instance there are **0** open (`Processed='N'`) schedules with `IsActive='N'` (0 carrying `QtyReserved>0`), so adding `IsActive='Y'` drops no real demand. It also aligns branch-2 with the on-hand branch, which already filters `IsActive='Y'`. Regression guard `availableForSales.feature` stays green (9/9).

Builds on the branch-1 rewrite in https://github.com/metasfresh/metasfresh/pull/24361.

PR #25073, opened 2026-07-10
ShipmentSchedule: fix intermittent wrong-close of a fully-shipped schedule (multi-InOut split)

teosarca:
## What

Fixes an intermittent, **permanent** wrong-close of a **fully-shipped** shipment schedule when a sales order's picked lines ship via **≥2 separate `M_InOut` documents** (the async shipment-generation pipeline groups schedules into separate workpackages). A real order-fulfilment correctness bug — a fully-delivered order line left `IsClosed=Y` — surfaced via a flaky mobile-webui E2E.

## Root cause

`ShipmentScheduleBL.closePartiallyShipped_ShipmentSchedules` runs synchronously at each shipment's `M_InOut` `TIMING_AFTER_COMPLETE` and closes every *not-fully-delivered* schedule of the whole order. It judged "delivered" from the schedule's persisted `QtyDelivered`, which is written by a **deferred, per-`M_InOut` `runAfterCommit` recompute**. So when a sibling `M_InOut` of the same order completes first, a schedule already fully shipped by an **earlier sibling** can still read `QtyDelivered=0` → it looks not-delivered → `closeShipmentSchedule`. **Nothing reopens it**, so it stays `IsClosed=Y`.

## Fix

Compute delivered qty from committed data only: **this shipment's own `MovementQty`** + the **committed processed-shipment-line ledger** (`IShipmentScheduleAllocDAO.retrieveQtyDelivered` — summed `MovementQty` of the schedule's `Processed` shipment lines) of already-completed siblings. The two summands are **disjoint**: at `TIMING_AFTER_COMPLETE` the current InOut's own lines are not yet `Processed` (`MInOut.completeIt` fires the validate before `setProcessed`), so the Processed-only ledger doesn't overlap them. The schedule's own `QtyDelivered` column is no longer read — it is only a deferred recompute of that same ledger and lags here.

## Validation

- Unit: a regression test reproducing the async-stale sibling race, plus the sibling-already-recomputed / solely-current-InOut / completely-unshipped shapes — 4/4; full `de.metas.swat.base` suite 482 passed / 0 failed.
- End-to-end: the `close_partially_shipped` mobile picking flow, looped 70× against a locally-rebuilt app server carrying this fix — 70/70 pass, **0** fully-delivered schedules left closed, versus a ~5% permanent-stuck rate on the unfixed backend.

### DIFF
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentScheduleBL.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentScheduleBL.java
@@ -20,9 +20,9 @@ import de.metas.inout.IInOutBL;
 import de.metas.inout.IInOutDAO;
 import de.metas.inout.ShipmentScheduleId;
 import de.metas.inoutcandidate.api.ApplyShipmentScheduleChangesRequest;
+import de.metas.inoutcandidate.api.IShipmentScheduleAllocDAO;
 import de.metas.inoutcandidate.api.IShipmentScheduleBL;
 import de.metas.inoutcandidate.api.IShipmentScheduleEffectiveBL;
-import de.metas.inoutcandidate.api.IShipmentScheduleAllocDAO;
 import de.metas.inoutcandidate.api.IShipmentSchedulePA;
 import de.metas.inoutcandidate.api.OlAndSched;
 import de.metas.inoutcandidate.api.ShipmentScheduleAllowConsolidatePredicateComposite;
@@ -739,18 +739,8 @@ public class ShipmentScheduleBL implements IShipmentScheduleBL
 
 				final BigDecimal effectiveQtyOrdered = shipmentScheduleEffectiveBL.computeQtyOrdered(schedule);
 
-				final BigDecimal qtyDeliveredPersisted = CoalesceUtil.coalesce(schedule.getQtyDelivered(), BigDecimal.ZERO);
 				final BigDecimal qtyDeliveredThisShipment = qtyDeliveredByThisShipmentByScheduleId.getOrDefault(scheduleId, BigDecimal.ZERO);
-				// The schedule's own QtyDelivered is written by a deferred, per-InOut runAfterCommit
-				// recompute (M_InOut_Shipment.invalidateShipmentSchedulesForLines), so when this order's
-				// picked lines ship via >=2 sibling InOuts, a schedule fully delivered by an EARLIER
-				// sibling InOut can still read QtyDelivered=0 here - and that sibling's line is not part of
-				// qtyDeliveredThisShipment either. retrieveQtyDelivered sums the MovementQty of the
-				// PROCESSED shipment lines actually linked to the schedule (this InOut + all committed
-				// siblings), independent of the async recompute, so it sees that delivery. Take the max so
-				// a schedule whose persisted QtyDelivered is already ahead of the ledger still counts.
-				final BigDecimal qtyDeliveredCommitted = shipmentScheduleAllocDAO.retrieveQtyDelivered(schedule);
-				final BigDecimal qtyDeliveredTotal = qtyDeliveredPersisted.add(qtyDeliveredThisShipment).max(qtyDeliveredCommitted);
+				final BigDecimal qtyDeliveredTotal = computeQtyDeliveredForCloseDecision(schedule, qtyDeliveredThisShipment);
 
 				if (qtyDeliveredTotal.compareTo(effectiveQtyOrdered) >= 0)
 				{
@@ -763,6 +753,25 @@ public class ShipmentScheduleBL implements IShipmentScheduleBL
 				closeShipmentSchedule(schedule);
 			}
 		}
+	}
+
+	/**
+	 * Effective delivered qty used to decide whether {@link #closePartiallyShipped_ShipmentSchedules}
+	 * may close a schedule. The schedule's own {@code QtyDelivered} is written by a deferred, per-{@code
+	 * M_InOut} {@code runAfterCommit} recompute, so when an order's picked lines ship via several sibling
+	 * {@code M_InOut}s it can still read 0 for a line already fully shipped by an earlier sibling (and
+	 * that sibling's line is not part of {@code qtyDeliveredThisShipment} either). Reading the committed
+	 * processed-shipment-line ledger ({@link IShipmentScheduleAllocDAO#retrieveQtyDelivered}) sees that
+	 * delivery regardless of the async recompute. The {@code max} keeps the persisted figure for the rare
+	 * case it is ahead of the ledger (e.g. a manually-created shipment with no {@code M_ShipmentSchedule_QtyPicked} row).
+	 */
+	private BigDecimal computeQtyDeliveredForCloseDecision(
+			@NonNull final I_M_ShipmentSchedule schedule,
+			@NonNull final BigDecimal qtyDeliveredThisShipment)
+	{
+		final BigDecimal qtyDeliveredPersisted = CoalesceUtil.coalesce(schedule.getQtyDelivered(), BigDecimal.ZERO);
+		final BigDecimal qtyDeliveredCommitted = shipmentScheduleAllocDAO.retrieveQtyDelivered(schedule);
+		return qtyDeliveredPersisted.add(qtyDeliveredThisShipment).max(qtyDeliveredCommitted);
 	}
 
 	public void applyShipmentScheduleChanges(@NonNull final ApplyShipmentScheduleChangesRequest request)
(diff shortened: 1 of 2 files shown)


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentScheduleBL.java
```diff
@@ -754,6 +755,25 @@ public void closePartiallyShipped_ShipmentSchedules(@NonNull final I_M_InOut ino
 		}
 	}
 
+	/**
+	 * Effective delivered qty used to decide whether {@link #closePartiallyShipped_ShipmentSchedules}
+	 * may close a schedule. The schedule's own {@code QtyDelivered} is written by a deferred, per-{@code
+	 * M_InOut} {@code runAfterCommit} recompute, so when an order's picked lines ship via several sibling
```
teosarca: when that deferred code (runAfterCommit) timing will change, i think the logic from here will be broken. is there any automated test that will catch that?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentScheduleBL.java
```diff
@@ -754,6 +755,25 @@ public void closePartiallyShipped_ShipmentSchedules(@NonNull final I_M_InOut ino
...
+	 * that sibling's line is not part of {@code qtyDeliveredThisShipment} either). Reading the committed
+	 * processed-shipment-line ledger ({@link IShipmentScheduleAllocDAO#retrieveQtyDelivered}) sees that
+	 * delivery regardless of the async recompute. The {@code max} keeps the persisted figure for the rare
+	 * case it is ahead of the ledger (e.g. a manually-created shipment with no {@code M_ShipmentSchedule_QtyPicked} row).
+	 */
+	private BigDecimal computeQtyDeliveredForCloseDecision(
+			@NonNull final I_M_ShipmentSchedule schedule,
+			@NonNull final BigDecimal qtyDeliveredThisShipment)
+	{
+		final BigDecimal qtyDeliveredPersisted = CoalesceUtil.coalesce(schedule.getQtyDelivered(), BigDecimal.ZERO);
```
teosarca: schedule.getQtyDelivered() never returns null

pls learn that BigDecimal returning generated methods never return null

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentScheduleBL.java
```diff
@@ -754,6 +755,25 @@ public void closePartiallyShipped_ShipmentSchedules(@NonNull final I_M_InOut ino
...
+	 * Effective delivered qty used to decide whether {@link #closePartiallyShipped_ShipmentSchedules}
+	 * may close a schedule. The schedule's own {@code QtyDelivered} is written by a deferred, per-{@code
+	 * M_InOut} {@code runAfterCommit} recompute, so when an order's picked lines ship via several sibling
+	 * {@code M_InOut}s it can still read 0 for a line already fully shipped by an earlier sibling (and
+	 * that sibling's line is not part of {@code qtyDeliveredThisShipment} either). Reading the committed
+	 * processed-shipment-line ledger ({@link IShipmentScheduleAllocDAO#retrieveQtyDelivered}) sees that
+	 * delivery regardless of the async recompute. The {@code max} keeps the persisted figure for the rare
+	 * case it is ahead of the ledger (e.g. a manually-created shipment with no {@code M_ShipmentSchedule_QtyPicked} row).
+	 */
+	private BigDecimal computeQtyDeliveredForCloseDecision(
```
teosarca: i feel like this approach is more like a mambo jambo "there, i've fixed it" code smell.
isn't any other easy and clean way of doing it?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentScheduleBL.java
teosarca: **Resolution (dev):** The close decision no longer reads `QtyDelivered` at all, so a change in that deferred `runAfterCommit` timing can't break it. As of 43f8a844c4a it computes `delivered = this shipment's MovementQty + retrieveQtyDelivered(schedule)` (the committed processed-shipment-line ledger) and never touches the schedule's own `QtyDelivered` column. The regression test `doesNotCloseScheduleFullyDeliveredBySibling_whenQtyDeliveredNotYetRecomputed` locks it in: it sets `QtyDelivered=0` (recompute not landed) and asserts the schedule stays open purely from the committed ledger — so any reintroduced dependency on `QtyDelivered` would fail that test.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentScheduleBL.java
teosarca: **Resolution (dev):** Fixed in 43f8a844c4a — the reworked code no longer reads `schedule.getQtyDelivered()` at all (it reads the committed ledger), so the `CoalesceUtil.coalesce(...)` is gone. rule added to the Java coding-rules: generated numeric model getters (`BigDecimal`/`int` on `X_*`) never return null → never `coalesce`/null-check them, with a Review-rule.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentScheduleBL.java
teosarca: **Resolution (dev):** Agreed, reworked in 43f8a844c4a. `delivered = this shipment's MovementQty + retrieveQtyDelivered(schedule)` — a plain sum of two **disjoint** sources: this `M_InOut`'s lines are not `Processed` yet at `TIMING_AFTER_COMPLETE` (`completeIt` fires the validate before `setProcessed`), so the processed-line ledger doesn't overlap them. Dropped the persisted `QtyDelivered` (it is only a deferred recompute of that same ledger) and the `max`. `retrieveQtyDelivered` is the exact query `ShipmentScheduleUpdater` uses to populate `QtyDelivered`, so this reads the canonical value synchronously instead of a heuristic — and it is strictly more correct than the `max` form, which under-counted the mixed "part by this shipment + part by an async-stale sibling" case.

backend/de.metas.swat/de.metas.swat.base/src/test/java/de/metas/inoutcandidate/api/impl/ShipmentScheduleBL_closePartiallyShipped_Test.java
```diff
@@ -147,6 +170,122 @@ public void closePartiallyShipped_ShipmentSchedules_doesNotCloseScheduleAlreadyF
...
+	 * {@code M_InOutLine.MovementQty}; if the close trusts the stale {@code QtyDelivered} it wrongly
+	 * closes the fully-shipped schedule, which nothing ever reopens.
+	 */
+	@Test
+	public void closePartiallyShipped_ShipmentSchedules_doesNotCloseScheduleFullyDeliveredBySibling_whenQtyDeliveredNotYetRecomputed()
+	{
+		final OrgId orgId = OrgId.ANY;
+
+		Services.get(ISysConfigBL.class).setValue(
+				"M_ShipmentSchedule_Close_PartiallyShipped",
```
teosarca: pls use de.metas.inoutcandidate.api.impl.ShipmentScheduleBL#SYS_Config_M_ShipmentSchedule_Close_PartiallyShipped.

pls refactor this code snippet to it's own method.

teosarca: or u can introduce de.metas.inoutcandidate.api.impl.ShipmentScheduleBL#setCloseIfPartiallyShipped, as counter part of isCloseIfPartiallyShipped.

teosarca: **Resolution (dev):** Done in 072f528ff66 — made `ShipmentScheduleBL.SYS_Config_M_ShipmentSchedule_Close_PartiallyShipped` package-private, and the 4 tests now enable the sysconfig through a shared `enableCloseIfPartiallyShipped(orgId)` helper that references that constant, replacing the hardcoded `"M_ShipmentSchedule_Close_PartiallyShipped"` literal + 5-line boilerplate that was repeated in each test.

PR #25088, opened 2026-07-11
eInvoice — buyer-reference relabel, seller-contact precedence, silent-factoring IBAN (BT-84)

metas-ts:
Follow-up hardening of the CII (XRechnung/ZUGFeRD) e-invoice mapping.

## What changed

**Buyer reference (BT-10) — relabel the field**
- Migration `5813340_sys_EInvoice_BuyerReference_Rename.sql`: relabels `C_BPartner.EInvoice_BuyerReference` from "Leitweg-ID" to "Leitweg-ID / Käuferreferenz" (de_DE/de_CH) / "Buyer Reference" (en_US) and adds help text (previously empty), via `AD_Element_Trl` + `update_TRL_Tables_On_AD_Element_TRL_Update` propagation. No enforcement change — BT-10 stays required for XRechnung (KoSIT BR-DE-15). The field is the general buyer reference (any agreed reference satisfies BR-DE-15), not a public-authority-only Leitweg-ID.
- `CiiValidatorTest`: adds coverage that an arbitrary, non-Leitweg-format buyer reference passes BR-DE-15 (empty still fails).

**Seller contact (BG-6) — controllable selection**
- `CiiMapper.buildSellerContact` now selects among the seller org BPartner's **active** contacts by precedence: `IsSalesContact` → `IsDefaultContact` → first by `SeqNo`/`AD_User_ID` (preferring a contact with a phone). An explicit-flag match is used even without a phone; inactive contacts are never selected. Previously: first-with-phone-else-first, no active filter, not user-controllable.

**Silent factoring — factorer IBAN in BT-84**
- `CiiMapper.buildPaymentMeans`: when the invoice's bill partner has `IsFactoring=Y`, BT-84 carries the org's factorer (`IsFactorer=Y`) default bank-account IBAN instead of the seller's; the seller/payee identity is unchanged (no BG-10 Payee party). Misconfiguration (no factorer / multiple factorers / factorer without an IBAN) fails fast with a user-validation error — the seller IBAN is never emitted on a factored invoice. The factorer lookup mirrors the existing printed-invoice report function so PDF and XML select the same factorer.
- Adds `CiiMapperTest` unit coverage (factoring, non-factoring regression, and the three misconfiguration cases) and a cucumber E2E asserting the emailed XRechnung carries the factorer IBAN.

**Unit code (BT-130)** — no code change: an audit confirmed the in-use unit of measure carries a valid UN/ECE Rec 20 code (`PCE`, remapped to `C62`); a broader UOM-code cleanup is tracked separately.


backend/de.metas.einvoice.base/src/main/java/de/metas/einvoice/cii/CiiMapper.java
```diff
@@ -996,6 +1052,78 @@ private List<TradeSettlementPaymentMeansType> buildPaymentMeans(
 		return Collections.singletonList(paymentMeans);
 	}
 
+	/**
+	 * Resolves the payee IBAN for a factored ({@code C_BPartner.IsFactoring=Y}) invoice: looks up the
```
metas-ts: WAY too much prose here

backend/de.metas.einvoice.base/src/main/java/de/metas/einvoice/cii/CiiMapper.java
```diff
@@ -585,6 +589,42 @@ private TradeContactType buildSellerContact(@NonNull final I_C_BPartner sellerBP
 		return tradeContact;
 	}
 
+	/**
+	 * Selects the seller contact to use for BG-6 among the given (already active-filtered) contacts,
+	 * which are expected to be ordered by {@code SeqNo, AD_User_ID}. See {@link #buildSellerContact}
+	 * for the full precedence description.
+	 */
+	private I_AD_User selectSellerContact(@NonNull final List<I_AD_User> activeContacts)
```
metas-ts, line 597: don'T spaghetty noew code. use existing org.compiere.apps.search.UserQuery 
extend that if needed

backend/de.metas.einvoice.base/src/main/java/de/metas/einvoice/cii/CiiMapper.java
metas-ts: **Resolution (dev):** Trimmed from ~22 lines to 5 — kept only the one load-bearing caveat (the factorer lookup mirrors the printed-invoice report SQL `getFactorer_BankDetails` so PDF and XML pick the same factorer; single/default bank account assumed for IBAN parity) and the fail-fast note. Done in https://github.com/metasfresh/metasfresh/commit/cd50dfa85d9270ad750b2e1cdcd99b5035d045a3.

backend/de.metas.einvoice.base/src/main/java/de/metas/einvoice/cii/CiiMapper.java
metas-ts: **Resolution (dev):** Looked into both — a couple of findings before I refactor, since the drop-in isn't clean:

- `org.compiere.apps.search.UserQuery` is a *saved-user-search* value object (package-private; id/caption/AD_User_ID/restrictions, used to build WebUI document filters). It doesn't select a contact, so it isn't the right tool here.
- The existing prioritized contact retrieval is `IBPartnerDAO.retrieveContact(ctx, bpartnerId, isSOTrx, trxName)`. For `isSOTrx=true` it returns the first active contact with `IsSalesContact='Y' AND IsSalesContact_Default='Y'`, and **null** otherwise.

That differs from what this change needs in two ways: the CII seller contact (BG-6) must **always** be present — XRechnung BR-DE-2 mandates a `DefinedTradeContact` — and the agreed selection here is `IsSalesContact → IsDefaultContact → first active (SeqNo/phone)`. `retrieveContact` returns null when there is no *default* sales contact, and keys on `IsSalesContact_Default` (not `IsDefaultContact`), so reusing it as-is would change the agreed behaviour and drop the mandatory-contact fallback.

Two ways forward — your call:
1. **Keep the agreed precedence, move it out of the mapper**: add a reusable `IBPartnerDAO` method encapsulating `IsSalesContact → IsDefaultContact → first-active` (active-filtered, always returns one if any exists) and have the mapper delegate to it. Removes the in-mapper loops, keeps behaviour + tests.
2. **Conform to `retrieveContact`'s `IsSalesContact_Default` semantics** — this changes the agreed selection rule (and needs a `DefinedTradeContact`-always fallback added to it), so I'd confirm with the requirements owner first.

I'd go with (1) unless you prefer (2). Which do you want?

metas-ts: go with de.metas.bpartner.service.IBPartnerDAO#retrieveContact. make sure there is a proper error-message if no sales contact is found

metas-ts, line 597: going with dedicated method for this particular eInvoice requirement

metas-ts: **Resolution (dev):** Understood — going with a dedicated `IBPartnerDAO` method for this eInvoice seller-contact requirement (based on `retrieveContact`'s sales-contact selection), delegating `CiiMapper.buildSellerContact` to it, with a clear user-validation error when no sales contact is found. Since this PR is already merged, it ships as a follow-up PR on `deep_tundra_release`; I'll link it here and in the R&I.

metas-ts: **Resolution (dev):** Understood — going with a dedicated `IBPartnerDAO` method for this eInvoice seller-contact requirement (based on `retrieveContact`'s sales-contact selection), delegating `CiiMapper.buildSellerContact` to it, with a clear user-validation error when no sales contact is found. Since this PR is already merged, it ships as a follow-up PR on `deep_tundra_release`; I'll link it here and in the R&I.

metas-ts: **Resolution (dev):** Update — for this issue we are keeping the current seller-contact selection as-is; no follow-up PR will be opened. Noted here so the thread is not left expecting one.

PR #25109, opened 2026-07-13
eInvoice — tax-included invoices: emit tax-exclusive EN16931 amounts (BR-S-08)

metas-ts:
## Problem

On a **tax-included** invoice (`C_Invoice.IsTaxIncluded = Y`), `CiiMapper` emitted the EN16931 line/total amounts from the **gross** metasfresh columns:

- BT-146 (item net price) ← `C_InvoiceLine.PriceActual` (gross)
- BT-131 (line net amount) ← `C_InvoiceLine.LineNetAmt` (gross)
- BT-106 / BT-109 (sum of line net amounts / tax basis total) ← `C_Invoice.TotalLines` (gross)

while BT-116 (`C_InvoiceTax.TaxBaseAmt`) was already **net**. The line total (BT-131) therefore did not equal the VAT-breakdown taxable base (BT-116), so KoSIT schematron rejected the document with **BR-S-08** at invoice completion — and, on multi-line invoices, **BR-CO-10**.

## Fix

On the **tax-included path only** (the tax-excluded path is byte-identical to before):

- BT-146 and BT-131 are converted to the tax-exclusive net via `Tax.calculateBaseAmt`.
- BT-131 is **reconciled per `C_Tax_ID` group** so the group's line-net sum equals that tax's `C_InvoiceTax.TaxBaseAmt` **exactly** — the last line of each group absorbs the rounding delta. This is required because a naive per-line conversion sum-of-rounds can differ by a cent from the breakdown's round-of-sum `TaxBaseAmt`, which would still fail BR-S-08 / BR-CO-10.
- BT-106 / BT-109 are sourced from `SUM(C_InvoiceTax.TaxBaseAmt)` instead of the gross `TotalLines`.
- The line-net gate keys on the invoice-level `IsTaxIncluded` flag, consistent across BT-131/BT-146/BT-106/BT-109.
- Fail-fast (with `C_Tax_ID` + `C_Invoice_ID`) if a line references a tax that has no `C_InvoiceTax` breakdown row, rather than silently emitting broken XML.

## Tests

`de.metas.einvoice.base` (`CiiMapperTest` 27, `CiiValidatorTest` 7, all green):

- single-line tax-included: BT-131/BT-146/BT-106/BT-109 emitted net (not gross);
- multi-line tax-included: per-line BT-131 reconciles exactly to `TaxBaseAmt` (rounding-delta absorption);
- KoSIT schematron validation of a multi-line tax-included XRechnung: no fatal/error rules, specifically **no BR-S-08 / no BR-CO-10**;
- tax-excluded whole-tax line: BT-146 remains the raw `PriceActual` (regression guard for the byte-identical tax-excluded path);
- missing `C_InvoiceTax` breakdown row: fail-fast throw carrying both IDs.


backend/de.metas.einvoice.base/src/main/java/de/metas/einvoice/cii/CiiMapper.java
```diff
@@ -242,20 +262,128 @@ private SupplyChainTradeTransactionType buildTradeTransaction(
...
+
+			// The last line of the group absorbs the rounding delta so the group's BT-131 sum
+			// matches this tax's TaxBaseAmt (BT-116) EXACTLY — this is what satisfies BR-S-08/BR-CO-10.
+			final BigDecimal delta = targetTaxBaseAmt.subtract(sumOfNaturalNets);
+			final I_C_InvoiceLine lastLine = groupLines.get(groupLines.size() - 1);
+			for (final I_C_InvoiceLine line : groupLines)
+			{
+				final InvoiceLineId lineId = InvoiceLineId.ofRepoId(line.getC_InvoiceLine_ID());
+				final BigDecimal naturalNet = naturalNetByLineId.get(lineId);
+				final BigDecimal net = line == lastLine ? naturalNet.add(delta) : naturalNet;
```
metas-ts, line 380: i guess one can derive a max possible delta using the currency's precision, right? 
please add an assertion (using `de.metas.common.util.Check#assume...`) so we can be sure it's fine

backend/de.metas.einvoice.base/src/main/java/de/metas/einvoice/cii/CiiMapper.java
```diff
@@ -242,20 +262,128 @@ private SupplyChainTradeTransactionType buildTradeTransaction(
...
+	 * {@code M_PriceList_ID}-keyed {@code AbstractInvoiceBL.getAmountPrecision}, since a {@link I_C_Invoice}
+	 * reaching this mapper is not guaranteed to carry a price list reference.
+	 */
+	@NonNull
+	private CurrencyPrecision resolveAmountPrecision(@NonNull final I_C_Invoice invoice)
+	{
+		final CurrencyId currencyId = CurrencyId.ofRepoIdOrNull(invoice.getC_Currency_ID());
+		return currencyId != null
+				? currencyDAO.getStdPrecision(currencyId)
+				: CurrencyPrecision.TWO;
```
metas-ts: this is nonsense: the code won'T be executed for an invoice without C_Currency

backend/de.metas.einvoice.base/src/main/java/de/metas/einvoice/cii/CiiMapper.java
metas-ts: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/58de98898a957ca8b47b99d7f5cb30661652444e. The delta absorbed on the group's last line can only be the per-line rounding gap, so I derive `maxDelta = oneMinorUnit × lineCount` (`oneMinorUnit = BigDecimal.ONE.movePointLeft(amountPrecision.toInt())`, amountPrecision = the invoice currency's std precision) and `Check.assume(delta.abs() <= maxDelta, …)` before applying it — so a divergence beyond rounding (e.g. gross fed into the reconciliation, or an inconsistent breakdown) now fails loudly with C_Tax_ID + C_Invoice_ID instead of silently distorting a line. Covered by a RED→GREEN test `tax_included_reconciliationDeltaExceedsBound_throwsWithIds` (TaxBaseAmt 90.00 vs derived net 84.03 → delta 5.97 ≫ 0.01 bound → throws).

backend/de.metas.einvoice.base/src/main/java/de/metas/einvoice/cii/CiiMapper.java
metas-ts: **Resolution (dev):** Agreed — dropped in https://github.com/metasfresh/metasfresh/commit/58de98898a957ca8b47b99d7f5cb30661652444e. `C_Currency_ID` is mandatory on `C_Invoice` (and required for BT-5 / BR-CO-15), so `resolveAmountPrecision` now resolves directly via `CurrencyId.ofRepoId(invoice.getC_Currency_ID())` + `getStdPrecision`; the `CurrencyPrecision.TWO` fallback branch is removed.

backend/de.metas.einvoice.base/src/main/java/de/metas/einvoice/cii/CiiMapper.java
metas-ts: **Resolution (dev):** One note on the package: I used `de.metas.util.Check` rather than `de.metas.common.util.Check` (https://github.com/metasfresh/metasfresh/commit/396b7d2db7bc7dbb781c9a4e080766a273bd8a95) — identical `assume(boolean, String, Object...)`, but it's the dominant variant in the backend and this file already imports `de.metas.util.Services` from the same dependency. Happy to switch to the common one if you'd prefer.

PR #25142, opened 2026-07-14
Cost Revaluation: CopyFromCostElement source (value-neutral switch to Moving Average Invoice)

teosarca:
## What

Adds a **`CopyFromCostElement`** source to the Cost Revaluation document, to switch an accounting schema's costing method to **Moving Average Invoice** *value-neutrally* — by seeding the target (MAI) cost element from a source cost element's current cost at a cutoff, without any GL impact.

> ⚠️ **Draft / work-in-progress.** This is scaffolded incrementally (9 tasks). Tasks 1–5 are in; Tasks 6–9 (GL-posting suppression, reversal + guard, edge-case/idempotency guards, cucumber E2E) are still to come. Not ready to merge.

## Approach

`M_CostRevaluation` gets two header columns:
- **`RevaluationSource`** (ref-list): `Calculated` (existing default) · `CopyFromCostElement` (this feature).
- **`CopyFrom_M_CostElement_ID`**: the source element to copy from (the existing `M_CostElement_ID` stays the target).

On **Create Lines** (`CopyFromCostElement`): one line per source `M_Cost`, previewing the source's current cost/qty.
On **Complete**: directly set the target element's `M_Cost` to the source's current own price + LL + qty (read atomically, fresh), and write ONE anchoring `M_CostDetail` dated at the cutoff with `Qty=0`, `Amt=0`, `IsChangingCosts=Y`, and `Prev_*` = the seeded opening. That anchor is the make-or-break: it lets `product_costs_recreate_from_date` restore `M_Cost` from `Prev_*` when there are no later transactions (survives the recompute).

The method-flip (`C_AcctSchema.CostingMethod`) stays a separate action — this document maintains the cost *data* only.

## Commits so far

| Commit | |
|---|---|
| add `RevaluationSource` + `CopyFrom_M_CostElement_ID` (+ review fixes) | AD migration |
| regen `M_CostRevaluation` model | generated `I_/X_` |
| `RevaluationSource` enum (`ReferenceListAwareEnum`) | |
| CopyFromCostElement line creation | + createLines |
| guard `RevaluationSource`/`CopyFrom` against edit while lines exist | interceptor + test |
| extract shared line-query helpers | dedup |
| CopyFromCostElement complete-time seed + opening anchor | core seed logic |

> A correctness follow-up — **seed reads the full source snapshot atomically (own+LL+qty) at complete time** instead of mixing stale create-lines values with a fresh LL read — is landing into this PR now (found in review; unit-test-covered, GREEN locally).

## Tests

Unit tests in `CostRevaluationServiceTest` / `M_CostRevaluationTest` (`de.metas.business`), all TDD (RED verified before each). Covers: line creation (incl. zero-on-hand + rerun), the interceptor guard, the complete-time direct-set + opening-anchor shape (`Prev_*`=opening / qty=0 / amt=0), and that the seed uses a *fresh atomic* source snapshot rather than stale create-lines-time values.

## Still to do (Tasks 6–9)

- Suppress GL posting for `CopyFromCostElement` (value-neutral → zero `Fact_Acct`).
- Reversal + guard (refuse if a later cost event exists).
- Edge-case guards: idempotency (refuse if target already seeded), self-copy no-op, zero/neg stock, org/multi-ASI seeding.
- Cucumber end-to-end (recompute survival + value-neutral + forward MAI).


backend/de.metas.business/src/main/java/de/metas/costing/impl/CostingService.java
```diff
@@ -552,4 +554,39 @@ public CostsRevaluationResult revaluateCosts(@NonNull CostsRevaluationRequest re
...
+	 * bypassing the {@code Calculated} history-replay of {@link #revaluateCosts(CostsRevaluationRequest)}, and
+	 * write ONE anchoring {@code M_CostDetail} dated {@code anchorDate} with {@code Qty=0}, {@code Amt=0},
+	 * {@code IsChangingCosts=Y} and {@code Prev_*} = the SAME {@code opening}.
+	 * <p>
+	 * The shared {@code opening} object is the correctness crux: the anchor's {@code Prev_*} snapshot equals the
+	 * seeded {@code M_Cost}, so {@code product_costs_recreate_from_date} (which restores {@code M_Cost} from the
+	 * first in-range cost detail's {@code Prev_*}) reproduces exactly this state when there are no later transactions.
+	 */
+	public void seedCurrentCostFromOpening(
+			@NonNull final CostSegmentAndElement targetSegAndElem,
```
teosarca: pls avoid such abbreviations. favor targetSegmentAndElement
capture learning.

backend/de.metas.business/src/main/java/de/metas/costrevaluation/CostRevaluation.java
```diff
@@ -47,4 +47,8 @@ public class CostRevaluation
 	@NonNull InstantAndOrgId dateAcct;
 
 	@NonNull DocStatus docStatus;
+
+	@NonNull RevaluationSource revaluationSource;
+	/** Source cost element for {@link RevaluationSource#CopyFromCostElement}; {@code null} for {@link RevaluationSource#Calculated}. */
+	CostElementId copyFromCostElementId;
```
teosarca: annotate with `@Nullable`.
capture learning

backend/de.metas.business/src/main/java/de/metas/costrevaluation/CostRevaluationService.java
```diff
@@ -224,4 +286,62 @@ private void createDetails(@NonNull final CostRevaluation costRevaluation, @NonN
...
+	 * Builds ONE {@link CostDetailPreviousAmounts opening} object and reuses it for BOTH the {@code M_Cost} seed and the
+	 * anchor's {@code Prev_*} — that shared snapshot is the recompute-survival crux. Own price and qty come from the line;
+	 * the lower-level (LL/component) price is re-read fresh from the SOURCE element's {@code M_Cost} (it is not carried on
+	 * the line, mirroring the {@code Calculated} path).
+	 */
+	private void createDetailsForCopyFromCostElement(
+			@NonNull final CostRevaluation costRevaluation,
+			@NonNull final CostRevaluationLine line)
+	{
+		final CostElementId sourceCostElementId = costRevaluation.getCopyFromCostElementId();
```
teosarca, line 401: make sure M_CostRevalution.CopyFrom_CostElement_ID has MandatoryLogic whhich makes it manatoty when reval source is copy

backend/de.metas.business/src/main/java/de/metas/costrevaluation/RevaluationSource.java
```diff
@@ -0,0 +1,85 @@
...
+	/** Cost is recalculated for this revaluation (default, existing behaviour). */
+	Calculated(X_M_CostRevaluation.REVALUATIONSOURCE_Calculated), //
+	/** Cost is copied unchanged from the cost element selected under CopyFrom_M_CostElement_ID (value-neutral). */
+	CopyFromCostElement(X_M_CostRevaluation.REVALUATIONSOURCE_CopyFromCostElement) //
+	;
+
+	@Getter
+	private final String code;
+
+	RevaluationSource(final String code)
```
teosarca: use all args ctor annotation
capture learning

backend/de.metas.business/src/main/java/de/metas/costrevaluation/RevaluationSource.java
```diff
@@ -0,0 +1,85 @@
...
+public enum RevaluationSource implements ReferenceListAwareEnum
+{
+	/** Cost is recalculated for this revaluation (default, existing behaviour). */
+	Calculated(X_M_CostRevaluation.REVALUATIONSOURCE_Calculated), //
+	/** Cost is copied unchanged from the cost element selected under CopyFrom_M_CostElement_ID (value-neutral). */
+	CopyFromCostElement(X_M_CostRevaluation.REVALUATIONSOURCE_CopyFromCostElement) //
+	;
+
+	@Getter
+	private final String code;
```
teosarca: use NonNull annottion.
capture learning

backend/de.metas.business/src/main/java/de/metas/costrevaluation/RevaluationSource.java
```diff
@@ -0,0 +1,85 @@
...
+		return code != null ? ofCode(code) : null;
+	}
+
+	@Nullable
+	public static String toCodeOrNull(@Nullable final RevaluationSource source)
+	{
+		return source != null ? source.getCode() : null;
+	}
+
+	private static final ImmutableMap<String, RevaluationSource> typesByCode = ReferenceListAwareEnums.indexByCode(values());
```
teosarca: use ValuesIndex
capture learning

backend/de.metas.business/src/main/java/de/metas/costrevaluation/RevaluationSource.java
```diff
@@ -0,0 +1,85 @@
...
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+/**
+ * NOTE to developers: Please keep in sync with list reference "M_CostRevaluation RevaluationSource" {@code AD_Reference_ID=542117}
+ */
+public enum RevaluationSource implements ReferenceListAwareEnum
```
teosarca, line 41: the enum have several problems. see below.
pls find out an example how a proper enum shall be and pls improve corpus so next time when u have to define it you will do it 100% flawless

backend/de.metas.business/src/test/java/de/metas/costrevaluation/interceptor/M_CostRevaluationTest.java
```diff
@@ -0,0 +1,316 @@
...
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+/**
+ * Package-private on purpose: needs to reach {@link M_CostRevaluation} (package-private interceptor class).
+ * Setup mirrors {@code de.metas.costrevaluation.CostRevaluationServiceTest#beforeEach} (copied, kept self-contained).
+ */
+@ExtendWith(AdempiereTestWatcher.class)
+class M_CostRevaluationTest
```
teosarca, line 97: i prefer cucumber tests and aovid junit. read the doctrine

backend/de.metas.business/src/test/java/de/metas/costrevaluation/CostRevaluationServiceTest.java
```diff
@@ -0,0 +1,389 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+@ExtendWith(AdempiereTestWatcher.class)
+public class CostRevaluationServiceTest
```
teosarca, line 100: i prefer cucumber tests and aovid junit. read the doctrine

backend/de.metas.business/src/main/java/de/metas/costing/impl/CostingService.java
teosarca: **Resolution (dev):** Renamed to full words — `targetSegAndElem`→`targetSegmentAndElement`, `sourceSegAndElem`→`sourceSegmentAndElement` — in https://github.com/metasfresh/metasfresh/commit/30bd8a75a2afa4af6744ac8cf61501032d7cdece. Capturing a no-abbreviations naming rule in the corpus.

backend/de.metas.business/src/main/java/de/metas/costrevaluation/CostRevaluation.java
teosarca: **Resolution (dev):** Added `@Nullable` to `CostRevaluation.copyFromCostElementId` (null for `Calculated`) in https://github.com/metasfresh/metasfresh/commit/30bd8a75a2afa4af6744ac8cf61501032d7cdece. Captured with the annotation conventions.

backend/de.metas.business/src/main/java/de/metas/costrevaluation/CostRevaluationService.java
teosarca: **Resolution (dev):** Will add a migration setting `MandatoryLogic` on `M_CostRevaluation.CopyFrom_M_CostElement_ID` so it is mandatory when `RevaluationSource='CopyFromCostElement'` (today enforced only by a runtime AdempiereException). Coming as the next migration commit in this PR.

backend/de.metas.business/src/main/java/de/metas/costrevaluation/RevaluationSource.java
teosarca: **Resolution (dev):** Enum now uses `@AllArgsConstructor` (manual constructor removed) — https://github.com/metasfresh/metasfresh/commit/30bd8a75a2afa4af6744ac8cf61501032d7cdece.

backend/de.metas.business/src/main/java/de/metas/costrevaluation/RevaluationSource.java
teosarca: **Resolution (dev):** `@NonNull` on the `code` field (null-check generated by `@AllArgsConstructor`) — https://github.com/metasfresh/metasfresh/commit/30bd8a75a2afa4af6744ac8cf61501032d7cdece.

backend/de.metas.business/src/main/java/de/metas/costrevaluation/RevaluationSource.java
teosarca: **Resolution (dev):** Replaced the manual `ImmutableMap`+`indexByCode` with `ReferenceListAwareEnums.ValuesIndex` (`index.ofCode`/`ofNullableCode`) — https://github.com/metasfresh/metasfresh/commit/30bd8a75a2afa4af6744ac8cf61501032d7cdece.

backend/de.metas.business/src/main/java/de/metas/costrevaluation/RevaluationSource.java
teosarca: **Resolution (dev):** Rewrote `RevaluationSource` to the canonical `ReferenceListAwareEnum` pattern — `@AllArgsConstructor`, `@NonNull code`, `ReferenceListAwareEnums.ValuesIndex` — matching `ESRType`/`ServiceRepairProjectTaskType` (https://github.com/metasfresh/metasfresh/commit/30bd8a75a2afa4af6744ac8cf61501032d7cdece). Capturing this canonical enum recipe in the corpus so it's done right first time.

backend/de.metas.business/src/test/java/de/metas/costrevaluation/interceptor/M_CostRevaluationTest.java
teosarca: **Resolution (dev):** Acknowledged the cucumber-first preference. The end-to-end behaviour (complete → value-neutral seed → recompute survival → forward MAI) is planned as the Task-9 cucumber E2E of this feature. This JUnit currently gives fast unit coverage of the interceptor guard. Do you want the unit-level JUnit dropped in favour of cucumber-only, or kept alongside the Task-9 cucumber E2E? Will follow your call.

backend/de.metas.business/src/test/java/de/metas/costrevaluation/CostRevaluationServiceTest.java
teosarca: **Resolution (dev):** Same as the interceptor-test note — this JUnit covers the unit logic (line seeding, the value-neutral seed + opening-anchor shape); the full behaviour is the planned Task-9 cucumber E2E. Keep as fast unit coverage or convert to cucumber-only? Your call.

backend/de.metas.business/src/main/java/de/metas/costrevaluation/CostRevaluationService.java
```diff
@@ -62,17 +65,70 @@ public void createLines(@NonNull final CostRevaluationId costRevaluationId)
 	{
 		final CostRevaluation costRevaluation = costRevaluationRepository.getById(costRevaluationId);
 
+		if (costRevaluation.getRevaluationSource().isCopyFromCostElement())
+		{
+			createLinesFromCopyFromCostElement(costRevaluationId, costRevaluation);
+			return;
+		}
```
teosarca, line 71: in such cases always go with if/else.
the code below and wihch shall be in "else" part shall be actually refactored to a new method similar name to createLinesFromCopyFromCostElement.

capture learning

backend/de.metas.business/src/main/java/de/metas/costrevaluation/CostRevaluationService.java
```diff
@@ -125,6 +179,12 @@ private void createDetails(@NonNull final CostRevaluation costRevaluation, @NonN
 			throw new AdempiereException("Line already revaluated: " + line.getId());
 		}
 
+		if (costRevaluation.getRevaluationSource().isCopyFromCostElement())
+		{
+			createDetailsForCopyFromCostElement(costRevaluation, line);
+			return;
+		}
```
teosarca, line 283: in such cases always go with if/else.
the code below and wihch shall be in "else" part shall be actually refactored to a new method similar name to createDetailsForCopyFromCostElement.

capture learning

backend/de.metas.business/src/main/java/de/metas/costrevaluation/CostRevaluationService.java
teosarca: and a DisplayLogic ...

backend/de.metas.business/src/test/java/de/metas/costrevaluation/interceptor/M_CostRevaluationTest.java
teosarca: our doctrine/corpus is clearing telling u how to proceed.
favor e2e tests.
avoid overlapping.

again a question u could answer by simply checking the corpus

backend/de.metas.business/src/test/java/de/metas/costrevaluation/CostRevaluationServiceTest.java
teosarca: our doctrine/corpus is clearing telling u how to proceed.
favor e2e tests.
avoid overlapping.

again a question u could answer by simply checking the corpus

backend/de.metas.business/src/main/java/de/metas/costrevaluation/CostRevaluationService.java
teosarca: **Resolution (dev):** Done — `createLines` now uses an explicit if/else and the `Calculated` branch is extracted into `createLinesFromCalculated` (mirroring `createLinesFromCopyFromCostElement`). https://github.com/metasfresh/metasfresh/commit/665e7b9554b13e6724d3f19537329936fc929fa9 — 14/14 business costrevaluation tests green. Learning captured as an internal coding rule: when a method branches on a discriminator into alternative strategies, prefer explicit if/else with each branch in a symmetrically-named private method, not early-return-on-one-branch + fall-through.

backend/de.metas.business/src/main/java/de/metas/costrevaluation/CostRevaluationService.java
teosarca: **Resolution (dev):** Done — `createDetails` now uses an explicit if/else and the `Calculated` branch is extracted into `createDetailsForCalculated` (mirroring `createDetailsForCopyFromCostElement`). https://github.com/metasfresh/metasfresh/commit/665e7b9554b13e6724d3f19537329936fc929fa9 . Same learning captured as an internal coding rule (if/else + symmetric extracted methods over early-return + fall-through).

backend/de.metas.business/src/main/java/de/metas/costrevaluation/CostRevaluationService.java
teosarca: **Resolution (dev):** Already implemented — migration `5813920_sys_gh26253_M_CostRevaluation_CopyFrom_MandatoryLogic.sql` sets `AD_Column.MandatoryLogic = '@RevaluationSource@=''CopyFromCostElement'''` on `CopyFrom_M_CostElement_ID`, so the field is mandatory exactly when RevaluationSource = CopyFromCostElement (and optional for Calculated). Verified on the local stack.

backend/de.metas.business/src/test/java/de/metas/costrevaluation/CostRevaluationServiceTest.java
teosarca: **Resolution (dev):** Doctrine acknowledged — favour cucumber e2e, avoid JUnit for flows an e2e can prove. The CopyFromCostElement service flow (seed the target M_Cost from the source element at cutoff -> survive a recompute-from-date -> value-neutral / no GL -> forward MovingAverageInvoice) is being built as a cucumber e2e (the active next task). Once that e2e is green, the overlapping flow-level tests in this JUnit class are removed — sequenced replace-then-remove so there is no coverage gap in between. JUnit is kept only where it is genuinely unit-unique (e.g. the RevaluationSource enum).

backend/de.metas.business/src/test/java/de/metas/costrevaluation/interceptor/M_CostRevaluationTest.java
teosarca: **Resolution (dev):** Doctrine acknowledged — same plan. The interceptor guard (RevaluationSource / CopyFrom_M_CostElement_ID immutable once active lines exist) will be exercised end-to-end by driving the guarded change through the document in the cucumber e2e; this JUnit class is then removed once that e2e is green (replace-then-remove, no coverage gap).

backend/de.metas.business/src/main/java/de/metas/costrevaluation/CostRevaluationService.java
teosarca: **Resolution (dev):** DisplayLogic is already in place — `AD_Field` 781421 (`CopyFrom_M_CostElement_ID` on tab 546464) has `DisplayLogic=@RevaluationSource@=CopyFromCostElement`, set in the first migration `5813790_sys_gh26253_M_CostRevaluation_RevaluationSource.sql` (line 208). So the field is hidden unless `RevaluationSource=CopyFromCostElement`, and — via the MandatoryLogic added in `5813920` — mandatory once shown.

backend/de.metas.business/src/main/java/de/metas/costing/impl/CostDetailRepository.java
```diff
@@ -417,6 +419,31 @@ public boolean hasCostDetailsByProductId(@NonNull final ProductId productId)
...
+		// ONE batched, indexed query over the given product set (no per-product N+1). The M_CostRevaluationLine_ID is set
+		// ONLY on cost details written by a completed M_CostRevaluation line (see #updateRecordFromDocumentRef), so its
+		// presence for (acctSchema, cost element, product) means a completed cost-revaluation line has already written a
+		// detail here — regardless of RevaluationSource (this is the broad, source-agnostic signal, NOT restricted to a
+		// prior CopyFromCostElement switch).
+		return queryBL.createQueryBuilder(I_M_CostDetail.class)
+				.addEqualsFilter(I_M_CostDetail.COLUMNNAME_C_AcctSchema_ID, acctSchemaId)
+				.addEqualsFilter(I_M_CostDetail.COLUMNNAME_M_CostElement_ID, costElementId)
+				.addInArrayFilter(I_M_CostDetail.COLUMNNAME_M_Product_ID, productIds)
+				.addCompareFilter(I_M_CostDetail.COLUMNNAME_M_CostRevaluationLine_ID, Operator.GREATER, 0)
```
teosarca: imho this shall be checked to not be null.
capture this learning when checking IDs.

backend/de.metas.business/src/main/java/de/metas/costing/impl/CostingService.java
```diff
@@ -552,4 +560,97 @@ public CostsRevaluationResult revaluateCosts(@NonNull CostsRevaluationRequest re
...
+	 * average has moved on, so restoring the opening would corrupt it. The seed's own anchor is dated AT the
+	 * cut-off, so it is excluded by the strictly-after filter and never counts as a forward event.
+	 */
+	@Override
+	public void reverseSeededCurrentCost(
+			@NonNull final CostSegmentAndElement targetSegmentAndElement,
+			@NonNull final Instant cutoffDate,
+			@NonNull final CostRevaluationLineId lineId)
+	{
+		final boolean hasPostCutoffCostEvents = costDetailsService.stream(
```
teosarca: imho u shall push this down to repository and name the method exists.
so u don't need to mambo jambo stream->findAny->isPresent.

backend/de.metas.business/src/main/java/de/metas/costing/impl/CostDetailRepository.java
teosarca: **Resolution (dev):** Done. Changed the skip-guard filter to `.addNotNull(COLUMNNAME_M_CostRevaluationLine_ID)`. The FK is nullable and unset is stored as SQL NULL (verified on the instance: 271 NULL / 2919 positive / 0 zeros), so `IS NOT NULL` is the correct, intention-revealing "a revaluation line is linked" check. Captured the learning as an internal coding rule: in a query filter, a nullable-FK is-set uses `addNotNull` (SQL NULL for unset), distinct from the Java-getter `> 0` idiom. Local build/tests verifying; code commit URL to follow.

backend/de.metas.business/src/main/java/de/metas/costing/impl/CostingService.java
teosarca: **Resolution (dev):** Done. Pushed it down to the repository as a named method: added `hasCostDetails(CostDetailQuery)` to `ICostDetailRepository`/`CostDetailRepository` (`toSqlQuery(query).anyMatch()`, mirroring the existing `hasCostDetailsByProductId`) and exposed it via `ICostDetailService`/`CostDetailService`. `reverseSeededCurrentCost` now calls `costDetailsService.hasCostDetails(query)` instead of `stream(query).findAny().isPresent()`. Local build/tests verifying; code commit URL to follow.

backend/de.metas.business/src/main/java/de/metas/costing/impl/CostDetailRepository.java
teosarca: **Resolution (dev):** Landed in https://github.com/metasfresh/metasfresh/commit/002c4e926d9 — the nullable FK is-set check now uses `addNotNull`. Build + CostRevaluationServiceTest/M_CostRevaluationTest green (12/12).

backend/de.metas.business/src/main/java/de/metas/costing/impl/CostingService.java
teosarca: **Resolution (dev):** Landed in https://github.com/metasfresh/metasfresh/commit/002c4e926d9 — named `hasCostDetails(CostDetailQuery)` repository method; `reverseSeededCurrentCost` calls it instead of the inline stream→findAny→isPresent. Build + tests green (12/12).

backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/DocLine_CostRevaluation.java
```diff
@@ -34,31 +37,41 @@ public CostAmount getCreateCosts(@NonNull final AcctSchema as)
...
-			// 				.acctSchemaId(as.getId())
-			// 				.reversalDocumentRef(CostingDocumentRef.ofCostRevaluationLineId(get_ID()))
-			// 				.initialDocumentRef(CostingDocumentRef.ofCostRevaluationLineId(getReversalLine_ID()))
-			// 				.date(getDateAcctAsInstant())
-			// 				.build())
-			// 		.getTotalAmountToPost(as);
+			// Not reachable for this document type: M_CostRevaluationLine has no Reversal_ID, so no reversal line is
+			// ever posted. Reversal of a CopyFromCostElement switch is value-neutral and handled in-place by
+			// CostRevaluationDocumentHandler#reverseCorrectIt, not through posting. Fail fast if ever hit.
+			throw new UnsupportedOperationException("Posting a M_CostRevaluation reversal line is not supported");
```
teosarca, line 43: if posting is not supportd, we might prevent reversing the document - poka-yoke

teosarca: **Resolution (dev):** Reversal *is* supported for CopyFromCostElement and is value-neutral + in-place — `reverseCorrectIt` → `reverseSeededCurrentCost` (removes the opening anchor, resets M_Cost), it never posts a reversal *line*. So we do not want to prevent reversing a supported document. The genuinely-unsupported case is already prevented at the document boundary (poka-yoke): `reverseCorrectIt` refuses a non-Completed doc, and `reverseDetails` refuses a `Calculated`-source doc **upfront, before any state change**, with a clear message ("Reversal is only implemented for the CopyFromCostElement source"). This `throw` guards a *different*, unreachable path — posting a reversal *line* — which cannot occur: `M_CostRevaluationLine` has no `Reversal_ID`, so `isReversalLine()` is always false (the comment documents that). Net: no way to reach the throw, and the real unsupported-reversal case is already blocked early. If you would rather drop the dead branch entirely than keep it as a fail-fast landmine, say so and I will remove it.

PR #25144, opened 2026-07-14
EDI EPCIS: exactly-once, no-orphan SSCC transmission (per-LU close-gate + transmission ledger + reverse guard)

metas-ts:
## What & why

Fixes two production defects in the EPCIS SSCC event export (per-physical-pallet events sent to an EPCIS receiver), so that **every physical SSCC pallet is transmitted to the receiver exactly once, complete, after it closes**:

- **Q1 — no-orphan:** a shipment's fully-shipped *standalone* pallets were never transmitted when a *shared* pallet on the same shipment was still open. The outbound-selection gate was all-or-nothing: any not-yet-covered LU on the shipment suppressed the whole shipment's events, orphaning the fully-covered pallets.
- **Q2 — no-duplicate:** the same physical SSCC could be transmitted twice — e.g. when a shipment was reactivated and re-completed, or re-sent — because nothing recorded what had already been transmitted, and nothing prevented re-sending it.

## How

- **Per-LU close-gate** in `get_epcis_events_json_fn`: emit each LU as soon as it is fully covered by completed shipments, independent of other still-open LUs on the same shipment. The closer of a shared LU emits it. (Q1)
- **Per-SSCC transmission ledger** (`EDI_EPCIS_Transmitted_SSCC`), written on a confirmed successful send via an `IExternalSystemInvocationSuccessListener`, plus a **ledger-exclusion** in `get_epcis_events_json_fn` so an already-transmitted physical SSCC is never re-emitted. (Q2)
- **Reverse/reactivate/void guard** on the shipment: rejected once the shipment's EPCIS SSCCs were transmitted **or while an export is in-flight** (Enqueued/SendingStarted) — closing the window between dispatch-to-receiver and the async success callback. Deactivating the ledger row (or a stuck in-flight status row) on the shipment's EPCIS tabs is the sanctioned escape-hatch; both tabs are deactivate-capable and expose no hard-delete.

## Tests

- Function-level scenarios for the per-LU close-gate, ledger-exclusion, deactivated-row-ignored, and header-reference scoping (`edi/epcisExport.feature`).
- End-to-end send-path scenarios (`edi/epcisScriptedExportStatus.feature`): mixed standalone+shared shipment (each SSCC transmitted exactly once), re-send produces no duplicate, reverse blocked once transmitted, reverse blocked while in-flight then released by deactivating the status row. Full feature green locally (10/10).

## Migrations

New ledger table + indexes; per-LU close-gate + ledger-exclusion + header-ref scoping in the export function (DDL file kept in sync); read-only diagnostic ledger tab + deactivate-capable status/ledger tabs on the shipment window.


backend/de.metas.edi/src/main/java/de/metas/edi/model/validator/M_InOut.java
```diff
@@ -86,6 +89,38 @@ public void updateEdiExportStatusOnReverse(final I_M_InOut inOut)
...
+			ModelValidator.TIMING_BEFORE_VOID })
+	public void forbidReverseWhenEpcisTransmitted(final I_M_InOut inOut)
+	{
+		final InOutId inOutId = InOutId.ofRepoId(inOut.getM_InOut_ID());
+		if (!epcisReverseGuardService.isEpcisTransmittedOrInFlight(inOutId))
+		{
+			return;
+		}
+
+		throw new AdempiereException("This shipment's EPCIS SSCC events were already transmitted (or are currently being"
```
metas-ts: as always, this needs to be a userfriend and localized AD_Message.  I assume it already is, but if not, then please add this to your corpus (also as review-rule)

backend/de.metas.edi/src/main/sql/postgresql/system/80-de.metas.edi/5813870_sys_EPCIS_StatusTab_deactivatable.sql
```diff
@@ -0,0 +1,57 @@
+-- Make the EPCIS-Exportstatus diagnostic tab (AD_Tab 549295, table
```
metas-ts, line 1: these two SQLs should be in the same filder as the others

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisTransmittedSsccSuccessListener.java
```diff
@@ -0,0 +1,119 @@
...
+	@Override
+	public void onInvocationSuccess(
+			@NonNull final PInstanceId pInstanceId,
+			@NonNull final ExternalSystemInvocationContext context,
+			@NonNull final HttpStatus httpStatus)
+	{
+		final ScriptedExportConversionStatus status = exportStatusService.getLatestByPInstanceId(pInstanceId).orElse(null);
+		if (status == null)
+		{
+			logger.debug("No scripted-export status row found for pInstanceId={} - not a scripted export, skipping", pInstanceId);
```
metas-ts: log all this also to an ILoggable via Loggables.withLogger(...

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisEventsJsonDAO.java
```diff
@@ -0,0 +1,95 @@
...
+	/**
+	 * Returns the SSCC18 value of every pallet in the EPCIS events JSON for the given shipment.
+	 * Returns an empty list when the function yields no pallets (including the {@code '{}'} shape
+	 * returned for a shipment with no EPCIS-relevant events).
+	 */
+	@NonNull
+	public ImmutableList<String> getPalletSscc18s(@NonNull final InOutId inOutId)
+	{
+		final String sql = "SELECT \"de.metas.edi\".get_epcis_events_json_fn(?)::text";
+		final String json = DB.getSQLValueStringEx(ITrx.TRXNAME_None, sql, inOutId.getRepoId());
```
metas-ts: wait, what? we run the whole SQL again, just to thet the SSCCs? i would have expected that we get them fromt he **actual** JSON returned by the actual process invocation

backend/de.metas.edi/src/main/java/de/metas/edi/model/validator/M_InOut.java
metas-ts: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/8005f4aa3d4 — the guard now throws a localized `AD_Message` (`AdMessageKey.of("EPCIS_ReverseBlocked_SSCCTransmitted")`, migration `5813900_sys_AD_Message_EPCIS_ReverseBlocked.sql`, German base + en_US translation, `MsgType=E` + `ErrorCode`) instead of the hardcoded string — it was not yet localized. I also added a corpus Review-rule so code review flags a hardcoded user-facing `AdempiereException("…")` in future (the principle was documented in the AD-message reference but not framed as a greppable review-rule).

backend/de.metas.edi/src/main/sql/postgresql/system/80-de.metas.edi/5813870_sys_EPCIS_StatusTab_deactivatable.sql
metas-ts: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/8005f4aa3d4 — moved both migrations (`5813870_sys_EPCIS_StatusTab_deactivatable.sql`, `5813880_sys_EPCIS_StatusTab_no_hard_delete.sql`) into `backend/de.metas.edi/src/main/sql/postgresql/system/80-de.metas.edi/`, alongside the other 30916 EPCIS migrations.

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisTransmittedSsccSuccessListener.java
metas-ts: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/8005f4aa3d4 — the listener now logs via `Loggables.withLogger(logger, Level.…).addLog(...)` (ILoggable): each skip-reason (no status row / not-the-EPCIS-config / non-M_InOut source / no pallets) and a final `recorded N transmitted SSCC(s)` line now surface on the process / async-workpackage log, not only the slf4j logger.

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisEventsJsonDAO.java
metas-ts: **Resolution (dev):** Confirmed and done — the export process attaches the JSON it sent to the M_InOut (`M_InOut_EPCIS_Export_JSON.afterPostgRESTCall()` → `attachmentEntryService.createNewAttachment(record, "M_InOut_EPCIS_Export_JSON_<id>.json", bytes)`), so the actually-sent payload IS available at callback time. My earlier reply here was wrong — I'd checked the success-callback signature (`pInstanceId` only) and the send-path return value, but missed that the export persists the payload as an M_InOut attachment.

Reworked in https://github.com/metasfresh/metasfresh/commit/4516273691f: `EpcisEventsJsonDAO.getPalletSscc18s` now reads the SSCCs from the shipment's latest `M_InOut_EPCIS_Export_JSON_*` attachment (unwrapping the `{"embedded_json":{…}}` send envelope) instead of re-running the SQL. The `get_epcis_events_json_fn` call remains only as a **defensive fallback** (logged) for the case where that attachment is absent — which shouldn't happen in the scripted-export send path — so the exactly-once ledger is still written. Full feature 10/10 green.

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisEventsJsonDAO.java
```diff
@@ -0,0 +1,116 @@
...
+public class EpcisEventsJsonDAO
+{
+	private final ObjectMapper objectMapper = new ObjectMapper();
+
+	/**
+	 * Recomputes the EPCIS events JSON for the shipment via the SQL function and returns the SSCC18 of
+	 * every pallet. Empty when the function yields no pallets (including the {@code '{}'} shape).
+	 */
+	@NonNull
+	public ImmutableList<String> getPalletSscc18sFromFunction(@NonNull final InOutId inOutId)
```
metas-ts: why do we still need this method?

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisTransmittedSsccSuccessListener.java
```diff
@@ -0,0 +1,168 @@
...
+	private ImmutableList<String> getTransmittedSscc18s(
+			@NonNull final TableRecordReference sourceRecord,
+			@NonNull final InOutId inOutId,
+			@NonNull final ILoggable loggable)
+	{
+		final String epcisFilenamePrefix = M_InOut_EPCIS_Export_JSON.class.getSimpleName();
+		final Optional<AttachmentEntry> sentPayload = attachmentEntryService.getByReferencedRecord(sourceRecord)
+				.stream()
+				.filter(entry -> entry.getFilename() != null && entry.getFilename().startsWith(epcisFilenamePrefix))
+				.findFirst();
```
metas-ts: do the **know** that this returns the most recent EPCIS-Attachment-JSON for the current inout?

backend/de.metas.edi/src/main/java/de/metas/edi/model/validator/M_InOut.java
```diff
@@ -86,6 +92,36 @@ public void updateEdiExportStatusOnReverse(final I_M_InOut inOut)
 		inOut.setEDI_ExportStatus(EDIExportStatus.DontSend.getCode());
 	}
 
+	/**
+	 * Forbids reversing/reactivating/voiding a shipment whose EPCIS SSCC events are already at — or
```
metas-ts: way too much prose here

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/edi/epcisScriptedExportStatus.feature
```diff
@@ -439,3 +439,791 @@ Feature: EPCIS scripted-export status — success, error and re-send flows
     Then after not more than 30s, ExternalSystem_ScriptedExportConversion_Status is found:
       | M_InOut_ID | ExternalSystem_Config_ScriptedExportConversion_ID | ExportStatus |
       | ioB_040    | scriptedCfg_040                                   | U            |
+
+
+  @from:cucumber
```
metas-ts, line 444: please compare epcisScriptedExportStatus.feature and epcisExport.feature
 as list it's differences and where they repeat each other. check if they shouldnt be combinded into one, of maybe epcisScriptedExportStatus should be much more focussed

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisEventsJsonDAO.java
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Extracts the physical SSCC18 values (one per pallet, {@code pallets[].sscc}) of the EPCIS events
+ * payload for a shipment, so a successful EPCIS send can record each transmitted SSCC in the
+ * {@code EDI_EPCIS_Transmitted_SSCC} ledger (see {@link EpcisTransmittedSsccRepository}).
+ *
+ * <p>Two entry points, both pure of any business decision (the "which source" decision lives in the
+ * service-layer caller, {@code EpcisTransmittedSsccSuccessListener}):
+ * <ul>
+ *   <li>{@link #extractPalletSscc18s(String)} — parse an already-obtained EPCIS events JSON string
```
metas-ts: this javadoc is already outdated. make checking that a review rule!

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisTransmittedSsccSuccessListener.java
```diff
@@ -0,0 +1,168 @@
...
+
+		if (sentPayload.isPresent())
+		{
+			final String json = new String(attachmentEntryService.retrieveData(sentPayload.get().getId()), StandardCharsets.UTF_8);
+			return epcisEventsJsonDAO.extractPalletSscc18s(json, inOutId);
+		}
+
+		loggable.addLog("EPCIS ledger: no {} attachment on M_InOut_ID={} - recomputing the EPCIS events JSON via the export function",
+				epcisFilenamePrefix, inOutId.getRepoId());
+		return epcisEventsJsonDAO.getPalletSscc18sFromFunction(inOutId);
```
metas-ts: *why* write a ledger if ther is no attachment? i mean, how would anything be send if there is no attachment ath this point?

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisExportConfigMatcher.java
```diff
@@ -0,0 +1,75 @@
...
+	{
+		final AdProcessId epcisExportProcessId = getProcessIdOrNull();
+		// Fail SAFE, not fail-open: a null EPCIS export process means EPCIS export is not (uniquely)
+		// registered, so NO config can be the EPCIS export — never match a config just because its
+		// own outbound process happens to be null. The null case is logged below for visibility.
+		return epcisExportProcessId != null
+				&& Objects.equals(config.getOutboundDataProcessId(), epcisExportProcessId);
+	}
+
+	private AdProcessId getProcessIdOrNull()
```
metas-ts, line 64: it a method can return null, then annotate with Nullable (from javax.. package). make this a review-rule!

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisExportProcess.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Objects;
+
+/**
+ * Single source of truth for "is this scripted-export config the EPCIS outbound export?" — i.e. does
+ * its outbound-data process resolve to {@link M_InOut_EPCIS_Export_JSON}. Shared by every consumer
+ * (the success listener that writes the transmission ledger, and the reverse guard) so the gating
+ * cannot silently drift between them.
+ */
+@Component
+public class EpcisExportProcess
```
metas-ts: by out convention, a class ending with Process is an JavaProcess (table AD_Process), so this is confusing. if you want to keep the class, find a better name

metas-ts (commented):
this is a partial review
please address all individual comments

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisEventsJsonDAO.java
metas-ts: **Resolution (dev):** Removed in https://github.com/metasfresh/metasfresh/commit/8a0c74a51ad — it was the SQL "fallback" you also flag below. Since the export always attaches its payload on the success path, there is no case where we'd recompute via the function, so the method is gone.

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisTransmittedSsccSuccessListener.java
metas-ts: **Resolution (dev):** Made it explicit in https://github.com/metasfresh/metasfresh/commit/8a0c74a51ad — instead of relying on the query's ordering, the listener now selects the attachment with the highest `AttachmentEntryId` (`.max(Comparator.comparingInt(e -> AttachmentEntryId.getRepoId(e.getId())))`), which is the most recent send for that shipment. (`getByReferencedRecord` does return youngest-first, but the code no longer depends on that guarantee.)

backend/de.metas.edi/src/main/java/de/metas/edi/model/validator/M_InOut.java
metas-ts: **Resolution (dev):** Trimmed to 4 lines in https://github.com/metasfresh/metasfresh/commit/8a0c74a51ad.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/edi/epcisScriptedExportStatus.feature
metas-ts: **Resolution (dev):** Compared both files.

**What each tests — different layers:**
- `epcisExport.feature` — the `get_epcis_events_json_fn` **JSON contract**: every scenario calls the function directly (`the EPCIS JSON export function is called for M_InOut …`) and asserts the produced JSON — pallet/item structure, DESADV/PO references, the per-LU close-gate returning `{}` vs. the covered pallet, ledger-exclusion, deactivated-row-ignored. No send, no status.
- `epcisScriptedExportStatus.feature` — the **send path + status lifecycle**: completion → scripted-export status (`Enqueued`/`Sent`/`Error`/re-send) → `/ok` callback → transmission-ledger write → the reverse / in-flight guard. It asserts status rows, the ledger, and `epcis_has_events` relevance — **not** the JSON structure (0 direct function-JSON assertions).

**Where they repeat:**
- **Setup** — both build the HU→TU→LU→SSCC pallet topologies (intrinsic; both need pallets).
- **Behaviour proven at two layers** — a few Q1/Q2 behaviours are covered both at the function level and end-to-end: mixed standalone+shared (`S30916_010` fn ↔ `S30916_120` send-path), already-transmitted-excluded (`S30916_020` fn ↔ `S30916_130` re-send), shared-LU closer (`S30558_010/020` fn ↔ `S30558_040` status).

**Recommendation:** keep them separate. They exercise genuinely different layers (pure function contract vs. the send/status/ledger/guard integration); merging would put two concerns in one file. The layered coverage of the exactly-once behaviour (function-level *plus* end-to-end) is deliberate for a data-integrity guarantee, and the real duplication is setup, not assertions.

If you'd still like `epcisScriptedExportStatus` more focused, I can trim `S30916_120/130` to their send-specific assertions (the ledger-per-SSCC + guard), leaning on `S30916_010/020` for the close-gate/exclusion coverage — but that drops some end-to-end AC proof, so I'd want your OK first (test-integrity). Which direction do you prefer?

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisEventsJsonDAO.java
metas-ts: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/8a0c74a51ad — the class is now `EpcisEventsJsonParser` with accurate Javadoc; the dangling `{@link #extractPalletSscc18s(String)}` is gone (the method took an `InOutId` param). Also added a corpus Review-rule (java-general.md §38b) so a Javadoc that references a method/class/parameter/signature the current code no longer has is flagged in review.

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisTransmittedSsccSuccessListener.java
metas-ts: **Resolution (dev):** Removed the fallback in https://github.com/metasfresh/metasfresh/commit/8a0c74a51ad — confirmed your point: the export attaches its payload unconditionally, so on the success callback the attachment always exists. A missing attachment therefore means nothing was actually sent, so writing a ledger row would claim a transmission that never happened. The listener now logs a WARN and records nothing in that (impossible-on-success) case.

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisExportConfigMatcher.java
metas-ts: **Resolution (dev):** Added `@Nullable` (`javax.annotation`) to `getProcessIdOrNull` in https://github.com/metasfresh/metasfresh/commit/8a0c74a51ad. Also added a corpus Review-rule (java-general.md §3b): a method or field that can be `null` must carry `@Nullable`.

backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/EpcisExportProcess.java
metas-ts: **Resolution (dev):** Renamed to `EpcisExportConfigMatcher` in https://github.com/metasfresh/metasfresh/commit/8a0c74a51ad — it only answers "is this scripted-export config the EPCIS export?", no AD_Process / JavaProcess involved, so the `…Process` suffix was misleading.

PR #25179, opened 2026-07-16
Shipment Schedule invalidation — dedupe accumulated segments (OOM fix)

teosarca:
## What

Fixes an app-server OOM / GC-death-spiral caused by the shipment-schedule invalidation-segment accumulator under heavy batch load.

**Root cause:** `ShipmentScheduleSegmentChangedProcessor` accumulated invalidation segments in a plain `ArrayList` on the batch transaction, flushed only at `AFTER_COMMIT`. When `CreateMissingShipmentSchedules` runs as one long-lived transaction, thousands of segments — the great majority identical, each holding an enumerated per-warehouse locator `Set` (multiple MB when a warehouse has tens of thousands of locators) — pile up until the heap is exhausted.

Confirmed via JVM heap-dump dominator-tree analysis (Eclipse MAT): a single batch transaction retained ~94% of the heap through this processor's `segments` list, the dominators being thousands of `ImmutableShipmentScheduleSegment` instances holding duplicate enumerated-locator sets.

## The fix — three complementary parts (all in `de.metas.swat.base`)

**A — Dedupe the accumulator.** `segments`: `ArrayList` → `LinkedHashSet`. `ImmutableShipmentScheduleSegment` is a Lombok `@Value` (value equals/hashCode), so equal segments collapse; insertion order is preserved. `addSegment` first normalizes each segment via the existing `ImmutableShipmentScheduleSegment.copyOf(...)`, so segment classes that use identity equals (the HU-derived `ShipmentScheduleSegmentFromHU/-Storage/-Attribute`, which reach the accumulator directly) are also deduped — otherwise the high-churn HU invalidation path would degrade to unbounded list-like growth.

**B — Warehouse identity instead of enumerated locators.** A warehouse-derived segment now carries the warehouse id (`getWarehouseIds()`) instead of a copy of every locator of that warehouse; the recompute WHERE clause matches `COALESCE(M_Warehouse_Override_ID, M_Warehouse_ID) IN (:warehouseIds)` — equivalent to the old `EXISTS(M_Locator …)` for warehouse-derived segments. Single-locator segments stay locator-granular. This shrinks each such segment from MB to a handful of ids.

**C — Threshold flush.** The accumulator flushes mid-batch once it reaches a configurable size (sysconfig `de.metas.inoutcandidate.ShipmentScheduleSegmentFlushThreshold`, default 1000; `<=0` disables), not only at `AFTER_COMMIT` — so a single non-committing batch transaction can no longer accumulate without bound. Safe because the flush's matching SQL runs on its own connection (`TRXNAME_None`) and matches only committed schedules.

## Behavior preserved (AC4)

The exact same set of shipment schedules is flagged for recompute as before — none missed, none spuriously added — regardless of the new flush timing and the warehouse-identity vs enumerated-locators representation.

**One intentional exception (a latent-bug fix, not strict preservation):** for a warehouse that has **zero** locators, the old path enumerated an empty locator set, which made the segment `isInvalid()` and silently matched no schedule — a latent under-invalidation. Matching by the warehouse column now correctly invalidates schedules in such a warehouse. This is safer (a superset — extra recompute is an idempotent no-op), affects only the pathological zero-locator case, and never *misses* an invalidation.

The mid-batch threshold flush is **best-effort**: `process()` flags-then-clears, and a mid-batch flush failure is caught + logged (segments retained for the AFTER_COMMIT retry) so it can never abort the enclosing business transaction — matching the historical post-commit semantics of the AFTER_COMMIT flush.

## Test

TDD throughout. `ShipmentScheduleSegmentChangedProcessorTest`: value-equal segments ×1000 → retained 1; 1000 distinct → retained 1000; identity-equals (HU-shaped) value-equal segments → deduped to 1; mid-batch threshold flush → multiple bounded flushes whose union equals the distinct set. `ImmutableShipmentScheduleSegmentTest`: warehouse segment carries `warehouseIds` and no locators; locator segment unchanged. `ShipmentScheduleInvalidateRepositoryTest`: warehouse-segment WHERE clause matches by warehouse column with no locator sub-select; locator-segment unchanged. Full `de.metas.swat.base` shipment-schedule/invalidation suite green (88 tests).

### DIFF
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/invalidation/impl/ShipmentScheduleSegmentChangedProcessor.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/invalidation/impl/ShipmentScheduleSegmentChangedProcessor.java
@@ -2,7 +2,9 @@ package de.metas.inoutcandidate.invalidation.impl;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Set;
 
 import org.adempiere.ad.trx.api.ITrx;
 import org.adempiere.ad.trx.api.ITrxListenerManager.TrxEventTiming;
@@ -83,7 +85,7 @@ final class ShipmentScheduleSegmentChangedProcessor
 		return processor;
 	}
 
-	private final List<IShipmentScheduleSegment> segments = new ArrayList<>();
+	private final Set<IShipmentScheduleSegment> segments = new LinkedHashSet<>();
 	private final ShipmentScheduleInvalidateBL shipmentScheduleInvalidator;
 
 	private ShipmentScheduleSegmentChangedProcessor(@NonNull final ShipmentScheduleInvalidateBL shipmentScheduleInvalidator)
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/invalidation/impl/ShipmentScheduleSegmentChangedProcessor.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/invalidation/impl/ShipmentScheduleSegmentChangedProcessor.java
@@ -11,9 +11,11 @@ import org.adempiere.ad.trx.api.ITrxListenerManager.TrxEventTiming;
 import org.adempiere.ad.trx.api.ITrxManager;
 import org.adempiere.ad.trx.api.OnTrxMissingPolicy;
 import org.adempiere.service.ISysConfigBL;
+import org.slf4j.Logger;
 
 import de.metas.inoutcandidate.invalidation.segments.IShipmentScheduleSegment;
 import de.metas.inoutcandidate.invalidation.segments.ImmutableShipmentScheduleSegment;
+import de.metas.logging.LogManager;
 import de.metas.util.Services;
 import lombok.NonNull;
 import lombok.ToString;
@@ -43,6 +45,8 @@ import lombok.ToString;
 @ToString(of = "segments")
 final class ShipmentScheduleSegmentChangedProcessor
 {
+	private static final Logger logger = LogManager.getLogger(ShipmentScheduleSegmentChangedProcessor.class);
+
 	private static final String TRX_PROPERTYNAME = ShipmentScheduleSegmentChangedProcessor.class.getName();
 
 	/**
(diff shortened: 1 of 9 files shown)


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/invalidation/impl/ShipmentScheduleSegmentChangedProcessor.java
```diff
@@ -83,12 +99,20 @@ private static ShipmentScheduleSegmentChangedProcessor getOrCreate(
...
+	 * Mid-batch flush threshold, resolved once when this per-trx processor is created (the value is stable for the
+	 * lifetime of the batch, and reading it once bounds the per-{@link #addSegment} cost). A value {@code <= 0}
+	 * disables the mid-batch flush — only the AFTER_COMMIT listener flushes then.
+	 */
+	private final int flushThreshold;
+
 	private ShipmentScheduleSegmentChangedProcessor(@NonNull final ShipmentScheduleInvalidateBL shipmentScheduleInvalidator)
 	{
 		this.shipmentScheduleInvalidator = shipmentScheduleInvalidator;
+		this.flushThreshold = Services.get(ISysConfigBL.class).getIntValue(SYSCONFIG_FlushThreshold, DEFAULT_FlushThreshold);
```
teosarca: sysconfigBL or the flushThreshold shall be params too, like shipmentScheduleInvalidator

teosarca: **Resolution (dev):** Done. `flushThreshold` is now resolved once in the `getOrCreate` factory (via `ISysConfigBL.getIntValue`) and passed to the constructor as an `int` param, so the constructor is a pure value assignment and no longer reaches into the `Services` registry itself — consistent with how `shipmentScheduleInvalidator` is threaded in. https://github.com/metasfresh/metasfresh/commit/b606c9b859a2bfb30a9d7661e5eb24e97c200f3b

teosarca: **Resolution (dev):** Refined per follow-up feedback — the previous change only moved the `Services.get(ISysConfigBL.class)` read from the constructor up into the static factory, which was still a registry call site in the processor. Now the threshold is fully **owned by `ShipmentScheduleInvalidateBL`** (which already owns the invalidation concern and its collaborators): it holds the `ISysConfigBL` field + the `SYSCONFIG_SegmentFlushThreshold` constant and exposes `getSegmentFlushThreshold()`. `ShipmentScheduleSegmentChangedProcessor` obtains the value from `shipmentScheduleInvalidator` (the instance already threaded in) — no `Services`/sysconfig access in the processor at all. The unit test now stubs the threshold on the mock invalidator instead of mutating a SYSTEM sysconfig. https://github.com/metasfresh/metasfresh/commit/aaeb1ba76a18abd8c9528f1e306476e66babab55

PR #25182, opened 2026-07-16
sFTP + REST scripted import: endpoint as single source, window parity, retire EndpointName

metas-ts:
## What

Make the external-systems framework's scripted-import processor usable end-to-end for both **SFTP** and **REST** transports: `ExternalSystem_Endpoint` becomes the single source of a channel's identity + transport config, both config windows reach parity, the legacy free-text `EndpointName` column is retired, and a transport-agnostic Start/Stop control lets a parent config manage import children on different transports.

## Changes

**Config UI + endpoint as single source of truth**
- `getParameters` sources the channel name from the linked endpoint's `Value` (mandatory FK); the config value object drops `endpointName`.
- Application dictionary (19 migrations): bring the parent config window's child tab to parity with the dedicated endpoint window (missing UI wiring for the endpoint FK + SFTP fields); standardise the endpoint FK reference type across import/export tabs; group the external-system windows under a dedicated menu folder; retire the unused `Art`/Type field; rename `OutboundHttpEP` → `HttpEndPoint` (bidirectional label); move the SFTP polling/directory settings from the config onto the endpoint; rename the "Importeur" field to "Import User"; backfill `EndpointName` into an endpoint FK, make the FK mandatory at the DB level, and drop the column; add direction-tagged field descriptions and translation fixes (DE/EN/FR/IT); grid-view column tweaks.

**Per-child Start/Stop control**
- A parent config can have several import children on different endpoints (e.g. one REST, one SFTP), so the user-facing command is now a transport-agnostic Start/Stop *intent*; the concrete enable/disable command is resolved per child from that child's endpoint transport, and a failure on one child no longer aborts the others.
- Registered the missing `ExternalSystem_Service` row for SFTP polling so it can actually be enabled/disabled by the reconciler.

**Runtime fixes**
- Renamed and relabeled the two archive-directory endpoint columns as transport-agnostic LOCAL folders (they are used by both the SFTP and the REST import flow, not SFTP-only as originally implied), and relocated their UI placement out of the SFTP-only field group.
- Fixed an `IsEncrypted` flag drift on the endpoint's private-key column that broke model regeneration for that table.
- SFTP now consumes the remote file by delete (no remote move/rename); the payload is archived to the local processed/error directory for both the SFTP and the REST flow.
- Degrade gracefully instead of throwing when no `PInstanceId` is available to attach an error log to (e.g. a continuous polling consumer).
- Replaced an opaque deserialization error with an actionable message when a script's output doesn't match the expected shape.
- Reset the SFTP polling interval when an endpoint's transport is switched away from SFTP.
- Batch-load endpoints in one query instead of one query per item (N+1) when resolving a collection.

**Tests**
- New Playwright spec covering config creation on both windows (RED before the AD fix, GREEN after).
- Hardened Camel SFTP + REST inbound integration tests: a real JS transform driving OLCand dispatch, and file lifecycle assertions (no longer a no-op transform).
- New unit tests for the endpoint-sourced channel name, the auth-token guard, the error-report route's graceful-degradation path, and the Start/Stop command resolution.

## Verification

- Backend module compiles (Java 8); service- and interceptor-level unit tests green.
- Camel scripted-adapter module test suite green.
- Playwright: config-creation (both windows) + endpoint-window specs green against a customer-faithful stack.


backend/de.metas.externalsystem/src/main/sql/postgresql/system/80-de.metas.externalsystem/5814120_externalsystem_menu_folder.sql
```diff
@@ -0,0 +1,91 @@
+-- #############################################################################
+-- Migration: group the three external-system windows under a dedicated "External
```
metas-ts, line 2: please add *all* external-system-related windows

metas-ts: **Resolution (dev):** Done — extended the menu-folder migration to reparent **all** external-system windows (every `de.metas.externalsystem` window with a menu node, 12 total), not just the three scripted-import/endpoint ones. Added: Externes System Log, Shopware 6, Service + Service Instanz, Leich + Mehl, PCM, Externes System, Skriptbasierte Exportkonvertierung, and PLU-Datei Konfiguration (this last one was parked under the generic "Einstellungen" folder). Verified on a customer-faithful DB: all 12 now sit under the "External System" folder, 0 outside. Node ids absent on a given instance (customer-specific window not installed) are simply not matched by the IN-list — safe no-op. Commit https://github.com/metasfresh/metasfresh/commit/83d93c53fd7

misc/services/camel/de-metas-camel-externalsystems/de-metas-camel-scriptedadapter/src/main/java/de/metas/camel/externalsystems/scriptedadapter/convertmsg/to_mf/ScriptedImportConversionSftpDynamicRouteBuilder.java
```diff
@@ -81,10 +94,36 @@ public void configure()
 							.process(this::handleItemInList)
 						.end()
 					.endChoice()
-				.end();
+				.end()
+				.process(this::archiveLocallyOnSuccess);
 		//@formatter:on
 	}
 
+	private void archiveLocallyOnSuccess(@NonNull final Exchange exchange)
```
metas-ts: there is a pendant for this and the next 2 methods also in ScriptedImport ScriptedImportConversionDynamicRouteBuilder. they look almost the same.\
- can you deduplicate this code? please check if feasible.
- why is such code in here and in ScriptedImportConversionDynamicRouteBuilder, but not in ScriptedImportConversionRestAPIRouteBuilder ?

misc/de-metas-common/de-metas-common-externalsystem/src/main/java/de/metas/common/externalsystem/ExternalSystemConstants.java
```diff
@@ -220,8 +220,11 @@ public class ExternalSystemConstants
 	public static final String PARAM_SFTP_POLLING_ENDPOINT_AUTH_TYPE = "sftpAuthType";
 	public static final String PARAM_SFTP_POLLING_ENDPOINT_REMOTE_PATH = "sftpRemotePath";
 	public static final String PARAM_SFTP_POLLING_INTERVAL_MS = "sftpPollingIntervalMs";
-	public static final String PARAM_SFTP_POLLING_PROCESSED_DIR = "sftpProcessedDirectory";
-	public static final String PARAM_SFTP_POLLING_ERROR_DIR = "sftpErrorDirectory";
+	// Local, transport-agnostic archive folders (used by both SFTP and REST import) -- the
+	// constant names and wire keys keep their historic values; only the endpoint's AD column
+	// names changed (SftpProcessedDirectory/SftpErrorDirectory -> ProcessedDirectory/ErrorDirectory).
+	public static final String PARAM_PROCESSED_DIR = "sftpProcessedDirectory";
```
metas-ts: please remove the "sftp" prefix, also at PARAM_ERROR_DIR

misc/services/camel/de-metas-camel-externalsystems/de-metas-camel-scriptedadapter/src/main/java/de/metas/camel/externalsystems/scriptedadapter/convertmsg/to_mf/ScriptedImportConversionSftpDynamicRouteBuilder.java
metas-ts: **Resolution (dev):** Deduplicated — extracted a shared abstract base `AbstractScriptedImportConversionArchivingRouteBuilder` (commit https://github.com/metasfresh/metasfresh/commit/ab4d625058c) that both `ScriptedImportConversionSftpDynamicRouteBuilder` and `ScriptedImportConversionDynamicRouteBuilder` now extend. It holds the shared fields plus `archiveLocallyOnSuccess`/`OnError`, `archiveLocally`, `handleItemInList`, `getErrorMessage`, `resolveCamelEndpointUri`, and the nested `ResponseAggregationStrategy`. Each subclass keeps only its own `configure()` (genuinely different: `sftp:` vs `direct:` source, and SFTP's `onException(...).handled(true)`) plus the one real difference, exposed as `protected abstract String archiveFileName(Exchange)` — SFTP preserves the real remote `Exchange.FILE_NAME` (fallback `endpointName_timestamp`), the on-demand route synthesizes `timestamp_endpointName.json`. Scriptedadapter module suite green (62 tests).

On why the archiving/dispatch code is here and in `ScriptedImportConversionDynamicRouteBuilder` but not in `ScriptedImportConversionRestAPIRouteBuilder`: those two are **asynchronous per-endpoint** routes — they consume a payload (a downloaded SFTP file / an on-demand `direct:` body), transform → dispatch, then archive the consumed payload to a local processed/error folder for audit/retry. `ScriptedImportConversionRestAPIRouteBuilder` is the **synchronous REST** endpoint: the caller POSTs and gets an immediate HTTP status (`OK` / `MULTI_STATUS` / `500`) back as the acknowledgement, so there is no consumed file to move aside and no local-archiving step to perform there.

misc/de-metas-common/de-metas-common-externalsystem/src/main/java/de/metas/common/externalsystem/ExternalSystemConstants.java
metas-ts: **Resolution (dev):** Done — dropped the `sftp` prefix (commit https://github.com/metasfresh/metasfresh/commit/1037292bc56): `PARAM_PROCESSED_DIR` / `PARAM_ERROR_DIR` values are now `processedDirectory` / `errorDirectory`. Since these are wire keys, I verified producer and consumer stay matched — the producer (`ExternalSystemScriptedImportConversionService.getParameters`) and every camel consumer/test already reference the constants, never the literal, so `grep -rn 'sftpProcessedDirectory\|sftpErrorDirectory'` over the worktree is now 0 hits. I also updated a stale test fixture (`1_SftpExternalSystemRequest.json`) that carried the old keys. Both affected module suites green (de.metas.externalsystem 126, scriptedadapter 62).

PR #25188, opened 2026-07-16
externalsystem: one status row per export attempt

metas-ts:
## What

The scripted-export status table (`ExternalSystem_ScriptedExportConversion_Status`) shipped with a UNIQUE index on `(config, AD_Table_ID, Record_ID)`, forcing a single in-place-upserted row per source record. The intended design is **one row per export ATTEMPT**: each enqueue / re-send is its own row, so the status tab shows the attempt log newest-first and a re-send never overwrites the prior attempt's outcome.

## Changes

- **migration `5814350`** — `DROP` the single-row unique index so attempt rows coexist.
- **repo** — `upsert()` → `insertNewAttempt()`: always inserts a fresh row; lifecycle transitions update the row via `update()` / `updateLatestByPInstanceId()`.
- **repo `update()`** — targets the **pInstance-bound** attempt when the status carries a pInstance (every outcome transition does), falling back to the latest attempt-by-key only for the pre-enqueue Pending→Enqueued binding. Without this, a late/duplicate callback for a superseded attempt would clobber a newer sibling attempt of the same `(config, sourceRecord)`.
- **service** — `recordDontSend` / `recordPending` / `recordPendingAsResend` insert new attempts.

## Tests

- **repo** — per-attempt row-count assertion + a RED→GREEN guard that a transition for one attempt's pInstance updates *its own* row, never the newest sibling.
- **cucumber `S30088_030`** — asserts **two rows coexist** after a re-send (errored first attempt + Sent re-send) — the observable that distinguishes per-attempt history from the former single-row upsert (would be 1 under the old behavior).

Backend `de.metas.externalsystem` unit tests: 25 green (10 repo + 15 service), JDK8, run locally.

### DIFF
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/edi/epcisScriptedExportStatus.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/edi/epcisScriptedExportStatus.feature
@@ -6,7 +6,7 @@ Feature: EPCIS scripted-export status — success, error and re-send flows
 ## Verifies the ExternalSystem_ScriptedExportConversion_Status lifecycle for the EPCIS export config:
 ## (a) shipment completed → invocation enqueued → /ok callback → status row Sent + roll-up M_InOut.EPCIS_ExportStatus=Sent
 ## (b) same path but /error callback → status row Error + AD_Issue_ID linked + roll-up Error
-## (c) re-send of an errored shipment → status row flipped to Pending+IsResend=Y (single-row upsert) → Sent via /ok
+## (c) re-send of an errored shipment → a NEW Pending+IsResend=Y attempt row (per-attempt history, prior attempt kept) → Sent via /ok
 
   Background:
     Given infrastructure and metasfresh are running
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/edi/epcisScriptedExportStatus.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/edi/epcisScriptedExportStatus.feature
@@ -160,7 +160,7 @@ Feature: EPCIS scripted-export status — success, error and re-send flows
   @allure.label.epic:E0292_EDI
   @allure.label.feature:F00353_EDI_DESADV_InOut_Link
   @Id:S30088_030
-  Scenario: S30088_030 — re-send of an errored shipment flips the single status row to Pending+IsResend=Y
+  Scenario: S30088_030 — re-send of an errored shipment adds a NEW attempt row (per-attempt history), keeping the errored one
 
     And metasfresh contains C_Orders:
       | Identifier | IsSOTrx | C_BPartner_ID | DateOrdered | POReference                |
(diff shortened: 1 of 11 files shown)


backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/edi/epcisScriptedExportStatus.feature
```diff
@@ -213,6 +214,10 @@ Feature: EPCIS scripted-export status — success, error and re-send flows
       | M_InOut_ID | EPCIS_ExportStatus |
       | io_030     | S                  |
 
+    # Per-attempt history: TWO rows now coexist for this shipment+config — the errored first attempt
+    # AND the Sent re-send attempt. (Under the former single-row upsert this count would be 1.)
+    And after not more than 10s, ExternalSystem_ScriptedExportConversion_Status row count for shipment io_030 and config scriptedCfg_es is 2
```
metas-ts: verify the actual data of the two rows. i mean the data that's displayed in the grid-view

metas-ts: Done in 492a3cfdf1b — replaced the count-only check with a data-table assertion that verifies both attempt rows' **actual data, newest-first** (exactly as the status tab's grid orders them):

| ExportStatus | IsResend | HttpResponseCode | HasAD_Issue |
|---|---|---|---|
| S | Y | 200 | N |
| E | N |  | Y |

So it now asserts the successful re-send row (Sent, IsResend=Y, HTTP 200, no issue) sits on top and the errored first attempt (Error, IsResend=N, its AD_Issue retained) beneath it. The row count is implied by the exact row-for-row match.

metas-ts: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/492a3cfdf1b — replaced the count-only check with a data-table assertion that verifies both attempt rows' actual data, newest-first (as the status tab's grid orders them):

| ExportStatus | IsResend | HttpResponseCode | HasAD_Issue |
|---|---|---|---|
| S | Y | 200 | N |
| E | N |  | Y |

It asserts the successful re-send row (Sent, IsResend=Y, HTTP 200, no issue) on top and the errored first attempt (Error, IsResend=N, its AD_Issue retained) beneath it. Row count is implied by the exact row-for-row match.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/edi/epcisScriptedExportStatus.feature
```diff
@@ -213,6 +214,15 @@ Feature: EPCIS scripted-export status — success, error and re-send flows
       | M_InOut_ID | EPCIS_ExportStatus |
       | io_030     | S                  |
 
+    # Per-attempt history: the status tab's grid now shows TWO attempt rows (newest-first), each with
+    # its own data — the successful re-send on top (Sent, IsResend=Y, HTTP 200, no issue) and the
+    # errored first attempt beneath it (Error, IsResend=N, its AD_Issue retained). (Under the former
+    # single-row upsert there would be a single row.)
+    And after not more than 10s, ExternalSystem_ScriptedExportConversion_Status rows for shipment io_030 and config scriptedCfg_es are (newest first):
```
metas-ts: consolidate this step with "after not more than 10s, ExternalSystem_ScriptedExportConversion_Status is found:" and deduplicate the two into one.

metas-ts: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/f6a943aede6 — agreed, that second step was a needless duplicate. I generalized the existing `...Status is found:` step to accept N data-table rows instead of one (each row matched newest-first by its `ExportStatus`, with optional `IsResend` / `HttpResponseCode` / `HasAD_Issue`) and deleted the separate step. Single-row usages keep the prior "latest attempt" behaviour; the S30088_030 re-send check now uses the same one step with two rows:

| M_InOut_ID | ExternalSystem_Config_ScriptedExportConversion_ID | ExportStatus | IsResend | HttpResponseCode | HasAD_Issue |
|---|---|---|---|---|---|
| io_030 | scriptedCfg_es | S | Y | 200 | N |
| io_030 | scriptedCfg_es | E | N |  | Y |

PR #25194, opened 2026-07-16
CreateMissingShipmentSchedules — batch missing-schedule creation via bounded re-enqueued work packages

teosarca:
## Problem

`CreateMissingShipmentSchedulesWorkpackageProcessor` created ALL missing shipment schedules in a single unbounded transaction, and its handler retrieve materialized a guaranteed-iterator selection over the ENTIRE missing-order-line backlog. On a large backlog this holds one long-running transaction and exhausts heap (`OutOfMemoryError`), stalling the async work-package queue.

## Change

Process missing order lines in **bounded, re-enqueued batches** (mirrors `UpdateInvalidInvoiceCandidatesWorkpackageProcessor`):

- Processor runs **out of transaction** (`isRunInTransaction() = false`).
- Each run processes at most `MaxToProcess` models inside one short batch run via `trxManager.callInThreadInheritedTrx(() -> processOneBatch(...))` (starts + commits/closes a new trx when none is inherited). Batch size via sysconfig `de.metas.inoutcandidate.async.CreateMissingShipmentSchedulesWorkpackageProcessor.MaxToProcess` (code default 500).
- When work remains, it enqueues a follow-up work package (carrying the `AsyncBatchId`); draining the chain creates all missing schedules.
- `IShipmentScheduleHandlerBL.createMissingCandidates(ctx, QueryLimit)` returns `CreateMissingCandidatesResult { createdShipmentScheduleIds, limitReached }`. The no-arg overload is now `@Deprecated` (delegates unlimited — unchanged behaviour; the unbounded path can OOM on a large backlog).
- **The batch limit is pushed down into the retrieve**: `ShipmentScheduleHandler#retrieveModelsWithMissingCandidates(ctx, trxName, QueryLimit)` now applies `.setLimit(limit)` (`OrderLineShipmentScheduleHandler`, `SubscriptionShipmentScheduleHandler`), so each batch materializes at most `limit` rows instead of the whole backlog. ORDER BY is preserved → deterministic first-N → forward progress.

## Result

- No single transaction spans the whole backlog, and the retrieve no longer buffers the entire backlog → no OOM, no multi-hour transaction.
- `getByIds` runs on a bounded (≤ limit) id set — no unbounded IN-clause.
- End result identical to the unbounded run: same schedules created, none missed or duplicated. `limitReached` uses budget-exhaustion; an exact-multiple-of-batch backlog causes at most one extra empty follow-up run, which terminates.

## Tests

- New cucumber scenario `createMissingShipmentSchedulesBatching.feature`: with `MaxToProcess = 2` and a 5-line sales order, one processor run creates exactly 2 schedules and re-enqueues a follow-up WP; draining creates all 5 with no pending WP left. A deterministic single-work-package-run step drives the processor directly to avoid background-drain flakiness.
- Unit tests for `createMissingCandidates(ctx, QueryLimit)` / the `Budget` cross-handler logic (bounded path, exact-multiple edge, NO_LIMIT, cross-handler threading).
- Existing `CreateMissingShipmentSchedulesWorkpackageProcessorTest` unchanged and green.

## Notes

- A manual, admin-triggered path (`M_ShipmentSchedule_Update` → `ShipmentScheduleUpdater`) still calls the deprecated unbounded overload; it is not the automated feed that caused the incident and is documented for a possible follow-up.

## Known limitation

A single order line that throws during creation rolls back its **entire batch transaction** (every creation in that run, up to `MaxToProcess`, is undone) and the work package errors — halting the re-enqueue chain at that batch until the offending line is corrected. No infinite loop; blast radius is one batch, and the drain resumes once the poison line is fixed. Still strictly better than the previous whole-backlog single-transaction failure. To be addressed separately.

### DIFF
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/IShipmentScheduleHandlerBL.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/IShipmentScheduleHandlerBL.java
@@ -6,6 +6,7 @@ import de.metas.inoutcandidate.model.I_M_ShipmentSchedule;
 import de.metas.inoutcandidate.spi.ModelWithoutShipmentScheduleVetoer;
 import de.metas.inoutcandidate.spi.ShipmentScheduleHandler;
 import de.metas.util.ISingletonService;
+import org.adempiere.ad.dao.QueryLimit;
 import org.compiere.model.I_C_OrderLine;
 
 import java.util.Properties;
@@ -51,8 +52,19 @@ public interface IShipmentScheduleHandlerBL extends ISingletonService
 
 	/**
 	 * Invokes all registered {@link ShipmentScheduleHandler}s to create missing InOut candidates.
+	 * <p>
+	 * Unlimited variant of {@link #createMissingCandidates(Properties, QueryLimit)}; delegates with {@link QueryLimit#NO_LIMIT}.
 	 */
 	Set<ShipmentScheduleId> createMissingCandidates(Properties ctx);
+
+	/**
+	 * Invokes all registered {@link ShipmentScheduleHandler}s to create missing InOut candidates, processing at most
+	 * {@code maxToProcess} models (created-or-vetoed, not schedules created) across all handlers combined.
+	 *
+	 * @param maxToProcess budget of models to process; use {@link QueryLimit#NO_LIMIT} for unlimited.
+	 * @return the created shipment schedule ids, plus whether the budget was exhausted with more work remaining.
+	 */
+	CreateMissingCandidatesResult createMissingCandidates(Properties ctx, QueryLimit maxToProcess);
 
 	/**
 	 * Invokes the given <code>sched</code>'s {@link ShipmentScheduleHandler} to get a {@link IDeliverRequest} instance.
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/IShipmentScheduleHandlerBL.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/IShipmentScheduleHandlerBL.java
@@ -54,7 +54,11 @@ public interface IShipmentScheduleHandlerBL extends ISingletonService
 	 * Invokes all registered {@link ShipmentScheduleHandler}s to create missing InOut candidates.
 	 * <p>
 	 * Unlimited variant of {@link #createMissingCandidates(Properties, QueryLimit)}; delegates with {@link QueryLimit#NO_LIMIT}.
+	 *
+	 * @deprecated unbounded: processes the whole backlog in one go and can OOM on a large backlog. Use the bounded
+	 * {@link #createMissingCandidates(Properties, QueryLimit)} overload instead.
 	 */
+	@Deprecated
 	Set<ShipmentScheduleId> createMissingCandidates(Properties ctx);
 
 	/**
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/async/CreateMissingShipmentSchedulesWorkpackageProcessor.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/async/CreateMissingShipmentSchedulesWorkpackageProcessor.java
@@ -31,7 +31,7 @@ import de.metas.async.api.IWorkPackageQueue;
 import de.metas.async.model.I_C_Queue_WorkPackage;
 import de.metas.async.processor.IWorkPackageQueueFactory;
 import de.metas.async.spi.WorkpackageProcessorAdapter;
-import de.metas.inout.ShipmentScheduleId;
+import de.metas.inoutcandidate.api.CreateMissingCandidatesResult;
 import de.metas.inoutcandidate.api.IShipmentScheduleBL;
 import de.metas.inoutcandidate.api.IShipmentScheduleHandlerBL;
 import de.metas.inoutcandidate.api.IShipmentSchedulePA;
@@ -42,14 +42,17 @@ import de.metas.util.ILoggable;
 import de.metas.util.Loggables;
 import de.metas.util.Services;
 import lombok.NonNull;
+import org.adempiere.ad.dao.QueryLimit;
+import org.adempiere.ad.trx.api.ITrxManager;
 import org.adempiere.model.InterfaceWrapperHelper;
+import org.adempiere.service.ISysConfigBL;
 import org.adempiere.util.lang.IContextAware;
 import org.slf4j.Logger;
 
 import javax.annotation.Nullable;
 import java.util.Collection;
 import java.util.Properties;
-import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Workpackage used to create missing shipment schedules.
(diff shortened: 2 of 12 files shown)


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/async/CreateMissingShipmentSchedulesWorkpackageProcessor.java
```diff
@@ -125,26 +131,76 @@ private static boolean _scheduleIfNotPostponed(final IContextAware ctxAware, @Nu
...
-		for (final I_M_ShipmentSchedule scheduleRecord : scheduleRecords)
+		final Properties ctx = InterfaceWrapperHelper.getCtx(workpackage);
+		final QueryLimit maxToProcess = QueryLimit.ofInt(getMaxToProcess());
+
+		// Create+invalidate one bounded batch of missing shipment schedules in its own transaction. Justification for
+		// runInNewTrx (see the "runInNewTrx is a hack" gotcha): we deliberately want a SHORT, bounded transaction here
+		// instead of one unbounded transaction for the whole backlog (which OOMs on a large backlog) -- that is the
+		// whole point of this processor's batching.
+		final AtomicReference<CreateMissingCandidatesResult> resultHolder = new AtomicReference<>();
+		trxManager.runInNewTrx(() -> {
```
teosarca: always use runInThreadInheritedTrx instead of runInNewTrx.
If there is no outer trx, runInThreadInheritedTrx starts a new trx and commits/closes it.

teosarca: in this case use callInThreadInheritedTrx and refactor the lambda as a new method.
so might be that resultHolder  won;t be needed either

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/async/CreateMissingShipmentSchedulesWorkpackageProcessor.java
```diff
@@ -125,26 +131,76 @@ private static boolean _scheduleIfNotPostponed(final IContextAware ctxAware, @Nu
...
+			final CreateMissingCandidatesResult result = inOutCandHandlerBL.createMissingCandidates(ctx, maxToProcess);
+			resultHolder.set(result);
+
+			// After shipment schedules where created, invalidate them because we want to make sure they are up2date.
+			final IShipmentScheduleInvalidateBL invalidSchedulesService = Services.get(IShipmentScheduleInvalidateBL.class);
```
teosarca: services, declare them as a class field

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/async/CreateMissingShipmentSchedulesWorkpackageProcessor.java
```diff
@@ -125,26 +131,76 @@ private static boolean _scheduleIfNotPostponed(final IContextAware ctxAware, @Nu
...
+	 * Enqueues a fresh workpackage (carrying over the current one's async batch) to create the shipment schedules that
+	 * remained after this run's bounded batch. Enqueues it directly instead of going through {@link #scheduleIfNotPostponed},
+	 * because that method's dedup guard (skip if &gt;1 processable workpackage already queued) does not apply here: we
+	 * KNOW there is more work left for this exact run and must not skip it.
+	 */
+	private void enqueueFollowUpWorkpackage(@NonNull final Properties ctx, @NonNull final I_C_Queue_WorkPackage workpackage)
+	{
+		final AsyncBatchId asyncBatchId = AsyncBatchId.ofRepoIdOrNull(workpackage.getC_Async_Batch_ID());
+
+		final IWorkPackageQueueFactory workPackageQueueFactory = Services.get(IWorkPackageQueueFactory.class);
```
teosarca: declare it as a class field

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/IShipmentScheduleHandlerBL.java
```diff
@@ -51,9 +52,20 @@ public interface IShipmentScheduleHandlerBL extends ISingletonService
 
 	/**
 	 * Invokes all registered {@link ShipmentScheduleHandler}s to create missing InOut candidates.
+	 * <p>
+	 * Unlimited variant of {@link #createMissingCandidates(Properties, QueryLimit)}; delegates with {@link QueryLimit#NO_LIMIT}.
 	 */
 	Set<ShipmentScheduleId> createMissingCandidates(Properties ctx);
```
teosarca, line 62: annotate it as deprecated.
also check/document on issue page because callers of this method might have problems

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/async/CreateMissingShipmentSchedulesWorkpackageProcessor.java
teosarca: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/88f1f2f0a24 — replaced `callInNewTrx` with `callInThreadInheritedTrx` (starts + commits/closes a new trx when none is inherited, i.e. the out-of-trx case here), and extracted the batch body into a private `processOneBatch(...)` invoked via method reference, so the `resultHolder`/AtomicReference is gone.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/async/CreateMissingShipmentSchedulesWorkpackageProcessor.java
teosarca: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/88f1f2f0a24 — the instance-scope services are `private final transient` class fields; the local `IWorkPackageQueueFactory` re-declaration in the follow-up-enqueue path was removed so it uses the field. (The one remaining local is in the `private static _scheduleIfNotPostponed` helper, which has no access to an instance field.)

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/async/CreateMissingShipmentSchedulesWorkpackageProcessor.java
teosarca: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/88f1f2f0a24 — the `IWorkPackageQueueFactory` used by the re-enqueue path now uses the class field (its local `Services.get` re-declaration removed).

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/IShipmentScheduleHandlerBL.java
teosarca: **Resolution (dev):** Done — no-arg `createMissingCandidates(Properties)` annotated `@Deprecated` (interface + impl), pointing callers to the bounded overload (https://github.com/metasfresh/metasfresh/commit/88f1f2f0a24). And per your other point, the batch limit is now pushed down into `retrieveModelsWithMissingCandidates(..., QueryLimit)` (https://github.com/metasfresh/metasfresh/commit/766a3f296cc), so `OrderLineShipmentScheduleHandler` does `.setLimit(limit)` and the guaranteed-iterator selection is bounded instead of materializing the whole backlog. Caller audit + OOM-risk assessment will be posted on the issue.

PR #25226, opened 2026-07-18
mobile-webui: fix barcode scanner stuck in default mode on late settings load (flaky case 19)

teosarca:
## What

Fixes flaky test **case 19** (flaky-test registry) — `barcode_scanner_modes.spec.js` "manual mode — visible editable input rendered…" intermittently timing out on `expect(manual-entry-input).toBeAttached()`.

## Root cause (production defect, not a test artifact)

`BarcodeScannerComponent` initialised `activeMode` from `defaultMode` via `useState` only. `defaultMode` is derived from backend settings, which `ApplicationRoot` fetches **fire-and-forget** (`getSettings().then(putSettingsAction)`) after login — with no ordering guarantee vs. when a scanner screen mounts the component.

When the component mounts **before** settings resolve, `useBarcodeScannerModes` returns hook defaults (`defaultMode=HARDWARE`) and the `useState` initializer freezes `activeMode` at HARDWARE. When settings later arrive (`defaultMode=manual`), `activeMode` never re-syncs -> `ManualModePanel` never mounts -> `manual-entry-input` never attaches -> `SLOW_ACTION_TIMEOUT`. The lucky ordering (settings resolve first) passes, which is why it "clears on re-run".

This is also a latent real-world defect: on a slow network an operator can be stuck in the wrong scanner mode until reload.

## Fix

Adopt `defaultMode` **once**, when settings first load — extracted into a dedicated hook `useSettingsDefaultModeAdoption({ defaultMode, isProcessing })` (defined in the same file, just below the component). It:
- adopts the settings-derived default only **once** (a one-shot latch, keyed off a new `useIsSettingsLoaded()` selector), then never again;
- **never overrides a mode the operator already chose** (an operator-interaction ref, set only by explicit mode selection);
- **defers adoption while a scan is processing** (`!isProcessing`) so a late settings arrival can never flip `activeMode` mid-scan (which would flicker the panel) — the pending adoption applies on the first calm render instead.

## Tests

- Deterministic component tests in `BarcodeScannerComponent.modeSettingsRace.test.js` covering: adoption after mount-before-settings, operator choice not overridden by a settings re-emit, and (new) adoption deferred while a scan is processing — each RED→GREEN.
- Full mobile-webui frontend unit suite: 254 passed / 23 suites. Lint clean.

Verified (independent review, HEAD https://github.com/metasfresh/metasfresh/commit/e8fdfd938a8b003277ceb6c6b983ac26965b0567): reverting only the `!isProcessing` guard and running `react-scripts test --watchAll=false BarcodeScannerComponent.modeSettingsRace.test.js` fails exactly the "defers adoption while a scan is processing" case (1 failed / 3 passed); restoring it -> 4 passed. This runtime RED→GREEN confirms the guard behaviour, not merely the presence of the code.

Base: `new_dawn_uat`. Down-ports to the hotfix/release branch groups noted as future low-priority; NOT propagated here.

### DIFF
--- a/misc/services/mobile-webui/mobile-webui-frontend/src/components/BarcodeScannerComponent.jsx
+++ b/misc/services/mobile-webui/mobile-webui-frontend/src/components/BarcodeScannerComponent.jsx
@@ -24,47 +24,10 @@ const BarcodeScannerComponent = ({
     invisible,
   });
 
-  const [activeMode, setActiveMode] = useState(defaultMode);
-
-  // Set true the moment the operator explicitly picks a mode (footer HW/CAM toggle, footer
-  // "enter manually", or the ManualModePanel "back to scanner" button). Read by the async-settings
-  // adoption below so a deliberate operator choice made during the settings-load window is never
-  // overridden. NOT set by the internal auto-returns to the default mode (post-scan / camera
-  // cancel) — those are not operator mode selections.
-  const hasOperatorSelectedModeRef = useRef(false);
-  const selectMode = useCallback((mode) => {
-    hasOperatorSelectedModeRef.current = true;
-    setActiveMode(mode);
-  }, []);
-
-  // ── Async-settings default adoption ─────────────────────────────────────────────────────────
-  // Settings load asynchronously (ApplicationRoot fetches them fire-and-forget after login). If
-  // this scanner mounts BEFORE they arrive, useBarcodeScannerModes returns its hook defaults so
-  // defaultMode resolves to HARDWARE, and the useState initializer above freezes activeMode there
-  // — it would NOT pick up the configured default once settings resolve (e.g. defaultMode=manual),
-  // leaving the operator stuck in hardware mode with no visible manual input (flaky e2e case:
-  // barcode_scanner_modes.spec.js "manual mode — visible editable input rendered…").
-  //
-  // Adjust the state DURING RENDER (React's documented "adjust state when an input changes"
-  // pattern) rather than in a useEffect, so the switch happens before paint — on a handheld this
-  // avoids a one-frame flash of the wrong-mode scanner UI. Guarded so it fires exactly once, on
-  // the not-loaded→loaded transition.
-  //
-  // Adopt the configured default ONLY if the operator has NOT explicitly picked a mode during the
-  // load window (the footer's "enter manually" / toggle buttons are live before settings load) —
-  // otherwise a late settings arrival would silently revert the operator's own selection.
-  const isSettingsLoaded = useIsSettingsLoaded();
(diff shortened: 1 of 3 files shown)


misc/services/mobile-webui/mobile-webui-frontend/src/components/BarcodeScannerComponent.jsx
```diff
@@ -26,6 +26,43 @@ const BarcodeScannerComponent = ({
 
   const [activeMode, setActiveMode] = useState(defaultMode);
 
+  // Set true the moment the operator explicitly picks a mode (footer HW/CAM toggle, footer
+  // "enter manually", or the ManualModePanel "back to scanner" button). Read by the async-settings
+  // adoption below so a deliberate operator choice made during the settings-load window is never
+  // overridden. NOT set by the internal auto-returns to the default mode (post-scan / camera
+  // cancel) — those are not operator mode selections.
+  const hasOperatorSelectedModeRef = useRef(false);
```
teosarca: pls refactor the whole code chunk u introduced in a new hook, define it in the same file, below BarcodeScannerComponent.
give it an intention-revealing and descriptive but short name.
pls keep the comments concise, precise and simple! (closer to eli5 but not idiotic)

teosarca: **Resolution (dev):** Acting on all three points — (1) refactoring the introduced chunk into a new intention-revealing hook defined in this same file, below `BarcodeScannerComponent`; (2) tightening the comments to concise/precise/simple; (3) also adding the `!isProcessing` guard so the settings-default adoption never flips the mode mid-scan. Implementation is in progress (with a RED→GREEN test for the guard + full unit suite + lint); I will follow up on this thread with the commit URL and test results once pushed.

teosarca: **Resolution (dev):** Done. Extracted the whole introduced chunk into a new hook `useSettingsDefaultModeAdoption({ defaultMode, isProcessing }) → { activeMode, setActiveMode, selectMode }`, defined in this same file just below `BarcodeScannerComponent`; the component body is now a single call to it. Rewrote the comments tight and plain — a short WHY (mount-before-settings race) plus a 3-bullet guard list, no wall-of-text. Also added the `!isProcessing` guard so a late settings arrival never flips `activeMode` mid-scan (the one-shot latch stays unset while a scan is processing and adoption defers to the first calm render). New RED→GREEN test in the modeSettingsRace suite covers it; full mobile-webui suite green (254/254), lint clean. Commit: https://github.com/metasfresh/metasfresh/pull/25226/commits/e8fdfd938a8b003277ceb6c6b983ac26965b0567

PR #25229, opened 2026-07-19
EPCIS: don't send an empty event on re-send when nothing is new

metas-ts:
## What

A manual **Re-send** (`M_InOut_ReSend_ScriptedExportConversion`) invoked the scripted-export conversion **unconditionally**, unlike the auto-complete path which gates on the config `WhereClause`. So re-sending a shipment whose SSCCs are all already in the `EDI_EPCIS_Transmitted_SSCC` ledger fired an **empty EPCIS event**. This makes the re-send send **nothing** when there is nothing new.

## Change

- **Resend gate (the fix)** — `M_InOut_ReSend_ScriptedExportConversion` (thin glue) delegates per config to `ExternalSystemScriptedExportConversionService.resendConfigIfRelevant`, which resolves the config (fail-fast) and checks `isConfigMatchingRecord` (the config's `WhereClause` = `epcis_has_events(m_inout_id)` for EPCIS) before invoking. Nothing new → records `DontSend`, does **not** invoke the adapter; matching → records `Pending(IsResend)` + invokes. Mirrors the auto path's `recordEligibilityAndInvoke`.

_(An earlier SQL "return `{}` on empty pallets" defense-in-depth layer was dropped: it regressed the `get_epcis_events_json_fn` header-envelope contract — unpicked shipments legitimately return a full header envelope with `pallets: []` (validated by EPCIS_010/020/030) — and it is redundant, since the resend gate already guarantees the adapter/JS is never invoked with nothing new.)_

## Tests

- **Unit (RED→GREEN):** `ExternalSystemScriptedExportConversionServiceResendTest` — not-matching → `DontSend`, invoke never called, returns false; matching → `Pending`+invoke, returns true. Process test is pure delegation.
- **Cucumber `S30916_150`:** complete → Enqueued → `/error` (resendable) → seed the SSCC in the ledger (`epcis_has_events` flips false) → run Re-send → asserts the status is **DontSend (`N`)** and the ledger stays at exactly one row (no send, no duplicate). Green in CI profile7.

`de.metas.externalsystem` module: 124 tests green.


backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/scriptedexportconversion/process/M_InOut_ReSend_ScriptedExportConversion.java
```diff
@@ -67,15 +65,12 @@ protected String doIt()
...
-
-			scriptedExportService.executeInvokeScriptedExportConversionActionAndGetResult(
-					config,
-					m_inout_id,
-					ExternalSystemInvocationContext.RESEND);
-
-			triggered++;
+			// The service gates the re-send on the config's WhereClause: nothing left to export
+			// (e.g. every SSCC already in the EPCIS ledger) → DontSend, no adapter invocation.
+			if (scriptedExportService.resendConfigIfRelevant(configId, sourceRecord, m_inout_id))
```
metas-ts, line 70: use ILoggable (Loggables.addMsg(....) ) to log to AD_PInstance on whether each sourceRecord was send or skipped

backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/scriptedexportconversion/ExternalSystemScriptedExportConversionService.java
```diff
@@ -291,28 +291,40 @@ public List<Exception> executeInvokeScriptedExportConversionAction(
...
+	 * status-row write, so a bad config throws before a Pending/DontSend row is created (no orphan row).
 	 *
-	 * @return the resolved config, ready for the follow-up
-	 *         {@link #executeInvokeScriptedExportConversionActionAndGetResult} call
+	 * @return {@code true} if the conversion was invoked (something to send), {@code false} if suppressed
+	 *         because nothing was left to export.
 	 */
-	@NonNull
-	public ExternalSystemScriptedExportConversionConfig resolveConfigAndRecordPendingAsResend(
+	public boolean resendConfigIfRelevant(
```
metas-ts, line 309: at one point the code skips over source-reords if they have no error-state. please make it instead also consider that have the ste "shall not be send"

backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/scriptedexportconversion/process/M_InOut_ReSend_ScriptedExportConversion.java
metas-ts: **Resolution (dev):** Adopting. Adding ambient `Loggables.addLog(...)` in `resendConfigIfRelevant` so each config + source-record logs to the AD_PInstance process log whether it was **invoked (sent)** or **skipped (nothing to export / WhereClause no longer matches)**. Implementing now — will confirm with the commit URL.

backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/scriptedexportconversion/ExternalSystemScriptedExportConversionService.java
metas-ts: **Resolution (dev):** Adopting. Extending `getResendableConfigsBySourceRecord` to also include the **DontSend** ("shall not be sent") state — via a new `ExternalSystemExportStatus.isResendable()` (Error | Invalid | DontSend) — so a suppressed record can be re-evaluated on re-send. It composes idempotently with the gate: a re-send re-checks `epcis_has_events`; still nothing new → DontSend again, now-something → sends. Adding a covering test. Implementing now — will confirm with the commit URL.

backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/scriptedexportconversion/process/M_InOut_ReSend_ScriptedExportConversion.java
metas-ts: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/e85956c39d7 — `resendConfigIfRelevant` now `Loggables.addLog`s per config+record: `→ invoked` (sent) or `→ skipped (nothing to export / WhereClause no longer matches)`, visible in the AD_PInstance process log.

backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/scriptedexportconversion/ExternalSystemScriptedExportConversionService.java
metas-ts: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/e85956c39d7 — `getResendableConfigsBySourceRecord` now selects `Error | Invalid | DontSend` via new `ExternalSystemExportStatus.isResendable()`; a DontSend record is re-evaluated on re-send (idempotent with the gate). Covering test `getResendableConfigs_includes_dontSendConfig` (RED against the old filter). `computeRollUp` keeps `isErrorOrInvalid` (roll-up, unchanged).

PR #25247, opened 2026-07-20
BPartner v2 upsert: persist under the effective org (path-only) and reject path/body org mismatch; show org in External Reference window

mrie04:
## Problem
The v2 bpartner composite upsert (`PUT /api/v2/bpartner/{orgCode}`, `JsonPersisterService.persistWithinTrx`) resolved the org **twice**: the existing-record lookup used the **path** `{orgCode}`, but the writes (C_BPartner + its external references) used the **body** `bpartnerComposite.orgCode`, falling back to the context/auth-token org. A path-only request (org given only in the path) therefore persisted the partner and its `S_ExternalReference` under the context org, not the requested org → the next call for that org could not re-find them and inserted a **duplicate** partner + external reference.

## Fix
- Resolve **one** effective org `= coalesce(body orgCode, path orgCode, context org)` once in `persistWithinTrx`, and use it for the lookup **and** every write (C_BPartner, its locations/contacts, and both external-reference paths).
- When body and path orgCode are **both** present and resolve to **different** orgs → reject with **HTTP 422** and a user-friendly `JsonErrorItem` whose `errorCode = "BPartnerCompositeOrgMismatch"` (new AD_Message, de_DE/de_CH/en_US). Nothing is persisted. Body == path succeeds normally; neither present falls back to context org (unchanged).
- Scope: only the composite upsert path. Location/contact/bank-account sub-resource endpoints are already org-consistent (untouched).

## Also
- The **Organization** column is now shown in the **External Reference window** (`AD_Window` 540901, `S_ExternalReference`) in both single-record and grid view, so the org placement is directly inspectable.

## Tests
New cucumber feature `bpartnerV2OrgConsistency.feature` (executor4), validated GREEN end-to-end (twice, re-runnable) against a local stack:
- path-only PUT persists C_BPartner + C_BPartner_Location + AD_User + S_ExternalReference under the path org; repeat PUT → UPDATED, exactly one row each (no duplicates).
- body/path org mismatch → 422 + `errorCode=BPartnerCompositeOrgMismatch`, user-friendly.
- GET under the correct org returns the partner; a different org does not (isolation).

## Out of scope
- Cleanup of already-corrupted data (partners/external-refs previously written under the wrong org) — separate task if needed.


backend/de.metas.business.rest-api-impl/src/main/java/de/metas/rest_api/v2/bpartner/bpartnercomposite/jsonpersister/JsonPersisterService.java
```diff
@@ -200,28 +203,43 @@ private JsonResponseBPartnerCompositeUpsertItem persistWithinTrx(
...
 
-		final OrgId orgId = retrieveOrgIdOrDefault(orgCode);
+		final JsonRequestComposite jsonRequestComposite = requestItem.getBpartnerComposite();
+		final String bodyOrgCode = jsonRequestComposite.getOrgCode();
+
+		if (!isBlank(orgCode) && !isBlank(bodyOrgCode))
+		{
+			final OrgId pathOrgId = retrieveOrgIdOrDefault(orgCode);
+			final OrgId bodyOrgId = retrieveOrgIdOrDefault(bodyOrgCode);
+			if (!pathOrgId.equals(bodyOrgId))
```
mrie04: pls use de.metas.organization.OrgId#equals(de.metas.organization.OrgId, de.metas.organization.OrgId)

mrie04: **Resolution (dev):** Adopted — switched the path/body org comparison to the null-safe static `OrgId.equals(pathOrgId, bodyOrgId)` in https://github.com/metasfresh/metasfresh/commit/5705e9f3013a4adbdca561fd8bc7319db00a26a0

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/createbpartner/CreateBPartnerV2_StepDef.java
```diff
@@ -176,9 +302,27 @@ public void verify_contact_is_created_for_bpartner_v2(@NonNull final String acti
...
+			final JsonResponseContact contact;
+			final StepDefDataIdentifier contactOrgIdIdentifier = row.getAsOptionalIdentifier(I_AD_User.COLUMNNAME_AD_Org_ID).orElse(null);
+			if (contactOrgIdIdentifier != null)
+			{
+				final OrgId orgId = orgTable.getId(contactOrgIdIdentifier);
+				final ExternalIdentifier contactExtId = ExternalIdentifier.of(contactIdentifier);
+				contact = jsonRetrieverService
+						.getJsonBPartnerComposite(orgId, ExternalIdentifier.of(bpartnerIdentifier))
+						.flatMap(composite -> composite.getContacts().stream()
+								.filter(c -> isContactMatches(orgId, c, contactExtId))
```
mrie04: pls contact instead of c

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/createbpartner/CreateBPartnerV2_StepDef.java
```diff
@@ -196,8 +340,41 @@ public void verify_contact_is_created_for_bpartner_v2(@NonNull final String acti
...
+
 			softly.assertAll();
 		});
 	}
 
+	/**
+	 * Mirrors {@code BPartnerEndpointService#isJsonContactMatches} for use in this step def when we have an {@link OrgId} directly
+	 * (to avoid the secondary org-code DB lookup that the service performs via {@code RestUtils.retrieveOrgIdOrDefault}).
+	 */
+	private boolean isContactMatches(
```
mrie04: only orgCode lookup isn't reason enough for me to duplicate the logic. Harder to maintain on changes. so why not use existing?

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/bpartner/bpartnerV2OrgConsistency.feature
```diff
@@ -0,0 +1,168 @@
...
+    And verify that S_ExternalReference was created
+      | ExternalSystem | Type     | ExternalReference | AD_Org_ID |
+      | Test_System    | BPartner | 001               | org002    |
+    # no-duplicate: repeat PUT must not create a second C_BPartner or S_ExternalReference under any org
+    And verify that exactly 1 S_ExternalReference exists for:
+      | ExternalSystem | Type     | ExternalReference |
+      | Test_System    | BPartner | 001               |
+    And verify that exactly 1 C_BPartner exists for external identifier:
+      | ExternalSystem | ExternalReference |
+      | Test_System    | 001               |
```
mrie04: why this new stepdefs instead of Then the metasfresh REST-API responds with

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/bpartner/bpartnerV2OrgConsistency.feature
```diff
@@ -0,0 +1,168 @@
...
+    }
+  ],
+  "syncAdvise": {
+    "ifNotExists": "CREATE",
+    "ifExists": "UPDATE_MERGE"
+  }
+}
+"""
+    # GET by path org 002 should find the partner (org isolation: only the correct org returns it)
+    Then the metasfresh REST-API endpoint path 'api/v2/bpartner/002/ext-Test_System-001' receives a 'GET' request with the headers from context, expecting status='200'
```
mrie04, line 175: maybe also Then the metasfresh REST-API responds with ?

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/createbpartner/CreateBPartnerV2_StepDef.java
mrie04: **Resolution (dev):** Renamed the lambda parameter to `contactCandidate` in https://github.com/metasfresh/metasfresh/commit/33d2f57ff8cacd5b1774434eaa55fadb84dfc4bd. It can't be `contact` literally — that would shadow the enclosing result variable `contact` being assigned (Java compile error), so I used the closest non-shadowing descriptive name.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/createbpartner/CreateBPartnerV2_StepDef.java
mrie04: **Resolution (dev):** Agreed — dropped the two custom DB-count step-defs (`verify that exactly 1 …`) in https://github.com/metasfresh/metasfresh/commit/33d2f57ff8cacd5b1774434eaa55fadb84dfc4bd. The no-duplicate invariant is now asserted at the API-contract level via the existing `Then the metasfresh REST-API responds with` step: the repeat PUT must return `syncOutcome: UPDATED` (with the bug it would report CREATED and land a duplicate, failing the assertion). No duplicated lookup logic to maintain.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/bpartner/bpartnerV2OrgConsistency.feature
mrie04: **Resolution (dev):** Replaced the custom count step with the existing `Then the metasfresh REST-API responds with` asserting `syncOutcome: UPDATED` on the repeat PUT (mirrors createBPartnerV2.feature) — proves the upsert re-found and updated the same partner rather than creating a second. https://github.com/metasfresh/metasfresh/commit/33d2f57ff8cacd5b1774434eaa55fadb84dfc4bd

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/bpartner/bpartnerV2OrgConsistency.feature
mrie04: **Resolution (dev):** Added `And the metasfresh REST-API responds with` after the GET /002 → 200, asserting the returned partner (name). https://github.com/metasfresh/metasfresh/commit/33d2f57ff8cacd5b1774434eaa55fadb84dfc4bd

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/createbpartner/CreateBPartnerV2_StepDef.java
mrie04: how is this related to javadoc of this function and my
  question about it (to avoid the secondary org-code DB lookup that the service performs via {@code RestUtils.retrieveOrgIdOrDefault)

mrie04: **Resolution (dev):** You are right that the parallel branch is the wrong shape — but I verified that the literal "use the existing `retrieveBPartner(orgCode)`" does not work here, and the reasoning matters:

`retrieveBPartner(orgCode)` resolves the org by **Value** (`RestUtils.retrieveOrgIdOrDefault` → `IOrgDAO.retrieveOrgIdBy(OrgQuery.ofValue(orgCode))`). The freshly-created **test** org `002` (made via `metasfresh contains AD_Org:`) is not resolvable by Value in the cucumber context → it throws `MissingResourceException: organisation (val-)002 could not be found` (confirmed with a RED). In production Genesis sends a real orgCode that resolves fine; only the synthetic test org fails. So the step must retrieve by the `OrgId` it already holds (from the step-def `AD_Org_StepDefData`), not re-resolve the Value.

To remove the duplication you flagged while keeping a working retrieval, I am adding `OrgId`-based overloads to `BPartnerEndpointService` (`retrieveBPartner/Location/Contact(OrgId, …)`); the existing `String orgCode` methods delegate to them after resolving the Value. The step then calls the `OrgId` overload — using the **existing** match logic, so the mirrored `isContactMatches` helper and the `jsonRetrieverService` field are removed entirely. Commit + updated resolution to follow.

mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/7484d451433a278243c38f25539d789aec36c650 — using the existing `retrieveBPartner/Location/Contact(orgCode)` as you asked, no parallel path.

Root cause (per your `setRequiredAccess(Access.READ)` pointer): the verify steps make **direct** in-process service calls, but `RESTUtil.getAuthToken` only set `loggedUserId` + `org=MAIN` on the test thread — it never loaded the role`s `UserRolePermissions` (the token only authorises HTTP requests, via `UserAuthTokenFilter`). So `OrgDAO.retrieveOrgIdBy(...).setRequiredAccess(READ)` ran without the WebUI role and could not see org `002` → `MissingResourceException`. The role itself is fine (`IsAccessAllOrgs=Y`); the thread just wasn`t running as it.

Fix: `getAuthToken` now also sets `AD_Role_ID` + `AD_Client_ID` on the thread (same context the HTTP filter establishes per request), so `retrieveBPartner(orgCode)` resolves org `002` normally. The three verify steps now call the existing `retrieveBPartner/Location/Contact(orgCode)`; the parallel `jsonRetrieverService` branches, the mirrored `isContactMatches`, the field and 6 imports are removed. All 3 scenarios GREEN.

mrie04:

PR #25261, opened 2026-07-21
Enable multi selection for `Purchase Cockpit - Jump To Purchase Candidate` process

riadmen:


riadmen:
@claude review

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
```diff
@@ -108,21 +125,77 @@ protected IZoomSource createZoomSource(@NonNull final TableRecordReference recor
 		return process;
 	}
 
+	// Constructor for testing multi-selection: inject the selected record refs and a per-ref zoom source
+	@VisibleForTesting
+	public static RelationTypeInOverlayProcess newInstanceForUnitTesting(
```
mrie04: Why a second newInstanceForUnitTesting overload rather than one factory all tests share?

The two variants differ only in that this one overrides getSelectedSourceRecordRefs() while the 4-arg one leaves it real. Two consequences make me think
one map-based factory would be better:

1. The // multi-selection framing isn't the real axis. throwsFriendlyNoRelatedDocs_whenTheSingleSelectedSourceRowCannotBeLoaded is a single-selection test
   that already borrows this "multi" factory (1-element list) purely to get a zoom source that throws — the 4-arg variant's fixed zoom can't. So the actual
   distinction is "throwable per-ref zoom" vs "fixed zoom", not single vs multi.
2. Because this factory stubs getSelectedSourceRecordRefs(), the real production resolution — the IQueryBL…listIds() multi-row branch — is never
   exercised. The tests that do hit the real method (getSelectedSourceRecordRefs_*) only cover the single-ref / id-0 / @NoSelection@ branches.

Could this collapse into a single factory that takes the Map<TableRecordReference, IZoomSource> (covers the throwing case for every test) and drives
selection through a real ProcessInfo + in-memory records instead of overriding the method? OrderAttachmentRowsLoaderTest in this module shows newInstance
+ save + real IQueryBL running in-memory under AdempiereTestHelper, so the multi tests could save in-memory C_OrderLine rows + set a selection filter and
  let the real getSelectedSourceRecordRefs() resolve them. Net: one factory shared by all tests, and the currently-untested DB-resolution branch gets
  covered too.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
```diff
@@ -135,11 +208,139 @@ else if (relatedDocumentGroups.size() > 1)
...
+		}
+
+		final String tableName = getProcessInfo().getTableNameOrNull();
+		final IQueryFilter<Object> selectionFilter = getProcessInfo().getQueryFilterOrElse(null);
+		if (tableName == null || selectionFilter == null)
+		{
+			throw new AdempiereException("@NoSelection@");
+		}
+
+		return Services.get(IQueryBL.class)
```
mrie04: querybl should never be used outside DAO/repo

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
```diff
@@ -135,11 +208,139 @@ else if (relatedDocumentGroups.size() > 1)
...
+	 *     which we run against the source table to get all selected record refs.
+	 * </ul>
+	 */
+	protected List<TableRecordReference> getSelectedSourceRecordRefs()
+	{
+		// Note: getRecordRefOrNull() accepts record_id 0 (it only rejects negative ids), so a view row that resolves to
+		// e.g. RV_PurchaseCockpit/0 (a row with no single backing record) would otherwise be treated as a loadable single
+		// record. Require a real (>0) id here; otherwise fall through to the view selection where-clause below.
+		final TableRecordReference singleRecordRef = getProcessInfo().getRecordRefOrNull();
+		if (singleRecordRef != null && singleRecordRef.getRecord_ID() > 0)
```
mrie04: use repoidaware

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
```diff
@@ -135,11 +208,139 @@ else if (relatedDocumentGroups.size() > 1)
...
-		getResult().setWebuiViewToOpen(WebuiViewToOpen.builder().viewId(viewId.getViewId()).target(getOpenTarget()).build());
+	/**
+	 * Opens a single combined view showing the union of the related documents of all selected source records.
+	 * <p>
+	 * We cannot rely on {@code CreateViewRequest.referencingDocumentPaths} here: {@code SqlViewFactory} derives the
+	 * related-documents sticky filter from {@code getSingleReferencingDocumentPathOrNull()} (i.e. only the first path),
+	 * which would collapse the union to a single source. Instead we OR the per-source related-documents SQL where-clauses
+	 * into one sticky filter.
+	 */
+	private ViewId createCombinedView(@NonNull final List<TableRecordReference> sourceRecordRefs, @NonNull final RelationTypeId relationTypeId)
```
mrie04, line 228: Is there a reason not to reuse the framework's setRecordsToOpen here, the way Material Cockpit already does?

MD_Cockpit_DocumentDetail_Display handles the same shape — multi-selected rows → open one combined view of related records — in three lines:

final Set<Integer> cockpitRowIds = getSelectedCockpitRecordIdsRecursively();
final List<TableRecordReference> detailRecords = retrieveCockpitDetailRecordReferences(cockpitRowIds); // IQueryBL … listIds()
getResult().setRecordsToOpen(detailRecords, WINDOWID_MaterialCockpit_Detail_String);

setRecordsToOpen(records, windowId) hands the framework a list of record refs + a target window and opens a GridView of exactly those rows — no
MQuery.getWhereClause, no (…) OR (…) string, no dedup set, no null/blank-where guards, no hand-built sticky DocumentFilter.

The resolve half here (getSelectedSourceRecordRefs() via IQueryBL…listIds()) is already the same pattern Material Cockpit uses; the only divergence is the
open half — which is exactly the ~70 lines of createCombinedView. It looks like it could execute each candidate's MQuery against its zoom table to
materialize the related target refs, union them, and call setRecordsToOpen(relatedRefs, targetWindowId) — dropping the OR-join, the LinkedHashSet dedup,
and both skip-guards.

Two caveats I can see, so this isn't a pure swap:

1. Eager vs lazy — the single-source path (createView with setReferencingDocumentPaths) and the combined sticky filter are both lazy (the grid pages
   through the filter). setRecordsToOpen materializes the full related-ID set up front. Fine for cockpit-detail sizes; a "jump from many rows" union could be
   larger.
2. Open target — setRecordsToOpen hardcodes GridView + SAME_TAB_OVERLAY, whereas this process honours getOpenTarget() (ModalOverlay / NewBrowserTab).
   Preserving that means building RecordsToOpen directly with the target rather than the convenience overload.

I get that the JavaDoc shows you hit the referencingDocumentPaths-reads-only-the-first-path wall; the framework-level fix (support multiple referencing
paths) is bigger, and setRecordsToOpen sidesteps it by materializing. So this is a design question, not a blocker — mainly: is the raw WHERE-union buying
us anything over the framework's records-to-open path that Material Cockpit already relies on?

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
```diff
@@ -202,10 +403,6 @@ private RelationTypeId getRelationTypeId()
 	@Override
 	public ProcessPreconditionsResolution checkPreconditionsApplicable(final @NonNull IProcessPreconditionsContext context)
 	{
-		if (!context.isSingleSelection())
-		{
-			return ProcessPreconditionsResolution.rejectBecauseNotSingleSelection();
-		}
 		if (context.getAdWindowId() == null)
```
mrie04, line 414: consider something like following **and** noSelection isn't guarded now 
final int maxSelectionSize = sysConfigBL.getIntValue(SYSCONFIG_MaxSelectionSize, DEFAULT_MaxSelectionSize);

if (maxSelectionSize > 0 && context.isMoreThanAllowedSelected(maxSelectionSize))
{
return ProcessPreconditionsResolution.rejectBecauseTooManyRecordsSelected(maxSelectionSize);
}

And also

mrie04 (changes requested):
pls check my comments

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
riadmen: Done, moved the query off `IQueryBL` and onto the framework helper `JavaProcess.retrieveSelectedRecordsQueryBuilder(Object.class)`.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
riadmen: Replaced the raw `getRecord_ID() > 0` with `getProcessInfo().isRecordSet()`. Since the source table is generic there's no concrete XxxId to build, so this is the closest repo-id-aware framework check that still rejects id 0.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
riadmen: Added both guards. `isNoSelection()` now rejects up front, and the max is sysconfig-driven `SYSCONFIG_MaxSelectionSize` (default 100, set 0 to disable) via `isMoreThanAllowedSelected → rejectBecauseTooManyRecordsSelected`.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
riadmen: 1. Point 1: Fixed, collapsed the two factories into one keyed by `Map<TableRecordReference, IZoomSource>`, with a nullable injected selection, so the single-vs-multi framing is gone.
2. point 2: After the `IQueryBL` fix, we no longer build the query. We hand that job to the framework method   
  `JavaProcess.retrieveSelectedRecordsQueryBuilder(...)` and just use its result.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
riadmen: Kept the sticky-filter path over `setRecordsToOpen` because `setRecordsToOpen` materializes the full related-id set up front (a many-row union can be large) and hardcodes `GridView` + `SAME_TAB_OVERLAY`, whereas the sticky filter stays lazy and honours `getOpenTarget()` (`ModalOverlay`/`NewBrowserTab`). 
This justification has been recorded in the Javadoc documentation for `createCombinedView`.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/process/RelationTypeInOverlayProcess.java
```diff
@@ -135,11 +210,143 @@ else if (relatedDocumentGroups.size() > 1)
...
+				continue;
+			}
+
+			for (final RelatedDocumentsCandidateGroup group : groups)
+			{
+				if (targetWindowId == null)
+				{
+					targetWindowId = group.getTargetWindowId();
+				}
+				else if (!Objects.equals(targetWindowId, group.getTargetWindowId()))
```
mrie04: consider de.metas.ui.web.window.datatypes.WindowId.equals(de.metas.ui.web.window.datatypes.WindowId, de.metas.ui.web.window.datatypes.WindowId)

riadmen: fix done

mrie04 (approved):
lgtm

PR #25271, opened 2026-07-21
Advance Payment no LC: proforma allocation without Letter-of-Credit + transport-order BL/ETA date follow-up

teosarca:
## What

Two enhancements to the Multiple Levels of Payment feature (complex purchase payment terms whose advance portion is paid via a proforma invoice).

### US01 — advance proforma works without a Letter-of-Credit break
Previously the proforma→purchase-order allocation was hard-coded to require exactly one Letter-of-Credit (LC) break; a term whose advance milestone is Order-Date / Invoice-Date (no LC) could not be allocated, and its advance line never reached *Paid*.

- The allocation gate no longer requires an LC break. A term with **no LC break** is allowed; its single **advance** line (the one non-material-receipt break — Order Date / Invoice Date) is marked **Paid** on proforma payment, amount driven by the proforma total — mirroring today's LC behaviour, without stamping `LC_Date`.
- A term with **more than one advance line and no LC** fails loud (`MultipleAdvanceBreaksUnsupported`).
- A term with **an LC break is byte-for-byte unchanged** (allocation, transition, amount, messages, `LC_Date` stamping). Selection is LC-first (`getSingleLCLine().orElseGet(getSinglePrepaidLine)`), so the advance path is reachable only when there is no LC line.

### US02 — the shipping line's due date follows the Bill-of-Lading date after completion
The BL/ETA pay-schedule line's due date is a placeholder (`9999-12-01`) until the goods' shipping date is known. That date lives on the Transport Order and is frequently entered/corrected **after** the Transport Order is completed.

- A new `@ModelChange(AFTER_CHANGE, {BLDate, ETA})` handler on `M_ShipperTransportation`, guarded to fire only when the Transport Order is **Completed/Closed**, re-runs the existing propagation. A **draft** Transport Order does not propagate.
- `OrderPaySchedule.updateStatusFromContext` is broadened to also refresh an **unsettled awaiting-pay material-receipt (BL/ETA) line** — needed for the post-completion *correction* case (the line is already awaiting-pay). Gated on `isMaterialReceiptDate()` so LC/Order-Date lines are provably untouched.

## Testing
- Unit (JUnit5 + AdempiereTestHelper): proforma allocation (no-LC succeeds; >1-advance rejected; LC still succeeds), advance-line→Paid + `LC_Date` stays null, `getSinglePrepaidLine`, and the `updateStatusFromContext` BL + ETA refresh with the LC line asserted untouched.
- Cucumber (`purchaseOrderComplexPaymentTerm.feature`): BL date entered / corrected after Transport-Order completion recomputes the shipping line; a draft Transport Order does not propagate. All green.
- Regression: the Multiple-Levels-of-Payment cucumber suite's LC-path scenarios remain green (LC behaviour unchanged).

## Notes for the reviewer
- A payment term is mandatory on a completed purchase order (`C_PaymentTerm_ID` NOT NULL), so the null-payment-term path in the allocation gate is unreachable — no guard added for it.
- The old `NoLCBreakInOrder` message key is removed in code; its `AD_Message` row is now unreferenced (harmless; a retire-message migration is a possible follow-up).


backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/steps/letter_of_credit/OrderPayScheduleLCStepService.java
```diff
@@ -109,8 +109,8 @@ private void recomputeLCStep(@NonNull OrderId orderId, @Nullable final Prepaymen
 		final OrderPaySchedule schedule = orderPayScheduleService.getByOrderId(orderId).orElse(null);
 		if (schedule == null) {return;} // no pay-schedule at all — no-op
 
-		final OrderPayScheduleLine lcStep = schedule.getSingleLCLine().orElse(null);
-		if (lcStep == null) {return;} // no LC break in payment term — no-op
+		final OrderPayScheduleLine target = schedule.getSingleLCLine().orElseGet(() -> schedule.getSingleAdvanceLine().orElse(null));
```
teosarca: why not taking the eligible steps in order and allocate the prepayment to those steps in order?

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/steps/letter_of_credit/OrderPayScheduleLCStepService.java
```diff
@@ -127,9 +127,12 @@ private void recomputeLCStep(@NonNull OrderId orderId, @Nullable final Prepaymen
 		if (proforma == null)
 		{
 			// No proforma allocation — reset to Pending
-			schedule.markAsPending(lcStep.getIdNotNull());
+			schedule.markAsPending(target.getIdNotNull());
 			orderPayScheduleService.save(schedule);
-			clearLCDateOnOrder(orderId);
+			if (target.isLetterOfCreditDate())
+			{
+				clearLCDateOnOrder(orderId);
```
teosarca, line 134: for my information, why do we have to clear the LC date, can u pls remind me?

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/purchaseOrder/purchaseOrderComplexPaymentTerm.feature
```diff
@@ -244,4 +244,169 @@ Feature: Purchase order with complex payment term
     Then the order identified by po3 has following pay schedules
       | C_PaymentTerm_Break_ID | DueDate    | DueAmt | Status |
       | PTB31                  | 2025-10-10 | 25.58  | WP     |
-      | PTB32                  | 2025-04-01 | 76.72  | WP     |
\ No newline at end of file
+      | PTB32                  | 2025-04-01 | 76.72  | WP     |
+
```
teosarca, line 248: do we also have cucumber tests for non-LC cases?

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/core/OrderPaySchedule.java
```diff
@@ -160,4 +163,29 @@ else if (lcLines.size() > 1)
...
+	/**
+	 * The single advance break of the payment term, i.e. the one line that is neither a material-receipt
+	 * (BL/ETA) line nor an LC line. LC lines are excluded explicitly so the method stays honest regardless
+	 * of the caller: an "advance line" is never an LC line, even on a schedule that happens to carry both
+	 * (LC lines are resolved via {@link #getSingleLCLine()}).
+	 */
+	public Optional<OrderPayScheduleLine> getSingleAdvanceLine()
+	{
+		final ImmutableList<OrderPayScheduleLine> advanceLines = lines.stream()
+				.filter(line -> !line.isMaterialReceiptDate() && !line.isLetterOfCreditDate())
```
teosarca: question: why excluding material receipt date lines?
(just answer to question, don't be sycophant)

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/core/OrderPayScheduleLine.java
```diff
@@ -132,6 +134,23 @@ else if (!OrderPayScheduleId.equals(this.id, id))
...
+	/**
+	 * True for an unsettled, still-open material-receipt (BL/ETA) line that the generic recompute is
+	 * allowed to refresh on a later reference-date correction: it is {@code Awaiting_Pay} (a Transport
+	 * Order already completed it, so it is no longer {@code Pending}), not yet paid, and not linked to a
+	 * committed downstream document. LC/OD lines are excluded ({@code isMaterialReceiptDate()}) — an LC
+	 * line is refreshed only via the dedicated LC step service. {@code !isPaid()} guards against a
+	 * hand-edited/legacy row where status and {@code isPaid} drift apart (isPaid is a separate DB column,
+	 * not derived from status).
+	 */
+	public boolean isUnsettledAwaitingPayMaterialReceipt()
```
teosarca, line 151: if find tihs a bit twisted logic. what are we actually trying to achieve?

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/steps/letter_of_credit/OrderPayScheduleLCStepService.java
teosarca: **Resolution (dev):** By design this iteration supports exactly one proforma-paid step per term — one LC break, or (no-LC) one advance break. `getSingleLCLine()` / `getSingleAdvanceLine()` fail loud (`MultipleLCBreaksUnsupported` / `MultipleAdvanceBreaksUnsupported`) if a term defines more than one, so at this point there is at most one eligible target — nothing to order. Allocating a prepayment across multiple eligible steps *in order* is the natural generalization, but it's out of F00994 iter-4 scope and belongs to the split-payment feature (me03 https://github.com/metasfresh/me03/issues/29369). Glad to bring that forward here if you'd prefer.

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/steps/letter_of_credit/OrderPayScheduleLCStepService.java
teosarca: **Resolution (dev):** When the target is an LC line and a proforma is allocated+paid, `stampLCDateOnOrder` sets `C_Order.LC_Date = proforma.DateInvoiced` (~line 177). This branch is the inverse: the proforma allocation is gone (`proforma == null` → reset to Pending), so the previously-stamped `LC_Date` must be cleared to null — otherwise a stale LC date lingers on the order after its source proforma is removed. Guarded to LC-type targets only, since only they stamp it.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/purchaseOrder/purchaseOrderComplexPaymentTerm.feature
teosarca: **Resolution (dev):** Yes for the no-LC pay-schedule + BL/ETA propagation — S30954_1/2/3 all use a no-LC term (`pt_PO_4` = OD 10% + BL 90%, no LC break), so the non-LC recompute path is cucumber-covered. The no-LC *proforma-pays-the-advance-line* flow (US01) is covered at JUnit level — `ProformaOrderAllocServiceTest.allocate_noLcBreak_succeeds` + `OrderPayScheduleGetSingleAdvanceLineTest` — not cucumber, consistent with the existing choice (feature comment ~lines 175-183) that the proforma-payment flow is exercised under `splitPaymentLC` / the split-payment feature (me03 https://github.com/metasfresh/me03/issues/29369). If you'd like a dedicated end-to-end no-LC proforma-payment cucumber scenario, I can add one.

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/core/OrderPaySchedule.java
teosarca: **Resolution (dev):** A material-receipt (BL/ETA) line is a delivery-driven step, due after goods receipt — it is never the step paid up-front against the proforma. The advance step is the order-date (OD) break paid in advance. So `getSingleAdvanceLine()` excludes material-receipt lines (and LC lines) to isolate that single advance break; the fail-loud guard then ensures there is exactly one.

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/core/OrderPayScheduleLine.java
teosarca: **Resolution (dev):** Goal: after a Transport Order completes and its BL/ETA date is later entered/corrected, move that delivery line's due date — but only while the line is still "open". The four conditions define exactly that open state: it is a material-receipt (BL/ETA) line, currently `Awaiting_Pay` (the TO already pushed it past `Pending`), not yet paid, and not linked to a committed goods-receipt/invoice. I extracted the inline 4-term boolean into `isUnsettledAwaitingPayMaterialReceipt()` to *name* that state rather than narrate it inline. If the name/shape still reads as twisted, happy to refactor further — suggestions welcome.

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/steps/letter_of_credit/OrderPayScheduleLCStepService.java
teosarca: ok

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/purchaseOrder/purchaseOrderComplexPaymentTerm.feature
teosarca: pls always favor e2e tests (cucumber in this case) instead of junit!

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/core/OrderPaySchedule.java
teosarca: imho  u shall
* rename getSingleAdvanceLine to getSinglePrepaidLine
* introduce de.metas.order.paymentschedule.core.OrderPayScheduleLine#isPrepaidLine which considers both LetterOfCreditDate and OrderDate as prepaid (returns true)
* it is ok to consider also LC lines here because we always check LC line first and then we call this one. maybe that checking wont be necesarry either and simply honor the first prepaid step in order.

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/core/OrderPayScheduleLine.java
teosarca: ok then maybe the javadoc can be more CSP

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/core/OrderPaySchedule.java
teosarca: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/f951a4c7ea7a1e511dcacc53e8e739d13199e23e.
- Renamed `getSingleAdvanceLine` → `getSinglePrepaidLine`.
- Added `OrderPayScheduleLine#isPrepaidLine()` = Letter-of-Credit **or** order-date (advance) break. An invoice-date break is a regular post-invoice term and a BL/ETA break is delivery-driven, so neither counts as prepaid. `getSinglePrepaidLine` filters on it, and the 4-line "why LC is excluded" Javadoc is gone.

On *"maybe drop the LC-first check and just honor the first prepaid step in order"*: I kept `getSingleLCLine().orElseGet(getSinglePrepaidLine)` for two reasons — (1) `getSingleLCLine` is still needed on its own (`OrderPayScheduleRegularInvoiceService` uses it for the LC-percent lookup, and `stampLCDateOnOrder`/`clearLCDateOnOrder` are gated on the target being an LC line); (2) "honor the first prepaid step in order" would replace the fail-loud >1-prepaid guard — the human-approved US01 requirement — with silent first-wins. I flagged it rather than relaxing the approved contract unilaterally; happy to do it as a follow-up if we want to change that requirement.

Note on scope: on a no-LC term `getSinglePrepaidLine` now matches order-date only (before: any non-material-receipt, i.e. also invoice-date). Every term the allocate-gate currently accepts is unaffected; the only differing shape (OD + invoice-date, no LC) is still rejected by `ProformaOrderAllocateCommand.validate()` before this method is reached, so behaviour is unchanged. If you'd like that gate to also move to the prepaid concept (→ accept OD + invoice-date terms), that's a deliberate behaviour change I can make and de-risk against the live terms.

backend/de.metas.business/src/main/java/de/metas/order/paymentschedule/core/OrderPayScheduleLine.java
teosarca: **Resolution (dev):** Tightened the `isUnsettledAwaitingPayMaterialReceipt` Javadoc to 5 lines while keeping each condition's rationale (Awaiting_Pay = a Transport Order completed it so it is no longer Pending; `!isPaid()` guards a legacy row whose status and `isPaid` column have drifted; downstream-link exclusion). Done in https://github.com/metasfresh/metasfresh/commit/f951a4c7ea7a1e511dcacc53e8e739d13199e23e.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/purchaseOrder/purchaseOrderComplexPaymentTerm.feature
teosarca: **Resolution (dev):** Agreed. Adding a Cucumber e2e scenario (`@Id:S30954_4`) for the no-LC proforma-pays-advance flow now (this case was previously JUnit-only). I'll update this thread with the scenario + commit URL once it runs green. Noted the favour-e2e-over-JUnit preference going forward.

teosarca: **Resolution (dev):** Done — Cucumber scenario `@Id:S30954_4` *"Advance paid via a proforma marks the order-date step Paid on a no-LC payment term"* added to `purchaseOrderComplexPaymentTerm.feature` (https://github.com/metasfresh/metasfresh/commit/ae7d7b19d38ad3d1a770535df90f0a4be56c974c). Faithful e2e of the no-LC flow: no-LC term (OD 10% + BL 90%) → PO completed → purchase proforma completed + allocated → full prepayment completed → the OD pay-schedule step goes **Paid** while the BL step stays open. Green in provided-infra mode against the `new_dawn_uat` stack (1 scenario / 36 steps, 0 failures). Exercises `OrderPayScheduleLCStepService.recomputeLCStepAfterPaymentCompleted` → `getSinglePrepaidLine()`.

PR #25273, opened 2026-07-22
HU org: derive M_HU.AD_Org_ID from locator warehouse at creation + repair migration

metas-ts:
## Problem

A physical Handling Unit lives in a locator → warehouse → exactly one real org, yet an `M_HU`'s `AD_Org_ID` is never tied to that physical location: a new HU inherits its org from the creation **context**. Any HU-creating flow running under an org-0 (`*`/ANY) context therefore leaves a physical HU with `AD_Org_ID=0`, even though its warehouse belongs to a real org.

Such org-0 HUs break stock reconciliation: `MD_Stock_From_HUs_V` groups HU stock by `hu.AD_Org_ID` while taking the warehouse from the locator, so an org-0 HU produces an org-0 aggregate row for a real-org warehouse that reconciles against no `MD_Stock` row. `MD_Stock_Update_From_M_HUs` then tries to create an `MD_Stock` row / fire a stock event for org 0, and the whole run fails.

## Fix

**Enforce the invariant at the single source (creation):** a new `@ModelChange(BEFORE_NEW)` interceptor on `M_HU` derives `AD_Org_ID` from the HU's locator → warehouse org. No locator ⇒ context org is left untouched (such HUs are not physical stock in the view). This covers every creation flow (`HUBuilder` and any `newInstance(I_M_HU.class)`), not one call site.

**Repair existing corrupt rows:** an idempotent system migration sets `AD_Org_ID` to the locator's warehouse org for every active `M_HU` currently at `AD_Org_ID=0` whose warehouse has a real org. Touches only such rows; a re-run corrects nothing further.

Scope is **creation-time only** — no re-derivation on locator change (cross-org HU movements do not occur).

## Verification

- TDD unit test `HUBuilderTests#huCreatedUnderOrgZeroContextGetsWarehouseOrg` (RED→GREEN): an HU built under an org-0 context at a real-org locator gets the warehouse org; no-regression case (context org already correct) unchanged. Existing HU suite green.
- Migration applied against a local customer-faithful DB (applies cleanly, registered, idempotent); repair logic verified on a synthetic org-0 HU (org 0 → warehouse org; second run 0 rows).
- End-to-end verification that the stock-update process completes on corrected data is a manual UAT on the target instance after merge-up.

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/sql/postgresql/system/70-de.metas.handlingunits/5815330_sys_repair_M_HU_org_zero.sql
+++ b/backend/de.metas.handlingunits.base/src/main/sql/postgresql/system/70-de.metas.handlingunits/5815330_sys_repair_M_HU_org_zero.sql
@@ -14,5 +14,4 @@ UPDATE m_hu hu
   JOIN m_warehouse w ON w.m_warehouse_id = l.m_warehouse_id
  WHERE hu.m_locator_id = l.m_locator_id
    AND hu.ad_org_id    = 0
-   AND hu.isactive     = 'Y'
    AND w.ad_org_id     > 0;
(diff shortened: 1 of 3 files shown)


backend/de.metas.handlingunits.base/src/main/sql/postgresql/system/70-de.metas.handlingunits/5815330_sys_repair_M_HU_org_zero.sql
```diff
@@ -0,0 +1,18 @@
...
+
+UPDATE m_hu hu
+   SET ad_org_id = w.ad_org_id,
+       updated   = TO_TIMESTAMP('2026-07-22 07:00:00', 'YYYY-MM-DD HH24:MI:SS'),
+       updatedby = 99
+  FROM m_locator l
+  JOIN m_warehouse w ON w.m_warehouse_id = l.m_warehouse_id
+ WHERE hu.m_locator_id = l.m_locator_id
+   AND hu.ad_org_id    = 0
+   AND hu.isactive     = 'Y'
```
metas-ts:
```suggestion
```

PR #25275, opened 2026-07-22
S_ExternalReference: per-org external-reference uniqueness (add AD_Org_ID to unique index)

mrie04:
## What

Make `S_ExternalReference` uniqueness **per-org**, aligning the enforced DB uniqueness with the org-scoped lookup.

`ExternalReferenceRepository` resolves external references filtered by `AD_Org_ID` (the lookup is org-scoped, `ExternalReferenceQuery.orgId` is `@NonNull`), but the unique index `idx_s_externalreference_externalsystem_type_externalreference` was **org-agnostic**: `UNIQUE(ExternalReference, Type, ExternalSystem_ID) WHERE isactive='Y'`. So two different orgs using the same external system + reference code collided on the global index even though the lookup treats them as distinct per org (the second org's write failed with a unique-constraint violation).

## Change

- Migration `5815340`: add `AD_Org_ID` (SeqNo 40, tenant column last) to the index — both the `AD_Index_Table` definition (`AD_Index_Column` row) and the physical index → `UNIQUE(ExternalReference, Type, ExternalSystem_ID, AD_Org_ID) WHERE isactive='Y'`. Loosening change: every row that satisfied the old narrower index still satisfies the new one, so no data migration/dedup is needed.
- The sibling index `(Type, ExternalSystem_ID, Record_ID)` is intentionally left unchanged — `Record_ID` is a metasfresh id that already belongs to exactly one org, so it cannot collide across orgs.
- Cucumber `bpartnerV2OrgConsistency.feature` TC4: the same external reference is created under two different orgs and both persist. The `verify that S_ExternalReference was created` step def is made org-aware (scopes the lookup by `AD_Org_ID` when provided) so per-org references can be asserted individually.

## Validation

Index-level RED→GREEN proven locally (transactional, rolled back): with the current global index the second cross-org insert fails with `duplicate key ... (externalreference, type, externalsystem_id)=(...)`; after the migration DDL both inserts succeed. End-to-end TC4 + compilation validated by CI.

### DIFF
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/externalreference/S_ExternalReference_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/externalreference/S_ExternalReference_StepDef.java
@@ -156,7 +156,7 @@ public class S_ExternalReference_StepDef
 
 			// scope by org when provided, so the same external reference can be asserted independently per org
 			row.getAsOptionalIdentifier(I_S_ExternalReference.COLUMNNAME_AD_Org_ID)
-					.ifPresent(orgIdentifier -> queryBuilder.addEqualsFilter(I_S_ExternalReference.COLUMNNAME_AD_Org_ID, orgTable.getIdAsInt(orgIdentifier)));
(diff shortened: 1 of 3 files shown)


mrie04:
known flaky mobile test, merging

mrie04:

PR #25287, opened 2026-07-22
InArrayQueryFilter — render oversized IN lists inline to stay under the JDBC 32767 bind-param limit

metas-ts:
## Problem
`IQueryBuilder.addInArrayFilter` / `addNotInArrayFilter` render `col IN (?,?,…)` — one JDBC bind parameter per value. PostgreSQL's wire protocol caps bind parameters at 32767 (a signed 2-byte count) per statement, so a data-volume-driven list overflows with `Tried to send an out-of-range integer as a 2-byte value`. Both methods build the same shared `InArrayQueryFilter`, so this affects all of its ~547 call sites.

## Fix
`InArrayQueryFilter.buildSql()` now renders the values **inline** (via the existing `embedSqlParams` path, `DB.TO_SQL`) automatically once a list exceeds `MAX_BIND_PARAMS_BEFORE_EMBED` (30000, headroom below 32767) — producing zero bind parameters for oversized lists, so the overflow is impossible. Small lists keep the `?` form unchanged (plan-cacheable — the common case). Covers both `IN` and `NOT IN` (same filter class); NULL handling (`… IS NULL`) is preserved.

`DB.TO_SQL` gains a general `Number` branch so inline rendering accepts every value type the `?` bind path did (e.g. `Long`), matching the prior behavior for lists in the 30001–32767 range.

_Note: this bounds a single IN filter; a statement combining several near-threshold IN filters could still exceed the per-statement cap — such a query should bind the list as an array (`= ANY(?)`), a fuller future improvement._

## Tests
`InArrayQueryFilterTest` (13/13 green): oversized list renders inline with zero bind params; oversized `Long` list renders without throwing (regression guard); oversized NULL-mixed list renders `(… IN (…) OR … IS NULL)` inline; small list keeps `?`; cap invariant `≤ 32767`.

### DIFF
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/util/DB.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java-legacy/org/compiere/util/DB.java
@@ -1906,6 +1906,13 @@ public class DB
 		{
 			return TO_STRING(DisplayType.toBooleanString((Boolean)param));
 		}
+		else if (param instanceof Number)
+		{
+			// Long/Short/Byte/AtomicInteger/Double/... — the JDBC '?' bind path accepts any Number, so inline rendering
+			// must too (else switching a large IN list to inline would reject a type that used to bind). Integer and
+			// BigDecimal are handled by their specific branches above; this covers the rest.
+			return param.toString();
+		}
 		else
 		{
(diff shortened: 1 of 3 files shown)


teosarca:
on a quick look, lgtm

PR #25294, opened 2026-07-22
externalsystems scripted-import — post-rollout follow-up fixes

metas-ts:
## externalsystems scripted-import — post-rollout follow-up fixes

Collected follow-up fixes found during UAT of the SFTP scripted-import processor (delivered via https://github.com/metasfresh/metasfresh/pull/25182). Bundled into one PR rather than one-per-fix.

### Included

1. **Startup 500 — `invoke` rejects the concrete command code.** The generic external-system invoke path (`…/externalsystem/invoke/{type}/{value}/{request}`) and `ExternalSystem_Service.EnableCommand`/`DisableCommand` pass the concrete command (`enableSftpPolling`/`disableSftpPolling`/`enableRestAPI`/`disableRestAPI`); the camel reconciler fires this on startup. `InvokeScriptedImportConversionAction` only accepted the Start/Stop *intent* → `No ScriptedImportConversionIntent for code` → HTTP 500 on startup. Fix: the action now resolves an intent from **either** an intent code (`start`/`stop`, the manual process param) **or** a concrete command code (mapping enable*→Start, disable*→Stop). Adds `ScriptedImportConversionCommand.ofCode` + intent mapping. TDD-covered.

2. **`Skript-Kennung` (ScriptIdentifier) field description.** Corrected the misleading "name of the JavaScript file" wording (which implied entering the `.js`) to state the value is the script identifier — the file name **without** the `.js` extension (the loader appends `.js`). Applied to the shared AD element + its dependent columns/fields/UI-elements across both scripted export + import configs, all languages.

3. **Stale SFTP route on endpoint change (Stop/Start leaks the old poller).** _(in progress)_ After changing an import endpoint's SFTP host and Stop→Start via the process, both the old and new SFTP servers are polled — Stop does not remove the previously-started dynamic route. Fix keys the dynamic route lifecycle on a stable identifier so Stop reliably tears down the running route regardless of connection-detail changes.

_(Further UAT follow-up fixes may be added to this PR.)_

### DIFF
--- a/backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/process/InvokeScriptedImportConversionAction.java
+++ b/backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/process/InvokeScriptedImportConversionAction.java
@@ -120,10 +120,8 @@ public class InvokeScriptedImportConversionAction extends AlterExternalSystemSer
 	}
 
 	/**
-	 * Resolve the Start/Stop intent from {@code External_Request}. Two callers pass different codes here:
-	 * the manual "call" process passes a Start/Stop intent (AD_Reference 541998), while the generic
-	 * external-system infra (startup reconciler) and the {@code ExternalSystem_Service} enable/disable
-	 * commands pass the CONCRETE transport command (e.g. {@code enableSftpPolling}). Accept both.
+	 * The manual process passes a Start/Stop intent, but the generic infra + {@code ExternalSystem_Service}
+	 * enable/disable commands pass the CONCRETE transport command here — so accept both.
 	 */
 	@VisibleForTesting
 	static ScriptedImportConversionIntent resolveIntent(@NonNull final String externalRequest)
--- a/backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/process/InvokeScriptedImportConversionAction.java
+++ b/backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/process/InvokeScriptedImportConversionAction.java
@@ -22,6 +22,7 @@
 
 package de.metas.externalsystem.process;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import de.metas.common.externalsystem.JsonExternalSystemName;
 import de.metas.common.externalsystem.JsonExternalSystemRequest;
@@ -40,11 +41,13 @@ import de.metas.externalsystem.scriptedimportconversion.ExternalSystemScriptedIm
 import de.metas.externalsystem.scriptedimportconversion.ExternalSystemScriptedImportConversionConfigId;
 import de.metas.externalsystem.scriptedimportconversion.ExternalSystemScriptedImportConversionService;
 import de.metas.externalsystem.scriptedimportconversion.ExternalSystemScriptedImportConversionService.ResolvedChildCommand;
+import de.metas.externalsystem.scriptedimportconversion.ScriptedImportConversionCommand;
 import de.metas.externalsystem.scriptedimportconversion.ScriptedImportConversionIntent;
 import de.metas.logging.LogManager;
 import de.metas.process.IProcessPreconditionsContext;
 import de.metas.process.PInstanceId;
 import de.metas.process.ProcessPreconditionsResolution;
+import lombok.NonNull;
 import org.adempiere.exceptions.AdempiereException;
 import org.compiere.SpringContextHolder;
 import org.slf4j.Logger;
@@ -74,7 +77,7 @@ public class InvokeScriptedImportConversionAction extends AlterExternalSystemSer
 	@Override
 	protected String doIt() throws Exception
 	{
-		final ScriptedImportConversionIntent intent = ScriptedImportConversionIntent.ofCode(externalRequest);
+		final ScriptedImportConversionIntent intent = resolveIntent(externalRequest);
 
 		final ImmutableList<ResolvedChildCommand> resolvedCommands = resolveChildCommands(intent);
 		if (resolvedCommands.isEmpty())
@@ -114,6 +117,36 @@ public class InvokeScriptedImportConversionAction extends AlterExternalSystemSer
 		}
 
 		return MSG_OK + " (" + succeeded + "/" + resolvedCommands.size() + ")";
+	}
+
+	/**
+	 * Resolve the Start/Stop intent from {@code External_Request}. Two callers pass different codes here:
+	 * the manual "call" process passes a Start/Stop intent (AD_Reference 541998), while the generic
+	 * external-system infra (startup reconciler) and the {@code ExternalSystem_Service} enable/disable
+	 * commands pass the CONCRETE transport command (e.g. {@code enableSftpPolling}). Accept both.
+	 */
+	@VisibleForTesting
+	static ScriptedImportConversionIntent resolveIntent(@NonNull final String externalRequest)
+	{
+		try
+		{
+			return ScriptedImportConversionIntent.ofCode(externalRequest);
+		}
+		catch (final AdempiereException intentNotFound)
+		{
+			try
+			{
+				return ScriptedImportConversionCommand.ofCode(externalRequest).getIntent();
+			}
+			catch (final AdempiereException commandNotFound)
+			{
+				throw new AdempiereException("No ScriptedImportConversion intent or command for External_Request")
+						.appendParametersToMessage()
+						.setParameter("External_Request", externalRequest)
+						.setParameter("acceptedIntents", "start, stop")
+						.setParameter("acceptedCommands", "enableRestAPI, disableRestAPI, enableSftpPolling, disableSftpPolling");
+			}
+		}
 	}
 
 	private ImmutableList<ResolvedChildCommand> resolveChildCommands(final ScriptedImportConversionIntent intent)
--- a/backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/process/InvokeScriptedImportConversionAction.java
+++ b/backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/process/InvokeScriptedImportConversionAction.java
@@ -126,25 +126,23 @@ public class InvokeScriptedImportConversionAction extends AlterExternalSystemSer
 	@VisibleForTesting
 	static ScriptedImportConversionIntent resolveIntent(@NonNull final String externalRequest)
 	{
-		try
+		final ScriptedImportConversionIntent intent = ScriptedImportConversionIntent.ofCodeOrNull(externalRequest);
+		if (intent != null)
 		{
-			return ScriptedImportConversionIntent.ofCode(externalRequest);
+			return intent;
 		}
-		catch (final AdempiereException intentNotFound)
+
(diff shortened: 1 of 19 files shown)


misc/services/camel/de-metas-camel-externalsystems/de-metas-camel-scriptedadapter/src/main/java/de/metas/camel/externalsystems/scriptedadapter/convertmsg/to_mf/ScriptedImportConversionSftpDynamicRouteBuilder.java
```diff
@@ -41,7 +41,14 @@ public class ScriptedImportConversionSftpDynamicRouteBuilder extends AbstractScr
 {
 	@NonNull private final String sftpUri;
 
+	/**
+	 * Stable per-child route id. Keyed on the child config id (never on the endpoint host/{@code Value}),
+	 * so enable/disable can find and tear down the running poller even after the child's endpoint changed.
+	 */
+	@NonNull private final String routeKey;
```
metas-ts, line 48: assume i change the endpoint in metasfresh and then do a stop and start. will it work? will the previous poller be shut down?

metas-ts: **Resolution (dev):** Yes — that's exactly the scenario this commit fixes (before it, the old poller leaked). Change the endpoint, then Stop→Start (or just Start): the previous poller is shut down.

Why it works now: the route is keyed on `routeKey = "ScriptedImportConversion-" + childConfigId` — the child config id, which is **stable across an endpoint/host/Value change** (backend `ExternalSystemScriptedImportConversionService.getParameters`, sent identically on the enable AND disable requests). So:

- **Stop (disable):** `removePollingRoute(routeKey)` → `stopRoute`+`removeRoute` on that stable key → matches the running poller regardless of the changed endpoint, and tears it down (+ deletes its ssh-key temp file).
- **Start (enable):** pre-cleans any route already running under `routeKey` (`removePollingRoute`) before `addRoutes`+`startRoute` — so even a Start *without* a preceding Stop replaces the old poller rather than adding a second one.

Before the fix the key was `endpoint.getValue()`, so after an endpoint change Stop recomputed the *new* value and stopped nothing → both the old and new SFTP servers kept polling (your UAT observation).

Covered by the integration test `InboundSftpIntegrationTest.endpointChanged_previousPollerTornDown_onlyNewPollerRemains` (RED before the fix: `expected 1 but was 2` — both pollers alive; green after). Multiple SFTP children still get one poller each (distinct child ids ⇒ distinct keys).

backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/process/InvokeScriptedImportConversionAction.java
```diff
@@ -116,6 +119,34 @@ protected String doIt() throws Exception
...
+	 * enable/disable commands pass the CONCRETE transport command here — so accept both.
+	 */
+	@VisibleForTesting
+	static ScriptedImportConversionIntent resolveIntent(@NonNull final String externalRequest)
+	{
+		try
+		{
+			return ScriptedImportConversionIntent.ofCode(externalRequest);
+		}
+		catch (final AdempiereException intentNotFound)
```
metas-ts: not sure it's cool to catch an unspecific Adempiere exception. 
maybe better have a method `ScriptedImportConversionIntent.ofCodeOrNull()` ? and then have proper ifs..or maybe work with optionals if you like.
my point is: that resolveIntent get a command is *not* an exception as we now know

backend/de.metas.externalsystem/src/main/sql/postgresql/system/80-de.metas.externalsystem/5815620_ScriptIdentifier_description_stem_without_js.sql
```diff
@@ -0,0 +1,51 @@
...
+    Updated = TO_TIMESTAMP('2026-07-22 10:00:03', 'YYYY-MM-DD HH24:MI:SS'), UpdatedBy = 100
+WHERE AD_Element_ID = 584103;
+
+-- 5) Propagate the corrected element text to AD_Column(_Trl) and AD_Field(_Trl) for all languages
+SELECT update_TRL_Tables_On_AD_Element_TRL_Update(584103);
+
+-- 6) AD_UI_Element has no _Trl companion and is not covered by propagation -> correct directly.
+--    All four UI elements render the ScriptIdentifier field (export tabs + import tabs).
+UPDATE AD_UI_Element
+SET Description = 'Kennung des JavaScript-Konvertierungsskripts: der Dateiname ohne die Endung .js (z. B. eddyson2m_orders für die Datei eddyson2m_orders.js).',
```
metas-ts:
```suggestion
SET Description = 'Kennung des JavaScript-Konvertierungsskripts: der Dateiname ohne die Endung .js (z. B. edifact2olcand für die Datei edifact2olcand.js).',
```

backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/process/InvokeScriptedImportConversionAction.java
metas-ts: **Resolution (dev):** Agreed — catching an unspecific `AdempiereException` for control flow is wrong (a command code here is expected, not exceptional, and the broad catch could swallow unrelated `AdempiereException`s). Refactoring to add `ScriptedImportConversionIntent.ofCodeOrNull` (+ symmetric `ScriptedImportConversionCommand.ofCodeOrNull`) and rewriting `resolveIntent` with plain ifs — intent code → intent; else command code → its intent; else throw a clear "unknown request code" error (that case *is* a genuine error). No try/catch. Commit landing on this branch shortly; will confirm the SHA here.

backend/de.metas.externalsystem/src/main/sql/postgresql/system/80-de.metas.externalsystem/5815620_ScriptIdentifier_description_stem_without_js.sql
metas-ts: **Resolution (dev):** Adopting — switching the example to the neutral `edifact2olcand` (dropping the customer-specific `eddyson2m_orders`), applied consistently across **all** the description strings in the migration (de_DE/de_CH, en_US, fr_CH/it_CH, the base `AD_Element`, and the `AD_UI_Element` rows), not just the one line — so the wording is uniform everywhere it renders. Commit landing on this branch shortly; will confirm the SHA here.

misc/services/camel/de-metas-camel-externalsystems/de-metas-camel-scriptedadapter/src/main/java/de/metas/camel/externalsystems/scriptedadapter/convertmsg/to_mf/ScriptedImportConversionSftpRouteBuilder.java
```diff
@@ -180,7 +190,9 @@ private void enableSftpPollingProcessor(@NonNull final Exchange exchange) throws
 			final FileAttribute<Set<PosixFilePermission>> attr = PosixFilePermissions.asFileAttribute(ownerOnly);
 			final Path tempKeyFile = Files.createTempFile("sftp_key_" + endpointName + "_", ".pem", attr);
```
metas-ts: please research if this is the proper way and best practice to handle the key

backend/de.metas.externalsystem/src/main/java/de/metas/externalsystem/process/InvokeScriptedImportConversionAction.java
metas-ts: **Resolution (dev):** Done — https://github.com/metasfresh/metasfresh/commit/a0b53ff14b0. Added `ScriptedImportConversionIntent.ofCodeOrNull` (+ symmetric `ScriptedImportConversionCommand.ofCodeOrNull`) and rewrote `resolveIntent` with plain ifs: intent code → intent; else command code → `getIntent()`; else throw a clear "unknown External_Request" error (the only genuinely-exceptional case). No `try/catch (AdempiereException)`. `ofCode` is reimplemented on top of `ofCodeOrNull` so the matching logic isn't duplicated. de.metas.externalsystem suite green (156).

backend/de.metas.externalsystem/src/main/sql/postgresql/system/80-de.metas.externalsystem/5815620_ScriptIdentifier_description_stem_without_js.sql
metas-ts: **Resolution (dev):** Done — https://github.com/metasfresh/metasfresh/commit/b0842ea10ff. Applied `edifact2olcand` (dropping the customer-specific `eddyson2m_orders`) to every occurrence in the migration — all 5 `Description` strings (de_DE/de_CH, en_US, fr_CH/it_CH, base `AD_Element`, `AD_UI_Element`) plus the header comment — so the example is uniform and neutral.

misc/services/camel/de-metas-camel-externalsystems/de-metas-camel-scriptedadapter/src/main/java/de/metas/camel/externalsystems/scriptedadapter/convertmsg/to_mf/ScriptedImportConversionSftpRouteBuilder.java
metas-ts: **Resolution (dev):** Researched against the Apache Camel docs (camel-ftp `sftp:` component + camel's key-auth guidance). Findings:

- The SFTP endpoint accepts the private key three ways, in priority order: **`privateKey`** (a `byte[]`, in-memory, highest priority — bound in the registry and referenced as `&privateKey=#bean:<id>`), **`privateKeyUri`** (`classpath:`/`file:`), and **`privateKeyFile`** (a filesystem path — lowest). Camel's own best-practice example loads the key material from a secret manager and passes it as **bytes in memory**, never to disk.
- So the current temp-`.pem` approach is **not** the recommended way. It is *mitigated* — `0600` (owner-only) perms and the key file is deleted on disable/replace (rekeyed on the stable route key by the B3 fix) — but it still writes the plaintext key to the container filesystem for the route's lifetime and can leave a stray `.pem` if the process is killed mid-run.

**Recommendation:** switch to the in-memory `privateKey` (`byte[]`) option — bind the key bytes in the camel registry under the stable route key and reference them via `&privateKey=#bean:<routeKey>`, removing the bean on disable/replace (same lifecycle the temp file has now). That eliminates the on-disk key entirely.

It's a security-sensitive change to the SFTP-auth path and would want a key-auth integration-test case, so it's a bit meatier than the other two fixes. **Do you want it folded into this PR, or tracked as a separate hardening follow-up?** I'm happy to implement it here if you prefer.

PR #25297, opened 2026-07-22
TU package-dimension calculation + nShift mandatory-field validation

mrie04:
## What

Implements Transport-Unit package-dimension calculation and nShift mandatory-field validation.

**Part A — TU dimension calculation**
- New `PackageDimensionCalcMethod` ref-list + enum (Strapping / Repacking / Nesting).
- New `M_HU_PI_Version.PackageDimensionCalcMethod` column (shown on the PI-version tab, TU-only via display-logic).
- Calc math on `PackageDimensions.ofItems(mode, items)` for a mixed TU (stacking / volume-repack / nesting).
- `@ModelChange` interceptor confining the calc method to TU pi-versions.
- `HUPackageBL.getPackageDimensions` now aggregates a multi-item TU via the selected mode and no longer gates dimensions on `IsSelfPacked`; the `SelfPackedProductWithNoDefinedSizes` throw is removed (missing dims now yield `UNSPECIFIED`, validated at nShift build — see Part B).
- 3 mode packing-instructions shipped inactive in core (activated per-customer in the customer repo).
- Dedicated cucumber feature `HUPackage_DimensionCalculation.feature` (5 scenarios: all modes + TU guard + non-self-packed).

**Part B — nShift mandatory-field validation**
- `NShiftUtil` validates sender + receiver Phone / Email / Attention and parcel dimensions at request build, across all four nShift build paths, failing early with a message naming the missing field.

## Testing
- Unit: `PackageDimensionsTest`, `HUPackageBL_DimensionCalcTest`, `M_HU_PI_VersionInterceptorTest`, `NShiftMandatoryFieldValidationTest` — all green (TDD, RED observed first).
- Cucumber: `HUPackage_DimensionCalculation.feature` — 5/5 green against a `deep_tundra_release` stack.

## Notes
- Per-customer activation of the 3 mode-PIs ships in the customer-repo PR (separate).
- Config windows verified reachable in the menu tree (no AD_Menu change needed).

### DIFF
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/hu/M_HU_PI_Version_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/hu/M_HU_PI_Version_StepDef.java
@@ -23,6 +23,7 @@
 package de.metas.cucumber.stepdefs.hu;
 
 import de.metas.common.util.CoalesceUtil;
+import de.metas.cucumber.stepdefs.DataTableRow;
 import de.metas.cucumber.stepdefs.DataTableRows;
 import de.metas.cucumber.stepdefs.StepDefDataIdentifier;
 import de.metas.handlingunits.model.I_M_HU_PI;
@@ -30,8 +31,11 @@ import de.metas.handlingunits.model.I_M_HU_PI_Version;
 import de.metas.util.Services;
 import io.cucumber.datatable.DataTable;
 import io.cucumber.java.en.And;
+import io.cucumber.java.en.Then;
 import lombok.NonNull;
+import javax.annotation.Nullable;
 import org.adempiere.ad.dao.IQueryBL;
+import org.adempiere.exceptions.AdempiereException;
 import org.adempiere.model.InterfaceWrapperHelper;
 
 import static de.metas.handlingunits.model.I_M_HU_PI_Version.COLUMNNAME_HU_UnitType;
@@ -40,6 +44,7 @@ import static de.metas.handlingunits.model.I_M_HU_PI_Version.COLUMNNAME_IsCurren
 import static de.metas.handlingunits.model.I_M_HU_PI_Version.COLUMNNAME_M_HU_PI_ID;
 import static de.metas.handlingunits.model.I_M_HU_PI_Version.COLUMNNAME_M_HU_PI_Version_ID;
 import static de.metas.handlingunits.model.I_M_HU_PI_Version.COLUMNNAME_Name;
+import static de.metas.handlingunits.model.I_M_HU_PI_Version.COLUMNNAME_PackageDimensionCalcMethod;
 import static org.adempiere.model.InterfaceWrapperHelper.saveRecord;
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -51,6 +56,10 @@ public class M_HU_PI_Version_StepDef
 	private final M_HU_PI_Version_StepDefData huPiVersionTable;
 	private final M_HU_PackagingCode_StepDefData huPackagingCodeTable;
 
+	/** Captures the last exception thrown by {@link #add_M_HU_PI_Version_expectingError}. */
+	@Nullable
+	private AdempiereException lastSaveException;
+
 	public M_HU_PI_Version_StepDef(
 			@NonNull final M_HU_PI_StepDefData huPiTable,
 			@NonNull final M_HU_PI_Version_StepDefData huPiVersionTable,
@@ -61,50 +70,129 @@ public class M_HU_PI_Version_StepDef
 		this.huPackagingCodeTable = huPackagingCodeTable;
 	}
 
+	/**
+	 * Creates or upserts {@code M_HU_PI_Version} records.
+	 *
+	 * @cucumber.stepdef
+	 * @cucumber.columns
+	 *   <b>M_HU_PI_Version_ID</b> — (required) alias for cross-step reference<br>
+	 *   <b>M_HU_PI_ID</b> — (required, identifier-ref) parent packing instruction<br>
+	 *   <b>HU_UnitType</b> — (required) TU, LU, or V<br>
+	 *   <b>IsCurrent</b> — (optional) default true<br>
+	 *   <b>IsActive</b> — (optional) default true<br>
+	 *   <b>M_HU_PackagingCode_ID</b> — (optional, identifier-ref) packaging code<br>
+	 *   <b>PackageDimensionCalcMethod</b> — (optional) S (Strapping), R (Repacking), N (Nesting); only valid on TU versions<br>
+	 * @cucumber.example
+	 * <pre>
+	 * And metasfresh contains M_HU_PI_Version:
+	 *   | M_HU_PI_Version_ID | M_HU_PI_ID | HU_UnitType | IsCurrent | PackageDimensionCalcMethod |
+	 *   | tuVersion          | tuPi       | TU          | Y         | S                          |
+	 * </pre>
+	 */
 	@And("metasfresh contains M_HU_PI_Version:")
 	public void add_M_HU_PI_Version(@NonNull final DataTable dataTable)
 	{
 		DataTableRows.of(dataTable)
 				.setAdditionalRowIdentifierColumnName(COLUMNNAME_M_HU_PI_Version_ID)
-				.forEach(row -> {
-					final I_M_HU_PI huPi = row.getAsIdentifier(COLUMNNAME_M_HU_PI_ID).lookupNotNullIn(huPiTable);
-					final String name = row.suggestValueAndName(null, huPi::getName).getName();
-					final String huUnitType = row.getAsString(COLUMNNAME_HU_UnitType); //dev-note: HU_UNITTYPE_AD_Reference_ID=540472;
-					final boolean isCurrent = row.getAsOptionalBoolean(COLUMNNAME_IsCurrent).orElseTrue();
-					final boolean active = row.getAsOptionalBoolean(COLUMNNAME_IsActive).orElseTrue();
+				.forEach(row -> buildAndSavePiVersion(row, huPiVersionTable));
+	}
 
-					final I_M_HU_PI_Version existingPiVersion = queryBL.createQueryBuilder(I_M_HU_PI_Version.class)
-							.addEqualsFilter(COLUMNNAME_M_HU_PI_ID, huPi.getM_HU_PI_ID())
-							.addStringLikeFilter(COLUMNNAME_Name, name, true)
-							.addEqualsFilter(COLUMNNAME_HU_UnitType, huUnitType)
-							.addEqualsFilter(COLUMNNAME_IsActive, active)
-							.create()
-							.firstOnly(I_M_HU_PI_Version.class);
+	/**
+	 * Variant of {@link #add_M_HU_PI_Version} that captures the first {@link AdempiereException}
+	 * thrown during save (e.g. when {@code PackageDimensionCalcMethod} is set on a non-TU version)
+	 * instead of propagating it. The captured exception can be asserted via
+	 * {@link #assertLastSaveExceptionWasThrown()}.
+	 *
+	 * @cucumber.stepdef
+	 * @cucumber.example
+	 * <pre>
+	 * When metasfresh contains M_HU_PI_Version expecting error:
+	 *   | M_HU_PI_Version_ID | M_HU_PI_ID | HU_UnitType | PackageDimensionCalcMethod |
+	 *   | luVersion          | luPi       | LU          | S                          |
+	 * Then an AdempiereException was thrown when saving the M_HU_PI_Version
+	 * </pre>
+	 */
+	@And("metasfresh contains M_HU_PI_Version expecting error:")
+	public void add_M_HU_PI_Version_expectingError(@NonNull final DataTable dataTable)
+	{
+		lastSaveException = null;
+		try
+		{
+			DataTableRows.of(dataTable)
+					.setAdditionalRowIdentifierColumnName(COLUMNNAME_M_HU_PI_Version_ID)
+					.forEach(row -> buildAndSavePiVersion(row, null));
+		}
+		catch (final AdempiereException e)
+		{
+			lastSaveException = e;
+		}
+	}
 
-					final I_M_HU_PI_Version piVersion = CoalesceUtil.coalesceSuppliers(
-							() -> existingPiVersion,
-							() -> InterfaceWrapperHelper.newInstanceOutOfTrx(I_M_HU_PI_Version.class)
-					);
-					assertThat(piVersion).isNotNull();
+	/**
+	 * Asserts that the most recent {@link #add_M_HU_PI_Version_expectingError} step did throw an {@link AdempiereException}.
+	 *
+	 * @cucumber.stepdef
+	 * @cucumber.example
+	 * <pre>
+	 * Then an AdempiereException was thrown when saving the M_HU_PI_Version
+	 * </pre>
+	 */
+	@Then("an AdempiereException was thrown when saving the M_HU_PI_Version")
+	public void assertLastSaveExceptionWasThrown()
+	{
+		assertThat(lastSaveException)
+				.as("Expected an AdempiereException to be thrown when saving M_HU_PI_Version, but none was thrown")
+				.isNotNull();
+	}
 
-					piVersion.setM_HU_PI_ID(huPi.getM_HU_PI_ID());
-					piVersion.setName(name);
-					piVersion.setHU_UnitType(huUnitType);
-					piVersion.setIsCurrent(isCurrent);
-					piVersion.setIsActive(active);
+	private void buildAndSavePiVersion(
+			@NonNull final DataTableRow row,
+			@Nullable final M_HU_PI_Version_StepDefData versionTableToStore)
+	{
+		final I_M_HU_PI huPi = row.getAsIdentifier(COLUMNNAME_M_HU_PI_ID).lookupNotNullIn(huPiTable);
+		final String name = row.suggestValueAndName(null, huPi::getName).getName();
+		final String huUnitType = row.getAsString(COLUMNNAME_HU_UnitType); //dev-note: HU_UNITTYPE_AD_Reference_ID=540472;
+		final boolean isCurrent = row.getAsOptionalBoolean(COLUMNNAME_IsCurrent).orElseTrue();
+		final boolean active = row.getAsOptionalBoolean(COLUMNNAME_IsActive).orElseTrue();
 
-					row.getAsOptionalIdentifier(I_M_HU_PI_Version.COLUMNNAME_M_HU_PackagingCode_ID)
-							.ifPresent(huPackagingCodeIdentifier -> {
-								final int huPackagingCodeId = huPackagingCodeIdentifier.isNullPlaceholder()
-										? -1
-										: huPackagingCodeTable.get(huPackagingCodeIdentifier).getM_HU_PackagingCode_ID();
+		final I_M_HU_PI_Version existingPiVersion = queryBL.createQueryBuilder(I_M_HU_PI_Version.class)
+				.addEqualsFilter(COLUMNNAME_M_HU_PI_ID, huPi.getM_HU_PI_ID())
+				.addStringLikeFilter(COLUMNNAME_Name, name, true)
+				.addEqualsFilter(COLUMNNAME_HU_UnitType, huUnitType)
+				.addEqualsFilter(COLUMNNAME_IsActive, active)
+				.create()
+				.firstOnly(I_M_HU_PI_Version.class);
 
-								piVersion.setM_HU_PackagingCode_ID(huPackagingCodeId);
-							});
+		final I_M_HU_PI_Version piVersion = CoalesceUtil.coalesceSuppliers(
+				() -> existingPiVersion,
+				() -> InterfaceWrapperHelper.newInstanceOutOfTrx(I_M_HU_PI_Version.class)
+		);
+		assertThat(piVersion).isNotNull();
 
-					saveRecord(piVersion);
-					huPiVersionTable.put(row.getAsIdentifier(), piVersion);
+		piVersion.setM_HU_PI_ID(huPi.getM_HU_PI_ID());
+		piVersion.setName(name);
+		piVersion.setHU_UnitType(huUnitType);
+		piVersion.setIsCurrent(isCurrent);
+		piVersion.setIsActive(active);
+
+		row.getAsOptionalIdentifier(I_M_HU_PI_Version.COLUMNNAME_M_HU_PackagingCode_ID)
+				.ifPresent(huPackagingCodeIdentifier -> {
+					final int huPackagingCodeId = huPackagingCodeIdentifier.isNullPlaceholder()
+							? -1
+							: huPackagingCodeTable.get(huPackagingCodeIdentifier).getM_HU_PackagingCode_ID();
+
+					piVersion.setM_HU_PackagingCode_ID(huPackagingCodeId);
 				});
+
+		row.getAsOptionalString(COLUMNNAME_PackageDimensionCalcMethod)
+				.ifPresent(piVersion::setPackageDimensionCalcMethod);
+
+		saveRecord(piVersion);
+
+		if (versionTableToStore != null)
+		{
+			versionTableToStore.putOrReplace(row.getAsIdentifier(), piVersion);
+		}
 	}
 
 	@And("load M_HU_PI_Version:")
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/hu/M_HU_PI_Version_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/hu/M_HU_PI_Version_StepDef.java
@@ -143,8 +143,11 @@ public class M_HU_PI_Version_StepDef
 	{
 		assertThat(lastSaveException)
 				.as("Expected the TU-only guard AdempiereException when saving M_HU_PI_Version, but none was thrown")
-				.isNotNull()
-				.hasMessageContaining("can only be set on Transport Unit");
+				.isNotNull();
+		// Assert the language-independent errorCode (the message text is translated: cucumber runs de_DE).
+		assertThat(lastSaveException.getErrorCode())
+				.as("Expected the TU-only guard error code")
+				.isEqualTo("M_HU_PI_Version_CalcMethodOnlyOnTU");
 	}
 
 	private void buildAndSavePiVersion(
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/hu/M_HU_PI_Version_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/hu/M_HU_PI_Version_StepDef.java
@@ -28,6 +28,7 @@ import de.metas.cucumber.stepdefs.DataTableRows;
 import de.metas.cucumber.stepdefs.StepDefDataIdentifier;
 import de.metas.handlingunits.model.I_M_HU_PI;
 import de.metas.handlingunits.model.I_M_HU_PI_Version;
+import de.metas.product.PackageDimensionCalcMethod;
 import de.metas.util.Services;
 import io.cucumber.datatable.DataTable;
 import io.cucumber.java.en.And;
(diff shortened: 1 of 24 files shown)


backend/de.metas.shipper.client.nshift/src/main/java/de/metas/shipper/client/nshift/NShiftUtil.java
```diff
@@ -109,11 +109,28 @@ public static JsonAddress buildAddressWithAttentionFromMappings(
...
+		}
+
+		if (contact == null || Check.isBlank(contact.getPhone()))
+		{
+			throw new IllegalStateException(role + " Phone is mandatory but is missing or blank.");
+		}
+		if (contact.getEmailAddress() == null || Check.isBlank(contact.getEmailAddress()))
+		{
+			throw new IllegalStateException(role + " Email is mandatory but is missing or blank.");
+		}
```
mrie04: Why not de.metas.util.Check#assumeNotEmpty(java.lang.String, java.lang.String, java.lang.Object...) ?

backend/de.metas.handlingunits.base/src/main/sql/postgresql/system/70-de.metas.handlingunits/5815840_sys_gh30361_PackageDimensions_CheckIsSelfPacked_sysconfig.sql
```diff
@@ -0,0 +1,24 @@
...
+    (AD_SysConfig_ID, AD_Client_ID, AD_Org_ID,
+     Name, Value, Description,
+     EntityType, ConfigurationLevel,
+     IsActive,
+     Created, CreatedBy, Updated, UpdatedBy)
+VALUES
+    (541837 /*From ID Server*/, 0, 0,
+     'de.metas.handlingunits.PackageDimensions.CheckIsSelfPacked',
+     'N',
+     'When Y, the legacy IsSelfPacked gate is active: non-self-packed products return PackageDimensions.UNSPECIFIED instead of using their named dimensions. Default N = flag-independent behaviour.',
```
mrie04: timeless wording pls, no history or legacy assumptions

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipping/impl/HUPackageBL.java
```diff
@@ -230,24 +237,20 @@ public List<I_M_Package> createM_Packages(@NonNull final CreatePackageForHUReque
...
-	 * {@code getPackageDimensions}.
+	 * Single-unit dimensions: each parcel carries the product's named dimensions verbatim
+	 * (no qty-based sort/scale). Returns {@link PackageDimensions#UNSPECIFIED} when the product
+	 * has no dims. When {@value IHUPackageBL#SYSCONFIG_CHECK_IS_SELF_PACKED}='Y', also returns
+	 * {@link PackageDimensions#UNSPECIFIED} for non-self-packed products (legacy gate).
 	 */
 	private PackageDimensions resolveSingleUnitDimensions(@NonNull final Product product)
 	{
-		if (!product.isSelfPacked())
+		final boolean checkSelfPacked = sysConfigBL.getBooleanValue(SYSCONFIG_CHECK_IS_SELF_PACKED, false);
```
mrie04: consider moving into own function isCheckSelfPacked

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/hu/M_HU_PI_Version_StepDef.java
```diff
@@ -61,50 +70,129 @@ public M_HU_PI_Version_StepDef(
...
+	 * Asserts that the most recent {@link #add_M_HU_PI_Version_expectingError} step did throw an {@link AdempiereException}.
+	 *
+	 * @cucumber.stepdef
+	 * @cucumber.example
+	 * <pre>
+	 * Then an AdempiereException was thrown when saving the M_HU_PI_Version
+	 * </pre>
+	 */
+	@Then("an AdempiereException was thrown when saving the M_HU_PI_Version")
+	public void assertLastSaveExceptionWasThrown()
```
mrie04: why as seperate step? and why not check the errorCode like we do it in other places?

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/hu/M_HU_PI_Version_StepDef.java
```diff
@@ -61,50 +70,129 @@ public M_HU_PI_Version_StepDef(
...
+		row.getAsOptionalIdentifier(I_M_HU_PI_Version.COLUMNNAME_M_HU_PackagingCode_ID)
+				.ifPresent(huPackagingCodeIdentifier -> {
+					final int huPackagingCodeId = huPackagingCodeIdentifier.isNullPlaceholder()
+							? -1
+							: huPackagingCodeTable.get(huPackagingCodeIdentifier).getM_HU_PackagingCode_ID();
+
+					piVersion.setM_HU_PackagingCode_ID(huPackagingCodeId);
 				});
+
+		row.getAsOptionalString(COLUMNNAME_PackageDimensionCalcMethod)
```
mrie04: why not go via getAsOptionalEnum to make sure the value is valid?

backend/de.metas.shipper.client.nshift/src/main/java/de/metas/shipper/client/nshift/NShiftUtil.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/4e4087f4b890924a2657537ac0d52398a6768aa3. Switched the sender/receiver Attention/Phone/Email checks to Check.assumeNotEmpty(value, IllegalStateException.class, "<field> is mandatory ..."). Using the explicit-exception-class overload keeps the plain IllegalStateException, so the module stays decoupled from AdempiereException (not on this module's classpath) and the type-asserting tests stay green. NShiftMandatoryFieldValidationTest 20/20 pass.

backend/de.metas.handlingunits.base/src/main/sql/postgresql/system/70-de.metas.handlingunits/5815840_sys_gh30361_PackageDimensions_CheckIsSelfPacked_sysconfig.sql
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/4e4087f4b890924a2657537ac0d52398a6768aa3. Reworded the SysConfig Description (and the header comment) to timeless wording — dropped 'legacy'/'new default'/'restore' framing. Now: 'Controls the IsSelfPacked dimension gate. When Y, a non-self-packed product yields PackageDimensions.UNSPECIFIED; when N, product dimensions are used regardless of the self-packed flag.'

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipping/impl/HUPackageBL.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/4e4087f4b890924a2657537ac0d52398a6768aa3. Extracted isCheckSelfPacked() in HUPackageBL and used it at both dimension-resolution sites (getPackageDimensions single-product branch + resolveSingleUnitDimensions), removing the duplicated getBooleanValue local. HUPackageBL_DimensionCalcTest green.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipper/HUPackage_DimensionCalculation.feature
```diff
@@ -0,0 +1,321 @@
...
+      | cu_strapping_b | tu_strapping | 2   |
+    And metasfresh contains C_Orders:
+      | Identifier        | IsSOTrx | C_BPartner_ID    | DateOrdered | M_Warehouse_ID | M_Shipper_ID | AD_User_ID          |
+      | so_strapping      | true    | dimcalc_customer | 2022-12-12  | dimcalc_wh     | dimcalc_ship | dimcalc_custContact |
+    And metasfresh contains C_OrderLines:
+      | Identifier        | C_Order_ID    | M_Product_ID   | QtyEntered |
+      | so_strapping_l_a  | so_strapping  | dimcalc_prod_a | 3          |
+      | so_strapping_l_b  | so_strapping  | dimcalc_prod_b | 2          |
+    When the order identified by so_strapping is completed
+    And after not more than 60s, M_ShipmentSchedules are found:
```
mrie04, line 121: pls also validate carrier product to prevent flakiness, same on other scenarios

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipper/HUPackage_DimensionCalculation.feature
```diff
@@ -0,0 +1,321 @@
...
+  Scenario: LU version guard — PackageDimensionCalcMethod is rejected on non-TU versions
+    # The M_HU_PI_Version interceptor throws an AdempiereException when the calc method
+    # is set on an LU version. Only TU versions may carry a PackageDimensionCalcMethod.
+    When metasfresh contains M_HU_PI_Version expecting error:
+      | M_HU_PI_Version_ID | M_HU_PI_ID    | HU_UnitType | IsCurrent | PackageDimensionCalcMethod |
+      | dimcalc_LU_Version | dimcalc_LU_PI | LU          | Y         | S                          |
+    Then an AdempiereException was thrown when saving the M_HU_PI_Version
+
+  @Id:S30361_TC5
+  Scenario: Non-self-packed single-product TU — product dims contribute regardless of IsSelfPacked
```
mrie04: this the tests before already test automatically. Alternative activate sysCfg and test this case. (Reset of sysCfg after) or leave it out

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/hu/M_HU_PI_Version_StepDef.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/3a13aa2dcf0781934bfef4b6e5d5a2a19131da5b. Switched to row.getAsOptionalEnum(COLUMNNAME_PackageDimensionCalcMethod, PackageDimensionCalcMethod.class) then setPackageDimensionCalcMethod(calcMethod.getCode()) — this validates the DataTable value is a real S/R/N calc method at parse time (fail-fast on a typo) instead of storing an arbitrary string, matching the established pattern (e.g. C_BPartner_StepDef with PaymentRule/InvoiceRule). Cucumber module test-compiles on JDK8.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/hu/M_HU_PI_Version_StepDef.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/3a13aa2dcf0781934bfef4b6e5d5a2a19131da5b. (b) The assertion now checks the SPECIFIC error: assertThat(lastSaveException).isNotNull().hasMessageContaining("can only be set on Transport Unit") — so it verifies the TU-only guard actually fired, not merely that some exception was thrown. On a true error CODE: the T4 interceptor throws a raw-message AdempiereException with no AdMessageKey (the raw-message choice we made for that guard, matching the sibling M_HU_PI_Item_Product interceptor), so there is no code to assert against; if you'd prefer a real AD_Message code here I'll add an AdMessageKey to the guard + an AD_Message migration — just say so. (a) The two steps are a given/then split — the 'expecting error' given performs the save and captures the exception, the then asserts it; happy to fold them into a single step if you prefer that.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/model/validator/M_HU_PI_Version.java
```diff
@@ -24,23 +24,48 @@
...
+	{
+		final PackageDimensionCalcMethod calcMethod = PackageDimensionCalcMethod.ofNullableCode(piVersion.getPackageDimensionCalcMethod());
+		if (calcMethod == null)
+		{
+			return;
+		}
+		final HuUnitType huUnitType = HuUnitType.ofNullableCode(piVersion.getHU_UnitType());
+		if (huUnitType == null || !huUnitType.isTU())
+		{
+			throw new AdempiereException("PackageDimensionCalcMethod can only be set on Transport Unit (TU) packing instruction versions")
```
mrie04: why no translated adMsgKey. then it's automatically marked as userfriendly as well

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipping/impl/HUPackageBL.java
```diff
@@ -230,24 +237,20 @@ public List<I_M_Package> createM_Packages(@NonNull final CreatePackageForHUReque
...
-		else if (product.getPackageDimensions().isUnspecified())
-		{
-			throw new AdempiereException(MSG_SELF_PACKED_PRODUCT_WITH_NO_DEFINED_SIZES, product.getValue());
-		}
-		else
-		{
-			return product.getPackageDimensions();
-		}
+		final PackageDimensions dims = product.getPackageDimensions();
+		return dims.isUnspecified() ? PackageDimensions.UNSPECIFIED : dims;
```
mrie04: what is the benefit of this? if dims.isUnspecified() -> PackageDimensions.UNSPECIFIED == dims

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipper/HUPackage_DimensionCalculation.feature
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/dcc97c42950. Added Carrier_Product_ID + Carrier_Goods_Type_ID (resolved to dimcalc_cp1 / dimcalc_cgt1 — the only carrier product/goods-type configured in the Background) to the 'M_ShipmentSchedules are found' wait in all three mode scenarios (Strapping / Repacking / Nesting), so each now waits for the async carrier advise to land before asserting the M_Packages — same anti-flakiness pattern as delayShipmentNotificationUntilCarrierTrackingReady.feature. TC4 (LU guard) has no shipment flow so it is unaffected. Note: I couldn't re-run cucumber locally (the release stack I'd used was torn down), so CI (executor4) is the verification for this.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/shipper/HUPackage_DimensionCalculation.feature
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/dcc97c42950 — removed the non-self-packed scenario (TC5). Coverage is not lost: the three mode scenarios already exercise dimensions contributing from default IsSelfPacked=N products, and the single-product path plus the SysConfig gate (CheckIsSelfPacked=Y -> UNSPECIFIED, with reset) are both covered by the HUPackageBL_DimensionCalcTest unit tests. I chose 'leave it out' over an E2E sysCfg-gate scenario because that would just duplicate the unit coverage; happy to add the E2E gate scenario instead if you'd prefer it end-to-end.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipping/IHUPackageBL.java
```diff
@@ -17,6 +17,13 @@
 
 public interface IHUPackageBL extends ISingletonService
 {
+	/**
+	 * SysConfig (system level, default {@code N}): when {@code Y}, package-dimension resolution re-applies
+	 * the legacy {@code IsSelfPacked} gate (a non-self-packed product yields {@link PackageDimensions#UNSPECIFIED}).
+	 * Default {@code N} = product dimensions are used regardless of the flag.
+	 */
+	String SYSCONFIG_CHECK_IS_SELF_PACKED = "de.metas.handlingunits.PackageDimensions.CheckIsSelfPacked";
```
mrie04: why on interface?

backend/de.metas.handlingunits.base/src/main/sql/postgresql/system/70-de.metas.handlingunits/5815420_sys_M_HU_PI_Version_PackageDimensionCalcMethod_column.sql
```diff
@@ -0,0 +1,229 @@
...
+-- ============================================================
+-- 3. Physical DDL — add column to M_HU_PI_Version
+-- ============================================================
+-- New nullable column (no default). Routed through db_alter_table() so the function manages
+-- view dependencies (the bare-ALTER exception is only for NOT NULL DEFAULT backfill columns).
+SELECT public.db_alter_table('M_HU_PI_Version','ALTER TABLE public.M_HU_PI_Version ADD COLUMN IF NOT EXISTS PackageDimensionCalcMethod CHAR(1)');
+
+-- Check constraint to enforce the ref-list values (also via db_alter_table for view-dependency safety)
+SELECT public.db_alter_table('M_HU_PI_Version','ALTER TABLE public.M_HU_PI_Version DROP CONSTRAINT IF EXISTS PackageDimensionCalcMethod_Check');
+SELECT public.db_alter_table('M_HU_PI_Version','ALTER TABLE public.M_HU_PI_Version ADD CONSTRAINT PackageDimensionCalcMethod_Check CHECK (PackageDimensionCalcMethod IN (''S'', ''R'', ''N''))');
```
mrie04: only reflist values and null are selectable. this just makes it harder to extend imho

backend/de.metas.handlingunits.base/src/main/sql/postgresql/system/70-de.metas.handlingunits/5815420_sys_M_HU_PI_Version_PackageDimensionCalcMethod_column.sql
```diff
@@ -0,0 +1,229 @@
...
+--   AD_Column     592977 /*From ID Server*/  (M_HU_PI_Version.PackageDimensionCalcMethod)
+--   AD_Field      781766 /*From ID Server*/  (tab 540823, win 540344)
+--   AD_Field      781767 /*From ID Server*/  (tab 540505, win 540188)
+--   AD_UI_Element 652698 /*From ID Server*/  (tab 540823, win 540344)
+-- Migration prefix: 5815420
+
+-- ============================================================
+-- 1. AD_Element
+-- ============================================================
+INSERT INTO AD_Element
```
mrie04, line 16: add description otherwise it isn't clear calc method for what?

backend/de.metas.handlingunits.base/src/main/sql/postgresql/system/70-de.metas.handlingunits/5815630_sys_gh30361_DimensionCalc_mode_PIs_inactive.sql
```diff
@@ -0,0 +1,219 @@
...
+--   M_HU_PI               540013 (Strapping), 540014 (Repacking), 540015 (Nesting)
+--   M_HU_PI_Version        540013 (Strapping), 540014 (Repacking), 540015 (Nesting)
+--   M_HU_PI_Item           540018 (Strapping), 540019 (Repacking), 540020 (Nesting)
+--   M_HU_PI_Item_Product   540023 (Strapping), 540021 (Repacking), 540022 (Nesting)
+
+-- ============================================================
+-- STRAPPING mode (S)
+-- ============================================================
+
+INSERT INTO M_HU_PI
```
mrie04, line 15: pls use german names for everything here

backend/de.metas.picking.rest-api/src/main/java/de/metas/picking/workflow/PackedHUCarrierAdviseService.java
```diff
@@ -440,9 +444,19 @@ JsonDeliveryAdvisorRequestParcel buildRequestParcel(
 					.map(weight -> uomConversionBL.convertToKilogram(weight, product.getId()))
 					.map(Quantity::getAsBigDecimal)
 					.orElse(BigDecimal.ZERO);
-			dimensions = product.isSelfPacked()
-					? product.getPackageDimensions()
-					: PackageDimensions.UNSPECIFIED;
+			// Use product dims (IsSelfPacked gate is SysConfig-controlled, default off).
+			final boolean checkSelfPacked = sysConfigBL.getBooleanValue(IHUPackageBL.SYSCONFIG_CHECK_IS_SELF_PACKED, false);
```
mrie04: also into function as in other place

backend/de.metas.shipper.client.nshift/src/main/java/de/metas/shipper/client/nshift/NShiftShipmentService.java
```diff
@@ -225,9 +226,14 @@ private static JsonLine buildNShiftLine(@NonNull final JsonDeliveryOrderParcel d
 	{
 		// nShift expects weight in grams and dimensions in millimeters.
 		final int weightGrams = deliveryLine.getGrossWeightKg().multiply(BigDecimal.valueOf(1000)).intValue();
-		final int lengthMM = deliveryLine.getPackageDimensions().getLengthInCM() * 10;
-		final int widthMM = deliveryLine.getPackageDimensions().getWidthInCM() * 10;
-		final int heightMM = deliveryLine.getPackageDimensions().getHeightInCM() * 10;
+		final JsonPackageDimensions dims = deliveryLine.getPackageDimensions();
+		if (dims.getLengthInCM() == 0 && dims.getWidthInCM() == 0 && dims.getHeightInCM() == 0)
```
mrie04: why not also on unspecified then?

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/model/validator/M_HU_PI_Version.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/5af4b8af1ac. The guard now throws new AdempiereException(AdMessageKey.of("M_HU_PI_Version_CalcMethodOnlyOnTU")) — so it is automatically a user-friendly, translated message. AD_Message added in migration 5815870 (de_DE + en_US). Diagnostic params (HU_UnitType, PackageDimensionCalcMethod, M_HU_PI_Version_ID) kept via setParameter. The interceptor unit test's 3 message assertions were updated to the key (in a no-messages-loaded unit test the message resolves to the key string) — 6/6 green; the cucumber TC4 assertion checks the loaded en_US text and is validated by CI.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipping/impl/HUPackageBL.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/940700d0e55. You're right — isUnspecified() is UNSPECIFIED.equals(this) and PackageDimensions is a value object, so the ternary returned a value-equal object either way (a no-op). Simplified to 'return product.getPackageDimensions();'.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipping/IHUPackageBL.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/17f2d71093e. Removed SYSCONFIG_CHECK_IS_SELF_PACKED from IHUPackageBL — it is now a private constant in each consumer (HUPackageBL + PackedHUCarrierAdviseService), read via each class's isCheckSelfPacked() helper. (A prior review pass had flagged the duplicated literal and I'd extracted it onto the interface; reverted per your preference — the two private copies are the trade-off.)

backend/de.metas.handlingunits.base/src/main/sql/postgresql/system/70-de.metas.handlingunits/5815420_sys_M_HU_PI_Version_PackageDimensionCalcMethod_column.sql
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/17f2d71093e. Dropped the CHECK (PackageDimensionCalcMethod IN ('S','R','N')) constraint — the AD_Reference (List) already limits selectable values to the ref-list entries + null, and the DB CHECK only made extension harder. Column stays CHAR(1).

backend/de.metas.handlingunits.base/src/main/sql/postgresql/system/70-de.metas.handlingunits/5815420_sys_M_HU_PI_Version_PackageDimensionCalcMethod_column.sql
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/17f2d71093e. Added a Description on the AD_Element (+ de_DE/de_CH/en_US Trl): 'Calculation method for a Transport Unit's package dimensions when it holds multiple items (Strapping / Repacking / Nesting).'

backend/de.metas.handlingunits.base/src/main/sql/postgresql/system/70-de.metas.handlingunits/5815630_sys_gh30361_DimensionCalc_mode_PIs_inactive.sql
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/17f2d71093e. Translated all names/descriptions in the mode-PI migration to German — Maßberechnung - Bändern / Umverpacken / Verschachteln, consistently across M_HU_PI / M_HU_PI_Version / M_HU_PI_Item_Product for the three modes.

backend/de.metas.picking.rest-api/src/main/java/de/metas/picking/workflow/PackedHUCarrierAdviseService.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/17f2d71093e. Extracted a private isCheckSelfPacked() helper in PackedHUCarrierAdviseService (mirrors HUPackageBL); the call site now uses it.

backend/de.metas.shipper.client.nshift/src/main/java/de/metas/shipper/client/nshift/NShiftShipmentService.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/17f2d71093e. Changed the three '== 0' guards to '<= 0' so the check also rejects the unspecified (-1) sentinel, not just all-zero; message updated to 'zero or unspecified'.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/hu/M_HU_PI_Version_StepDef.java
mrie04: you changed it to adMsgkey on other review comment and still why not pass error code in name for example here and get rid of seperate step? `@And("metasfresh contains M_HU_PI_Version expecting error:")` . Looks like errorcode is missing on adMsg migration add it if needed

mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/pull/25297/commits/f4177b4336a. Consolidated into ONE step — `Then metasfresh contains M_HU_PI_Version expecting error "M_HU_PI_Version_CalcMethodOnlyOnTU":` — which saves the rows, catches the AdempiereException, and asserts its errorCode equals the passed code; removed the separate `Then an AdempiereException was thrown ...` step and the lastSaveException field. It asserts getErrorCode() (the language-independent AD_Message key) rather than the message text, because cucumber runs in de_DE so the text is German. On the migration: ErrorCode is already present — `ErrorCode = 'M_HU_PI_Version_CalcMethodOnlyOnTU'` at line 21 of 5815870 — so nothing to add there.

mrie04:
known mobile flake merging

PR #25298, opened 2026-07-23
Costing: ActivityControl cost collectors post no facts when resource has no cost product

teosarca:
## What

Manufacturing **ActivityControl** (routing / labour-time) cost collectors whose manufacturing resource has no cost product cannot post: the resource → cost-product lookup `IResourceProductService.getProductIdByResourceId` threw `"No product found"` (current build) or, on older builds, returned `null` and produced a message-less `NullPointerException` downstream in `CostingService`. When the resource has no cost product there is no activity cost to compute, so such cost collectors must post nothing rather than abort posting.

## Change

- `IProductDAO.getProductIdByResourceId` and `IResourceProductService.getProductIdByResourceId` now return `Optional<ProductId>` instead of throwing. Every existing caller that requires a product preserves today's behaviour with `.orElseThrow(() -> new AdempiereException("No product found for " + resourceId))`.
- The ActivityControl branch of the four `Manufacturing*CostingMethodHandler`s (`Standard`, `AveragePO`, `LastPO`, `MovingAverageInvoice`) now returns `CostDetailCreateResultsList.EMPTY` when the resource has no cost product — a graceful no-op mirroring the existing `if (actualResourceId.isNoResource()) return ...` guard directly above it. The empty result propagates to the existing no-fact guard in `Doc_PPCostCollector.createFacts_ActivityControl`, so posting succeeds with zero `Fact_Acct` lines.

`EMPTY` (not bare `null`) is returned deliberately: `CostingService` streams the handler result, so a `null` would NPE there and re-create the original message-less failure.

Behaviour is unchanged for every other cost-collector type (MaterialReceipt, ComponentIssue, the Variances) and for the UsageVariance branch, which still throws when a resource product is required.

## Covering test

`ManufacturingActivityControlNoProductTest` reproduces the failure first and passes after the fix: posting an ActivityControl cost collector whose resource has no cost product yields an empty result with no exception.

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/product/IProductDAO.java
+++ b/backend/de.metas.business/src/main/java/de/metas/product/IProductDAO.java
@@ -204,7 +204,7 @@ public interface IProductDAO extends ISingletonService
 
 	String getProductCategoryNameById(ProductCategoryId id);
 
-	ProductId getProductIdByResourceId(ResourceId resourceId);
+	Optional<ProductId> getProductIdByResourceId(ResourceId resourceId);
 
 	void updateProductsByResourceIds(Set<ResourceId> resourceIds, Consumer<I_M_Product> productUpdater);
 
--- a/backend/de.metas.business/src/main/java/de/metas/product/impl/ProductDAO.java
+++ b/backend/de.metas.business/src/main/java/de/metas/product/impl/ProductDAO.java
@@ -484,19 +484,15 @@ public class ProductDAO implements IProductDAO
 
 	@Cached(cacheName = I_M_Product.Table_Name + "#by#" + I_M_Product.COLUMNNAME_S_Resource_ID)
 	@Override
-	public ProductId getProductIdByResourceId(@NonNull final ResourceId resourceId)
+	public Optional<ProductId> getProductIdByResourceId(@NonNull final ResourceId resourceId)
 	{
-		final ProductId productId = queryBL
-				.createQueryBuilderOutOfTrx(I_M_Product.class)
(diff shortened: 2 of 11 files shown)


teosarca:
## Merging with `mobile test` red — pre-existing harness flake, not caused by this change

All checks are green **except `mobile test`**. Documenting why that red check is being accepted:

**This is a backend-only costing change.** The diff touches only `de.metas.business`, `de.metas.manufacturing`, and `de.metas.servicerepair.base` (manufacturing cost-collector posting; `getProductIdByResourceId` → `Optional<ProductId>`). No frontend/mobile files, no migration.

**The failing mobile specs cannot reach this code.** Every failure bottoms out at `tests/utils/screens/picking/GetQuantityDialog.js:204` (`expectComponentsDisabled` → `toBeDisabled()`, element-not-found) inside picking/distribution specs. The changed method `getProductIdByResourceId` has **zero callers** in picking / distribution / handlingunits / mobile — there is no code path from this diff to the failing UI assertion.

**Failure signature is a timing flake, not a regression.** Across 6 attempts the failing spec *set* shifts run-to-run (2 → 6 → 2 → 3 → 7 specs) while the assertion point stays fixed at the line-204 race. A regression fails the *same* specs deterministically.

**Mobile is broadly degraded this week, independent of this branch.** Base branch `intensive_care_hotfix` last ran `mobile test` green on 2026-07-18 (before the degradation) and has not re-run since; `new_dawn_uat` `mobile test` is red 2026-07-21→23 at a *different* timing seam. The "base green vs this PR red" gap is time, not code.

Root cause is a test-harness read-before-settle race at `GetQuantityDialog.expectComponentsDisabled`. Admin-merging over the red `mobile test` check on this basis.

PR #25301, opened 2026-07-23
Block closing a shipment schedule that has an unfinished picking job

teosarca:
## What

Closing a shipment schedule while it still has an **unfinished (Drafted) picking job** now hard-blocks
with a translated error message, instead of silently closing the schedule.

The guard applies only to the **user-initiated Close process**
(`M_ShipmentSchedule_CloseShipmentSchedules`). The automatic / BL close path
(`IShipmentScheduleBL.closeShipmentSchedule`) is unchanged and still closes normally.

## How

- **Detection SPI** `IShipmentSchedulePickingInfoService` (interface in `de.metas.swat.base`,
  implementation in `de.metas.handlingunits.base`) exposes a single `IQueryFilter<I_M_ShipmentSchedule>`
  matching schedules referenced by a Drafted `M_Picking_Job` (via `M_Picking_Job_Line` ∪
  `M_Picking_Job_Step`). It is resolved via `SpringContextHolder`, keeping the module boundary intact
  (no new `swat.base → handlingunits.base` dependency). Expressed as two `IN (subquery)` predicates
  OR'd together, so it folds straight into the caller's own selection query — there is no id
  round-trip and no JDBC bind-parameter limit, so the user selection size is unbounded.
- **Guard**, all-or-nothing over the full user selection: the filter is folded into the selection
  query, so the offending schedules come from a single query; if any exist, the whole close is
  rejected and nothing is closed. The offending schedules are listed in the message by order
  `DocumentNo` (falling back to `M_ShipmentSchedule_ID`).
- **Pre-existing eligibility unchanged**: among a picking-clean selection, only schedules with
  `QtyPickList = 0` are closed (schedules with picked-but-unshipped qty are skipped) — via the SQL
  `QtyPickList = 0` predicate on the close query (which excludes NULL, as before).
- **TOCTOU-safe**: the close query re-applies the unfinished-picking filter negated, so a schedule
  that turns picking-busy between the guard query and the close query is skipped rather than wrongly
  closed. Schedules are closed streaming (not materialized).
- New translated `AD_Message` for the block.

## Tests

- Cucumber `closeWithUnfinishedPicking.feature`: block / allow / all-or-nothing / automatic-path-still-closes.
- Unit tests for the detection filter (Line/Step reference, Completed/Voided/no-job exclusion, mixed
  selection) and the message-formatting helpers (order `DocumentNo`, `M_ShipmentSchedule_ID` fallback,
  shared-order dedup).


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/spi/IShipmentSchedulePickingInfoService.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Reports whether a shipment schedule still has picking work in progress.
+ * <p>
+ * The bridge exists because {@code de.metas.swat.base} (where the close process lives) has no dependency on
+ * {@code de.metas.handlingunits.base} (where {@code M_Picking_Job} lives), so the close process cannot name the
+ * picking-job tables directly. The {@code @Service} implementation lives in the handlingunits module and is resolved
+ * via {@code SpringContextHolder}.
+ */
+public interface IShipmentSchedulePickingInfoService
```
teosarca, line 35: why do we need a service only for that method?
why do we need an interface?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/process/M_ShipmentSchedule_CloseShipmentSchedules.java
```diff
@@ -46,27 +65,91 @@ protected String doIt() throws Exception
...
+	 */
+	@VisibleForTesting
+	static String toHumanReadableIdentifiersCsv(final List<I_M_ShipmentSchedule> offendingSchedules)
+	{
+		final ImmutableSet<OrderId> orderIds = offendingSchedules.stream()
+				.map(schedule -> OrderId.ofRepoIdOrNull(schedule.getC_Order_ID()))
+				.filter(Objects::nonNull)
+				.collect(ImmutableSet.toImmutableSet());
+
+		final Map<OrderId, String> documentNoByOrderId = Services.get(IOrderDAO.class).getByIds(orderIds)
```
teosarca: always declare services as field and not inlined Services.get(IOrderDAO.class)

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/process/M_ShipmentSchedule_CloseShipmentSchedules.java
```diff
@@ -46,27 +65,91 @@ protected String doIt() throws Exception
...
+	/**
+	 * Mirrors the pre-existing SQL {@code QtyPickList = 0} eligibility filter EXACTLY: only schedules whose
+	 * {@code QtyPickList} is set to zero are closed. A NULL {@code QtyPickList} is NOT eligible — the SQL {@code = 0}
+	 * predicate excludes NULLs — so the raw column value is read via {@link InterfaceWrapperHelper#getValueOrNull}:
+	 * the generated {@code getQtyPickList()} masks NULL as {@link BigDecimal#ZERO} and would wrongly include those rows.
+	 */
+	@VisibleForTesting
+	static boolean isEligibleForClose(final I_M_ShipmentSchedule schedule)
+	{
+		final BigDecimal qtyPickList = InterfaceWrapperHelper.getValueOrNull(schedule, I_M_ShipmentSchedule.COLUMNNAME_QtyPickList);
```
teosarca: why not simply schedule.getQtyPickList().signum() == 0?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/process/M_ShipmentSchedule_CloseShipmentSchedules.java
```diff
@@ -46,27 +65,91 @@ protected String doIt() throws Exception
 
 		final IQueryFilter<I_M_ShipmentSchedule> userSelectionFilter = getProcessInfo().getQueryFilterOrElseFalse();
 
-		IQueryBuilder<I_M_ShipmentSchedule> queryBuilderForShipmentScheduleSelection = shipmentSchedulePA.createQueryForShipmentScheduleSelection(getCtx(), userSelectionFilter);
+		// 1) Hard-block, all-or-nothing, over the FULL user selection: if any selected schedule still has an
```
teosarca, line 72: pls simplify the code from this class.... is still a big mess!

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/spi/IShipmentSchedulePickingInfoService.java
teosarca: **Resolution (dev):** Kept — the interface is structurally required, not optional. `de.metas.swat.base` (where this Close process lives) has no dependency on `de.metas.handlingunits.base` (where `M_Picking_Job*` lives); the dependency runs the other way (HU → swat, verified in the poms). So the process cannot name the picking-job tables directly — an interface defined in `swat.base` with a `@Service` impl in `handlingunits.base`, resolved via `SpringContextHolder`, is the only bridge. No existing swat-reachable service carries shipment-schedule-picking semantics (the only other swat-interface/HU-impl bridge, `OrderLineShipmentScheduleHandlerExtension`, is a write-path fan-out — wrong host). It is a one-method SPI + a one-line delegate impl, which is the minimal shape. I did trim the interface's javadoc so the lower-level interface no longer names its implementing module. Commit: https://github.com/metasfresh/metasfresh/commit/1b79ab9d1fe37be414250cde3ef36317a6518844

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/process/M_ShipmentSchedule_CloseShipmentSchedules.java
teosarca: **Resolution (dev):** Done. Hoisted `IShipmentSchedulePA`, `IShipmentScheduleBL` and `IOrderDAO` to `private final` fields; `toHumanReadableIdentifiersCsv` now uses the `orderDAO` field instead of the inline `Services.get(IOrderDAO.class)`. Commit: https://github.com/metasfresh/metasfresh/commit/1b79ab9d1fe37be414250cde3ef36317a6518844

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/process/M_ShipmentSchedule_CloseShipmentSchedules.java
teosarca: **Resolution (dev):** `getQtyPickList().signum() == 0` would be a regression: the generated `getQtyPickList()` masks a NULL column as `BigDecimal.ZERO`, so it would treat a NULL-`QtyPickList` schedule as eligible, whereas the original SQL `QtyPickList = 0` predicate excludes NULL. So instead of an in-memory check I moved `QtyPickList = 0` back into the SQL selection query (`addEqualsFilter`) — the original approach: simpler and NULL-correct. `isEligibleForClose` and its unit tests are removed. Commit: https://github.com/metasfresh/metasfresh/commit/1b79ab9d1fe37be414250cde3ef36317a6518844

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/process/M_ShipmentSchedule_CloseShipmentSchedules.java
teosarca: **Resolution (dev):** Simplified. `doIt()` is now: guard query (selection folded with the unfinished-picking filter) → if any offending, throw → close query (SQL `QtyPickList = 0` + the negated picking filter, streamed) → close each. Removed `isEligibleForClose` (eligibility is now the SQL predicate), hoisted the three services to fields, and dropped the redundant `markAsUserValidationError`. The earlier commits on this PR had already removed the `Set<ShipmentScheduleId>` round-trip, the 30000-id chunking (+ its test), and the in-memory intersection — the SPI now hands back one filter the process folds into its own query. Commit: https://github.com/metasfresh/metasfresh/commit/1b79ab9d1fe37be414250cde3ef36317a6518844

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/process/M_ShipmentSchedule_CloseShipmentSchedules.java
```diff
@@ -25,48 +25,113 @@
...
-
 		final IQueryFilter<I_M_ShipmentSchedule> userSelectionFilter = getProcessInfo().getQueryFilterOrElseFalse();
 
-		IQueryBuilder<I_M_ShipmentSchedule> queryBuilderForShipmentScheduleSelection = shipmentSchedulePA.createQueryForShipmentScheduleSelection(getCtx(), userSelectionFilter);
+		// 1) Hard-block, all-or-nothing, over the FULL user selection: if any selected schedule still has an
+		// unfinished (Drafted) picking job, reject the whole close and close NOTHING. The unfinished-picking check
+		// is a subquery filter folded into the selection query, so the offending schedules come from a single query
+		// (no id round-trip / in-memory intersection).
+		final List<I_M_ShipmentSchedule> offendingSchedules = shipmentSchedulePA.createQueryForShipmentScheduleSelection(getCtx(), userSelectionFilter)
+				.filter(pickingInfoService.newUnfinishedPickingFilter())
```
teosarca: have also a limit!

teosarca: consider extracting this snippet as a assertNoOffendingSchedules method

teosarca: **Resolution (dev):** Done both.
1. **Limit** — capped the offending-schedules query with `setLimit(100)` (`MAX_OFFENDING_SCHEDULES_TO_REPORT`). It feeds only the rejection message, so a single offender still rejects the whole all-or-nothing close; the cap just bounds how many get named instead of loading an unbounded selection.
2. **Extraction** — pulled the guard block out of `doIt()` into a private `assertNoOffendingSchedules(userSelectionFilter)` method; `doIt()` now reads: `assertNoOffendingSchedules(...)` → close query. Behavior unchanged.
Commit: https://github.com/metasfresh/metasfresh/commit/20821a901b4e78f0913eae80c5f84ddf6851bf6c

PR #25302, opened 2026-07-23
WebUI: surface relevant save-error reasons in input dialogs (and keep them visible)

teosarca:
# WebUI: surface relevant save-error reasons in input dialogs (and keep them visible)

## Problem
When a save is rejected **server-side** in a WebUI input dialog while the input is complete and each field is individually valid (e.g. a new/edited record whose value violates a DB unique index), the friendly translated reason was **not shown while the dialog was open**, or only **flashed for ~1 s and then reverted** — leaving the user unable to read why the save was refused. Two distinct causes:

1. **Frontend — the reason was swallowed.** The shared `computeSaveStatusFlags` promoted the error indicator only when `presentInDatabase` was true, so a **not-yet-persisted new record** whose relevant save failed never surfaced the reason — on the included-tab modal *and* on the main-window (non-modal) header.
2. **Backend — the reason was self-healed away.** For an **edit of an already-persisted record**, the document-cache self-heal (evict the errored root on child invalidation — https://github.com/metasfresh/metasfresh/pull/23672 / https://github.com/metasfresh/metasfresh/pull/24697) also evicted a **user-validation** error state, reloading a clean document — so the just-shown reason and the rejected value reverted after the websocket round-trip.

## Fix
**Do not swallow relevant save errors; do not spam on incomplete/invalid input; keep a user-fixable rejection visible until corrected.**

**Frontend** (`frontend/src/reducers/windowHandler.js`, `Modal.js`)
- `isRelevantSaveError(saveStatus)` = `error === true && exception.userFriendlyError === true` — a user-fixable business rejection, distinct from a mandatory-missing/incomplete state (which has an error flag but no user-friendly exception and stays quiet, handled by the existing field-level cues).
- The shared core `computeSaveStatusFlags` promotes the indicator to ERROR for a relevant save error in **window context** (main-window master + window modal), reusing `isRelevantSaveError` (single predicate, no duplication). An `isWindowContext` guard excludes **process** modals so their Start button is not disabled.

**Backend** (`de.metas.ui.web.base` — `DocumentCollection`, `DocumentSaveStatus`, `DocumentValidStatus`)
- A computed `userValidationError` flag (from `AdempiereException.isUserValidationError(exception)`, null-safe) on both status types.
- `DocumentCollection.shouldInvalidateRootOnChildInvalidation` no longer treats a **user-validation** error as an eviction reason (covering both the save-status-error and valid-status-invalid triggers), so the errored root is kept and the frontend keeps mirroring the reason. A **system/technical** save error, a system invalid state, and an explicit full-invalidation still evict — so the self-heal that https://github.com/metasfresh/metasfresh/pull/23672 / https://github.com/metasfresh/metasfresh/pull/24697 introduced is preserved for the cases it was built for.

## Tests
- Backend: `DocumentCollectionShouldInvalidateRootTest` (full system×user-validation × save-error×valid-status matrix + full-invalidation-still-evicts + lazy-supplier), `DocumentSaveStatusTest`, `DocumentValidStatusTest`. Full `de.metas.ui.web.base` module green.
- Frontend: `WindowHandler.test.js` — `isRelevantSaveError` and core `computeSaveStatusFlags` surfacing (window/master relevant-error → ERROR; mandatory-missing → quiet; process-modal → unaffected). Full frontend jest suite green.
- E2E: `pricelist-version-colliding-date.spec.js` — new-record collision (reason shown on screen + document not lost) and edit-existing collision (the rejection persists, document not self-heal-evicted).


e2e/frontend-webui/tests/spec/pricelist-version-colliding-date.spec.js
```diff
@@ -24,10 +24,63 @@ import { DateWidget } from '../utils/widgets/DateWidget';
...
+  const existingRows = Array.isArray(rowsBody) ? rowsBody : rowsBody.result || rowsBody.documents || [];
+  // Guard the self-heal: the clean 2015 seed PLV MUST be recognized before we delete anything. If a
+  // future backend change alters the row-value shape (adds a time/zone component, changes the envelope),
+  // isCleanSeedRow would stop matching the seed and the loop below would delete the shared fixture both
+  // tests depend on — with no failure pinpointing the cause. Fail loud here instead, before any delete.
+  expect(
+    existingRows.some(isCleanSeedRow),
+    'the clean 2015 seed PLV must be recognized before self-heal — else the delete loop would remove the shared fixture'
+  ).toBe(true);
+  for (const r of existingRows) {
```
teosarca: pls always use intention revealing names

e2e/frontend-webui/tests/spec/pricelist-version-colliding-date.spec.js
```diff
@@ -136,11 +203,178 @@ new child PLV.
...
+with no 404 on that path.
+    `);
+
+    test.setTimeout(120000);
+
+    // --- Network instrumentation on the PLV document path ---
+    // Records EVERY response on the PLV path (PATCH saves + GET re-fetches) in arrival order, with the
+    // server-side saveStatus.error of each returned document — so we can prove the post-failure GET retained
+    // the error (the carve-out) rather than returning a clean, evicted-and-rebuilt document.
+    const events = [];      // { seq, method, status, error, userFriendlyError, reason, saved }
```
teosarca: pls consider extracting and using smaller functions with intention revealing names.
also don't mix multiple layers in the same function

e2e/frontend-webui/tests/spec/pricelist-version-colliding-date.spec.js
teosarca: **Resolution (dev):** Renamed the cryptic locals to intention-revealing names (r -> row, rid -> rowId, del -> deleteResponse) and kept this loop inside the named helper `normalizePriceListToSeed`. Commit https://github.com/metasfresh/metasfresh/commit/4902357af3426579dc0cabdd042c5c9c8fce0a41

e2e/frontend-webui/tests/spec/pricelist-version-colliding-date.spec.js
teosarca: **Resolution (dev):** Split the inline per-test setup into single-layer named helpers so each test now reads as a sequence of named steps instead of mixed network/UI/assertion code: network layer `recordPlvPathResponses` (+ `failedSaveResponses`), UI layer `loginAndOpenCleanPriceList` / `addNewPlvRow` / `openSchemaDropdownAndAwaitList` / `setValidFrom`, and test-data `uniqueFarFutureValidFrom`. Both tests re-run green locally. Commit https://github.com/metasfresh/metasfresh/commit/4902357af3426579dc0cabdd042c5c9c8fce0a41

PR #25311, opened 2026-07-23
Costing: ActivityControl cost collectors post zero facts (fix NPE + not-yet-supported)

teosarca:
## What

Fix `PP_Cost_Collector` posting failures for **ActivityControl** cost collectors. Two failure modes could occur when posting an ActivityControl cost collector:

1. **`NullPointerException`** — when the cost collector's `S_Resource_ID` equals `ResourceId.NO_RESOURCE` (the no-resource sentinel), `isNoResource()` was true and the handler's `createCost` returned `null`. `CostingService` then dereferenced that null at `handler.createOrUpdateCost(request).stream()`.
2. **`AdempiereException: Computing activity costs is not yet supported`** — ActivityControl on an AveragePO / MovingAverageInvoice / LastPO cost element reached a `createActivityControl` that unconditionally threw.

Intended behaviour: an ActivityControl cost collector posts **zero facts** (graceful no-op) across all costing methods.

## Changes

- `ManufacturingStandardCostingMethodHandler` — the two `isNoResource()` guards (ActivityControl + no-BOM UsageVariance) return `CostDetailCreateResultsList.EMPTY` instead of `null`. The real `createActivityControl` cost posting is unchanged.
- `ManufacturingAveragePOCostingMethodHandler` — `isNoResource()` guard returns `EMPTY`; `createActivityControl` is a zero-fact no-op instead of throwing.
- `ManufacturingMovingAverageInvoiceCostingMethodHandler`, `ManufacturingLastPOCostingMethodHandler` — `createActivityControl` is a zero-fact no-op instead of throwing.
- `CostingService.createCostDetailUsingHandlersAndStream` — defensively treats a null handler result as an empty stream, so a null return can never NPE the pipeline again.

## Tests

New `ManufacturingActivityControlNoResourceTest` (TDD RED→GREEN): the NO_RESOURCE-sentinel ActivityControl path for Standard + AveragePO, and the product-linked-resource "not yet supported" path for AveragePO/LastPO/MovingAverage. All fail against the pre-fix code for the intended reason (null return / thrown exception) and pass after. Full `de.metas.manufacturing` costing test package green (83/83).

## Note

Supersedes #25298, whose guard sat after the `return null` and was never reached for the sentinel case.


backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/ManufacturingMovingAverageInvoiceCostingMethodHandler.java
```diff
@@ -227,8 +227,8 @@ private CostDetailCreateResult createActivityControl(
 			final CostDetailCreateRequest ignoredRequest,
 			final Duration ignoredTotalDuration)
 	{
-		// TODO Auto-generated method stub
-		throw new AdempiereException("Computing activity costs is not yet supported");
+		// Activity-control costs are not tracked for this costing method -> post zero facts (me03#29817)
```
teosarca: comments shall be timeless (no me03). btw we have a hook to prevent such commits, how this one escaped?!

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/ManufacturingMovingAverageInvoiceCostingMethodHandler.java
```diff
@@ -227,8 +227,8 @@ private CostDetailCreateResult createActivityControl(
...
+		return null;
```
teosarca: why returning null and not CostDetailCreateResultsList.EMPTY?

backend/de.metas.manufacturing/src/test/java/de/metas/costing/methods/ManufacturingActivityControlNoResourceTest.java
```diff
@@ -0,0 +1,342 @@
...
+ *     <li>the {@code S_Resource_ID = ResourceId.NO_RESOURCE} (540011) sentinel collision, which made
+ *     {@code createCost} return {@code null} and NPE the pipeline, and</li>
+ *     <li>a real, product-linked resource reaching {@code createActivityControl}, which used to
+ *     {@code throw new AdempiereException("Computing activity costs is not yet supported")}.</li>
+ * </ul>
+ * Both must be graceful no-ops returning {@link CostDetailCreateResultsList#EMPTY}, for every
+ * costing method.
+ */
+@ExtendWith(AdempiereTestWatcher.class)
+public class ManufacturingActivityControlNoResourceTest
```
teosarca: why not cucumbers and get rid of junit?

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/ManufacturingMovingAverageInvoiceCostingMethodHandler.java
teosarca: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/854ac03d9ead6bf7da57439a4e29b28d003ce3c2 — removed the non-timeless reference from every ActivityControl no-op comment and from the NoResource test Javadoc; comment/Javadoc only, no behaviour change. Note: this checkout has no active commit-msg hook (only the git *sample* hooks are present), which is how it slipped through locally.

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/ManufacturingMovingAverageInvoiceCostingMethodHandler.java
teosarca: **Resolution (dev):** Not applicable as literally stated — `createActivityControl(...)` returns a single `CostDetailCreateResult` (nullable), not a `CostDetailCreateResultsList`, so `CostDetailCreateResultsList.EMPTY` would not type-check there. `null` is the established "no cost detail produced" idiom for that return type: it is wrapped by `return CostDetailCreateResultsList.ofNullable(result)` at the end of the dispatch (→ EMPTY), and matches the sibling `result = null` branches (UsageVariance/etc.) in the same method. The list-typed early-return guards do return `CostDetailCreateResultsList.EMPTY`. Happy to also drop the now-vestigial helper and `return ...EMPTY` straight from the ActivityControl branch if you prefer that shape — that is a larger change than the approved minimal fix, so left as-is unless you want it.

backend/de.metas.manufacturing/src/test/java/de/metas/costing/methods/ManufacturingActivityControlNoResourceTest.java
teosarca: **Resolution (dev):** Kept as JUnit. The de.metas.manufacturing test tree is JUnit-only (no feature files), and the direct sibling covering the no-cost-product path is also a JUnit test, so this matches the module convention. The bug is a handler-level null-return / thrown-exception on a specific dispatch branch — a focused handler unit test reproduces it exactly (RED against the pre-fix code) and is the right granularity. A Cucumber scenario would need the full manufacturing-order to cost-collector pipeline stood up end to end, which is disproportionate for a no-op/null guard and has no footprint in this module today. If you would still rather have end-to-end coverage here, I will add it — just confirm and I will scope it.

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/ManufacturingMovingAverageInvoiceCostingMethodHandler.java
teosarca: **Resolution (dev):** Done — the `isActivityControl()` branch now returns `CostDetailCreateResultsList.EMPTY` directly in all three no-op handlers (AveragePO, MovingAverageInvoice, LastPO), and the now-vestigial `createActivityControl` helper (which only ever returned `null`) has been removed. Commit https://github.com/metasfresh/metasfresh/commit/908404faae8

backend/de.metas.manufacturing/src/test/java/de/metas/costing/methods/ManufacturingActivityControlNoResourceTest.java
teosarca: **Resolution (dev):** Done — the JUnit test has been dropped and replaced by an end-to-end Cucumber scenario in `manufacturingCostCollectorPosting.feature` that exercises the real posting path: it closes a manufacturing order whose routing activities resolve to the no-resource placeholder, so closing creates the ActivityControl cost collectors, then asserts they post successfully (`Posted='Y'`) with zero `Fact_Acct` rows. Proven RED (`NullPointerException` posting error against the pre-fix code) then GREEN after the fix. Commit https://github.com/metasfresh/metasfresh/commit/c52cc0725c1

PR #25314, opened 2026-07-24
Costing — back-dated MovingAverageInvoice switch: as-of-cut-off opening, opening-anchor guard, batched recompute driver

teosarca:
Makes a **back-dated switch to MovingAverageInvoice costing** survivable: the switch opens the target cost element with the source's cost **as of the cut-off**, the recompute that follows **refuses** to destroy that opening, and a full-installation recompute runs as a **batched, resumable procedure** that releases its documents to the reposting queue in document-date order.

Four independent streams, each with its own tests. They share one branch because they are only correct together — the opening is worthless if the recompute deletes it, and the recompute cannot run at all while a pre-range inbound aborts it.

---

## 1. Recompute no longer aborts on a weighted-average inbound before the range

`de_metas_acct.product_costs_recreate_from_date` → `m_costdetail_delete_from_date` aborted with `Extracting current costs from an inbound transaction is not implemented` when a product's last cost detail before the recompute start date was a **non-inventory, changing-costs inbound** (`M_MatchPO`, `M_InOutLine` non-material costs, `PP_Cost_Collector` receipt). The qty-classification only handled such an inbound when it was an inventory line (case 2.4); every other one fell to an unimplemented `ELSE … RAISE`.

New **case 2.5** in `m_costdetail_delete_from_date` reconstructs the current cost price by **weighted average** `round((prevPrice·prevQty + amt)/(prevQty + qty), costingPrecision)`, mirroring `de.metas.costing.CurrentCost.addWeightedAverage`. Case 2.4 is unchanged; the original `ELSE … RAISE` stays as the fallback.

### Gated on the cost element's costing method

Case 2.5 fires only for `costingmethod IN ('A','I','M','S')`, because only those move the price by averaging:

| Method | Behaviour on a changing-costs inbound | In gate |
|---|---|---|
| `A` AveragePO, `I` AverageInvoice, `M` MovingAverageInvoice | handlers call `CurrentCost.addWeightedAverage` | yes |
| `S` StandardCosting | records `amt = prevPrice·qty`, so the average provably reproduces `prevPrice` — a no-op, not an approximation | yes |
| `p` LastPOPrice, `i` LastInvoice | on `M_MatchPO` / `M_MatchInv` the handlers **replace** the price with `amt/qty` | no |
| `L` Lifo, `F` Fifo, `U` UserDefined, `x` ExternalProcessing | not established | no |

Without the gate, a LastPOPrice/LastInvoice-costed product would get a silently **wrong** `M_Cost.CurrentCostPrice`: with `prevPrice=20, prevQty=10` and an inbound `qty=10, amt=400`, replace yields `40` but averaging yields `600/20 = 30`.

Excluded methods fall through to the pre-existing `RAISE`, i.e. exactly the behaviour they have today — so this change can never convert a loud abort into a silent mis-valuation. A `NULL` costingmethod (unresolvable cost element) likewise falls through rather than averaging blindly. The `RAISE` message now reports the costing method so a future abort is diagnosable.

**Known residual gap (deliberate, out of scope):** for `p` and `i` the price movement depends on the **document**, not on the method alone, so a method-only rule cannot model them. Variants reachable with `qty>0` + `IsChangingCosts='Y'` + no inventory line — i.e. all matching case 2.5's shape — include: replace with `amt/qty` (base handler, `M_MatchPO`/`M_MatchInv`); a weighted average that *is* a no-op (`ManufacturingLastPOCostingMethodHandler.createMainProductOrCoProductReceipt`, where `amt = currentCostPrice·qty`); a weighted average that is *not* a no-op (same class, `createComponentIssue` reversal, where `amt/qty` is the price at the original issue); and price left **untouched** with only qty moving (`createOutboundCostDefaultImpl` reversal → `CurrentCost.addToCurrentQtyAndCumulate`). That list is examples, not exhaustive — it was incomplete twice.

So no single no-op variant is licence to add `p`/`i` to the gate. Excluding them is safe for a reason that does **not** depend on enumerating variants: the only fallback is the loud `RAISE`, never a silently wrong `CurrentCostPrice`. Consequence: a `p`/`i` element whose last pre-range detail has this shape still aborts instead of being reconstructed. Closing that needs a `(method, document)` key.

---

## 2. A point-in-time cost read, and the switch uses it

`ICostingService.getCurrentCost` only ever answers with the **live** `M_Cost` row. For a back-dated cut-off that is the cost *after* every movement booked since, so a `CopyFromCostElement` cost revaluation dated `EvaluationStartDate = <cut-off>` opened the target element with a value the source never had at the cut-off — and stamped it as of the cut-off.

New read:

- `ICostingService#getCostAsOf(costSegmentAndElement, asOfDate)` → `Optional<CostDetailPreviousAmounts>`. It reconstructs the state from the `Prev_*` columns of the **first changing-costs `M_CostDetail` dated after `asOfDate`** — those columns hold exactly the state the element was in immediately before that movement. With no such detail nothing moved after the cut-off, so the live `M_Cost` row *is* the state as of the date.
- `ICostDetailRepository#getFirstChangingCostsDetailAfter(...)`, expressed as a `CostDetailQuery` filter (`changingCosts`, `dateAcctRage`, `DATE_ACCT_ASC` + `ID_ASC`). Deliberately the same algorithm as the SQL function `getCurrentCostInfo`, so a point-in-time valuation read done in SQL and one done in Java agree by construction.
- The boundary is strictly `>`, never `>=`: the switch writes its own opening anchor dated exactly **at** the cut-off, and that anchor must never count as a forward event — otherwise the switch would seed itself from its own anchor's `Prev_*`.

`CostRevaluationService` now uses it on **both** sides, so the drafted lines preview exactly the numbers complete-time will write:

- `createLinesFromCopyFromCostElement` restates each source `M_Cost` as of `EvaluationStartDate` before creating the lines;
- `createDetailsForCopyFromCostElement` seeds `M_Cost` + the opening anchor from that same as-of read (own price, lower-level price and qty from **one** read — never a stale-own/qty + fresh-LL mix).

Both fail loudly (`AdempiereException`) when the source has no cost as of the cut-off, instead of silently falling back to the live cost — that fallback is the very bug this read exists to remove.

---

## 3. The recompute refuses to delete a cost-revaluation opening anchor

`m_costdetail_delete_from_date` now refuses, before its `DELETE`, when the delete range contains an `M_CostDetail` with `M_CostRevaluationLine_ID IS NOT NULL` — the single zero-delta opening anchor a completed `CopyFromCostElement` revaluation writes at the cut-off. Same range predicate as the `DELETE`, so the guard fires exactly when the `DELETE` would hit an anchor, and it sits at the choke point, so it protects direct callers too, not only `product_costs_recreate_from_date`.

Concrete failure it prevents: December is re-opened for year-end adjustments and someone recomputes a product from a December document, i.e. from a date at or before the cut-off. What makes it dangerous is that **nothing visibly breaks** — the anchor is `IsChangingCosts='Y'` and its `Prev_*` *are* the seeded opening, so case 1 picks the anchor, `M_Cost` is updated back to the very same numbers and survives at the correct price. The damage is the silent loss of the anchor **row**: the `M_CostRevaluationLine` survives with no corresponding `M_CostDetail`, which destroys (a) the reversal basis (`reverseSeededCurrentCost` undoes the switch by deleting that anchor), (b) the only record of what the opening balance was, and (c) the double-seed guard (`retrieveProductIdsWithCostRevaluationSeed` keys on exactly that column), so a re-run of the switch re-seeds an already-in-use cost.

Refusing costs nothing: the recompute is re-runnable with a start date strictly after the cut-off, which is the only correct range anyway — the opening is a given, not something a recompute may replay. Until now the *only* thing preventing this was that the previous year's periods happen to be closed.

---

## 4. `product_costs_recreate_all_from_date` — batched, resumable, order-preserving

New procedure `de_metas_acct.product_costs_recreate_all_from_date(p_C_AcctSchema_ID, p_M_CostElement_ID, p_StartDateAcct, p_ProductsPerCommit = 100, p_Resume = 'N')`, plus two new tables, `de_metas_acct.accounting_docs_to_repost_staging` and `de_metas_acct.product_costs_recreate_progress`.

Why a procedure and not a function: a full run over a real installation is thousands of products and hundreds of thousands of documents and cannot be one transaction — a single-transaction attempt ran 2h13m and then died with a backend I/O error, rolling everything back. So the work is cut into commit-batches of `p_ProductsPerCommit` products; each batch rewinds its products, stages its documents and records itself `DONE` in **one** transaction, and `p_Resume='Y'` skips the batches already done.

Ordering is the whole point of the staging table. Costs are computed *while* reposting, so documents must reach the accounting server strictly in document-date order — but the server's poller drains `accounting_docs_to_repost` continuously, so anything inserted there as it is found would be reposted early, out of order, and would freeze wrong costs. The rows are therefore parked in the staging table and released in **one** step at the end, with `SeqNo` assigned explicitly by `row_number() OVER (ORDER BY dateacct, tablename_prio, record_id)` over a block of sequence values reserved in a single `setval(nextval(…) + n - 1)` statement. `SeqNo` is never left to the column's `nextval` default: the order in which `INSERT … SELECT` consumes a sequence is not guaranteed by an `ORDER BY` in the `SELECT`. Nothing in the real queue, its poller, or `fact_acct_unpost` changes.

Guards, all of them before the first mutation:

- **one run at a time** — a *session*-level advisory lock (a transaction-level one would be released by the first per-batch commit, leaving the rest of a multi-hour run unprotected);
- **every period in range open** — `assert_period_open` per distinct `(dateacct, docbasetype, org)`, because the repost throws `@PeriodClosed@` for an already-posted document in a closed period, and finding that out 40 batches in would leave the installation mid-rewind;
- **a resume must prove it continues the run it resumes** — `product_costs_recreate_progress` carries the four run-identity columns (`c_acctschema_id`, `m_costelement_id`, `startdateacct`, `products_per_commit`), written in the same transaction that marks a batch `DONE`. `batch_no` is *not* a stable name for a set of products — it is recomputed on every call from the caller's own `p_ProductsPerCommit`, and the product population is re-derived from the documents in range. So a resume with a different products-per-commit, a resume that meets the `DONE` rows of a *completed* earlier run, or a resume after a new document brought a new product into range, would each read "batch 3 is DONE" for a **different** set of products: never rewound, never staged, and the run reports success. Both halves of the check (parameters, then population) therefore run **up front** — the population half sits at the last mutation-free moment, because the target-element purge below commits, and a refusal raised after it would already have destroyed the rows a later correct resume needs.

The run also purges the target MovingAverageInvoice element's in-range cost details first (raw `DELETE`, then `m_costdetail_delete_from_date` for the same products): posting explodes a document into *every* active Material cost element, so the target may already carry in-range details built before it had an opening balance, and those are reused as-is by `CostingMethodHandlerTemplate#createOrUpdateCost` and win over the anchor in case 1. Only products that actually have such details are passed to `m_costdetail_delete_from_date` — for a seeded-but-not-moved product it would find nothing to restore from and delete the `M_Cost` row outright.

### Deliberate, evidenced deviation from a coding rule

`backend/de.metas.acct.base/CLAUDE.md:81` and `docs/coding-rules/production-database-support.md:65` both state: *never insert directly into `Accounting_Docs_To_Repost` — always use the `de_metas_acct.fact_acct_unpost*` functions.* **The release step inserts directly, knowingly.** Not an oversight, and not a rule amendment either — the rule stands; this run documents why it does not apply. `fact_acct_unpost` bundles two safety-relevant side effects with the enqueue, and both are redundant for a queue-driven repost:

- **deleting the stale `Fact_Acct` rows** — the repost does it itself: `PostingService.postNow` calls `doc.post(request.isForce(), true)` with `repost` **hardcoded** `true` (`backend/de.metas.acct.base/src/main/java/de/metas/acct/api/impl/PostingService.java:97`), and `Doc.post` does `if (repost) { deleteAcct(); }` before saving the new facts (`backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc.java:360-363`). So no bulk `fact_acct` `DELETE` is needed;
- **resetting `Posted='N'` so the document can be locked again** — `SqlAcctDocLockService.lock()` adds the `AND Posted='N'` condition only when `repost` is false (`SqlAcctDocLockService.java:29-32`), so a document left at `Posted='Y'` locks and reposts fine.

What `fact_acct_unpost` cannot do is enqueue a whole set of documents with a chosen `SeqNo` ordering in one statement — and that ordering is the entire correctness contract of a cost recompute. Worse, it inserts straight into `accounting_docs_to_repost`, handing documents to the poller one by one as they are found: exactly the out-of-order reposting this procedure exists to prevent. Using it here would trade a provable ordering guarantee for two no-ops.

The driver therefore only deletes `M_CostDetail`, fixes `M_Cost` and enqueues in chronological order — it never touches `Fact_Acct` and never enters the real repost queue except through the ordered release.

**Retained caveat:** the one `fact_acct_unpost` effect that is *not* redundant is the period check. `Doc.java:333-339` still throws `@PeriodClosed@` for an already-posted document in a closed period, so every period in the run's range must be open — which is why the driver asserts that up front (see the guards above).

---

## Migration scripts

| Script | Content |
|---|---|
| `5816000` | `m_costdetail_delete_from_date` — case 2.5 |
| `5816900` | `m_costdetail_delete_from_date` — the opening-anchor guard |
| `5816910` | `accounting_docs_to_repost_staging` table |
| `5816920` | `product_costs_recreate_progress` table |
| `5816930` | `product_costs_recreate_all_from_date` procedure |
| `5816940` / `5816950` | run-identity columns + the resume check that uses them |
| `5817060` | the population pre-check hoisted before the purge |

Each function/table also has its source-of-truth `ddl/functions/` resp. `ddl/table/` mirror in `backend/de.metas.acct.base/src/main/sql/postgresql/`.

## Tests added

| Test | Covers |
|---|---|
| `CostingServiceAsOfTest` (3 tests) | `getCostAsOf`: reconstruction from the first detail strictly after the date, fallback to the live `M_Cost`, and a detail *on* the date counted as before it |
| `CostRevaluationServiceTest#seedsTheSourceCostAsOfTheEvaluationStartDate` | the seed reads the source as of the cut-off, not its live value |
| `@Id:S26253_TC7` | back-dated switch opens with the source cost as of the cut-off, end to end (incl. an `M_CostRevaluation` step-def asserting the anchor's `DateAcct`) |
| `@Id:S26253_TC8` | a recompute reaching back over the opening anchor is refused instead of deleting it |
| `@Id:S26253_TC9` | recompute reconstructs the current cost from a weighted-average `M_MatchPO` inbound dated before the start date (case 2.5) |
| `@Id:S26253_TC10` | the batched driver stages its documents, releases them in document-date order, and resumes after a crash |
| `@Id:S26253_TC11` | a resume that regroups the products into different commit-batches is refused |
| `@Id:S26253_TC12` | a resume refused because a product joined the population purges nothing on its way to the refusal |

`TC9` is the case-2.5 regression scenario, in its own feature file; `TC7`, `TC8`, `TC10`–`TC12` extend the existing `switchToMovingAverageInvoice.feature` (which already carried `TC1`–`TC6` on the base branch). `TC10`–`TC12` are driven by a new `ProductCostsRecreateAll_StepDef`.

`TC9` distinguishes three outcomes: RED (the old function aborts), a naive "copy case 2.4" fix (would leave 20 CHF) and the correct weighted-average fix (30 CHF). Its `M_CostDetails are found` step asserts an **exact** set, so it lists all four details the two receipts book — the two cost-changing `M_MatchPO` rows and the two non-cost-changing `M_InOutLine` rows.

Coverage note: case 2.5 is exercised for method `A` only; the `I` / `M` / `S` claims above are established from the handler sources, not from a scenario in this PR.

The originally failing case was additionally reproduced and confirmed fixed by an isolated single-product dry-run on a customer's hotfix instance.

## Status

Draft. Each of `@Id:S26253_TC7`, `TC8`, `TC9`, `TC10`, `TC11`, `TC12` was run locally (`scenarios=1 failures=0` each) against a stack carrying this branch's migrations, and the module's JUnit suites are green.

CI on the current head is still running. On the previous head https://github.com/metasfresh/metasfresh/commit/00bc4289933 (run https://github.com/metasfresh/metasfresh/actions/runs/30526098979) everything was green — cucumber 724/724, JUnit backend 8299 passed / 0 failed, frontend-webui 19/19, camel 192, jest 338, `db-apply-migrations`, `migration-cli`, `db-images`, `compatibility-images`, CodeQL — **except `mobile test`**, which is red for reasons unrelated to this diff:

- `git diff origin/intensive_care_hotfix...HEAD -- e2e/ misc/services/mobile-webui/ frontend/` is **empty** — this PR changes no frontend, mobile or E2E code at all;
- the suite is red on **67 %** of recent runs of the base branch `intensive_care_hotfix` itself (4 of the last 6), with none of this PR's code, and on 52 % of runs on this branch (11 of 21);
- the failures are 1–2 of ~201 specs and land on three distinct client-side seams: `barcode_scanner_modes.spec.js:396` (camera-toggle case 3 — fixed on `new_dawn_uat` by https://github.com/metasfresh/metasfresh/pull/25307, not ported to this base branch), `distribution/distribution.spec.js:161` (`#WFProcessScreen` job-start seam) and `picking/picking.spec.js:703` (`expectErrorToast` grace-window race).

### DIFF
--- a/backend/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/m_costdetail_delete_from_date.sql
+++ b/backend/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/m_costdetail_delete_from_date.sql
@@ -159,14 +159,22 @@ BEGIN
                 --        LastInvoiceCostingMethodHandler.createCostForMatchInvoice_MaterialCosts), so averaging
                 --        would silently write a WRONG CurrentCostPrice. Example: prevPrice=20, prevQty=10,
                 --        inbound qty=10 amt=400 -> replace gives 40, averaging gives 600/20 = 30.
-                --        CAVEAT for 'p': ManufacturingLastPOCostingMethodHandler.createMainProductOrCoProductReceipt
-                --        DOES call addWeightedAverage for a PP_Cost_Collector main/co-product receipt - but there
-                --        amt = currentCostPrice*qty, so that average is a no-op (same shape as 'S'). So 'p' behaves
-                --        two ways depending on the DOCUMENT, and a costingmethod-only gate cannot separate them;
-                --        'p' is therefore excluded wholesale, which is safe for BOTH sub-cases. Known residual
-                --        gap: a 'p' element whose last pre-range detail is a PP_Cost_Collector receipt still
-                --        RAISEs instead of being reconstructed. Widening the gate to a (method, document) key
-                --        would close it - deliberately out of scope here.
+                --        CAVEAT - 'p' is NOT uniform: it has at least three reachable shapes, because the
+                --        manufacturing handler (ManufacturingLastPOCostingMethodHandler) overrides the base one.
+                --          (1) M_MatchPO                       -> REPLACE with amt/qty (base handler, above).
+                --          (2) PP_Cost_Collector main/co-product receipt (createMainProductOrCoProductReceipt)
+                --              -> addWeightedAverage, but with amt = currentCostPrice*qty, so a no-op.
+                --          (3) PP_Cost_Collector component-issue REVERSAL (createComponentIssue, isReversal)
+                --              -> addWeightedAverage on the reversal's own amt/qty. An issue is outbound, so its
+                --                 reversal is qty>0 with ischangingcosts='Y' and no inventory line - it reaches
+                --                 this branch shape - and amt/qty is the price at the ORIGINAL issue, which need
+                --                 NOT equal the current price. That one is a GENUINE average, not a no-op.
+                --        So do NOT read the (2) no-op as licence to add 'p' to the gate. Excluding 'p' wholesale
+                --        is safe for a different and stronger reason: the ONLY fallback is the loud RAISE below,
+                --        never a silently wrong CurrentCostPrice - which holds for every shape, including (3).
+                --        Known residual gap: a 'p' element whose last pre-range detail is any PP_Cost_Collector
+                --        row therefore still RAISEs instead of being reconstructed. Closing it needs a
+                --        (method, document) key, not a method-only one - deliberately out of scope here.
                 --   'L' Lifo / 'F' Fifo / 'U' UserDefined / 'x' ExternalProcessing are likewise excluded.
                 -- Everything excluded falls through to the RAISE below - i.e. exactly the behaviour those
                 -- methods already have today - so this change can never turn a loud abort into a silent
--- a/backend/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/m_costdetail_delete_from_date.sql
+++ b/backend/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/m_costdetail_delete_from_date.sql
@@ -153,15 +153,26 @@ BEGIN
                 --        -> a changing-costs inbound records amt = prevPrice*qty
                 --           (StandardCostingMethodHandler.createCostForMaterialReceipt), so the weighted
                 --           average provably reproduces prevPrice: it is a no-op, not an approximation.
-                --   'p' LastPOPrice / 'i' LastInvoice are DELIBERATELY EXCLUDED: their handlers REPLACE the
-                --        price with amt/qty (LastPOCostingMethodHandler.createCostForMatchPO,
+                --   'p' LastPOPrice / 'i' LastInvoice are DELIBERATELY EXCLUDED: on the M_MatchPO /
+                --        M_MatchInv path their handlers REPLACE the price with amt/qty
+                --        (LastPOCostingMethodHandler.createCostForMatchPO,
                 --        LastInvoiceCostingMethodHandler.createCostForMatchInvoice_MaterialCosts), so averaging
                 --        would silently write a WRONG CurrentCostPrice. Example: prevPrice=20, prevQty=10,
                 --        inbound qty=10 amt=400 -> replace gives 40, averaging gives 600/20 = 30.
+                --        CAVEAT for 'p': ManufacturingLastPOCostingMethodHandler.createMainProductOrCoProductReceipt
+                --        DOES call addWeightedAverage for a PP_Cost_Collector main/co-product receipt - but there
+                --        amt = currentCostPrice*qty, so that average is a no-op (same shape as 'S'). So 'p' behaves
+                --        two ways depending on the DOCUMENT, and a costingmethod-only gate cannot separate them;
+                --        'p' is therefore excluded wholesale, which is safe for BOTH sub-cases. Known residual
+                --        gap: a 'p' element whose last pre-range detail is a PP_Cost_Collector receipt still
+                --        RAISEs instead of being reconstructed. Widening the gate to a (method, document) key
+                --        would close it - deliberately out of scope here.
                 --   'L' Lifo / 'F' Fifo / 'U' UserDefined / 'x' ExternalProcessing are likewise excluded.
                 -- Everything excluded falls through to the RAISE below - i.e. exactly the behaviour those
                 -- methods already have today - so this change can never turn a loud abort into a silent
                 -- mis-valuation.
+                -- A NULL costingmethod (unresolvable cost element) makes the IN test NULL, not TRUE, so it
+                -- also falls through to the RAISE rather than averaging blindly.
                 v_next_costs_found := TRUE;
                 v_next_costs_debugInfo := 'case 2.5: based on prev costs of last cost details (inbound trx, weighted average), with changing costs';
                 v_next_currentqty   := v_firstCostDetail.prev_currentqty   + v_firstCostDetail.qty;
--- a/backend/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/m_costdetail_delete_from_date.sql
+++ b/backend/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/m_costdetail_delete_from_date.sql
@@ -150,9 +150,17 @@ BEGIN
                 --   'A' AveragePO / 'I' AverageInvoice / 'M' MovingAverageInvoice
                 --        -> their handlers call CurrentCost.addWeightedAverage, so the formula matches.
                 --   'S' StandardCosting
-                --        -> a changing-costs inbound records amt = prevPrice*qty
-                --           (StandardCostingMethodHandler.createCostForMaterialReceipt), so the weighted
-                --           average provably reproduces prevPrice: it is a no-op, not an approximation.
+                --        -> INVARIANT: every Standard-costing handler posts amt = currentCostPrice*qty by
+                --           construction - it multiplies the CURRENT cost price of the very same cost
+                --           segment+element by the quantity and never derives a price from the document:
+                --             StandardCostingMethodHandler.createCostForMaterialReceipt /
+                --                 .createOutboundCostDefaultImpl / .createCostForMatchInvoice_MaterialCosts
+                --             ManufacturingStandardCostingMethodHandler.createIssueOrReceipt /
+                --                 .createActivityControl / .createUsageVariance
+                --           Substituting amt = prevPrice*qty into the weighted average gives
+                --           (prevPrice*prevQty + prevPrice*qty) / (prevQty + qty) = prevPrice, so the
+                --           reconstruction provably reproduces prevPrice for EVERY such inbound: it is a
+                --           no-op, not an approximation.
                 --   'p' LastPOPrice / 'i' LastInvoice are DELIBERATELY EXCLUDED: on the M_MatchPO /
                 --        M_MatchInv path their handlers REPLACE the price with amt/qty
                 --        (LastPOCostingMethodHandler.createCostForMatchPO,
--- a/backend/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/m_costdetail_delete_from_date.sql
+++ b/backend/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/m_costdetail_delete_from_date.sql
@@ -159,21 +159,23 @@ BEGIN
                 --        LastInvoiceCostingMethodHandler.createCostForMatchInvoice_MaterialCosts), so averaging
                 --        would silently write a WRONG CurrentCostPrice. Example: prevPrice=20, prevQty=10,
                 --        inbound qty=10 amt=400 -> replace gives 40, averaging gives 600/20 = 30.
-                --        CAVEAT - 'p' is NOT uniform: it has at least three reachable shapes, because the
-                --        manufacturing handler (ManufacturingLastPOCostingMethodHandler) overrides the base one.
-                --          (1) M_MatchPO                       -> REPLACE with amt/qty (base handler, above).
-                --          (2) PP_Cost_Collector main/co-product receipt (createMainProductOrCoProductReceipt)
-                --              -> addWeightedAverage, but with amt = currentCostPrice*qty, so a no-op.
-                --          (3) PP_Cost_Collector component-issue REVERSAL (createComponentIssue, isReversal)
-                --              -> addWeightedAverage on the reversal's own amt/qty. An issue is outbound, so its
-                --                 reversal is qty>0 with ischangingcosts='Y' and no inventory line - it reaches
-                --                 this branch shape - and amt/qty is the price at the ORIGINAL issue, which need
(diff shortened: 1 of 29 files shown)


backend/de.metas.acct.base/src/main/sql/postgresql/ddl/functions/m_costdetail_delete_from_date.sql
```diff
@@ -34,10 +34,21 @@ DECLARE
     v_next_cumulatedamt       numeric;
     v_next_cumulatedqty       numeric;
     v_m_cost_id               numeric;
+    v_costingmethod           char(1);
+    --
+    v_anchor_m_product_id     numeric;
```
teosarca, line 39: pls describe this newly introduced "anchor" term which i saw u used it everywhere. what's the purpose, what's the meaning, what's the concept behind it? be CSP ELI5

teosarca: **Resolution (dev):** Valid point — the term is used 97 times across 12 files in this diff and is defined nowhere. Definition below; it is also being written into the code at the place the anchor is created (`CostRevaluationService`, the `CopyFromCostElement` seed), with a one-line pointer from this function.

**What it is.** When a product is switched to a new costing method with a back-dated cut-off, the `CopyFromCostElement` revaluation writes **exactly one** `M_CostDetail` row, dated at the cut-off (`EvaluationStartDate`), carrying the *previous* method's cost **as of that date** — not its later value. That single row is the new method's **opening balance**.

**Why "anchor".** It is the fixed point the product's cost history under the new method hangs from: every later cost detail is computed forward from it.

**How to recognise one.** `M_CostRevaluationLine_ID IS NOT NULL`, `IsChangingCosts = 'Y'`, and its `Prev_*` fields equal the seeded opening.

**Its purpose — three jobs.**
1. It is the only record of what the opening balance was.
2. It is the row `reverseSeededCurrentCost` deletes to undo the switch.
3. It is the marker the double-seed guard keys on (`CostDetailRepository.retrieveProductIdsWithCostRevaluationSeed`, via `M_CostRevaluationLine_ID IS NOT NULL`), so a product already switched is never re-seeded.

**Why this function guards it.** A recompute reaching back over the cut-off would otherwise delete it, and the loss is silent. Note the damage is *not* what an earlier version of the plan claimed: deleting the anchor does **not** leave the new method zero-based — because the anchor is `IsChangingCosts = 'Y'` and its `Prev_*` equal the seeded opening, Approach 1 restores `M_Cost` to identical numbers. What is actually lost is **the row**: no opening-balance record, no basis for reversing the switch, and job 3 goes blind — so a re-switch would re-seed a cost that is already in use.

teosarca: **Resolution (dev):** Correcting one detail in my reply above: I said a one-line pointer would be added *from this function*. It was not, deliberately — this `ddl/functions/*.sql` body is byte-identical to its shipped migration `5816900`, so a comment-only edit here would cost a fresh migration that re-`DROP`s and re-`CREATE`s the function on every instance purely to move an explanation. Not worth it, and not needed: this file already carries an in-place explanation of the anchor at lines 232-259, right where the guard fires.

The canonical definition landed in Java instead, at the one place the anchor is created — `CostRevaluationService#createDetailsForCopyFromCostElement` — with pointers from `ICostingService#seedCurrentCostFromOpening` and `CostingService`: https://github.com/metasfresh/metasfresh/commit/de32a588b24a81135f9551d221e3e590670415f3

Two things I tightened versus my reply above, because the stricter version is the accurate one:

- **`M_CostRevaluationLine_ID IS NOT NULL` alone does not identify an anchor.** `CostingDocumentRef.ofCostRevaluationLineId` also has a production caller in GL posting (`DocLine_CostRevaluation.java:55`), and the existing comments call that signal deliberately broad and source-agnostic. Recognition is the FK **plus** `IsChangingCosts='Y'`, own `Qty`/`Amt` zero, and `Prev_*` equal to the seeded opening.
- **The double-seed guard is reached through a chain**, not directly: `CostRevaluationService#retrieveAlreadySeededProductIds` → `ICostingService#retrieveProductIdsAlreadySeededOnCostElement` → `ICostDetailService#retrieveProductIdsWithCostRevaluationSeed` → `CostDetailRepository#retrieveProductIdsWithCostRevaluationSeed` (`CostDetailRepository.java:437-460`).

PR #25319, opened 2026-07-25
DD_Order replenishment — aggregate by product group instead of per picking-job assignment

teosarca:
Aggregate replenishment `DD_Order`s by product group instead of one order per picking-job assignment.

Today the `deep_tundra` replenishment path creates one `DD_Order` per `M_Picking_Job_Schedule`, so an operator scans the same product once per assignment. This branch reshapes generation so all open assignments sharing (product, source locator, target locator, UOM) are served by a single `DD_OrderLine`, with the moved quantity attributed back to each contributing assignment.

## Delivered

- **Aggregate-then-split generation**: the request is re-keyed from a single assignment to a product group (`DDOrderReplenishmentGroupKey`); the greedy allocation runs once over the summed group demand, so two deliveries can no longer each claim the same on-hand units. Shares are attributed per contributor, ordered by effective `PriorityRule` → preparation date → older assignment.
- **Association table `DD_OrderLine_PickingJobSchedule`** (line ↔ assignment, with `Qty` / `C_UOM_ID`) + generated model classes, and a partial index `m_picking_job_schedule (c_workplace_id) WHERE processed = 'N'`. This is a pure back-end association table with **no operator-facing window**: it is written and read only by the replenishment service (contributor attribution + watchdog served-ness), never edited by a user, so it ships with no `AD_Window`/`AD_Tab`/`AD_Field`.
- **Effective `PriorityRule` accessor** on `IShipmentScheduleEffectiveBL` (COALESCE of the schedule override and the order's rule) plus a `PriorityRule.HIGH_TO_LOW` comparator, used to order attribution.
- **Change guards widened** to the whole contributor set: an assignment is refused a change once the shared order has goods in transit/delivered or any contributor has been picked; refusals name the assignment, its shipment schedule, the distribution order and the qty already moved.
- **Drift watchdog** re-keyed to the product group; served-ness resolves through the contributor association; one reconcile pass per group needing work, idempotent on re-run.
- **Duplicate grouping keys resolve safely**: the lookup is deterministic, and a duplicate holding goods in transit/delivered is **disconnected (`IsPickingDisconnected=Y`), never voided** — the mover finishes what is in his hands while the group re-plans around it.
- **Single-owner FK columns dropped** from `DD_Order` / `DD_OrderLine`; navigation (shipment schedule ↔ distribution order) now resolves through the association, preserving the existing behaviour.
- **Rollout backfill** (`5816390`): pre-rollout orders get contributor rows so the watchdog's already-served check answers correctly; nothing is re-planned or re-quantified.
- Test surface: cucumber across aggregation, netting, shortfall order, close-out re-issue, group-key change, picker-busy refusal, cross-order split, the rollout backfill, and the duplicate-disconnect case; plus unit coverage of the greedy allocation, the contributor repository and the priority comparator; and a mobile-webui Playwright end-to-end spec proving two same-product orders are offered to the picker as one aggregated job.

## Implementation notes

- **Already-carried quantity is netted off, not planned twice.** Because a duplicate holding moved goods is disconnected rather than voided (above), that quantity is still in the operator's hands. The group's demand is summed **before** the allocation runs, the allocation runs once, and the disconnected lines' already-carried share is then netted off each contributor's remaining demand — so it is never planned a second time. The uncovered-remainder log is reported net of it too, otherwise it would overstate a shortfall the final line quantities do not have.
- **Navigation for the no-UI association table** comes from two new `IRelatedDocumentsProvider` implementations — `ShipmentScheduleDDOrderRelatedDocumentsProvider` (delivery schedule to distribution order) and `DDOrderShipmentScheduleRelatedDocumentsProvider` (the reverse) — which is why the table needs no window of its own. The column-drop migration was layout-checked against a migrated database: no empty UI section, column or element group, no orphaned `AD_UI_ElementField`, and no `SeqNo` gap introduced — both removed UI elements sat last in their element group, at `SeqNo` 990.
- **`isDeadDDOrder` treats an order as gone when it is inactive, `Voided` or `Reversed`.** Reverse-Correct runs the same cleanup as Void but leaves `DocStatus='RE'`, so testing for Void alone would miss it. **`Closed` is deliberately kept**, so navigation from a delivery to an already-completed movement survives.

## Notes

- The mobile completion gate is intentionally **not** included (cut per product decision); completion stays unconditional.
- Every replenishment order (merged or single) carries the org's own `C_BPartner` on its header — an accepted deviation from today's single-contributor behaviour, which changes `DeliveryRule`/print format/print language to the org's.
- No-performance-regression `EXPLAIN` on the index-served path is verified post-rollout, since the served volume only exists after rollout; the read-only verification SQL is prepared and handed over internally.
- Extracting the reconciliation code out of `DDOrderPickingReplenishmentService` is tracked as a follow-up (class-size refactor), not part of this PR. That follow-up also carries the **pre-existing** N+1s left in the class — `disposeCloseOut`'s per-order in-progress-schedules check, and the per-order line retrieval near the top. All five N+1s that this change itself introduced were removed here (batched id lookups).

## Known gap — live on this branch (AC15)

Aggregation introduces one hazard that the per-assignment model did not have: a shared order can now be rewritten by a re-plan triggered by a *different* contributor, including while an operator is mid-move on it (picked from the source locator, not yet dropped — an `IN_PROGRESS` `DD_Order_MoveSchedule`). Three concrete defects follow, none of them reachable on `deep_tundra_release`, where every assignment owns its own order that no other assignment's change can touch:

1. **Grow.** `updateDDOrderLineQtyInPlace` writes a grown quantity unconditionally (only shrinks are ever refused), and nothing regenerates `DD_Order_MoveSchedule` rows on a line-quantity change — so the added quantity gets no move step, the operator drops the old quantity, and the job reads COMPLETED short with `QtyPicked` / `QtyOrdered` left inconsistent.
2. **Shrink.** The only in-place shrink refusal, `isShrinkRefusedByDeliveredQty`, keys on `DD_OrderLine.QtyDelivered > 0`, a column no production flow writes (as the class's own javadoc states) — so it never fires, and an in-flight line can be shrunk below the quantity the operator is physically carrying.
3. **Void.** The re-plan's zero-remaining-chunk branch calls `voidDDOrder` without the `holdsMovedGoods` check every disposal site applies; the `DD_Order` interceptor's `BEFORE_VOID` / `BEFORE_CLOSE` veto `clearSchedules` then throws, and because the reconcile is one transaction the whole pass rolls back — so the newly-arrived demand never gets its order either. The batch-level `findBlockingPickingWork` pre-check does not cover it: an active picker is a strictly narrower signal than a move under way.

A newly-approved acceptance criterion, **AC15**, covers all three: an order under movement is **frozen** — no re-plan grows it, shrinks it or disposes of it. Its quantity counts as serving its contributors, the group's remaining demand is planned onto another source locator or left uncovered and recovered by the existing close-short/re-issue path, and the freeze is transient — it releases itself when the operator drops, and the group converges back. It is therefore *not* the duplicate-disconnect above: the order stays in the group and is simply never written.

**AC15 is not implemented in this PR**, and no follow-up PR is committed to shipping it — a follow-up may implement it if and when it fits. **Residual risk, accepted knowingly:** until AC15 ships, this gap is live wherever this branch runs. A shared order can be grown, shrunk or disposed of by a re-plan while an operator is mid-move on it, with the three consequences described above.


backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/alloc/DDOrderLineContributorRepository.java
```diff
@@ -0,0 +1,200 @@
...
+	/**
+	 * Rewrites the given line's complete contributor set: the line's existing rows are deleted and the given ones
+	 * inserted. An empty list leaves the line with no contributors.
+	 * <p>
+	 * Runs in the caller's transaction and deliberately does NOT open its own — the group reconcile writes the alloc
+	 * rows and the line quantity atomically, and it is the single writer of both.
+	 */
+	public void replaceContributors(@NonNull final DDOrderLineId lineId, @NonNull final List<DDOrderLineContributor> contributors)
+	{
+		deleteByLineIds(ImmutableSet.of(lineId));
```
teosarca: i think u shall load all existing records, index them by M_Picking_Job_Schedule_ID.
update existing
create new if not existing
delete remaining.
aim: minimize DELETEs/INSERTs.

pls eval

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/alloc/DDOrderLineContributorRepository.java
```diff
@@ -0,0 +1,200 @@
...
+				.addOnlyActiveRecordsFilter()
+				.addInArrayFilter(I_DD_OrderLine_PickingJobSchedule.COLUMNNAME_DD_OrderLine_ID, lineIds)
+				.orderBy(I_DD_OrderLine_PickingJobSchedule.COLUMNNAME_M_Picking_Job_Schedule_ID)
+				.create()
+				.stream()
+				.map(DDOrderLineContributorRepository::fromRecord)
+				.collect(ImmutableList.toImmutableList());
+	}
+
+	public ImmutableSet<PickingJobScheduleId> getContributorIds(@NonNull final Collection<DDOrderLineId> lineIds)
```
teosarca: getPickingJobScheduleIds

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/alloc/DDOrderLineContributorRepository.java
```diff
@@ -0,0 +1,200 @@
...
+
+	/**
+	 * The active alloc rows sitting on the given {@code DD_OrderLine}s, for use as a sub-query filter whose
+	 * {@code M_Picking_Job_Schedule_ID} column is the set of assignments those lines serve.
+	 * <p>
+	 * Handed the lines of the live DD_Orders, this is the drift watchdog's "already served" set — the association's
+	 * answer to a question a single-owner back-reference column can only answer for one contributor of a
+	 * consolidated order.
+	 */
+	public IQuery<I_DD_OrderLine_PickingJobSchedule> queryContributorsOfLines(@NonNull final IQuery<I_DD_OrderLine> ddOrderLinesQuery)
```
teosarca: generally if the repo is about "Contributors", you could lose "Contributors" from method names. capture learning!

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/event/DDOrderReplenishmentEventHandler.java
```diff
@@ -44,11 +44,13 @@ public void onEvent(@NonNull final IEventBus eventBus, @NonNull final Event even
 
 		try (final IAutoCloseable ignored = switchCtx(request))
 		{
-			final PickingJobScheduleId pickingJobScheduleId = request.getPickingJobScheduleId();
-
+			// The request identifies the product GROUP, so N assignments of one group collapse into one reconcile and
```
teosarca: pls eval and get rid of AI slop comments in the entire PR

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
```diff
@@ -99,12 +106,31 @@ public class DDOrderPickingReplenishmentService
...
 	@NonNull private final WarehouseRepository warehouseRepository;
+	@NonNull private final DDOrderLineContributorRepository contributorRepository;
 	@NonNull private final IShipmentScheduleBL shipmentScheduleBL = Services.get(IShipmentScheduleBL.class);
 	@NonNull private final IShipmentScheduleEffectiveBL shipmentScheduleEffectiveBL = Services.get(IShipmentScheduleEffectiveBL.class);
 	@NonNull private final IDocTypeDAO docTypeDAO = Services.get(IDocTypeDAO.class);
 	@NonNull private final IHandlingUnitsBL handlingUnitsBL = Services.get(IHandlingUnitsBL.class);
 	@NonNull private final IUOMConversionBL uomConversionBL = Services.get(IUOMConversionBL.class);
 
+	/**
+	 * Refuses a change to a workstation assignment whose product group is already being replenished — the same rule
```
teosarca: wow, cannot read 1 page of AI slop. pls avoid such crap by all means!

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
```diff
@@ -850,10 +1645,175 @@ public void assertWarehouseConfigurationIsValid(@NonNull final I_M_Warehouse war
 		}
 	}
 
-	public boolean isPickerBusy(@NonNull final DDOrderId ddOrderId)
+	/**
+	 * The work a refusal has to name so the reader can go and resolve it: the blocking workstation assignment and the
```
teosarca: AI slop crap, better drop it than writting like this.
but anyways when wrtting class javadoc keep it to 1 paragraph. always ask yourself, is it explaining what the class does and whats the purpose. If thats not clear then the javadoc is pointless.

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
```diff
@@ -828,18 +1585,56 @@ public void rebuildDrift()
...
 
+	/**
+	 * The assignments {@link #rebuildDrift()} currently considers unserved, i.e. the watchdog's input set.
+	 *
+	 * <p>The read-only diagnostic seam over the served-ness predicate: "which deliveries does the watchdog think
+	 * still have no distribution order?" is otherwise only observable indirectly, through the documents a rebuild
+	 * pass happens to produce — and on an already-served group a pass produces nothing at all, so the predicate
+	 * cannot be told apart from an idempotent reconcile by looking at the outcome.</p>
+	 */
+	public ImmutableSet<PickingJobScheduleId> getAssignmentIdsNeedingDDOrder()
```
teosarca: not used. anyways avoid running sets on possible huge amount of data

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
```diff
@@ -8,9 +8,12 @@
 import de.metas.bpartner.BPartnerId;
```
teosarca: this class got way to big. 
consider extracting code like reconciliation to some commands.
make sure SQL N+1 is completelly avoided.
keep in mind we are working on HUGE datasets!

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
```diff
@@ -850,10 +1645,175 @@ public void assertWarehouseConfigurationIsValid(@NonNull final I_M_Warehouse war
...
+	private ImmutableList<PickingJobSchedule> contributorsOf(@NonNull final Set<DDOrderLineId> lineIds)
+	{
+		final ImmutableSet<PickingJobScheduleId> contributorIds = contributorRepository.getContributorIds(lineIds);
+		if (contributorIds.isEmpty())
+		{
+			return ImmutableList.of();
+		}
+
+		return pickingJobScheduleService.getByIds(contributorIds).stream()
+				.sorted(Comparator.comparing(contributor -> contributor.getId().getRepoId()))
```
teosarca: no need to call getRepoId()

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
```diff
@@ -850,10 +1645,175 @@ public void assertWarehouseConfigurationIsValid(@NonNull final I_M_Warehouse war
...
+				.collect(ImmutableList.toImmutableList());
+	}
+
+	private AdempiereException pickerBusyException(@NonNull final DDOrderId ddOrderId, @NonNull final BlockingWork blocking)
```
teosarca: newPickerBusyException

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
```diff
@@ -850,10 +1645,175 @@ public void assertWarehouseConfigurationIsValid(@NonNull final I_M_Warehouse war
...
+				asMessageValue(blocking.getJobScheduleId().getRepoId()),
+				asMessageValue(blocking.getShipmentScheduleId().getRepoId()));
+	}
+
+	/**
+	 * Record ids go into an AD_Message as text, never as a number: the message is rendered through
+	 * {@code MessageFormat}, which would push an {@code Integer} through the reader's locale number format and turn
+	 * the id into "1.234.567" — no longer the value they can search the system for.
+	 */
+	private static String asMessageValue(final int repoId)
```
teosarca: why not RepoIdAware as paraM?

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderReplenishmentGroupKey.java
```diff
@@ -0,0 +1,28 @@
...
+ * <p>The source locator is deliberately NOT part of the key: it is an <i>outcome</i> of the stock-aware
+ * allocation that runs once over the group's summed demand, not an input to the grouping.
+ *
+ * <p>The target locator is the workstation's configured pick-from locator, falling back to the workplace
+ * warehouse's default locator; two workplaces sharing a target locator therefore share one group, which is
+ * intended - the mover's trip is defined by where the goods land.
+ */
+@Value
+@Builder(toBuilder = true)
+public class DDOrderReplenishmentGroupKey
```
teosarca, line 17: can it be package level?

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderReplenishmentGroupKey.java
```diff
@@ -0,0 +1,28 @@
...
+{
+	@NonNull ProductId productId;
+	@NonNull LocatorId locatorToId;
```
teosarca, line 20: why locatorFromId is not part of the groupping key?

teosarca: u wrote huge amount of AI slop crap comments but not what is really relevant.

teosarca: **Resolution (dev):** Deliberate — the source locator is an *output* of the allocation, not an input to the grouping.

The group's demand is summed first, and the existing greedy stock allocation then runs **once** over that sum to decide which source locator(s) cover it (`computeRequiredAllocation`), emitting one `DD_Order` per contributing source locator. So the locator is not known at grouping time; it is what the split produces.

Putting `locatorFromId` into the key would invert that and lose the over-allocation fix: two demand schedules would again be netted independently per locator, which is precisely the double-allocation this change removes (AC6 / deviation D1 of the approved requirements).

The end state you are probably checking for still holds: demand that resolves to different source locators stays on separate documents, one line per locator — "physically distinct picks" is explicitly out of scope for merging.

If you would rather group locator-first, that is a contract-level change to AC6/D1, not a local edit — say so and I will bring the diff rather than change it unilaterally.

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderReplenishmentGroupKey.java
teosarca: **Resolution (dev):** Can't — it is referenced from outside its own package, so package-private would not compile. Verified by grep; three packages use it:

- `de.metas.distribution.ddorder.replenishment` (its own)
- `de.metas.distribution.ddorder.replenishment.event` — the request/converter/handler chain carries the key as its payload identity
- `de.metas.handlingunits.picking.job_schedule.service` — the contributor lookup takes the key as its group selector

A sub-package does not get package-private access either, so even the `event` usage alone rules it out.

Keeping it public. If you want the visibility narrowed, the way to get there is to stop passing the key across those boundaries (e.g. have the event carry the three plain components and reconstruct in-package) — a design change, not a modifier change, so tell me if it is worth it and I will bring the diff.

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
teosarca: **Resolution (dev):** Agreed on all three — the class is 1884 lines and that is indefensible.

Taking it as its own change, deliberately not mixed into the review-comment fixes so the structural diff stays readable:

1. **Extract the reconciliation** out of the service into commands, so the service keeps orchestration only.
2. **N+1**: the per-order `findBlockingPickingWork` re-resolution is the known offender (already logged from an earlier review round as a medium). Going through every query on the reconcile and delete paths, not just that one — the batched-lookup shape is the target everywhere.
3. **Huge datasets**: same pass removes the unbounded in-memory set you flagged separately (`getAssignmentIdsNeedingDDOrder`, ~2100 open assignments and unbounded), and I will check the remaining collectors on those paths for the same problem rather than only that one call site.

Sequencing it after the smaller comment fixes on purpose: those touch the same file, and doing the move first would churn them.

I will not claim this one done until the extraction is committed and the query counts are actually measured rather than asserted.

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/alloc/DDOrderLineContributorRepository.java
teosarca: **Resolution (dev):** Implemented in `DDOrderLineContributorRepository.replaceByLineId` — existing rows are loaded and indexed by `M_Picking_Job_Schedule_ID`, survivors are UPDATEd in place, only genuinely new keys are INSERTed and only departed ones DELETEd. RED first in https://github.com/metasfresh/metasfresh/commit/2fdcf7c66a7ffc65d03ff5d7f5da769fd086a4b2 ("a surviving contributor's alloc row must keep its identity"), turned green by https://github.com/metasfresh/metasfresh/commit/0ba331a1c1edd94f447312a535d2d2f5866abf08. `DDOrderLineContributorRepositoryTest`: 13 tests, 0 failures.

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/alloc/DDOrderLineContributorRepository.java
teosarca: **Resolution (dev):** Renamed to `getPickingJobScheduleIds` in https://github.com/metasfresh/metasfresh/commit/0ba331a1c1edd94f447312a535d2d2f5866abf08.

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/alloc/DDOrderLineContributorRepository.java
teosarca: **Resolution (dev):** Renames applied in https://github.com/metasfresh/metasfresh/commit/0ba331a1c1edd94f447312a535d2d2f5866abf08, following the `<verb>By<Key>` convention the sibling methods already use: `replaceContributors`→`replaceByLineId`, `getContributors`→`getByLineId`, `getContributorsOfLines`→`getByLineIds`, `getContributorIds`→`getPickingJobScheduleIds`, `queryContributorsOfLines`→`queryByLines`. Learning captured in the team corpus as `java-general.md` §21a ("don't restate the class's own subject in method names") including the discriminating-part carve-out — which is why the id-returning method keeps the `PickingJobScheduleIds` qualifier instead of collapsing to a bare `getIds()`.

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/event/DDOrderReplenishmentEventHandler.java
teosarca: **Resolution (dev):** Swept 27 files in the branch diff (Java, `.feature`, migration SQL) — roughly 1490 comment/javadoc lines removed, across https://github.com/metasfresh/metasfresh/commit/9faddc37accca13194f48f03a8b99165bdd75d8f and https://github.com/metasfresh/metasfresh/commit/505e606cc133c3b647cf45e6ec6b145022e17f69. What stayed is a handful of single-line *why* notes for things the code cannot state: the deferrable FKs, the partial index, the module boundary, the empty-`IN` footgun, the `MessageFormat` locale trap, and the pre-departure busy verdict. Two latent defects surfaced during the sweep and are fixed in the same commit: `alreadyCarriesQty`'s javadoc actually described `saveDraftDDOrder`, and `DDOrderLineContributorRepository` still documented the delete-then-insert that its diff-upsert had already replaced. One class javadoc the sweep left standing — the interceptor's — is cut in https://github.com/metasfresh/metasfresh/commit/9e8b5605bf017167cbaafd4aa9b23e4344e44f89.

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
teosarca: **Resolution (dev):** Cut from 18 lines to 1. https://github.com/metasfresh/metasfresh/commit/505e606cc133c3b647cf45e6ec6b145022e17f69

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
teosarca: **Resolution (dev):** Dropped rather than reworded — the class name and its single public entry point carry what the paragraph was restating. https://github.com/metasfresh/metasfresh/commit/505e606cc133c3b647cf45e6ec6b145022e17f69

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
teosarca: **Resolution (dev):** It is used: two cucumber `@Then` steps call it, backing three assertions in `DDOrderReplenishment_aggregate.feature` (lines 146, 164, 177), so deleting it would delete test coverage.

The performance objection is resolved instead. It is now `retainAssignmentsNeedingDDOrder(Set<PickingJobScheduleId>)`: the result is bounded by the caller's own id set, and it short-circuits on empty input before issuing any query — so no set is ever built over the full data volume. https://github.com/metasfresh/metasfresh/commit/505e606cc133c3b647cf45e6ec6b145022e17f69

If you still want it gone, say so and I will drop it together with the three assertions.

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
teosarca: **Resolution (dev):** Dropped from all three comparators — `RepoIdAware` already extends `Comparable`, so `Comparator.comparing(PickingJobSchedule::getId)` sorts directly. https://github.com/metasfresh/metasfresh/commit/505e606cc133c3b647cf45e6ec6b145022e17f69

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
teosarca: **Resolution (dev):** Renamed to `newPickerBusyException`; 5 call sites updated. https://github.com/metasfresh/metasfresh/commit/505e606cc133c3b647cf45e6ec6b145022e17f69

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
teosarca: **Resolution (dev):** `asMessageValue` now takes `RepoIdAware`, per `java-general.md` §6c; 6 call sites pass the typed id instead of unwrapping it. https://github.com/metasfresh/metasfresh/commit/505e606cc133c3b647cf45e6ec6b145022e17f69

teosarca:
**Resolution (dev):** Re-audited every review thread against the current head https://github.com/metasfresh/metasfresh/commit/931dc5240748a48a6c9fade522c4599f37e24cde — all 15 carry a resolution reply, and every file:line each reply points at was re-checked at that head rather than at the head it was written against.

Two threads are waiting on your verdict rather than on further work:

- **Grouping key / `locatorFromId`** (https://github.com/metasfresh/metasfresh/pull/25319#discussion_r3656883560) — the rationale is on that thread and stated in the `DDOrderReplenishmentGroupKey` class javadoc. Grouping locator-first is a contract-level change, so nothing is changed unilaterally.
- **Architecture** (https://github.com/metasfresh/metasfresh/pull/25319#discussion_r3656858036) — the N+1 half is done and enumerated per call site. Still open by your call: the extraction of reconciliation into commands is NOT done (the class is 1612 lines), and one pre-existing N+1 remains — `disposeCloseOut` calls `hasInProgressSchedules` once per order (`DDOrderPickingReplenishmentService.java:1301`, probe at `:1306`). Say whether either lands in this PR.

Landed since the previous version of this comment:

- `PriorityRule.HIGH_TO_LOW` comparator, used by `attributionOrder`, plus the three-key ordering test that was missing — https://github.com/metasfresh/metasfresh/commit/59149e9da187c0c496121a4be52d81789fa4c006
- `DDOrderLowLevelDAO.retrieveLines(Set<I_DD_Order>)`, replacing the per-order fetch at five call sites — https://github.com/metasfresh/metasfresh/commit/99980cf33d3dbcd25812ca44f73bd4d0581f6233
- The contributor diff-upsert is now index-backed, with obsolete rows deleted before the reused rows are saved — https://github.com/metasfresh/metasfresh/commit/f3478bbccbaedf40a33a668d315db603d94c6ea6
- That unique index moved into its own migration script `5816730`; appended to `5816200` it would have been skipped on every database that had already applied 5816200, leaving the invariant unenforced exactly where the table already existed — https://github.com/metasfresh/metasfresh/commit/422e2b716b67af05415ad9e5b5cb7445ddf4812d

One thread you already resolved (public-vs-package visibility of the group key). The remaining 12 are done and are yours to resolve when you agree.

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
```diff
@@ -119,149 +129,254 @@ public void assertCanChange(@NonNull final I_M_Picking_Job_Schedule jobSchedule)
...
+				qtyToPick,
+				candidate.getId().getRepoId());
+		return false;
+	}
+
+	/** The id tiebreak keeps the order stable across passes, so a group whose demand did not change does not rewrite its alloc rows. */
+	private Comparator<PickingJobSchedule> attributionOrder(@NonNull final Map<ShipmentScheduleId, I_M_ShipmentSchedule> schedules)
+	{
+		return Comparator
+				.comparing((final PickingJobSchedule contributor) -> shipmentScheduleEffectiveBL.getPriorityRule(schedules.get(contributor.getShipmentScheduleId())).getCode())
```
teosarca: pls introduce&use to PriorityRule enum an LOW_TO_HIGH or HIGH_TO_LOW Comparator<PriorityRule>

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
```diff
@@ -119,149 +129,254 @@ public void assertCanChange(@NonNull final I_M_Picking_Job_Schedule jobSchedule)
...
+				&& formerContributors.stream().allMatch(PickingJobSchedule::isProcessed);
 	}
 
-	private boolean isOnAutoDistributionOrder(@NonNull final PickingJobSchedule jobSchedule)
+	private ImmutableSet<DDOrderLineId> lineIdsOf(@NonNull final List<I_DD_Order> ddOrders)
 	{
-		final I_M_ShipmentSchedule schedule = shipmentScheduleBL.getById(jobSchedule.getShipmentScheduleId());
-		return isOnAutoDistributionOrder(schedule);
+		return ddOrders.stream()
+				.flatMap(ddOrder -> ddOrderLowLevelDAO.retrieveLines(ddOrder).stream())
```
teosarca: pls introduce to ddOrderLowLevelDAO retrieveLines(Set<I_DD_Order> ddOrders)

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
teosarca: **Resolution (dev):** Adopted. `PriorityRule` gets a named `Comparator<PriorityRule>` constant on the enum itself, based on the enum's own semantic ranking (not `ordinal()` unless the declaration order *is* the documented order), and `attributionOrder` in `DDOrderPickingReplenishmentService` now uses it instead of the inline expression. The three-key ordering (priority rule → preparation date → assignment id) is preserved exactly — this is a refactor, no behaviour change.

Related: that ordering had no test at all. The JUnit `attribution_*` cases feed `attribute(...)` an already-sorted list, and the shortfall cucumber varies only the preparation date with two contributors, so the priority-rule leg and the id tiebreak could both be deleted with the suite still green. AC7 asks for three contributors and all three keys, so a real ordering test is landing together with this change.

Commit link follows in this thread once it is pushed.

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
teosarca: **Resolution (dev):** Adopted. `DDOrderLowLevelDAO` gets `retrieveLines(Set<I_DD_Order> ddOrders)` — a single `addInArrayFilter` query replacing the per-order fetch at the call site, with the empty-input case guarded so an empty set cannot match everything. Return shape is chosen from what the call site actually consumes rather than adding an unused shape, and any ordering the current per-order loop implied is preserved explicitly.

Commit link follows in this thread once it is pushed.

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
teosarca: **Resolution (dev):** Follow-up on the bound: the id set is now pushed into the query instead of being applied in Java after the scan, so the DB does the restricting — `PickingJobScheduleRepository.streamAssignmentsNeedingDDOrder(IQuery, Set)` adds an `addInArrayFilter` on `M_Picking_Job_Schedule_ID` (PickingJobScheduleRepository.java:310, query built at :322-341). https://github.com/metasfresh/metasfresh/commit/4611141f4c5cab539ee3e6538ebc1838ccfb52b9

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
teosarca: **Resolution (dev):** Two halves, different answers.

**Extraction — deferred, per your call. Not done:** `DDOrderPickingReplenishmentService` is still one class (1613 lines, down from 1884) and reconciliation still lives in it.

**N+1 — done.** The 2-query-per-order `isPickerBusy(DDOrderId)` probe is gone entirely; batching goes through new `DDOrderLowLevelDAO.getByIds` (:96), `getLinesByIds` (:381), `retrieveLines(Set<I_DD_Order>)` (:225) and `DDOrderLineContributorRepository.getPickingJobScheduleIdsByLineId` (:128). Per site, queries vs. group size N (line numbers in `DDOrderPickingReplenishmentService`):

- `assertCanChange` (:128) — 3/order (busy probe 2 + lines 1) → 5 flat
- reconcile's pre-mutation busy guard (:454) — 2/order → 4 flat
- `voidAllDDOrders` (:1284), the old per-order `voidDDOrderFor` probe — 2/order → 4 flat
- `voidDDOrdersLeftWithoutContributor` (:1239) — 3 flat for the whole line set, not per line
- `findBlockingPickingWorkByLineId` (:1463) — 3 flat for the whole line set, not per line
- `rebuildDrift` (:1345) — the shipment schedules are one `getByIds` batch, not one load per open assignment

**One N+1 remains and I have not fixed it:** `disposeCloseOut` (:1302) still calls `hasInProgressSchedules` once per order. It predates this branch and I parked it with the deferred extraction — say the word if you want it batched in this PR instead.

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
teosarca: **Resolution (dev):** Landed as `PriorityRule.HIGH_TO_LOW` (PriorityRule.java:55), ranking by `getCode()` — the constants are declared High, Medium, Low, Urgent, Minor, so an `ordinal()`-based comparator would put Urgent fourth; `PriorityRuleTest.highToLow_ranksMostUrgentFirst_notInDeclarationOrder` pins that. `attributionOrder` uses it at DDOrderPickingReplenishmentService.java:383, and the three-key ordering test I owed you is `attributionOrder_ranksByPriorityRule_thenPreparationDate_thenOlderAssignment` (DDOrderPickingReplenishmentServiceGreedyAllocateTest.java:620), fed in reverse of the expected order so no key can pass by accident. https://github.com/metasfresh/metasfresh/commit/59149e9da187c0c496121a4be52d81789fa4c006

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/DDOrderPickingReplenishmentService.java
teosarca: **Resolution (dev):** Landed as `DDOrderLowLevelDAO.retrieveLines(Set<I_DD_Order>)` (DDOrderLowLevelDAO.java:225) — one `addInArrayFilter` query, empty input returns without querying, and it sets the `DD_Order` back-link like the single-order flavour does. Used at five call sites in `DDOrderPickingReplenishmentService` (:154, :406, :473, :1043, :1427); no per-order `retrieveLines` is left in that class. https://github.com/metasfresh/metasfresh/commit/99980cf33d3dbcd25812ca44f73bd4d0581f6233

backend/de.metas.handlingunits.base/src/main/java/de/metas/distribution/ddorder/replenishment/alloc/DDOrderLineContributorRepository.java
teosarca: **Resolution (dev):** Addendum at the current head — the diff-upsert you asked for is now index-backed, and the reuse path had two defects that the first version of it hid.

- A partial unique index `ddorderline_pjs_active_uidx (DD_OrderLine_ID, M_Picking_Job_Schedule_ID) WHERE IsActive='Y'` now enforces the per-pair invariant the settlement code already assumed (`5816200_sys_DD_OrderLine_PickingJobSchedule.sql:349`).
- An index is not deferrable, so the obsolete rows are deleted **before** the reused rows are saved (`DDOrderLineContributorRepository.java:58`); the previous ordering would have failed the reused row's save with 23505 instead of at commit.
- A row whose assignment no longer contributes is now recognised as obsolete in the same pass (`:49`), and two shares of ONE assignment are rejected outright (`extractPickingJobScheduleIds`, `:76`) — the pair carries a single `Qty` column, so summing them in the repository would silently mask a broken attribution.

Four tests pin it: `replaceByLineId_reconcilesAnInactiveRowAndItsActiveSibling_withoutEverHoldingTwoActiveRows`, `_reactivatesAPreExistingInactiveRow_insteadOfInsertingASecondOne`, `_deletesTheInactiveRowOfADepartedContributor`, `_rejectsTwoSharesOfTheSameAssignment` (`DDOrderLineContributorRepositoryTest.java:171,188,203,220`). https://github.com/metasfresh/metasfresh/commit/f3478bbccbaedf40a33a668d315db603d94c6ea6

The UPDATE-survivors / INSERT-new / DELETE-departed shape you asked for is unchanged — this only makes it safe against the pre-existing duplicate rows the missing constraint allowed.

teosarca: **Resolution (dev):** Correction to the index pointer in the reply above — the index moved after that reply was written, and the move fixed a real defect rather than merely relocating a statement.

`ddorderline_pjs_active_uidx` had been appended to `5816200_sys_DD_OrderLine_PickingJobSchedule.sql`, the script that creates the table. Migration idempotency keys on the script NAME, so every database that had already applied script 5816200 skipped the amended file entirely and silently never got the index — the per-pair invariant this diff-upsert relies on was unenforced exactly where the table already existed. It now lives in its own script, `5816730_sys_DD_OrderLine_PickingJobSchedule_active_unique_index.sql:10`, sequenced after the table (5816200) and the row backfill (5816390), and guarded with `IF NOT EXISTS` so a database that already carries it applies the script as a no-op instead of erroring; 5816200 is restored byte-for-byte to its pre-amendment content. https://github.com/metasfresh/metasfresh/commit/422e2b716b67af05415ad9e5b5cb7445ddf4812d

The rest of that reply is unchanged at the current head: obsolete rows are deleted before the reused rows are saved (`DDOrderLineContributorRepository.java:56-58`), a row whose assignment no longer contributes is recognised as obsolete in the same pass (`:45-53`), two shares of ONE assignment are rejected outright (`extractPickingJobScheduleIds`, `:76`), and the four pinning tests are at `DDOrderLineContributorRepositoryTest.java:171,188,203,220`.

PR #25328, opened 2026-07-27
Distribution mobile — fix QR_WRONG_TYPE on auto-advance (conditional huQRCode carry-forward) + clearer HU-vs-GTIN scan prompts

teosarca:
## Problem

In mobile Distribution, after a pick-from completes and the app auto-advances to the next DD order, the scanned source-HU QR code was dropped. The operator was landed on the product-scan screen and — scanning a product GTIN there — got `QR_WRONG_TYPE` (`EAN13HUQRCode`), a dead-end.

## Root cause

`postDistributionPickFromThunk.js` auto-advance navigation omitted `huQRCode`. Carrying it forward *unconditionally* is wrong: when the next order draws from a **different** physical HU (distinct HU per order, even sharing a source locator) the operator must re-scan, so an unconditional carry regresses the packing-table and jobs-list flows.

## Fix

Frontend-only. On auto-advance, carry the scanned `huQRCode` forward to the next order's pick-from screen **only when that next order's pre-allocated move plan draws from the same source HU** (HU-QR identity, not locator). Gated three ways:
- same source HU → carry forward → product-scan continues (the "sweep" flow: one staging LU feeding several orders);
- different source HU → omit → Scan-HU;
- no pre-allocated plan (`allowPickingAnyHU=true`) → omit → Scan-HU (safe default).

The just-picked identity is re-resolved from the actually-scanned barcode (the same resolution the backend's pick uses), because a partial pick overwrites the completed step's own `pickFromHU` to the ephemeral split HU.

## UX hardening

Clearer scan prompts so the operator knows what to scan (the human-error behind this ticket): the step-level pick-from screen now labels its HU-scan and product-scan steps explicitly — `Scan HU barcode` / `HU-Barcode scannen` and `Scan product (GTIN)` / `Artikel (GTIN) scannen`.

## Testing

- New Playwright E2E `sweep_scan_product_after_autoAdvance` reproduces the bug (RED before the fix) and is GREEN after — asserts product-scan continues and the backend end-state.
- Non-regression verified locally (clean-tree baseline) and by the full mobile-webui suite in CI.


e2e/mobile-webui/tests/utils/screens/Backend.js
```diff
@@ -133,6 +133,50 @@ export const Backend = {
...
+     * end-state of a distribution pick when the picked HU has no masterdata identifier — a partial
+     * pick off a shared staging LU creates a NEW split HU only at pick time, so it can only be
+     * referenced via the QR code the job step reports (`Backend.expect`'s HU matcher resolves a raw
+     * QR-code-JSON string as a fallback identifier).
+     *
+     * A distribution job can have several lines (one per DD Order line), so `lineId` is REQUIRED
+     * whenever the job has more than one line; it may be omitted only for a genuinely single-line
+     * job (this throws rather than silently guessing the wrong line's steps otherwise).
+     */
+    getDistributionPickedHUQRCode: async ({ wfProcessId, lineId }) => await test.step(`Backend: get distribution picked HU QR code for wfProcess "${wfProcessId}"${lineId != null ? ` line "${lineId}"` : ''}`, async () => {
```
teosarca: imho this function shall not be there but somewhere in the distribution folder.

e2e/mobile-webui/tests/spec/distribution/sweep_scan_product_after_autoAdvance.spec.js
```diff
@@ -0,0 +1,156 @@
...
+//
+// Mobile distribution profile (same as the mirror spec):
+//   navigateToJobsListAfterPickFromComplete: true — auto-advance to the next order after pick-from
+//   requireScanningProductCode: true              — operator must scan the product GTIN
+//   completeJobAutomatically: true                — auto-complete a job once fully moved
+//   allowStartNextJobOnly: true                   — start jobs strictly in offered order
+//   orderBys: 'Priority, LocatorPriority'         — offer orders sorted by priority, then source-locator priority
+//
+
+const N = 3;
```
teosarca: intention revealing names pls

e2e/mobile-webui/tests/utils/screens/Backend.js
teosarca: **Resolution (dev):** Fixed — relocated the helper out of the generic `Backend.js` into `tests/utils/screens/distribution/DistributionUtils.js` as `DistributionUtils.getPickedHUQRCode` (call site + import in the spec updated). https://github.com/metasfresh/metasfresh/commit/f8e42160462c72dbe69fa2c511997f2ef6b6c904

e2e/mobile-webui/tests/spec/distribution/sweep_scan_product_after_autoAdvance.spec.js
teosarca: **Resolution (dev):** Fixed — renamed `N` → `ORDER_COUNT` (all references in the spec). https://github.com/metasfresh/metasfresh/commit/f8e42160462c72dbe69fa2c511997f2ef6b6c904

PR #25332, opened 2026-07-27
`Transport Order`: auto-default ETD/ETA/ATD/ATA/B-L date from first assigned Purchase Order

riadmen:
On a purchase (inbound) Transport Order (M_ShipperTransportation), default the five date fields when the first Purchase Order is assigned; each value stays user-overridable and the sales flow is left untouched:

- ETD = 1st PO's DatePromised (provisioning date)
- ETA = ETD + vendor delivery time (C_BPartner_Product.DeliveryTime_Promised of the 1st order line's product+vendor)
- ATD = ETD
- ATA  = ETA
- B/L date = ATD


riadmen:
@claude review

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
```diff
@@ -63,18 +66,23 @@
...
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import static org.adempiere.model.InterfaceWrapperHelper.saveRecord;
```
mrie04: This shouldn't be in an service, use at least dao/repo save method. Or alternative update

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
```diff
@@ -135,6 +149,9 @@ public void addPurchaseOrdersToShipperTransportation(@NonNull final ShipperTrans
 		final ImmutableList<OrderId> validPurchaseOrdersIds = availableOrderIds
 				.stream()
 				.filter(orderId -> !alreadyAssignedOrderIds.contains(orderId))
+				// Sort by C_Order_ID so the "first order" that seeds the transport order's default dates is deterministic
+				// even when several purchase orders are assigned in a single call (the DB/set encounter order is not meaningful).
+				.sorted(Comparator.comparingInt(OrderId::getRepoId))
```
mrie04: Are you sure it shouldn't be by prepDate?

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
```diff
@@ -241,6 +267,95 @@ private void addPurchaseOrderLines(final @NonNull I_M_ShipperTransportation ship
...
+	{
+		final BPartnerId vendorId = BPartnerId.ofRepoId(order.getC_BPartner_ID());
+		final OrgId orgId = OrgId.ofRepoId(order.getAD_Org_ID());
+
+		// Intentionally re-query ALL of the order's lines rather than reusing the caller's list: addOrderLinesToShipperTransportation()
+		// passes only a SELECTED subset of lines, and the spec requires the vendor delivery time of the PO's *first* line (lowest Line).
+		// This runs once per transport order (first-order assignment only), so the extra query is negligible.
+		return orderDAO.retrieveOrderLines(order)
+				.stream()
+				.filter(ol -> ol.getM_Product_ID() > 0)
```
mrie04: ProductId.ofRepoIdOrNull() != null

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
```diff
@@ -241,6 +267,95 @@ private void addPurchaseOrderLines(final @NonNull I_M_ShipperTransportation ship
...
+				.min(Comparator.comparingInt(I_C_OrderLine::getLine)) // lowest Line = the PO's first line, independent of DAO sort order
```
mrie04: same as for order, here maybe highest date offset is relevant? for me this isn't clear from re. And why we need to calculate it here as the purchase orders already did calculate datePromised with it in `de.metas.tourplanning.api.impl.OrderDeliveryDayBL#computePreparationDateAndTour0` and this would then also for customers using tour.
AFAIS you also could end up with skipped line data here

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
```diff
@@ -241,6 +267,95 @@ private void addPurchaseOrderLines(final @NonNull I_M_ShipperTransportation ship
...
+	private void applyDefaultDatesFromFirstOrder(@NonNull final I_M_ShipperTransportation shipperTransportation, @NonNull final I_C_Order order)
+	{
+		if (shipperTransportation.isSOTrx())
+		{
+			return; // sales behaviour on the transport order must keep working unchanged
+		}
+
+		// DatePromised is guaranteed non-null here: this runs only for completed/closed purchase orders and the caller
+		// already dereferences order.getDatePromised() when building the base package request, so a null would fail earlier.
+		final Timestamp etd = order.getDatePromised();
```
mrie04: shouldn't this be provisioning so prep?

mrie04 (changes requested):
pls check my comments

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
riadmen: Right: the model treats `DatePromised` as the promised *arrival* date, with `PreparationDate = DatePromised − vendor transport days` (`OrderDeliveryDayBL.computePreparationDateAndTour0`, and `PurchaseCandidateRequestedHandler`: `DateOrdered = DatePromised − leadTime`). My original code set `ETD = DatePromised` and then `ETA = DatePromised + deliveryTime`, which both inverted the departure/arrival direction and double-counted a lead time already baked into the PO's dates. Reworked to:

- `ETA = order.getDatePromised()` (promised arrival)
- `ETD = order.getPreparationDate()` when set — i.e. the PO's own provisioning date, so tour customers get the tour-computed value — falling back to `DatePromised − IOrderBL.getMaxPurchaseTransportDays(order)` when `PreparationDate` is unset.

So the vendor lead time now comes from the PO's own dates rather than being recomputed/re-added.

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
riadmen: Switched from the first line's delivery time to `getMaxPurchaseTransportDays`, which takes the **max across all PO lines** — so it no longer keys off a single (possibly skipped) line. The custom first-line lookup is deleted entirely.

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
riadmen: The line you flagged lived in that deleted method; `getMaxPurchaseTransportDays` already uses `ProductId.ofRepoIdOrNull(...)`.

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
riadmen: Moved to a new `IShipperTransportationDAO.save(I_M_ShipperTransportation)` and call that from the service.

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
riadmen: I kept `C_Order_ID` for determinism: `PreparationDate` is nullable for POs (only populated via tour logic), so ordering the "first order" by it would be non-deterministic when unset.

riadmen:
@claude review

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
mrie04: This comment is for ordering by id, why not earliest date?

backend/de.metas.business/src/test/java/de/metas/shipping/PurchaseOrderToShipperTransportationServiceTest.java
```diff
@@ -542,6 +544,368 @@ public void addPurchaseOrderLines_someLinesSkipped_warnsAndAddsRest()
...
+		final ShipperTransportationId transportationId = ShipperTransportationId.ofRepoId(shipperTransportation.getM_ShipperTransportation_ID());
+
+		final BPartnerLocationId bpartnerAndLocation = createBPartnerAndLocation("VendorDefaults", "addressDefaults");
+		final OrderId orderId = createOrder(bpartnerAndLocation);
+
+		createOrderLine(orderId, StockQtyAndUOMQtys.createConvert(BigDecimal.valueOf(2), product1, uom1), Money.of(10, chf));
+
+		// the PO already carries its provisioning date (PreparationDate) as computed elsewhere; ETD is taken straight from it
+		final I_C_Order order = load(orderId, I_C_Order.class);
+		final java.sql.Timestamp preparationDate = TimeUtil.asTimestamp(LocalDate.of(2019, 6, 1), orgDAO.getTimeZone(OrgId.ofRepoId(order.getAD_Org_ID())));
```
mrie04: fqdy, why not import

mrie04 (approved):
lgtm

PR #25337, opened 2026-07-27
WebUI view auto-invalidation on configured table change

adi-stefan:
## WebUI view auto-invalidation on configured table change

Generic, config-driven mechanism to auto-refresh (websocket push) an aggregate SQL-backed WebUI view when a record of a configured trigger table has an after-commit cache reset.

### Why
An aggregate SQL view (synthetic/hash PK) cannot be mapped by the default per-table view-invalidation advisor: the changed records' ids never match the view's synthetic row ids, so the view never refreshes when its underlying data changes. This adds a config-driven full-invalidate for such views.

### What
- New config table `WEBUI_ViewInvalidateOnChange` — one row = "when a record of `AD_Table_ID` has an after-commit cache reset, full-invalidate all views of `AD_Window_ID`". Empty by default (no-op on a vanilla instance; no customer identifier in core).
- `WebuiViewInvalidateOnChangeRepository` — table-scoped `CCache` reader (IncotermsRepository pattern), self-invalidating when the config table changes.
- `ConfiguredViewInvalidationListener` — an `ICacheResetListener` that, after commit (debounced), full-invalidates the configured window's frontend-watched views via `IViewsRepository.invalidateView`. Piggy-backs the existing after-commit dispatcher, so it reflects committed data even when the triggering work is asynchronous.

### Tests
`ConfiguredViewInvalidationListenerTest` (4 cases): trigger-table match invalidates the configured window's watched views; unrelated table is a no-op.

### Consumers
Config rows are seeded per customer in the consuming customer repo (no customer identifier in core). The first consumer maps a purchasing-overview view (`RV_PurchaseCockpit`) to `M_ReceiptSchedule`, so it auto-refreshes `QtyOrdered` after purchase orders are created — via a separate seed migration in that customer repo.

### DIFF
--- a/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/invalidation/ConfiguredViewInvalidationListener.java
+++ b/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/invalidation/ConfiguredViewInvalidationListener.java
@@ -71,8 +71,8 @@ public class ConfiguredViewInvalidationListener implements ICacheResetListener
 	private static final String SYSCONFIG_DelayInMillis = "webui.ConfiguredViewInvalidationListener.debouncer.delayInMillis";
 	private static final int DEFAULT_DelayInMillis = 100;
 
-	private final ISysConfigBL sysConfigBL = Services.get(ISysConfigBL.class);
-	private final ITrxManager trxManager = Services.get(ITrxManager.class);
+	@NonNull private final ISysConfigBL sysConfigBL = Services.get(ISysConfigBL.class);
+	@NonNull private final ITrxManager trxManager = Services.get(ITrxManager.class);
 
 	private final IViewsRepository viewsRepository;
 	private final WebuiViewInvalidateOnChangeRepository configRepository;
--- a/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/invalidation/ConfiguredViewInvalidationListener.java
+++ b/backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/invalidation/ConfiguredViewInvalidationListener.java
@@ -1,0 +1,172 @@
+/*
+ * #%L
+ * de.metas.ui.web.base
+ * %%
+ * Copyright (C) 2026 metas GmbH
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+package de.metas.ui.web.view.invalidation;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import de.metas.cache.CacheMgt;
+import de.metas.cache.ICacheResetListener;
+import de.metas.cache.model.CacheInvalidateMultiRequest;
+import de.metas.logging.LogManager;
+import de.metas.ui.web.view.IView;
+import de.metas.ui.web.view.IViewsRepository;
+import de.metas.ui.web.window.datatypes.WindowId;
+import de.metas.util.Services;
+import de.metas.util.async.Debouncer;
+import lombok.NonNull;
+import org.adempiere.ad.trx.api.ITrx;
+import org.adempiere.ad.trx.api.ITrxManager;
+import org.adempiere.ad.trx.api.OnTrxMissingPolicy;
+import org.adempiere.service.ISysConfigBL;
+import org.slf4j.Logger;
+import org.springframework.stereotype.Component;
+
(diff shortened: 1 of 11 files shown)


backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/invalidation/ConfiguredViewInvalidationListener.java
```diff
@@ -0,0 +1,172 @@
...
+	private final Debouncer<String> debouncer;
+
+	public ConfiguredViewInvalidationListener(
+			@NonNull final IViewsRepository viewsRepository,
+			@NonNull final WebuiViewInvalidateOnChangeRepository configRepository)
+	{
+		this.viewsRepository = viewsRepository;
+		this.configRepository = configRepository;
+
+		final ISysConfigBL sysConfigBL = Services.get(ISysConfigBL.class);
```
adi-stefan: extract as field

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/invalidation/ConfiguredViewInvalidationListener.java
```diff
@@ -0,0 +1,172 @@
...
+		this.debouncer = Debouncer.<String>builder()
+				.name(ConfiguredViewInvalidationListener.class.getSimpleName() + "-debouncer")
+				.bufferMaxSize(sysConfigBL.getIntValue(SYSCONFIG_BufferMaxSize, 500))
+				.delayInMillis(sysConfigBL.getIntValue(SYSCONFIG_DelayInMillis, 100))
```
adi-stefan: Magic number, please extract as `DEFAULT_...` constant

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/invalidation/ConfiguredViewInvalidationListener.java
```diff
@@ -0,0 +1,172 @@
...
+				.filter(triggerTableNames::contains)
+				.collect(ImmutableSet.toImmutableSet());
+		if (changedTriggerTableNames.isEmpty())
+		{
+			return 0;
+		}
+
+		//
+		// Defer after commit (committed data) + debounce (coalesce bulk changes).
+		final ITrxManager trxManager = Services.get(ITrxManager.class);
```
adi-stefan: as field

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/invalidation/ConfiguredViewInvalidationListener.java
adi-stefan: **Resolution (dev):** Extracted `ISysConfigBL` as a `private final` field (Services.get) instead of a constructor-local lookup. Commit https://github.com/metasfresh/metasfresh/commit/1c497f0b928

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/invalidation/ConfiguredViewInvalidationListener.java
adi-stefan: **Resolution (dev):** Extracted the debouncer defaults as constants: `DEFAULT_BufferMaxSize = 500` and `DEFAULT_DelayInMillis = 100`. Commit https://github.com/metasfresh/metasfresh/commit/1c497f0b928

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/invalidation/ConfiguredViewInvalidationListener.java
adi-stefan: **Resolution (dev):** Extracted `ITrxManager` as a `private final` field (Services.get) instead of a method-local lookup in reset(). Commit https://github.com/metasfresh/metasfresh/commit/1c497f0b928

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/view/invalidation/ConfiguredViewInvalidationListener.java
```diff
@@ -0,0 +1,175 @@
...
+ */
+@Component
+public class ConfiguredViewInvalidationListener implements ICacheResetListener
+{
+	private static final Logger logger = LogManager.getLogger(ConfiguredViewInvalidationListener.class);
+
+	private static final String SYSCONFIG_BufferMaxSize = "webui.ConfiguredViewInvalidationListener.debouncer.bufferMaxSize";
+	private static final int DEFAULT_BufferMaxSize = 500;
+	private static final String SYSCONFIG_DelayInMillis = "webui.ConfiguredViewInvalidationListener.debouncer.delayInMillis";
+	private static final int DEFAULT_DelayInMillis = 100;
```
adi-stefan, line 72: I don't see these 2 constants anywhere else in the repository?
If you use them here, then they must also be created in an SQL using those default value.

adi-stefan: **Resolution (dev):** Added migration `5816590_sys_WEBUI_ConfiguredViewInvalidationListener_debouncer_sysconfig.sql` seeding the two AD_SysConfig rows (`...debouncer.bufferMaxSize=500`, `...debouncer.delayInMillis=100`), mirroring the existing WebsocketSender debouncer seed (`5605840`). ConfigurationLevel=S, EntityType=de.metas.ui.web. Applied + verified locally. Commit https://github.com/metasfresh/metasfresh/commit/32ee2ac5b80

PR #25342, opened 2026-07-28
Packing Instruction — honour validity date in EDI/OLCand GTIN lookup

mrie04:
## What

The EDI / order-candidate (OLCand) product lookup resolved a product's packing instruction (`M_HU_PI_Item_Product`, which carries the CU/TU capacity) purely by GTIN/EAN/UPC, **ignoring the packing instruction's validity window**. When a product's packing instruction is versioned with a future `ValidFrom`, the future instruction was applied **before** its start date — and then propagated onto the order line, shipment schedule, delivery disposition, and picking list.

## Fix

- **`HUPIItemProductDAO`** — extracted a reusable `IHUPIItemProductDAO.createValidOnDateFilter(date)` holding the canonical predicate `ValidFrom <= date AND (ValidTo >= date OR ValidTo IS NULL)` (null date ⇒ match-all, behaviour preserved). The DAO's own query reuses it — no duplication.
- **`ExternalIdentifierProductLookupService.lookupProductByGTIN`** — applies that validity filter to the `M_HU_PI_Item_Product` lookup and orders by `ValidFrom DESC` so that, among the rows valid on the reference date, the one with the latest `ValidFrom` wins.
- **v2 OLCand path** (`JsonConverters` → `MasterdataProvider` → `ProductMasterDataProvider`) — threads the order's promised delivery date (`C_OLCand.DatePromised`, the same date used for price-list-version / price finding) into the lookup, and makes `ProductCacheKey` date-aware so orders with different delivery dates do not share a stale cached result.

## Behaviour

- Before a version's `ValidFrom` the currently-valid instruction is used; on/after, the new one. No `ValidTo` needs to be set manually on the old row.
- Non-OLCand GTIN lookups (warehouse, BOM, purchase candidates) pass no date and consume only the resolved product id, so their outcome is unchanged.

## Testing

- `ExternalIdentifierProductLookupServiceTest` (new) — reproduces the pre-switch wrong selection (RED) and verifies correct selection before/after the switch plus the latest-`ValidFrom` tie-break, and that the cache key is date-aware.
- `HUPIItemProductDAOTest` — unchanged and green (the extraction is behaviour-preserving).
- Full `de.metas.business.rest-api-impl` unit suite green (158 tests, 0 failures).


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/ordercandidate/spi/impl/OLCandProductFromPIIPvalidator.java
```diff
@@ -78,6 +90,61 @@ else if (productId.getRepoId() != huPIItemProduct.getM_Product_ID())
...
+	 * barcode-lookup view that picks the newest-created row regardless of its validity window, so a
+	 * future-dated instruction can otherwise be applied before its start date. This validator has the
+	 * lowest {@link #getSeqNo() seqNo}, so the corrected instruction is in place before pricing reads it.
+	 * <p>
+	 * A user-set {@code M_HU_PI_Item_Product_Override_ID} is never touched. Idempotent for the REST path
+	 * (already resolved on {@code DatePromised}); a no-op when no row is valid on the date.
+	 */
+	private void resolveValidPackingInstructionForDatePromised(@NonNull final I_C_OLCand olCand)
+	{
+		if (olCand.getM_HU_PI_Item_Product_Override_ID() > 0)
```
mrie04: use HUPIItemProductId.ofRepoId.ornull

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/ordercandidate/spi/impl/OLCandProductFromPIIPvalidator.java
```diff
@@ -78,6 +90,61 @@ else if (productId.getRepoId() != huPIItemProduct.getM_Product_ID())
...
+		// legitimate selection made among several instructions that share one barcode (e.g. the EDI
+		// lookup view's per-BPartner/StoreGLN choice); we only correct one that is NOT valid on the date.
+		if (isValidOnDatePromised(current, datePromised))
+		{
+			return;
+		}
+
+		huPIItemProductDAO.findFirstByGtin(GTIN.ofString(barcode), datePromised)
+				.map(ProductAndHUPIItemProductId::getHupiItemProductId)
+				.filter(validId -> !validId.equals(currentId))
```
mrie04: de.metas.handlingunits.HUPIItemProductId#equals(de.metas.handlingunits.HUPIItemProductId, de.metas.handlingunits.HUPIItemProductId)

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/ordercandidate/spi/impl/OLCandProductFromPIIPvalidator.java
```diff
@@ -78,6 +90,61 @@ else if (productId.getRepoId() != huPIItemProduct.getM_Product_ID())
...
+				.ifPresent(validId -> olCand.setM_HU_PI_Item_Product_ID(validId.getRepoId()));
+	}
+
+	private static boolean isValidOnDatePromised(@NonNull final I_M_HU_PI_Item_Product piip, @NonNull final ZonedDateTime datePromised)
+	{
+		final Instant date = datePromised.toInstant();
+		final Instant validFrom = TimeUtil.asInstant(piip.getValidFrom());
+		final Instant validTo = TimeUtil.asInstant(piip.getValidTo());
+		return (validFrom == null || !validFrom.isAfter(date))
```
mrie04: validFrom should be mandatory so never null

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/ordercandidate/spi/impl/OLCandProductFromPIIPvalidator.java
mrie04: **Resolution (dev):** Done — the override check now uses the value-object idiom `HUPIItemProductId.ofRepoIdOrNull(olCand.getM_HU_PI_Item_Product_Override_ID()) != null` instead of the raw `> 0`. Applied in the working tree; pushing after the in-flight local unit+E2E verification (will follow up with the commit SHA).

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/ordercandidate/spi/impl/OLCandProductFromPIIPvalidator.java
mrie04: **Resolution (dev):** Done — changed to `!HUPIItemProductId.equals(validId, currentId)` (the static null-safe equals). Applied in the working tree; pushing after local verification (SHA to follow).

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/ordercandidate/spi/impl/OLCandProductFromPIIPvalidator.java
mrie04: **Resolution (dev):** Resolved by refactor — the hand-rolled validity check (which carried the `validFrom == null` guard) was removed. Validity now goes through a new `IHUPIItemProductDAO.isValidOnDate(id, date)` that reuses the DAO's own `createValidOnDateFilter`, so the mandatory `ValidFrom`/`ValidTo` are compared in the DB exactly as `findFirstByGtin` does — no hand-rolled/timezone-sensitive `Instant` comparison. Applied in the working tree; pushing after local verification (SHA to follow).

PR #25348, opened 2026-07-28
EDI export cucumber: port the basicGet pull-loop fix for the ack-vs-channel-close race (test-only; unblocks PR 25319)

teosarca:
## What

Ports the **already-merged `new_dawn_uat` test-harness fix** https://github.com/metasfresh/metasfresh/commit/4991c8e7a5966b4c4a0bfb4eb29dfa92490f48c5 ("EDI export cucumber: drop racy push-consumer for a synchronous basicGet pull-loop", https://github.com/metasfresh/metasfresh/pull/24612) to `deep_tundra_release`, which never received it.

Straight `git cherry-pick -x` — **no rewrite, no local variant**. Both Java files are now byte-identical to `origin/new_dawn_uat`, so the eventual `deep_tundra_release → new_dawn_uat` propagation merge is a no-op on them.

**Test-only.** Zero production classes touched. Diff is 3 files, all under `src/test`:

- `backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/edi/MetasfreshToEDIRabbitMQ_StepDef.java`
- `backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/externalsystem/MetasfreshToExternalSystemRabbitMQ_StepDef.java`
- `backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/edi/ediExport.feature`

## Root cause (harness, not product)

`pollDocumentFromQueue` used an asynchronous push consumer (`basicConsume` + `DefaultConsumer.handleDelivery`) with **no prefetch limit**, and inside `handleDelivery` called `countDownLatch.countDown()` **before** `channel.basicAck(...)` / `channel.basicCancel(...)`. The countdown releases the main thread parked on `countDownLatch.await(60s)`, whose `finally` runs `channel.close()` — so the ack can be lost, the broker **requeues** the delivery, and the next `pollDocumentFromQueue` re-reads the **same** document. The step binds whatever it polled to the identifier named in the data-table row (arrival-order binding), so the wrong document lands under the wrong identifier.

### Log evidence

Failing job: https://github.com/metasfresh/metasfresh/actions/runs/30362621686/job/90294199165 — `test (cucumber) (profile5)`, scenario `S29231_150 — Two orders, one consolidated shipment → both DESADVs exported via the Replication Interface (RPL) path` (`ediExport.feature:264`), failing at `:437`:

```
org.opentest4j.AssertionFailedError: [QtyCUsPerTU]
expected: "20"
 but was: "10"
	at de.metas.cucumber.stepdefs.edi.EDI_Desadv_StepDef.validateEDIExpDesadvPack(EDI_Desadv_StepDef.java:262)
```

From that run's raw profile5 log, inside the `S29231_150` window:

- exactly **two** `*** Queue: ediExport_150, received message:` dumps, **10.7 ms apart** (`13:55:28.9765`, `13:55:28.9872`);
- `POReference>PO_A_S29231_150` appears **4×**, `POReference>PO_B_S29231_150` appears **0×** (each DESADV XML carries its POReference twice ⇒ 2 dumps × 2 = 4, all of them **A**) ⇒ both polls received **DESADV A**; B's message was never received;
- immediately after the second dump: `[WARN] MetasfreshToEDIRabbitMQ_StepDef - basicCancel failed (ignored): channel is already closed due to clean channel shutdown; protocol method: #method<channel.close>(reply-code=200, reply-text=OK, …)` — the channel was already closed before the consumer finished its ack/cancel sequence, i.e. the lost ack.

So `e_dA_150` = A (passes, `QtyCUsPerTU=10`) and `e_dB_150` = **A again** → `expected: "20" but was: "10"`, exactly as observed. The feature deliberately gives A `QtyCUsPerTU=10` and B `QtyCUsPerTU=20` to prove per-source-DESADV projection, which is why the mis-bind surfaces with those precise numbers.

**Production exported both DESADVs correctly** — the harness lost one and double-counted the other. The sibling surface recorded on three branches (`com.rabbitmq.client.ShutdownSignalException: clean channel shutdown … Closed due to exception from Consumer … handleDelivery`) is the identical close-vs-callback collision seen from the AMQP client side.

## The fix (as authored upstream)

- `pollDocumentFromQueue` / `pollRequestFromQueue` rewritten as a **synchronous `basicGet` pull-loop on the test thread**: one message per `basicGet`, `basicAck` issued **before** the method returns and before the channel is closed. No consumer-callback thread ⇒ no prefetch storm, no ack-vs-close race, no requeue, so the redelivery that mis-binds the documents cannot occur. This is a strictly stronger harness, not a relaxed one — no assertion was weakened, none skipped, none disabled.
- Foreign/unparseable leftovers on the shared durable queue are acked-and-skipped (with the body logged) instead of crashing the consumer or being returned as the wrong document.
- New Background isolation step `the EDI export RabbitMQ queue '<name>' is purged`, applied to `ediExport` and `ediExport_150`, so each scenario starts from a known-empty queue regardless of what an earlier scenario/feature on the same executor left behind. No-op when the queue does not exist yet.
- Channel/connection lifecycle tightened: connection closed in `finally` (the old code leaked it), `channel.close()` guarded by `isOpen()` so a broker force-close cannot suppress the real failure, `Thread.sleep` restores the interrupt flag.

## Why this must merge before https://github.com/metasfresh/metasfresh/pull/25319 can go green

This flake is **the only thing keeping https://github.com/metasfresh/metasfresh/pull/25319 red.** That PR is *not* test-only, so it cannot use the flaky-red carve-out — the harness fix has to land on its base branch `deep_tundra_release` first.

Evidence the flake is not caused by that PR: the same scenario fails on the **diff-free base** `deep_tundra_release` (runs https://github.com/metasfresh/metasfresh/actions/runs/30009980484 and https://github.com/metasfresh/metasfresh/actions/runs/30343940854), the PR touches zero EDI/DESADV files, and its own immediately-preceding commit `39ec8ea0aef` was cucumber-green with the full production diff already in place (https://github.com/metasfresh/metasfresh/actions/runs/30360221405).

**⇒ Please merge this PR, then re-run CI on https://github.com/metasfresh/metasfresh/pull/25319.**

## Tracking

Internal flaky-test registry, case 34 (verified mechanism + fix recipe; 7 scenario-failures across 6 runs on 3 branches, intermittency proven by same-SHA green re-runs). The registry also records a still-unverified second surface (`[EDI_Exp_Desadv_Pack] Expecting actual not to be null` on `ediExport.feature:175` / `:424`, an async pack-materialization hypothesis); the Background queue-purge plus the skip-foreign-message tolerance in this port address the cross-scenario-bleed half of that surface, and the case will be updated with this PR's CI outcome.

## Validation

- Ported by `git cherry-pick -x` and verified byte-identical to `origin/new_dawn_uat` for both step-def classes; the only feature-file delta vs `new_dawn_uat` is `new_dawn_uat`-only extra GLN assertions this branch does not carry (the purge lines are identical).
- Checked on this branch: no duplicate cucumber step expression for the new purge step; queue names `ediExport` / `ediExport_150` match the feature's `routingKey` parameters; `amqp-client 5.10.0` (`misc/parent-pom/pom.xml`) provides `Channel#basicGet` / `GetResponse`; the removed `DefaultConsumer` / `Envelope` / `CountDownLatch` imports have no remaining references; every checked exception thrown in the new code is declared by its method.
- `pollRequestFromQueue` behaviour is preserved: the pre-port version consumed with `autoAck=true`, so non-qualifying messages were already removed — the new ack-and-skip is equivalent, and the count assertion was tightened (`isEqualTo` instead of `>=`), not loosened.
- **Local execution was not possible** for this change: no Maven build and no local stack were available in this worktree (the workspace's `.m2-local` is exclusively owned by a concurrent build in the sibling checkout, and running a second build would have raced it). Verification was therefore static, plus the strongest available runtime evidence: this exact code has been running on `new_dawn_uat` since 2026-06-18 and **every** recorded occurrence of this flake since then is on a branch that lacks it (`deep_tundra_release`, `soft_panda_hotfix`, `intensive_care_hotfix`) — none on `new_dawn_uat`. CI on this PR is the runtime gate.

## Residual, reported honestly

The pull-loop removes the *redelivery* that caused the observed mis-bind, but the step still binds the polled document to the data-table identifier **by arrival order**, not by content. If the two independent RPL export workpackages ever published B before A, `S29231_150` would fail as `expected: "10" but was: "20"` on the first pack row. That shape has **never** been observed (all recorded failures are the duplicate-A shape). Content-matching the polled document against its expected identifier is deliberately **not** added here: it would fork this file away from `new_dawn_uat` and conflict on the propagation train. If that shape ever appears, the content-match belongs upstream on `new_dawn_uat` first and then ports down.

### DIFF
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/edi/MetasfreshToEDIRabbitMQ_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/edi/MetasfreshToEDIRabbitMQ_StepDef.java
@@ -26,8 +26,7 @@ import com.rabbitmq.client.AMQP;
 import com.rabbitmq.client.Channel;
 import com.rabbitmq.client.Connection;
 import com.rabbitmq.client.ConnectionFactory;
-import com.rabbitmq.client.DefaultConsumer;
-import com.rabbitmq.client.Envelope;
+import com.rabbitmq.client.GetResponse;
 import de.metas.CommandLineParser;
 import de.metas.ServerBoot;
 import de.metas.cucumber.stepdefs.DataTableUtil;
@@ -37,6 +36,7 @@ import de.metas.esb.edi.model.I_EDI_cctop_invoic_v;
 import de.metas.logging.LogManager;
 import de.metas.util.Services;
 import io.cucumber.datatable.DataTable;
+import io.cucumber.java.en.Given;
 import io.cucumber.java.en.Then;
 import lombok.NonNull;
 import org.adempiere.ad.dao.IQueryBL;
@@ -54,7 +54,6 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
@@ -91,6 +90,74 @@ public class MetasfreshToEDIRabbitMQ_StepDef
 		metasfreshToRabbitMQFactory.setPort(commandLineOptions.getRabbitPort());
 		metasfreshToRabbitMQFactory.setUsername(commandLineOptions.getRabbitUser());
 		metasfreshToRabbitMQFactory.setPassword(commandLineOptions.getRabbitPassword());
+	}
+
+	/**
+	 * Purges the given durable EDI-export RabbitMQ queue so the scenario starts from a known-empty
+	 * state, regardless of what an earlier scenario (or an earlier feature on the same CI executor)
+	 * left behind.
+	 * <p>
+	 * Why this is needed: the EDI export queues (e.g. {@code ediExport}) are <b>durable</b> and
+	 * <b>shared</b> by every scenario in a feature and by every feature that exports through the same
+	 * routing key on one CI executor (all run sequentially against one RabbitMQ broker). The consumer
+	 * in {@link #pollDocumentFromQueue(String)} returns the message at the head of the queue, so a
+	 * single leftover message (a foreign DESADV/INVOIC from a sibling scenario, or a message re-queued
+	 * by the consumer's {@code basicNack(requeue=true)} extra-message path) makes the next scenario
+	 * consume the <i>wrong</i> document. That manifests as the flaky
+	 * {@code [EDI_Exp_Desadv_Pack] Expecting actual not to be null} (a foreign INVOIC has no pack
+	 * element) and the sibling {@code ShutdownSignalException}. Purging in the Background (per the
+	 * setUp-isolation convention) removes the bleed at its source.
+	 * <p>
+	 * The purge is a no-op when the queue does not exist yet (first run on a fresh broker): we
+	 * passively check existence first and skip silently, because "no queue" already means "nothing to
+	 * bleed".
+	 * <p>
+	 * DataTable-free step. Example:
+	 * <pre>
+	 * Given the EDI export RabbitMQ queue 'ediExport' is purged
+	 * </pre>
+	 */
+	@Given("the EDI export RabbitMQ queue {string} is purged")
+	public void edi_export_queue_is_purged(@NonNull final String queueName) throws IOException, TimeoutException
+	{
+		final Connection connection = metasfreshToRabbitMQFactory.newConnection();
+		try
+		{
+			final Channel channel = connection.createChannel();
+			try
+			{
+				try
+				{
+					// queueDeclarePassive throws (and closes the channel) when the queue does not exist yet;
+					// in that case there is nothing to purge, so we treat it as a no-op and return. The dead
+					// channel needs no explicit close (the broker already closed it; the finally below guards
+					// it via isOpen()), and the outer finally closes the connection.
+					channel.queueDeclarePassive(queueName);
+				}
+				catch (final IOException queueAbsent)
+				{
+					logger.info("EDI export queue {} does not exist yet -> nothing to purge", queueName);
+					return;
+				}
+
+				// queuePurge is OUTSIDE the queueAbsent catch on purpose: the queue exists here, so a purge
+				// failure (e.g. ACCESS_REFUSED, or a protocol error) is a real problem and must propagate
+				// rather than be mislogged as "does not exist yet" and silently leave the queue unpurged.
+				final AMQP.Queue.PurgeOk purgeOk = channel.queuePurge(queueName);
+				logger.info("Purged {} message(s) from EDI export queue {}", purgeOk.getMessageCount(), queueName);
+			}
+			finally
+			{
+				if (channel.isOpen())
+				{
+					channel.close();
+				}
+			}
+		}
+		finally
+		{
+			connection.close();
+		}
 	}
 
 	@Then("RabbitMQ receives a EDI_cctop_invoic_v")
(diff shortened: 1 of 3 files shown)


teosarca:
## CI status — the fix's target job is green; the one red is an unrelated mobile spec

Run https://github.com/metasfresh/metasfresh/actions/runs/30382245339 (commit `2abc3a3fbdc`), attempt 1.

**`test (cucumber) (profile5)` — the job this PR exists for — PASSED** (19m58s): https://github.com/metasfresh/metasfresh/actions/runs/30382245339/job/90360413778. From that run's own `junit-results-cucumber-profile5` artifact: `<testsuite name="Cucumber" tests="312" failures="0" errors="0" skipped="0">`, with all three `ediExport` scenarios present in the testcase list and green — `S29231_150 — Two orders, one consolidated shipment → both DESADVs exported via the Replication Interface (RPL) path`, `As a user I want to export EDI_Exp_Desadv`, and `As a user I want to export C_Invoice using EDI_cctop_invoic_v export format`. `skipped=0` is stated explicitly because a green job would otherwise also be consistent with the scenarios never having run.

Also green on this commit: cucumber `profile1`, `profile2`, `profile3`, `profile4`, `profile6`, `profile7`, `catchall`, `report`; `test (java)`; `health check`; and all three `frontend webui test` shards.

### The single red: `mobile test`

Job: https://github.com/metasfresh/metasfresh/actions/runs/30382245339/job/90358879721 — **1 failed / 279 passed**.

The failing spec in **this** run is `e2e/mobile-webui/tests/spec/picking/picking_catchWeight.spec.js:452` › `EAN13 with prefix 29 and not matching product`:

```
› PickingJobsListScreen - Start job by documentNo 0078 › PickingJobScreen - Wait for screen
TimeoutError: locator.waitFor: Timeout 20000ms exceeded.
Call log:
  - waiting for locator('#WFProcessScreen') to be visible
  at e2e/mobile-webui/tests/utils/screens/picking/PickingJobScreen.js:25
```

i.e. the mobile job-start screen gate, reached via `PickingJobsListScreen.startJob`.

### Why this red is not caused by this PR

1. **This PR cannot reach mobile code.** Its entire diff is 3 files, all under `backend/de.metas.cucumber/src/test/` — two cucumber step-def classes and one `.feature` file. Zero `e2e/mobile-webui` files, zero frontend files, zero production classes.
2. **The identical failing step appears on an unrelated PR, at a different commit, with a completely different diff.** https://github.com/metasfresh/metasfresh/pull/25319 (branch `deep_tundra_release_DDOrderAggregateByProduct`, commit `49a9a0934a9`) run https://github.com/metasfresh/metasfresh/actions/runs/30382458509 attempt 1, job https://github.com/metasfresh/metasfresh/actions/runs/30382458509/job/90359804531 — 1 failed / 279 passed, failing at the **same** `PickingJobsListScreen - Start job by documentNo … › PickingJobScreen - Wait for screen` step with the same `TimeoutError: locator.waitFor: Timeout 20000ms exceeded`.

   To be precise, and verified from each run's own `junit-results-playwright-mobile` artifact rather than assumed: it is a **different spec** there — `e2e/mobile-webui/tests/spec/picking/picking.spec.js:346` › `Picking Job Completion › Should succeed when partial picking and allowCompletingPartialPickingJob = Y`. What the two share is the failing **step and seam**, not the spec.

   Two unrelated diffs failing at the same mobile job-start seam in the same CI wave points at base-branch / harness state, not at either change.
3. **The base commit of this branch passed `mobile test`.** Base `deep_tundra_release` at `b94f261363a` — the commit this branch is cut from — was green in run https://github.com/metasfresh/metasfresh/actions/runs/30381002550, so the seam is intermittent rather than newly broken.

### Honest caveat about "known flake"

This particular spec (`picking_catchWeight.spec.js`) is **newly observed today and is not currently a registered known flake** — please do not treat it as pre-documented. The *signature* (mobile `#WFProcessScreen` job-start gate timing out) is a recurring mobile-webui job-start race that has been seen in other specs, but this spec's occurrence is new and is being triaged separately. It is called out here explicitly so the merge decision is not resting on an assumption that it was already documented.

A re-run of the failed job is already in flight (attempt 2); no further re-runs are being triggered from my side.

### Side observation that speaks to this PR's effect

On the **same** CI wave and the **same** base branch, with an identical scenario set (`tests=312` in both), the two branches diverge exactly as expected:

| branch | carries this fix? | `test (cucumber) (profile5)` |
|---|---|---|
| `deep_tundra_release_FixEdiDesadvPollAckRace` (this PR) | yes | **312 / 312 passed**, `failures=0` |
| `deep_tundra_release_DDOrderAggregateByProduct` (https://github.com/metasfresh/metasfresh/pull/25319) | no | **`failures=1`** — `S29231_150`, `clean channel shutdown; protocol method: #method<channel.close>(reply-code=200, reply-text=Closed due to exception from Consumer … method handleDelivery …)` |

That failure message is precisely the ack-vs-channel-close surface this PR removes: the consumer callback acting on a channel the poll method's `finally` had already closed. Read from each run's own `junit-results-cucumber-profile5` artifact.

PR #25358, opened 2026-07-29
Return pricing: resolve origin-line Packvorschrift so returns of PI-keyed-priced products invoice correctly

mrie04:
## Problem

A product **return** that references its original document fails to invoice — the invoice candidate errors **"Produkt nicht auf der Preisliste"** — when the product's active price is keyed on a **Packvorschrift** (`M_HU_PI_Item_Product_ID`). The only workaround was a manual 0.00 price-list entry.

## Root cause

For a return, `M_InOutLine_Handler` prices from the **origin document line** (`getReturn_Origin_InOutLine()`), to reproduce the original price at the original date. A vendor **receipt** line carries no Packvorschrift in any of its columns — the PI lives on the linked **purchase-order line** (receipt lines are not PI-stamped, unlike HU shipment lines). `HUPricing.getPackingMaterialId()` reads only the line's own columns → null → the PI-keyed price is never matched → "Produkt nicht auf der Preisliste".

(The single-price case is masked by `UniqueProductPriceFallbackRule`; the defect surfaces when the product has more than one active price.)

**Verified (local cucumber + DB):** the new `vendorReturn.feature` scenario reproduces it — before the fix the return's `C_Invoice_Candidate` is `IsError=Y` and its origin receipt line has `M_HU_PI_Item_Product_ID` (base/override/calculated) all `NULL` while the linked order line carries the PI (confirmed by `psql` on the local stack). After the fix the same scenario is GREEN (`IsError=N`, `PriceActual` = the original PO price). The receipt/order candidate priced correctly throughout, isolating the defect to the return→origin pricing path.

## Fix (return-scoped)

- `IPricingContext` / `IEditablePricingContext` / `PricingContext`: an optional explicit Packvorschrift carrier (default none; carried across `copy()`).
- `HUPricing.getPackingMaterialId`: honor the explicit Packvorschrift when set (and regular), otherwise the existing referenced-object extraction — unchanged when not set.
- `InOutBL` / `IInOutBL`: a `getProductPrice(inOutLine, explicitPackingInstruction)` overload; the existing single-arg method delegates with `null`.
- `M_InOutLine_Handler`: at both origin-line pricing sites (candidate creation + recompute), when the line is a return (`Return_Origin_InOutLine_ID > 0`), resolve the origin line's order-line Packvorschrift and pass it explicitly. Non-return pricing passes `null` and is byte-for-byte unchanged; returns without an order line stay unchanged.

**Direct shipment/receipt invoicing (incl. shipment-IC and mixed-PI shipment lines) is untouched** — the explicit value is set only on the return→origin path.

## Scope

The defect is **vendor-return-specific**: customer returns price from an origin **shipment** line, which *is* PI-stamped by `ShipmentLineBuilder`, so they already price correctly (confirmed: the customer-return scenario passes before and after the fix).

## Tests (cucumber, `de.metas.cucumber`)

- **Vendor return** scenario in `purchaseOrder/vendorReturn.feature` — RED before the fix, GREEN after. Reproduction + regression test.
- **Customer return** scenario in `shipment/customerReturn.feature` — passes before and after (regression guard).
- `C_Invoice_Candidate_StepDef`: optional `IsError` / `ErrorMsg` / `PriceActual` assertions + an optional `M_Product_ID` filter on the credit-memo-candidate lookup.

No AD migration, no data backfill.

### DIFF
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
@@ -624,6 +624,15 @@ public class C_Invoice_Candidate_StepDef
 									softly.assertThat(finalInvoiceCandidate.getC_Project_ID()).as("C_Project_ID").isEqualTo(project.getC_Project_ID());
 								});
 
+						row.getAsOptionalBoolean(I_C_Invoice_Candidate.COLUMNNAME_IsError)
+								.ifPresent(expected -> softly.assertThat(finalInvoiceCandidate.isError()).as("IsError").isEqualTo(expected));
+
+						row.getAsOptionalString(I_C_Invoice_Candidate.COLUMNNAME_ErrorMsg)
+								.ifPresent(expected -> softly.assertThat(finalInvoiceCandidate.getErrorMsg()).as("ErrorMsg").contains(expected));
+
+						row.getAsOptionalBigDecimal(I_C_Invoice_Candidate.COLUMNNAME_PriceActual)
+								.ifPresent(expected -> softly.assertThat(finalInvoiceCandidate.getPriceActual()).as("PriceActual").isEqualByComparingTo(expected));
+
 						softly.assertAll();
 					}
 					catch (final Throwable e)
@@ -718,13 +727,40 @@ public class C_Invoice_Candidate_StepDef
 
 	}
 
+	/**
+	 * Loads the invoice candidate for the given return inout, optionally filtered by product.
+	 *
+	 * <p>Required columns:
+	 * <ul>
+	 *   <li>{@code M_InOut_ID.Identifier} – identifier of the vendor/customer return inout</li>
+	 *   <li>{@code C_Invoice_Candidate_ID.Identifier} – identifier under which to store the found IC</li>
+	 * </ul>
+	 *
+	 * <p>Optional columns:
+	 * <ul>
+	 *   <li>{@code OPT.M_Product_ID.Identifier} – when present, filters ICs by this product; required when
+	 *       the return generates multiple ICs (e.g. packing-material lines alongside the product line)</li>
+	 * </ul>
+	 *
+	 * @return {@code true} if exactly one matching IC was found and stored; {@code false} if none found yet
+	 */
 	private boolean loadCreditMemoCandidate(@NonNull final Map<String, String> row)
 	{
 		final String customerReturnIdentifier = DataTableUtil.extractStringForColumnName(row, I_M_InOut.COLUMNNAME_M_InOut_ID + "." + TABLECOLUMN_IDENTIFIER);
 		final int customerReturnId = shipmentTable.get(customerReturnIdentifier).getM_InOut_ID();
 
-		final Optional<I_C_Invoice_Candidate> invoiceCandidate = queryBL.createQueryBuilder(I_C_Invoice_Candidate.class)
-				.addEqualsFilter(I_C_Invoice_Candidate.COLUMNNAME_M_InOut_ID, customerReturnId)
+		final IQueryBuilder<I_C_Invoice_Candidate> queryBuilder = queryBL.createQueryBuilder(I_C_Invoice_Candidate.class)
+				.addEqualsFilter(I_C_Invoice_Candidate.COLUMNNAME_M_InOut_ID, customerReturnId);
+
+		// Optional product filter — required when the return generates multiple ICs (packing-material lines alongside the product IC)
+		final String productIdentifierRaw = DataTableUtil.extractStringOrNullForColumnName(row, "OPT." + COLUMNNAME_M_Product_ID + "." + TABLECOLUMN_IDENTIFIER);
+		if (!EmptyUtil.isBlank(productIdentifierRaw))
+		{
+			final I_M_Product product = productTable.get(productIdentifierRaw);
+			queryBuilder.addEqualsFilter(COLUMNNAME_M_Product_ID, product.getM_Product_ID());
+		}
+
+		final Optional<I_C_Invoice_Candidate> invoiceCandidate = queryBuilder
 				.create()
 				.firstOnlyOptional(I_C_Invoice_Candidate.class);
 
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
@@ -268,6 +268,21 @@ public class C_Invoice_Candidate_StepDef
 		return invoiceCandidateHolder.getValueNotNull();
 	}
 
+	/**
+	 * Polls (up to {@code timeoutSec}) for the invoice candidate(s) of a vendor/customer return inout and
+	 * stores each under its identifier. Each DataTable row is delegated to {@link #loadCreditMemoCandidate(Map)}.
+	 *
+	 * <p>Required columns per row:
+	 * <ul>
+	 *   <li>{@code M_InOut_ID.Identifier} – the return inout</li>
+	 *   <li>{@code C_Invoice_Candidate_ID.Identifier} – identifier under which to store the found IC</li>
+	 * </ul>
+	 * <p>Optional columns per row:
+	 * <ul>
+	 *   <li>{@code OPT.M_Product_ID.Identifier} – filters the lookup to this product; required when the return
+	 *       generates several ICs (e.g. HU packing-material lines alongside the product line)</li>
+	 * </ul>
+	 */
 	@And("^after not more than (.*)s, credit memo candidates are found:$")
 	public void find_credit_memo_candidates(final int timeoutSec, @NonNull final DataTable dataTable) throws InterruptedException
 	{
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
@@ -771,6 +771,8 @@ public class C_Invoice_Candidate_StepDef
 				.map(productIdentifier -> productIdentifier.lookupNotNullIn(productTable))
 				.ifPresent(product -> queryBuilder.addEqualsFilter(COLUMNNAME_M_Product_ID, product.getM_Product_ID()));
 
+		// firstOnlyOptional throws when >1 IC matches: pass OPT.M_Product_ID.Identifier to narrow the lookup
+		// whenever the return generates several ICs (e.g. HU packing-material lines alongside the product line).
 		final Optional<I_C_Invoice_Candidate> invoiceCandidate = queryBuilder
 				.create()
 				.firstOnlyOptional(I_C_Invoice_Candidate.class);
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
@@ -270,7 +270,7 @@ public class C_Invoice_Candidate_StepDef
 
 	/**
 	 * Polls (up to {@code timeoutSec}) for the invoice candidate(s) of a vendor/customer return inout and
-	 * stores each under its identifier. Each DataTable row is delegated to {@link #loadCreditMemoCandidate(Map)}.
+	 * stores each under its identifier. Each DataTable row is delegated to {@link #loadCreditMemoCandidate(DataTableRow)}.
 	 *
 	 * <p>Required columns per row:
 	 * <ul>
@@ -286,7 +286,7 @@ public class C_Invoice_Candidate_StepDef
 	@And("^after not more than (.*)s, credit memo candidates are found:$")
 	public void find_credit_memo_candidates(final int timeoutSec, @NonNull final DataTable dataTable) throws InterruptedException
 	{
-		for (final Map<String, String> row : dataTable.asMaps())
+		for (final DataTableRow row : DataTableRows.of(dataTable).toList())
 		{
 			StepDefUtil.tryAndWait(timeoutSec, 500, () -> loadCreditMemoCandidate(row));
 		}
@@ -759,21 +759,17 @@ public class C_Invoice_Candidate_StepDef
 	 *
 	 * @return {@code true} if exactly one matching IC was found and stored; {@code false} if none found yet
 	 */
-	private boolean loadCreditMemoCandidate(@NonNull final Map<String, String> row)
+	private boolean loadCreditMemoCandidate(@NonNull final DataTableRow row)
 	{
-		final String customerReturnIdentifier = DataTableUtil.extractStringForColumnName(row, I_M_InOut.COLUMNNAME_M_InOut_ID + "." + TABLECOLUMN_IDENTIFIER);
-		final int customerReturnId = shipmentTable.get(customerReturnIdentifier).getM_InOut_ID();
+		final int returnInOutId = row.getAsIdentifier(I_M_InOut.COLUMNNAME_M_InOut_ID).lookupNotNullIn(shipmentTable).getM_InOut_ID();
 
 		final IQueryBuilder<I_C_Invoice_Candidate> queryBuilder = queryBL.createQueryBuilder(I_C_Invoice_Candidate.class)
-				.addEqualsFilter(I_C_Invoice_Candidate.COLUMNNAME_M_InOut_ID, customerReturnId);
+				.addEqualsFilter(I_C_Invoice_Candidate.COLUMNNAME_M_InOut_ID, returnInOutId);
 
 		// Optional product filter — required when the return generates multiple ICs (packing-material lines alongside the product IC)
-		final String productIdentifierRaw = DataTableUtil.extractStringOrNullForColumnName(row, "OPT." + COLUMNNAME_M_Product_ID + "." + TABLECOLUMN_IDENTIFIER);
-		if (!EmptyUtil.isBlank(productIdentifierRaw))
-		{
-			final I_M_Product product = productTable.get(productIdentifierRaw);
-			queryBuilder.addEqualsFilter(COLUMNNAME_M_Product_ID, product.getM_Product_ID());
-		}
+		row.getAsOptionalIdentifier(COLUMNNAME_M_Product_ID)
+				.map(productIdentifier -> productIdentifier.lookupNotNullIn(productTable))
+				.ifPresent(product -> queryBuilder.addEqualsFilter(COLUMNNAME_M_Product_ID, product.getM_Product_ID()));
 
 		final Optional<I_C_Invoice_Candidate> invoiceCandidate = queryBuilder
 				.create()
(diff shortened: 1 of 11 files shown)


backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
```diff
@@ -718,13 +742,40 @@ public void check_not_marked_as_to_recompute(final int timeoutSec, @NonNull fina
...
+	 * @return {@code true} if exactly one matching IC was found and stored; {@code false} if none found yet
+	 */
 	private boolean loadCreditMemoCandidate(@NonNull final Map<String, String> row)
 	{
 		final String customerReturnIdentifier = DataTableUtil.extractStringForColumnName(row, I_M_InOut.COLUMNNAME_M_InOut_ID + "." + TABLECOLUMN_IDENTIFIER);
 		final int customerReturnId = shipmentTable.get(customerReturnIdentifier).getM_InOut_ID();
 
-		final Optional<I_C_Invoice_Candidate> invoiceCandidate = queryBL.createQueryBuilder(I_C_Invoice_Candidate.class)
-				.addEqualsFilter(I_C_Invoice_Candidate.COLUMNNAME_M_InOut_ID, customerReturnId)
+		final IQueryBuilder<I_C_Invoice_Candidate> queryBuilder = queryBL.createQueryBuilder(I_C_Invoice_Candidate.class)
```
mrie04, line 766: pls modernise to DataTableRows

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/shipment/M_InOut_StepDef.java
```diff
@@ -1180,6 +1181,52 @@ public void generateVendorReturnFromReceipt(@NonNull final DataTable dataTable)
...
+			final I_M_InOut receipt = row.getAsIdentifier(I_M_InOut.COLUMNNAME_M_InOut_ID).lookupNotNullIn(inoutTable);
+			final List<I_M_HU> hus = huInOutBL.retrieveHandlingUnits(receipt);
+			assertThat(hus)
+					.as("Receipt %s must have assigned HUs before creating a vendor return via the HU path", receipt.getM_InOut_ID())
+					.isNotEmpty();
+			final List<Integer> receiptLineIds = inOutBL.getLines(receipt).stream()
+					.map(org.compiere.model.I_M_InOutLine::getM_InOutLine_ID)
+					.collect(Collectors.toList());
+			assertThat(receiptLineIds).as("Receipt %s must have lines", receipt.getM_InOut_ID()).isNotEmpty();
+			final ReturnsServiceFacade returnsServiceFacade = SpringContextHolder.instance.getBean(ReturnsServiceFacade.class);
```
mrie04: why not as field

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/shipment/M_InOut_StepDef.java
```diff
@@ -1180,6 +1181,52 @@ public void generateVendorReturnFromReceipt(@NonNull final DataTable dataTable)
...
+	 *   | receipt_VR_PI | vendorReturn_VR_PI |
+	 * </pre>
+	 */
+	@And("generate vendor return from receipt HUs")
+	public void generateVendorReturnFromReceiptHUs(@NonNull final DataTable dataTable)
+	{
+		DataTableRows.of(dataTable).forEach(row ->
+		{
+			final I_M_InOut receipt = row.getAsIdentifier(I_M_InOut.COLUMNNAME_M_InOut_ID).lookupNotNullIn(inoutTable);
+			final List<I_M_HU> hus = huInOutBL.retrieveHandlingUnits(receipt);
```
mrie04: why not use identifier of hu from When create material receipt

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
```diff
@@ -985,19 +998,54 @@ public static PriceAndTax calculatePriceAndTax(
...
+	 * identical to before (e.g. standalone returns / empties without an order line stay unchanged).
+	 */
+	@Nullable
+	private static HUPIItemProductId resolveReturnOriginPackingInstruction(@NonNull final org.compiere.model.I_M_InOutLine originLine)
+	{
+		if (originLine.getC_OrderLine_ID() <= 0)
+		{
+			return null;
+		}
+		final de.metas.interfaces.I_C_OrderLine orderLine = create(originLine.getC_OrderLine(), de.metas.interfaces.I_C_OrderLine.class);
```
mrie04: dao or bl should be used

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
```diff
@@ -353,8 +354,10 @@ private I_C_Invoice_Candidate createInvoiceCandidateForInOutLineOrNull(
 
 		//
 		// Pricing Informations
-		final org.compiere.model.I_M_InOutLine inOutLineRecordToUse = inOutLineRecord.getReturn_Origin_InOutLine_ID() > 0 ? inOutLineRecord.getReturn_Origin_InOutLine() : inOutLineRecord;
-		calculatePriceAndTaxAndUpdate(icRecord, inOutLineRecordToUse);
+		final boolean isReturnFromOrigin = inOutLineRecord.getReturn_Origin_InOutLine_ID() > 0;
+		final org.compiere.model.I_M_InOutLine inOutLineRecordToUse = isReturnFromOrigin ? inOutLineRecord.getReturn_Origin_InOutLine() : inOutLineRecord;
+		final HUPIItemProductId explicitPackingInstruction = isReturnFromOrigin ? resolveReturnOriginPackingInstruction(inOutLineRecordToUse) : null;
```
mrie04: i thought this would be scoped to de.metas.inout.impl.InOutBL#isVendorReturn (new version that accepts InoutId -> getById -> call existing)

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
```diff
@@ -353,8 +354,10 @@ private I_C_Invoice_Candidate createInvoiceCandidateForInOutLineOrNull(
 
 		//
 		// Pricing Informations
-		final org.compiere.model.I_M_InOutLine inOutLineRecordToUse = inOutLineRecord.getReturn_Origin_InOutLine_ID() > 0 ? inOutLineRecord.getReturn_Origin_InOutLine() : inOutLineRecord;
-		calculatePriceAndTaxAndUpdate(icRecord, inOutLineRecordToUse);
+		final boolean isReturnFromOrigin = inOutLineRecord.getReturn_Origin_InOutLine_ID() > 0;
```
mrie04: In general in this file use Repoidaware != null no int compare

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
```diff
@@ -919,16 +922,26 @@ private void setBPartnerData(final I_C_Invoice_Candidate ic, final I_M_InOutLine
 	public PriceAndTax calculatePriceAndTax(@NonNull final I_C_Invoice_Candidate icRecord)
 	{
 		final I_M_InOutLine inoutLine = getM_InOutLine(icRecord);
-		final org.compiere.model.I_M_InOutLine inOutLineRecordToUse = inoutLine.getReturn_Origin_InOutLine_ID() > 0 ? inoutLine.getReturn_Origin_InOutLine() : inoutLine;
+		final boolean isReturnFromOrigin = inoutLine.getReturn_Origin_InOutLine_ID() > 0;
+		final org.compiere.model.I_M_InOutLine inOutLineRecordToUse = isReturnFromOrigin ? inoutLine.getReturn_Origin_InOutLine() : inoutLine;
```
mrie04: dao or bl

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
```diff
@@ -985,19 +998,54 @@ public static PriceAndTax calculatePriceAndTax(
...
+	/**
+	 * Resolves the effective Packvorschrift (M_HU_PI_Item_Product) for a return's origin line via its linked
+	 * order line. This mirrors the order-line fallback of {@code InOutLineHUPackingAware} (which lives in
+	 * de.metas.handlingunits.base and is not visible from this module): a vendor-receipt origin line carries
+	 * no PI in its own columns — it lives on the purchase order line — so pricing must recover it there.
+	 * Returns {@code null} when there is no order line or it carries no PI, leaving pricing behaviour
+	 * identical to before (e.g. standalone returns / empties without an order line stay unchanged).
+	 */
+	@Nullable
+	private static HUPIItemProductId resolveReturnOriginPackingInstruction(@NonNull final org.compiere.model.I_M_InOutLine originLine)
```
mrie04: still should check inoutline first, for case set in the future, or explain why always orderline one. Purchase Order IC is initial price

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/invoicecandidate/C_Invoice_Candidate_StepDef.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/becb1e45658 — `loadCreditMemoCandidate` now takes a `DataTableRow`; the caller iterates `DataTableRows.of(dataTable).toList()`, the return inout is resolved via `row.getAsIdentifier(M_InOut_ID).lookupNotNullIn(shipmentTable)` and the optional product via `row.getAsOptionalIdentifier(M_Product_ID)` — no more `Map<String,String>` / `DataTableUtil.extract…`.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/shipment/M_InOut_StepDef.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/becb1e45658 — `ReturnsServiceFacade` is now a class field (`private final … = SpringContextHolder.instance.getBean(ReturnsServiceFacade.class)`), matching the other service fields in this step def; the per-call `getBean` is gone.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/shipment/M_InOut_StepDef.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/becb1e45658 — the step now takes an `M_HU_ID` identifier column and returns the HU the `When create material receipt` step already created (`hu_VR_PI`), resolved from `huTable` (comma-separated list supported). This drops the `huInOutBL.retrieveHandlingUnits(receipt)` re-retrieval and mirrors the real flow where the user selects specific HUs to return. The receipt (`M_InOut_ID`) is kept only to locate the created return via its line ids.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/becb1e45658 — the order line is now loaded through `IOrderDAO.getOrderLineById(OrderLineId)` instead of `create(originLine.getC_OrderLine(), …)`.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/becb1e45658 — added `IInOutBL.isVendorReturn(InOutId)` (getById → the existing `isVendorReturn(I_M_InOut)`), and both handler sites now gate the explicit-Packvorschrift injection on `inOutBL.isVendorReturn(InOutId.ofRepoId(inOut.getM_InOut_ID()))`.

On the design question you raised (resolve at `createPricingCtx` build vs. pass explicit down): we kept the explicit route deliberately. `createPricingCtx` is handed the origin **receipt** line (the handler already swapped it in for the return), so it cannot tell "this is for a vendor return"; resolving own-PI-else-orderline unconditionally there would also change pricing for a non-sales-order shipment that carries a PI and is invoiced via shipment config. The `isVendorReturn` gate confines the change to vendor returns.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/becb1e45658 — replaced the int comparison with RepoId-aware ids: `InOutLineId.ofRepoIdOrNull(getReturn_Origin_InOutLine_ID()) != null` (return-origin selection) and `OrderLineId.ofRepoIdOrNull(getC_OrderLine_ID())` in the resolver.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/becb1e45658 — the origin line is now loaded via `inOutBL.getLineByIdInTrx(returnOriginLineId)` instead of the `getReturn_Origin_InOutLine()` model-nav getter.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/becb1e45658 — `resolvePackingInstruction` now checks the inout line's **own** Packvorschrift first (`getValueOverrideOrValue(line, "M_HU_PI_Item_Product_ID")`, override-aware) and only falls back to the order-line PI when the line carries none — future-proof if receipt lines ever get stamped. Today the vendor-receipt origin line carries no PI in any of its own columns, so it resolves the PO-line PI, which carries the initial PO price (as you note, the Purchase-Order IC is the initial price). The own-PI is read generically because its typed getter lives on the handlingunits `I_M_InOutLine`, not importable from this module.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
```diff
@@ -919,16 +933,30 @@ private void setBPartnerData(final I_C_Invoice_Candidate ic, final I_M_InOutLine
...
 	{
 		final I_M_InOutLine inoutLine = getM_InOutLine(icRecord);
-		final org.compiere.model.I_M_InOutLine inOutLineRecordToUse = inoutLine.getReturn_Origin_InOutLine_ID() > 0 ? inoutLine.getReturn_Origin_InOutLine() : inoutLine;
-
-		return calculatePriceAndTax(icRecord, inOutLineRecordToUse);
+		final InOutLineId returnOriginLineId = InOutLineId.ofRepoIdOrNull(inoutLine.getReturn_Origin_InOutLine_ID());
+		final org.compiere.model.I_M_InOutLine inOutLineRecordToUse = returnOriginLineId != null
+				? inOutBL.getLineByIdInTrx(returnOriginLineId)
+				: inoutLine;
+		final HUPIItemProductId explicitPackingInstruction = inOutBL.isVendorReturn(InOutId.ofRepoId(inoutLine.getM_InOut_ID()))
```
mrie04: why use id version here if inout is already loaded. replace load above with proper getbyid

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
```diff
@@ -985,19 +1013,75 @@ public static PriceAndTax calculatePriceAndTax(
...
+			return ownPackingInstruction;
+		}
+
+		final OrderLineId orderLineId = OrderLineId.ofRepoIdOrNull(inOutLine.getC_OrderLine_ID());
+		if (orderLineId == null)
+		{
+			return null;
+		}
+
+		final de.metas.interfaces.I_C_OrderLine orderLine = Services.get(IOrderDAO.class).getOrderLineById(orderLineId);
```
mrie04: why inline instead of field?

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
```diff
@@ -985,19 +1013,75 @@ public static PriceAndTax calculatePriceAndTax(
...
+	 * where a vendor-receipt origin line keeps the PI (the receipt line itself carries none). Returns {@code null}
+	 * when neither carries a PI (e.g. empties / returns without an order line), leaving pricing unchanged.
+	 * <p>
+	 * The line's own PI is read generically because its typed getter lives on the handlingunits {@code I_M_InOutLine},
+	 * which this module cannot import; the order-line PI is read through {@link IOrderDAO} rather than a model getter.
+	 */
+	@Nullable
+	private static HUPIItemProductId resolvePackingInstruction(@NonNull final org.compiere.model.I_M_InOutLine inOutLine)
+	{
+		final Integer ownPackingInstructionRepoId = getValueOverrideOrValue(inOutLine, COLUMNNAME_M_HU_PI_Item_Product_ID);
```
mrie04: this should bl or dao method, isn't there an existing

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/5e7876bdbfd — both sites now use the loaded `I_M_InOut` record for `isVendorReturn`: site 1 uses the already-loaded `inOut`; the recompute site loads it once via `inOutBL.getById(...)` and passes the record. The `isVendorReturn(InOutId)` overload is now unused and has been removed.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/5e7876bdbfd — the resolution moved into `InOutBL`, which uses its existing `orderDAO` **field**; there is no more inline `Services.get(IOrderDAO.class)` (that whole resolver no longer lives in the handler).

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inout/invoicecandidate/M_InOutLine_Handler.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/5e7876bdbfd — moved into a BL method `InOutBL.resolvePIForVendorReturnPricingCtx(returnOriginLine)` (`CoalesceUtil.coalesce(ownPI, orderLinePI)`), so the generic `InterfaceWrapperHelper` read + order-line lookup live in the BL, not the handler. On "isn't there an existing": the canonical resolver `InOutLineHUPackingAware` lives in `de.metas.handlingunits.base`, which sits **above** both `de.metas.swat.base` and `de.metas.business`, so it can't be imported here — hence a new BL method. It is deliberately named/scoped for vendor-return pricing (not a general effective-PI resolver), and the own-PI is read generically only because that column's typed getter also lives in the non-importable handlingunits `I_M_InOutLine`.

PR #25359, opened 2026-07-30
Workstation MobileUI — re-read operator workstation and workplace on screen re-entry

teosarca:
## Problem

On the Mobile-UI, a screen that displays the operator's current workstation / active workplace read
those values **once per mount**. `useCurrentWorkstation` (`src/api/workstation.js`) and
`useCurrentWorkplace` (`src/api/workplace.js`) both fetched inside `useEffect(…, [])`, with nothing
listening for focus or visibility changes.

Navigating in and out happened to work, because the launchers screen unmounts. A screen that **stays
open** — a second app instance, or an OS-level app switch — showed the value it had at mount, forever.
Operators scanned a new workstation, switched back to the Produktion screen, and saw the previous
station indefinitely, while the server and the server-filtered job list were already on the new one.
Every re-scan therefore looked like it had failed.

Reproduced live on the target instance across a 21-variant sweep: the navigation route was uniformly
correct; only the screen-stays-mounted scenarios were stale.

## Fix

A small `useReentryKey()` hook returns a counter that increments whenever the app is re-entered
(window `focus`, or `visibilitychange` back to visible). Both effects take that key as a dependency.

Deliberately minimal and behaviour-preserving: no change to loading/error/spinner semantics, no state
moved into redux, no polling. `useCurrentWorkstation` additionally gains
`requiresWorkstationIfAvailable` as a dependency — it arrives from the applications store, so an effect
that ran once while it was still falsy would never fetch at all.

`useCurrentTrolley` has the same one-shot shape but is not named by any acceptance criterion; it is
left alone deliberately.

Migrating the two hooks to `useQuery` (the newer pattern the trolley hook uses) was considered and
rejected for this bugfix: `useQuery` has no error branch and starts `isPending: false`, so it would
change both hooks' spinner/toast behaviour for no gain on this defect.

## Verification

Local mobile-webui Playwright, driven against a dev server serving this branch's frontend:

- `workstation_switches_workplace.spec.js` — new test **RED before** the hook change (header stayed on
  `WS1` / `wpA` after the switch to `WS2` / `wpB`), **GREEN after**; the pre-existing test in the file
  passes throughout. The same RED was reproduced against the prebuilt base-branch frontend first, to
  confirm it is the production defect and not a harness artifact.
- Regression: `distribution/filter_by_workplace.spec.js` (3) and `manufacturing/manufacturing.spec.js`
  (4) — all pass.
- Jest: 4 test methods on `useReentryKey` (starts at 0; increments on `focus`; increments on
  `visibilitychange` when visible; does **not** increment while hidden). Observed failing before the
  hook existed.
- `yarn lint` clean.

## Note for the reviewer

The change also makes a **failed** read retry on the next re-entry, instead of leaving the header blank
until the screen is remounted. That is a second defect found during the reproduction, and it is not
named by any acceptance criterion — it comes with the same one dependency-array entry rather than being
separable. Say the word if it should be excluded.

### DIFF
--- a/e2e/mobile-webui/tests/spec/manufacturing/workstation_switches_workplace.spec.js
+++ b/e2e/mobile-webui/tests/spec/manufacturing/workstation_switches_workplace.spec.js
@@ -7,6 +7,7 @@ import { ApplicationsListScreen } from "../../utils/screens/ApplicationsListScre
 import { WorkstationManagerScreen } from "../../utils/screens/workstationManager/WorkstationManagerScreen";
 import { WorkplaceManagerScreen } from "../../utils/screens/workplaceManager/WorkplaceManagerScreen";
 import { ManufacturingJobsListScreen } from "../../utils/screens/manufacturing/ManufacturingJobsListScreen";
+import { AppLifecycleComponent } from "../../utils/components/AppLifecycleComponent";
 
 const createMasterdata = async () => {
     return await Backend.createMasterdata({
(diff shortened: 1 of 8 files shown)


teosarca:
Closing: this does not fix the defect it was opened for.

The investigation established that the stale workstation/workplace is **not** a refresh problem in the
launchers screen. The screen was freshly mounted each time and did re-issue both reads; the reads never
reached the backend. A re-entry refresh therefore changes nothing for that path — it re-issues the same
request, which is short-circuited on the client side before it leaves the device.

What is in here is still real and still green: `useCurrentWorkstation` and `useCurrentWorkplace` read once
per mount with no refresh path, so session-global operator state can go stale on a screen that stays
mounted (e.g. the assignment being changed by someone else while the screen sits open). That is a genuine
gap, just a different one — it is not what the reported symptom was.

Not merging a change whose premise turned out to be wrong, even though CI is green. The branch
`intensive_care_hotfix_WorkstationContextRefresh` is left in place, so the hook (`useReentryKey`, with its
Jest coverage) and the e2e scaffolding can be picked up if the remaining staleness gap is scheduled on its
own.

PR #25361, opened 2026-07-30
`Manufactuirng Order`: default `Plant / Work Station` from `Product Category`

riadmen:
Add S_Resource_ID (Plant) and WorkStation_ID fields to M_Product_Category and derive PP_Order.S_Resource_ID / PP_Order.WorkStation_ID from the selected product's category when not already set.                                                                                                  
       
  - migration 5817090: add M_Product_Category.S_Resource_ID + WorkStation_ID                                                                         
    (FK to S_Resource)                                                                                                                               
  - regenerate I_/X_M_Product_Category model                                                                                                         
  - PP_Order interceptor (beforeSave): fill plant/workstation from the  product's category as a fallback, without overriding values already set (e.g. from Produktionsdispo/PP_Product_Planning)

### DIFF
--- a/backend/de.metas.manufacturing/src/main/java/org/eevolution/callout/PP_Order.java
+++ b/backend/de.metas.manufacturing/src/main/java/org/eevolution/callout/PP_Order.java
@@ -13,6 +13,8 @@ import de.metas.material.planning.pporder.IPPRoutingRepository;
 import de.metas.material.planning.pporder.PPRoutingId;
 import de.metas.organization.OrgId;
 import de.metas.product.IProductBL;
+import de.metas.product.IProductDAO;
+import de.metas.product.ProductCategoryId;
 import de.metas.product.ProductId;
 import de.metas.product.ResourceId;
 import de.metas.quantity.Quantity;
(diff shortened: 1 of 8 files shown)


backend/de.metas.manufacturing/src/main/java/org/eevolution/callout/PP_Order.java
```diff
@@ -53,6 +56,7 @@ public class PP_Order extends CalloutEngine
 	private final IWarehouseBL warehouseBL = Services.get(IWarehouseBL.class);
 	private final IDocTypeBL docTypeBL = Services.get(IDocTypeBL.class);
 	private final IProductBL productBL = Services.get(IProductBL.class);
+	private final IProductDAO productsRepo = Services.get(IProductDAO.class);
```
mrie04: don't add dao in addition to bl, instead delegate via bl (create function in bl, that calls the dao one

mrie04 (changes requested):
pls check my comment

backend/de.metas.manufacturing/src/main/java/org/eevolution/callout/PP_Order.java
riadmen: done

mrie04 (approved):
lgtm

PR #25365, opened 2026-07-30
Transport order: make B/L Date and ETA editable while completed

teosarca:
## Problem

A completed `M_ShipperTransportation` is `Processed='Y'`, and the WebUI renders every field of a processed document read-only unless the column carries `AD_Column.IsAlwaysUpdateable='Y'` (`DocumentReadonly.computeFieldReadonly`). `M_ShipperTransportation.BLDate` and `M_ShipperTransportation.ETA` were both left at `'N'`, so neither date could be entered or corrected once the transport order was completed — although the bill of lading is issued by the carrier only **after** loading, i.e. normally after the transport order is done.

This also left the completed/closed branch of `M_ShipperTransportation.syncOrderDatesOnEdit` unreachable from the WebUI. That handler recomputes the linked purchase orders' pay-schedule due dates when `BLDate` or `ETA` is edited on an already-completed transport order — an edit the UI never permitted. The only workaround was to reactivate and re-complete the document, which resets `M_ShippingPackage.Processed` and nulls/re-derives `M_Package.ShipDate` on every linked package.

## Fix

Set `AD_Column.IsAlwaysUpdateable='Y'` on the two columns.

This is an alignment, not a new behaviour: every other logistics field on the same tab is already always-updateable — among them `M_ShipperTransportation.ATA`, `M_ShipperTransportation.ETD` and `M_ShipperTransportation.IsBLReceived` ("B/L erhalten"), the companion flag of the B/L date. `BLDate` and `ETA` were the only two outliers.

## Scope / blast radius

- `M_ShipperTransportation.BLDate` and `.ETA` are read in exactly one place — the transport-order validator → `OrderBL.syncDatesFromTransportOrder` → `C_Order.BLDate`/`C_Order.ETA` → the order's pay schedule. No posting, accounting or material-planning consumer.
- `M_ReceiptSchedule.BLDate` is a virtual `ColumnSQL` reading `max(st.BLDate)` live, so it follows automatically.
- `AD_Column.IsAlwaysUpdateable` makes the fields editable in every document state, not only Completed. That is harmless here: `syncOrderDatesOnEdit` is guarded on `docStatus.isCompletedOrClosed()`, so an edit in any other state does not propagate. A draft transport order's tentative dates still do **not** reach the pay schedule.

## Verification

Application dictionary probed on a customer-faithful DB (all core + customer migrations applied), window 540020 "Transport Auftrag", tab 540096 "Speditionslieferung":

| Column | `IsUpdateable` | `IsAlwaysUpdateable` before |
|---|---|---|
| `M_ShipperTransportation.ATA` | Y | Y |
| `M_ShipperTransportation.ETD` | Y | Y |
| `M_ShipperTransportation.IsBLReceived` | Y | Y |
| `M_ShipperTransportation.BLDate` | Y | **N** |
| `M_ShipperTransportation.ETA` | Y | **N** |

Both fields already carry `AD_Field.IsReadOnly='N'` with no read-only logic, so the column flag was the only thing gating them. `AD_Window.Overrides_Window_ID = 540020` returns no row — there is no override window, so the column-level change reaches the window that is actually opened.

### DIFF
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/purchaseOrder/purchaseOrderComplexPaymentTerm.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/purchaseOrder/purchaseOrderComplexPaymentTerm.feature
@@ -247,6 +247,26 @@ Feature: Purchase order with complex payment term
       | PTB32                  | 2025-04-01 | 76.72  | WP     |
 
 
+  # ---------------------------------------------------------------------------------------------
+  # S30954_1..S30954_3 own the BL-date -> pay-schedule propagation. They are the ONLY home for it.
+  #
+  # Why they are meaningful: the "update transport order" step writes through the model layer
+  # (M_ShipperTransportation_StepDef#updateTransportOrder -> record.setBLDate(...) -> saveRecord),
+  # so saving the record fires the real @ModelChange interceptor
+  # de.metas.shipping.model.validator.M_ShipperTransportation#syncOrderDatesOnEdit, which propagates
+  # onto the C_Order and in turn fires C_Order#updateOrderPaySchedules. That is exactly the
+  # production chain — nothing here is stubbed, and no scenario may be rewritten to poke
+  # C_OrderPaySchedule (or any table on the chain) directly: doing so would bypass the very
+  # interceptors these scenarios exist to prove, and they would keep passing after the chain broke.
+  #
+  # What these scenarios deliberately do NOT cover: whether the WebUI even lets a user type the B/L
+  # date once the transport order is completed. The step above goes straight to saveRecord and never
+  # touches the WebUI Document layer, where DocumentReadonly#computeFieldReadonly blanks every field
+  # of a Processed='Y' document unless its AD_Column.IsAlwaysUpdateable='Y'. That read-only gate is
+  # covered — once, in a real browser — by the Playwright spec
+  # e2e/frontend-webui/tests/spec/transport-order-dates-editable-when-completed.spec.js.
+  # Keep the split: model/business chain here, WebUI editability there, no overlap.
+  # ---------------------------------------------------------------------------------------------
   @from:cucumber
 @allure.label.epic:E0140_Purchasing
 @allure.label.feature:F00600_Purchase_Order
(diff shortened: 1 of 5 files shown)


backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/purchaseOrder/purchaseOrderComplexPaymentTerm.feature
```diff
@@ -412,6 +433,63 @@ Feature: Purchase order with complex payment term
       | PTB62                  | 2025-10-23 | 92.07  | WP     |
 
 
+  @from:cucumber
+@allure.label.epic:E0140_Purchasing
+@allure.label.feature:F00600_Purchase_Order
+@allure.label.feature:F00994_Multiple_Levels_of_Payment
+@F00600
+@Id:S30954_5
+  Scenario: ETA corrected after transport order completion recomputes the shipping line
```
teosarca, line 442: we need the same test with B/L date too

teosarca: **Resolution (dev):** Already covered — `@Id:S30954_2` at `purchaseOrderComplexPaymentTerm.feature:330-331` ("BL date corrected after transport order completion recomputes the shipping line") is the exact BL-date twin of this ETA scenario. `S30954_5` was deliberately written as its single-axis mirror: same payment-term shape (one `OD` break + one date-driven break), same order/transport-order fixture, same three-step assertion — set the date, complete the transport order, then **change the date again on the already-completed transport order** and re-assert the recomputed due date. `S30954_2` varies only `BLDate` (2025-10-15 -> 2025-10-22, due date 2025-10-20 -> 2025-10-27); `S30954_5` varies only `ETA` on an `ET`-referenced break (2025-10-15 -> 2025-10-22, due date 2025-10-20 -> 2025-10-27) and never touches `BLDate`.

BL date is in fact the better-covered of the two axes — three scenarios to ETA's one:
- `@Id:S30954_1` (`:276-277`) — BL date entered *after* completion
- `@Id:S30954_2` (`:330-331`) — BL date *corrected* after completion (the twin of this one)
- `@Id:S30954_3` (`:387-388`) — draft transport order does *not* propagate its BL date

ETA had zero propagation coverage before `S30954_5`, which is why this iteration added the ETA side only — `M_ShipperTransportation.syncOrderDatesOnEdit` fires on both columns, so the ETA half was the untested one.

Editability (the actual subject of this PR — `AD_Column.IsAlwaysUpdateable` on `M_ShipperTransportation.BLDate` and `.ETA`) is covered for **both** columns in `e2e/frontend-webui/tests/spec/transport-order-dates-editable-when-completed.spec.js`, with a `DateDoc` control that stays disabled in both states so the test cannot pass by simply finding "some field enabled".

If you meant something the above does not cover — e.g. a scenario that varies `BLDate` and `ETA` together on one transport order, or a BL-date case on the mobile/WebUI path — say the word and I will add it.

PR #25372, opened 2026-07-31
Purchase order reactivation: a proforma allocation alone must not block it

teosarca:
## What changed

A proforma invoice allocated to a purchase order no longer prevents that order from being reactivated.

Reactivation is now blocked **only** by committed downstream activity — a goods receipt (`M_InOut_ID`) or a matched invoice (`C_Invoice_ID`) linked to a pay-schedule line. `OrderPayScheduleProformaService` is no longer consulted by the reactivate guard at all, so the dependency was removed rather than left injected-but-unused.

## Why the guard removal alone was not enough

Reactivate deletes every `C_OrderPaySchedule` row. Before this change a proforma allocation always blocked reactivation, so the reactivate → re-complete round trip on a proforma-linked order was unreachable. Relaxing the guard makes it reachable — and re-complete rebuilt the schedule from the payment-term breaks alone, silently dropping the proforma-derived state (Paid status, `DueAmt_Actual`, `ReferenceDate`, `DueDate`, and the order's `LC_Date`).

So the second half of this PR re-derives that state: `de.metas.order.model.interceptor.C_Order#createOrderPaySchedules` now calls `OrderPayScheduleLCStepService#recomputeLCStepAfterOrderCompleted` immediately after `orderPayScheduleService.createOrderPaySchedules(order)`, where it observes the freshly rebuilt schedule. The allocation and its prepayment survive reactivation untouched, so the round trip is lossless.

The call could not go inside `OrderPayScheduleService.createOrderPaySchedules` — `OrderPayScheduleLCStepService` already depends on `OrderPayScheduleService`, which would close a dependency cycle. The interceptor is a leaf, so injecting there is safe.

## Regression introduced by this PR, and fixed inside it

The recompute call started out **unconditional**, and that reset a just-built **order-date** pay-schedule line on completion.

The recompute resolves its target as *the single LC line, else the single prepaid line* — and `isPrepaidLine()` admits order-date lines. So on a payment term with an order-date break and **no** LC break, and an order with no proforma allocation at all, `getSingleLCLine()` found nothing, the prepaid fallback picked the order-date line that `createOrderPaySchedules` had just built correctly, and the no-proforma branch reset it to `Pending` / `9999-12-01` with a null `ReferenceDate`.

That reset is load-bearing for de-allocation, so it stayed as it is. Instead the completion call site was routed through the guarded `OrderPayScheduleLCStepService#recomputeLCStepAfterOrderCompleted`, which early-returns when the order has no proforma allocation — i.e. it re-derives only when there is something to restore. It reuses `OrderPayScheduleProformaService#getIdByOrderId`, the same query the branch it guards uses, so the two cannot diverge.

Symptom and fix: https://github.com/metasfresh/metasfresh/commit/1fbb6157f3dbe47b296afd9c5ade82344148decd. The unconditional call broke 6 of the 9 scenarios in `purchaseOrder/purchaseOrderComplexPaymentTerm.feature` — precisely the 6 whose payment term has an order-date break and no LC break (`[DueDate] expected 2025-10-10 but was 9999-12-01`, `[Status] expected Awaiting_Pay but was Pending`); all 9 are green with the guard in place. `S30621_TC8` is the covering scenario in this PR's own feature file, and it is the first one there in which the order-date line is actually the recompute target — every earlier term carries an LC break, so `getSingleLCLine()` always wins and a wrongly reset order-date row stays invisible.

## Message text

`Order_Reactivate_Blocked_By_PaySchedule_Activity` no longer tells the user to de-allocate the proforma invoice, since that is no longer a remedy. Reworded in DE and EN (and `de_CH`, kept in sync with the German base as the two prior migrations for this message did).

## Testing

**Unit** — `C_Order_Test`, 6 tests covering the guard predicate: `Tests run: 6, Failures: 0`. `OrderPayScheduleLCServiceTest`: `Tests run: 13, Failures: 0, Errors: 0, Skipped: 0` (includes the two cases pinning both sides of the new completion guard: no allocation leaves the advance line untouched, allocation present still re-derives it). `OrderTest`: `Tests run: 10, Failures: 0, Errors: 0, Skipped: 0`.

**Cucumber** — `orderReactivatePayScheduleGuard.feature`, all **8** scenarios green (`@Id:S30621_TC1` … `TC8`):

- `TC1` Reactivate allowed — Awaiting_Pay OD row with no downstream (regression guard)
- `TC2` Reactivate allowed — proforma allocation exists, nothing committed downstream *(this scenario previously asserted the opposite; inverting it is the behaviour change under test)*
- `TC3` Reactivate blocked — goods receipt links a pay-schedule line
- `TC4` Reactivate then re-complete preserves the paid-proforma pay-schedule state
- `TC5` Reactivate then re-complete preserves the paid-proforma pay-schedule state on a B/L term
- `TC6` Reactivate then re-complete preserves the awaiting-pay proforma pay-schedule state
- `TC7` Reactivate then re-complete leaves an ordinary order's pay schedule untouched
- `TC8` Reactivate then re-complete preserves the paid-proforma state on the order-date step of a no-LC term

`purchaseOrder/purchaseOrderComplexPaymentTerm.feature`, all 9 scenarios green — the feature the regression above broke, and the reason the completion call is guarded.

`S30621_TC4` was verified to actually be diagnostic rather than incidentally green: with the recompute call temporarily commented out, it fails on `Status` (`Paid` → `Awaiting_Pay`), `DueAmt_Actual` (null) **and** `DueDate` (`2026-04-29` → `2026-04-24`). The proforma deliberately uses a payment term with a 5-day offset so its due date diverges from what the naive rebuild path would produce; without that, three of the five asserted columns would have passed with the fix absent.

**CI** — https://github.com/metasfresh/metasfresh/actions/runs/30618383055: every job green, Merge Gate pass.

## Accepted risk — retired, no longer applies

This section previously disclosed an accepted risk: because the completion interceptor called `recomputeLCStep` on **every** order completion, a payment term carrying more than one letter-of-credit break, or more than one advance/order-date break, would make `OrderPaySchedule#getSingleLCLine` / `#getSinglePrepaidLine` throw and abort **every** completion of every order on that term (`PaymentTerm.assertValid()` only checks that break percentages sum to 100%, so it does not reject such a term).

That is no longer the case. https://github.com/metasfresh/metasfresh/commit/1fbb6157f3dbe47b296afd9c5ade82344148decd routes the completion call through the guarded `recomputeLCStepAfterOrderCompleted`, which early-returns when the order has no proforma allocation. The fail-fast is therefore reachable again only for orders that carry a proforma allocation — exactly the blast radius it had before this PR. Nothing about the risk was left open; it was closed by the guard, not waived.

For orders that never had a proforma the call is inert by construction: it returns after the single allocation lookup, without reading or writing the pay schedule at all.

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
@@ -796,7 +796,8 @@ public class C_Order
 		// losing any proforma/prepayment-derived state (Paid/Awaiting_Pay, DueAmt_Actual, ReferenceDate,
 		// DueDate, LC_Date). Re-derive it here from the surviving C_Proforma_Order_Alloc / completed
 		// prepayment so a paid proforma allocation is restored rather than silently dropped.
-		orderPayScheduleLCStepService.recomputeLCStep(OrderId.ofRepoId(order.getC_Order_ID()));
+		// Only orders that carry a proforma allocation are touched — see the method's javadoc.
+		orderPayScheduleLCStepService.recomputeLCStepAfterOrderCompleted(OrderId.ofRepoId(order.getC_Order_ID()));
 	}
 
 	@DocValidate(timings = ModelValidator.TIMING_AFTER_REACTIVATE)
--- a/backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
+++ b/backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
@@ -60,6 +60,7 @@ import de.metas.user.UserGroupRepository;
 import de.metas.user.UserRepository;
 import de.metas.order.location.OrderLocationsUpdater;
 import de.metas.order.paymentschedule.core.service.OrderPayScheduleService;
+import de.metas.order.paymentschedule.steps.letter_of_credit.OrderPayScheduleLCStepService;
 import de.metas.promotioncode.PromotionCodeId;
 import de.metas.organization.IOrgDAO;
 import de.metas.organization.OrgId;
@@ -128,6 +129,7 @@ public class C_Order
 	@NonNull private final IDocumentLocationBL documentLocationBL;
 	@NonNull private final PurchaseOrderToShipperTransportationService purchaseOrderToShipperTransportationService;
 	@NonNull private final OrderPayScheduleService orderPayScheduleService;
+	@NonNull private final OrderPayScheduleLCStepService orderPayScheduleLCStepService;
 
 	@VisibleForTesting
 	public static final String AUTO_ASSIGN_TO_SALES_ORDER_BY_EXTERNAL_ORDER_ID_SYSCONFIG = "de.metas.payment.autoAssignToSalesOrderByExternalOrderId.enabled";
@@ -142,7 +144,8 @@ public class C_Order
 			@NonNull final IDocumentLocationBL documentLocationBL,
 			@NonNull final BPartnerSupplierApprovalService partnerSupplierApprovalService,
 			@NonNull final PurchaseOrderToShipperTransportationService purchaseOrderToShipperTransportationService,
-			@NonNull final OrderPayScheduleService orderPayScheduleService)
+			@NonNull final OrderPayScheduleService orderPayScheduleService,
+			@NonNull final OrderPayScheduleLCStepService orderPayScheduleLCStepService)
 	{
 		this.bpartnerBL = bpartnerBL;
 		this.bpartnerEffectiveBL = bpartnerEffectiveBL;
@@ -152,6 +155,7 @@ public class C_Order
 		this.documentLocationBL = documentLocationBL;
 		this.purchaseOrderToShipperTransportationService = purchaseOrderToShipperTransportationService;
 		this.orderPayScheduleService = orderPayScheduleService;
+		this.orderPayScheduleLCStepService = orderPayScheduleLCStepService;
 
 		final IProgramaticCalloutProvider programmaticCalloutProvider = Services.get(IProgramaticCalloutProvider.class);
 		programmaticCalloutProvider.registerAnnotatedCallout(this);
@@ -170,7 +174,8 @@ public class C_Order
 				DocumentLocationBL.newInstanceForUnitTesting(),
 				new BPartnerSupplierApprovalService(new BPartnerSupplierApprovalRepository(), new UserGroupRepository()),
 				PurchaseOrderToShipperTransportationService.newInstanceForUnitTesting(),
-				OrderPayScheduleService.newInstanceForUnitTesting()));
+				OrderPayScheduleService.newInstanceForUnitTesting(),
+				OrderPayScheduleLCStepService.newInstanceForUnitTesting()));
 	}
 
 	@ModelChange(timings = { ModelValidator.TYPE_BEFORE_NEW, ModelValidator.TYPE_BEFORE_CHANGE }, ifColumnsChanged = { I_C_Order.COLUMNNAME_M_PriceList_ID })
@@ -785,6 +790,13 @@ public class C_Order
 	public void createOrderPaySchedules(final I_C_Order order)
 	{
 		orderPayScheduleService.createOrderPaySchedules(order);
+
+		// Reactivate deletes every C_OrderPaySchedule row (see deleteOrderPaySchedules below), so a
+		// reactivate -> re-complete round trip rebuilds the schedule from the payment-term breaks alone,
+		// losing any proforma/prepayment-derived state (Paid/Awaiting_Pay, DueAmt_Actual, ReferenceDate,
+		// DueDate, LC_Date). Re-derive it here from the surviving C_Proforma_Order_Alloc / completed
+		// prepayment so a paid proforma allocation is restored rather than silently dropped.
+		orderPayScheduleLCStepService.recomputeLCStep(OrderId.ofRepoId(order.getC_Order_ID()));
 	}
 
 	@DocValidate(timings = ModelValidator.TIMING_AFTER_REACTIVATE)
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/order/orderReactivatePayScheduleGuard.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/order/orderReactivatePayScheduleGuard.feature
@@ -88,10 +88,10 @@ Feature: PO reactivation guard — downstream-activity check
 
   @from:cucumber
   @Id:S30621_TC2
-  Scenario: Reactivate blocked — proforma allocation exists
+  Scenario: Reactivate allowed — proforma allocation exists, nothing committed downstream
     # A buyer allocates a purchase proforma invoice to the order (LC step → Awaiting_Pay).
-    # The guard detects the proforma allocation and blocks reactivation. The buyer must
-    # de-allocate the proforma before they can reactivate.
+    # The allocation link and its prepayment both survive reactivation, so nothing downstream
+    # would be lost — the guard must allow reactivation without requiring a de-allocation first.
 
     And metasfresh contains C_Orders:
       | Identifier | IsSOTrx | C_BPartner_ID | DateOrdered | DocBaseType | M_Warehouse_ID | C_PaymentTerm_ID |
@@ -115,8 +115,8 @@ Feature: PO reactivation guard — downstream-activity check
       | ReferenceDateType | DueAmt   | DueAmt_Actual | Status |
       | LC                | 21000.00 | 21000.00      | WP     |
 
-    # Guard blocks — proforma allocation exists; user must de-allocate first
-    And the order identified by po cannot be reactivated
+    # Guard allows reactivation — proforma allocation with no goods-receipt/invoice link does not block
+    And the order identified by po is reactivated
 
 
   @from:cucumber
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/order/orderReactivatePayScheduleGuard.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/order/orderReactivatePayScheduleGuard.feature
@@ -40,11 +40,14 @@ Feature: PO reactivation guard — downstream-activity check
 
     # pt_od: LC 30% + OD 70% — OrderDate break resolves at completion, OD row immediately WP
     # pt_bl: LC 30% + BL 70% — BillOfLadingDate break resolves only on goods receipt
+    # pt_net5: TC4's proforma term — a real 5-day offset so the proforma's DueDate diverges
+    # from its DateInvoiced, making DueDate diagnostic of whether the LC recompute fired.
     And metasfresh contains C_PaymentTerm
-      | Identifier   |
-      | pt_od        |
-      | pt_bl        |
-      | pt_immediate |
+      | Identifier   | OPT.NetDays |
+      | pt_od        |             |
+      | pt_bl        |             |
+      | pt_immediate |             |
+      | pt_net5      | 5           |
     And metasfresh contains C_PaymentTerm_Break
       | Identifier  | C_PaymentTerm_ID | Percent | OffsetDays | ReferenceDateType | SeqNo |
       | ptb_od_lc   | pt_od            | 30      | 0          | LC                | 10    |
@@ -194,10 +197,11 @@ Feature: PO reactivation guard — downstream-activity check
       | poL1       | po         | product      | 700        |
     And the order identified by po is completed
 
-    # Allocate a proforma invoice to the order (LC row -> Awaiting_Pay)
+    # Allocate a proforma invoice to the order (LC row -> Awaiting_Pay). pt_net5 gives the
+    # proforma a real 5-day offset (DueDate = DateInvoiced + 5), unlike pt_immediate.
     And metasfresh contains C_Invoice:
       | Identifier | C_BPartner_ID | C_DocTypeTarget_ID.Name       | DateInvoiced | IsSOTrx | C_Currency_ID | C_PaymentTerm_ID |
-      | proforma   | vendor        | Proforma-Rechnung (Lieferant) | 2026-04-24   | false   | EUR           | pt_immediate     |
+      | proforma   | vendor        | Proforma-Rechnung (Lieferant) | 2026-04-24   | false   | EUR           | pt_net5          |
     And metasfresh contains C_InvoiceLines
       | Identifier | C_Invoice_ID | M_Product_ID | QtyInvoiced | Price    |
       | proformaL1 | proforma     | product      | 1 PCE       | 21000.00 |
@@ -212,16 +216,20 @@ Feature: PO reactivation guard — downstream-activity check
 
     Then the order identified by po has following pay schedule lines by ReferenceDateType
       | ReferenceDateType | DueAmt   | DueAmt_Actual | ReferenceDate | DueDate    | Status |
-      | LC                | 21000.00 | 21000.00      | 2026-04-24    | 2026-04-24 | P      |
+      | LC                | 21000.00 | 21000.00      | 2026-04-24    | 2026-04-29 | P      |
 
     # Reactivate (allowed - proforma allocation no longer blocks) and re-complete the order
     And the order identified by po is reactivated
     And the order identified by po is completed
 
-    # Re-complete must restore the same proforma-derived Paid state
+    # Load-bearing columns: Status, DueAmt_Actual and DueDate actually prove the LC recompute
+    # fired on re-complete (DueDate = proforma.DueDate, 5 days past DateInvoiced — the naive
+    # create-path that only reads LC_Date + pt_od's zero OffsetDays cannot reproduce it).
+    # ReferenceDate and LC_Date coincide with DateInvoiced either way, so they document the
+    # intended state rather than prove the recompute ran.
     Then the order identified by po has following pay schedule lines by ReferenceDateType
       | ReferenceDateType | DueAmt   | DueAmt_Actual | ReferenceDate | DueDate    | Status |
-      | LC                | 21000.00 | 21000.00      | 2026-04-24    | 2026-04-24 | P      |
+      | LC                | 21000.00 | 21000.00      | 2026-04-24    | 2026-04-29 | P      |
     And validate the created orders
       | Identifier | LC_Date    |
       | po         | 2026-04-24 |
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/order/orderReactivatePayScheduleGuard.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/order/orderReactivatePayScheduleGuard.feature
@@ -171,3 +171,57 @@ Feature: PO reactivation guard — downstream-activity check
 
     # Guard blocks — pay-schedule line carries a downstream goods-receipt link
     And the order identified by po cannot be reactivated
+
+
+  @from:cucumber
+  @Id:S30621_TC4
+  Scenario: Reactivate then re-complete preserves the paid-proforma pay-schedule state
+    # A buyer completes a PO, allocates a proforma invoice, and pays it in full (payment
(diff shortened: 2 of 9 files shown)


backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
```diff
@@ -785,6 +790,13 @@ public void setBillBPartnerIdFromEffectiveResolution(final I_C_Order order)
 	public void createOrderPaySchedules(final I_C_Order order)
 	{
 		orderPayScheduleService.createOrderPaySchedules(order);
+
+		// Reactivate deletes every C_OrderPaySchedule row (see deleteOrderPaySchedules below), so a
+		// reactivate -> re-complete round trip rebuilds the schedule from the payment-term breaks alone,
+		// losing any proforma/prepayment-derived state (Paid/Awaiting_Pay, DueAmt_Actual, ReferenceDate,
+		// DueDate, LC_Date). Re-derive it here from the surviving C_Proforma_Order_Alloc / completed
+		// prepayment so a paid proforma allocation is restored rather than silently dropped.
+		orderPayScheduleLCStepService.recomputeLCStep(OrderId.ofRepoId(order.getC_Order_ID()));
```
teosarca: IMHO we shall not call orderPayScheduleLCStepService directly, but we shall have a OrderPayScheduleService.recomputeLCStep instead.

more, the name sounds misleading because i would expect to be applied for all steps that are affected by the proforma.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/order/orderReactivatePayScheduleGuard.feature
```diff
@@ -171,3 +174,62 @@ Feature: PO reactivation guard — downstream-activity check
 
     # Guard blocks — pay-schedule line carries a downstream goods-receipt link
     And the order identified by po cannot be reactivated
+
+
+  @from:cucumber
+  @Id:S30621_TC4
+  Scenario: Reactivate then re-complete preserves the paid-proforma pay-schedule state
```
teosarca, line 189: do we have a test for the B/L date case too? are there any other cases?

backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
teosarca: **Resolution (dev):** Both points taken, but the first one needs your call because the literal form closes a dependency cycle — I don't want to redesign around your comment silently.

**On the facade.** `OrderPayScheduleLCStepService` already constructor-injects `OrderPayScheduleService` (field `orderPayScheduleService`). So `OrderPayScheduleService.recomputeLCStep(...)` delegating to the step service would make `OrderPayScheduleService → OrderPayScheduleLCStepService → OrderPayScheduleService`. That is exactly why the call sits in the interceptor today — the interceptor is a leaf, so it can depend on both without a cycle.

Ways out, if you want the facade:
1. Break the back-edge: find what `OrderPayScheduleLCStepService` actually needs from `OrderPayScheduleService` and pull that into something both can depend on. Cleanest, largest.
2. Keep the orchestration out of both and give the step family a dedicated orchestrator (see the naming point below) that the interceptor calls. Middling.
3. Lazy/`ObjectProvider` injection to tolerate the cycle. Works, but hides the coupling rather than fixing it — I'd rather not.

**On the name — you're right, and it's the more interesting half.** `recomputeLCStep` is LC-specific, but the trigger isn't: reactivate deletes *every* `C_OrderPaySchedule` row, so re-complete has to re-derive whatever any referenced document had contributed. There is already a sibling step package (`steps/material_receipt` beside `steps/letter_of_credit`), so the honest shape is a step-family recompute — something like `recomputeStepsFromReferencedDocuments(orderId)` — that fans out to each step service, rather than the interceptor naming one step.

That also subsumes option 2 above: the orchestrator is the natural home for the facade, and it makes the call site say what it means.

I'd rather do this properly than patch the name. Tell me which you want:
- **(a)** ship as-is now (the behaviour is correct and covered) and open a follow-up for the orchestrator + rename;
- **(b)** do the orchestrator + rename in this PR before merge.

I've held off touching it until you decide.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/order/orderReactivatePayScheduleGuard.feature
teosarca: **Resolution (dev):** Real gap — confirmed, and I'm closing it.

Checked the feature file: the round-trip scenario is built on `pt_od` (LC 30% + OD 70%). `pt_bl` (LC 30% + BL 70%) appears only in the *blocked* scenario, as the carrier for the goods-receipt link. So there is currently **no** reactivate → re-complete test for a bill-of-lading term, and the B/L path is materially different: the BL break resolves only on goods receipt, so its row is `Pending` with `DueDate` at the infinite-future sentinel at completion time, where OD resolves immediately.

Adding a B/L variant of the round trip. It has to prove the same thing the OD one does — that the assertion fails without the fix — so I'll verify it the same way, by temporarily reverting the recompute call and confirming it goes red for the right reason rather than passing incidentally.

**Other cases I checked while I was in there**, so this is a considered answer rather than just the one you named:

- **Unpaid proforma round trip** — allocated but no completed payment, so the LC row is `Awaiting_Pay` rather than `Paid`. Distinct branch in the recompute (`awaitingPayment()` vs `paid()`) and currently untested across the round trip. Worth adding.
- **Reactivate → re-complete with no proforma at all** — the inert path. Every completion now runs the recompute, so this is the regression guard for "we didn't break ordinary orders". Not covered by a round-trip scenario today.
- **Goods receipt + proforma together** — covered at unit level (two of the six surviving tests pair a proforma with an in/out or invoice link), and the blocked scenario covers the receipt link without a proforma. I think that combination is adequately pinned; say if you disagree.

I'll add the B/L round trip plus the two gaps above unless you'd rather scope it to B/L only.

backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
teosarca: if the feature is working OK for LC and B/L and any other proforma affected step, i am fine shipping it as is now. pls confirm

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/order/orderReactivatePayScheduleGuard.feature
teosarca: pls confirm when u have the test(s) and not before!

backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
teosarca: **Resolution (dev):** Understood — shipping as-is, and I will not claim it works until tests say so.

Tests in flight now, covering exactly what you named: a B/L-term reactivate → re-complete round trip, an unpaid-proforma (`Awaiting_Pay`) round trip, and a no-proforma round trip as the regression guard for ordinary orders. Each will be proved diagnostic by temporarily reverting the recompute call and checking it goes red for the right reason — except the no-proforma one, which must stay green with the call removed since it guards the inert path.

I will confirm here with the per-scenario results, not before.

One thing I can already state as a code fact rather than a test result, so you can judge the shape of the claim: B/L rows are never a resolution target of the recompute. It resolves either the single LC line or the single prepaid line (`isPrepaidLine()` = letter-of-credit-date or order-date), while a B/L row is material-receipt-typed (`isMaterialReceiptDate()` = bill-of-lading-date or ETA-date). So B/L is untouched by construction — and an order that actually has a B/L *receipt* cannot be reactivated at all, because the goods-receipt link still blocks. That is the reasoning the B/L test has to confirm empirically, not a substitute for it.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/order/orderReactivatePayScheduleGuard.feature
teosarca: **Resolution (dev):** Acknowledged — not confirming before the tests exist.

Three scenarios are being written and run now: B/L-term round trip, unpaid-proforma (`Awaiting_Pay`) round trip, and a no-proforma round trip. I will post the per-scenario verdicts here, plus the temporarily-reverted run showing the first two fail without the fix, once they are green.

Separately, and unrelated to your point: CI on this PR went red on `test (java)` in `de.metas.business.rest-api-impl` and `de.metas.swat.base`. That is a real break, not a flake — both modules hold tests that construct `de.metas.order.model.interceptor.C_Order`, whose constructor gained `OrderPayScheduleLCStepService`. Being reproduced locally and fixed; not merge-ready until it is.

backend/de.metas.business/src/main/java/de/metas/order/model/interceptor/C_Order.java
teosarca: sorry i was talking about the "order date" steps, not the B/L date steps.
i mean all steps which are affected by a proforma+its payment.

------

on followup needed for the renames.

teosarca: **Resolution (dev):** taking your correction first, because my previous answer in this thread proved the wrong thing.

`OrderPayScheduleLCStepService` resolves its target as *LC line, else single prepaid line*, and `OrderPayScheduleLine.isPrepaidLine()` is `isLetterOfCreditDate() || referenceDateType.isOrderDate()`. So the set of steps a proforma and its payment can affect is exactly {LC step, order-date advance step}; B/L and ETA lines are `isMaterialReceiptDate()` and are never a target. My earlier reply argued that B/L is untouched — true, but beside the point you were making. The mistake was answering the literal words "B/L date" instead of first enumerating which `ReferenceDateType`s `isPrepaidLine()` admits. Your "the name sounds misleading" objection is simply correct: `recomputeLCStep` recomputes the order-date advance step too.

**That correction found a real defect in this PR — stating it before you have to.**

On a payment term with an **order-date break and no LC break**, the target *is* the order-date line, `proforma == null` holds on an ordinary completion, and the unconditional `recomputeLCStep` in the `AFTER_COMPLETE` handler resets the line that `createOrderPaySchedules` had correctly built one statement earlier — back to `Pending` / `9999-12-01`, reference date nulled.

Six scenarios in `purchaseOrder/purchaseOrderComplexPaymentTerm.feature` (a file this PR does not touch) are red on exactly that:

```
[DueDate] expected: 2025-10-10 but was: 9999-12-01
[Status]  expected: Awaiting_Pay but was: Pending
```

https://github.com/metasfresh/metasfresh/actions/runs/30610090698/job/91094154584

The split is not incidental: of the 9 scenarios in that file, the 6 whose term has an order-date break and no LC break fail (`pt_PO_3`, `pt_PO_4`, `pt_PO_5`, `pt_PO_6`, `pt_PO_7`, `pt_PO_8`), and the 3 whose term carries an LC break pass (`pt_PO`, `pt_PO_2`). `test (cucumber) (profile4)` was green on this branch's exact merge-base, so this is caused here.

**Why the reactivate feature could not catch it.** Every payment term in `order/orderReactivatePayScheduleGuard.feature` carries an LC break (`pt_od` = LC 30% + OD 70%, `pt_bl` = LC 30% + BL 70%), so `getSingleLCLine()` always wins and the order-date line is never the recompute target in any of TC1–TC7. The order-date-as-target path is covered only at unit level, plus one non-round-trip scenario in `purchaseOrderComplexPaymentTerm.feature` — which is itself one of the 6 currently red.

**Call-site audit of the reset**, since it decides where the fix belongs: `ProformaOrderAllocService#deallocate` is the only caller that can reach `proforma == null` *and* wants the reset — that is its whole purpose. `ProformaOrderAllocateCommand`, `recomputeLCStepAfterPaymentCompleted` and `recomputeLCStepAfterPaymentReversed` cannot reach it. The new `AFTER_COMPLETE` call reaches it on every completion of every order without a proforma. So the reset semantics stay untouched and the new call site is what gets guarded.

**Committed so far**

- https://github.com/metasfresh/metasfresh/commit/06aead357406bea4a133a1a866e9b67306ac252e — three scenarios on the guard feature: the paid-proforma round trip on a B/L term, the same round trip with the proforma allocated but **not** paid (so the `Awaiting_Pay` branch is exercised, not only `paid()`), and an ordinary-order regression guard. All 7 scenarios of the feature pass locally (`tests=7 failures=0 errors=0 skipped=0`).
- https://github.com/metasfresh/metasfresh/commit/cbf108793beb5f15318c8f6ee2a1d359090c049f — the `test (java)` break: `C_Order.newInstanceForUnitTesting()` now transitively resolves `ProductTaxCategoryService` from the spring context, so both interceptor tests must register that bean.

**Not done — no confirmation claimed on any of these**

- The guard at the new call site is not committed.
- No reactivate → re-complete scenario exists yet on a no-LC, order-date-break term — i.e. the case you are actually asking about. That is the RED for the fix.
- CI has not yet run a build containing either commit above.

I am not confirming the feature until the order-date path is green in a round-trip test rather than at unit level only, and I will post the per-scenario verdicts here when I have observed them.

**Renames:** understood — follow-up, not this PR. I will get the follow-up agreed before opening it and link it here.

teosarca: **Resolution (dev):** the order-date path is now covered and green. Observed results, as promised.

**The defect is fixed** — https://github.com/metasfresh/metasfresh/commit/1fbb6157f3dbe47b296afd9c5ade82344148decd

The `AFTER_COMPLETE` call site now goes through a guarded sibling of the two existing `recomputeLCStepAfter<Event>` methods:

```java
public void recomputeLCStepAfterOrderCompleted(@NonNull final OrderId orderId)
{
    if (!proformaService.getIdByOrderId(orderId).isPresent())
    {
        return;
    }
    recomputeLCStep(orderId);
}
```

`recomputeLCStep`'s `proforma == null` branch is untouched, because `ProformaOrderAllocService#deallocate` legitimately depends on that reset — it is the only caller that both reaches it and wants it. The guard uses the same query the guarded branch itself uses (`getByOrderId` is `getIdByOrderId(orderId).map(this::getById)`, backed by `findProformaInvoiceIdByOrderId` with `addOnlyActiveRecordsFilter()`), so guard and branch cannot diverge.

**Order-date coverage** — https://github.com/metasfresh/metasfresh/commit/27090b7129de9a123e73c18307faa3dfb50490a6

`S30621_TC8`: a new no-LC term `pt_od_nolc` (OD 25% at +1 day, BL 75%), advance paid via proforma, reactivate → re-complete. This is the first scenario in the file where the order-date line is actually the recompute target — every earlier term carries an LC break, so `getSingleLCLine()` always won and the OD line was never reached.

RED with the production fix reverted (`git checkout 06aead35740 -- <3 files>`, full rebuild, then TC8 alone) — `FAILED: 1 of 1 scenarios failed`:

```
-- failure 1 --  [ReferenceDate] expected: 2026-04-24 but was: null
-- failure 2 --  [DueDate]       expected: 2026-04-25 but was: 9999-12-01
-- failure 3 --  [Status]        expected: Awaiting_Pay but was: Pending
```

Fix restored and rebuilt → green.

**Per-scenario verdicts, local runs at `27090b7129d`**

`order/orderReactivatePayScheduleGuard.feature` — `All 8 scenarios passed`:

| | Scenario | |
|---|---|---|
| TC1 | `Awaiting_Pay` OD row, no downstream | PASS |
| TC2 | proforma allocation exists, nothing downstream | PASS |
| TC3 | blocked — goods receipt links a pay-schedule line | PASS |
| TC4 | preserves the paid-proforma state | PASS |
| TC5 | …on a B/L term | PASS |
| TC6 | preserves the awaiting-pay proforma state | PASS |
| TC7 | ordinary order left untouched | PASS |
| TC8 | paid-proforma state on the order-date step of a no-LC term | PASS |

`purchaseOrder/purchaseOrderComplexPaymentTerm.feature` — `All 9 scenarios passed` (the 6 that the regression broke, plus the 3 that were never affected).

JUnit:

```
Tests run: 13, Failures: 0, Errors: 0, Skipped: 0 -- OrderPayScheduleLCServiceTest
Tests run: 10, Failures: 0, Errors: 0, Skipped: 0 -- OrderTest
```

The 13 include two new tests pinning both sides of the guard.

**The "accepted risk" in the PR description is gone.** It said the multiple-LC-break / multiple-advance-break fail-fast now trips on every completion of such a term. With the guard in place that is again reachable only for orders carrying a proforma allocation — i.e. as it was before this PR. I am updating the description accordingly.

CI is running on `27090b7129d`. `test (cucumber) (profile7)` will be red there and is not from this PR: its 18 failures are present on base `deep_tundra_release`, correlating with https://github.com/metasfresh/metasfresh/pull/25365 on every build that contains it and absent on every build that does not, and that commit adds no profile7 scenarios.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/order/orderReactivatePayScheduleGuard.feature
teosarca: **Resolution (dev):** confirming now that the tests exist and are green — replacing my previous reply in this thread, which was a promise and should not have been posted as a resolution.

**B/L date case** — `@Id:S30621_TC5` *"Reactivate then re-complete preserves the paid-proforma pay-schedule state on a B/L term"*, added in https://github.com/metasfresh/metasfresh/commit/06aead357406bea4a133a1a866e9b67306ac252e.

**The other cases you asked about** — the feature file now carries all eight:

| | Scenario | Local verdict at PR HEAD |
|---|---|---|
| TC1 | Reactivate allowed — `Awaiting_Pay` OD row, no downstream | PASS |
| TC2 | Reactivate allowed — proforma allocation exists, nothing committed downstream | PASS |
| TC3 | Reactivate blocked — goods receipt links a pay-schedule line | PASS |
| TC4 | Reactivate → re-complete preserves the paid-proforma state | PASS |
| TC5 | …on a B/L term | PASS |
| TC6 | …preserves the awaiting-pay (unpaid) proforma state | PASS |
| TC7 | …leaves an ordinary order's pay schedule untouched | PASS |
| TC8 | …preserves the paid-proforma state on the order-date step of a no-LC term | PASS |

`All 8 scenarios passed`. TC5/TC6/TC7 came in with https://github.com/metasfresh/metasfresh/commit/06aead357406bea4a133a1a866e9b67306ac252e, TC8 with https://github.com/metasfresh/metasfresh/commit/27090b7129de9a123e73c18307faa3dfb50490a6.

**They are diagnostic, not incidentally green.** With the recompute call commented out, TC4/TC5/TC6 fail and TC7 stays green — the signature a correct guard must produce. TC8 was RED-verified separately against the reverted production fix:

```
-- failure 1 --  [ReferenceDate] expected: 2026-04-24 but was: null
-- failure 2 --  [DueDate]       expected: 2026-04-25 but was: 9999-12-01
-- failure 3 --  [Status]        expected: Awaiting_Pay but was: Pending
```

**Adjacent coverage** — `purchaseOrder/purchaseOrderComplexPaymentTerm.feature`: `All 9 scenarios passed` (the 6 the regression broke, plus the 3 never affected). JUnit: `OrderPayScheduleLCServiceTest` `Tests run: 13, Failures: 0, Errors: 0, Skipped: 0` (12 methods — 11 plain plus one `@EnumSource`-parameterized over the two allocation states; two of the 13 pin both sides of the new completion guard), `OrderTest` `Tests run: 10, Failures: 0`, `C_Order_Test` 6 tests on the guard predicate.

**Two corrections to my earlier reply in this thread.** The `test (java)` red is fixed (https://github.com/metasfresh/metasfresh/commit/cbf108793beb5f15318c8f6ee2a1d359090c049f registers the interceptor's new collaborator as a JUnit bean), so the "not merge-ready" caveat no longer stands. And I predicted `test (cucumber) (profile7)` would be red at that HEAD — it passed: https://github.com/metasfresh/metasfresh/actions/runs/30618383055/job/91121731253. CI at PR HEAD https://github.com/metasfresh/metasfresh/commit/27090b7129de9a123e73c18307faa3dfb50490a6 is green on every job with `Merge Gate` pass: https://github.com/metasfresh/metasfresh/actions/runs/30618383055.

PR #25373, opened 2026-07-31
Costing - plain repost driver: accounting_docs_repost_all_from_date

teosarca:
## What

Adds `de_metas_acct.accounting_docs_repost_all_from_date(p_C_AcctSchema_ID, p_StartDateAcct)` — a **plain repost** driver — plus its recorded migration `5817140` and cucumber coverage.

## Why

When an accounting schema's **costing method is switched**, the cost details are already correct but every `Fact_Acct` row is still valued by the OLD method: the amount a document posts is filtered to the cost element accountable under the schema's *current* costing configuration (`CostDetailCreateResultsList#getAmtAndQtyToPost`). So every affected document has to be reposted — and nothing else. Recomputing would delete and rebuild exactly the cost details the switch just established.

There was no way to ask for that. `product_costs_recreate_all_from_date` always deletes the cost details in range so that reposting rebuilds them, and `fact_acct_unpost` cannot do it at scale (see below).

## How

The new function is the counterpart of `product_costs_recreate_all_from_date` and deliberately mirrors it where it matters:

- **same document set** — same `accountable_docs_and_lines_v` predicate (`m_product_id IS NOT NULL`, `tablename <> 'C_Invoice'`, `dateacct >= p_StartDateAcct`, client-scoped), collapsed to one row per document;
- **same release order and SeqNo discipline** — the block of SeqNos is reserved by advancing the queue's own sequence in a single statement, and each SeqNo is assigned *explicitly* via `row_number() OVER (ORDER BY dateacct, tablename_prio, record_id)`. Never left to the column's `nextval` default: the order in which `INSERT … SELECT` evaluates rows, and therefore consumes a sequence, is not guaranteed by an `ORDER BY` in the `SELECT`;
- **same session advisory lock** — the same lock class and key as the recompute driver, so the two exclude *each other*, not merely their own kind. Enqueuing documents while a recompute has them parked in its staging table would repost them against half-rewound costs.

And differs where it must: it deletes nothing (no cost detail, no `M_Cost` rewind, no `Fact_Acct` purge), stages nothing, records no progress, and has no resume — the enqueue is a single `INSERT` in a single transaction, so the run either happened or it did not.

One thing it must do that the recompute driver gets for free: it **sweeps `assert_period_open` over the whole range before enqueuing anything**. `Doc#post` rejects an already-posted document whose period is closed even when reposting, and every document this function enqueues is already posted. Without the sweep the run would look like a success — the rows are on the queue — and then fail document by document in the accounting server, one `AD_Issue` at a time, with the ledger left half re-valued.

### Known, deliberate rule deviation

`docs/coding-rules/production-database-support.md` and `backend/de.metas.acct.base/CLAUDE.md` both say: never `INSERT` directly into `Accounting_Docs_To_Repost`, always use `de_metas_acct.fact_acct_unpost*`. This function breaks that knowingly, on the same grounds as the recompute driver's release step, and **says so in its own body** rather than hiding it. Of the three effects `fact_acct_unpost` bundles with the enqueue, two are provably redundant here (the `Fact_Acct` delete — the repost does it itself via `PostingService#postNow` → `Doc#post` → `deleteAcct()`; and the `Posted='N'` reset — `SqlAcctDocLockService#lock` only adds `AND Posted='N'` when `repost` is false) and the third, the period check, is replaced by the explicit sweep, which is strictly stronger: it covers the whole range up front instead of one document at a time.

The `fact_acct_unpost` loop was considered and **rejected**, also recorded in the body: at its default `p_Force:='N'` it aborts on the first document already `Processing='Y'` (over a range holding hundreds of thousands of documents), it deletes every `Fact_Acct` row up front — work the repost redoes anyway, leaving the ledger empty until the queue drains — and it enqueues one document at a time with the SeqNo left to the sequence, so no contiguous ordered block can be reserved.

The coding rule itself is deliberately **not** amended — that is not this function's call to make.

## Test

A new scenario in `switchToMovingAverageInvoice.feature`. It books stock counts whose accounting-date order is deliberately the *opposite* of the product order, switches the schema to `MovingAverageInvoice` back-dated to a cut-off, and then asserts that the repost:

- enqueues exactly the in-range documents, in accounting-date order, none of them staged;
- gives every queued document its own SeqNo (the poller consumes by SeqNo and the queue has no uniqueness constraint on it);
- leaves the `M_CostDetail` row count **unchanged** — both immediately after the call *and* again after the queue has drained, which is where a recompute would have rebuilt the costs;
- leaves the pre-cut-off opening anchors intact.

Verified locally against a local docker stack: RED first (the scenario fails at the driver step with `function … does not exist`), then GREEN after applying migration `5817140`.

### DIFF
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/costing/moving_average_invoice/switchToMovingAverageInvoice.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/costing/moving_average_invoice/switchToMovingAverageInvoice.feature
@@ -821,3 +821,18 @@ Feature: Switch to Moving Average Invoice
       | repostFirst  | MovingAverageInvoice | 2025-12-31 | 0 PCE | 0 CHF |
       | repostSecond | MovingAverageInvoice | 2025-12-31 | 0 PCE | 0 CHF |
       | repostThird  | MovingAverageInvoice | 2025-12-31 | 0 PCE | 0 CHF |
+    #
+    # And finally what the whole driver exists for: the general ledger. Each reposted stock count carries its
+    # two facts again - the stock going up against the warehouse differences account, valued from the cost
+    # details above. The match is exhaustive per document, so a repost that wiped the ledger, or one that
+    # posted a second time on top of the first, fails here.
+    #
+    And Wait until documents invRepost1, invRepost2, invRepost3 are posted
+    And Fact_Acct records are matching
+      | Record_ID  | AccountConceptualName | M_Product_ID | AmtAcctDr | AmtAcctCr | Qty     |
+      | invRepost1 | P_Asset_Acct          | repostFirst  | 100       | 0         | 10 PCE  |
+      | invRepost1 | W_Differences_Acct    | repostFirst  | 0         | 100       | -10 PCE |
+      | invRepost2 | P_Asset_Acct          | repostSecond | 100       | 0         | 10 PCE  |
+      | invRepost2 | W_Differences_Acct    | repostSecond | 0         | 100       | -10 PCE |
+      | invRepost3 | P_Asset_Acct          | repostThird  | 100       | 0         | 10 PCE  |
+      | invRepost3 | W_Differences_Acct    | repostThird  | 0         | 100       | -10 PCE |
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/costing/moving_average_invoice/switchToMovingAverageInvoice.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/costing/moving_average_invoice/switchToMovingAverageInvoice.feature
@@ -814,7 +814,11 @@ Feature: Switch to Moving Average Invoice
       | M_InventoryLine | invPreCutRepost3_l1 | 100 CHF | 10 PCE |
       | M_InventoryLine | invRepost3_l1       | 100 CHF | 10 PCE |
     #
-    # Including the opening anchors the switch created: a repost must never reach back over the cut-off.
+    # Including the opening anchors the switch created: a repost must never reach back over the cut-off. Each one
+    # is still there exactly once and still dated AT the cut-off - surviving the repost is what this asserts.
+    # Their own Qty and Amt are zero BY DESIGN, as in every scenario above: an anchor is value-neutral and books
+    # no GL, so the copied opening lives in M_Cost and in the anchor's Prev_* columns, never in its own delta
+    # (seedCurrentCostFromOpening writes qty/amt as .toZero() and passes the opening as the Prev_* amounts).
     #
     And the cost revaluation identified by switchRepost seeded opening cost details:
       | M_Product_ID | M_CostElement_ID     | DateAcct   | Qty   | Amt   |
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/costing/moving_average_invoice/switchToMovingAverageInvoice.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/costing/moving_average_invoice/switchToMovingAverageInvoice.feature
@@ -698,3 +698,126 @@ Feature: Switch to Moving Average Invoice
       | lateJoiner        | MovingAverageInvoice | 2025-12-31 | 0 PCE | 0 CHF |
       | doneBatchProduct  | MovingAverageInvoice | 2025-12-31 | 0 PCE | 0 CHF |
       | crashBatchProduct | MovingAverageInvoice | 2025-12-31 | 0 PCE | 0 CHF |
+
+  @Id:S26253_TC13
+  Scenario: A plain repost enqueues the documents in accounting-date order and rebuilds no cost at all
+    #
+    # Once the accounting schema's costing method has been switched, the already-computed cost details have to
+    # reach the general ledger - and only that. Reposting regenerates Fact_Acct from the cost details that
+    # already exist; recomputing them again would be wrong, because they are what the switch just established.
+    #
+    # Same isolation technique as the batched-recompute scenarios above: this driver also has no product
+    # parameter, so the scenario is separated from the shared test database by ACCOUNTING DATE - it books its
+    # stock counts into the LAST THREE WEEKS of 2027 and reposts from 10.12.2027 onwards, later than every
+    # range the scenarios above reach into. It stays inside 2027, whose end is also the end of the test
+    # database's fiscal calendar.
+    #
+    Given after not more than 60s, the repost queue is drained
+    And metasfresh contains M_Products:
+      | Identifier   |
+      | repostFirst  |
+      | repostSecond |
+      | repostThird  |
+    #
+    # Pre-cut-off history on the prior method (AveragePO), so the switch has a non-trivial opening to copy.
+    #
+    And update current costs
+      | M_Product_ID | M_CostElement_ID | CurrentCostPrice |
+      | repostFirst  | AveragePO        | 10 CHF           |
+      | repostSecond | AveragePO        | 10 CHF           |
+      | repostThird  | AveragePO        | 10 CHF           |
+    And metasfresh contains single line completed inventories
+      | M_Inventory_ID   | M_InventoryLine_ID  | MovementDate | M_Warehouse_ID | M_Product_ID | QtyBook | QtyCount | UOM.X12DE355 |
+      | invPreCutRepost1 | invPreCutRepost1_l1 | 2025-12-15   | warehouseStd   | repostFirst  | 0       | 10       | PCE          |
+      | invPreCutRepost2 | invPreCutRepost2_l1 | 2025-12-15   | warehouseStd   | repostSecond | 0       | 10       | PCE          |
+      | invPreCutRepost3 | invPreCutRepost3_l1 | 2025-12-15   | warehouseStd   | repostThird  | 0       | 10       | PCE          |
+    #
+    # Switch to MovingAverageInvoice, back-dated to the 31.12.2025 cut-off: one opening anchor per product.
+    #
+    And metasfresh contains M_CostRevaluation:
+      | Identifier   | C_AcctSchema_ID | M_CostElement_ID     | RevaluationSource   | CopyFrom_M_CostElement_ID | EvaluationStartDate | DateAcct   |
+      | switchRepost | acctSchema      | MovingAverageInvoice | CopyFromCostElement | AveragePO                 | 2025-12-31          | 2025-12-31 |
+    And create lines for cost revaluation switchRepost
+    And the cost revaluation identified by switchRepost is completed
+    #
+    # One stock count per product inside the repost's range, dated OPPOSITE to the product order: the driver
+    # must release by accounting date, so the two orders have to disagree to prove anything.
+    #
+    When metasfresh contains single line completed inventories
+      | M_Inventory_ID | M_InventoryLine_ID | MovementDate | M_Warehouse_ID | M_Product_ID | QtyBook | QtyCount | UOM.X12DE355 |
+      | invRepost1     | invRepost1_l1      | 2027-12-28   | warehouseStd   | repostFirst  | 10      | 20       | PCE          |
+      | invRepost2     | invRepost2_l1      | 2027-12-20   | warehouseStd   | repostSecond | 10      | 20       | PCE          |
+      | invRepost3     | invRepost3_l1      | 2027-12-12   | warehouseStd   | repostThird  | 10      | 20       | PCE          |
+    #
+    # Wait until they are posted. Their cost details are what the repost has to REUSE - so they must exist
+    # before it runs, and the very same rows must still be there afterwards.
+    #
+    And after not more than 60s, M_CostDetails are found for product repostFirst and cost element AveragePO
+      | TableName       | Record_ID           | Amt     | Qty    |
+      | M_InventoryLine | invPreCutRepost1_l1 | 100 CHF | 10 PCE |
+      | M_InventoryLine | invRepost1_l1       | 100 CHF | 10 PCE |
+    And after not more than 60s, M_CostDetails are found for product repostSecond and cost element AveragePO
+      | TableName       | Record_ID           | Amt     | Qty    |
+      | M_InventoryLine | invPreCutRepost2_l1 | 100 CHF | 10 PCE |
+      | M_InventoryLine | invRepost2_l1       | 100 CHF | 10 PCE |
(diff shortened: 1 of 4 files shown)


backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/costing/moving_average_invoice/switchToMovingAverageInvoice.feature
```diff
@@ -698,3 +698,141 @@ Feature: Switch to Moving Average Invoice
...
+      | M_InventoryLine | invRepost2_l1       | 100 CHF | 10 PCE |
+    And after not more than 60s, M_CostDetails are found for product repostThird and cost element AveragePO
+      | TableName       | Record_ID           | Amt     | Qty    |
+      | M_InventoryLine | invPreCutRepost3_l1 | 100 CHF | 10 PCE |
+      | M_InventoryLine | invRepost3_l1       | 100 CHF | 10 PCE |
+    #
+    # Including the opening anchors the switch created: a repost must never reach back over the cut-off.
+    #
+    And the cost revaluation identified by switchRepost seeded opening cost details:
+      | M_Product_ID | M_CostElement_ID     | DateAcct   | Qty   | Amt   |
```
teosarca, line 824: why qty and amt are zero for all ?

teosarca: **Resolution (dev):** Already-correct — zero is what that row is *supposed* to carry. Added the missing inline explanation so the next reader does not have to ask: https://github.com/metasfresh/metasfresh/commit/4b0efe7cc6e72892513fb54592a618b934e0eb8e

**Why zero.** The opening anchor is value-neutral by construction. `CostingService.seedCurrentCostFromOpening` puts the copied opening into `M_Cost` and into the anchor's `Prev_*` columns, and forces the anchor's *own* `Qty`/`Amt` to zero — `backend/de.metas.business/src/main/java/de/metas/costing/impl/CostingService.java:596-613`:

```java
final CurrentCost currentCost = currentCostsRepo.getOrCreate(targetSegmentAndElement);
currentCost.setFrom(opening);          // <- 10 CHF / 10 PCE / 100 CHF land HERE
currentCostsRepo.save(currentCost);
...
utils.createCostDetailRecordWithChangedCosts(
        CostDetailCreateRequest.builder()
                ...
                .qty(opening.getQty().toZero())            // <- own Qty forced to 0
                .amt(opening.getCumulatedAmt().toZero())   // <- own Amt forced to 0
                .date(anchorDate)
                .build(),
        opening);                                          // <- opening -> Prev_* columns
```

`.toZero()` keeps UOM/currency and zeroes only the value — hence `0 PCE` / `0 CHF` rather than blank. It is part of the anchor's documented recognition signature (`CostRevaluationService.java:447-449`: "`M_CostRevaluationLine_ID IS NOT NULL` …, `IsChangingCosts='Y'`, own `Qty` and `Amt` zero, and `Prev_*` equal to the seeded opening"), and it is what makes the seed post no GL (`DocLine_CostRevaluation.java:55-70` — `.qty(costRevaluationLine.getCurrentQty().toZero())`, "the copy books nothing").

**So the zeros are constants of the row type, not scenario values** — same shape as the 17 already-merged `0 PCE | 0 CHF` rows in TC7/TC8/TC10/TC11/TC12 on `intensive_care_hotfix`. TC7 is where the *value* side is pinned instead: `validate current costs` → `10 CHF | 10 PCE | 100 CHF` sitting next to a `0 PCE | 0 CHF` anchor.

**What the step therefore discriminates here in TC13** is the other two columns, not the amounts: exactly ONE anchor per product scoped to *this* document's own revaluation line (`getSingleAnchor` → `assertThat(anchors).hasSize(1)`), and `DateAcct` still AT the 31.12.2025 cut-off. That is the assertion that the 2027 repost did not reach back over the cut-off and delete them.

**Left for you to decide** — the step asserts `DateAcct`/`Qty`/`Amt` but not `Prev_*`, so the opening *value* is not pinned at the anchor itself. Pinning it means adding `Prev_*` columns to `M_CostRevaluation_StepDef.costRevaluation_seededOpeningCostDetails`, which is shared by the 5 merged scenarios above — happy to do it in this PR if you want it, but I did not widen the shared step def on my own.

Commit `4b0efe7cc6e` is comment-only: verified with the Gherkin parser that the parsed structure of the feature (12 scenarios, every tag, step text and data-table row count) is identical to `6f2ecf86ef8`.

teosarca: ok but the standard case is that 
* we were using AveragePO before, we already have trxs for years
* we are switching to MAI, so on costrevaluation copy we will have actual cost prices

teosarca: **Resolution (dev):** Confirmed — the standard case is exactly as you describe, and TC13 does set one up, yet the scenario never asserts the actual copied price anywhere, only the zero delta. That is a real gap in what I wrote, so: why it happened first, then the options. I have changed nothing yet because the fix is a scope call.

**How it got there.** TC13 reused the anchor-assertion table verbatim from the merged scenarios instead of deriving the assertion from what TC13 has to prove. Those zero columns are the anchor row's *signature* — byte-identical in all six scenarios — so the table reads like a verification of the copied opening when in fact only its `DateAcct` and its existence discriminate anything. The shortcut was pattern-matching an established table shape rather than asking which column is load-bearing here. The check I will apply when reusing an assertion step in a new scenario: name the column that discriminates for THIS scenario, and if every column is a constant across the scenarios sharing the step, treat the step as documentation rather than verification and add a value-bearing assertion. Run against TC13 at authoring time that check fails at once — `Qty`/`Amt` are constants, `DateAcct` is the sole discriminator — and it lands on Option A below.

**The years of AveragePO history are not lost by the zero anchor.** The copy carries the source's *position* as of the cut-off — price, qty and cumulated amount — into two places, and only the anchor's own delta is zeroed:

| where | what it holds |
|---|---|
| `M_Cost` of the MAI element | `CurrentCostPrice` / `CurrentQty` / `CumulatedAmt` = the AveragePO position at the cut-off |
| anchor `M_CostDetail`, `Prev_*` columns | the same opening (`Prev_CurrentCostPrice`, `Prev_CurrentQty`, `Prev_CumulatedAmt`, `Prev_CumulatedQty`) |
| anchor `M_CostDetail`, own `Qty`/`Amt` | zero — so the anchor itself moves no stock and books no GL |

So the years of transactions are represented as an **opening balance**, not replayed — `CostRevaluationService.createDetailsForCopyFromCostElement` builds one `opening` from `getCostAsOf(source, EvaluationStartDate)` and hands the same object to both `M_Cost` and the anchor's `Prev_*`.

**TC13's own setup is not a zero-price case either:** AveragePO seeded at `10 CHF` plus a pre-cut-off stock count of `10 PCE` on 2025-12-15 per product, both already asserted green (`| M_InventoryLine | invPreCutRepost1_l1 | 100 CHF | 10 PCE |`). So at the 31.12.2025 cut-off the copied opening is a real `10 CHF / 10 PCE / 100 CHF` — it is just invisible in the scenario, because the only step that looks at the anchor asserts `DateAcct` + own `Qty`/`Amt` and never `Prev_*`. TC7 is currently the only scenario that pins the copied value at all, via `validate current costs` → `MovingAverageInvoice | 10 CHF | 10 PCE | 100 CHF`.

**Option A — assert the copied opening inside TC13, with the existing step.** Add `validate current costs` for the three products (`MovingAverageInvoice | 10 CHF | 10 PCE | 100 CHF`) after the switch, and again after the repost to prove a plain repost leaves it untouched.
*Pro:* confined to TC13, no shared step def touched, nothing merged has to be re-verified. *Con:* the anchor row's own `Prev_*` still go unasserted, so the next reader of the anchor table sees the same zeros. *Cost:* one local TC13 cucumber run to confirm GREEN before pushing.

**Option B — extend the shared step def with `Prev_*` columns.** Add `Prev_CurrentCostPrice` / `Prev_CurrentQty` / `Prev_CumulatedAmt` to `M_CostRevaluation_StepDef.costRevaluation_seededOpeningCostDetails`, so every anchor assertion shows the actual copied price next to the zero delta.
*Pro:* fixes it at the source — the actual price becomes visible in all six scenarios that assert an anchor, which is exactly the gap you are pointing at. *Con:* the step is shared by five already-merged scenarios (TC7, TC8, TC10, TC11, TC12), so their tables change too and all of them need re-running, not just TC13. *Cost:* a local run of the six scenarios.

**Recommendation: B**, because the zeros being the only visible number is the actual defect and B removes it everywhere rather than working around it in one scenario; A is the right pick only if you want zero churn on the merged scenarios. Both are small — the difference is blast radius, not effort. Happy to do A+B together if you want the post-repost invariant pinned as well.

teosarca: **Resolution (dev):** Correcting my own earlier reply in this thread — https://github.com/metasfresh/metasfresh/pull/25373#discussion_r3690332470 — which conceded a coverage gap and put two options to you. That framing was wrong, and I do not want it left on the record.

### Root cause of the wrong framing

I assessed the gap from **TC13 alone** — the scenario I had just written — instead of from the feature file as a whole. TC13 does not re-assert the copied MovingAverageInvoice price, so I reported that price as unasserted. It is in fact already asserted, in **TC7**: the scenario that models exactly the case you raised. The step I skipped was checking the feature for an existing `validate current costs` on the *target* element before claiming any assertion was missing — a single grep, which would have surfaced `:303` immediately.

### The evidence: TC7 is your standard case, and it pins the copied price

`backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/costing/moving_average_invoice/switchToMovingAverageInvoice.feature`, TC7 — real AveragePO purchase history, then a back-dated switch. It is on `intensive_care_hotfix` and untouched by this PR (the PR's change to this file is `142 insertions, 0 deletions`, all of it the new TC13):

- `:263` — AveragePO at the cut-off, from an actual 2025-12-15 purchase + material receipt of 10 PCE @ 10 CHF:
  `| acctSchema | product | AveragePO | 10 CHF | 10 PCE | 100 CHF |`
- `:288` — a post-cut-off purchase (2026-01-15, 10 PCE @ 30 CHF) moves the still-active AveragePO on to a value it never had at the cut-off:
  `| acctSchema | product | AveragePO | 20 CHF | 20 PCE | 400 CHF |`
- `:303` — **the copied MovingAverageInvoice opening, asserted:**
  `| acctSchema | product | MovingAverageInvoice | 10 CHF | 10 PCE | 100 CHF |`
- `:309` — the zero-delta anchor, in the **same** scenario:
  `| product | MovingAverageInvoice | 2025-12-31 | 0 PCE | 0 CHF |`

So a copy that took the wrong price — the live post-cut-off `20 CHF | 20 PCE | 400 CHF` instead of the as-of-the-cut-off `10 CHF | 10 PCE | 100 CHF` — **fails a test today**, at `:303`. The actual cost prices you describe are covered, and they are covered in the scenario whose subject is precisely that.

### The zeros are scenario-independent, not merely correct-by-design

`CostRevaluationService.createDetailsForCopyFromCostElement` builds ONE fully-priced `opening` from the source's cost as of the cut-off — `getCostAsOf` is where "trxs for years" is actually read (`backend/de.metas.business/src/main/java/de/metas/costrevaluation/CostRevaluationService.java:481-498`):

```java
final CostDetailPreviousAmounts sourceCostAsOfCutoff = costingService.getCostAsOf(sourceSegmentAndElement, costRevaluation.getEvaluationStartDate())
...
final CostAmount cumulatedAmt = sourceCostPrice.getOwnCostPrice().multiply(sourceQty);
final CostDetailPreviousAmounts opening = CostDetailPreviousAmounts.builder()
        .costPrice(sourceCostPrice)     // 10 CHF
        .qty(sourceQty)                 // 10 PCE
        .cumulatedAmt(cumulatedAmt)     // 100 CHF
        .cumulatedQty(sourceQty)
        .build();
```

That same object then goes to two places, with **no conditional on how rich it is** (`backend/de.metas.business/src/main/java/de/metas/costing/impl/CostingService.java:596-613`):

```java
currentCost.setFrom(opening);                        // <- the ACTUAL prices land in M_Cost, in full
currentCostsRepo.save(currentCost);
...
utils.createCostDetailRecordWithChangedCosts(
        CostDetailCreateRequest.builder()
                ...
                .qty(opening.getQty().toZero())           // <- anchor's OWN qty: always 0
                .amt(opening.getCumulatedAmt().toZero())  // <- anchor's OWN amt: always 0
                .date(anchorDate)
                .build(),
        opening);                                        // <- opening -> the anchor's Prev_*
```

`.toZero()` is applied to whatever arrives, so a years-of-history opening of `10 CHF | 10 PCE | 100 CHF` and an empty one of `10 CHF | 0 PCE | 0 CHF` produce the **identical** `0 PCE | 0 CHF` anchor, differing only in `M_Cost` and in `Prev_*`. The zeros therefore cannot indicate a thin fixture — they are a constant of the row type, and part of the anchor's documented recognition signature (`CostRevaluationService.java:447-449`).

TC10/TC11 corroborate that empirically rather than by reading the code: both seed 10 PCE of priced pre-cut-off stock per product, and both still assert `0 PCE | 0 CHF` anchors. TC13 does the same (`:724-733` gives each product `10 CHF` AveragePO plus a 2025-12-15 stock count of 10 PCE, and `:757` asserts the resulting `100 CHF | 10 PCE` AveragePO cost detail).

### What is genuinely not asserted anywhere

The anchor's own `Prev_*` columns — the audit trail of *what was replaced*. `M_Cost` after the copy is asserted (`:303`), and the anchor's own `Qty`/`Amt`/`DateAcct` are asserted (`:309` and the equivalent rows in TC10/TC11/TC12/TC13), but no test reads the anchor's `Prev_*` back. Covering it would mean widening `M_CostRevaluation_StepDef.costRevaluation_seededOpeningCostDetails`, which five already-merged scenarios share.

You have weighed that and decided the coverage is sufficient as it stands: no change to the shared step def, no change to TC13, nothing re-run. Recording it here so the choice is deliberate and visible rather than an oversight.

**Outcome: no code or test change for this thread.** The only commits it produced are the two comment-only ones already posted — https://github.com/metasfresh/metasfresh/commit/4b0efe7cc6e72892513fb54592a618b934e0eb8e and https://github.com/metasfresh/metasfresh/commit/7d3dd04a2f1fe117487f73866fe24f90aceb0a89 — and TC13's inline note at `:819-821` remains accurate as written.

PR #25386, opened 2026-07-31
MobileUI Picking: harden the whole-TU over-delivery guard (in-flight guard, CU parity test, catch-weight regression cover)

teosarca:
Post-merge hardening of the whole-TU over-delivery guard, following up on https://github.com/metasfresh/metasfresh/pull/25374. Every item below is a place where the whole-TU path and its CU sibling still diverged, or where their parity was unproven.

## 1. The confirm-Yes handler did not await the booking

`ScanHUAndGetQtyComponent`'s confirm-Yes branch called `fireOnResult` **un-awaited**, so the dialog closed and the always-listening HU scanner re-armed while the pick POST was still in flight — no spinner, no disabled state. On a handheld over warehouse WiFi that is a double-pick. The direct path already awaited the same call.

`fireOnResult` now mirrors the CU sibling `GetQuantityDialog.fireOnQtyChange` 1:1: set `isProcessing`, return the promise so the direct path keeps its sequencing, and clear the flag in a `finally` — **including on the error branch**, so a failed pick lets the operator scan again. While processing, the component renders a `Spinner` via early return, which *unmounts* the scanner rather than overlaying it (the app's documented mechanism for suppressing an always-listening scanner).

**TDD:** three Jest tests written first and confirmed RED against the unfixed code — the spinner assertion, the double-`onResult` assertion (`Expected 1, received 2` — the double-pick itself), and the clears-on-error assertion.

## 2. CU / whole-TU parity test

The two paths must behave identically — one physical action for the operator. The agreed enforcement mechanism is *shared helper + a parity test*, and the test did not exist.

It asserts the **structural** property: `validateQtyEntered` (CU) and `requestQtyOrReportResult` (whole-TU) both resolve their ceiling through the single `validateQtyAgainstMax` helper, there is exactly one definition of it, and `DEFAULT_MSG_qtyAboveMax` is referenced nowhere outside it. Structural rather than outcome-based on purpose: on `new_dawn_uat`'s tip the ceiling currently exists as two *behaviour-identical* copies, so an outcome-only test would be green there and would miss a re-duplication. Verified by mutation — re-inlining the CU ceiling exactly as `new_dawn_uat` has it turns the test red.

No production refactor was needed; `validateQtyAgainstMax` already existed and both paths already called it.

## 3. Regression coverage — a weight-only excess must not trigger the guard

Two Playwright scenarios (piece count exactly on target, catch-weight label decisively above nominal), one with the confirmation prompt on and one with it off. The merged fix resolves the whole-TU quantity through a backend lookup; if that ever returned the catch-UOM quantity instead of the stock-UOM piece count, the guard would fire on weight variance — which is explicitly out of scope for this feature. These scenarios make that regression visible.

## 4. Smaller items

- `PickingMobileApplication.resolvePickFromHU`'s `callerId` used the jetbrains `@NotNull` (no runtime check) where the class and the whole delegate chain use Lombok `@NonNull`.
- Comment volume in `PickingJobService`, `PickLineScanScreen` and `ScanHUAndGetQtyComponent` over the project's comment-volume rule. Comments carrying a non-obvious *why* were kept; redundant narration and development-history narration were removed.

## Explicitly out of scope

- Booking semantics — this hardens *how* the guard is reached, not what it books.
- `PickingJobService`'s two-hop `line.getScheduleId().getShipmentScheduleId()`: it matches two peer call sites, so the real fix is a one-hop accessor applied to all three.
- The `uiTrace` gap on the whole-TU rejection paths (they only `console.warn`). Pre-existing at both sites.

### DIFF
--- a/misc/services/mobile-webui/mobile-webui-frontend/src/__tests__/components/ScanHUAndGetQtyComponent.test.js
+++ b/misc/services/mobile-webui/mobile-webui-frontend/src/__tests__/components/ScanHUAndGetQtyComponent.test.js
@@ -205,11 +205,9 @@ const arrowFunctionBody = (name) => {
 
 const countOccurrences = (haystack, needle) => haystack.split(needle).length - 1;
 
-// The CU and the whole-TU path must stay behaviourally identical (REQUIREMENTS.md 31290 § 1), and the
-// mechanism that keeps them so is the single validateQtyAgainstMax both of them call. That is a
-// STRUCTURAL property, and it is deliberately asserted structurally: when the ceiling was last
-// duplicated (new_dawn_uat tip 171414313f5) the two copies were behaviour-identical, so an
-// outcome-comparing test was green throughout and would not have caught the re-duplication.
+// The CU and whole-TU paths must stay behaviourally identical, enforced structurally by the single
+// validateQtyAgainstMax both of them call. Hence the structural assertions: re-duplicating the ceiling
+// into two behaviour-identical copies keeps an outcome-only test green, while these fail.
 describe('ScanHUAndGetQtyComponent: the CU and whole-TU ceilings resolve through one helper', () => {
   it('reaches the ceiling from the CU path via validateQtyAgainstMax', () => {
     expect(arrowFunctionBody('validateQtyEntered')).toContain('validateQtyAgainstMax(');
--- a/misc/services/mobile-webui/mobile-webui-frontend/src/__tests__/components/ScanHUAndGetQtyComponent.test.js
+++ b/misc/services/mobile-webui/mobile-webui-frontend/src/__tests__/components/ScanHUAndGetQtyComponent.test.js
@@ -21,208 +21,89 @@ jest.mock('../../utils/toast', () => ({
   toastErrorFromObj: jest.fn(),
 }));
 
-const HU_SCANNER_INPUT = 'scanHUBarcode-input';
+// The operator scans a whole TU holding 3 pieces onto a line with 2 still to pick, so the component
+// raises the over-delivery confirmation, and answering "Ja" fires the pick.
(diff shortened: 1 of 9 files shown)


misc/services/mobile-webui/mobile-webui-frontend/src/__tests__/components/ScanHUAndGetQtyComponent.test.js
```diff
@@ -0,0 +1,230 @@
+import fs from 'fs';
```
teosarca, line 1: what ever!? i cannot read this crap. to me playwright e2e are the real tests!

teosarca: **Resolution (dev):** Agreed — the real coverage belongs in Playwright, and it is being added rather than argued about.

New E2E scenario going into `e2e/mobile-webui/tests/spec/picking/overPickingPrompt_wholeHU.spec.js`, driving the in-flight guard along the actual operator path: over-pick a whole HU, answer Yes on the prompt, then attempt a second submit while the first is still in flight — asserting the pick is booked exactly once (one `M_ShipmentSchedule_QtyPicked` row, correct qty) and no second dialog appears. Gated on a genuine RED first: the scenario must fail with the in-flight guard reverted before it counts.

On this Jest file, the two groups are being judged against that E2E:
- the in-flight guard tests become redundant once the E2E covers the same race deterministically;
- the structural parity tests are the only thing that can catch the ceiling being re-duplicated into two behaviour-identical copies — a behavioural test stays green through exactly that regression.

Neither group gets deleted without your explicit go. If you want the whole Jest file gone once the E2E is green, say so and it goes.

PR #25416, opened 2026-08-04
nShift COO: scope shipment QtyPicked split to shipment-line-config attributes

metas-ts:
## What

Fixes a `deep_tundra_release` / `new_dawn_uat` build red: the mobile E2E `e2e/mobile-webui/tests/spec/picking/overPickingPrompt_wholeHU.spec.js` fails deterministically — a catch-weight whole-HU pick books **two** `M_ShipmentSchedule_QtyPicked` rows where **one** is expected (`Expected qty picked records size to be <1> but was <2>`).

## Root cause

`HUShipmentScheduleBL.createCandidatesForQtyPicked` splits the picked-qty allocation into one `M_ShipmentSchedule_QtyPicked` row per distinct **`UseInASI` attribute fingerprint** (`computeUseInASIFingerprint`). That fingerprint used the **full** `UseInASI` attribute set — diverging from the downstream `M_InOutLine` aggregation in `ShipmentScheduleWithHU#computeAttributeValues`, which groups only by attributes **whitelisted in `M_ShipmentSchedule_AttributeConfig`** (via `ShipmentScheduleHandler#attributeShallBePartOfShipmentLine`).

`WeightNet` is `UseInASI` **and** a per-piece **measurement** (`LinearDistribution` splitter + `Sum` aggregation): when a whole HU of N catch-weight pieces is split, the total weight is distributed with rounding, so the pieces carry unequal values (e.g. `0.333 / 0.333 / 0.334`). Two distinct values → two fingerprint groups → two `QtyPicked` rows — even though the InOutLine aggregation correctly ignores weight and merges them back into a single line. The extra `QtyPicked` row backs no distinct shipment line.

Introduced by the nShift Country-of-Origin work (https://github.com/metasfresh/metasfresh/pull/24363), whose per-attribute expansion of `createCandidatesForQtyPicked` used the raw `UseInASI` flag instead of the shipment-line attribute config.

## Fix

Apply the **same** `attributeShallBePartOfShipmentLine` (`M_ShipmentSchedule_AttributeConfig`) filter to `computeUseInASIFingerprint` that the InOutLine aggregation already uses. The QtyPicked split now uses the **identical criterion** as the InOutLine aggregation:
- classification attributes (country of origin, lot, best-before, …) still split — as configured;
- measurement attributes (weight) no longer do.

Whether a customer wants per-attribute lines stays a config decision in `M_ShipmentSchedule_AttributeConfig` — the fix makes the QtyPicked layer honor that config, like the rest of the pipeline.

## Test

Covered by the existing E2E `e2e/mobile-webui/tests/spec/picking/overPickingPrompt_wholeHU.spec.js` — RED before this change (it asserts the `QtyPicked` row count that fails), GREEN after. Reproduced locally: the three picked VHUs carry `WeightNet` `0.333 / 0.333 / 0.334`, producing the two-row split the spec rejects.

## AD metadata

None — pure Java change, no `AD_*` migration or new field.

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
@@ -173,6 +173,7 @@ public class HUShipmentScheduleBL implements IHUShipmentScheduleBL
 	private final IHUPackingAwareBL huPackingAwareBL = Services.get(IHUPackingAwareBL.class);
 	private final IHUCapacityBL huCapacityBL = Services.get(IHUCapacityBL.class);
 	private final IAttributeDAO attributeDAO = Services.get(IAttributeDAO.class);
+	private final IShipmentScheduleHandlerBL shipmentScheduleHandlerBL = Services.get(IShipmentScheduleHandlerBL.class);
 
 	private static final String SYSCONFIG_ShipmentConsolidationPeriod = "de.metas.handlingunits.shipmentschedule.api.impl.HUShipmentScheduleBL.ShipmentConsolidationPeriod";
 	private static final String DEFAULT_ShipmentConsolidationPeriod = null;
@@ -1229,14 +1230,15 @@ public class HUShipmentScheduleBL implements IHUShipmentScheduleBL
 	/**
 	 * Predicate deciding whether an attribute splits the shipment line, i.e. whether it is whitelisted in
 	 * {@code M_ShipmentSchedule_AttributeConfig} for the schedule's handler — the same rule the M_InOutLine
-	 * aggregation applies. Resolved via {@code getHandlerForOrNull} because this runs eagerly while picking,
-	 * where the schedule may not have a resolvable handler yet; then it falls back to "no restriction" (the
-	 * split is refined at shipment generation, where the handler resolves).
+	 * aggregation applies. Resolved via {@code getHandlerForOrNull}: handlers are registered once at server
+	 * startup ({@code InOutCandidateValidator#onInit}), so during real picking the handler is present and the
+	 * whitelist restricts the split. It is null only in narrow init/test contexts that bypass that
+	 * registration; there we fall back to "no restriction" rather than fail.
 	 */
 	private Predicate<AttributeId> attributeSplitsShipmentLinePredicate(@NonNull final I_M_ShipmentSchedule_QtyPicked qtyPicked)
 	{
 		final de.metas.inoutcandidate.model.I_M_ShipmentSchedule shipmentSchedule = qtyPicked.getM_ShipmentSchedule();
-		final ShipmentScheduleHandler handler = Services.get(IShipmentScheduleHandlerBL.class).getHandlerForOrNull(shipmentSchedule);
+		final ShipmentScheduleHandler handler = shipmentScheduleHandlerBL.getHandlerForOrNull(shipmentSchedule);
 		if (handler == null)
 		{
 			return attributeId -> true;
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
@@ -99,7 +99,6 @@ import org.adempiere.util.lang.IContextAware;
 import org.adempiere.warehouse.LocatorId;
 import org.adempiere.warehouse.WarehouseId;
 import org.compiere.model.I_C_UOM;
-import org.compiere.model.I_M_Attribute;
 import org.compiere.model.X_C_DocType;
 import org.compiere.model.X_M_InOut;
 import org.slf4j.Logger;
@@ -116,7 +115,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -173,7 +171,6 @@ public class HUShipmentScheduleBL implements IHUShipmentScheduleBL
 	private final ISysConfigBL sysConfigBL = Services.get(ISysConfigBL.class);
 	private final IHUPackingAwareBL huPackingAwareBL = Services.get(IHUPackingAwareBL.class);
 	private final IHUCapacityBL huCapacityBL = Services.get(IHUCapacityBL.class);
-	private final IShipmentScheduleHandlerBL shipmentScheduleHandlerBL = Services.get(IShipmentScheduleHandlerBL.class);
 
 	private static final String SYSCONFIG_ShipmentConsolidationPeriod = "de.metas.handlingunits.shipmentschedule.api.impl.HUShipmentScheduleBL.ShipmentConsolidationPeriod";
 	private static final String DEFAULT_ShipmentConsolidationPeriod = null;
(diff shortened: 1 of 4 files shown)


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -1050,6 +1056,18 @@ public List<ShipmentScheduleWithHU> createCandidatesForQtyPicked(
 		final IHUStorageFactory storageFactory = huContext.getHUStorageFactory();
 		final IAttributeStorageFactory attrFactory = huContext.getHUAttributeStorageFactory();
 
+		// The picked-qty allocation may only be split by attributes that are whitelisted for the shipment
+		// line in M_ShipmentSchedule_AttributeConfig — the SAME criterion the downstream M_InOutLine
+		// aggregation applies (see ShipmentScheduleWithHU#computeAttributeValues). Without this filter every
+		// UseInASI attribute splits the allocation, including per-piece measurements such as catch-weight
+		// WeightNet whose values differ only by distribution rounding (e.g. 0.333/0.333/0.334) — producing
+		// extra M_ShipmentSchedule_QtyPicked rows that back no distinct shipment line.
```
metas-ts:
```suggestion
		// The picked-qty allocation may only be split by attributes that are whitelisted for the shipment
		// line in M_ShipmentSchedule_AttributeConfig — the SAME criterion the downstream M_InOutLine
		// aggregation applies (see ShipmentScheduleWithHU#computeAttributeValues).
```

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -1050,6 +1056,18 @@ public List<ShipmentScheduleWithHU> createCandidatesForQtyPicked(
...
+		final de.metas.inoutcandidate.model.I_M_ShipmentSchedule shipmentSchedule =
```
metas-ts: why not private?
why not static? then you could avoid passing it as parameter

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
metas-ts: **Resolution (dev):** Applied — comment trimmed to the criterion + the ShipmentScheduleWithHU#computeAttributeValues reference. https://github.com/metasfresh/metasfresh/commit/77ef858a0204ee9efc347ceeb150d30318f29f2e

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
metas-ts: **Resolution (dev):** Refactored to avoid threading the predicate: static computeUseInASIFingerprint / groupProductStoragesByFingerprint now receive the M_ShipmentSchedule and resolve the handler via Services.get(IShipmentScheduleHandlerBL.class).getHandlerFor(...) internally — no Predicate parameter, and the hoisted shipmentScheduleHandlerBL field is dropped. getHandlerFor is a cheap AD_Table_ID->singleton lookup, resolved once per VHU. https://github.com/metasfresh/metasfresh/commit/77ef858a0204ee9efc347ceeb150d30318f29f2e

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -1050,6 +1052,11 @@ public List<ShipmentScheduleWithHU> createCandidatesForQtyPicked(
 		final IHUStorageFactory storageFactory = huContext.getHUStorageFactory();
 		final IAttributeStorageFactory attrFactory = huContext.getHUAttributeStorageFactory();
 
+		// The picked-qty allocation may only be split by attributes that are whitelisted for the shipment
+		// line in M_ShipmentSchedule_AttributeConfig — the SAME criterion the downstream M_InOutLine
+		// aggregation applies (see ShipmentScheduleWithHU#computeAttributeValues).
+		final de.metas.inoutcandidate.model.I_M_ShipmentSchedule shipmentSchedule = qtyPicked.getM_ShipmentSchedule();
```
metas-ts: look for an alternative to working with model-classes. this is against our role for contraining them to DAO and reposititories

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
```diff
@@ -1185,29 +1192,36 @@ private static Stream<Map.Entry<AttributesKey, IHUProductStorage>> groupProductS
...
+			@NonNull final IAttributeStorageFactory factory,
+			@NonNull final de.metas.inoutcandidate.model.I_M_ShipmentSchedule shipmentSchedule)
 	{
 		// IAttributeStorage does not implement getAttributeValueIdOrNull, so we cannot use
 		// AttributesKeys.createAttributesKeyFromAttributeSet here — list-type attributes
 		// would be silently dropped. Instead, we build the key directly from IAttributeValue,
 		// using the value string for all attribute types (consistent for grouping purposes).
+		// getHandlerForOrNull (not getHandlerFor): this runs eagerly while picking, where a schedule may not
+		// have a resolvable handler yet — then fall back to no restriction (the split is refined at shipment
+		// generation, where the handler resolves and the M_ShipmentSchedule_AttributeConfig whitelist applies).
```
metas-ts: check if this whole comment block can be more shorter and concise

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
metas-ts: **Resolution (dev):** Reworked to keep model classes out of the fingerprint helpers. The split-relevance decision is resolved once into a value-typed `Predicate<AttributeId>` (attributeSplitsShipmentLinePredicate), and groupProductStoragesByFingerprint / computeUseInASIFingerprint now take that predicate instead of I_M_ShipmentSchedule. The one place a model is still needed — attributeShallBePartOfShipmentLine intrinsically requires the schedule (its onlyIfInReferencedRecordAsi branch) — is confined to that small builder, and the attribute record is fetched via IAttributeDAO (i.e. through the repository, per the rule). https://github.com/metasfresh/metasfresh/commit/11eef21dfed8441908476e158285ca8c96eb42b3

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/shipmentschedule/api/impl/HUShipmentScheduleBL.java
metas-ts: **Resolution (dev):** Shortened. The entry-point comment is now two lines; the getHandlerForOrNull rationale moved to the (shorter) javadoc of the new attributeSplitsShipmentLinePredicate helper, and the verbose block inside computeUseInASIFingerprint is gone. https://github.com/metasfresh/metasfresh/commit/11eef21dfed8441908476e158285ca8c96eb42b3

PR #25423, opened 2026-08-04
Invoice candidates — retry the model lock instead of dying on contention

adi-stefan:
## What

Invoice-candidate generation dies when two async workpackages reach the same model concurrently.
This opts the lock acquisition into the retry mechanism `SqlLockDatabase` already implements.

## The defect

`InvoiceCandidateHandlerBL.createForModel0` acquires the model lock like this:

```java
final ILock lock = lockManager.lock()
        .setOwner(lockOwner)
        .setRecordByModel(model)
        .setAutoCleanup(true)
        .setFailIfAlreadyLocked(true)   // no retry, no wait
        .acquire();
```

Three facts compose the failure:

1. **No retry.** `setFailIfAlreadyLocked(true)` with the default `retryOnFailure = 0` gives
   `maxTrials = 1` — a single attempt.
2. **Owners can never coincide.** `LockOwner.newOwner(prefix)` appends a `UUID.randomUUID()` suffix,
   so two concurrent invocations of the *same* handler hold *different* owners and cannot share a lock.
3. **The database rejects the second inserter.** `t_lock_reference_singleowner` is
   `UNIQUE (record_id, ad_table_id) WHERE isallowmultipleowners = 'N'`.

So the loser of the race cannot serialise, cannot share, and is rejected outright. It dies with

```
de.metas.lock.exceptions.LockFailedException:
    Record was already locked: TableRecordReference{tableName=C_Order, recordId=…}
  Caused by: DBUniqueConstraintException:
    duplicate key value violates unique constraint "t_lock_reference_singleowner"
  at InvoiceCandidateHandlerBL.createMissingCandidatesFor(InvoiceCandidateHandlerBL.java:152)
  at CreateMissingInvoiceCandidatesWorkpackageProcessor.processWorkPackage(…)
```

the workpackage fails, the invoice candidates are never created, and the caller eventually sees a
`400` from `PUT api/v2/orders/sales/candidates/process` — several minutes later, which is what makes
this look like a timeout rather than a lock problem.

Note the `.acquire()` call sits **outside** the surrounding `try/catch`, so the method's own error
logging never sees it either.

## The fix

One line: `.retryOnFailure(10)`.

Nothing new is introduced. `SqlLockDatabase` already implements the retry loop
(`maxTrials = 1 + getRetryOnFailure()`, `sleepBeforeRetry()` at 200 ms); it is simply inert unless a
caller opts in, and this caller never did. `ShipmentScheduleEnqueuer` is the existing precedent and
already uses `retryOnFailure(10)` — the value here matches it rather than inventing a new one.

The lock is held only for one generate-and-save, so contention is transient by construction; 10
attempts across 2 s is far longer than the window.

Retry rather than skip-on-contention: skipping would silently drop the work if the lock holder's
transaction later rolled back, and retry is this codebase's established answer to exactly this
situation.

## Evidence

Surfaced as an intermittent `test (cucumber) (profile3)` failure in
`automateOrderCandToOrderAndShipmentAndInvoice.feature`, where the failing scenario moved between
runs — the classic look of a flaky test.

It is not flaky. Reproduced locally, single machine, no CI involved: **6 failures out of 84
scenarios**, with **102** `t_lock_reference_singleowner` violations and **34** `LockFailedException`
in one run, across many distinct `C_Order` records. That is a higher failure rate than CI produces,
so the contention needs no special infrastructure — ordinary concurrency is enough.

The load-sensitivity explains the intermittency: collision probability rises with parallelism, so the
same commit can pass on a quiet runner and fail on a busy one.

## Scope

Only the main line is fixed here. Older branches that predate the `SqlLockDatabase` retry loop
(the retry mechanism is absent there entirely) need that mechanism ported before this one-liner can
help them.

### DIFF
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandidateHandlerBL.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/invoicecandidate/api/impl/InvoiceCandidateHandlerBL.java
@@ -318,11 +318,16 @@ public class InvoiceCandidateHandlerBL implements IInvoiceCandidateHandlerBL
 		for (final InvoiceCandidateGenerateRequest request : requests)
 		{
 			// Lock the "model" to make sure nobody else would generate invoice candidates for it.
+			// retryOnFailure: two workpackages can legitimately reach the same model concurrently, and the
+			// lock is held only for this one generate-and-save. Without retries the loser dies on the
+			// t_lock_reference_singleowner unique index (LockOwner.newOwner mints a fresh owner per call,
+			// so they can never share the lock), the workpackage fails, and the candidates are never created.
 			final ILock lock = lockManager.lock()
 					.setOwner(lockOwner)
 					.setRecordByModel(model)
 					.setAutoCleanup(true)
 					.setFailIfAlreadyLocked(true)
+					.retryOnFailure(10)
 					.acquire();
 
 			try (final ILockAutoCloseable ignored = lock.asAutoCloseable())


adi-stefan:
## ⚠️ Do not merge — local verification shows this change causes DB deadlocks

I reproduced the original lock race locally on `big_coconut_uat` (where it fails 6/84 reliably) and
then measured the identical `.retryOnFailure(10)` change against that reproduction. The retry does
remove the lock race — but it replaces it with something worse.

Both runs: `profile3` cucumber, same branch, same machine, run sequentially and alone.

| Metric | Without the change | With the change |
|---|---|---|
| Scenarios | 84 | 84 |
| **Failures** | **6** | **30** |
| `t_lock_reference_singleowner` violations | 102 | **0** |
| `LockFailedException` | 34 | **0** |
| `DBUniqueConstraintException` | 46 | 4 |
| **`DBDeadLockDetectedException`** | **0** | **22** |
| Elapsed | 2,547 s | 3,616 s |

### Mechanism

All 28 deadlock reports are on `c_invoice_candidate`:

```
Process 3130 waits for ShareLock on transaction 22845; blocked by process 3133.
Process 3133 waits for ShareLock on transaction 22837; blocked by process 3130.
  Where: while updating tuple (1,7) in relation "c_invoice_candidate"
```

The fail-fast acquisition this PR removes was doing real work: by aborting the loser immediately it
kept the "no hold-and-wait" condition satisfied. Making the loser *wait* lets two workpackages each
hold `c_invoice_candidate` row locks while waiting on the other, and PostgreSQL's deadlock detector
fires. The failures move from the original `automateOrderCand` scenarios to the whole
`invoicecandidate/packingmaterial/shipments.feature` file, which times out waiting for invoice
candidates that are never created.

### Why CI here is green

`profile3` on `new_dawn_uat` passes, but that is absence of evidence rather than evidence of
absence — the underlying race does not reproduce on this branch in the first place, so this PR's CI
cannot exercise the path where the change does harm.

Holding this until a design that doesn't reintroduce hold-and-wait is agreed. The two remaining
candidate directions from the original analysis are a stable/shared lock owner so concurrent
invocations coalesce, or de-duplicating enqueued workpackages per `C_Order` so two never target one
record — neither of which requires anyone to wait while holding rows.

adi-stefan:
Closing: the change in this PR fixes a non-problem and makes things worse.

The `C_Order` lock contention it targets is real but already **handled** — `CreateMissingInvoiceCandidatesWorkpackageProcessor` converts `LockFailedException` into the framework's `WorkpackageSkipRequestException`, and every batch in the reproduction shows `erroredWPSize: 0`. A diagnostic that cut the skip-retry window 20x (5000 -> 250 ms) moved the failure count only from 6 to 5, i.e. noise.

Meanwhile the retry introduced 22 `DBDeadLockDetectedException` on `c_invoice_candidate` (measured, clean run, same branch/machine): the fail-fast acquisition this PR removes was keeping Coffman's no-hold-and-wait condition satisfied.

The real root cause of the failing scenarios is in `AsyncBatchObserver`: overlapping `observeOn` calls on one `AsyncBatchId` overwrite the previous `BatchProgress` in a plain id-keyed map (orphaning the first waiter's `CompletableFuture` -> 300 s timeout -> REST 400), and `removeObserver`'s `unlockAll()` on the shared per-batch lock owner drops the monitoring lock a concurrent observer still needs (-> the `not monitored! Return empty list!` storm, notifications dropped).

That will be fixed separately. Full evidence in the earlier comment on this PR.

PR #25427, opened 2026-08-04
BPartner: auto-generate debtor/creditor numbers on V2 upsert

mrie04:
## What

Auto-generate `C_BPartner.DebtorId` (customers) / `CreditorId` (vendors) on business-partner creation, return them on the BPartner V2 upsert response, and accept explicit values (for data migration from an old system).

A guarded `C_BPartner` model interceptor (before-new / before-change) calls a core number-generation service that, per org config:
- draws the next value from a per-org `AD_Sequence` (`DocumentNo` mechanism), or
- delegates to an optional per-org override DB function, or
- is a no-op when unconfigured (core stays customer-agnostic — no config ships active).

Explicit values advance the target sequence instead of generating, so generated numbers stay above any migrated range. A per-org unique index is the backstop.

## Highlights

- `DebtorId`/`CreditorId` RepoIdAware value types (null-not-zero discipline).
- Dedicated `JsonResponseBPartnerUpsertItem` — `debtorId`/`creditorId` surface under `responseItems[n].responseBPartnerItem`, omitted when null.
- Optional explicit `debtorId`/`creditorId` on the upsert request.
- `BPartnerNumberSequenceDAO` (atomic draw/advance) + `BPartnerNumberGenerator` (3-branch selection; override function name validated against SQL injection, all args bound).
- `C_BPartner_NumberGen` interceptor; saver copy-back so a generated number reaches the response.
- AD migration: sysconfig keys (blank / per-org) + two deactivated default sequences at org 1000000; no `AD_Sequence`/`MSequence` schema change.
- Shared parameterized cucumber stepdefs + a core feature (TC1-TC10 + both-roles) — green end-to-end.


backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5817520_sys_BPartner_NumberGen.sql
```diff
@@ -0,0 +1,95 @@
...
+-- Sysconfig: name of the AD_Sequence to use for debtor numbers (blank = feature off)
+INSERT INTO AD_SysConfig (AD_Client_ID, AD_Org_ID, AD_SysConfig_ID, ConfigurationLevel,
+                          Created, CreatedBy, Updated, UpdatedBy,
+                          EntityType, IsActive, Name, Value, Description)
+SELECT 0, 0, 541843 /*From ID Server*/, 'O',
+       TO_TIMESTAMP('2026-08-04 10:00:01', 'YYYY-MM-DD HH24:MI:SS'), 100,
+       TO_TIMESTAMP('2026-08-04 10:00:01', 'YYYY-MM-DD HH24:MI:SS'), 100,
+       'D', 'Y',
+       'de.metas.bpartner.DebtorNoSequence',
+       '',
```
mrie04: why not default set and inactive?

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5817520_sys_BPartner_NumberGen.sql
```diff
@@ -0,0 +1,95 @@
...
+-- Sysconfig: name of the AD_Sequence to use for creditor numbers (blank = feature off)
+INSERT INTO AD_SysConfig (AD_Client_ID, AD_Org_ID, AD_SysConfig_ID, ConfigurationLevel,
+                          Created, CreatedBy, Updated, UpdatedBy,
+                          EntityType, IsActive, Name, Value, Description)
+SELECT 0, 0, 541844 /*From ID Server*/, 'O',
+       TO_TIMESTAMP('2026-08-04 10:00:02', 'YYYY-MM-DD HH24:MI:SS'), 100,
+       TO_TIMESTAMP('2026-08-04 10:00:02', 'YYYY-MM-DD HH24:MI:SS'), 100,
+       'D', 'Y',
+       'de.metas.bpartner.CreditorNoSequence',
+       '',
```
mrie04: why not default set and inactive?

misc/de-metas-common/de-metas-common-bpartner/src/main/java/de/metas/common/bpartner/v2/response/JsonResponseBPartnerUpsertItem.java
```diff
@@ -0,0 +1,122 @@
...
+	Integer debtorId;
+
+	@ApiModelProperty(value = "Creditor number persisted for this business partner (C_BPartner.CreditorId). Null when not set.")
+	@JsonProperty("creditorId")
+	@JsonInclude(Include.NON_NULL)
+	@Nullable
+	Integer creditorId;
+
+	@JsonCreator
+	public JsonResponseBPartnerUpsertItem(
```
mrie04, line 60: why not `@Builder` ?

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5817520_sys_BPartner_NumberGen.sql
mrie04: **Resolution (dev):** Adopted. `DebtorNoSequence` now ships `Value=556615` (the AD_Sequence_ID of `BPartner_DebtorNo_Default`) + `IsActive='N'`. SysConfigDAO's value lookup filters `WHERE IsActive='Y'`, so the row reads as "not configured" (feature off) until a customer flips `IsActive='Y'` — which activates the shipped default sequence with no id-lookup. https://github.com/metasfresh/metasfresh/commit/22df2456effc92d3b00aa19e22c477513b3b0db6

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5817520_sys_BPartner_NumberGen.sql
mrie04: **Resolution (dev):** Adopted. `CreditorNoSequence` now ships `Value=556616` (AD_Sequence_ID of `BPartner_CreditorNo_Default`) + `IsActive='N'`, same rationale as the debtor sysconfig: inactive default-set row = feature off (SysConfigDAO filters `IsActive='Y'`) until a customer activates it. https://github.com/metasfresh/metasfresh/commit/22df2456effc92d3b00aa19e22c477513b3b0db6

misc/de-metas-common/de-metas-common-bpartner/src/main/java/de/metas/common/bpartner/v2/response/JsonResponseBPartnerUpsertItem.java
mrie04: **Resolution (dev):** Adopted. Replaced the hand-written builder with `@Builder(builderMethodName = "bpartnerUpsertItemBuilder")` on the `@JsonCreator` constructor (−55 lines). The custom method name is required because a plain `builder()` would clash with the parent `JsonResponseUpsertItem`'s Lombok-generated `builder()` (different return type, same signature). Compiles + serialization test green. https://github.com/metasfresh/metasfresh/commit/22df2456effc92d3b00aa19e22c477513b3b0db6

backend/de.metas.business/src/main/java/de/metas/bpartner/service/BPartnerNumberGenerator.java
```diff
@@ -0,0 +1,155 @@
...
+	 * {@link Optional#empty()} when no number-generation is configured for this org.
+	 */
+	@NonNull
+	public Optional<Integer> generateNext(@NonNull final BPartnerNumberContext ctx)
+	{
+		final int adClientId = ctx.getClientId().getRepoId();
+		final int adOrgId = ctx.getOrgId().getRepoId();
+
+		// 1. Override branch
+		final String overrideFn = sysConfigBL.getValue(SYSCONFIG_OVERRIDE, (String)null, adClientId, adOrgId);
```
mrie04: why not getValue(SYSCONFIG_OVERRIDE, ClientAndOrgid.of(orgid, ClientId.METASFRESH) ? Same for others

backend/de.metas.business/src/main/java/de/metas/bpartner/service/BPartnerNumberGenerator.java
```diff
@@ -0,0 +1,155 @@
...
+	 * Validated as {@code [A-Za-z_][A-Za-z0-9_]*(\.[A-Za-z_][A-Za-z0-9_]*)?} before use.
+	 * Example: {@code public.fn_bpartner_no}.
+	 */
+	public static final String SYSCONFIG_OVERRIDE = "de.metas.bpartner.NumberResolverOverride";
+	/** Per-org sysconfig: {@code AD_Sequence_ID} (integer) for debtor numbers. */
+	public static final String SYSCONFIG_DEBTOR_SEQ = "de.metas.bpartner.DebtorNoSequence";
+	/** Per-org sysconfig: {@code AD_Sequence_ID} (integer) for creditor numbers. */
+	public static final String SYSCONFIG_CREDITOR_SEQ = "de.metas.bpartner.CreditorNoSequence";
+
+	@NonNull private final BPartnerNumberSequenceDAO dao;
```
mrie04: to general name dao

backend/de.metas.business/src/main/java/de/metas/bpartner/service/BPartnerNumberGenerator.java
mrie04: **Resolution (dev):** Adopted. Now uses the `ClientAndOrgId` overload (https://github.com/metasfresh/metasfresh/commit/413860dfa905dae72eda2a49acd693a17494bbbf) and, per your follow-up, the no-default `getValue(name, clientAndOrgId)` form (https://github.com/metasfresh/metasfresh/commit/bd27600e84d688c161a60bda1d34bebc05327d70) — dropped the redundant `(String)null`. (`getIntValue` keeps its `-1` default; there is no no-default overload for that signature.)

backend/de.metas.business/src/main/java/de/metas/bpartner/service/BPartnerNumberGenerator.java
mrie04: **Resolution (dev):** Partially adopted; confirming the rest. Renamed the class `BPartnerNumberSequenceDAO` -> `BPartnerNumberDAO` (https://github.com/metasfresh/metasfresh/commit/413860dfa905dae72eda2a49acd693a17494bbbf) since it is no longer sequence-only. Per your follow-up ("should be a service using existing daos/repos"): I will rework it into a service that draws via the existing `IDocumentNoBuilderFactory.forSequenceId(...)` instead of raw `UPDATE AD_Sequence`. Two points I flagged in chat need your call before I commit the reshape: (a) advance-past has no existing "advance a sequence past N" API — model-layer read/modify/save reuses the repo but loses the atomic single-statement guarantee (acceptable since it is migration-time/low-contention?); (b) the override `SELECT <fn>(...)` is inherently a raw DB-function call (no existing DAO for an arbitrary customer function) so it stays, housed in the service. Confirming, then implementing.

mrie04: **Resolution (dev):** Fully adopted. The DAO is now the service `BPartnerNumberService` (commit https://github.com/metasfresh/metasfresh/commit/4850d3ef1c9688c161a60bda1d34bebc05327d70), reworked to reuse framework sequence APIs instead of bespoke SQL:

- **draw** → `IDocumentNoBuilderFactory.forSequenceId(seqId).setClientId(clientId).build()` — the same atomic `UPDATE AD_Sequence … RETURNING` the whole system uses for document numbers (row-lock serialized, thread-safe);
- **advance-past** → a new atomic `IDocumentSequenceDAO.advanceCurrentNextPast(seqId, value)` in the core sequence DAO (`UPDATE AD_Sequence SET CurrentNext = GREATEST(CurrentNext, ?) …`, single statement, race-free) — no framework API for this existed, so it lives beside `DocumentNoBuilder`'s own atomic draw;
- **override** → kept as SQL (no framework API for an arbitrary customer function) but rebuilt via a `sqlParams`-list builder (the `PaymentAllocationRepository#buildSelectPaymentsToAllocateSql` pattern); name validated, args bound.

Behavior-neutral: `BPartnerNumberGeneratorTest` 13/13 + the cucumber feature 11/11 green locally; CI green (Merge Gate SUCCESS). code-reviewer pass CLEAN.

backend/de.metas.business/src/main/java/de/metas/bpartner/composite/BPartner.java
```diff
@@ -189,8 +189,10 @@ public class BPartner
 	private final String referrer;
 	@Nullable private final CampaignId campaignId;
 
-	private final Integer creditorId;
-	private final Integer debtorId;
+	// non-final so @Data generates a setter: the v2 BPartner REST persister sets it from the request's creditorId
+	private @Nullable Integer creditorId;
```
mrie04: `@Nullable private Creditorid creditorId`

misc/de-metas-common/de-metas-common-bpartner/src/main/java/de/metas/common/bpartner/v2/response/JsonResponseBPartnerUpsertItem.java
```diff
@@ -0,0 +1,71 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * BPartner-specific upsert response item.
+ * Extends {@link JsonResponseUpsertItem} with debtor/creditor numbers persisted for this business partner.
+ */
+@Value
+@EqualsAndHashCode(callSuper = true)
+public class JsonResponseBPartnerUpsertItem extends JsonResponseUpsertItem
```
mrie04, line 44: why this new one? we are already in bpartner package here

backend/de.metas.business/src/main/java/de/metas/bpartner/service/BPartnerNumberContext.java
```diff
@@ -0,0 +1,71 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Immutable context object passed to {@link BPartnerNumberGenerator}.
+ * Carries all per-request data needed to resolve the right sequence or override function
+ * for a given org and business-partner role.
+ */
+@Value
+@Builder
+public class BPartnerNumberContext
```
mrie04, line 39: why not some helper here? isDebtor, isCreditor, ofBPartner(I_C_BPartner)

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/bpartner/BPartnerNumberGen_StepDef.java
```diff
@@ -0,0 +1,649 @@
...
+
+	/**
+	 * Asserts that {@code responseItems[index].responseBPartnerItem.debtorId} is in the inclusive range [{@code from}..{@code to}].
+	 *
+	 * @param index zero-based index
+	 * @param from  inclusive lower bound
+	 * @param to    inclusive upper bound
+	 */
+	@Then("responseItems[{int}].responseBPartnerItem.debtorId is within {int}..{int}")
+	public void responseItems_responseBPartnerItem_debtorId_is_within(
```
mrie04: why we need a range version. if seq is defined right before we should have an exact expectation

backend/de.metas.business/src/main/java/de/metas/bpartner/composite/BPartner.java
mrie04: **Resolution (dev):** Adopted — you meant the type, not just the annotation order. Composite `BPartner.debtorId`/`creditorId` are now typed `DebtorId`/`CreditorId` (were `Integer`), with the ripples wrapped/unwrapped at the boundaries: `BPartnerCompositesLoader` uses `ofNullableNo` (replacing `NumberUtils.graterThanZeroOrNull`), `BPartnerCompositeSaver` uses `.toInt()` to the record and `ofNullableNo` on copy-back, and `JsonRetrieverService`/`JsonPersisterService` convert at the JSON edge. The JSON DTOs stay `Integer` (their module `de-metas-common-bpartner` can't depend on `de.metas.business` where the value types live) and `I_C_BPartner` stays generated `int`. https://github.com/metasfresh/metasfresh/commit/ca799b7058d2f6441e6fa9105917bca288006a05

misc/de-metas-common/de-metas-common-bpartner/src/main/java/de/metas/common/bpartner/v2/response/JsonResponseBPartnerUpsertItem.java
mrie04: **Resolution (dev):** Kept as a subclass, with reasoning — happy to change if you prefer. `responseBPartnerItem` was `JsonResponseUpsertItem`, which is *also* the element type of `responseLocationItems`/`responseContactItems`/`responseBankAccountItems`. Adding `debtorId`/`creditorId` to that shared type would surface them (as null) on location/contact/bank items too, so `JsonResponseBPartnerUpsertItem extends JsonResponseUpsertItem` scopes the two fields to the bpartner item only. Alternative if you'd rather not have the extra class: add the two fields to the bpartner-package `JsonResponseUpsertItem` and accept the null fields on the other item types — tell me which you prefer.

backend/de.metas.business/src/main/java/de/metas/bpartner/service/BPartnerNumberContext.java
mrie04: **Resolution (dev):** Adopted. `BPartnerNumberContext` now has `isDebtor()`/`isCreditor()` and a `ofBPartner(I_C_BPartner, Kind)` factory; `C_BPartner_NumberGen` uses `ofBPartner(bpartner, DEBTOR|CREDITOR)`, dropping the two hand-rolled builder blocks. (Kept it `(record, Kind)` rather than a single `ofBPartner(record)` because debtor vs creditor produce different contexts and `C_BPartner_ID` is null at `BEFORE_NEW`.) https://github.com/metasfresh/metasfresh/commit/ca799b7058d2f6441e6fa9105917bca288006a05

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/bpartner/BPartnerNumberGen_StepDef.java
mrie04: **Resolution (dev):** Adopted. The core feature now asserts exact values (e.g. `debtorId is 10000`, `next is 15001`) instead of ranges — each scenario seeds its own fresh `AD_Sequence` `StartNo` immediately before the upsert, so the first draw is deterministic. The `is within` step is kept only because the fm206 override feature legitimately uses a range (its resolver assigns within a company/non-company band). https://github.com/metasfresh/metasfresh/commit/ca799b7058d2f6441e6fa9105917bca288006a05

misc/de-metas-common/de-metas-common-bpartner/src/main/java/de/metas/common/bpartner/v2/response/JsonResponseBPartnerUpsertItem.java
mrie04: **Resolution (dev):** Confirmed — keeping **A** (the `JsonResponseBPartnerUpsertItem` subclass) per your call, so `debtorId`/`creditorId` stay scoped to the bpartner item and do not appear on the location/contact/bank items. No further change.

backend/de.metas.business/src/main/java/de/metas/bpartner/model/interceptor/C_BPartner_NumberGen.java
```diff
@@ -0,0 +1,105 @@
...
+ * in the base module (which handles other concerns), focusing solely on number generation.
+ */
+@Interceptor(I_C_BPartner.class)
+@Component
+@RequiredArgsConstructor
+public class C_BPartner_NumberGen
+{
+	@NonNull private final BPartnerNumberGenerator bpartnerNumberGenerator;
+
+	@ModelChange(timings = { ModelValidator.TYPE_BEFORE_NEW, ModelValidator.TYPE_BEFORE_CHANGE })
```
mrie04: needs to be after_new to get the id, but as you don't rly need ot for resolving it should stay before new

backend/de.metas.business/src/main/java/de/metas/bpartner/model/interceptor/C_BPartner_NumberGen.java
```diff
@@ -0,0 +1,105 @@
...
+	public void generateOrReserveNumbers(@NonNull final I_C_BPartner bpartner)
+	{
+		final boolean isNew = InterfaceWrapperHelper.isNew(bpartner);
```
mrie04: where is the early exit if feature is turned off. default off

backend/de.metas.business/src/main/java/de/metas/bpartner/service/BPartnerNumberGenerator.java
```diff
@@ -0,0 +1,154 @@
...
+		final String overrideFn = sysConfigBL.getValue(SYSCONFIG_OVERRIDE, clientAndOrgId);
+		if (!Check.isBlank(overrideFn))
+		{
+			numberService.callOverrideFunction(overrideFn, ctx, explicitValue);
+			return;
+		}
+
+		// 2. Sequence branch
+		final int seqId = sysConfigBL.getIntValue(seqSysconfigName(ctx.getKind()), -1, clientAndOrgId);
+		if (seqId > 0)
```
mrie04: repoidaware not null check instead

backend/de.metas.business/src/main/java/de/metas/bpartner/service/BPartnerNumberService.java
```diff
@@ -0,0 +1,194 @@
...
+		}
+		if (!FUNCTION_NAME_PATTERN.matcher(trimmed).matches())
+		{
+			throw new IllegalArgumentException(
+					"Override function name is not a valid SQL identifier (must match [A-Za-z_][A-Za-z0-9_]*(.[A-Za-z_][A-Za-z0-9_]*)? ): " + trimmed);
+		}
+
+		// bPartnerId may be null at TYPE_BEFORE_NEW (native-sequence mode: ID not yet assigned).
+		// Pass 0 in that case — the override function must treat 0 as "not yet known".
+		final BPartnerId bPartnerId = ctx.getBPartnerId();
```
mrie04: why we have it here at all then, new is the normal case

backend/de.metas.business/src/main/java/de/metas/bpartner/model/interceptor/C_BPartner_NumberGen.java
mrie04: **Resolution (dev):** Corrected — generation runs at `TYPE_BEFORE_NEW`, not `AFTER_NEW`. Cucumber showed `AFTER_NEW` is infeasible here: re-saving the record inside its own `AFTER_NEW` event throws `"Object is already involved in a model change event … AFTER_NEW, BEFORE_NEW"` (the PO keeps its `createNew` flag set until `saveFinish`). And the assigned id is not actually needed — `p_c_bpartner_id` is declared-but-unused by the override resolver (resolution is by org / kind / company-flag). At `BEFORE_NEW` the numbers are written by the INSERT (no re-save); `C_BPartner_ID` is `0`, passed as `0` to the override. All 12 cucumber scenarios pass.

backend/de.metas.business/src/main/java/de/metas/bpartner/model/interceptor/C_BPartner_NumberGen.java
mrie04: **Resolution (dev):** Adopting — adding a master enable flag (sysconfig, default `N`) shipped by the core migration, checked first in the interceptor so it early-exits with no per-save sysconfig-branch lookups on the (vast majority of) instances that don't use the feature. Feature ships OFF.

backend/de.metas.business/src/main/java/de/metas/bpartner/service/BPartnerNumberGenerator.java
mrie04: **Resolution (dev):** Adopting — replacing the raw `int seqId > 0` with the RepoIdAware null-check `DocSequenceId.ofRepoIdOrNull(sysConfigBL.getIntValue(...)) != null`, in both `generateNext` and `reserveExplicit`.

backend/de.metas.business/src/main/java/de/metas/bpartner/service/BPartnerNumberService.java
mrie04: **Resolution (dev):** Corrected — `bPartnerId` is kept **nullable** (no `assumeNotNull`). This follows from the `BEFORE_NEW` timing (see the interceptor thread): at `BEFORE_NEW` the native sequence has not assigned `C_BPartner_ID` yet (`0`/null), so the override receives `0` for `p_c_bpartner_id`, which it does not use.

backend/de.metas.business.rest-api-impl/src/test/java/de/metas/rest_api/v2/bpartner/bpartnercomposite/JsonPersisterServiceDebtorCreditorTest.java
```diff
@@ -0,0 +1,206 @@
...
+				new JsonRequestConsolidateService());
+
+		final I_C_BP_Group bpGroupRecord = newInstance(I_C_BP_Group.class);
+		bpGroupRecord.setC_BP_Group_ID(C_BP_GROUP_ID);
+		bpGroupRecord.setName(BP_GROUP_RECORD_NAME);
+		bpGroupRecord.setValue(BP_GROUP_RECORD_NAME);
+		saveRecord(bpGroupRecord);
+
+		// Create the org record + OrgInfo that JsonPersisterService requires
+		final I_AD_Org org = newInstance(I_AD_Org.class);
```
mrie04, line 107: isn't this part of adempiere init already?

backend/de.metas.business/src/main/java/de/metas/bpartner/model/interceptor/C_BPartner_NumberGen.java
```diff
@@ -0,0 +1,147 @@
...
+		final BPartnerNumbers numbers = bpartnerNumberGenerator.generateNumbers(bpartner);
+
+		for (final BPartnerNumberContext.Kind kind : BPartnerNumberContext.Kind.values())
+		{
+			// unwrap to the raw int only here, at the model-column boundary; BEFORE_NEW ⇒ written by the INSERT
+			numbers.getNo(kind).ifPresent(no -> applyNo(bpartner, kind, no));
+		}
+	}
+
+	private static void applyNo(@NonNull final I_C_BPartner bpartner, @NonNull final BPartnerNumberContext.Kind kind, final int no)
```
mrie04: why this isn't in BPartnerNumbers

backend/de.metas.business/src/main/java/de/metas/bpartner/service/BPartnerNumberService.java
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	/**
+	 * Advances the sequence so that {@code value} will not be re-issued by a later {@link #drawNext}.
+	 * Delegates to the atomic {@link IDocumentSequenceDAO#advanceCurrentNextPast(DocSequenceId, int)}.
+	 *
+	 * @param seqId the sequence to advance
+	 * @param value advance the sequence past this value
+	 */
+	public void advancePast(@NonNull final DocSequenceId seqId, final int value)
```
mrie04, line 119: Did you consider de.metas.util.lang.SeqNo instead of int?

backend/de.metas.business/src/main/java/de/metas/bpartner/service/BPartnerNumberService.java
```diff
@@ -0,0 +1,206 @@
...
+		{
+			// build() returns NO_DOCUMENTNO (null) WITHOUT throwing when the sequence has IsAutoSequence='N'
+			// (setFailOnError only rethrows the exception paths). A debtor/creditor sequence must be an auto-sequence.
+			throw new IllegalStateException("AD_Sequence_ID=" + seqId.getRepoId()
+					+ " returned no number; a debtor/creditor sequence must have IsAutoSequence='Y'");
+		}
+		try
+		{
+			// Debtor/creditor sequences are plain numeric (no prefix/suffix/decimal pattern), so build() yields a bare integer string.
+			return Integer.parseInt(documentNo.trim());
```
mrie04: de.metas.common.util.NumberUtils#asInt

backend/de.metas.business/src/main/java/de/metas/bpartner/model/interceptor/C_BPartner_NumberGen.java
mrie04: **Resolution (dev):** Moved — `BPartnerNumbers` now owns applying itself to the record via `applyTo(I_C_BPartner)` (each typed value unwrapped to its `int` column at the model boundary; a role with no generated number is left untouched). The interceptor is now a one-liner — `bpartnerNumberGenerator.generateNumbers(bpartner).applyTo(bpartner)` — and the `applyNo` kind-switch is gone. (commit 058e43cf4ad)

backend/de.metas.business.rest-api-impl/src/test/java/de/metas/rest_api/v2/bpartner/bpartnercomposite/JsonPersisterServiceDebtorCreditorTest.java
mrie04: **Resolution (dev):** No — `AdempiereTestHelper.init()` creates only org `0` (`OrgId.ANY`) and its OrgInfo; this test runs under `AD_ORG_ID = 10` (the `BPartnerRecordsUtil` convention), which init does not create. Verified empirically: removing the block makes all 3 tests error (the only OrgInfo present is org 0's). The sibling composite tests obtain org 10 via `BPartnerRecordsUtil.createBPartnerData(0)`, but that also seeds users/bpartners/contacts/locations this create-focused test doesn't need, so I kept the minimal inline org + OrgInfo.

backend/de.metas.business/src/main/java/de/metas/bpartner/service/BPartnerNumberService.java
mrie04: **Resolution (dev):** Adopted — `drawNext` now parses via `NumberUtils.asInt(documentNo)`, replacing the hand-rolled `Integer.parseInt` + try/catch. `asInt` trims internally (`StringUtils.trimBlankToNull`) and throws a clear "Cannot convert `<value>`" error naming the offending string, so both the descriptive wrapper and the `.trim()` are no longer needed. (commit d0ce4136b67)

backend/de.metas.business/src/main/java/de/metas/bpartner/service/BPartnerNumberService.java
mrie04: **Resolution (dev):** Considered, but `SeqNo` doesn't fit here. `de.metas.util.lang.SeqNo` models a document-line ordinal with a fixed STEP of 10 (`first()=10`, `next()` rounds up to the next multiple of 10) — that is line positioning, not a business account number drawn with the sequence's own `IncrementNo`. The produced number is already given a domain type (`DebtorId`/`CreditorId`); the `int` at this boundary is just the raw value the DB function returns before it is wrapped into that type. So `int` is the right transport here, with `DebtorId`/`CreditorId` carrying the typing.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/bpartner/BPartnerNumberGen_StepDef.java
```diff
@@ -0,0 +1,607 @@
...
+ * Given a debtor sequence for org "001" starting at 10000
+ * When I upsert a "non-company" "customer" "TC1-cust"
+ * Then responseItems[0].responseBPartnerItem.debtorId is 10000
+ * </pre>
+ */
+@RequiredArgsConstructor
+public class BPartnerNumberGen_StepDef
+{
+	/** Value column of the default org in the standard seed DB. */
+	private static final String DEFAULT_ORG_VALUE = "001";
```
mrie04: why not add it here? de.metas.cucumber.stepdefs.StepDefConstants

mrie04: **Resolution (dev):** Adopted — moved `"001"` to `StepDefConstants.ORG_VALUE` (companion to the existing `StepDefConstants.ORG_ID = OrgId.MAIN`, which is repoId 1000000 = seed org 001). The step def references it instead of a local `DEFAULT_ORG_VALUE`. (commit 5d375e6e803)

mrie04:
known flaky, merging

mrie04:

PR #25431, opened 2026-08-05
Async: release the workpackage lock when a queue processor cannot take it

adi-stefan:
## Problem

`AbstractQueueProcessor.processLockedWorkPackage` bails out early when the queue processor has no
permits left (`isAvailableToWork()` is false), but it left `LockedAt` set on the workpackage.

`QueueDAO`'s polling query only reconsiders a locked workpackage once `LockedAt` is older than
**10 minutes**, while a caller waiting on that workpackage's async batch gives up after the
**5-minute** `de.metas.async.AsyncBatchObserver.WaitTimeOutMS`. A dropped workpackage therefore
could never be re-polled before the waiter timed out.

The log line on that path even claimed `WorkpackageProcessorTask` would clear the lock — it cannot,
because the task is only constructed inside `processWorkPackage(..)`, which the bail-out never reaches.

## Evidence

`automateOrderCandToOrderAndShipmentAndInvoice.feature` (`@topic:orderCandidate`), before the fix:
**7 scenarios, 5 failed** — `PUT api/v2/orders/sales/candidates/process` returning 400 after a
~300 s stall. Instrumenting the timeout path showed all 8 timeout events had **exactly one**
outstanding workpackage — `processed=false error=false readyForProcessing=true skippedAt=null` —
and each of those workpackage ids appears exactly once in the
`processLockedWorkPackage was not successful` log: offered to its processor once, dropped, never retried.

After the fix, same filter: **7 scenarios, 0 failed**, and wall clock 30:25 → 6:12 min (the four
300 s stalls are gone). The drop itself still occurs 4 times — it is a legitimate outcome of the
check-then-dispatch race — but each dropped workpackage is now released and re-polled.

## Changes

- `AbstractQueueProcessor.processLockedWorkPackage` — clear `LockedAt` via
  `WorkPackageLockHelper.unlockNoFail(..)` in the unsuccessful branch, as every sibling path already
  does (`QueueProcessorPlanner` for invalid / no-processor, `processWorkPackage` for submit failure,
  `WorkpackageProcessorTask` for processed / skipped / errored).
- `AsyncBatchService.isWaitingForAsyncBatchAllowed()` — new detector, and `executeBatch` now WARNs
  when it is called from a thread that is itself processing a workpackage. That pattern holds a
  queue-processor thread for up to `WaitTimeOutMS` and is how a processor runs out of permits in the
  first place. **Deliberately not an exception**: callers doing exactly this exist today
  (`AutoProcessingOLCandService`'s order / shipment / invoice steps all run inside a `ProcessOLCands`
  workpackage), so turning it into a throw needs those call sites reworked first.
- `AsyncBatchObserver.waitToBeProcessed` — attach `AsyncBatchId` to the timeout rethrow, matching the
  sibling `catch (Exception)` branch, so a stuck batch is identifiable from the exception alone.

## Tests

- `AbstractQueueProcessor_UnlockOnNotProcessed_Test` — a planner-locked workpackage handed to a
  processor with no permits must come back unlocked. Fails on the old code (`LockedAt` stayed set).
- `AsyncBatchServiceTest` — the detector predicate, on and off a workpackage-processing thread.
- `de.metas.async` suite: 56 tests, 0 failures.


backend/de.metas.async/src/main/java/de/metas/async/service/AsyncBatchService.java
```diff
@@ -114,6 +133,19 @@ public void checkProcessed(@NonNull final AsyncBatchId asyncBatchId, @Nullable f
...
+		{
+			// dev-note: not thrown, because callers doing exactly this exist today (AutoProcessingOLCandService's
+			// order/shipment/invoice steps all run inside a ProcessOLCands workpackage). Deliberately WARN and not
+			// DEBUG: this pattern holds a queue processor's thread for up to WaitTimeOutMS and is how a processor
+			// runs out of permits in the first place - it must stay visible until the call sites are reworked.
+			Loggables.withLogger(logger, Level.WARN).addLog(
+					"*** executeBatch: waiting for C_Async_Batch_ID: {} from a thread that is itself processing workpackage-batch {}."
+							+ " This blocks a queue-processor thread until the batch completes or times out.",
+					asyncBatchId.getRepoId(),
+					Services.get(IWorkpackageProcessorContextFactory.class).getThreadInheritedWorkpackageAsyncBatch());
```
adi-stefan: IWorkpackageProcessorContextFactory as field

backend/de.metas.async/src/test/java/de/metas/async/processor/impl/AbstractQueueProcessor_UnlockOnNotProcessed_Test.java
```diff
@@ -0,0 +1,95 @@
...
+public class AbstractQueueProcessor_UnlockOnNotProcessed_Test extends QueueProcessorTestBase
+{
+	@Test
+	public void givenProcessorHasNoPermitsLeft_whenProcessLockedWorkPackage_thenLockedAtIsCleared()
+	{
+		// given
+		final I_C_Queue_Processor processorDef = helper.createQueueProcessor("test", 1, 1000);
+		helper.assignPackageProcessor(processorDef, StaticMockedWorkpackageProcessor.class);
+
+		final IWorkPackageQueue queue = Services.get(IWorkPackageQueueFactory.class)
```
adi-stefan: IWorkPackageQueueFactory as field

backend/de.metas.async/src/main/java/de/metas/async/service/AsyncBatchService.java
adi-stefan: **Resolution (dev):** Applied in https://github.com/metasfresh/metasfresh/commit/22f67a0f8cb6af25caaf85ea5c24541ff1930dcd

Added `private final IWorkpackageProcessorContextFactory workpackageProcessorContextFactory = Services.get(IWorkpackageProcessorContextFactory.class);` as an instance field.

One consequence worth calling out: `isWaitingForAsyncBatchAllowed()` was `static`, so it could not read an instance field — and `docs/coding-rules/service-injection.md` §2 forbids `static` service fields outright. So the field necessarily replaces the static predicate: the check is now inlined in `executeBatch` against the field, and the static method is gone. Side benefit — the WARN now logs the enclosing batch's repo id instead of the `AsyncBatchId` object's `toString()`.

Injection mechanism verified against the implementation rather than the interface, per that rule: `WorkpackageProcessorContextFactory` carries no Spring component annotation and `IWorkpackageProcessorContextFactory extends ISingletonService`, so `Services.get()` is correct here (not constructor injection).

This also removed `AsyncBatchServiceTest`, which covered only that static predicate. The fix itself stays covered by `AbstractQueueProcessor_UnlockOnNotProcessed_Test`. Happy to add a test against the context factory instead if you would rather keep one at that level.

backend/de.metas.async/src/test/java/de/metas/async/processor/impl/AbstractQueueProcessor_UnlockOnNotProcessed_Test.java
adi-stefan: **Resolution (dev):** Applied in https://github.com/metasfresh/metasfresh/commit/22f67a0f8cb6af25caaf85ea5c24541ff1930dcd

`IWorkPackageQueueFactory` is now a `private final` instance field on the test class and the inline `Services.get(..)` at the use site is gone.

I checked the one risk empirically rather than assuming it: a field initializer runs before `@Before AdempiereTestHelper.get().init()`, so resolving the service at field-init time could have been an ordering problem. It is not — the test still passes.

PR #25432, opened 2026-08-05
gh31289: recompute batch skips a handler-less shipment schedule instead of aborting

metas-ts:
## What

A cucumber seed fixture (`M_ShipmentSchedule_StepDef.seedShipmentSchedulesWithUntaggedRecomputeMarker`, from https://github.com/metasfresh/metasfresh/pull/25200) raw-INSERTs `M_ShipmentSchedule` rows with a **self-referential `AD_Table_ID`** (= the `M_ShipmentSchedule` table itself) plus an untagged recompute marker. The always-on `UpdateInvalidShipmentSchedulesWorkpackageProcessor` sweeps those markers DB-wide in one-transaction batches; the handler-less row makes `ShipmentScheduleHandlerBL.getHandlerFor` throw, which **aborts the whole batch** — so no schedule in the recompute queue gets recomputed. Symptom: profile6/profile7 picking scenarios time out on `M_ShipmentSchedules are found: | IsToRecompute | N |` (their schedule's recompute never settles); orphan rows + `AD_Issue`s were also observed accumulating on the base branch.

## Fix — two parts

1. **Product robustness — `ShipmentSchedulePA.createOlAndScheds`.** Resolve each schedule's handler via `getHandlerForOrNull`; a handler-less schedule is logged + recorded as an `AD_Issue` and **skipped** instead of aborting the batch. One bad row (test poison, or genuine data corruption / a schedule whose source-module handler is not loaded) can no longer jam the whole recompute queue.
2. **Fixture hermeticity — `M_ShipmentSchedule_StepDef`.** Track the seeded IDs; an `@After` deletes their recompute markers (any tag state) then the schedule rows, so the poison cannot leak into another scenario's recompute sweep. The https://github.com/metasfresh/metasfresh/pull/25200 direct-seed design (deliberately not using the order pipeline, to keep the tag-count assertions deterministic) is otherwise untouched.

## Test

New JUnit `ShipmentSchedulePA_createOlAndScheds_handlerlessSchedule_Test`: a handler-less schedule + a resolvable schedule in the same batch. **RED** before (throws `No shipment schedule handler defined for M_ShipmentSchedule`, aborting the batch); **GREEN** after (handler-less skipped, resolvable still processed). Run locally (JDK 8) — 1 run, 0 failures.

## Why it was intermittent (not branch-specific)

https://github.com/metasfresh/metasfresh/pull/25200 is an ancestor of **both** this base branch and `new_dawn_uat` (byte-identical fixture). It only fails when CI's per-run **profile partitioning** places the batching feature in the same profile as picking scenarios (shared JVM/DB) — which is why a green run and a red run can share the same code.

### DIFF
--- a/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentSchedulePA.java
+++ b/backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentSchedulePA.java
@@ -9,10 +9,13 @@ import de.metas.cache.CacheMgt;
 import de.metas.cache.model.CacheInvalidateMultiRequest;
 import de.metas.inout.ShipmentScheduleId;
 import de.metas.inout.model.I_M_InOutLine;
+import de.metas.error.IErrorManager;
 import de.metas.inoutcandidate.api.IShipmentScheduleAllocDAO;
+import de.metas.inoutcandidate.api.IShipmentScheduleHandlerBL;
 import de.metas.inoutcandidate.api.IShipmentSchedulePA;
 import de.metas.inoutcandidate.api.OlAndSched;
 import de.metas.inoutcandidate.exportaudit.APIExportStatus;
+import com.google.common.annotations.VisibleForTesting;
 import de.metas.inoutcandidate.invalidation.IShipmentScheduleInvalidateRepository;
 import de.metas.inoutcandidate.model.I_M_ShipmentSchedule;
 import de.metas.interfaces.I_C_OrderLine;
@@ -301,7 +304,8 @@ public class ShipmentSchedulePA implements IShipmentSchedulePA
 		return OrderAndLineId.ofRepoIdsOrNull(shipmentSchedule.getC_Order_ID(), shipmentSchedule.getC_OrderLine_ID());
 	}
 
-	private List<OlAndSched> createOlAndScheds(final List<I_M_ShipmentSchedule> shipmentSchedules)
+	@VisibleForTesting
+	List<OlAndSched> createOlAndScheds(final List<I_M_ShipmentSchedule> shipmentSchedules)
 	{
 		final IOrderDAO orderDAO = Services.get(IOrderDAO.class);
 
@@ -318,8 +322,20 @@ public class ShipmentSchedulePA implements IShipmentSchedulePA
 
 		final List<OlAndSched> result = new ArrayList<>();
 
+		final IShipmentScheduleHandlerBL shipmentScheduleHandlerBL = Services.get(IShipmentScheduleHandlerBL.class);
+
 		for (final I_M_ShipmentSchedule schedule : shipmentSchedules)
 		{
+			// gh31289: a schedule whose AD_Table_ID has no registered handler cannot be recomputed. Skip it
+			// (surfaced as an AD_Issue) instead of letting OlAndSched -> getHandlerFor throw and abort the
+			// WHOLE batch -- one such row (a #25200 self-referential test-seed row, or genuine data
+			// corruption) would otherwise jam every other schedule in the recompute queue.
+			if (shipmentScheduleHandlerBL.getHandlerForOrNull(schedule) == null)
+			{
+				reportHandlerlessScheduleAndSkip(schedule);
+				continue;
+			}
+
 			final OrderAndLineId orderLineId = extractOrderAndLineId(schedule);
 			final I_C_OrderLine orderLine;
 			final I_C_Order order;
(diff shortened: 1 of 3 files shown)


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentSchedulePA.java
```diff
@@ -344,6 +360,32 @@ private List<OlAndSched> createOlAndScheds(final List<I_M_ShipmentSchedule> ship
...
+				"Skipping M_ShipmentSchedule with no registered ShipmentScheduleHandler for its AD_Table_ID"
+						+ " (it cannot be recomputed): " + schedule);
+		logger.warn(issue.getLocalizedMessage(), issue);
+
+		// A handler-less schedule cannot be recomputed; since the pass no longer aborts on it, raise the
+		// AD_Issue explicitly to keep it visible. Best-effort: creating the issue must never itself break
+		// the recompute pass.
+		try
+		{
+			Services.get(IErrorManager.class).createIssue(issue);
```
metas-ts: Also here - Services generally shall be declared as fields.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentSchedulePA.java
```diff
@@ -318,8 +322,20 @@ private List<OlAndSched> createOlAndScheds(final List<I_M_ShipmentSchedule> ship
 
 		final List<OlAndSched> result = new ArrayList<>();
 
+		final IShipmentScheduleHandlerBL shipmentScheduleHandlerBL = Services.get(IShipmentScheduleHandlerBL.class);
```
metas-ts: Also here - Services generally shall be declared as fields. 
Unless there is a good reason (i can't see one here)
Please create a drift report for this behavior.

metas-ts: **Resolution (dev):** Done — declared as a field (`shipmentScheduleHandlerBL`) in https://github.com/metasfresh/metasfresh/commit/05ca9c665c3da9f6aeee1a55ac4e2dbfbc5a8179.

There is no good reason to deviate, and no drift report is needed. I had assumed keeping `IErrorManager` (below) as a local was necessary — a field initializer runs `Services.get(IErrorManager.class)` at construction, which I thought would break the POJO unit test that doesn't register an error manager. I verified that assumption and it's false: the unit test passes with it as a field. So both `Services` lookups are now fields.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentSchedulePA.java
metas-ts: **Resolution (dev):** Done — this is now the `errorManager` field, in the same commit https://github.com/metasfresh/metasfresh/commit/05ca9c665c3da9f6aeee1a55ac4e2dbfbc5a8179. Verified the POJO unit test still passes with it field-initialized.

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/api/impl/ShipmentSchedulePA.java
```diff
@@ -344,6 +360,32 @@ private List<OlAndSched> createOlAndScheds(final List<I_M_ShipmentSchedule> ship
...
+	 * {@link de.metas.inoutcandidate.spi.ShipmentScheduleHandler}, so it stays visible while the recompute
+	 * batch continues. Its {@code M_ShipmentSchedule_Recompute} marker is cleared with the rest of the batch
+	 * (tagged with this pass's selection) at the end of the pass, so it is not re-swept forever.
+	 */
+	private void reportHandlerlessScheduleAndSkip(@NonNull final I_M_ShipmentSchedule schedule)
+	{
+		final AdempiereException issue = new AdempiereException(
+				"Skipping M_ShipmentSchedule with no registered ShipmentScheduleHandler for its AD_Table_ID"
+						+ " (it cannot be recomputed): " + schedule);
+		logger.warn(issue.getLocalizedMessage(), issue);
```
metas-ts, line 374: this already creates an AD_Issue, so now this code creates two issues

metas-ts: **Resolution (dev):** Not fixed — flagging that plainly rather than closing the thread.

The observation stands: `reportHandlerlessScheduleAndSkip` constructs an `AdempiereException` and then also calls `errorManager.createIssue(issue)`, so the handler-less schedule is reported twice.

It is not fixed here because there is nowhere in this change to fix it: the PR is merged, and the issue it belongs to is closed. A follow-up needs either a new issue or a small direct PR against the base branch, and that is a scope decision for the issue owner rather than something to settle inside a resolved thread.

Surfaced again while working an unrelated issue, because the unanswered-comment check is repo-wide. Recorded so it is not lost.

PR #25435, opened 2026-08-05
QR-bill: emit structured (type S) addresses in invoice & dunning QR codes

riadmen:
Swiss ISO-20022 2019 mandate: payment files with combined (type K) addresses are rejected on submission; the creditor and debtor must be encoded as structured (type S) with separate street name, building number, postal code and town.

Both Swiss QR functions still emitted type K with street packed into one line and postal+town concatenated into another (structured fields left empty), so the recipient's town was never present in structured form.

  - Docs_Sales_Invoice_QR_Code / Docs_Sales_Dunning_QR_Code: emit 'S', split the street via extract_street_and_number, and place postal and town in their own fields. Per-party field count is unchanged (7 lines), so the overall SPC payload layout is preserved.
  - Add the extract_street_and_number helper (identical to core, for a conflict-free merge-up).
  - Invoice debtor name: was only the salutation (first line of bpartneraddress, e.g. "Herr"); now the partner's real name (company name, or first + last name), mirroring Docs_Sales_Dunning_QR_Code.
  - Existing invoice-function fixes (IBAN fallback, amount mask, QRR/SCOR reference) are left untouched.

### DIFF
--- a/backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/extract_street_and_number.sql
+++ b/backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/extract_street_and_number.sql
@@ -1,0 +1,50 @@
+/*
+ * #%L
+ * de.metas.fresh.base
+ * %%
+ * Copyright (C) 2025 metas GmbH
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as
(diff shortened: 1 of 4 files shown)


riadmen:
Superseded by #25437 (re-targeted to `inner_silence_uat`). This is a feature for the ISO-20022 2019 mandate, not an urgent production hotfix, so it belongs on the UAT line, not hotfix. Nothing was merged to `inner_silence_hotfix`.

PR #25438, opened 2026-08-06
Make `TotalAmount`, `Weight` and `DeliveryRule` invoice report elements hideable per `C_DocType`

riadmen:
Extend the ReportElements reference (541946) with TotalAmount, Weight and DeliveryRule, so these elements of the sales invoice report can be hidden per document type via C_DocType_ReportElement (same mechanism as Delivery_Via_Rule).                                                                   
                                                                                                                                                     
  - Docs_Sales_Invoice_Details: expose Is_TotalAmount_Hidden and Is_Weight_Hidden columns from report.IsHiddenReportElement                                                                                                        
  - Docs_Invoice_Details_Footer: NULL out deliveryrule when hidden                                                                                   
  - report_details_v2.jrxml: printWhenExpression to hide the total-amount footer band and the weight/catchweight fields accordingly


cristinamghita (approved):
lgtm

PR #25463, opened 2026-08-10
Cucumber: quiesce the shipment-schedule recompute queue before the batching fixture seeds

metas-ts:
## Problem

`updateInvalidShipmentSchedulesRecomputeBatching.feature` is flaky on `test (cucumber) (profile7)`.

Its scenarios assert on **global** counts in `M_ShipmentSchedule_Recompute`, while a real
`UpdateInvalidShipmentSchedules` recompute pass — enqueued on commit by some earlier scenario sharing the
executor's DB, and picked up by the 500 ms background queue planner — can still be in flight. That pass
claims markers through the same `M_ShipmentSchedule_TagToRecompute` DB function the feature asserts on, and
the claiming SQL is unconditionally global: it tags whatever untagged markers exist. The fixture therefore
cannot be made invisible to a concurrent pass, and draining the queue first is the only test-side fix.

Two failure shapes, both reproduced:

```
✘ And 6 M_ShipmentSchedule_Recompute markers remain untagged
    [Number of untagged M_ShipmentSchedule_Recompute markers remaining]
    expected: 6
     but was: 1
```
```
✘ Then 2 M_ShipmentSchedule_Recompute markers are tagged for recompute selection recomputePass:
    [Number of M_ShipmentSchedule_Recompute markers tagged for selection recomputePass …]
    Expected size: 2 but was: 0 in: []
```

## Fix

Test-side only — no production code, no feature-file change, no assertion change.

`M_ShipmentSchedule_StepDef.deleteAllUntaggedRecomputeMarkers()` — the Background's isolation step, i.e. the
step that *consumes* a quiet queue — now waits until neither `UpdateInvalidShipmentSchedules` nor
`CreateMissingShipmentSchedules` has a pending workpackage before it deletes untagged markers and the fixture
seeds. Once the queue is quiet it stays quiet for the rest of the scenario: a pass is only ever enqueued on
transaction commit (`WorkpackagesOnCommitSchedulerTemplate`), this feature completes no document, and the
seeding step inserts its rows with raw SQL.

Both processors are checked in **one** predicate so they are zero at the same poll — a create-missing run
enqueues a recompute pass, so clearing them one after the other could pass on a queue that is not quiet.

The pending-workpackage query is deliberately **not** widened to include not-ready workpackages: a
workpackage bound to a rolled-back transaction never becomes ready, so waiting on those would trade an
intermittent flake for a deterministic 60 s timeout. A workpackage currently *being processed* is still
counted — the queue clears `IsReadyForProcessing` only while building the workpackage, never during
processing — so an in-flight pass is waited for, not missed.

The query itself already existed in `C_Queue_WorkPackage_StepDef`; the first commit lifts it into
`WorkPackageQueueUtil` (unchanged semantics) so both step defs can share it. cucumber-picocontainer hands
both step-def classes the same per-scenario instance.

## Evidence

Reproduced with a temporary harness (never committed) that enqueues a real recompute pass into the
`Background`, so every scenario starts with a pass already queued:

| | runs | result |
|---|---|---|
| harness, no fix | 10 | **3 red** — both signatures above |
| harness + fix | 10 | **10 green** |
| fix only, harness reverted | 1 | 5/5 green |

Judged from the full maven logs, not the reports — the suite executes twice per run and only the second
execution writes reports. Zero `timeout` / `tryAndWait` occurrences across the ten green runs, so the drain
never approaches its deadline and is not masking the race behind a stall.

## Scope

Three files, all under `backend/de.metas.cucumber/src/test/`:

- `stepdefs/workpackage/WorkPackageQueueUtil.java` — new, shared pending-workpackage query
- `stepdefs/workpackage/C_Queue_WorkPackage_StepDef.java` — delegates to it
- `stepdefs/shipmentschedule/M_ShipmentSchedule_StepDef.java` — the drain wait

### DIFF
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/shipmentschedule/M_ShipmentSchedule_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/shipmentschedule/M_ShipmentSchedule_StepDef.java
@@ -367,8 +367,8 @@ public class M_ShipmentSchedule_StepDef
 	 * <p>
 	 * The wait closes a real race: a recompute pass enqueued by an earlier scenario claims markers through the
 	 * same DB function this feature asserts on, and that claiming SQL is unconditionally global. Both processors
-	 * are checked in ONE predicate so they are zero at the same poll — a create-missing run enqueues a recompute
-	 * pass.
+	 * are checked in ONE predicate so they are zero at the same poll: a create-missing run enqueues a recompute
+	 * pass, so checking them one after the other could pass on a queue that is not quiet.
 	 * <p>
 	 * Do NOT widen {@link WorkPackageQueueUtil#countPendingWorkPackages(String)} to not-ready workpackages: one
 	 * bound to a rolled-back transaction never becomes ready, so waiting on it would turn this intermittent flake
--- a/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/shipmentschedule/M_ShipmentSchedule_StepDef.java
+++ b/backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/shipmentschedule/M_ShipmentSchedule_StepDef.java
@@ -361,39 +361,18 @@ public class M_ShipmentSchedule_StepDef
 	}
 
 	/**
-	 * Isolation cleanup: waits until the shipment-schedule work queue is drained, then removes any leftover untagged
-	 * {@code M_ShipmentSchedule_Recompute} markers ({@code AD_PInstance_ID IS NULL}) that a prior scenario sharing
-	 * this DB may have left behind, so a whole-product recompute-batching scenario starts from a known, empty
-	 * recompute backlog that stays empty. Belongs at the start of the Background.
+	 * Isolation cleanup: waits until the shipment-schedule work queue is drained, then deletes any leftover
+	 * untagged {@code M_ShipmentSchedule_Recompute} markers ({@code AD_PInstance_ID IS NULL}). Belongs at the
+	 * start of the Background.
 	 * <p>
-	 * <b>Why the wait — the concrete failure it prevents.</b> A REAL recompute pass enqueued by an earlier scenario
-	 * on this executor is picked up by the background queue planner (500ms poll interval, see
-	 * {@code CucumberLifeCycleSupport}) and can still be in flight while this feature seeds its fixture and counts
-	 * markers GLOBALLY. That pass claims markers through the same {@code M_ShipmentSchedule_TagToRecompute} DB
-	 * function the feature asserts on, and the claiming SQL is unconditionally global — it tags whatever untagged
-	 * markers exist, so the fixture cannot be made invisible to it and draining first is the only test-side fix.
-	 * Reproduced 3 red out of 10 runs with a pass enqueued in the Background, in both variants:
-	 * <ul>
-	 *   <li>the concurrent pass tag-claimed the seeded markers —
-	 *       {@code [Number of untagged M_ShipmentSchedule_Recompute markers remaining] expected: 6 but was: 1}</li>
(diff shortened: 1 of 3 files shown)


backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/shipmentschedule/M_ShipmentSchedule_StepDef.java
```diff
@@ -336,10 +361,39 @@ private void waitUntilValid(final int timeoutSec, final Collection<ShipmentSched
 	}
 
 	/**
-	 * Isolation cleanup: removes any leftover untagged {@code M_ShipmentSchedule_Recompute} markers
-	 * ({@code AD_PInstance_ID IS NULL}) that a prior scenario sharing this DB may have left behind, so a
-	 * whole-product recompute-batching scenario starts from a known, empty recompute backlog. Belongs at the
-	 * start of the Background.
+	 * Isolation cleanup: waits until the shipment-schedule work queue is drained, then removes any leftover untagged
```
metas-ts: *way* too much prose here

metas-ts: **Resolution (dev):** Agreed — trimmed from 34 lines to 11 in 5e55306291c.

Kept only the two facts a future reader needs to not break it: *why* the wait exists (a concurrent recompute pass claims markers through the same unconditionally-global DB function this feature asserts on), and *do not widen the filter to not-ready workpackages* (one bound to a rolled-back transaction never becomes ready, so waiting on it would turn the intermittent flake into a deterministic timeout).

Dropped: the verbatim reproduction signatures and run counts, the enqueue-on-commit reasoning, and the in-flight-still-counted note. Those are evidence about the fix rather than guidance for the reader, and they live in the PR description. Also cut the chatty last sentence of the sibling `waitForRecomputeWorkQueueToDrain` javadoc.

metas-ts: Correction to the reply above, from the re-review of the trim: one of the things I dropped as "evidence" was actually guidance, and I have put it back in `5181977a57a`.

The fact that a workpackage *currently being processed* is still counted — the queue clears `IsReadyForProcessing` only while building the workpackage, never during processing — is what makes the drain's guarantee hold at all. Without it, "count reached zero" could be read as "nothing queued yet" rather than "nothing running". It now lives as one clause on `WorkPackageQueueUtil.countPendingWorkPackages()` instead of in this step def, so both callers get it once.

Also restored the consequence clause on the ONE-predicate sentence, which the trim had left dangling ("…so checking them one after the other could pass on a queue that is not quiet").

Net effect on this javadoc: still 11 lines, down from 34.

metas-ts: **Resolution (dev):** Final state of this thread — the javadoc block is **20 lines, down from 41**.

- Trimmed in https://github.com/metasfresh/metasfresh/commit/5e55306291c: dropped the verbatim reproduction signatures, run counts, poll-interval mechanics and enqueue-on-commit reasoning.
- Corrected in https://github.com/metasfresh/metasfresh/commit/5181977a57a: the "an in-flight workpackage is still counted" fact was guidance, not evidence — it is what makes the drain guarantee "nothing running", not merely "nothing queued". It now sits as one clause on `WorkPackageQueueUtil.countPendingWorkPackages()` so both callers get it once. The ONE-predicate sentence also got its consequence clause back, which the trim had left dangling.

Kept in the step def: why the wait exists, and do-not-widen-the-filter.

(Earlier replies in this thread said "11 lines, down from 34" — that was the diff-stat insert/delete count, not the resulting block size. The block is 20 lines, of which 6 are the unchanged `@cucumber.stepdef` / `@cucumber.example` tail.)

PR #25464, opened 2026-08-10
Purchase candidates: do not auto-create POs from material-disposition candidates on sales-order completion

mrie04:
## What
On sales-order completion — and re-completion after a reactivation — the `C_Order` after-complete interceptor auto-generated purchase orders from **material-disposition** purchase candidates (`Source=MaterialDisposition`), even when product planning is configured to create candidates but **not** auto-create purchase orders.

## Root cause
The "manual candidate" query in `PurchaseCandidateRepository` (`retrieveManualPurchaseCandidateIdsBySalesOrderId*`) selected candidates by `IsAggregatePO='N'`, which is the **vendor's** aggregate setting (`C_BPartner.IsAggregatePO`, default `'N'`) — not a manual-vs-dispo marker. Dispo candidates default to `IsAggregatePO='N'`, so they leaked through the "manual" filter and were auto-ordered. The purpose-built discriminator `C_PurchaseCandidate.Source` (`MaterialDisposition` / `SalesOrder` / `Api`) was never applied.

Config responsibility is otherwise correct and untouched: the dispo path (`PurchaseCandidateRequestedHandler.scheduleGeneratePurchaseOrderIfNeeded`) already gates PO generation on `PP_Product_Planning.IsCreatePlan`. Only this interceptor — a second, redundant path meant for manually-staged candidates — ignored it.

## Why it's intermittent (a race the fix closes deterministically)
The unwanted PO only appears when the interceptor picks up a **still-eligible** dispo candidate on re-completion:
- Candidate creation and the qty reset/redistribution (`PurchaseCandidateAdvisedEventCreator.doDecreaseQty`) run on the **async** material-event thread (post-commit).
- The interceptor runs **synchronously, in the completion transaction**.
- If the interceptor fires **before** the async reset lands, it sees the leftover candidate at `QtyToPurchase>0` and orders it → PO. If the reset lands first (`QtyToPurchase=0`), nothing is eligible → no PO.

So it's a race — more likely to trigger **under load** (async backlog) — which is why the symptom is intermittent and hard to reproduce on an idle instance. The fix removes the race entirely for dispo candidates: with `Source=SalesOrder`, they are never eligible, so the outcome is deterministic regardless of async timing.

## Fix (Part 1 — the interceptor)
Add a `Source=SalesOrder` filter to both query methods so only manually-created candidates (Purchase Planning view) are auto-ordered on SO completion. Dispo candidates remain candidates; the draft-vs-complete behaviour for manual candidates is unchanged. `Api`-sourced candidates are never linked to a sales order, so they were already out of scope.

## Fix (Part 2 — guarantee `Source` is always present)
The fix relies on `Source` reliably identifying manual vs dispo, but the column was nullable with no backfill (legacy rows `NULL`). This part makes it mandatory and truthful:
- New `Unknown` (`UNK`) value on the PO Sources reference + `PurchaseCandidateSource` enum (regenerated `I_/X_C_PurchaseCandidate`).
- Backfill legacy `NULL` → `UNK` (`backup_table` first), then `Source NOT NULL` + `AD_Column.IsMandatory` in sync (three migrations, data/DDL split).
- Persist path fails loudly on a null `Source`; the Purchase Planning display-builder sets `SalesOrder`.
- Interceptor unchanged: `Source=SalesOrder` already excludes `UNK` (and `MD`/`API`), so legacy/unknown candidates are never auto-ordered.

## Tests
`materialDispo_purchaseOrder_autoCreateFromProductPlanning.feature`:
- **S0265_35** — product-specific planning, `IsCreatePlan=N`: complete → candidate, no PO; reactivate → recomplete → still no PO (RED before the fix — it pins the race timing so the leftover candidate is present when the interceptor runs).
- **S0265_36** — generic (no-product, "all products") planning, same path.

`PP_Product_Planning` step-def now supports no-product plannings; the feature's DataTable columns are modernized (bare identifiers for `DataTableRow` steps) and comments made timeless.

All 6 scenarios verified locally green (S0265_10/20/25/30 unchanged + S0265_35/36 new), including after the mandatory-`Source` migrations.

## Not in this PR
`resolveIsDocComplete` orders candidate-planning lookups by `SeqNo` with no product-specificity preference (affects only draft-vs-complete for manual candidates when a specific and a generic planning coexist) — pre-existing, low priority, to be handled separately on `new_dawn_uat`.


backend/de.metas.purchasecandidate.base/src/main/java/de/metas/purchasecandidate/interceptor/C_PurchaseCandidate.java
```diff
@@ -79,4 +81,19 @@ public void updateSalesOrderLineProfit(@NonNull final I_C_PurchaseCandidate purc
...
 		saveRecord(salesOrderLineRecord);
 	}
+
+	@ModelChange(timings = ModelValidator.TYPE_BEFORE_NEW)
+	public void rejectLegacyUnknownSourceOnNew(@NonNull final I_C_PurchaseCandidate purchaseCandidateRecord)
+	{
+		// 'Unknown' is a backfill-only marker for pre-Source-column candidates. A new one carrying it
+		// would be excluded from both the sales-order interceptor and dispo auto-ordering, so a genuine
+		// purchase demand would never be ordered.
+		if (PurchaseCandidateSource.Unknown.getCode().equals(purchaseCandidateRecord.getSource()))
```
mrie04: introduce PurchaseCandidateSource.isUnknown()

mrie04: **Resolution (dev):** Applied — introduced `PurchaseCandidateSource.isUnknown()` and the guard now uses it (https://github.com/metasfresh/metasfresh/commit/5929b81750834f546de3e8b83c0319b66515f13e).

Follow-up on `ofCode` vs `ofCodeOrNull`: since `Source` has no column default and no callout, any legitimate creation path has already set it by `BEFORE_NEW`, so the guard now uses strict `PurchaseCandidateSource.ofCode(...).isUnknown()` — a missing Source is a programming error that fails loud rather than being tolerated, consistent with the repo load path (https://github.com/metasfresh/metasfresh/commit/c7a7583ec51c37b57be838d48c7034abb68c8757).

PR #25466, opened 2026-08-10
Manufacturing costing — Distribute action for the WIP cost residual (CostDifferenceDistribution)

adi-stefan:
## Manufacturing costing — post-calculation root-cause fix + the "Nachberechnung" action for the WIP cost residual

Adds a controller-driven **PP_Order_PostCalculation** action on a completed-but-not-closed manufacturing order that discharges the WIP cost residual left when the components were issued at a different cost than the finished good was received at — and fixes the two defects that made that residual impossible to compute correctly in the first place.

### Root-cause fixes

Both are long-standing and are the reason the residual formula agreed in review could not be implemented as written.

1. **`PPOrderCostTrxType`'s constructor took an `outboundCost` parameter and never assigned it** (since 2018), so `MainProduct` / `CoProduct` / `ByProduct` — all declared outbound — reported `isOutboundCost() == false`. `updatePostCalculationAmountsForCostElement` therefore folded the receipt side into `totalInboundCostAmount`, double-counting it.
2. **All three `Manufacturing*CostingMethodHandler.createComponentIssue` accumulated the original `request`** (whose amount is zero) instead of the computed `requestEffective`. `MaterialIssue` rows consequently carried a quantity but never an amount, so the issue side was absent from `PP_Order_Cost` entirely. This also corrects a second, previously undetected sign bug in the same method's reversal branch.

With both fixed, the main product's `postCalculationAmount` holds the order's total **input** cost, and the residual is `postCalculationAmount − accumulatedAmount` — exposed as `PPOrderCosts.getResidualCost()`.

The amount/quantity same-sign check in `PPOrderCost.addingAccumulatedAmountAndQty` is removed: a component issue legitimately keeps the movement direction in a negative qty while accumulating a positive input cost, and discharging the residual moves an amount against no qty at all. Documented in javadoc.

### What the action does
- New PP cost-collector type **`CostDifferenceDistribution`** (`AD_Ref_List` value `170`).
- The action computes the residual from `PP_Order_Cost` and splits it: **capitalizes** the still-in-stock portion onto the finished good's `M_Cost.CurrentCostPrice` (weighted-average, qty delta 0), and **spills** the already-shipped portion to COGS.
- Posting books a **balanced** entry: `DR P_Asset (capitalized) + DR P_COGS (spill) / CR P_WIP (residual)`, and the reverse for a negative residual. Every fact line carries **qty ZERO**, so the inventory-valuation (Lagerwert) report is unaffected.
- The collector itself carries neither product nor quantity — everything posted is recomputed from `PP_Order_Cost` at posting time.
- Discharging accumulates the residual back onto the main product's `PP_Order_Cost` line, so a re-run finds nothing to do; reversing the distribution takes it back off. Distributing also **closes** the order.

Worked example: residual 40 over 10 manufactured, 8 in stock, price 30 → capitalize 32 (Asset), spill 8 (COGS), CR WIP 40, price 30 → 34.

### Tests
- Unit (`de.metas.manufacturing`, 109 green): `PPOrderCostsTest` (post-calculation with a non-zero main-product accumulated amount, and `getResidualCost`), `PPOrderCostDifferenceDistributorTest` (split math), `PPOrderCostDifferenceDistributorCostDetailsTest` (cost details, per-AcctSchema scoping, the discharge, and the reversal re-opening the residual), `Doc_PPCostCollectorCostDifferenceDistributionTest` (balanced posting legs), `ManufacturingComponentIssueReversalTest` (the component-issue accumulation and its reversal, across all three handlers), `PP_Order_PostCalculationTest` (the precondition gate, including the standard-costing rejection and that the costing method is resolved from the accounting schema rather than the product).
- Cucumber end-to-end, **run and verified on a customer-flavored local stack** — `manufacturingCostCollectorPosting.feature`, 5 scenarios green:
  - `@Id:S30811_TC3` on **AveragePO**: issue component @34, receive finished @30 → residual 40, sell 10 leaving 8 in stock, drive the real process.
  - `@Id:S30811_TC4` on **MovingAverageInvoice** — the costing method actually in production use, which the earlier scenario did not exercise: issue @25, receive @20 → residual 50 over 10 manufactured with 6 in stock → capitalize 30, spill 20.
  - Both assert the collector and all three fact legs at qty 0, `P_WIP_Acct` netting to zero across the order's collectors, `DocStatus = CL`, and an unchanged inventory-valuation quantity before and after.

### Known limitations
- **Costing methods.** Only AveragePO, MovingAverageInvoice and LastPO accumulate into `PP_Order_Cost`; Standard costing values everything at standard and accumulates nothing, so it has no residual to discharge. The action is therefore **not offered** on a Standard-costed order rather than silently doing nothing. Note that Standard costing does not clear that difference by another route either — `MPPOrder.closeIt()` creates only usage variances, and the rate/method variance calls are commented out — so wiring Standard costing up is worth its own issue.
- **Co-products / by-products.** The discharge touches only the main-product line, so an order carrying co-products or by-products does not net to zero after discharging.
- **Delayed posting.** Posting recomputes the split from **live** `M_Cost.CurrentQty` while the cost price is moved at distribution time. If stock moves between distributing and posting, the posted split can diverge from what was capitalized. The immediate distribute→post flow (the real path, covered by the cucumber) is unaffected.

### Notes
- Core-only. Touches no `AD_Column`, so it carries no dependency on the sibling D1/D2 PR https://github.com/metasfresh/metasfresh/pull/25256.


backend/de.metas.manufacturing/src/main/java/de/metas/manufacturing/acct/Doc_PPCostCollector.java
```diff
@@ -388,4 +395,97 @@ private List<Fact> createFacts_Variance(
...
+		@NonNull CostAmount amt;
+
+		boolean isDebit()
+		{
+			return amt.signum() > 0;
+		}
+
+		CostAmount getAbsAmt()
+		{
+			return amt.signum() < 0 ? amt.negate() : amt;
```
adi-stefan: use either `amt.negateIf(amt.signum() < 0)` or a dedicated dedicated method on the CostAmount that returns the absolute amount

backend/de.metas.manufacturing/src/main/java/org/eevolution/process/PP_Order_PostCalculation.java
```diff
@@ -0,0 +1,79 @@
...
+ * not yet closed.
+ */
+public class PP_Order_Distribute extends JavaProcess implements IProcessPrecondition
+{
+	private final PPOrderCostDifferenceDistributor costDifferenceDistributor = SpringContextHolder.instance.getBean(PPOrderCostDifferenceDistributor.class);
+
+	@Override
+	public ProcessPreconditionsResolution checkPreconditionsApplicable(final IProcessPreconditionsContext context)
+	{
+		final I_PP_Order ppOrder = context.getSelectedModel(I_PP_Order.class);
```
adi-stefan, line 55: think a `rejectIfNotSingleSelection()` would not hurt here

backend/de.metas.manufacturing/src/main/java/de/metas/manufacturing/acct/Doc_PPCostCollector.java
adi-stefan: Adopted `amt.negateIf(amt.signum() < 0)`. A grep showed this was the only such ternary on a `CostAmount`, so a dedicated `abs()` on the value class would have had one caller. 8704350fe88

backend/de.metas.manufacturing/src/main/java/org/eevolution/process/PP_Order_PostCalculation.java
adi-stefan: Added `rejectBecauseNotSingleSelection().toInternal()`, matching `PP_Order_Weighting_Run_Process`. It was a real gap: `doIt()` acts on `getRecord_ID()` while the precondition read an arbitrary order via `getSelectedModel`. 8704350fe88

adi-stefan: `PP_Order_UnClose` now carries the same guard — same mismatch. I swept the rest of `de.metas.manufacturing` for the shape; these two were the only ones. 03e9575d8e0

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
```diff
@@ -0,0 +1,363 @@
...
+					|| !materialCostElementId.equals(cost.getCostElementId()))
+			{
+				continue;
+			}
+
+			final PPOrderCostTrxType trxType = cost.getTrxType();
+			if (trxType == PPOrderCostTrxType.MaterialIssue)
+			{
+				// MaterialIssue rows accumulate a negative qty, hence the negate().
+				issued = issued.add(cost.getPrice().multiply(cost.getAccumulatedQty()).negate());
```
teosarca: why not getAccumulatedAmt()?

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
```diff
@@ -0,0 +1,363 @@
...
+
+		final ResidualAndManufacturedQty residualAndQty = computeResidualAndManufacturedQtyForOrder(orderId, acctSchemaId);
+		if (residualAndQty == null || residualAndQty.getResidual().isZero())
+		{
+			return;
+		}
+
+		costCollectorsService.createCostDifferenceDistribution(
+				order,
+				ProductId.ofRepoId(order.getM_Product_ID()),
```
teosarca: productId param is reduntant

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
```diff
@@ -0,0 +1,363 @@
...
+		}
+
+		final CurrentCost currentCost = utils.getCurrentCost(request);
+		final CostAmountDetailed split = computeSplit(
+				residualAndQty.getResidual(),
+				residualAndQty.getManufacturedQty(),
+				currentCost);
+
+		final CostDetailCreateResult mainResult = utils.createCostDetailRecordNoCostsChanged(
+				request.withAmountAndType(split.getMainAmt(), CostAmountType.MAIN),
```
teosarca, line 171: shall we also make sure the qty is either residualAndQty.getManufacturedQty() or ZERO?

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
```diff
@@ -0,0 +1,363 @@
...
+		final ResidualAndManufacturedQty residualAndQty = computeResidualAndManufacturedQtyForOrder(orderId, acctSchemaId);
+		if (residualAndQty == null || residualAndQty.getResidual().isZero())
+		{
+			return;
+		}
+
+		costCollectorsService.createCostDifferenceDistribution(
+				order,
+				ProductId.ofRepoId(order.getM_Product_ID()),
+				residualAndQty.getManufacturedQty());
```
teosarca: because later we are calc again the residual amount + received qty, shall we alwyas use ZERO here (and remove the param)?

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
```diff
@@ -0,0 +1,363 @@
...
+
+			final PPOrderCostTrxType trxType = cost.getTrxType();
+			if (trxType == PPOrderCostTrxType.MaterialIssue)
+			{
+				// MaterialIssue rows accumulate a negative qty, hence the negate().
+				issued = issued.add(cost.getPrice().multiply(cost.getAccumulatedQty()).negate());
+			}
+			else if (trxType == PPOrderCostTrxType.MainProduct
+					|| trxType == PPOrderCostTrxType.CoProduct
+					|| trxType == PPOrderCostTrxType.ByProduct)
```
teosarca: i tihnk co/by product need to be subtracted. so either document is out of scope either do it right.

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	/**
+	 * Recomputes {@code residual = issued - received} over the order's {@code PP_Order_Cost} rows of the given
+	 * schema and material cost element: {@code issued} sums {@code -accumulatedQty x price} over the MaterialIssue
+	 * rows, {@code received} sums the post-calculation amount over the MainProduct / CoProduct / ByProduct rows.
+	 */
+	@Nullable
+	@VisibleForTesting
+	static ResidualAndManufacturedQty computeResidualAndManufacturedQty(
```
teosarca: pls introduce a method in PPorderCosts and use it.
residual cost shall be postCalculationAmount - accumulatedAmount of the main product (finished goods) line.

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
```diff
@@ -0,0 +1,363 @@
...
+				received = received.add(cost.getPostCalculationAmount());
+			}
+		}
+
+		final CostAmount residual = issued.subtract(received);
+		return ResidualAndManufacturedQty.of(residual, mainProductCost.getAccumulatedQty());
+	}
+
+	@Nullable
+	private static PPOrderCost getMainProductCostOrNull(
```
teosarca: not needed or if need it it shall be a method of PPOrderCosts

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
```diff
@@ -0,0 +1,363 @@
...
+		{
+			throw new AdempiereException("Expected at most one main-product PP_Order_Cost row for acctSchema=" + acctSchemaId
+					+ ", costElement=" + materialCostElementId + " in " + orderCosts);
+		}
+
+		return mainProductCosts.isEmpty() ? null : mainProductCosts.get(0);
+	}
+
+	@Value(staticConstructor = "of")
+	static class ResidualAndManufacturedQty
```
teosarca: not needed

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
```diff
@@ -0,0 +1,363 @@
...
+				.map(CostElement::getId)
+				.collect(GuavaCollectors.singleElementOrThrow(
+						() -> new AdempiereException("Expected exactly one material cost element for costing method " + costingMethod + " but got " + materialElements)));
+	}
+
+	/** Splits the residual pro rata: the still-in-stock share is capitalized, the shipped remainder goes to COGS. */
+	@VisibleForTesting
+	static CostAmountDetailed computeSplit(
+			@NonNull final CostAmount residual,
+			@NonNull final Quantity manufacturedQty,
```
teosarca: might be that u will pass the main product's PPOrderCost as param

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
```diff
@@ -0,0 +1,363 @@
...
+		final CostingMethod costingMethod = acctSchema.getCosting().getCostingMethod();
+		final CostElementId materialCostElementId = getMaterialCostElementId(costingMethod);
+
+		final PPOrderCosts orderCosts = ppOrderCostsService.getByOrderId(orderId);
+		return computeResidualAndManufacturedQty(orderCosts, acctSchemaId, materialCostElementId);
+	}
+
+	private boolean isAlreadyDistributed(@NonNull final PPOrderId orderId)
+	{
+		return costCollectorsService.getByOrderId(orderId).stream()
```
teosarca: introduce isAlreadyDistributed to service and then to repositroy and use IQueryBUilder to check it

backend/de.metas.manufacturing/src/main/java/de/metas/manufacturing/acct/Doc_PPCostCollector.java
```diff
@@ -388,4 +395,97 @@ private List<Fact> createFacts_Variance(
...
+		{
+			final Account account = docLine.getAccount(leg.getAcctType(), as);
+			final CostAmount absAmt = leg.getAbsAmt();
+			fact.createLine()
+					.setDocLine(docLine)
+					.setAccount(account)
+					.setAmtSource(absAmt.getCurrencyId(),
+							leg.isDebit() ? absAmt.toBigDecimal() : null,
+							leg.isDebit() ? null : absAmt.toBigDecimal())
+					.setQty(leg.isDebit() ? qty : qty.negate())
```
teosarca: important: this shall be zero to make sure lagerwert (inventory valuation) report is not affected!

backend/de.metas.manufacturing/src/main/java/de/metas/manufacturing/acct/Doc_PPCostCollector.java
```diff
@@ -388,4 +395,97 @@ private List<Fact> createFacts_Variance(
...
+		}
+
+		final ImmutableList.Builder<CostDifferenceDistributionLeg> legs = ImmutableList.builder();
+		addLegIfNotZero(legs, ProductAcctType.P_Asset_Acct, split.getCostAdjustmentAmt());
+		addLegIfNotZero(legs, ProductAcctType.P_COGS_Acct, split.getAlreadyShippedAmt());
+		addLegIfNotZero(legs, ProductAcctType.P_WIP_Acct, residual.negate());
+		return legs.build();
+	}
+
+	private static void addLegIfNotZero(
```
teosarca, line 474: addCostDifferenceFactLine

backend/de.metas.manufacturing/src/main/java/org/eevolution/api/impl/PPCostCollectorBL.java
```diff
@@ -297,6 +297,29 @@ private I_PP_Cost_Collector createFinishedGoodsReceipt(@NonNull final ReceiptCos
...
+
+		return createCollector(
+				CostCollectorCreateRequest.builder()
+						.costCollectorType(CostCollectorType.CostDifferenceDistribution)
+						.order(order)
+						.productId(productId)
+						.locatorId(locatorId)
+						.attributeSetInstanceId(asiId)
+						.resourceId(ResourceId.ofRepoId(order.getS_Resource_ID()))
+						.movementDate(SystemTime.asZonedDateTime())
```
teosarca: no static calls here! pls provide it as param.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/pporder/PP_Order_StepDef.java
```diff
@@ -412,6 +417,25 @@ public void order_action(
 		}
 	}
 
+	/**
+	 * Runs {@link PP_Order_Distribute} on the given order. It is a plain {@code AD_Process}, not a document
```
teosarca: call the process PP_Order_PostCalculation or similar

teosarca (dismissed):
check my changes

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
adi-stefan: **Resolution (dev):** Dropped. `createCostDifferenceDistribution(order, movementDate)` derives the product from the order. b7d890f38f0

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
adi-stefan: **Resolution (dev):** Taking ZERO — the collector now carries no quantity at all, which settles your Lagerwert point below for free. The cucumber `MovementQty 10` assertion went with it. b7d890f38f0

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
adi-stefan: **Resolution (dev):** Done — the MAIN cost detail is created with an explicit zero qty rather than inheriting the request. b7d890f38f0

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
adi-stefan: **Resolution (dev):** Deleted; `PPOrderCosts.getMainProductCostOrNull` replaces it. b7d890f38f0

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
adi-stefan: **Resolution (dev):** Deleted. b7d890f38f0

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
adi-stefan: **Resolution (dev):** `computeSplit` now takes the main-product `PPOrderCost` and reads the qty off it. b7d890f38f0

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
adi-stefan: **Resolution (dev):** Moved to the repository as asked — then removed again, because it turned out to be wrong once the residual self-discharges: after `PP_Order_UnClose` plus a further issue or receipt a legitimate new residual exists, and the guard would block it permanently. The zero-residual early return is now the sole guard. Say the word if you want it back. b7d890f38f0

backend/de.metas.manufacturing/src/main/java/de/metas/manufacturing/acct/Doc_PPCostCollector.java
adi-stefan: **Resolution (dev):** Done — every leg carries qty 0, so inventory valuation is untouched. Asserted in the cucumber scenario. b7d890f38f0

backend/de.metas.manufacturing/src/main/java/de/metas/manufacturing/acct/Doc_PPCostCollector.java
adi-stefan: **Resolution (dev):** Extracted into `addCostDifferenceFactLine`. b7d890f38f0

backend/de.metas.manufacturing/src/main/java/org/eevolution/api/impl/PPCostCollectorBL.java
adi-stefan: **Resolution (dev):** Done — the movement date is a parameter now. b7d890f38f0

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/pporder/PP_Order_StepDef.java
adi-stefan: **Resolution (dev):** Renamed to `PP_Order_PostCalculation`, including the `AD_Process` row (migration 5820730). b7d890f38f0

backend/de.metas.manufacturing/src/main/java/de/metas/costing/methods/PPOrderCostDifferenceDistributor.java
adi-stefan: Implemented — see the reply on the thread below for what had to be fixed first. b7d890f38f0

adi-stefan: **Resolution (dev):** implemented as you wrote it — `PPOrderCosts.getResidualCost()` = `postCalculationAmount - accumulatedAmount` of the main-product line. Recording what had to be fixed first, because until then it returned a structural **0** on every order.

Two defects meant the inputs were never populated:

1. `PPOrderCostTrxType`'s constructor took `outboundCost` and never assigned it, so `MainProduct` / `CoProduct` / `ByProduct` reported `isOutboundCost() == false` and the receipt side was counted into `totalInboundCostAmount`.
2. All three `Manufacturing*CostingMethodHandler.createComponentIssue` accumulated the original `request` (amount zero) instead of `requestEffective`, so `MaterialIssue` rows never carried an amount.

Together those made `MR.postCalculationAmount == MR.accumulatedAmount`. Measured before: `MI cumulatedamt=0`, `MR 25/25` → residual 0. After: `MI cumulatedamt=340 cumulatedqty=-10`, `MR cumulatedamt=300 postcalculationamt=340` → **40**.

Fixing only the first would have been worse than fixing neither — the receipt leg drops out of `PP_Order.CostDifference` and both existing scenarios go from 15 / -5 to **-10**. Verified by running them.

Two consequences worth your view:

- **The same-sign check in `PPOrderCost.addingAccumulatedAmountAndQty` is gone** — a component issue keeps a negative qty against a positive input cost, and the discharge moves an amount against no qty. A narrower trx-type-aware guard fails two existing `PPOrderCostTest` fixtures that use `+100` against `+1`, a shape real data never has. Happy to add it plus corrected fixtures if you prefer.
- **`PP_Order.CostDifference` had to stop reading `postcalculationamt`** — once post-calculation is right, that field holds the *issued* total on the main-product line. It reads `cumulatedamt` now; that change is in #25256 so this PR has no cross-branch schema dependency.

PR #25469, opened 2026-08-10
AD_Issue: bound the error-handling recursion that can exhaust the heap

metas-ts:
## Problem

Persisting an `AD_Issue` is itself a database write that runs inside the transaction manager. When that write fails, `AbstractTrxManager.call0` logs the failure through the ambient `ILoggable`, passing the throwable as the **last** parameter:

```java
catch (final Throwable runException) {
    final ILoggable loggable = Loggables.withLogger(logger, Level.WARN);
    if (AdempiereException.isThrowableLoggedInTrxManager(runException)) {
        loggable.addLog("AbstractTrxManager.call0 - caught {} with message={}",
                        runException.getClass(), runException.getMessage(),
                        runException);
    }
```

`LoggableWithThrowableUtil.extractMsgAndAdIssue` treats a trailing `Throwable` as a request to create an `AD_Issue`. So the cycle closes:

```
ErrorManager.createIssue(Throwable)
 └ createIssueInTrx → InterfaceWrapperHelper.saveRecord(AD_Issue)
    └ PO.saveEx() → AbstractTrxManager.run → call0
       └ catch (Throwable) → loggable.addLog(…, runException)
          └ ApiAuditLoggable.addLog → createLogEntry
             └ LoggableWithThrowableUtil.extractMsgAndAdIssue
                └ errorManager.createIssue(exception)   ← recurses
```

Nothing bounded the depth. `extractMsgAndAdIssue` *does* wrap `createIssue` in `catch (final Exception …)`, but the recursion happens **below** that frame, so the guard never closes the loop.

Each level's stacktrace text embeds the level below it, so the strings grow without bound.

## Evidence

From a production heap dump — one Tomcat request thread:

- **4,222 stack frames**, containing **171** `ErrorManager.createIssueInTrx` and **170** `LoggableWithThrowableUtil.extractMsgAndAdIssue` frames
- **170 `DBException`** retaining 706,950,040 B and **171 `MIssue`** retaining 228,412,936 B — roughly **7.9 MB each**
- 1,289,453,896 B (39.51 % of the heap) held in that one thread's locals

It died inside the string work itself:

```
java.lang.OutOfMemoryError.<init>
 Arrays.copyOfRange → StringBuffer.toString → Matcher.replaceAll → String.replace
 MIssue.truncateExceptionsRelatedString    MIssue.java:100
 MIssue.setStackTrace                      MIssue.java:81
```

The triggering request came in through `ApiAuditFilter` → `ApiAuditService.processRequestSync`, i.e. an audited REST call, whose `ApiAuditLoggable` is the loggable that converts throwables into `AD_Issue`s.

## Fix

`ErrorManager` marks the region in which it persists an `AD_Issue`; `LoggableWithThrowableUtil` skips issue creation inside that region and only logs:

```java
try (final IAutoCloseable ignored = LoggableWithThrowableUtil.suppressAdIssueCreation())
{
    return trxManager.callInNewTrx(() -> createIssueInTrx(request));
}
```

This bounds the nesting at one level while preserving the log record and the formatted message. The region is **save/restore** rather than set/clear, so a nested region closing cannot re-enable issue creation early.

Deliberately placed at `ErrorManager.createIssue(IssueCreateRequest)` because both public entry points funnel through it, and no public contract changes — the return type and nullability of `createIssue` are untouched, so none of the ~15 call sites is affected.

## Tests

`LoggableWithThrowableUtilTest`:

- an `AD_Issue` **is** created when not suppressed (guards against the fix silently disabling issue creation everywhere)
- **no** `AD_Issue` is created while suppressed
- nested regions restore correctly, and issue creation works again after the outermost region closes
- the message is still formatted and returned while suppressed

with `CountingErrorManager`, a test double registered via `Services.registerService`, so no database is involved.

`ErrorManagerRecursionGuardTest` then pins the **integration point** — the wiring those unit tests do not cover, and the thing a refactor would silently break:

- the throwable's `getLocalizedMessage()` throws, so `createIssueInTrx` fails while building the summary (`AdempiereException.extractMessage` calls it unguarded) before any save is attempted;
- `AbstractTrxManager.call0` catches it and logs it through the ambient `ILoggable`;
- the installed loggable mimics `ApiAuditLoggable`, routing that throwable back into AD_Issue creation.

The counter **decorates the real `ErrorManager`** rather than stubbing it: a stub never runs `createIssueInTrx`, and counting inside the poisoned throwable cannot distinguish the fixed from the unfixed path, because the nested attempt operates on a different, renderable exception. Counting the calls that arrive via `Services.get` gives **1** with the guard and **2** without. This particular failure path needs no database: it fails in `buildIssueSummary` before any save is attempted, and `Trx.getConnection()` is lazy. (The module's test environment does provide one — the claim is about this path, not about the suite.)

## Validation status

**Not validated locally.** `repo.metasfresh.com` is returning **502** for every request, and it is the only public host for several build dependencies (`net.dev.java:pdf-renderer:0.9.0`, `org.jpedal:jpedal:1.0` — neither is on Maven Central), so no local module chain can be built at the moment. CI is the first real verdict on this branch. Flagging this rather than implying a local green.

## Relationship to the cache-label leak PR

Independent defect, separate PR. In the incident that surfaced both, the slow cache-label leak consumed the headroom and this recursion then converted a recoverable failure into an `OutOfMemoryError`. Whether this recursion can also fire on an otherwise-healthy instance is **not** established — it needs a failing `AD_Issue` persist, which is why it is worth fixing on its own merits.

### DIFF
--- a/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/error/LoggableWithThrowableUtil.java
+++ b/backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/error/LoggableWithThrowableUtil.java
@@ -8,6 +8,7 @@ import org.slf4j.Logger;
 import de.metas.logging.LogManager;
 import de.metas.util.ILoggable;
 import de.metas.util.Services;
+import org.adempiere.util.lang.IAutoCloseable;
 import de.metas.util.StringUtils;
 import lombok.NonNull;
 import lombok.Value;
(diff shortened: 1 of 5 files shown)


metas-ts (commented):
lgtm

PR #25498, opened 2026-08-12
ESR import: a duplicate incoming payment gets its own C_Payment

metas-ts:
## Problem

A **genuinely new** incoming ESR payment is mistaken for one already seen, because the candidate search matches on partner + amount (+ transaction date) only. The typical case is a recurring payment — rent for the following month: same partner, same amount, same invoice reference as last month, but a different transaction and real new money.

Instead of creating a new `C_Payment`, the import pulled out the **previous** payment and attached it to the new line, skipping payment creation entirely. So the second receipt was never booked, the bank statement could not be reconciled, and the import could not be completed.

The same code path also covers the case where money genuinely is sent twice. Either way the outcome must be the same: the line is flagged for the accountant to inspect, **and** the incoming money gets its own payment.

`ESRImportBL.handleEsrImportLine` did three things on a positive duplicate detection: flag the line, attach the payment it had just *found*, and `return true`. That early return marked the line "already handled", so it was excluded from `linesToProcess` and never reached payment creation.

## Fix

The candidate search is deliberately left **unchanged**: it cannot tell a recurring payment apart from a re-sent one (partner, amount, invoice reference and even the raw ESR line text can all be identical), so it must keep flagging the line for the accountant. What changes is the *consequence* of a match.

`handleEsrImportLine` now keeps the `Duplicate_Payment` flag and lets the line fall through to the regular handling, which creates its own payment. Three statements were removed:

- `line.setC_Payment_ID(paymentId.getRepoId())` — the payment found belongs to a **different**, earlier transaction; this line is new money.
- `return true` — it skipped payment creation.
- `handleUnsupportedTrxType(line)` — it can set `Processed=Y` under the `de.metas.payment.esr.ProcessUnspportedTrxTypes` sysconfig, which would defeat the requirement that a duplicate line remain unprocessed for the accountant to act on.

The line then reaches the same machinery a duplicate **within one file** has always used (`processLinesWithPaymentNoInvoice` → `createUnlinkedPaymentForLine`): it gets its own payment, left unallocated because the invoice is already paid. In other words, the cross-import case now converges on the behaviour `ESRImportTest#testDoublePayment_T02` has asserted for the same-file case all along.

Deliberately **not** changed, both considered and rejected with evidence:

- **Narrowing the candidate search** (a day-scoped `DateTrx` filter in `AbstractPaymentDAO.retrievePaymentIds`). Implemented and falsified: removing the candidate means the duplicate is never *detected*, so `ESR_Payment_Action` is never set and the accountant loses the warning.
- **Removing the sibling-line match** in `ESRImportDAO.findExistentPaymentId`. That branch is *detection*, not the defect; removing it leaves the same-day case unflagged.

## Second change — a re-entrancy artefact

`setInvoice` runs twice for one line: once while it is evaluated, and again from the `C_Payment_ID` `@ModelChange` interceptor after the line's payment is created. Everything it derives is idempotent except the appended match-error note, so the "invoice already paid" message was recorded twice once lines started falling through. Added `ESRDataLoaderUtil.addMatchErrorMsgIfNotPresent` and routed both call sites through one `markInvoiceAlreadyPaid` helper.

## Scope boundary — the fix is forward-only

`handleEsrImportLine` returns early for any line that already carries a payment. Import lines that the **old** code already mis-attached therefore keep the wrong payment after deployment, and the money they represent stays unbooked. Repairing existing rows is separate work and is **not** part of this PR.

## Tests

Three new tests, all written and shown RED before the fix:

| Test | Fixture |
|---|---|
| `DuplicatePaymentGetsOwnPayment.arrivingOnALaterDay` | second import one day after the first |
| `DuplicatePaymentGetsOwnPayment.arrivingOnTheSameDay` | both imports on the same payment date |
| `testReprocessSameImport_createsNoSecondPayment` | re-processing the same import header |

Each asserts the duplicate line gets its **own** payment (not the sibling's), for the right amount, unallocated (`C_Invoice_ID == 0`), flagged `Duplicate_Payment`, and left **unprocessed**.

`testReprocessSameImport_createsNoSecondPayment` pins the header-level `Processed=Y` interlock, which refuses re-processing before the duplicate search runs at all.

Also repaired: `testDuplicatePayments` closed with `assertThat(x.getC_Payment_ID(), is(x.getC_Payment_ID()))` — a self-comparison that could never fail. It now asserts which payment the line actually carries, and is the only test covering the null-`PaymentDate` fixture and the exactly-once match-error note.

**Verified locally** (2026-08-12, after a full `mvn_build.sh --skip-tests` reinstall so no stale jar could mask the result):

- `ESRImportTest` — 67 run, 0 failures
- `de.metas.payment.esr` module — 160 run, 0 failures, 1 pre-existing skip
- `de.metas.business` — 622 run, 0 failures
- `de.metas.banking.base` — 97 run, 0 failures, 2 pre-existing skips

### DIFF
--- a/backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/api/impl/ESRImportBL.java
+++ b/backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/api/impl/ESRImportBL.java
@@ -644,11 +644,12 @@ public class ESRImportBL implements IESRImportBL
 		final PaymentId paymentId = fetchDuplicatePaymentIfExists(line);
 		if (paymentId != null)
 		{
+			// Flag the line so the accountant still sees the duplicate, but do not attach the payment we
+			// just found: that payment belongs to the earlier transaction, while this line is money that
+			// arrived a second time and needs its own C_Payment. Falling through to the regular handling
+			// lets processLinesWithPaymentNoInvoice create it (unallocated, since the invoice is paid).
 			line.setESR_Payment_Action(X_ESR_ImportLine.ESR_PAYMENT_ACTION_Duplicate_Payment);
-			handleUnsupportedTrxType(line);
-			line.setC_Payment_ID(paymentId.getRepoId());
 			esrImportDAO.save(line);
-			return true;
 		}
 		return false;
 	}
--- a/backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/api/impl/ESRImportBL.java
+++ b/backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/api/impl/ESRImportBL.java
@@ -129,6 +129,8 @@ public class ESRImportBL implements IESRImportBL
 
 	private static final AdMessageKey ESR_NO_HAS_WRONG_ORG_2P = AdMessageKey.of("de.metas.payment.esr.EsrNoHasWrongOrg");
 
+	private static final AdMessageKey ESR_INVOICE_ALREADY_PAID_1P = AdMessageKey.of("de.metas.payment.esr.InvoiceAlreadyPaid");
+
 	/**
 	 * Filled by {@link #registerActionHandler(String, IESRActionHandler)}.
 	 */
@@ -1178,7 +1180,9 @@ public class ESRImportBL implements IESRImportBL
 	 */
 	private void markInvoiceAlreadyPaid(@NonNull final I_ESR_ImportLine importLine, @NonNull final I_C_Invoice invoice)
 	{
-		ESRDataLoaderUtil.addMatchErrorMsgIfNotPresent(importLine, "Rechnung " + invoice.getDocumentNo() + " wurde im System als bereits bezahlt markiert");
+		final Properties ctx = getCtx(importLine);
+		ESRDataLoaderUtil.addMatchErrorMsgIfNotPresent(importLine,
+														msgBL.getMsg(ctx, ESR_INVOICE_ALREADY_PAID_1P, new Object[] { invoice.getDocumentNo() }));
 		importLine.setESR_Document_Status(X_ESR_ImportLine.ESR_DOCUMENT_STATUS_PartiallyMatched);
 	}
 
--- a/backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/api/impl/ESRImportBL.java
+++ b/backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/api/impl/ESRImportBL.java
@@ -644,10 +644,9 @@ public class ESRImportBL implements IESRImportBL
 		final PaymentId paymentId = fetchDuplicatePaymentIfExists(line);
 		if (paymentId != null)
 		{
-			// Flag the line so the accountant still sees the duplicate, but do not attach the payment we
-			// just found: that payment belongs to the earlier transaction, while this line is money that
-			// arrived a second time and needs its own C_Payment. Falling through to the regular handling
-			// lets processLinesWithPaymentNoInvoice create it (unallocated, since the invoice is paid).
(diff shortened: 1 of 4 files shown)


backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/api/impl/ESRImportBL.java
```diff
@@ -1173,6 +1171,17 @@ else if (invoice.isPaid())
...
 	}
 
+	/**
+	 * {@code setInvoice} runs twice for one line — once while it is evaluated, once from the
+	 * {@code C_Payment_ID} interceptor after the line's payment is created — so this note has to be as
+	 * idempotent as the document status it accompanies.
+	 */
+	private void markInvoiceAlreadyPaid(@NonNull final I_ESR_ImportLine importLine, @NonNull final I_C_Invoice invoice)
+	{
+		ESRDataLoaderUtil.addMatchErrorMsgIfNotPresent(importLine, "Rechnung " + invoice.getDocumentNo() + " wurde im System als bereits bezahlt markiert");
```
metas-ts: good code-deduplication, but we also need an AD_Message here

metas-ts: **Resolution (dev):** Agreed — and the dedup is what makes it actionable, since the string now exists in exactly one place instead of two.

In progress, following the convention already in this file (`AdMessageKey` at `ESRImportBL.java:130`, `msgBL.getMsg(ctx, KEY, params)` at `:1162`):

- the hardcoded `"Rechnung " + invoice.getDocumentNo() + " wurde im System als bereits bezahlt markiert"` becomes an `AdMessageKey` constant resolved via `msgBL.getMsg(...)`, with a `{0}` parameter placeholder instead of string concatenation;
- the `AD_Message` + `AD_Message_Trl` rows ship as a migration script in the module's SQL tree, with every ID taken from the central ID server;
- the existing `de_DE` wording is preserved verbatim so nothing changes for German users, and an `en_US` translation is added.

`ESRImportTest#testDuplicatePayments` asserts that note appears exactly once (that invariant is the reason for the second commit on this branch), so it is adapted to the resolved-message contract without weakening the exactly-once check.

I will follow up on this thread with the commit link once it is pushed.

metas-ts: **Resolution (dev):** Done — https://github.com/metasfresh/metasfresh/commit/6645577eeb9

- `markInvoiceAlreadyPaid` now resolves the text via `msgBL.getMsg(getCtx(importLine), ESR_INVOICE_ALREADY_PAID_1P, new Object[]{ invoice.getDocumentNo() })`; no hardcoded literal remains in the method.
- New `AD_Message` `de.metas.payment.esr.InvoiceAlreadyPaid` (`AD_Message_ID` 545795, `MsgType='I'`, `EntityType='de.metas.payment.esr'` — matching the sibling `ESR_Reverse_Booking`, which is also written through `addMatchErrorMsg`). The `de_DE` text keeps the previous wording with a `{0}` placeholder for the document number; `en_US` added.
- Migration script `5818570_sys_AD_Message_ESR_InvoiceAlreadyPaid.sql`; both the message id and the script prefix came from the central ID server.

One consequence worth flagging: `ESRImportTest#testDuplicatePayments` previously asserted the literal German sentence and now asserts `de.metas.payment.esr.InvoiceAlreadyPaid_[654321]`. That is not a weakened assertion — `AdempiereTestHelper` enables unit-test mode, in which `Msg.lookup()` always short-circuits to the missing-message representation regardless of whether the `AD_Message` row exists, so no fixture registration can surface the real German text in a unit test. The assertion still proves the note appears exactly once (the invariant the preceding commit exists for) and still fails if it were duplicated or absent.

PR #25501, opened 2026-08-12
VAT-ID check — extract VAT-ID validation into its own module and add the VIES check foundations

metas-ts:
**Draft — foundations only. Further work follows on this branch.**

Extracts VAT-ID validation out of `adempiere.base` into a new module and adds the storage, configuration and audit foundations for an online VIES check. No VIES call is made yet, and nothing consumes the new configuration or status values yet — those arrive in later commits on this branch.

## New module

`de.metas.vatid`, mirroring `de.metas.vertical.creditscore`: a parent pom with `metasfresh-vatid-base` and `metasfresh-vatid-vies`, reaching the runtime through `de.metas.fresh.base`.

Moved into it, behaviour unchanged:

- `EUVatIdValidator`, `VATaxIDValidationUtil` and their test, from `de.metas.bpartner.vatid`.
- The `C_BPartner` and `C_BPartner_Location` VAT-ID interceptors, now `de.metas.vatid.interceptor.*`. The now-empty `C_BPartner_Location` interceptor in `adempiere.base` was removed.

## Status domain

- `AD_Reference` for the check status with six values: `NotChecked`, `RequestSent`, `Valid`, `Invalid`, `NotSupported`, `ServiceUnavailable`.
- `de.metas.vatid.VATaxIDStatus`, a reference-list-aware enum sourcing its codes from the generated model, with a predicate for whether a status counts as holding a tax certificate.
- A separate two-value reference for the on-service-unavailable fallback setting (`ServiceUnavailable` or `Invalid`), so a configuration choice cannot be set to a status that could never describe that outcome.

## Configuration

`VATaxID_Config`, one active row per organisation (partial unique index), with its window, menu entry under `Finanzen → Einstellungen`, and `VATaxIDConfigRepository`. Settings: format-check and VIES-check toggles, REST base URL, requester member-state code and number, recheck interval, and the fallback action.

## Check log

`VATaxID_CheckLog` plus its read-only window, menu entry under `Finanzen`, and `VATaxIDCheckRepository`.

- Append-only except for one transition: a `RequestSent` row is later completed to a final status. The repository exposes only those two operations, and completion claims the row under a `SELECT … FOR NO KEY UPDATE` lock before saving, so a concurrent second completion fails rather than silently overwriting.
- The window is locked at tab level following the `AD_ChangeLog` / `AD_Issue` precedent: no insert, no delete, fields read-only.
- `AD_PInstance_ID` and `AD_Session_ID` are indexed but deliberately carry no foreign key, because both parents are purge targets.
- Reverse zoom from the business partner is wired, so a partner's check history is reachable from the partner record.

## Tests

Unit tests for the enum and both repositories; Playwright specs for the configuration window and the check-log window, the latter covering menu reachability, the read-only guarantee, and reverse-zoom reachability from a partner.

## Migrations

21 scripts under `backend/de.metas.vatid/base/src/main/sql/postgresql/system/45-de.metas.vatid/`. Several correct earlier scripts in the same, not-yet-merged chain rather than editing them after they had been applied locally; the whole chain has been verified to apply cleanly from a virgin database and to converge on the intended end state.


backend/de.metas.vatid/base/src/main/java/de/metas/vatid/VATaxIDValidationUtil.java
```diff
@@ -34,7 +35,7 @@
  *
  * <p>{@link #SYSCONFIG_validateVATaxID} is the gate; reading it is the caller's responsibility (the
  * interceptors hold an {@code ISysConfigBL} instance field), so this helper itself stays free of any
- * service-locator call. {@link #validate(String)} performs the actual check via
+ * service-locator call. {@link #validate(VATIdentifier)} performs the actual check via
```
metas-ts: please add a **short** javadoc to SYSCONFIG_validateVATaxID. About what it means if this is Y resp N

metas-ts: **Resolution (dev):** Added, kept short as asked.

It now states what `Y` and `N` actually do — `Y`: a VAT-ID whose format is wrong for its country prefix is rejected on save and the record is not stored; `N`: the check is skipped, so legacy data that would otherwise be unsavable can be imported.

Two things I added because they were the non-obvious part:

- **What an absent/unset row counts as** — `Y`. The callers pass `true` as the default, and the shipped row is a single System-level `Y` with no organisation override. I confirmed that by querying `ad_sysconfig` on the local stack rather than inferring it from the DDL.
- **What it does *not* gate** — only the save-time check. The format check inside `VATaxIDCheckService` is governed by the per-organisation `VATaxID_Config` instead, so turning this off does not turn that off.

Commit: https://github.com/metasfresh/metasfresh/commit/a256a93dc3e

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/interceptor/C_BPartner_Location.java
```diff
@@ -0,0 +1,90 @@
...
+
+	/**
+	 * Unchanged synchronous format check: still {@code BEFORE_NEW}/{@code BEFORE_CHANGE}, still throws
+	 * hard on a malformed value, still gated by the same {@code C_BPartner.validateVATaxID} SysConfig. Not
+	 * touched by the after-commit trigger below — a value this rejects never reaches a save at all, let
+	 * alone the online check.
+	 */
+	@ModelChange(timings = { ModelValidator.TYPE_BEFORE_NEW, ModelValidator.TYPE_BEFORE_CHANGE },
+			ifColumnsChanged = I_C_BPartner_Location.COLUMNNAME_VATaxID)
+	public void validateVATaxID(@NonNull final I_C_BPartner_Location bpLocation)
```
metas-ts, line 65: are you assuming that this method is always run before #scheduleVATaxIDCheck ?

Asking because I don't think that's a safe assumption.

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/interceptor/C_BPartner.java
```diff
@@ -0,0 +1,92 @@
...
+	 * Unchanged synchronous format check: still {@code BEFORE_NEW}/{@code BEFORE_CHANGE}, still throws
+	 * hard on a malformed value, still gated by the same {@code C_BPartner.validateVATaxID} SysConfig. Not
+	 * touched by the after-commit trigger below — a value this rejects never reaches a save at all, let
+	 * alone the online check.
+	 */
+	@ModelChange(timings = { ModelValidator.TYPE_BEFORE_NEW, ModelValidator.TYPE_BEFORE_CHANGE },
+			ifColumnsChanged = I_C_BPartner.COLUMNNAME_VATaxID)
+	public void validateVATaxID(@NonNull final I_C_BPartner bpartner)
+	{
+		if (sysConfigBL.getBooleanValue(VATaxIDValidationUtil.SYSCONFIG_validateVATaxID, true))
```
metas-ts: please see my comment about validateVATaxID of the C_BPartner_Location.java MI

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/interceptor/C_BPartner.java
```diff
@@ -0,0 +1,92 @@
...
+		{
+			VATaxIDValidationUtil.validate(VATIdentifier.ofNullable(bpartner.getVATaxID()));
```
metas-ts, line 72: also: what if this bpartner's org has a config-record that sais isFormatCheckEnabled=false ??

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/interceptor/C_BPartner_Location.java
metas-ts: **Resolution (dev):** You're right, and I've removed the claim. Fixed in https://github.com/metasfresh/metasfresh/commit/209d8c1d0ea.

The javadoc asserted "a value this rejects never reaches a save at all, let alone the online check". That was wrong on two counts, and I checked both rather than just softening the wording:

1. **The framework ordering is fine, but it proves less than I implied.** `BEFORE_*` is ordered before `AFTER_*` within one save, so `validateVATaxID` is always *invoked* first. But that does not mean it *validated* anything — it is wrapped in the `C_BPartner.validateVATaxID` SysConfig check, so with that set to `N` it is a no-op and a malformed value passes straight through to the after-commit path.
2. **Inter-class interceptor ordering is not guaranteed**, so another interceptor's `BEFORE_*` handler could set `VATaxID` after mine already looked at it. I grepped for this: no production interceptor writes the column today (only cucumber step-defs and a REST test util), so there is no current instance — but it is not something the design should lean on.

**Why the code is nonetheless safe** — and this is what the javadoc now says instead of asserting an invariant it doesn't have: `VATaxIDCheckService#check` re-runs the format check itself, and `VATaxIDCheckTrigger` logs the resulting throw instead of propagating it. So a malformed value that reaches the after-commit path produces a skipped check plus a WARN, never a failed save and never a `VATaxID_CheckLog` row claiming a malformed value was checked.

The related point you raised about `VATaxID_Config.IsFormatCheckEnabled` is a genuine gap and I've answered it separately on the `C_BPartner` thread.

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/interceptor/C_BPartner.java
metas-ts: **Resolution (dev):** Applied here too — the same javadoc claim was on both interceptors and both are fixed in https://github.com/metasfresh/metasfresh/commit/209d8c1d0ea. Full reasoning is on the `C_BPartner_Location` thread you pointed at: https://github.com/metasfresh/metasfresh/pull/25501#discussion_r3775317509

Short version: the `BEFORE_*`→`AFTER_*` ordering holds, but it only guarantees `validateVATaxID` was *invoked*, not that it *validated* — it is SysConfig-gated and can be a no-op. The after-commit path no longer claims otherwise, and is safe because `VATaxIDCheckService#check` re-validates and the trigger logs rather than propagates.

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/interceptor/C_BPartner.java
metas-ts: **Resolution (dev):** Real gap, not defended — thanks for catching it. Here is exactly what happens today, and what I propose.

**Today, with the shipped configuration** (`C_BPartner.validateVATaxID` = a single System-level `Y`, no org override — verified by querying `ad_sysconfig`, not inferred):

- an org whose `VATaxID_Config` says `IsFormatCheckEnabled=false` **still cannot save a malformed VAT-ID**. This line reads only the global SysConfig, so the org-level setting has no effect on the save at all. The setting silently does not do what its name says.
- the same setting *is* honoured inside `VATaxIDCheckService#check`, which skips its own format check when the org disabled it. So the two halves of "is the format checked?" disagree.
- the mirror case: were the SysConfig ever set to `N` while an org's config said `true`, the save would accept a malformed value and the after-commit check would throw on it — caught and logged by the trigger, so a skipped check with a WARN, no failed save.

So there are **two independent gates for one business question**, and the interceptor is wired to the wrong one. That is inconsistent by design, not by accident of this PR: `VATaxIDCheckService.CONFIG_DEFAULT_WITHOUT_RECORD` already carries a "Known divergence" note about it, because `VATaxID_Config` did not exist when the save-time gate was written.

**Who owns the fix:** it is task 17 of the plan for this issue — "replace the SysConfig gate", i.e. make the save-time check read the same per-organisation `VATaxID_Config.IsFormatCheckEnabled` the service reads, so both halves resolve one configured value. It was scheduled after the online-check tasks precisely because it changes *save-time* behaviour for every customer, so it wants its own scenarios (org with the config off, org with no config record at all, and the regression that an unconfigured org behaves exactly as today).

**Question for you, since you raised it here:** shall I pull that forward into this PR now, or leave it in task 17 as planned? I'd default to leaving it where it is — it is a behaviour change on a shared core path and deserves its own commit and tests rather than riding along in the after-commit-trigger change — but if you'd rather not merge the inconsistency at all, say so and I'll do it now.

metas-ts: **Resolution (dev):** Decided — **as planned**, i.e. the gate unification stays in task 17 and this PR merges with the inconsistency present and documented.

Recording it here so the decision is visible on the PR rather than only in chat, and so the point cannot get lost between this thread and that task, the plan entry for task 17 now carries:

- the concrete defect statement (org-level `IsFormatCheckEnabled=false` currently has no effect on saves, because the interceptors read the global SysConfig while `VATaxIDCheckService` reads the org config), with the shipped `ad_sysconfig` state that was confirmed by query;
- **three mandatory scenarios**, because "config-then-SysConfig-fallback" can be implemented and still leave this exact defect untested — (1) org with the config off ⇒ a malformed VAT-ID saves, **and this one must fail before the fix**; (2) org with the config on ⇒ rejected as today; (3) org with no config record ⇒ behaves exactly as today;
- an instruction to delete the now-obsolete "Known divergence" note on `VATaxIDCheckService.CONFIG_DEFAULT_WITHOUT_RECORD` in the same commit, since a stale note describing a divergence that no longer exists is its own defect.

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/interceptor/VATaxIDCheckTrigger.java
```diff
@@ -0,0 +1,129 @@
...
+			checkService.check(VATaxIDCheckRequest.builder()
+					.bpartnerId(bpartnerId)
+					.bpartnerLocationId(bpartnerLocationId)
+					.vataxID(vataxID)
+					.adSessionId(adSessionId)
+					.build());
+		}
+		catch (final Exception ex)
+		{
+			// This catch is not optional and not cargo-cult: on the AUTOCOMMIT path — no ambient transaction
```
metas-ts: reduce the comment to just

Without this catch, a throwing online checker — or a
// malformed VAT-ID that only this format re-check rejects — would propagate straight back into
// the caller's save and fail it, even though the row is already committed.

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/interceptor/VATaxIDCheckTrigger.java
```diff
@@ -0,0 +1,129 @@
...
+			// when the triggering save ran, e.g. a save outside any explicit trx — ITrxManager#runAfterCommit
+			// resolves to AutoCommitTrxListenerManager, whose registerListener() calls this Runnable
+			// SYNCHRONOUSLY and wraps-and-rethrows any exception it throws (see
+			// AutoCommitTrxListenerManager#execute). Without this catch, a throwing online checker — or a
+			// malformed VAT-ID that only this format re-check rejects — would propagate straight back into
+			// the caller's save and fail it, even though the row is already committed. On the transactional
+			// path this catch is a no-op safety net: TrxListenerManager#fireAfterCommit already fires
+			// listeners with OnError.LogAndSkip, discarding a thrown exception itself. Either way, the save
+			// must never fail because of what happens here.
+			Loggables.addLog("VAT-ID check for bpartnerId={}, bpartnerLocationId={}, VATaxID={} failed after commit: {}",
```
metas-ts: use Loggables.withLogger(...).addLog(...)

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/interceptor/VATaxIDCheckTrigger.java
metas-ts: **Resolution (dev):** Reduced to exactly those three lines. https://github.com/metasfresh/metasfresh/commit/ac9ebcb5555

Everything else went: the `AutoCommitTrxListenerManager` mechanism walk-through, the note about the transactional path being `OnError.LogAndSkip`, and the closing "the save must never fail" line. The remaining sentence says what the catch prevents, which is the part a reader actually needs at that spot.

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/interceptor/VATaxIDCheckTrigger.java
metas-ts: **Resolution (dev):** Done — one `Loggables` call now instead of a `Loggables.addLog` plus a separate `logger.warn`. https://github.com/metasfresh/metasfresh/commit/ac9ebcb5555

I used **`Loggables.withWarnLoggerToo(logger)`** rather than the literal `withLogger(logger, Level.WARN)`. It is the same thing — `Loggables.java:131-134` delegates straight to `withLogger(logger, Level.WARN)` — but it avoids importing `ch.qos.logback.classic.Level` into this module, which is what `withLogger`'s `Level` parameter is typed as. Say the word if you'd rather have the explicit two-argument form and I'll swap it.

**One consequence worth your call:** the old `logger.warn(..., ex)` passed the exception itself, so the **stack trace** reached the log. The `ILoggable.addLog(String, Object...)` contract has no throwable slot, so the message now carries `ex.getMessage()` only — the reason is still in both sinks, the stack is not. For a swallowed after-commit failure that is exactly the diagnostic you'd want when it turns out to be something other than a format rejection. Want me to keep a stack-trace-carrying `logger.warn` alongside the `Loggables` line, or is the message enough?

Verified compiling: `de.metas.vatid/base`, BUILD SUCCESS, 19 sources.

backend/de.metas.business/src/main/java/de/metas/tax/api/impl/TaxDAO.java
```diff
@@ -393,12 +395,24 @@ private IQueryBuilder<I_C_Tax> getTaxQueryBuilder(@NonNull final TaxQuery taxQue
...
 		final BPartnerId bpartnerId = taxQuery.getBPartnerId();
 
-		final VATIdentifier bpVATaxID = Optional.ofNullable( taxQuery.getBPartnerLocationId())
+		final BPartnerLocationId bpartnerLocationId = Optional.ofNullable(taxQuery.getBPartnerLocationId())
 				.map(BPartnerLocationAndCaptureId::getBpartnerLocationId)
-				.flatMap(bpartnerBL::getVATTaxId)
 				.orElse(null);
 
-		final boolean bPartnerHasTaxCertificate = bpVATaxID != null;
+		final VATIdentifier bpVATaxID = bpartnerLocationId != null
```
metas-ts, line 402: we will need to somehow get the "contractor's" vatax-id - right??

metas-ts: **Resolution (dev):** Not resolved yet — I need one word from you, because "Contractor" doesn't map onto a field and I'd rather ask than guess. Everything below is queried against a customer-faithful database, not inferred.

**No field is called "Contractor".** On the customer's **live** order window `541886` (their override of core `143`), the partner-ish fields are labelled: `C_BPartner_ID` → *Auftraggeber* (and *Lieferung an* on a second tab), `C_BPartner_Location_ID` → *Auftraganschrift* / *Lieferanschrift*, `Bill_BPartner_ID` → *Rechnung an*, `DropShip_BPartner_ID` → *Lieferung an*, `HandOver_Partner_ID` → *Übergabe-Partner*. Nothing is *Auftragnehmer*, and no `AD_Element` in the database carries a Contractor/Auftragnehmer caption on any of those columns. So "Buyer"/"Contractor" is business vocabulary, not window captions — which is exactly why I don't want to map it myself.

Two readings, and they need different answers:

1. **"Contractor" = *Auftragnehmer*, i.e. us, the supplier.** Then its VAT-ID is our own organisation's, and we *already* need it — `VATaxID_Config.RequesterMemberStateCode` + `RequesterNumber`, sent to VIES so it returns the consultation number. Nothing changes at this line: `RequiresTaxCertificate` keys off the **customer's** certificate only, and AC5 doesn't ask us to change that.
2. **"Contractor" = the order's partner + address pair (*Auftraggeber* / *Auftraganschrift*).** Then we already read it, and this line is correct as written: `getVATTaxId` takes the **location's** `VATaxID` when set and falls back to the **partner's** (unless `IgnorePartnerVATID`), and the new `getVATaxIDStatusCode` reads the status off *whichever of those two records supplied the value* — never a different one. That was the whole point of sharing one resolver between them.

**What we do *not* read under either reading** — and this is the case where your instinct would be right: `DropShip_BPartner_ID` (*Lieferung an*) and `HandOver_Partner_ID` (*Übergabe-Partner*). If "Contractor" means the party the goods actually go to rather than the party that ordered, then no, we don't have that VAT-ID, and picking it up would be a real change — squarely AC2/AC9 territory, which is deferred.

So: **which party do you mean?** If it's (1) or (2), this line needs no change. If it's the delivery/handover party, that's new scope and I'd want it written into AC2 rather than slipped in here.

---

Unrelated data-hygiene observation, worth knowing since it bit me while checking the above: `ad_table.ad_window_id` for `C_Order` points at window **143, whose name is `Auftrag_OLD`** — the deprecated one. The usual recipe for "find the live window" therefore returns the `_OLD` window for `C_Order` on this database, and you only reach the real one via `overrides_window_id = 143`. My first label query read `143` for exactly that reason before I caught it. Not a defect in this PR; flagging it because any future field-placement work on the order window will hit the same trap.

metas-ts: ok, then we are good. If this comes from C_Order.C_BPartner, then it's fine with me.

metas-ts: **Resolution (dev):** Confirmed — it does come from `C_Order.C_BPartner`, so no change here. Recording the evidence rather than leaving the condition assumed, since it was the premise of the approval.

`OrderLineBL#setTax` builds its `TaxQuery` with `bPartnerLocationId(bpLocationId)`, where `bpLocationId` comes from `OrderLineDocumentLocationAdapterFactory.locationAdapter(orderLine).getBPartnerLocationAndCaptureId()` — i.e. the order **line's** own `C_BPartner_ID` / `C_BPartner_Location_ID`, which are populated from the order header at header time. `getEffectiveBillPartnerId` is consulted only for the separate `IsTaxExempt` flag, never for this. So the chain is order header → order line → tax query, and `Bill_BPartner_ID` never enters it.

From there, `getVATTaxId` prefers the **location's** `VATaxID` and falls back to the **partner's** (unless the `IgnorePartnerVATID` SysConfig says otherwise), and the new `getVATaxIDStatusCode` shares one private resolver with it — so the status is always read off whichever of those two records supplied the value, never a different one.

For completeness on what is still *not* read, in case it matters later: `DropShip_BPartner_ID` and `HandOver_Partner_ID` are not in the chain. If a future requirement needs the party the goods actually go to rather than the party that ordered, that is new scope and belongs in AC2.

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/interceptor/C_BPartner.java
metas-ts: **Resolution — this is now implemented, not deferred.**

Re-checking this thread while reconciling the PR: the concern here was that an organisation's `VATaxID_Config.IsFormatCheckEnabled` had no effect on the save-time format check, so the save-time gate and the online-check service could disagree. The reply chain recorded it as deferred to a later step, and it is worth stating plainly that the deferral was then closed by actually doing the work rather than by dropping it.

`C_BPartner#validateVATaxID` (and its `C_BPartner_Location` counterpart) now resolve the gate from the same repository the check service uses:

```java
final VATaxIDConfig config = configRepository.getByOrgId(OrgId.ofRepoId(bpartner.getAD_Org_ID()));
if (config.isFormatCheckEnabled())
{
    VATaxIDValidationUtil.validate(VATIdentifier.ofNullable(bpartner.getVATaxID()));
}
```

Three things worth noting for whoever reads this next:

- `VATaxIDConfigRepository#getByOrgId` is now `@NonNull` and is the **only** production reader of `IsFormatCheckEnabled` — a `git grep` over the module finds exactly one, so the save-time gate and the online check cannot diverge by construction rather than by convention.
- The organisation is taken from the record's own `AD_Org_ID`, never from ambient context.
- Absent a config record for the organisation, the value falls back to a SysConfig default, and the config cache is invalidated on `AD_SysConfig` changes as well as on `VATaxID_Config` ones — so flipping the default takes effect without a restart. Covered by unit tests in both directions on distinct organisation ids.

Cucumber coverage for the three cases this thread was about: format check disabled with a malformed value (must save), enabled with a malformed value (must be rejected), and no config record at all (unchanged default-on behaviour) — each mirrored for the address grain.

One process note, since it matters for how this thread reads: the earlier replies here were posted from an automation using the maintainer's own credentials, so the API shows them under the same login as the review comment. That means the visible history cannot by itself distinguish "reviewer accepted the deferral" from "the same actor answered its own question" — worth a human eye if the deferral itself needs confirming. The functional question, though, is closed: the behaviour asked for is in the branch and tested.

metas-ts: **Resolution (dev):** implemented, not deferred — the save-time format check now resolves the gate from the organisation's own `VATaxID_Config` via `VATaxIDConfigRepository#getByOrgId`, which is the single production reader of `IsFormatCheckEnabled`, so the save-time gate and the online check cannot diverge. Full detail, including the SysConfig fallback, cache invalidation on `AD_SysConfig` changes, and the cucumber coverage for all three cases, is in the preceding reply: https://github.com/metasfresh/metasfresh/pull/25501#discussion_r3786846012

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/IBPartnerBL.java
```diff
@@ -304,4 +304,21 @@ public enum IfNotFound
 	 */
 	@NonNull
 	Optional<VATIdentifier> getVATTaxId(@NonNull BPartnerLocationId bpartnerLocationId);
+
+	/**
+	 * Gets the raw {@code VATaxIDStatus} column value of whichever record actually supplied the
```
metas-ts: make this javadoc way shorter. just the important things

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/session/ISessionBL.java
```diff
@@ -15,6 +15,20 @@ public interface ISessionBL extends ISingletonService
...
 
+	/**
+	 * @return the current session's {@code AD_Session_ID}, or {@code null} if there is no current session
+	 * (e.g. a batch/import save with no logged-in user). The common "resolve once, pass down" shape a
+	 * caller needs instead of re-deriving it from {@link #getCurrentSession(Properties)} at every call
+	 * site — a save-time {@code @ModelChange} method resolving the acting user for an audit column is the
+	 * typical caller.
+	 */
+	@Nullable
+	default Integer getCurrentSessionIdOrNull(final Properties ctx)
```
metas-ts: Return a repoidaware SessionId. If we don't have one, create it

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/IBPartnerBL.java
metas-ts: **Resolution (dev):** shortened — https://github.com/metasfresh/metasfresh/commit/c703ae19883

Kept only the two facts a caller can't get from the signature: the status is resolved off the *same* record that supplied `getVATTaxId`'s value, and empty means "no VAT-ID found" **or** "found, but its status column is null" — so empty is not proof there is no VAT-ID. Dropped the rest (14 lines → 8).

This was part of a wider pass in the preceding commit https://github.com/metasfresh/metasfresh/commit/86a1b4cc798: the module's javadoc had reached 19.8% of its lines against 3–8.6% for comparable modules, with 16 blocks of 20+ lines where `de.metas.handlingunits` has 4 across 1103 files. That commit is comment-only (every added and removed line is a comment line) and cuts 443 lines. This file was missed there because the pass scoped to newly-added files and this one is modified — the two `IBPartnerDAO` blocks had the same gap and are now trimmed too.

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/session/ISessionBL.java
metas-ts: **Resolution (dev):** implemented — https://github.com/metasfresh/metasfresh/commit/c703ae19883

New `AdSessionId` (`RepoIdAware`, in `org.adempiere.ad.session`) replaces the bare `Integer`. `getCurrentSessionIdOrNull` is gone; `getCurrentOrCreateSessionId(ctx)` returns a `@NonNull AdSessionId` and delegates to the existing `getCurrentOrCreateNewSession`, so a save with no logged-in session (batch/import) now gets one created instead of yielding null.

Threaded through: both interceptors, `VATaxIDCheckTrigger` (the parameter is now `@NonNull`), `VATaxIDCheckRequest.adSessionId`, and the repository write — where the `-1` "no id" sentinel that the PO layer stores as SQL `NULL` (a `0` would persist as a literal zero) now lives in `AdSessionId.toRepoId` rather than as an inline ternary.

Two things worth flagging:

- **Cost of "create it":** `getCurrentOrCreateNewSession` caches into the context and `s_sessions`, so it creates once per context, not per call — no row-per-save.
- **The cucumber step that establishes a session up front is deliberately kept.** Its original rationale is now obsolete (the capture can no longer be vacuously null), but with the step in place TC9/TC10 still prove the trigger picks up a **pre-existing** session rather than merely creating one of its own — strictly better coverage than dropping it. Its javadoc and the interceptor's are updated where they described the old null-returning behaviour; both scenarios assert `AD_Session_ID | true`, which still holds.

Verified: `de.metas.adempiere.adempiere.base`, `de.metas.business`, `de.metas.vatid` and the cucumber test sources all compile; 159 `de.metas.vatid` unit tests pass. `getCurrentSessionIdOrNull` had no callers outside the two interceptors, so removing it breaks nothing.

PR #25566, opened 2026-08-16
VAT-ID check: refresh the order tax on the path that actually writes the status

metas-ts:
## What this PR does

**The feature:** an open sales order's tax is now refreshed after a VAT-ID check.

`VATaxIDCheckService.check()` — the convergence point of the save-triggered async check and the `C_BPartner_VATaxID_Check` process — never called the order-tax refresher. The only call site sat in the run service behind `newStatus != previousStatus`, compared against a snapshot taken when the run retrieved its targets. So the save-triggered check consumed the transition first, the process then saw `Valid == Valid`, and the refresh was dead in the normal flow.

The refresh now happens where the verdict is written, decided against the record's **stored** status. It fires only when the parent write actually happened — not when the verdict is abandoned because the `VATaxID` changed mid-flight, not when the org has the online check off, not when de-duplication's answer equals the stored status. Only orders with `C_Order.Processed = false` are touched.

Status write and refresh share one transaction (`callInThreadInheritedTrx`), so a refresh failure can never leave a committed status behind stale order tax. On the process path that nests in the existing per-target `callInNewTrx`; on the async path, where the work-package processor runs with no ambient transaction, it is the only thing making the pair atomic.

## Production defects found while building it

**The nightly run could not complete on a real dataset.** It materialised its whole selection as ids and re-queried by that list twice — one bind parameter per due record, unbounded and unchunked — dying with `An I/O error occurred while sending to the backend` before performing a single check. `MaxChecksPerRun` is applied *after* those queries, so it capped VIES calls, never the selection.

Now: per VIES-enabled organisation, one query per grain, due-filtered and ordered `VATaxIDLastAttemptedAt ASC NULLS FIRST` in SQL, walked with a guaranteed iterator (the run writes the very columns that query filters and orders by), 500-record pages, and it stops **reading** once the budget is spent. Zero bind parameters. `pendingCount` comes from a count query, since the run no longer reads to the end.

**A location's check was gated on the wrong organisation.** `CheckTarget` took the org from the parent partner, while the trigger and `VATaxIDParentStatusRepository` both use the location's own `AD_Org_ID`. Where they differed, a location whose own org had the check off could be selected forever without ever advancing `VATaxIDCheckedAt` — the starvation the ordering exists to prevent.

**`TaxDAO#findTaxCategoryId` lost its client scoping.** A cleanup dropped `addOnlyContextClient()`, which was the method's *only* client predicate; the inner `C_Tax` subquery was never scoped either. With eight live callers in pharma product import, `orderBy(Name).firstId()` could return another client's tax category. Restored.

## Also in here

- `VATaxIDCheckRunService` → `VATaxIDMassCheckService` (+ `-Request`/`-Result`): the old names differed by one word carrying the whole single-record-vs-batch distinction. Both class javadocs trimmed to the job plus the load-bearing invariants.
- A redundant `runInThreadInheritedTrx` removed from the order-tax refresher — its caller already guarantees the ambient transaction, so it only added a savepoint nothing resumed from.
- Cucumber: nightly-selection assertions moved from reading an in-memory id list to asserting a real budgeted run against the database, so the SQL ordering is actually exercised; the shared checker mock records unexpected VAT-IDs instead of throwing on a work-package thread (which used to destroy a *different* scenario's data); and the check queue is drained at scenario end plus before the two scenarios that mutate the clock, closing a cross-scenario straggler race.

## Verification

CI green on https://github.com/metasfresh/metasfresh/commit/d1ac7c33009 — run https://github.com/metasfresh/metasfresh/actions/runs/32819740212 — all cucumber profiles, `test (java)`, `test (frontend)`, Merge Gate.

163 `de.metas.vatid` + 17 `de.metas.vatid.vies` unit tests. New unit tests pin the two properties whose absence caused the outage: the nightly path never reaches the id-list overloads, and `MaxChecksPerRun` bounds what is *read*, not only what is checked.

## Reviewer notes

- Deliberate exception to `de.metas.cucumber/CLAUDE.md` rule 7 (drains belong in the consuming step def): two scenarios drain in the `.feature` file, because by the time the consuming step runs they have already mutated the clock and the stub that the pending task will read. Documented as a carve-out on rule 17.
- `VATaxIDCheckService` keeps three thin delegating methods to the config/check repositories. Deleting them as "pass-throughs" was tried and reverted: `docs/coding-rules/service-injection.md` §1 forbids a class holding both a service and that aggregate's repository, and prescribes exactly those delegates.
- Cucumber hooks force glue-class construction for **every** scenario in the suite, and a `@Before` runs before the step that boots Spring. The VAT-ID step-defs therefore resolve their beans lazily; eager field initialisers there take down every scenario in the fork.


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/IBPartnerDAO.java
```diff
@@ -282,24 +282,31 @@ public interface IBPartnerDAO extends ISingletonService
...
+	 * attempt first ({@code VATaxIDLastAttemptedAt} ascending, nulls — never attempted — first).
+	 *
+	 * <p>Due means: non-blank header {@code VATaxID}, and either no conclusive status yet or a
+	 * {@code VATaxIDCheckedAt} older than {@code staleBefore}. Pass {@code staleBefore == null} to mean
+	 * "no staleness window", i.e. every VAT-ID-bearing record of the organisation is due.
 	 *
-	 * <p>Deliberately not narrowed to organisations that enabled the check: the caller resolves each
-	 * partner's own configuration before touching it, so a disabled organisation costs only an in-memory
-	 * re-check — never a call, a write, or another organisation's policy.
+	 * <p>A guaranteed iterator: the caller writes {@code VATaxIDLastAttemptedAt} and
```
metas-ts: i understand "guaranteed iterator", but what am i supposed to make of 

> the very columns this query filters and orders by — on records it has already passed, so the id set must be fixed up front rather than re-read page by page.

?

metas-ts: Fair — that sentence named the mechanism without ever saying what goes wrong, so it was unreadable. Rewritten in https://github.com/metasfresh/metasfresh/commit/3ab2ad15cdb:

> Returns a guaranteed iterator, i.e. one that reads all matching ids once and then loads them in pages. That matters because the caller updates `VATaxIDLastAttemptedAt` and `VATaxIDCheckedAt` while it iterates, and this query filters and sorts on exactly those two columns. A plain iterator re-runs the query for every page, so each update would re-sort the rows it has not read yet, and records would be skipped or returned twice.

metas-ts: **Resolution (dev):** Rewritten in https://github.com/metasfresh/metasfresh/commit/3ab2ad15cdb — the old sentence named the mechanism ("the id set must be fixed up front rather than re-read page by page") but never said what goes wrong without it. It now reads: "Returns a guaranteed iterator, i.e. one that reads all matching ids once and then loads them in pages. That matters because the caller updates VATaxIDLastAttemptedAt and VATaxIDCheckedAt while it iterates, and this query filters and sorts on exactly those two columns. A plain iterator re-runs the query for every page, so each update would re-sort the rows it has not read yet, and records would be skipped or returned twice."

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/impl/BPartnerDAO.java
```diff
@@ -678,31 +683,88 @@ public ImmutableList<I_C_BPartner_Location> retrieveBPartnerLocationsWithVATaxID
...
+	}
+
+	@Override
+	public int countBPartnersDueForVATaxIDCheck(@NonNull final OrgId orgId, @Nullable final Instant staleBefore)
+	{
+		return createBPartnersDueForVATaxIDCheckQuery(orgId, staleBefore).count();
+	}
+
+	@NonNull
+	private IQuery<I_C_BPartner> createBPartnersDueForVATaxIDCheckQuery(@NonNull final OrgId orgId, @Nullable final Instant staleBefore)
```
metas-ts: reaname the parameter `staleBefore` to something like `lastCheckedBefore`

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/bpartner/service/impl/BPartnerDAO.java
```diff
@@ -678,31 +683,88 @@ public ImmutableList<I_C_BPartner_Location> retrieveBPartnerLocationsWithVATaxID
...
-				.create()
-				.listIds(BPartnerId::ofRepoId);
+				.addNotEqualsFilter(I_C_BPartner.COLUMNNAME_VATaxID, "");
+
+		if (staleBefore != null)
+		{
+			// A null VATaxIDStatus counts as never-checked, matching the caller's own resolveStatus.
+			queryBuilder.filter(queryBL.createCompositeQueryFilter(I_C_BPartner.class).setJoinOr()
+					.addEqualsFilter(I_C_BPartner.COLUMNNAME_VATaxIDStatus, null)
+					.addEqualsFilter(I_C_BPartner.COLUMNNAME_VATaxIDStatus, X_C_BPartner.VATAXIDSTATUS_NotChecked)
```
metas-ts: do we have one or more indexes to aid this filtering and avoid a sequential scan?

metas-ts: **Resolution (dev):** No — and you're right that it matters. Checked `pg_indexes` on a migrated DB: the only VAT-ID-related indexes on `C_BPartner` / `C_BPartner_Location` are `*_vataxid_checklog_id_idx` (the FK back to `VATaxID_CheckLog`), which this query never touches. Nothing on `VATaxID`, `VATaxIDStatus`, `VATaxIDCheckedAt`, and **nothing on `VATaxIDLastAttemptedAt`** — so as written this is a sequential scan plus a sort, per organisation, per run.

I did not measure a plan: the only DB I can reach has ~24 VAT-ID-bearing partners, where the planner would pick a seq scan whatever indexes existed, so an `EXPLAIN` there would prove nothing about a real dataset. The absence of the index is the fact; the plan cost is an inference from it.

What I'd add, one per grain — partial, because the overwhelming majority of partners carry no VAT-ID at all, and column-ordered so the `ORDER BY` is satisfied by the index and the sort disappears:

```sql
CREATE INDEX C_BPartner_VATaxID_Due_idx
    ON C_BPartner (AD_Org_ID, VATaxIDLastAttemptedAt, C_BPartner_ID)
 WHERE IsActive = 'Y' AND VATaxID IS NOT NULL AND VATaxID <> '';

CREATE INDEX C_BPartner_Location_VATaxID_Due_idx
    ON C_BPartner_Location (AD_Org_ID, VATaxIDLastAttemptedAt, C_BPartner_Location_ID)
 WHERE IsActive = 'Y' AND VATaxID IS NOT NULL AND VATaxID <> '';
```

Two caveats on that shape, so it isn't taken as more than it is:

- The staleness predicate is a disjunction (`VATaxIDStatus IS NULL OR = 'NotChecked' OR VATaxIDCheckedAt IS NULL OR VATaxIDCheckedAt < ?`), which no single btree can serve. The partial clause is what does the real work — it reduces the candidate set to VAT-ID-bearing rows before that disjunction is evaluated.
- Postgres's default `ASC` puts NULLs last, while the query orders `NULLS FIRST`. So an index in the form above supports the *filter* and the org lookup, but a backwards-compatible index-only ordering needs `VATaxIDLastAttemptedAt ASC NULLS FIRST` spelled out in the index too. Worth pinning that explicitly rather than relying on the default.

This needs a migration script, and the script ID has to come from the ID server rather than being invented — so tell me whether you want it in this PR or as a follow-up, and I'll add it properly.

metas-ts: - what does "grain" mean? are the two grains C_BPartner and C_BPartner_Location? then fine
- if you want to add `AND VATaxID <> ''`, then make sure the whereclause contains this. better, remove the `AND VATaxID <> ''` though
- AD_Org_ID must be last, not first
- it has of course to be in this PR

metas-ts: **Resolution (dev):** All four taken as given; the migration is being added in this PR.

1. Yes — the two are `C_BPartner` and `C_BPartner_Location`, one index each. "grain" was my jargon and it cost you time; dropping it.
2. The `VATaxID <> ''` condition comes out. Worth flagging that this is not only an index change: a partial index's predicate must match the query's WHERE clause or the planner cannot use the index at all, so `addNotEqualsFilter(..., "")` comes out of both query builders too, leaving `VATaxID IS NOT NULL` on both sides. I am verifying first whether a blank can actually persist in those columns or is normalised to NULL on save — if a blank *can* persist, then dropping the filter makes blank-VAT-ID records nightly candidates, and I will report that here rather than let it pass silently.
3. `AD_Org_ID` last: `(VATaxIDLastAttemptedAt, <id column>, AD_Org_ID)`. That is better than what I proposed — leading with the ordering column lets the index serve the `ORDER BY`, so the budget-limited run stops after `MaxChecksPerRun` rows without a sort, which is exactly this query's access pattern.
4. In this PR.

I am also spelling `ASC NULLS FIRST` out explicitly in the index rather than relying on the Postgres default, which is NULLS LAST and would not match the query's ordering — otherwise the sort-free scan is silently lost.

The script ID comes from the ID server. I will post the final DDL and the blank-string finding on this thread once it is in.

metas-ts: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/564834703bb — migration `5820250_sys_VATaxID_DueForCheck_Indexes.sql`, ID from the ID server.

```sql
CREATE INDEX IF NOT EXISTS C_BPartner_VATaxID_DueForCheck_idx
    ON C_BPartner (VATaxIDLastAttemptedAt ASC NULLS FIRST, C_BPartner_ID, AD_Org_ID)
 WHERE VATaxID IS NOT NULL;

CREATE INDEX IF NOT EXISTS C_BPartner_Location_VATaxID_DueForCheck_idx
    ON C_BPartner_Location (VATaxIDLastAttemptedAt ASC NULLS FIRST, C_BPartner_Location_ID, AD_Org_ID)
 WHERE VATaxID IS NOT NULL;
```

`AD_Org_ID` last as instructed, `NULLS FIRST` spelled out, `VATaxID <> ''` gone from both due-for-check builders so the WHERE clause matches the partial predicate. `retrieveBPartnerLocationsWithVATaxID` keeps its filter — not one of these queries. Applied to a local stack; `pg_indexes` confirms both with that column order and `NULLS FIRST` preserved.

**On removing `VATaxID <> ''` — you should know what it exposed, because it is worse than "blank rows become candidates".**

A blank *can* persist: `VATaxID` is not mandatory, and `PO` does no blank-to-null coercion for it (the only such coercion is hard-coded to the `Value` column). `BPartnerCompositeSaver` and `BPartnerImportHelper` both pass the value through guarding only against `null`, so a present-but-blank import column lands as `''`. `VATIdentifier.of` then throws on it, from a call site *outside* the only `try` on that path, and neither `run(...)` nor the process has a catch-all — so **one blank VAT-ID aborted the entire nightly run**, leaving the remaining records and remaining organisations unchecked.

I did not leave that in. `CheckTarget.ofPartner`/`ofLocation` now resolve via `VATIdentifier.ofNullable` and return `null` for a blank; both nightly loops skip it, spending no budget and writing no attempt stamp, after one WARN naming the record. The guard is in Java, so query and index stay in parity at `VATaxID IS NOT NULL`.

This also closes the same hole for whitespace-only values, which was **pre-existing**: `<> ''` never excluded `" "`, and `Check.isEmpty` on the manual path is null-or-length-0 only, so `" "` aborted both paths before today.

One behaviour change follows and cannot be avoided: on the manual-selection path a whitespace-only VAT-ID is now skipped rather than aborting the process. Preserving that abort would mean re-throwing on `null` — code whose only purpose is keeping a crash. For `null` and `''` the manual path behaves exactly as before. The skip is logged on the nightly path only, where the query guarantees `VATaxID IS NOT NULL` so a blank is a genuine data defect; on the manual path a missing VAT-ID is the ordinary case and would log once per partner in the selection.

Test written first and confirmed RED with `Invalid VAT ID` before the fix (blank record streamed ahead of a good one, budget 1). 164 `de.metas.vatid` tests pass.

No `EXPLAIN` offered as evidence — the DB I can reach has ~24 VAT-ID-bearing partners, so it would seq-scan whatever indexes exist.

metas-ts: **Resolution (dev):** Correction to my two replies above — one factual claim in them is wrong, and you should not carry it forward.

I wrote that a partial index's predicate must **match** the query's WHERE clause or the planner cannot use the index. It must be **implied** by it. `VATaxID IS NOT NULL AND VATaxID <> ''` implies `VATaxID IS NOT NULL`, so keeping `<> ''` would never have disabled either index. Removing it was your instruction and it is a fine simplification; it was not an index fix, and I presented it as one.

What that removal did expose is real and unchanged: it made blank-VAT-ID rows nightly candidates, and `VATIdentifier.of` threw on them from outside the only `try` on that path, so one blank aborted the whole run. That is fixed as described, plus the pre-existing whitespace-only case.

Also, for reading the diff: the two due-for-check queries and `CheckTarget` no longer live in `IBPartnerDAO`/`BPartnerDAO`. Your cleanup commit https://github.com/metasfresh/metasfresh/commit/5436a7af754 moved them to `de.metas.vatid.VATaxIdCheckTargetRepo`, so this thread's file no longer contains the code it is anchored to. The migration itself is unchanged from the DDL posted above.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/vatid/VATaxIDCheckProcess_StepDef.java
```diff
@@ -170,6 +181,28 @@ public void runProcessForSelection(@NonNull final String maxChecksPerRunText, @N
...
+	 * The scheduled run with {@code MaxChecksPerRun} pinned — the only way to assert that the nightly
+	 * ORDER BY actually decides WHICH records a budgeted run reaches, rather than merely what a full sweep
+	 * happens to contain.
+	 *
+	 * @cucumber.stepdef
+	 * @cucumber.example
+	 * <pre>
+	 * When the C_BPartner_VATaxID_Check process is run as scheduled with MaxChecksPerRun '1'
+	 * </pre>
+	 */
```
metas-ts:
```suggestion
```

useless javadoc

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/vatid/VATaxIDCheckProcess_StepDef.java
```diff
@@ -200,6 +237,72 @@ public void runProcessAsScheduled()
...
+	 * @cucumber.columns
+	 *   <b>C_BPartner_ID</b> — (required, identifier-ref) partner to leave untouched, together with its locations
+	 * @cucumber.depends StepDefData: C_BPartner_StepDefData
+	 * @cucumber.example
+	 * <pre>
+	 * Given every other VAT-ID check candidate has already been attempted, except C_BPartner:
+	 *   | C_BPartner_ID |
+	 *   | bp_broken     |
+	 *   | bp_pending    |
+	 * </pre>
```
metas-ts, line 246:
```suggestion
	 * Stamps {@code VATaxIDLastAttemptedAt} on every not-yet-attempted VAT-ID check candidate of the test
	 * organisation EXCEPT the partners named in the data table (and their locations), so that afterwards the
	 * named records are the only null-attempted ones left.
	 *
	 * <p>Why: this makes a budgeted {@link #runProcessAsScheduledWithBudget(String)} deterministic.
	 *
	 * <p>Covers both {@code C_BPartner_Location} and {@code C_BPartner} carrying a VAT-ID.
	 * <p>Only records the nightly query would actually select are touched
	 *
	 * @cucumber.stepdef
	 * @cucumber.columns
	 *   <b>C_BPartner_ID</b> — (required, identifier-ref) partner to leave untouched, together with its locations
	 * @cucumber.depends StepDefData: C_BPartner_StepDefData
	 * @cucumber.example
	 * <pre>
	 * Given every other VAT-ID check candidate has already been attempted, except C_BPartner:
	 *   | C_BPartner_ID |
	 *   | bp_broken     |
	 *   | bp_pending    |
	 * </pre>
```

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/vatid/VATaxIDCheckProcess_StepDef.java
metas-ts: **Resolution (dev):** Removed in https://github.com/metasfresh/metasfresh/commit/639e2871a7c. The suggestion was anchored on the closing `*/` only, so applying it literally would have left a dangling `*/`; removed the whole javadoc block, which is the evident intent.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/vatid/VATaxIDCheckProcess_StepDef.java
metas-ts: **Resolution (dev):** Applied verbatim in https://github.com/metasfresh/metasfresh/commit/639e2871a7c.

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/VATaxIDMassCheckService.java
```diff
@@ -0,0 +1,419 @@
...
+	 * <p>Deliberately {@code runInNewTrx} rather than {@code runInThreadInheritedTrx} — the one genuinely
+	 * required use of the hack {@code docs/coding-rules/java-general.md} otherwise warns against. Joining the
+	 * check-and-refresh transaction would let a rolled-back check erase the evidence that an attempt was
+	 * made, which is the whole defect this mechanism prevents.
+	 *
+	 * <p>A chronic failure of this write itself is an accepted residual risk: such a target sorts first every
+	 * run but consumes only its own slot, so it crowds out nobody. Diagnose it from the run log rather than
+	 * auto-excluding after N failures.
+	 */
+	private void stampAttemptInOwnTrx(@NonNull final CheckTarget checkTarget)
```
metas-ts: this code belongs also into VATaxIdCheckTargetRepo!

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/VATaxIDMassCheckService.java
```diff
@@ -0,0 +1,419 @@
...
+	 *
+	 * <p>Two {@code catch} blocks, not one: a stamp failure means the check was never attempted and leaves no
+	 * other trace, while a failure of the check itself means the attempt is already durably recorded. One
+	 * shared message would hide the rarer signal.
+	 *
+	 * @throws VATaxIDCheckRequestRejectedException when the checking service rejected the request itself
+	 * because of how this system is configured. Deliberately NOT swallowed here, unlike every other failure:
+	 * it would repeat identically for every remaining target, so {@link #run} aborts on it instead.
+	 */
+	private void checkOneInOwnTrx(@Nullable final PInstanceId pinstanceId, @NonNull final CheckTarget checkTarget)
```
metas-ts, line 334: make the javadoc shorter and more concise

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/process/C_BPartner_VATaxID_Check.java
```diff
@@ -82,35 +80,25 @@ public ProcessPreconditionsResolution checkPreconditionsApplicable(@NonNull fina
...
+		// user-triggered run passes its selection as a lazily-streamed query rather than a materialised id list:
+		// "select all" can carry tens of thousands of records, and binding one parameter per record is what
+		// produced `An I/O error occurred while sending to the backend`. retrieveSelectedRecordsQueryBuilder
+		// covers BOTH the multi-record pinstance selection and the single-record case, so the query preserves
+		// the exact selection semantics of either.
+		final IQuery<I_C_BPartner> selectedBPartnersQuery = nightlyRun
+				? null
+				: retrieveSelectedRecordsQueryBuilder(I_C_BPartner.class)
+						.orderBy(I_C_BPartner.COLUMNNAME_C_BPartner_ID)
+						.create();
```
metas-ts: call `.createSelection(PInstanceId)` and then pass that selection's PInstanceId rather than the query

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/VATaxIDMassCheckService.java
metas-ts: **Resolution (dev):** Moved in https://github.com/metasfresh/metasfresh/commit/7db3aae666e. The own-transaction attempt-stamp orchestration is now `VATaxIdCheckTargetRepo.stampVATaxIDCheckAttemptInOwnTrx(CheckTarget)` — the `SystemTime` read, the `runInNewTrx`, and the partner-vs-location branch sit next to the two `stampVATaxIDCheckAttempt` writes they drive; the service just calls it. The repo gained an `ITrxManager` field for it. The service keeps its own `trxManager` only for `callInNewTrx(checkService.check(...))`, which isolates each target's check-and-refresh unit — business logic, not persistence. 169 `de.metas.vatid` tests still pass.

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/VATaxIDMassCheckService.java
metas-ts: **Resolution (dev):** Shortened in https://github.com/metasfresh/metasfresh/commit/4e294bcb82f — down to the four load-bearing facts (stamp commits before the check; each check runs in its own `callInNewTrx` incl. the tax refresh; the two catches distinguish stamp-fail from check-fail; the rejected-request abort), the rest cut.

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/process/C_BPartner_VATaxID_Check.java
metas-ts: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/4e294bcb82f, and you're right it's the cleaner shape.

The process now materialises the selection into a `T_Selection` keyed by its own pinstance — `retrieveSelectedRecordsQueryBuilder(...).create().createSelection(getPinstanceId())`, the same idiom `CreateMissingReferenceNo` / `M_ShipmentSchedule_ChangeExportStatus` use — and the request carries only that `PInstanceId`; the live `IQuery` field is gone from the DTO. The repo streams with `setOnlySelection(selectionId)` (partner grain) and an `addInSubQueryFilter` over a `setOnlySelection` sub-select (location grain). Still zero bind parameters — `createSelection` writes `T_Selection` server-side — so "select all" can't rebuild the `IN`-list crash. Count and stream still share one query per grain, so the pendingCount stays drift-free. 169 `de.metas.vatid` tests pass (the repo test now materialises via `createSelection` and asserts count == stream over the real selection).

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/vatid/VATaxIDOnlineChecker_StepDef.java
```diff
@@ -63,14 +68,34 @@
  */
 public class VATaxIDOnlineChecker_StepDef
 {
+	/**
+	 * Resolved on first use, NOT in a field initialiser. This class carries a {@code @Before} hook, so cucumber
+	 * instantiates it for every scenario in the suite, and a {@code @Before} runs before the
+	 * {@code Given infrastructure and metasfresh are running} step that brings Spring up — anything resolved at
+	 * construction time would therefore fail on every scenario. Constructing this class must stay free.
```
metas-ts:
```suggestion
	 * Resolved on first use, to not inferfere with this class's {@code @Before} hook.
	 * This instance is created {@link VATaxIDTestServiceConfiguration}.
```

metas-ts: **Resolution (dev):** Applied in https://github.com/metasfresh/metasfresh/commit/4218e8bfc1e. Two small fixes to the suggestion so it doesn't ship a typo: `inferfere` → `interfere`, and `created {@link ...}` → `created in {@link ...}`.

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/VATaxIDCheckService.java
```diff
@@ -168,14 +266,15 @@ public VATaxIDStatus check(@NonNull final VATaxIDCheckRequest request)
 	 * <p>The comparison lives here rather than inside
 	 * {@link VATaxIDParentStatusRepository#updateParentStatus(VATaxIDCheckRequest, VATaxIDLastCheck)}
 	 * because it is a decision, and a repository must not make conditional writes based on business state
-	 * ({@code docs/coding-rules/architecture.md} §8). It also has to be visible here: the caller must learn
-	 * that nothing was stored, so that neither the status-change log line nor
-	 * {@code VATaxIDCheckRunService}'s order-line-tax refresh fires on a verdict that was never written.
+	 * ({@code docs/coding-rules/architecture.md} §8). It also has to be visible to the caller: neither the
```
metas-ts: the javadoc of this method needs to be **way** shorter

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/VATaxIdCheckTargetRepo.java
```diff
@@ -0,0 +1,604 @@
...
+	}
+
+	/**
+	 * Stamps {@code C_BPartner.VATaxIDLastAttemptedAt} unconditionally, whether the check that follows
+	 * succeeds, fails or throws — unlike {@code VATaxIDCheckedAt}, which advances only on a completed check.
+	 * Without it a permanently failing target would never advance any timestamp and would sort first of
+	 * every nightly run forever. The caller must commit this in its own transaction so it survives the
+	 * check's rollback.
+	 */
+	public void stampVATaxIDCheckAttempt(@NonNull final BPartnerId bpartnerId, @NonNull final Instant attemptedAt)
```
metas-ts: *every* public Stamp* method needs to take CheckTarget isntances, no Bpartner and BPArtner-location-Ids

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/VATaxIDCheckService.java
metas-ts: **Resolution (dev):** Shortened in https://github.com/metasfresh/metasfresh/commit/b2fbc24f944 — from ~35 lines to the essentials: what it does, the async enqueue-vs-process race and the harm it prevents (a stale verdict branding the record), the `VATaxID_CheckLog`-row-is-still-kept caveat, and why the decision lives in the service not the repository (§8). The residual-risk and sync-path paragraphs are gone.

backend/de.metas.vatid/base/src/main/java/de/metas/vatid/VATaxIdCheckTargetRepo.java
metas-ts: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/b2fbc24f944. The two `stampVATaxIDCheckAttempt(BPartnerId|BPartnerLocationId, Instant)` overloads collapse into one `stampVATaxIDCheckAttempt(CheckTarget, Instant)` that branches internally; `stampVATaxIDCheckAttemptInOwnTrx(CheckTarget)` just wraps it in `runInNewTrx`. So every public `stamp*` method now takes a `CheckTarget`.

The only non-test caller of the id-based methods was `VATaxIDCheckProcess_StepDef`, which used them to seed "already attempted" records — test data setup, so it now sets `VATaxIDLastAttemptedAt` on the record directly (setter + `saveRecord`) rather than reaching into the repo's production API. That also drops the step-def's now-unused `VATaxIdCheckTargetRepo` injection. Unit tests stub/verify by `CheckTarget` (`argThat(ct -> …equals(ct.getBpartnerId()))`); the chronic-stamp-failure test still drives a throwing write through the spy. 169 `de.metas.vatid` tests pass.

PR #25569, opened 2026-08-17
MobileUI manufacturing receipt — configurable receive targets, skip-target step and catch-weight capture

mrie04:
Adds four manufacturing-profile configuration flags that let a deployment simplify the mobile production-receipt flow. Every flag defaults to today's behaviour, so existing deployments are unaffected.

## The flags

| Flag | Default | Effect when switched away from the default |
|---|---|---|
| `IsAllowReceiveToLU` | `Y` | `N` — no load-unit (pallet) targets are offered |
| `IsAllowReceiveToTU` | `Y` | `N` — no transport-unit targets are offered |
| `IsSkipReceiveTargetStep` | `N` | `Y` — the new-Gebinde / scan-existing chooser is bypassed; the receive-target button goes straight to the packing-instruction list |
| `IsCaptureCatchWeightAtReceipt` | `Y` | `N` — no catch-weight field at receipt; the nominal quantity alone is captured |

Each flag exists twice, mirroring the established `MobileUIManufacturingConfig` pattern: on `MobileUI_MFG_Config` as a global `YesNo`, and on `MobileUI_UserProfile_MFG` as a `_YesNo` reference list where blank means "do not override the global". Merge precedence is user profile → global → built-in default.

## Scoping — main finished good only

All four effects are gated on `line.getCoProductBOMLineId() == null`. Co-/by-product lines are deliberately untouched: they keep their (possibly infinite-capacity) transport-unit targets, their chooser, and their catch-weight prompt. An infinite-capacity catch-weight co-product takes its *quantity* from the weight, so removing either would leave it unreceivable.

## `emptyReason` contract

A target list emptied **by configuration** is returned empty **without** an `emptyReason`, because the frontend renders that reason as the operator-facing "no receiving Gebinde" error — showing it next to perfectly good targets of the other structure would be wrong. When *both* structures are excluded, one list carries the existing localized reason, so the operator still gets the existing guidance rather than a blank screen. `JsonNewLUTargetsList.emptyWithoutReason()` names that first case at the call site.

## Change inventory

- migration: the four columns on both tables, plus `AD_Element`s (en_US / de_DE / de_CH), `AD_Field`s and `AD_UI_Element`s on the global configuration window and the user-profile tab
- regenerated model classes for both tables
- `MobileUIManufacturingConfig` / `…Repository`: four `OptionalBoolean` fields, effective getters, merge fallback, and the matching write path
- `MaterialReceiptActivityHandler.toJson`: the gating described above
- `JsonFinishedGoodsReceiveLine.skipReceiveTargetStep` (nullable; absent means today's behaviour)
- mobile frontend: skip-mode routing, and a skip-mode back target so the bypassed chooser stays unreachable
- frontend-testing masterdata API accepts the four flags
- Playwright coverage for the new behaviour, including a co-product negative control

## Merge ordering

A paired migration in the customer repo sets a deployment-specific profile using these columns. **This PR must merge first** — otherwise that migration runs against a base image without the columns and fails with `column "isallowreceivetolu" does not exist`. That failure is intentional and must not be guarded around: a customer migration must never create or alter a core-owned column.


backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5819440_sys_mobileMfgReceiveTargetFlags.sql
```diff
@@ -0,0 +1,440 @@
...
+;
+
+-- 2026-08-17 09:01:13 (de_DE override)
+UPDATE AD_Element_Trl SET Name='Fanggewicht beim Wareneingang erfassen', Description='Fanggewicht eines Fanggewicht-Produkts beim Produktions-Wareneingang erfassen.', Help='Wenn gesetzt (Standard), wird bei einem Fanggewicht-Produkt das Fanggewicht beim Produktions-Wareneingang abgefragt. Deaktivieren, damit beim Wareneingang kein Fanggewicht erfasst wird – es wird nur die nominale Menge eingegeben – und das Gewicht später erfasst wird (z. B. bei der Kommissionierung). Gilt nur für die Zeile des Hauptfertigprodukts; Kuppel- und Nebenprodukte (z. B. in ein Gebinde mit unbegrenzter Kapazität, bei dem das Fanggewicht die Menge ist) behalten die Fanggewicht-Erfassung immer.', IsTranslated='Y', Updated=TO_TIMESTAMP('2026-08-17 09:01:13.000000','YYYY-MM-DD HH24:MI:SS.US')::timestamp without time zone AT TIME ZONE 'UTC', UpdatedBy=100 WHERE AD_Element_ID=585308 AND AD_Language='de_DE'
```
mrie04: In german usually also Catch Weight is used and not Fanggewicht

backend/de.metas.manufacturing.rest-api/src/main/java/de/metas/manufacturing/workflows_api/activity_handlers/receive/json/JsonFinishedGoodsReceiveLine.java
```diff
@@ -19,6 +19,7 @@ public class JsonFinishedGoodsReceiveLine
 	@NonNull String id;
 
 	boolean coproduct;
+	@Nullable Boolean skipReceiveTargetStep;
```
mrie04: how this can be null?

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5819440_sys_mobileMfgReceiveTargetFlags.sql
mrie04: **Resolution (dev):** Fixed — swapped to the loanword. Verified your point against the DB rather than assuming: in `AD_Element_Trl` for `de_DE`, "Catch Weight" appears in several labels ("Catch Weight", "Abw. Catch Weight Menge", "TU Catch weight erlauben") while "Fanggewicht" appeared in exactly one — this new one. So it was the outlier.

All 28 German occurrences in the migration now use the loanword, across name/description/help for both `de_DE` and `de_CH`: the label reads `Catch Weight beim Wareneingang erfassen`, and compounds became `Catch-Weight-Produkt` / `Catch-Weight-Erfassung`. English text is untouched. Commit: https://github.com/metasfresh/metasfresh/commit/ea8a8c8add8

backend/de.metas.manufacturing.rest-api/src/main/java/de/metas/manufacturing/workflows_api/activity_handlers/receive/json/JsonFinishedGoodsReceiveLine.java
mrie04: **Resolution (dev):** It could only be null for a **co-product line**: the handler set it to `null` when `coProductBOMLineId != null`, so a co-product line's JSON stayed byte-identical to before this change. For the main finished good it was always populated.

On reflection that does not justify a nullable type, so it is now a primitive `boolean`. `false` is not a placeholder for a co-product — it is semantically correct, since a co-product genuinely does not skip the receive-target step. It also matches the sibling `coproduct` field on this same DTO, which is already a primitive. Handler now reads `.skipReceiveTargetStep(isMainFinishedGood && config.getIsSkipReceiveTargetStepEffective())`.

No behaviour change: the frontend treated absent and `false` identically (falsy). Compiles, and the module's 91 unit tests pass. Commit: https://github.com/metasfresh/metasfresh/commit/ea8a8c8add8

PR #25578, opened 2026-08-19
Product Import: accept all valid ProductType codes (fix Invalid ProductType for F/O/N)

metas-rc:
## Problem
Importing a product via `I_Product` with `ProductType = 'F'` (freight cost) fails with `ERR=Invalid ProductType`. The same happens for `'O'` (Online) and `'N'` (Food) — even though all three are valid `ProductType` reference-list codes and are assignable on `M_Product`.

## Root cause
`MProductImportTableSqlUpdater.dbUpdateErrorMessages()` hardcoded the accepted set as `ProductType NOT IN ('E','I','R','S')`, which had drifted from the 7-value `de.metas.product.ProductType` enum (`I, S, R, E, O, F, N`). A sibling method in the same class already enumerates all seven.

## Fix
Build the allowed-`ProductType` `IN (...)` list from `ProductType.values()`, so the import's accepted set is always the full `ProductType` reference list and can never drift from the enum again. Codes are single-character literals — no injection concern. Genuinely invalid codes (not in the reference list) are still rejected.

## Verification
Manual: import a product with `ProductType = 'F'` via the Import Product window — it now imports as a freight-cost product instead of erroring; a code outside the reference list is still rejected.

### DIFF
--- a/backend/de.metas.business/src/main/java/de/metas/product/impexp/MProductImportTableSqlUpdater.java
+++ b/backend/de.metas.business/src/main/java/de/metas/product/impexp/MProductImportTableSqlUpdater.java
@@ -57,8 +57,7 @@ public class MProductImportTableSqlUpdater
 	/**
 	 * SQL {@code IN (...)} list of every valid {@link ProductType} code (e.g. {@code 'E','F','I',...}).
 	 * Derived from the enum so the import's accepted set can never drift from the ProductType reference
-	 * list (the drift that let 'F'/'O'/'N' be wrongly rejected). Codes are single-char literals — no
-	 * injection concern.
+	 * list. Codes are single-char literals — no injection concern.
 	 */
 	private static final String VALID_PRODUCT_TYPE_CODES_SQL = Arrays.stream(ProductType.values())
 			.map(ProductType::getCode)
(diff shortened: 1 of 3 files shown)


backend/de.metas.business/src/main/java/de/metas/product/impexp/MProductImportTableSqlUpdater.java
```diff
@@ -51,6 +54,16 @@ public class MProductImportTableSqlUpdater
 {
 	private static final Logger logger = LogManager.getLogger(MProductImportTableSqlUpdater.class);
 
+	/**
+	 * SQL {@code IN (...)} list of every valid {@link ProductType} code (e.g. {@code 'E','F','I',...}).
+	 * Derived from the enum so the import's accepted set can never drift from the ProductType reference
+	 * list. Codes are single-char literals — no injection concern.
+	 */
+	private static final String VALID_PRODUCT_TYPE_CODES_SQL = Arrays.stream(ProductType.values())
```
mrie04: Consider introducing ProductType.getCodesAsSqlList()

mrie04 (approved):

PR #25608, opened 2026-08-25
Cucumber: cover the WorkStation hop of the lot-number line code

teosarca:
## What

Adds cucumber coverage for the WorkStation hop of the lot-number line code, and the two step-definition inputs such a test needs.

## Why

A lot-number provider resolves the production line as `PP_Order.WorkStation_ID` -> `S_Resource.LotNumberCode`. That hop had **no automated coverage at any layer**.

The existing provider scenario in `lotNumberSequenceProvider.feature` could not cover it, for two independent reasons:

1. `create PP_Order:` had no way to express a workstation at all — the step definition read `S_Resource_ID` (the plant) and nothing else.
2. Its test function hardcodes the line segment and ignores the order:

   ```sql
   CREATE OR REPLACE FUNCTION test_lotno_provider(p_pp_order_id numeric, p_at timestamptz)
     RETURNS text LANGUAGE sql AS
   $$ SELECT 'L' || to_char(p_at AT TIME ZONE 'UTC', 'DDD') || 'M2' $$
   ```

   `p_pp_order_id` is accepted and never read, so that scenario passes unchanged even if the workstation lookup is removed outright.

## Changes

- **`PP_Order_StepDef`** — `create PP_Order:` accepts an optional `OPT.WorkStation_ID.Identifier`, resolved against the resource table and passed to the existing `PPOrderCreateRequest.workstationId`. No behaviour change for tables that omit it.
- **`S_Resource_StepDef`** — `create S_Resource:` accepts an optional `LotNumberCode`.
- **New scenario** in `lotNumberSequenceProvider.feature`.

## Why the new scenario is discriminating

Two manufacturing orders share **one plant** and differ **only** by workstation, carrying `LotNumberCode` `5` and `7`. The provider function actually reads `p_pp_order_id` and joins `PP_Order.WorkStation_ID` -> `S_Resource.LotNumberCode`, and each received HU is asserted to carry its own line digit (`L0915` and `L0917`).

An implementation that hardcoded the digit, or that resolved the line from the **plant** instead of the workstation, cannot satisfy both assertions at once. That is the property the existing scenario lacks.

## Scope

Test-only. No production code is touched, and the two step-definition additions are optional columns, so every existing feature file keeps working unchanged.

### DIFF
--- a/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pporder/lotNumberSequenceProvider.feature
+++ b/backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pporder/lotNumberSequenceProvider.feature
@@ -159,24 +159,36 @@ Feature: Lot number stamped on HU during manufacturing receipt
       | PP_Product_BOM_ID.Identifier | OPT.LotNo_Sequence_ID.Identifier |
       | bom_ws                       | seq_lotno_ws                     |
 
-    # Both orders sit on the SAME plant (testResource) and differ ONLY by workstation.
+    # The first two orders sit on the SAME plant (testResource) and differ ONLY by workstation.
+    # The third runs on line 5 as well, but is received as TWO HUs (QtyTU=2) to cover the multi-HU receipt.
     And create PP_Order:
-      | PP_Order_ID.Identifier | DocBaseType | M_Product_ID.Identifier | QtyEntered | S_Resource_ID.Identifier | OPT.WorkStation_ID.Identifier | DateOrdered             | DatePromised            | DateStartSchedule       | completeDocument |
-      | ppOrder_lotno_ws5      | MOP         | finishedGoodsProd       | 10         | testResource             | wsLineFive                    | 2025-04-01T23:59:00.00Z | 2025-04-01T23:59:00.00Z | 2025-04-01T23:59:00.00Z | Y                |
-      | ppOrder_lotno_ws7      | MOP         | finishedGoodsProd       | 10         | testResource             | wsLineSeven                   | 2025-04-01T23:59:00.00Z | 2025-04-01T23:59:00.00Z | 2025-04-01T23:59:00.00Z | Y                |
+      | PP_Order_ID.Identifier  | DocBaseType | M_Product_ID.Identifier | QtyEntered | S_Resource_ID.Identifier | OPT.WorkStation_ID.Identifier | DateOrdered             | DatePromised            | DateStartSchedule       | completeDocument |
+      | ppOrder_lotno_ws5       | MOP         | finishedGoodsProd       | 10         | testResource             | wsLineFive                    | 2025-04-01T23:59:00.00Z | 2025-04-01T23:59:00.00Z | 2025-04-01T23:59:00.00Z | Y                |
+      | ppOrder_lotno_ws7       | MOP         | finishedGoodsProd       | 10         | testResource             | wsLineSeven                   | 2025-04-01T23:59:00.00Z | 2025-04-01T23:59:00.00Z | 2025-04-01T23:59:00.00Z | Y                |
+      | ppOrder_lotno_ws5_multi | MOP         | finishedGoodsProd       | 20         | testResource             | wsLineFive                    | 2025-04-01T23:59:00.00Z | 2025-04-01T23:59:00.00Z | 2025-04-01T23:59:00.00Z | Y                |
 
+    # The third row receives TWO HUs from ONE receipt: M_HU_ID.Identifier takes two comma-separated
+    # identifiers, bound positionally to the received HUs.
     And receive HUs for PP_Order with M_HU_LUTU_Configuration:
-      | PP_Order_ID       | M_HU_ID.Identifier | IsInfiniteQtyLU | QtyLU | IsInfiniteQtyTU | QtyTU | IsInfiniteQtyCU | QtyCUsPerTU | M_HU_PI_Item_Product_ID.Identifier |
-      | ppOrder_lotno_ws5 | hu_lotno_ws5       | N               | 0     | N               | 1     | N               | 10          | lotNoProduct_HUPI                  |
-      | ppOrder_lotno_ws7 | hu_lotno_ws7       | N               | 0     | N               | 1     | N               | 10          | lotNoProduct_HUPI                  |
+      | PP_Order_ID             | M_HU_ID.Identifier                  | IsInfiniteQtyLU | QtyLU | IsInfiniteQtyTU | QtyTU | IsInfiniteQtyCU | QtyCUsPerTU | M_HU_PI_Item_Product_ID.Identifier |
+      | ppOrder_lotno_ws5       | hu_lotno_ws5                        | N               | 0     | N               | 1     | N               | 10          | lotNoProduct_HUPI                  |
+      | ppOrder_lotno_ws7       | hu_lotno_ws7                        | N               | 0     | N               | 1     | N               | 10          | lotNoProduct_HUPI                  |
+      | ppOrder_lotno_ws5_multi | hu_lotno_ws5_multi_a,hu_lotno_ws5_multi_b | N         | 0     | N               | 2     | N               | 10          | lotNoProduct_HUPI                  |
 
     When complete planning for PP_Order:
-      | PP_Order_ID.Identifier |
-      | ppOrder_lotno_ws5      |
-      | ppOrder_lotno_ws7      |
+      | PP_Order_ID.Identifier  |
+      | ppOrder_lotno_ws5       |
+      | ppOrder_lotno_ws7       |
+      | ppOrder_lotno_ws5_multi |
 
     # Each HU must carry ITS OWN workstation's line digit -- 5 and 7, not one shared value.
+    # And BOTH HUs of the multi-HU receipt must be stamped, with the SAME lot number: the provider is
+    # invoked once per receipt (AbstractPPOrderReceiptHUProducer memoises the value), not once per HU.
+    # Note there is deliberately NO running counter to assert: DBFunctionSequenceNoProvider returns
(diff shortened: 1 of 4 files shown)


teosarca:
## CI triage — the one red check is a known pre-existing flake, not this PR

`mobile test` is the only non-green check. It is **not caused by this diff** and is already tracked as a known flaky test.

### What fails

Build `5.175-soft-panda-hotfix-LotNoWorkstationCoverage.42424` — mobile-webui **247 passed / 1 failed**.

Verified via https://test-reports.metasfresh.com/branches/soft-panda-hotfix-lotnoworkstationcoverage/builds/5.175-soft-panda-hotfix-LotNoWorkstationCoverage.42424/failures.json (suite totals and the failure entry below are read straight from it):

| | |
|---|---|
| Spec | `spec/humanager/huManager.spec.js:155` — `Bulk actions - Move` |
| Error | `TimeoutError: locator.waitFor: Timeout 40000ms exceeded` waiting for `#ApplicationsListScreen` |
| Sibling spec with the identical signature | `spec/humanager/by_ExternalBarcode_attribute.spec.js:123` — same test name, same seam |

Every other suite is green: cucumber 990/990, frontend-webui 121 passed / 6 skipped, backend JUnit 8989 passed, camel 239, jest 362.

### Why it is a flake and not a regression

Four attempts were run on the **single unchanged commit** https://github.com/metasfresh/metasfresh/commit/b86516fdb6b885f114255af545e1002a4dcb7f7c, and the failure **alternates between the two sibling specs** — exactly one of them fails per attempt, and each one has an observed pass on the very commit that fails it. That same-commit flip is the nondeterminism signature; a real breakage cannot pass and fail on identical code.

| Attempt | Job | `huManager.spec.js:155` | `by_ExternalBarcode_attribute.spec.js:123` |
|---|---|---|---|
| 1 | https://github.com/metasfresh/metasfresh/actions/runs/32840890910/job/97786102556 | **pass** (1.7s) | fail |
| 2 | https://github.com/metasfresh/metasfresh/actions/runs/32840890910/job/97796954824 | fail | **pass** |
| 3 | https://github.com/metasfresh/metasfresh/actions/runs/32840890910/job/97810231543 | **pass** (1.9s) | fail |
| 4 | https://github.com/metasfresh/metasfresh/actions/runs/32840890910/job/97820786006 | fail | **pass** |

Always 247/248, always the same `#ApplicationsListScreen` timeout. The alternation points at a shared timing-dependent resource rather than two independent defects.

### Why this diff cannot be the cause

This PR adds cucumber coverage only — a feature file plus step-definition support for the manufacturing-order workstation field. It touches nothing under `misc/services/mobile-webui`, and the failing spec exercises the HU-Manager mobile bulk-move flow, which shares no code path with it. The same failure is also present on the base branch with no PR branch involved.

**The scenario this PR actually adds ran and passed** — cucumber is 990/990 green.

### Suspected mechanism, and tracking

**Hypothesis (unverified) — mechanism: not yet determined at runtime.** The shape of the failure is consistent with a dropped-first-scan race: the keyboard-scanner `keydown` listener attaches in a `useEffect`, i.e. after the React commit, so a scan issued immediately after mount could be dropped — the target would then never be selected and `#ApplicationsListScreen` never appear, so the 40s `waitFor` expires. This rests on reading the scanner component and on the failure signature matching it; **no run has been captured that actually shows the listener attaching late**, so treat it as a lead, not a diagnosis. What *is* verified is only the nondeterminism (the table above). On present evidence it looks like a test-harness timing seam rather than a product defect, but that classification is part of the same unverified hypothesis.

**Verified (observation, not mechanism):** the same `#ApplicationsListScreen` timeout is not confined to this screen — it has also been observed on `by_ExternalBarcode_attribute.spec.js:123` (see the table above) and on `home_screen.spec.js:83` (`Scan Workplace code`) on a separate run. All three are scan steps. That breadth is why a per-screen re-scan loop would mitigate rather than close it, and why any durable fix would have to sit in the shared scanner component — a design decision, and out of scope for this PR either way.

Both specs are recorded in the team's flaky-test tracker with the evidence above; the re-run budget on this commit is spent (4 attempts). Per the flaky-test policy, a **test-only** PR whose sole red is a triaged known flake may be integrated without full green — that call is the integrator's.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pporder/lotNumberSequenceProvider.feature
```diff
@@ -116,3 +116,67 @@ Feature: Lot number stamped on HU during manufacturing receipt
...
+    When complete planning for PP_Order:
+      | PP_Order_ID.Identifier |
+      | ppOrder_lotno_ws5      |
+      | ppOrder_lotno_ws7      |
+
+    # Each HU must carry ITS OWN workstation's line digit -- 5 and 7, not one shared value.
+    Then M_HU_Attribute is validated
+      | M_HU_ID      | M_Attribute_ID.Value | Value |
+      | hu_lotno_ws5 | Lot-Nummer           | L0915 |
+      | hu_lotno_ws7 | Lot-Nummer           | L0917 |
```
teosarca: can u receive multiple HUs so to make sure sequnece is also working?

teosarca: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/8660107a4db — the scenario now receives a multi-HU receipt. One part of the ask turned out not to be assertable, so splitting the answer.

**What was blocking it:** the step itself could not express it. `M_HU_LUTU_Configuration_StepDef.create_M_HU_LUTU_Configuration_for_pp_order` hard-asserted `assertThat(hus).hasSize(1)` and bound only `hus.get(0)`, so any receipt packing into more than one HU failed the step regardless of the `QtyTU` given. Fixed using the convention already in the codebase: `M_HU_ID.Identifier` now takes comma-separated identifiers, bound positionally to the received HUs via `StepDefDataIdentifier.toCommaSeparatedList()`. A single identifier behaves exactly as before (list of one → `hasSize(1)`), so no existing scenario changes meaning.

**What the scenario now covers:** a third order on line 5, `QtyEntered=20` / `QtyTU=2`, received as two HUs, asserting **both** are stamped and **both** carry the same lot `L0915`. That locks an invariant with no coverage until now: the provider is invoked **once per receipt**, not once per HU, because `AbstractPPOrderReceiptHUProducer.getOrLoadLotNumberFromSeq()` memoises the value in a field. If someone moved that call inside the per-HU loop, all three assertions of this scenario would still pass today — now the multi-HU rows fail.

**On "so the sequence is also working" — the running-number half is deliberately not assertable here, and I did not fake it.** `DBFunctionSequenceNoProvider.isUseIncrementSeqNoAsPrefix()` returns `false`, so no incremental counter is appended: the lot value is exactly what the DB function returns, nothing more. `LotNumberBL.getAndIncrementLotNo` only builds the document-no through the sequence with the order as `Record_ID` context. So for this provider there is no running number that could advance between HUs, and the customer's real function is likewise a pure computation (day-of-year + shift + line digit). An assertion that a counter increments would assert behaviour the feature does not have, and would be a false regression guard.

What *is* verifiable about the sequence wiring, and what the scenario asserts, is that the provider is actually reached and its value applied to every HU of the receipt — plus the existing discrimination that the digit comes from the order's `WorkStation_ID` and not from its plant. If you did mean a genuinely incrementing lot number (i.e. the plain-sequence path, which is a separate scenario in this same feature file), say so and I will add coverage for that path instead — that one *does* have a counter.

`de.metas.cucumber` test-compiles clean locally (609 sources, output classes verified); CI is running the scenario now.

teosarca: **Resolution (dev):** Correction to my own reply above — one claim in it was wrong, and an independent review caught it. Fixed in https://github.com/metasfresh/metasfresh/commit/03b90b44bd6

**What I got wrong.** I wrote that the new multi-HU assertion locks the "provider invoked once per receipt, not once per HU" invariant, and that moving the call inside a per-HU loop would now fail the scenario. It would not. The Background freezes the clock (`SystemTime.setFixedTimeSource`), and `test_lotno_workstation` is a pure function of `(order, timestamp)` with no side-effecting counter — and on the `isUseIncrementSeqNoAsPrefix()==false` branch `DocumentNoBuilder` returns the provider's value verbatim, never touching `retrieveAndIncrementSeqNo`. So calling the provider once or once-per-HU issues identical SQL with identical bind parameters against a deterministic function: the same string comes back either way. Delete the memoisation in `AbstractPPOrderReceiptHUProducer.getOrLoadLotNumberFromSeq()` and this scenario stays green.

I asserted the guard without running the falsification check — would this test fail if the behaviour it claims to protect were broken? It would not, so it was not a regression guard for that invariant.

**What the scenario does prove**, and this part stands: a receipt that packs into more than one HU has every HU stamped, all with the same lot number. Nothing covered that path before — the step definition physically could not express it — so a stamping path that handled only the first HU would have gone unnoticed. The workstation-hop discrimination (line digit from `WorkStation_ID`, not the plant) is also still genuinely covered by the two single-HU orders.

The feature file now states both non-guarantees explicitly, so the next reader does not inherit the false guard: not a call-count assertion, and not a running-counter assertion. If the call-count invariant is worth locking, it needs a mock-based unit test on the producer asserting `ILotNumberBL.getAndIncrementLotNo` is invoked exactly once across a multi-HU receipt — say the word and I will add that instead of leaving it implied.

Two smaller review findings fixed in the same commit: the touched step definition now carries the method-level Javadoc this module requires (documenting the comma-separated multi-identifier form with a Gherkin example), and the DataTable pipes are realigned.

`de.metas.cucumber` test-compiles clean; CI is running on the new commit.

PR #25612, opened 2026-08-25
EDI DESADV: reach a terminal status on an under-delivery, and populate a retro-created DESADV

teosarca:
## What

In per-`M_InOut` DESADV export mode, an `EDI_Desadv` could sit at `Pending` forever once an order was
under-delivered: the terminal status was gated on `FulfillmentPercent >= 100`, which an under-delivery
never reaches. Closing the remaining `M_ShipmentSchedule`s expressed "nothing more will ship", but
nothing propagated that to the DESADV header.

Separately, an `EDI_Desadv` created on the fly while completing an `M_InOut` ended up with delivered
quantities of zero.

## How

1. **Delivery-closed predicate.** New per-`EDI_DesadvLine` check
   `QtyDeliveredInStockingUOM >= COALESCE(QtyOrdered_Override, QtyOrdered)`, added as an *additional
   alternative* beside the existing `FulfillmentPercent >= 100` gate in
   `DesadvBL.recomputeDesadvStatusFromInOuts`. A DESADV with no lines is deliberately **not**
   delivery-closed (`Stream.allMatch` is vacuously true on an empty stream, which would auto-close
   freshly created, still-empty DESADVs).

   `QtyOrdered_Override` needs care: the generated getter maps SQL `NULL` to `BigDecimal.ZERO`, so a
   naive `COALESCE` on it would read "no override" as ordered-qty 0 and trivially close *every* line.
   The distinction goes through `InterfaceWrapperHelper.isNull(...)`, the idiom already used for
   `M_ShipmentSchedule` in this class. Both sides are pinned by tests.

2. **Fires from every close route.** `updateQtyOrdered_OverrideFromShipSchedAndSave` — whose sole
   caller is the `M_ShipmentSchedule` `@ModelChange` interceptor — now recomputes the owning DESADV's
   status after writing the line. Stating the rule on the record rather than in any one process means
   it fires from the mass-close process, the REST shipment service, the warehouse service and the
   order path alike, with no interceptor signature change.

3. **Write only on change, and re-derive only once per transaction.** Two distinct costs, fixed
   separately:

   - *Writes.* All four status branches go through one guard, so closing an N-line order performs at
     most one write per transition instead of N. Evaluation order is unchanged — `Invalid` and
     `Error` aggregation keep priority over closing.
   - *Reads.* The write guard does nothing for the **reads**. The `M_ShipmentSchedule` interceptor
     fires once per record and `M_ShipmentSchedule_CloseShipmentSchedules` closes its selection row by
     row, so closing N lines re-derived the whole header N times — each pass costing three to four
     uncached round-trips (`retrieveById`, `isDESADVExternalSystemRecipient`, `retrieveAllInOuts`, and
     via `areAllDesadvLinesDeliveryClosed` a `retrieveAllDesadvLines`), i.e. roughly quadratic in
     DESADV line count. `recomputeDesadvStatusFromInOutsBeforeCommit` now accumulates the touched
     `EDI_Desadv_ID`s on the transaction and recomputes once per distinct id, via
     `ITrxManager.accumulateAndProcessBeforeCommit` — the shape used by
     `InvoicePayScheduleService.validateInvoiceBeforeCommit` and
     `PaymentTermService.validateBeforeCommit`.

     BEFORE_COMMIT rather than AFTER_COMMIT on purpose: the recompute writes the `EDI_Desadv`, so it
     stays inside the same transaction — the new status is still visible to whoever reads the DESADV
     once the close returns, and a failure fails the close instead of being logged and dropped. With
     no active transaction the shared helper runs the recompute immediately, so a stand-alone close is
     unchanged.

4. **Delivered quantities on a retro-created DESADV.** In `addToDesadvCreateForInOutIfNotExist` the
   `M_InOutLine` walk ran *before* the `C_Order_ID` fallback that creates the `EDI_DesadvLine`s the
   walk needs. The walk is now extracted and run a second time, only when the fallback actually
   created the header, so the fresh lines receive the shipment quantities. Order lines are re-read by
   id inside the walk rather than reused from a possibly-stale association.

## Not changed, deliberately

- **Nothing the EDI recipient receives changes value.** `EDI_Desadv.FulfillmentPercent` and
  `SumOrderedInStockingUOM` are `EXP_FormatLine`s of the DESADV export format; how either is computed
  is untouched.
- The automatic close **transmits nothing** — no export enqueue, no work package, no external-system
  call. Reopening a schedule reopens the DESADV and does not re-send.
- `EDI_Desadv_Close` keeps its current behaviour and stays available.
- The per-line SQL `IsDeliveryClosed` is untouched. No migration script, no `AD_*` change, no new
  `AD_SysConfig`.

## Tests

`de.metas.edi` unit suite green (37 tests). Coverage is split by tier deliberately: the three things
an end-to-end test cannot observe are unit-tested — the predicate's `NULL`-vs-zero behaviour, "no
write when the status is unchanged", and "N closes in one transaction cause exactly one
re-derivation" (counted by invocation, since the in-memory test store exposes no save or query
counter). The business flow is covered end-to-end by Cucumber: nine scenarios in
`edi/externalSystemBased/ediDesadvExportViaExternalSystem.feature`, covering the terminal-status flip
on an under-delivery, the still-open case, closing after export, the nothing-sent case, that the
close transmits nothing, the multi-line mass close, reopen-without-resend, location-level routing
precedence, and the delivered quantities on a DESADV created at shipment completion.

Note on CI: `test (cucumber) (profile5)` and `(profile6)` fail identically on the base commit this
branch starts from, so they are pre-existing and unrelated. `profile3`, which runs the DESADV export
feature this change touches, is green.

### DIFF
--- a/backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/DesadvBL.java
+++ b/backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/DesadvBL.java
@@ -1167,11 +1167,18 @@ public class DesadvBL
 	 * <ol>
 	 *   <li>Any linked InOut is Invalid → DESADV Invalid + aggregated error message</li>
 	 *   <li>Any linked InOut is Error (and none Invalid) → DESADV Error + aggregated error message</li>
-	 *   <li>All linked InOuts are Sent or DontSend AND (FulfillmentPercent >= 100% OR every {@code EDI_DesadvLine}
-	 *       is delivery-closed) → DESADV Sent/DontSend, clear error message</li>
-	 *   <li>Any linked InOut is Pending, Enqueued, or SendingStarted, OR neither of the above closing conditions
-	 *       holds → DESADV Pending, clear error message</li>
+	 *   <li>All linked InOuts are Sent or DontSend AND no further delivery is expected → DESADV Sent
+	 *       if at least one InOut is Sent, else DontSend; error message cleared.
+	 *       "No further delivery expected" means FulfillmentPercent &gt;= 100%, <b>or</b> every
+	 *       {@code EDI_DesadvLine} is delivery-closed
+	 *       ({@code QtyDeliveredInStockingUOM >= COALESCE(QtyOrdered_Override, QtyOrdered)}), which is
+	 *       what closing the line's {@code M_ShipmentSchedule} expresses. This is what lets an
+	 *       under-delivered DESADV reach its terminal status without running the
+	 *       {@code EDI_Desadv_Close} process.</li>
+	 *   <li>Otherwise → DESADV Pending, error message cleared</li>
 	 * </ol>
+	 * The computed status is written only when it (or the error message) actually differs from the
+	 * stored value, because closing an order's lines fires this recompute once per line.
 	 */
 	public void recomputeDesadvStatusFromInOuts(@NonNull final EDIDesadvId desadvId)
 	{
--- a/backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/DesadvBL.java
+++ b/backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/DesadvBL.java
@@ -376,29 +376,7 @@ public class DesadvBL
 		final Map<EDIDesadvId, EDIDesadvPackService.Sequences> sequencesByDesadv = new HashMap<>();
 
 		final List<I_M_InOutLine> inOutLines = inOutDAO.retrieveLines(inOut, I_M_InOutLine.class);
-		for (final I_M_InOutLine inOutLine : inOutLines)
-		{
-			if (inOutLine.getC_OrderLine_ID() <= 0)
-			{
-				continue; // the DESADV-Line needs to relate to an orderline to make sense
-			}
-
-			// Resolve the DESADV for this inOutLine via its order line → desadv line → desadv header.
-			// When the order line has no EDI_DesadvLine_ID set, this line cannot contribute to any
-			// source DESADV (addInOutLine would early-return anyway), so we skip it here.
-			final I_C_OrderLine orderLineRecord = InterfaceWrapperHelper.create(inOutLine.getC_OrderLine(), I_C_OrderLine.class);
-			final EDIDesadvLineId desadvLineId = EDIDesadvLineId.ofRepoIdOrNull(orderLineRecord.getEDI_DesadvLine_ID());
-			if (desadvLineId == null)
-			{
-				continue;
-			}
-			final I_EDI_DesadvLine desadvLineRecord = desadvDAO.retrieveLineById(desadvLineId);
-			final EDIDesadvId lineDesadvId = EDIDesadvId.ofRepoId(desadvLineRecord.getEDI_Desadv_ID());
-
-			final EDIDesadvPackService.Sequences lineSequences = sequencesByDesadv.computeIfAbsent(lineDesadvId, ediDesadvPackService::createSequences);
-
-			addInOutLine(inOutLine, recipientBPartnerId, lineSequences, desadvLineRecord);
-		}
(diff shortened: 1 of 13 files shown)


backend/de.metas.edi/src/main/java/de/metas/edi/api/impl/DesadvBL.java
```diff
@@ -798,6 +822,12 @@ public void updateQtyOrdered_OverrideFromShipSchedAndSave(@NonNull final I_M_Shi
 		final BigDecimal qtyOrdered_Override = getQtyOrdered_Override(schedule);
 		desadvLineRecord.setQtyOrdered_Override(qtyOrdered_Override);
 		desadvDAO.save(desadvLineRecord);
+
+		// The override we just wrote can complete (or re-open) the DESADV's last open line, so the
+		// header status has to be re-derived. This is stated here rather than in any close process
+		// because every close route goes through ShipmentScheduleBL.closeShipmentSchedule and thus
+		// through this interceptor-driven method.
+		recomputeDesadvStatusFromInOuts(EDIDesadvId.ofRepoId(desadvLineRecord.getEDI_Desadv_ID()));
```
teosarca: isn't this introduce SQL N+1 problem, pls eval

teosarca: **Resolution (dev):** Evaluated — the concern is valid, and it is not fully covered by the existing guards. Findings, with the call path:

`M_ShipmentSchedule` (the interceptor) is unchanged on this branch: it already fired `@ModelChange(TYPE_AFTER_CHANGE, ifColumnsChanged = {QtyOrdered_Override, IsClosed})` once per record. What is new here is the whole-aggregate recompute added at the tail of `updateQtyOrdered_OverrideFromShipSchedAndSave`. So closing N shipment schedules of one DESADV now re-derives the entire header N times.

Per firing, `recomputeDesadvStatusFromInOuts` issues two aggregate-wide queries that are neither cached nor bounded by the changed record:

- `DesadvDAO.retrieveAllInOuts` — `IQueryBL` over the `EDI_Desadv_M_InOut` junction, returning **all** linked shipments;
- `areAllDesadvLinesDeliveryClosed` -> `DesadvDAO.retrieveAllDesadvLines` — returning **all** `EDI_DesadvLine`s.

Since there is typically one shipment schedule per order line and one `EDI_DesadvLine` per order line, N scales with the line count, so the cost is roughly quadratic in DESADV size rather than a textbook N+1 (which would be one list query plus N single-row lookups). The shape is "per-item interceptor performing a full-aggregate recomputation".

Two things I expected to mitigate this do not:

1. `setDesadvStatusAndSaveIfChanged` guards repeated **writes** only. The repeated **reads** are unaffected.
2. The `allProcessed`-first short-circuit carries a comment claiming it spares the `areAllDesadvLinesDeliveryClosed` query "on the common path". That does not hold for the path this change exists to serve: on an under-delivery close, the shipments are already exported (`allProcessed == true`) and `FulfillmentPercent < 100`, so the `||` does not short-circuit and the second aggregate query runs on **every** firing — precisely when N is largest.

Also worth flagging honestly: the frozen acceptance criterion covering the once-per-line firing asserts "no repeated writes". It says nothing about repeated reads, so this cost is genuinely outside what the test suite checks rather than accidentally untested.

Proposed fix: collapse the recompute to once per DESADV per transaction — dedupe on `EDI_Desadv_ID` and run the recompute at commit boundary — which keeps the interceptor placement (chosen because every close route goes through `ShipmentScheduleBL.closeShipmentSchedule`) while making the aggregate queries O(1) per transaction instead of O(N). I am verifying the "every close route" claim before implementing, because if any close route bypasses the interceptor that constrains where the recompute may legally move. Will push the fix and report back here rather than leave this as an accepted cost.

PR #25615, opened 2026-08-25
ESR import: a system-set payment action is not a user decision; field + message fixes

metas-ts:
## What

Follow-up to https://github.com/metasfresh/metasfresh/pull/25605, which was squash-merged while this work was still in progress and therefore carried only the ref-list migration (`5819900`). This PR holds the rest.

Two independent fixes plus one migration, all on `ESR_ImportLine.ESR_Payment_Action`.

## 1. A system-set payment action is not a user decision

Found in UAT: a line the import had flagged `Duplicate_Payment` was **silently closed** by a later "Process ESR" run that the accountant triggered for a completely *different* line.

`ESRImportBL.handleEsrImportLine(String, I_ESR_ImportLine)` only tested that `ESR_Payment_Action` is non-null before running a handler and calling `setProcessed(true)`. Since the import sets some actions itself, that test cannot tell a system flag from an accountant's decision.

Now skipped in that pass:

| Action | Why |
|---|---|
| `Duplicate_Payment` | the import flags it; the accountant must still choose an overpayment action. It is **not selectable at all** (`ESRPaymentActionValidationRule` accepts it in no group), so skipping it cannot block a deliberate choice |
| `Control_Line`, `Reverse_Booking` | no handler is registered for either, so the pass only appended a `NotSupported` message on every run. Reverse bookings are documented as admin-handled |

`Unable_To_Assign_Income` is deliberately **not** skipped — it *is* offered in the dropdown, so it can be a genuine choice. `Fit_Amounts` needs no guard: those lines are already `Processed`, so the method returns early.

Money was never at risk: `AbstractESRActionHandler`'s guard (`!isAllocated && !isPaid`) is false for a settled invoice, so nothing was allocated even via the old path. What was lost was the accountant's prompt — a real overpayment left with no open todo.

## 2. Match-error messages no longer accumulate duplicates

`ESRDataLoaderUtil.addMatchErrorMsg` is now idempotent per message, and the separate `addMatchErrorMsgIfNotPresent` is gone (its single caller repointed). Every `evaluate*` entry point is re-runnable by design — import, a `@ModelChange` interceptor, then a second Process run — so the unconditional append showed the same sentence two or three times. UAT saw exactly that.

Fixing the one helper covers **all** call sites rather than patching most of them and leaving the trap for the next one. Known limitation, documented and pinned by a test: the check is `contains`, so a shorter message sitting inside a stored one is treated as present. Whole-message matching was considered and rejected — it only trades this for the mirror bug, since a message containing `"; "` would then duplicate again.

## 3. The field itself had no label, description or help

`AD_Element` 542031 still carried the raw column name, so the WebUI showed **`ESR_Payment_Action`** as the field label; Description/Help were NULL; `de_DE`/`de_CH` were untranslated; `en_US` said only "Action".

Now **Zahlungsaktion / Payment action**, with a description saying what the field decides and that it must be set before the import can complete. The Help documents why the action list can come up **empty** — a line with no payment offers nothing — because that is a dead end users otherwise hit with no explanation.

Notable: the propagation function `update_FieldTranslation_From_AD_Name_Element` sets `Name`/`Description` on the base `AD_Field` row but **never `Help`** (Help only reaches `AD_Field_Trl`), so the migration writes `AD_Field.Help` directly, set-based via the element, covering both routes the function itself covers.

## Testing

- `ESRImportTest#completingTheImportMustNotCloseTheFlaggedDuplicate` pins the exact UAT regression: `complete()` must leave a flagged duplicate open, keep its flag, and keep its own payment.
- `ESRDataLoaderUtilTest` pins de-duplication, null/blank safety, and the documented substring tradeoff.
- The migration was validated on a throwaway PostgreSQL reproducing live state — including a propagation stub that deliberately omits `Help` exactly as the real function does, and one `AD_Field` per route.

**The Java tests have not been executed locally** — this module does not compile in my environment (the generated `camt054` sources need `javax.xml.bind`). CI is their first run.

### DIFF
--- a/backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/api/impl/ESRImportBL.java
+++ b/backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/api/impl/ESRImportBL.java
@@ -1085,10 +1085,7 @@ public class ESRImportBL implements IESRImportBL
 
 		final String actionType = line.getESR_Payment_Action();
 
-		// A SYSTEM-SET flag is not a user decision. The import sets some actions itself, and this pass
-		// only tests that ESR_Payment_Action is non-null before running a handler and marking the line
-		// Processed -- so without this guard a later Process run (possibly triggered for a completely
-		// different line) silently closes such a line although nobody decided anything.
+		// A SYSTEM-SET flag is not a user decision:
 		//   Duplicate_Payment: the import flags it, the accountant must still pick an overpayment
 		//     action. It is not even selectable -- ESRPaymentActionValidationRule accepts it in no
 		//     group -- so skipping it here cannot block a deliberate choice.
--- a/backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/api/impl/ESRImportBL.java
+++ b/backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/api/impl/ESRImportBL.java
@@ -1019,6 +1019,16 @@ public class ESRImportBL implements IESRImportBL
 		}
 	}
 
+	/**
+	 * Actions the IMPORT sets by itself, which therefore must not be treated as a decision by
+	 * {@link #handleEsrImportLine(String, I_ESR_ImportLine)}. See the comment at the guard for why each
+	 * one is in here -- and why {@code Unable_To_Assign_Income} is not.
+	 */
+	private static final ImmutableSet<String> SYSTEM_SET_ACTIONS_AWAITING_DECISION = ImmutableSet.of(
+			X_ESR_ImportLine.ESR_PAYMENT_ACTION_Duplicate_Payment,
+			X_ESR_ImportLine.ESR_PAYMENT_ACTION_Control_Line,
+			X_ESR_ImportLine.ESR_PAYMENT_ACTION_Reverse_Booking);
+
 	private void handleEsrImportLine(final String message, final I_ESR_ImportLine line)
 	{
 		if (line.isProcessed())
@@ -1074,6 +1084,24 @@ public class ESRImportBL implements IESRImportBL
 		}
 
 		final String actionType = line.getESR_Payment_Action();
+
+		// A SYSTEM-SET flag is not a user decision. The import sets some actions itself, and this pass
+		// only tests that ESR_Payment_Action is non-null before running a handler and marking the line
+		// Processed -- so without this guard a later Process run (possibly triggered for a completely
+		// different line) silently closes such a line although nobody decided anything.
+		//   Duplicate_Payment: the import flags it, the accountant must still pick an overpayment
+		//     action. It is not even selectable -- ESRPaymentActionValidationRule accepts it in no
+		//     group -- so skipping it here cannot block a deliberate choice.
+		//   Control_Line / Reverse_Booking: no handler is registered for either, so this pass would
+		//     only append a "not supported" message to the line on every run. Reverse bookings are
+		//     documented as something the admin deals with manually.
+		// Unable_To_Assign_Income is deliberately NOT in this set: it IS offered in the dropdown, so
+		// it can be a genuine user choice and must keep being processed.
+		if (SYSTEM_SET_ACTIONS_AWAITING_DECISION.contains(actionType))
+		{
+			return;
+		}
+
 		if (Check.isEmpty(actionType, true))
 		{
 			final AdempiereException ex = AdempiereException.newWithTranslatableMessage("@" + ESRConstants.ERR_ESR_LINE_WITH_NO_PAYMENT_ACTION + "@");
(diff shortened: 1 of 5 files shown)


backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/api/impl/ESRImportBL.java
```diff
@@ -1074,6 +1084,24 @@ private void handleEsrImportLine(final String message, final I_ESR_ImportLine li
...
+		// Processed -- so without this guard a later Process run (possibly triggered for a completely
+		// different line) silently closes such a line although nobody decided anything.
+		//   Duplicate_Payment: the import flags it, the accountant must still pick an overpayment
+		//     action. It is not even selectable -- ESRPaymentActionValidationRule accepts it in no
+		//     group -- so skipping it here cannot block a deliberate choice.
+		//   Control_Line / Reverse_Booking: no handler is registered for either, so this pass would
+		//     only append a "not supported" message to the line on every run. Reverse bookings are
+		//     documented as something the admin deals with manually.
+		// Unable_To_Assign_Income is deliberately NOT in this set: it IS offered in the dropdown, so
+		// it can be a genuine user choice and must keep being processed.
```
metas-ts:
```suggestion
		// A SYSTEM-SET flag is not a user decision:
		//   Duplicate_Payment: the import flags it, the accountant must still pick an overpayment
		//     action. It is not even selectable -- ESRPaymentActionValidationRule accepts it in no
		//     group -- so skipping it here cannot block a deliberate choice.
		//   Control_Line / Reverse_Booking: no handler is registered for either, so this pass would
		//     only append a "not supported" message to the line on every run. Reverse bookings are
		//     documented as something the admin deals with manually.
		// Unable_To_Assign_Income is deliberately NOT in this set: it IS offered in the dropdown, so
		// it can be a genuine user choice and must keep being processed.
```

backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/dataimporter/ESRDataLoaderUtil.java
```diff
@@ -399,24 +399,29 @@ public String removeLeftZeros(final String value)
...
+	 * design -- the import evaluates a line, a {@code @ModelChange} interceptor evaluates it again, and a
+	 * second Process run evaluates it once more -- so an unconditional append shows the accountant the
+	 * same sentence two or three times over. It is idempotent per message, so a genuinely different
+	 * message is still appended.
+	 * <p>
+	 * Known limitation, deliberately kept: the check is {@code contains}, so a shorter message that
+	 * happens to sit inside one already stored is treated as present and dropped. Matching whole
+	 * messages instead (splitting on the {@code "; "} that {@link #addMsgToString(String, String)}
+	 * joins with) would only trade this for the opposite bug -- a message that itself contains
+	 * {@code "; "} would then duplicate again. Pinned by ESRDataLoaderUtilTest.
```
metas-ts:
```suggestion
```

backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/api/impl/ESRImportBL.java
metas-ts: **Resolution (dev):** Adopted verbatim — the four-line preamble is gone, the per-action bullets kept. Applied in the working tree but **not yet committed or pushed**, per your "don't push yet, there is more".

backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/dataimporter/ESRDataLoaderUtil.java
metas-ts: **Resolution (dev):** Adopted — lines 405-415 deleted (the "de-duplication is not a nicety" explanation and the "Known limitation" paragraph). Two notes: (1) your range started at 405 but 404 was `* <p>`, which left a dangling tag and malformed Javadoc, so I removed that line too — a mechanical consequence of the cut, say if you want it kept; (2) the `contains` substring limitation is now undocumented at the method, but stays pinned by `ESRDataLoaderUtilTest#aMessageThatIsASubstringOfAStoredOne_isSwallowed_knownTradeoff`, so the behaviour is locked even though it is no longer explained in place. Not yet committed or pushed, per your instruction.

backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/api/impl/ESRImportBL.java
metas-ts: **Resolution (dev):** Now committed and pushed: https://github.com/metasfresh/metasfresh/commit/f6f59c9210e

backend/de.metas.payment.esr/src/main/java/de/metas/payment/esr/dataimporter/ESRDataLoaderUtil.java
metas-ts: **Resolution (dev):** Now committed and pushed: https://github.com/metasfresh/metasfresh/commit/f6f59c9210e

PR #25618, opened 2026-08-25
`Customer delivery price` report: Add `Date To` period and `Base/Special` prices

riadmen:
Extend report.getCustomerDeliveryPriceOverview (AD_Process 585614) so it covers  a bounded delivery period and shows both the standard list price and the customer-specific special price per article.                                                                                                       
                                                                                                                                                     
  - New helper report.getSalesPriceSpecialAndBase(date, bpartner_location, product):                                                                 
    reuses the recursive BasePriceList_ID-chain resolution to return, in one row, the special price (customer's assigned depth-0 override list) and the base price (base list in the chain). Either may be NULL; special is empty when the assigned list is itself a base list (no override). getCurrentSalesPrice left untouched.                                                                   
                                                                                                                                                     
  - getCustomerDeliveryPriceOverview: add nullable p_DeliveryDateTo (NULL falls back to p_DeliveryDateFrom, i.e. the single From day), bound the delivery window, and replace the single Price/IsFromBasePriceList output with BaseListPrice and SpecialPrice columns (each with per-UOM + price-list-version) sourced from the new helper.                                                                                                                                      
                                                                                                                                                     
  - Update the AD_Process 585614 description (de + en) to describe the selectable period and the two price sources.


backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getSalesPriceSpecialAndBase.sql
```diff
@@ -0,0 +1,148 @@
...
+                   AND c.depth < 10),
+    -- For each PL in the chain, take its newest active PLV valid on p_Date
+    plv_per_pl AS (SELECT DISTINCT ON (c.M_PriceList_ID) c.depth,
+                                                         c.M_PriceList_ID,
+                                                         plv.M_PriceList_Version_ID
+                   FROM pl_chain c
+                            JOIN M_PriceList_Version plv ON plv.M_PriceList_ID = c.M_PriceList_ID
+                   WHERE plv.IsActive = 'Y'
+                     AND plv.ValidFrom <= p_Date
+                   ORDER BY c.M_PriceList_ID, plv.ValidFrom DESC),
```
mrie04: `bpl` / `pricing_system` / `pricelist_root` / `pl_chain` / `plv_per_pl` are verbatim from `getCurrentSalesPrice.sql` — a third copy of the chain walk. `public.getPriceListVersionsUpToBase(M_PriceList_ID, date)` (and `getpricelistversionsuptobase_ForPricelistVersion`) already returns the ordered PLV chain.

Also: after this PR `getCurrentSalesPrice` has no callers left in the repo — drop it, or build this function on the core resolver.

_(same in the identical `5820280` copy)_

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getSalesPriceSpecialAndBase.sql
```diff
@@ -0,0 +1,148 @@
...
+                      pp.PriceStd,
+                      pp.C_UOM_ID,
+                      plv.Name                       AS PLVName,
+                      COALESCE(pp.SeqNo, 2147483647) AS seqno_sort,
+                      pp.M_ProductPrice_ID
+               FROM plv_per_pl x
+                        JOIN M_PriceList_Version plv ON plv.M_PriceList_Version_ID = x.M_PriceList_Version_ID
+                        JOIN M_ProductPrice pp ON pp.M_PriceList_Version_ID = x.M_PriceList_Version_ID
+               WHERE pp.M_Product_ID = p_M_Product_ID
+                 AND pp.IsActive = 'Y'),
```
mrie04: **Blocking.** `priced` filters only `M_Product_ID` and `IsActive`, then `ORDER BY seqno_sort, M_ProductPrice_ID LIMIT 1` — so an attribute-dependent or packing-material-specific `M_ProductPrice` can win.

That contradicts the description this PR ships in `5820270`: _"No ASI/PI matching: price resolution ignores attribute-set-instance (ASI) specific prices and packaging (handling unit/PI) specific prices."_ It doesn't ignore them; it may select one.

`get_Product_Price.sql` in this same directory already has the filter set:

```sql
AND IsInvalidPrice <> 'Y'
AND IsAttributeDependant = 'N'
AND M_HU_PI_Item_Product_ID IS NULL
```

Canonical query agrees — `ProductPriceQuery.java:288` always excludes `IsInvalidPrice`, `:297` sets `IsAttributeDependant = false`.

_(same in the identical `5820280` copy)_

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getSalesPriceSpecialAndBase.sql
```diff
@@ -0,0 +1,148 @@
...
+    -- Base: the nearest base-list price. When the assigned list is an override, that is the
+    -- nearest depth>=1 row; when it is itself a base list, the depth-0 row IS the base.
+    base AS (SELECT pr.PriceStd, pr.C_UOM_ID, pr.PLVName
+             FROM priced pr,
+                  root_info ri
+             WHERE pr.depth >= (CASE WHEN ri.is_override THEN 1 ELSE 0 END)
+             ORDER BY pr.depth ASC,
+                      pr.seqno_sort ASC,
+                      pr.M_ProductPrice_ID ASC
+             LIMIT 1)
```
mrie04: Comment says "the nearest base-list price"; the predicate says `depth >= 1`. In an override → override → base chain, depth 1 is still an override, so `BasePriceStd` would carry an override price.

If "base list" is meant literally, key on `BasePriceList_ID IS NULL`; if `depth >= 1` is deliberate, the comment should read "nearest ancestor".

_(same in the identical `5820280` copy)_

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getCustomerDeliveryPriceOverview.sql
```diff
@@ -50,30 +60,33 @@ SELECT (SELECT value FROM C_BPartner WHERE C_BPartner_id = p_C_BPartner_ID) AS P
...
+       special_uom.name                                                     AS SpecialPricePerUOM,
+       pr.SpecialPLV                                                        AS SpecialPriceListVersion
 FROM m_shipmentschedule sched
          JOIN m_product p ON p.m_product_id = sched.m_product_id
          JOIN c_uom uom ON uom.c_uom_id = p.c_uom_id
          JOIN c_bpartner bp ON bp.c_bpartner_id = sched.c_bpartner_id
-         LEFT JOIN LATERAL report.getCurrentSalesPrice(NOW(), sched.c_bpartner_location_id, sched.m_product_id) pp ON TRUE
-         LEFT JOIN c_uom pp_uom ON pp_uom.c_uom_id = pp.c_uom_id
-         LEFT JOIN m_pricelist_version plv ON plv.m_pricelist_version_id = pp.m_pricelist_version_id
+         LEFT JOIN LATERAL report.getSalesPriceSpecialAndBase(NOW(), sched.c_bpartner_location_id, sched.m_product_id) pr ON TRUE
```
mrie04: **Blocking.** The price anchor is `NOW()`, so `p_DeliveryDateTo` has no influence on prices — it only scopes which shipment schedules are counted. Two consequences:

1. Not reproducible — the same call over the same closed period returns different prices once a newer price-list version exists.
2. A report over a past window can print a `BasePLV` / `SpecialPLV` whose `ValidFrom` lies *after* the window end — a version that priced nothing in the reported period.

Now that the function receives an explicit period, anchoring at its end keeps one price per article and removes both:

```sql
LEFT JOIN LATERAL report.getSalesPriceSpecialAndBase(
        COALESCE(p_DeliveryDateTo, p_DeliveryDateFrom) + INTERVAL '1 day' - INTERVAL '1 microsecond',
        sched.c_bpartner_location_id, sched.m_product_id) pr ON TRUE
```

matching the window bound on `:78`. ~~Also makes `STABLE` truthful — the function currently depends on wall-clock.~~

> **Correction (edited after the fact):** the struck sentence is wrong and should not be taken as a rule. `now()` is `STABLE` in PostgreSQL — it returns transaction start time, not wall-clock time; `clock_timestamp()` is the volatile one. A function calling `now()` is legitimately `STABLE`, so there was no volatility defect to fix here. Points 1 and 2 above are the actual reasons for the change, and they stand.

_(same in the identical `5820290` copy)_

mrie04 (commented):
Review of the two SQL functions and the migration copies. Two blocking items (`priced` filter set, and the `NOW()` price anchor), the rest are should-fix or questions. Line comments below.

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getCustomerDeliveryPriceOverview.sql
```diff
@@ -50,30 +60,33 @@ SELECT (SELECT value FROM C_BPartner WHERE C_BPartner_id = p_C_BPartner_ID) AS P
...
+         LEFT JOIN c_uom base_uom ON base_uom.c_uom_id = pr.BaseC_UOM_ID
+         LEFT JOIN c_uom special_uom ON special_uom.c_uom_id = pr.SpecialC_UOM_ID
 WHERE sched.processed = 'Y'
   AND sched.deliverydate >= p_DeliveryDateFrom
+  AND sched.deliverydate < COALESCE(p_DeliveryDateTo, p_DeliveryDateFrom) + INTERVAL '1 day'
```
mrie04: `DateTo` is `IsMandatory='N'` and NULL falls back to `DateFrom`, so a from-date-only call now returns a single day where the previous signature returned everything up to today.

If the old meaning should be preserved, `COALESCE(p_DeliveryDateTo, NOW())` covers both this bound and the price anchor above consistently.

_(same in the identical `5820290` copy)_

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getCustomerDeliveryPriceOverview.sql
```diff
@@ -50,30 +60,33 @@ SELECT (SELECT value FROM C_BPartner WHERE C_BPartner_id = p_C_BPartner_ID) AS P
...
   AND (p_C_BPartner_ID IS NULL OR bp.C_BPartner_ID = p_C_BPartner_ID)
 GROUP BY p.value, p.name, uom.name,
-         pp.pricestd, pp.c_currency_id,
-         pp.c_uom_id, pp.m_pricelist_id, pp.isfrombasepricelist,
-         plv.name, pp_uom.name, bp.value
+         pr.BasePriceStd, base_uom.name, pr.BasePLV,
+         pr.SpecialPriceStd, special_uom.name, pr.SpecialPLV,
+         bp.value
```
mrie04: Price resolves per `sched.c_bpartner_location_id` and all six price columns sit in the `GROUP BY`, so a partner whose locations resolve to different price lists yields two rows for the same article.

Pre-existing, but the surface grew from three price columns to six. Intended?

_(same in the identical `5820290` copy)_

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/70-de.metas.fresh/5820290_sys_gh29931_CustomerDeliveryPriceOverview.sql
```diff
@@ -0,0 +1,96 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program. If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+-- Task 3 (partial): add p_DeliveryDateTo (nullable; NULL falls back to p_DeliveryDateFrom -> single day)
+-- and bound the delivery window. Price columns unchanged for now (still via report.getCurrentSalesPrice).
```
mrie04: This comment contradicts the body below it — the shipped function uses `getSalesPriceSpecialAndBase`, not `getCurrentSalesPrice`.

It also makes this migration diverge from its `ddl/functions/` twin (this comment, the header year, the trailing newline), while the `getSalesPriceSpecialAndBase` pair is byte-identical. Drop it and re-sync the two copies.

mrie04 (changes requested):
pls check my comments

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/70-de.metas.fresh/5820290_sys_gh29931_CustomerDeliveryPriceOverview.sql
mrie04: also comments should not contain issue specific terms like task 3

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getSalesPriceSpecialAndBase.sql
riadmen: IMHO we should keep it, because dropping it in a hotfix needs a cross-repo check first => higher risk than the duplication.

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getCustomerDeliveryPriceOverview.sql
riadmen: This choice is documented in the AD_Process description: `if no Date to is given, only the Date from day is considered`

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getSalesPriceSpecialAndBase.sql
riadmen: Fixed:

- **Was:** `priced` filtered only `M_Product_ID` + `IsActive`, so an attribute-set-instance or packaging(HU/PI)-specific `M_ProductPrice` could win — contradicting the report's own "No ASI/PI matching" description.
- **Now:** added the canonical filter set (matches `get_Product_Price.sql` / `ProductPriceQuery.java`):
  ```sql
  AND pp.IsInvalidPrice <> 'Y'
  AND pp.IsAttributeDependant = 'N'
  AND pp.M_HU_PI_Item_Product_ID IS NULL
  ```

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getSalesPriceSpecialAndBase.sql
riadmen: Fixed:
- **Was:** comment said "nearest base-list price" but predicate was `depth >= 1`; in an override→override→base chain, depth 1 is still an override, so `BasePriceStd` could carry an override price.
- **Now:** `base` keys on the **true base list** — the price list whose `BasePriceList_ID IS NULL` — via a new `is_base_list` flag carried from `pl_chain`. A middle override level is never treated as base; comment rewritten to match.

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getCustomerDeliveryPriceOverview.sql
riadmen: Fixrd:
- **Was:** `getSalesPriceSpecialAndBase(NOW(), …)` — `p_DeliveryDateTo` never influenced prices; a closed-period report returned different prices once a newer PLV appeared, and could print a PLV whose `ValidFrom` was after the window.
- **Now:** anchored at the last instant of the delivery period —
  `getSalesPriceSpecialAndBase(COALESCE(p_DeliveryDateTo, p_DeliveryDateFrom) + INTERVAL '1 day' - INTERVAL '1 microsecond', …)`, matching the window upper bound. One reproducible price per article for a closed period; `STABLE` is now truthful.

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getCustomerDeliveryPriceOverview.sql
riadmen: Fixed:
- **Was:** price resolved per `c_bpartner_location_id` with all six price columns in `GROUP BY`, so a partner whose locations resolve to different (country-specific) price lists yielded several rows per article and split the `Qty`.
- **Now:** a `deliveries` CTE aggregates to **(business partner, product)** — `SUM(qtydelivered)` and a representative location (`MIN(c_bpartner_location_id)`); the price is resolved once via one `LATERAL` on that representative location and the outer query needs no `GROUP BY`. **Exactly one row per (partner, article)**;
- **Process description updated**: *"the delivered quantity is summed per article over the period; each article appears once, with a single base price and, if any, a single special price"*

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/system/70-de.metas.fresh/5820290_sys_gh29931_CustomerDeliveryPriceOverview.sql
riadmen: Fixed:
- **Was:** `5820290` carried `-- Task 3 (partial): … still via report.getCurrentSalesPrice`, which (a) contradicted the body (it calls `getSalesPriceSpecialAndBase`), (b) used the issue-internal term "Task 3", (c) made the migration diverge from its `ddl/functions/` twin (comment + copyright year + trailing newline).
- **Now:** comment removed; copyright aligned to 2026 and trailing newline normalised so migration `5820290` and `getCustomerDeliveryPriceOverview.sql` are **byte-identical** (verified), matching the `getSalesPriceSpecialAndBase` pair.

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getCustomerDeliveryPriceOverview.sql
mrie04: But why the choice is different to RE part. If single date is given we want to check until today, so now

riadmen: Done: Restore until-today when `Date To` is null

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getCustomerDeliveryPriceOverview.sql
mrie04: I think that isn't the right way to resolve it. Maybe even worse as random pick. wdyt about on deliveries CTE get effective c_location from coalesce(c_bpartner_location_value_override, c_bpartner_location_value), then the country and add it to groupBy

riadmen: Fixe:
- Resolve the price per delivery-location country (country-specific price lists), not per raw location. 
- The deliveries CTE derives the effective location (COALESCE(C_BP_Location_Override_ID, C_BPartner_Location_ID))                                                               
  and groups by its C_Country_ID, so same-country deliveries collapse to one row per article while genuinely different-country prices stay on separate rows.

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getSalesPriceSpecialAndBase.sql
riadmen: **Resolution (dev):** Fixed — base price is now taken from the TRUE base list (`BasePriceList_ID IS NULL`) via a new `is_base_list` flag carried from `pl_chain`, so a middle override level (`depth >= 1`) is never treated as base; the comment was rewritten to match. Commit https://github.com/metasfresh/metasfresh/commit/2df64f46f21305a688a481757d042b1f8be8f5f4

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/getSalesPriceSpecialAndBase.sql
riadmen: **Resolution (dev):** Won't-fix (deliberate, with reason). The `bpl`/`pricing_system`/`pl_chain` walk is kept as-is rather than delegated to `getPriceListVersionsUpToBase` / built on the core resolver, and `getCurrentSalesPrice` is left in place: this PR ships on a hotfix branch, and removing/rewiring a shared SQL resolver there would require a cross-repo impact check across all customer repos — higher risk than the local duplication. The refactor (drop `getCurrentSalesPrice`, share the chain walk) is the right follow-up on the mainline, not in this hotfix.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/pricing/M_PriceList_StepDef.java
```diff
@@ -174,6 +174,16 @@ private void createM_PriceList(@NonNull final DataTableRow row)
 		priceList.setIsSOPriceList(soTrx.toBoolean());
 		priceList.setC_Country_ID(country != null ? country.getC_Country_ID() : -1);
 
+		// Layering chain: link this list onto the list it falls back to (its base). getPriceListVersionsUpToBase
+		// walks BasePriceList_ID to reach the base list, so special/base price resolution can be exercised.
+		final StepDefDataIdentifier baseListIdentifier = row.getAsOptionalIdentifier("BasePriceList_ID")
```
mrie04: pls use `org.compiere.model.I_M_PriceList#COLUMNNAME_BasePriceList_ID`

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pricing/salesPriceSpecialAndBase.feature
```diff
@@ -0,0 +1,107 @@
...
+  Scenario: three-level chain takes the base price from the true base list, not the middle override
+    Given metasfresh contains M_PricingSystems
+      | Identifier              |
+      | layeredPricingSystem    |
+      | globalBasePricingSystem |
+    And metasfresh contains M_PriceLists
+      | Identifier   | M_PricingSystem_ID.Identifier | C_Currency.ISO_Code | SOTrx | BasePriceList_ID.Identifier |
+      | trueBaseList | globalBasePricingSystem       | EUR                 | false |                             |
+      | middleList   | layeredPricingSystem          | EUR                 | false | trueBaseList                |
+      | assignedList | layeredPricingSystem          | EUR                 | true  | middleList                  |
```
mrie04, line 89: why purchase and sales mixed?

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pricing/salesPriceSpecialAndBase.feature
```diff
@@ -0,0 +1,107 @@
...
+    Given metasfresh contains M_Products:
+      | Identifier  |
+      | productBoth |
+    And metasfresh contains M_ProductPrices
+      | M_PriceList_Version_ID.Identifier | M_Product_ID.Identifier | PriceStd | C_UOM_ID.X12DE355 | C_TaxCategory_ID.InternalName |
+      | baseListVersion                   | productBoth             | 100      | PCE               | Normal                        |
+      | overrideListVersion               | productBoth             | 90       | PCE               | Normal                        |
+    Then report.getSalesPriceSpecialAndBase returns:
+      | C_BPartner_Location_ID | M_Product_ID | Date       | SpecialPriceStd | BasePriceStd |
+      | customerLocation       | productBoth  | 2022-05-17 | 90              | 100          |
```
mrie04, line 46: i'm missing a dateTo case where prices exist before, in, after  range and newest in range is selected.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/pricing/M_PriceList_StepDef.java
riadmen: **Resolution (dev):** Done — replaced the `"BasePriceList_ID"` string literal with `I_M_PriceList.COLUMNNAME_BasePriceList_ID` (the class is already imported). Commit https://github.com/metasfresh/metasfresh/commit/20edb547443

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pricing/salesPriceSpecialAndBase.feature
riadmen: **Resolution (dev):** Fixed — the SO/non-SO mix was only a workaround for the `M_PriceList_UC_C_Country` unique index `(M_PricingSystem_ID, IsSOPriceList, COALESCE(C_Country_ID,0))`, which permits only one SO and one non-SO list per pricing system per country (country-less lists collapse to country 0). A sales-price report should walk sales lists only, so each rung now lives in its own pricing system (`layeredPricingSystem` -> `middlePricingSystem` -> `globalBasePricingSystem`), all `SOTrx=true`; `BasePriceList_ID` may cross pricing systems. Same assertion (Special 70 / Base 90). Commit https://github.com/metasfresh/metasfresh/commit/20edb547443

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/pricing/salesPriceSpecialAndBase.feature
riadmen: **Resolution (dev):** Added the scenario "the newest price-list version valid on the report date is used, ignoring older and future versions". baseList gets a version before (2022-01-01, price 100) and after (2022-09-01, price 120) the 2022-05-17 report date, around the in-range baseListVersion (2022-05-01, price 110). `getPriceListVersionsUpToBase` selects `MAX(ValidFrom) WHERE ValidFrom <= p_Date`, so 110 must win over both 100 and 120 -> asserts Base 110 (Special empty, the article is priced only on the base list). Commit https://github.com/metasfresh/metasfresh/commit/20edb547443

mrie04 (approved):
lgtm

PR #25627, opened 2026-08-26
VAT-ID follow-ups: format-invalid to Invalid, report overview, NotSupported relabel, check-log filters

metas-ts:
## What this PR does

Follow-up fixes to the VAT-ID online-check feature.

**Format-invalid VAT-IDs are now recorded as `Invalid`.** A syntactically-invalid VAT-ID reaching `VATaxIDCheckService.check()` previously threw at the format-check gate and the throw was swallowed, leaving the record `NotChecked` forever. When the format check is enabled and the value fails it, the check now records `Invalid` offline — no VIES call — via a check-log row marked as an offline format rejection (distinct from a VIES `valid:false`), flowing through the same store-and-refresh path so an open order's tax follows. The save-time interceptor still throws to block such a value at entry; this path handles values that were imported or predate the format check.

**Report `VATaxID_Config_Report`: new overview section** counting, per partner and per address, how many carry a VAT-ID versus none — so the per-status counts are read against the population that actually has a VAT-ID (a record that never had one reads as `NotChecked`, which otherwise dominates the count).

**`VATaxIDStatus` `NotSupported` relabelled** to "Außerhalb VIES-Abdeckung" / "Outside VIES coverage" (internal code unchanged) — the old "Nicht unterstützt" read like an error, when it means the country prefix is outside VIES coverage and the offline format check is its only validation.

**VATaxID_CheckLog window filter/grid tweaks:** VATaxID becomes filterable; the two date filters become from/to range filters (`AD_Column.FilterOperator='B'`); the Active column moves to the far right of the grid with its filter last.

**Also:** the external-system "enqueued to be synced" debug log lines now name their source class.

## Migrations
- `5820300` — report overview section (function replace; canonical `ddl/functions/VATaxID_Config_Report.sql` kept in sync)
- `5820310` — `NotSupported` relabel (de_DE / de_CH / en_US)
- `5820320` — check-log window filter/grid

## Verification
- `de.metas.vatid` unit tests green (170), including a new test pinning format-invalid → `Invalid` with no VIES call.
- Cucumber scenario added for the process path (an already-persisted malformed VAT-ID → `Invalid`).
- Migrations applied and verified on a customer-faithful DB; the check-log window layout was rule-checked and its Playwright spec extended for the grid-order change.

### DIFF
--- a/backend/de.metas.vatid/base/src/main/java/de/metas/vatid/VATaxIDMassCheckService.java
+++ b/backend/de.metas.vatid/base/src/main/java/de/metas/vatid/VATaxIDMassCheckService.java
@@ -172,9 +172,16 @@ public class VATaxIDMassCheckService
 	 *
 	 * <p>Due-ness and ordering are the query's ({@code IBPartnerDAO}), evaluated per organisation with that
 	 * organisation's own recheck window — organisations with the online check switched off are never queried
-	 * at all (only VIES-enabled organisations are swept — {@code getRecheckAfterDaysByViesEnabledOrgId}). By
-	 * design that means the nightly does NOT auto-detect a pre-existing / imported malformed VAT-ID on a
-	 * format-check-only organisation: there the manual {@code C_BPartner_VATaxID_Check} process is the path
+	 * at all (only VIES-enabled organisations are swept — {@code getRecheckAfterDaysByViesEnabledOrgId}).
+	 * <b>Why skip them:</b> the nightly exists solely to keep the <i>time-varying</i> VIES verdict current — a
+	 * VAT registration can be withdrawn or turned invalid after it was first checked, which is the whole
+	 * reason an online check is repeated at all. An organisation with VIES off has no such time-varying result
+	 * to refresh: its only validation is the offline format check, which is deterministic and already applied
+	 * at save time (and re-applied by the manual process), so a value's format verdict never changes on its
+	 * own between nights. Sweeping a VIES-off organisation would therefore re-examine values whose verdict
+	 * cannot have moved — pure work for no possible status change — so it is excluded by construction.
+	 * By design that also means the nightly does NOT auto-detect a pre-existing / imported malformed VAT-ID on
+	 * a format-check-only organisation: there the manual {@code C_BPartner_VATaxID_Check} process is the path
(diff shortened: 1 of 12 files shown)


backend/de.metas.vatid/base/src/main/java/de/metas/vatid/VATaxIDMassCheckService.java
```diff
@@ -171,10 +171,13 @@ public VATaxIDMassCheckResult run(@NonNull final VATaxIDMassCheckRequest request
 	 * The nightly sweep. Uses {@link VATaxIDMassCheckRequest#getMaxChecksPerRun()} and builds no selection.
 	 *
 	 * <p>Due-ness and ordering are the query's ({@code IBPartnerDAO}), evaluated per organisation with that
-	 * organisation's own recheck window — organisations with the check switched off are never queried at
-	 * all. The two grains run one after the other, headers then locations, each oldest-attempt-first; a
-	 * single combined ordering across both would need a merge, and raising {@code MaxChecksPerRun} is the
-	 * cheaper answer now that nothing is loaded up front.
+	 * organisation's own recheck window — organisations with the online check switched off are never queried
```
metas-ts, line 174: what's the functional reason (sense) of them never being checked?

metas-ts: **Resolution (dev):** The nightly sweep exists only to keep the *time-varying* online verdict current — a VAT registration can be withdrawn or turned invalid after it was first confirmed, which is the whole reason the online check is repeated on a schedule at all. An organisation with the online check switched off has no such time-varying result to refresh: its only validation is the offline format check, which is deterministic and already applied at save time (and re-applied by the manual process), so a value's format verdict can never move on its own between nights. Sweeping such an organisation would therefore re-examine values whose verdict cannot have changed — pure work for no possible status change — so it is excluded by construction.

The flip side, worth stating: a malformed value that was *imported* into an online-check-off organisation is consequently not auto-detected by the nightly; there the manual `C_BPartner_VATaxID_Check` process is the path that gives it its offline `Invalid` verdict.

I've expanded the `runNightly` javadoc to spell out this "why", not just the "what".

PR #25636, opened 2026-08-27
Delivery Planning — aggregate several delivery plannings in one delivery instruction

mrie04:
### Status

Implementation is complete through phase F.

| Phase | Scope | State |
|---|---|---|
| A | allocation table + indexes + generated models, the three-valued direction, the direction column on the instruction | **done** |
| B | `DeliveryPlanningList`, allocation CRUD, `Combine` / `Add to` / `Remove from`, `IsComplete` on `Generate`, the `IsB2B` retirement, the allocation backfill, dropping `M_ShipperTransportation.M_Delivery_Planning_ID` | **done** |
| C | lifecycle cascade, closed-planning guards, close/reopen interceptor, `IsSOTrx` retirement, doc-type value rules, deactivate-instead-of-delete on remove/move | **done** |
| D | the instruction owns the dates, `M_ShippingPackage.C_Order_ID` + null-guard, movement generation multi-select | **done** |
| E | report functions + JRXML, filterability + virtual columns, conditional read-only dates, the instruction window's two-tab shape | **done** |
| F | cucumber scenarios G1–G9 | **done** — 34 scenarios across 9 feature files: combine, add-to/move/remove, draft default, lifecycle cascade, closed-planning guards, the instruction's ownership of its plannings' dates (`deliveryInstructionDates.feature`), the consignment/history view invariants, and the aggregation-key lockstep against the AD |

**Resolved during review**: the complete / re-activate cascade mirrored the *pre-transition*
`DocStatus` onto its allocations, because `MMShipperTransportation` fires the `AFTER_COMPLETE` /
`AFTER_REACTIVATE` hook **before** `DocumentEngine` stamps the new status. The existing unit tests
passed only because they set that state themselves first — a fabricated precondition the real flow
never reaches. An allocation isn't a document, so it should never have carried a
`DocStatus`/`Processed` mirror at all. Fixed by removing the mirror entirely (columns, writers, the
complete/re-activate cascade) rather than by fixing the timing.

### Deleting is not an operation this feature offers

A delivery instruction is **cancelled or closed, never deleted** - the acceptance criteria know only
close, remove, void and cancel. Because allocations are retired (`IsActive='N'`) rather than deleted, a
retired row outlives the instruction that once carried it, and an earlier iteration reached for
`ON DELETE CASCADE` on the allocation's three foreign keys so those leftovers would stop blocking a
parent delete.

That was the wrong instrument. A cascade sees one undifferentiated set of child rows, so it erases a
**live** booking exactly as readily as retired history - silently, and with nowhere to put a check.
Deleting the parent is the moment where the live-vs-retired distinction matters most and the one place
the database cannot make it. The FKs are therefore back to plain `NO ACTION`, and the distinction is
made in application code:

- `interceptor/M_ShippingPackage` refuses the delete while **any** allocation points at the package,
  naming the instruction to cancel instead. Guarding the package also guards the instruction, because
  `PO.delete0()` runs `beforeDelete()` before it fires `TYPE_BEFORE_DELETE` and
  `MMShipperTransportation.beforeDelete()` force-deletes all of its package lines - so a separate guard
  on the instruction would be unreachable code.
- `interceptor/M_Delivery_Planning` refuses while any allocation is live, then removes the remaining
  retired rows itself. This is the one leg where a delete is legitimate: deleting a shipment or receipt
  schedule cascades into its plannings.

The guard is scoped by construction rather than by a filter - only delivery planning creates
allocations, so the transport-order and handling-units packages that share `M_ShippingPackage` never
match it and stay exactly as deletable as before.

Replaces the 1:1 link `M_ShipperTransportation.M_Delivery_Planning_ID` with an allocation row,
`M_Delivery_Planning_Alloc`, so one delivery instruction can carry several delivery plannings.
Modelled on `M_InOutLine_To_C_Customs_Invoice_Line` / `M_ReceiptSchedule_Alloc`: it holds the
planning, the instruction and the planning's shipping package, and nothing else. It carries no
`DocStatus`/`Processed` of its own — an allocation is not a document. Void, remove and move all
**deactivate** it (`IsActive='N'`); only deleting the planning itself deletes allocations. That is
what gives the history tab below something to show. It holds **no quantity** — a planning is one load, so
the link splits nothing.

### New AD_Table — how `M_Delivery_Planning_Alloc` is reachable (REVIEW.md § "New AD_Table")

`M_Delivery_Planning_Alloc` ships **without a window of its own**, deliberately: it is a link table,
never edited directly. It is reachable in two places, both read-only:

- **A history tab on the Delivery Instructions window (541657)** — `AD_Tab 549416`, over a view of
  **inactive** allocation rows, i.e. the plannings that were on this instruction and have since been
  re-booked away. Sorted newest-removed first. Before this, a retired allocation appeared in no UI
  surface at all: both shipped views join `IsActive='Y'`.
- **Related Documents** — `AD_RelationType 540505` surfaces the instruction's **currently allocated**
  plannings, mirroring the existing `C_Order_PO ↔ M_ShipperTransportation` relation pair.

The instruction window is therefore back to its pre-branch two-tab shape: lines plus history.
`AD_Tab 546754` is **parked** — still attached to `AD_Window 541657` at `SeqNo 30` (`ad_tab.ad_window_id`
is `NOT NULL`, so a tab cannot be detached), but set `IsActive='N'`, which is the park mechanism — reserved
for the multi-leg / N:N display it was reactivated in anticipation of.

### Carve-outs that ride along

Four, because the feature is wrong without them:
- the three-valued direction replacing `IsB2B` + `IsSOTrx`;
- `M_ShippingPackage.C_Order_ID` so the bill-of-lading date chain works, plus the null-guard;
- the closed-planning guards across all delivery-planning processes;
- the printed instruction rework + German translations.

### Explicitly deferred (not in this PR)

Consolidating a dropship's two plannings into one instruction. The three-valued direction (incl.
`Dropship`) ships; the consolidation does not. The sales-side planning keeps being created and the
paired shipment still generates with the receipt, exactly as before.

### Cross-repo impact — checked, none found

Two shared report functions in `de_metas_endcustomer_fresh_reports` changed shape:
`docs_deliveryinstructions_description` lost `orderno` / `referenceno` from its `RETURNS TABLE`, and
`docs_deliveryinstructions_productdetails` went from `(numeric)` to `(numeric, character varying)` and
gained those two columns. The in-repo JRXMLs were updated in lockstep. Because a customer repository
carrying its own copy of either JRXML would break on every delivery-instruction PDF, the whole
organisation was scanned rather than assumed:

- **43** repositories carry a `src/main/jasperreports/de/metas/docs` override tree; **none** contains a
  `deliveryinstructions` subdirectory.
- Code search for either function name returns hits only in this repository.

Limitation worth stating: the scan reads each repository's default branch, so an override living only
on a non-default branch would not have been seen.

### Report SQL — reconciliation (REVIEW.md § Reporting SQL)

`docs_deliveryinstructions_productdetails` now aggregates across the N plannings of one instruction
(`SUM(dp.plannedloadedquantity)`, `SUM(dp.qtyordered)`, grouped by product **and UOM**). Row
multiplication is impossible by construction: `M_Delivery_Planning_Alloc_Planning_UQ` is unique on
`M_Delivery_Planning_ID WHERE IsActive='Y'`, so a planning contributes to at most one active
allocation and each planning is summed once.

The UOM is part of the grouping deliberately. Grouping by product alone summed across different units
— proven on live data, where 4 Ea and 6 kg of one product printed as a single row reading qty 10
labelled "Ea". Differing units now print as separate lines rather than being silently added.

### Verification

- Every migration applied and verified by query against a local stack carrying the full migration chain.
- 194 unit tests in `de.metas.deliveryplanning.base`, up from 3 at branch point.
- The conditional read-only dates were verified on a **running instance**, not from metadata: the same
  planning reads read-only while an allocation is active, editable once it is deactivated, and
  read-only again once restored.
- The history tab was verified on a running instance: the instruction window serves exactly two child
  tabs, and the tab returns only inactive allocations, newest-removed first.
- `DateRemoved` is stamped once at deactivation and proven immune to later unrelated writes on the
  same row.
- **Cucumber coverage is complete for this iteration.** 34 scenarios across nine feature files —
  aggregation (combine, add-to, move, remove, draft-by-default), the delivery-instruction lifecycle
  cascade, and the closed-planning guards. Each was mutation-proven: its subject broken to show the
  scenario fails, then restored and re-run green.
- **Two closed-planning guards cannot be covered by cucumber and are manual instead** — receipt and
  shipment generation, whose guards live in a module that is not on the cucumber classpath. They are
  unit-covered at their own layer, and the end-to-end claim is evidenced by manual verification.
  (An earlier revision of this description said *three* guards and cited a classpath check. That was
  wrong: the shortage-overage guard sits in a module cucumber does reach, so it is coverable and a
  scenario is owed. Corrected rather than left standing.)
- Phase F is complete: dates, direction/dropship and the consignment/history view invariants all
  landed, and the existing delivery-planning features were kept passing throughout.
- Five checks are manual UAT rather than automated (the frontend suite is parked): allocated-planning
  dates read-only, the AddTo picker's contents, the grid filters and their defaults, the German
  printed instruction, and process visibility per document type.


backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5820480_sys_M_Delivery_Planning_Type_direction_caption.sql
```diff
@@ -0,0 +1,67 @@
...
+--   AD_MigrationScript 5820480 (this file). No new AD row is created, so no other ID is needed:
+--   AD_Element 540579 and AD_Field 708076 both already exist, and the AD_Element_Link row is
+--   (re)created by AD_Element_Link_Create_Missing_Field, which allocates its own id.
+
+-- ============================================================================
+-- AD_Window 541632 / AD_Tab 546674 "Lieferplanung" - AD_Field 708076
+-- ============================================================================
+-- Column: M_Delivery_Planning.M_Delivery_Planning_Type
+UPDATE AD_Field SET AD_Name_ID=540579, Description=NULL, Help=NULL, Name='Richtung',Updated=TO_TIMESTAMP('2026-08-27 01:00:00','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Field_ID=708076
+;
```
mrie04: why ad_name_id here? I don't see a reason to not name it the same in all 3 windows. looks like we don't need this script

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5821150_sys_M_Delivery_Planning_Filterability_And_Addresses.sql
```diff
@@ -0,0 +1,300 @@
...
+WHERE l.IsActive='Y' AND (l.IsSystemLanguage='Y' OR l.IsBaseLanguage='Y') AND t.AD_Element_ID=585385
+  AND NOT EXISTS (SELECT 1 FROM AD_Element_Trl tt WHERE tt.AD_Language=l.AD_Language AND tt.AD_Element_ID=t.AD_Element_ID)
+;
+
+UPDATE AD_Element_Trl SET Name='Loading Address', PrintName='Loading Address', Description='The address where goods are loaded for this delivery planning.', Help='The address where goods are loaded for this delivery planning.', IsTranslated='Y', Updated=TO_TIMESTAMP('2026-08-28 09:00:11','YYYY-MM-DD HH24:MI:SS'), UpdatedBy=100 WHERE AD_Element_ID=585385 AND AD_Language='en_US'
+;
+/* DDL */ select update_TRL_Tables_On_AD_Element_TRL_Update(585385,'en_US')
+;
+
+UPDATE AD_Element_Trl SET IsTranslated='N', Updated=TO_TIMESTAMP('2026-08-28 09:00:12','YYYY-MM-DD HH24:MI:SS'), UpdatedBy=100 WHERE AD_Element_ID=585385 AND AD_Language='de_DE'
```
mrie04: why transladed is set to N instead of translationg?

backend/de.metas.business/src/main/java/de/metas/handlingunits/impl/CreateShipperTransportationRequest.java
```diff
@@ -61,5 +60,14 @@ public class CreateShipperTransportationRequest
 	 */
 	boolean assignAnonymouslyPickedHUs;
 
-	SOTrx isSOTrx;
```
mrie04, line 64: why this is still here instead of transportationDirection

backend/de.metas.business/src/main/java/de/metas/handlingunits/impl/ShipperTransportationQuery.java
```diff
@@ -52,6 +52,13 @@ public class ShipperTransportationQuery
 	@NonNull @Singular Collection<OrderLineId> orderLineIds;
 	@Nullable Boolean processed;
 
+	/**
+	 * One of {@code X_M_ShipperTransportation.TRANSPORTDIRECTION_*}. Unset matches any direction -
+	 * a transport order's direction (Incoming/Outgoing/Dropship) is not implied by any of this query's
+	 * other fields, so a caller that cares which direction it matches must say so explicitly.
+	 */
+	@Nullable String transportDirection;
```
mrie04: why string and not REfListAwareEnum

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningService.java
```diff
@@ -519,13 +925,743 @@ public void generateDeliveryInstructions(final IQueryFilter<I_M_Delivery_Plannin
...
+	public void addTo(
+			@NonNull final IQueryFilter<I_M_Delivery_Planning> selectedDeliveryPlanningsFilter,
+			@NonNull final ShipperTransportationId targetDeliveryInstructionId)
+	{
+		final DeliveryPlanningList selectedDeliveryPlannings = getBySelection(selectedDeliveryPlanningsFilter);
+		if (selectedDeliveryPlannings.isEmpty())
+		{
+			// an invariant, not a user-facing rejection: the process's precondition already refuses an empty
+			// selection, and every rejection a planner can actually provoke is a translated message
+			throw new AdempiereException("No delivery planning selected");
```
mrie04: Why not Check.assumeNotEmpty

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningService.java
```diff
@@ -519,13 +925,743 @@ public void generateDeliveryInstructions(final IQueryFilter<I_M_Delivery_Plannin
...
+		final ImmutableList<DeliveryPlanningId> deliveryPlanningIds = selectedDeliveryPlannings.stream()
+				.filter(deliveryPlanning -> !targetDeliveryInstructionId.equals(deliveryPlanning.getDeliveryInstructionId()))
+				.map(DeliveryPlanning::getId)
+				.collect(ImmutableList.toImmutableList());
+		if (deliveryPlanningIds.isEmpty())
+		{
+			return;
 		}
+
+		trxManager.runInThreadInheritedTrx(() -> {
```
mrie04: why trxManager is needed? This kind should normally only be needed, if called from outOfTrx

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5820480_sys_M_Delivery_Planning_Type_direction_caption.sql
mrie04: **Resolution (dev):** Agreed, and investigating removal. `AD_Column 585005` and `593410` both point at `AD_Element 585383`, which already carries the "Richtung"/"Direction" caption — and `5820620:33` later NULLs `AD_Name_ID` on fields 783020/783021 for exactly that reason. So this per-field override is both redundant and inconsistent with its own siblings on this branch. A fix round is verifying that the caption is identical across all three windows with the override removed, and deleting the script if so. If any part turns out to still be needed I will say precisely which and why rather than keeping it silently.

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5821150_sys_M_Delivery_Planning_Filterability_And_Addresses.sql
mrie04: **Resolution (dev):** No change — this is the base-language convention, and the row is not missing a translation. `de_DE` is the base language (`AD_Language.IsBaseLanguage='Y'`), so its `AD_Element_Trl` row is the authored original rather than a translation of anything, and `IsTranslated='N'` is what the corpus uses for base-language rows — verified DB-wide at ~88% of them. The `en_US` row for the same element does carry `IsTranslated='Y'` with real English text, which is the row that represents actual translation work.

Worth noting this cuts the other way too: earlier on this branch these same rows were wrongly set to `'Y'`, and that was corrected — a base-language row marked `'Y'` claims a translation nobody made. A fix round is re-confirming against the DB that every language which should carry a real translation actually has one; if any is missing that is a defect and I will report it rather than close this out.

backend/de.metas.business/src/main/java/de/metas/handlingunits/impl/CreateShipperTransportationRequest.java
mrie04: **Resolution (dev):** You are right, and this is worse than a naming leftover — it is a functional limitation. `ShipperTransportationDAO.create()` derives the direction from this boolean, so **this path can only ever produce Incoming or Outgoing and never Dropship**. A two-valued field cannot express the three-valued domain the branch introduced.

The reason it survived: the `TransportDirection` enum lives in `de.metas.deliveryplanning.base`, and `de.metas.business` does not depend on that module (`grep -c deliveryplanning backend/de.metas.business/pom.xml` → 0), so it was unreachable from here. That is an explanation, not a justification.

Being fixed together with the `ShipperTransportationQuery` comment, since both have the same root cause: the enum describes `M_ShipperTransportation`, whose model and DAO live in `de.metas.business`, so it arguably belongs there. A fix round is assessing moving it down and converting both fields to the enum. If the move turns out not to be clean I will report exactly what blocks it and propose the next-best option rather than leaving a boolean here.

One property to preserve in whatever replaces it: it is a boxed `@NonNull Boolean` on purpose, so a caller that omits it fails loudly in `build()` instead of silently defaulting to `false` → Incoming — which is the same reason the column has no database default.

backend/de.metas.business/src/main/java/de/metas/handlingunits/impl/ShipperTransportationQuery.java
mrie04: **Resolution (dev):** Agreed. Same root cause as the `CreateShipperTransportationRequest.isSOTrx` comment, so both are being fixed as one change rather than separately.

The `TransportDirection` `ReferenceListAwareEnum` exists, but in `de.metas.deliveryplanning.base`, and `de.metas.business` does not depend on that module — so it was unreachable from this class and a `String` was used instead. That is the explanation for how it got here, not a defence of it: a raw `String` gives up the type safety and the ref-list validation the enum exists to provide.

A fix round is assessing moving the enum into `de.metas.business`, where the table it describes already lives, and converting this field and the request's boolean to it. If moving it creates a cycle or is otherwise not clean I will report precisely what blocks it and propose the next-best option, rather than leaving a `String` in place quietly.

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningService.java
mrie04: **Resolution (dev):** Agreed, adopting `Check.assumeNotEmpty`. The code's own comment says this is *"an invariant, not a user-facing rejection"* — the process precondition already refuses an empty selection and every rejection a planner can actually provoke is a translated message — which is exactly the case `Check.assume*` exists for. Throwing a bare `AdempiereException` with an untranslated literal both understates that it is an invariant and reads like a user-facing message that simply forgot its `AD_Message`.

A fix round is applying it and sweeping the class for the same shape elsewhere, so this does not get corrected in one spot and left standing in three.

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningService.java
mrie04: **Resolution (dev):** Likely right, and being removed unless a caller proves otherwise. The one caller I have traced is `M_Delivery_Planning_AddToDeliveryInstruction.doIt():118`, a `JavaProcess` that already runs inside a transaction — so `runInThreadInheritedTrx` there is a no-op wrapper that mostly obscures the fact that the block is already transactional.

A fix round is enumerating every caller of the enclosing method to check whether any genuinely arrives out-of-transaction. If none does, the wrapper goes. If one does, it stays and I will name that caller in a comment, so the next reader gets the answer instead of asking the same question — which is the outcome I would rather have than a silent removal that a future out-of-trx caller quietly breaks.

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5821150_sys_M_Delivery_Planning_Filterability_And_Addresses.sql
mrie04: **Resolution (dev):** Correcting my previous reply — it was wrong, and the row should be `'Y'`.

`IsTranslated` marks that a trl row **is correct for its language**, not that it needed translating. The decisive evidence, which I should have consulted before answering: `SetBaseLanguage.sql` — the function that *changes* the base language — writes `'Y'` into every `_Trl` row it materialises (line 21, `now(), 0, now(), 0, ''Y'', bt.IsActive`). The base language is configuration and can change; a flag whose correctness depends on which language happens to be base today is being used wrongly.

My "88% of rows DB-wide" was prevalence mistaken for authority. It measures how the field *is populated*, which is silent on what it *means* — a field set wrongly for years is 88% wrong, not 88% right.

Scope is wider than this line: these rows were originally `'Y'`, an earlier review on this branch "corrected" them to `'N'` on the same false premise, and I ratified it. A fix round is now sweeping **every** `_Trl` row this branch writes or updates and restoring `'Y'`, and will report the full list. I had also queued a corpus rule built on the false claim — that is now marked rejected so it cannot be adopted.

backend/de.metas.business/src/main/java/de/metas/shipping/api/impl/ShipperTransportationDAO.java
```diff
@@ -130,20 +129,31 @@ public ShipperTransportationId create(@NonNull final CreateShipperTransportation
 		final I_M_ShipperTransportation shipperTransportation = newInstance(I_M_ShipperTransportation.class);
 
 		shipperTransportation.setAD_Org_ID(request.getOrgId().getRepoId());
+		shipperTransportation.setTransportDirection(deriveTransportDirection(request));
```
mrie04: as of other request, should be correctly provided and not derived here, same on other calls, dervie function should go

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
```diff
@@ -294,7 +295,7 @@ private void addPurchaseOrderLines(final @NonNull I_M_ShipperTransportation ship
 	 */
 	private void applyDefaultDatesFromFirstOrder(@NonNull final I_M_ShipperTransportation shipperTransportation, @NonNull final I_C_Order order)
 	{
-		if (shipperTransportation.isSOTrx())
+		if (X_M_ShipperTransportation.TRANSPORTDIRECTION_Outgoing.equals(shipperTransportation.getTransportDirection()))
```
mrie04: TransportDirection.ofCode(...).isOutgoing()

backend/de.metas.business/src/main/java/de/metas/handlingunits/impl/CreateShipperTransportationRequest.java
mrie04: **Resolution (dev):** Done, with one correction to my own reply above.

Landed in https://github.com/metasfresh/metasfresh/commit/6c4be9436b7. `TransportDirection` moved down to `de.metas.business` (`de/metas/shipping`, next to `ShipperTransportationDAO`), and this field is now `@NonNull TransportDirection transportDirection`.

The move is down the existing chain — `de.metas.deliveryplanning.base` → `de.metas.handlingunits.base` → `de.metas.business` — verified in the poms; `de.metas.business` depends on neither of the other two, so no cycle. The enum now reads its codes off `X_M_ShipperTransportation`; both columns share `AD_Reference_Value_ID=541689`, and `TransportDirectionTest` pins that equality so a migration touching only one model fails there instead of silently splitting the domain.

**Correction:** I called this "worse than a naming leftover — a functional limitation". That overstates it. The expressiveness limit is real — a boolean cannot carry a three-valued domain — but no live caller is affected: `Dropship` applies only to the purchase-side leg, which is generated exclusively through the new repository path and never through this legacy DAO. So this was a latent limit, not a reachable defect, and the change is justified on type-safety grounds rather than on fixing a live bug.

backend/de.metas.business/src/main/java/de/metas/handlingunits/impl/ShipperTransportationQuery.java
mrie04: **Resolution (dev):** Done.

Landed in https://github.com/metasfresh/metasfresh/commit/6c4be9436b7 together with the `CreateShipperTransportationRequest` change, since both had the same cause. This field is now `TransportDirection` rather than a bare `String`, so a direction filter is no longer an unchecked literal. `ShipperTransportationDAO.deriveTransportDirection()` is gone — the direction is provided rather than derived, which also answers the separate comment on `ShipperTransportationDAO.java:132`.

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningService.java
mrie04: **Resolution (dev):** Done.

Landed in https://github.com/metasfresh/metasfresh/commit/88211003dfe.

Applied as a class, not only at this line: every site on this branch where an invariant was stated as a bare `AdempiereException` with an untranslated literal was reviewed and converted where `Check.assume*` is the right tool, leaving genuine user-facing rejections as translated `AD_Message`-backed exceptions.

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningService.java
mrie04: **Resolution (dev):** Done.

Landed in https://github.com/metasfresh/metasfresh/commit/88211003dfe. Two `runInThreadInheritedTrx` wrappers removed after enumerating the callers: each already runs inside a transaction, so the wrapper was a no-op that obscured the fact that the block was already transactional.

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5820480_sys_M_Delivery_Planning_Type_direction_caption.sql
mrie04: **Resolution (dev):** Done. Script removed in https://github.com/metasfresh/metasfresh/commit/53a25c83145. Your read was right: the per-field `AD_Name_ID` override was redundant with the element's own caption and inconsistent with its siblings on this branch.

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningList.java
```diff
@@ -0,0 +1,270 @@
...
+		}
+		return false;
+	}
+
+	/**
+	 * Everything the delivery-instruction header can hold only one of, and which therefore has to match across
+	 * the whole selection. Derived from the header columns the generation writes, so a column the header holds
+	 * once joins the rule by construction.
+	 */
+	public enum AdmissibilityField
```
mrie04: why not call it key or even aggregationkey, because this is what it's

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/process/M_Delivery_Planning_GenerateDeliveryInstruction.java
```diff
@@ -51,9 +57,9 @@ public ProcessPreconditionsResolution checkPreconditionsApplicable(@NonNull fina
 		{
 			return ProcessPreconditionsResolution.reject(msgBL.getTranslatableMsgText(DeliveryPlanningService.MSG_M_Delivery_Planning_NoForwarder));
 		}
-		final boolean isExistsOpenDeliveryPlannings = deliveryPlanningService.isExistsOpenDeliveryPlannings(selectedDeliveryPlanningsFilter);
+		final DeliveryPlanningList selectedDeliveryPlannings = deliveryPlanningService.getBySelection(selectedDeliveryPlanningsFilter);
```
mrie04: Consider DeliveryPlanningProcessHelper for duplicated Preconditions across procesesses

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningRepository.java
```diff
@@ -301,11 +382,9 @@ public void generateDeliveryPlanning(@NonNull final DeliveryPlanningCreateReques
 		deliveryPlanningRecord.setReleaseNo(request.getReleaseNo());
 		deliveryPlanningRecord.setTransportDetails(request.getTransportDetails());
 
-		deliveryPlanningRecord.setIsB2B(request.isB2B());
-
 		deliveryPlanningRecord.setM_MeansOfTransportation_ID(MeansOfTransportationId.toRepoId(request.getMeansOfTransportationId()));
 		deliveryPlanningRecord.setOrderStatus(OrderStatus.toCodeOrNull(request.getOrderStatus()));
-		deliveryPlanningRecord.setM_Delivery_Planning_Type(DeliveryPlanningType.toCodeOrNull(request.getDeliveryPlanningType()));
+		deliveryPlanningRecord.setTransportDirection(TransportDirection.toCodeOrNull(request.getTransportDirection()));
```
mrie04: not nullable anmyore

backend/de.metas.business/src/main/java/de/metas/shipping/api/impl/ShipperTransportationDAO.java
mrie04: **Resolution (dev):** Done — `deriveTransportDirection()` is gone, in https://github.com/metasfresh/metasfresh/commit/6c4be9436b7.

`ShipperTransportationDAO.create()` and `getOrCreate()` now take the direction from the request instead of deriving it. That was only possible once `TransportDirection` moved into `de.metas.business` (`de/metas/shipping`), which the same commit does — before that the type lived in `de.metas.deliveryplanning.base`, which `de.metas.business` does not depend on, so this class could not name it and fell back to deriving from a boolean.

The follow-up https://github.com/metasfresh/metasfresh/commit/c32a906b1ed retired the last two raw `X_M_ShipperTransportation.TRANSPORTDIRECTION_*` string comparisons, so the whole module is now on the enum rather than on codes.

backend/de.metas.business/src/main/java/de/metas/shipping/PurchaseOrderToShipperTransportationService.java
mrie04: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/c32a906b1ed.

One thing checked before applying it, since the swap is only safe if the partition is identical: this line must return early for `Outgoing` **only**, leaving `Incoming` *and* `Dropship` to fall through to the purchase-side defaults. `isOutgoing()` is `this == Outgoing`, so that holds exactly — had the helper grouped `Dropship` with `Outgoing`, the swap would have silently changed behaviour on dropship orders.

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningRepository.java
mrie04: **Resolution (dev):** Correct, fixed.

`DeliveryPlanningCreateRequest.transportDirection` is `@NonNull` (line 73), so `toCodeOrNull` was both unnecessary and actively misleading — it advertises a nullable domain that the request type rules out. Now calls `request.getTransportDirection().getCode()` directly.

Worth noting the same reasoning does *not* apply to the `OrderStatus.toCodeOrNull(...)` line just above it: that request field genuinely is nullable, so its `…OrNull` call is correct and stays.

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningList.java
mrie04: **Resolution (dev):** Adopted — renamed in https://github.com/metasfresh/metasfresh/commit/0194c7bde0c.

`AdmissibilityField` → `AggregationKeyField` (36 references, 9 files), `AdmissibilityFieldLabels` → `AggregationKeyFieldLabels`. You are right that this *is* the key: it is precisely the tuple that has to be identical for two plannings to end up on one instruction, and naming it after the check obscured that.

One deliberate limit on the rename: `admissibilityMismatches` and its siblings are **not** renamed. Those name the *check result* — which key fields failed to match — and they stay accurate read as "the aggregation-key fields that mismatched during the admissibility check". Renaming those too would collapse the distinction between the key (a property of the selection) and the check (an operation on it), which seemed worth keeping. Say the word if you'd rather they moved as well.

No migration needed: the `AD_Message` keys under `de.metas.deliveryplanning.CombineIntoDeliveryInstruction.Field.*` are built from the enum *constant* names, not the enum type name, so the label lookup is untouched.

Verified: `de.metas.deliveryplanning.base` install and `de.metas.cucumber` test-compile both green under Java 8 — the rename reaches four test classes and two step-def files, so the compiler covers every reference.

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/process/M_Delivery_Planning_GenerateDeliveryInstruction.java
mrie04: **Resolution (dev):** Agreed in principle, and I measured it before answering — the duplication is real but shallower than it looks, so I'd like your call on whether it earns a change on this PR.

**What is actually shared, across 9 process classes** (`AddTo`, `RemoveFrom`, `Close`, `ReOpen`, `Combine`, `Generate`, `ReGenerate`, `CancelDeliveryInstruction`, `CreateAdditionalLines`):

1. the `context.isNoSelection()` → `rejectBecauseNoSelection()` guard — byte-identical in all 9;
2. the selection-loading line `deliveryPlanningService.getBySelection(context.getQueryFilter(I_M_Delivery_Planning.class))` — in most of them.

That is roughly 4 lines × 9. **Everything after it genuinely differs per process** and is the part that carries the business rules — `Close` checks `anyOpen` / release-no / blocked-partner; `AddTo` checks `MAX_SELECTION_SIZE` and then delegates to `getAddToRejectionReason`. So a helper would factor out the boilerplate but leave the interesting half where it is, and it would need a parameter for the max-selection guard that only some processes apply.

My read: worth doing, but it is a ~9-file refactor whose payoff is boilerplate removal rather than a correctness or duplication-of-*logic* win. On a branch this size I'd rather not fold it in silently.

**Two options, your call:**
- **This PR** — the seam is obvious and the compiler covers it; adds a 9-file diff and another review round.
- **Next iteration** — alongside the `DeliveryPlanningRepository` split you've already taken, which touches the same area and would make one coherent cleanup rather than two overlapping ones.

I lean to the second for exactly that reason, but I'm not deciding it unilaterally.

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningService.java
```diff
@@ -392,6 +581,130 @@ public boolean isExistDeliveryPlanningsWithReleaseNo(final IQueryFilter<I_M_Deli
...
+				.type(transportDirection)
+				.shipperId(ShipperId.ofRepoIdOrNull(record.getM_Shipper_ID()))
+				.incotermsId(IncotermsId.ofRepoIdOrNull(record.getC_Incoterms_ID()))
+				.incotermLocation(record.getIncotermLocation())
+				.meansOfTransportationId(MeansOfTransportationId.ofRepoIdOrNull(record.getM_MeansOfTransportation_ID()))
+				.loadingLocationId(extractShipFromLocationIdOrNull(record, transportDirection, addresses))
+				.deliveryLocationId(extractShipToLocationIdOrNull(record, transportDirection, addresses))
+				.etd(TimeUtil.asInstant(record.getETD()))
+				.closed(record.isClosed())
+				.deliveryInstructionId(allocatedInstructionIds.get(deliveryPlanningId))
```
mrie04: why not ImmutableList.of(DeliveryPlanningAlloc). covers more usecases. IsAllocated can also be derived from it.

mrie04: **Resolution (dev):** Not changed yet — you asked to discuss this one before I continue, so this records the findings rather than a fix.

You are right that the single-id shape is the problem, and my first counter-argument was circular, so discarding it up front: I was going to point at the partial unique index `m_delivery_planning_alloc_planning_uq ON (m_delivery_planning_id) WHERE isactive='Y'`, which today makes planning→active-allocation strictly 1:1 at the DB level. But that index is exactly what multi-leg would remove, so it cannot justify baking 1:1 into the domain type.

**What the schema already supports.** `m_delivery_planning_alloc` is `(alloc_id, planning_id, shippertransportation_id, shippingpackage_id, lineno, isactive, dateremoved)` — N rows per planning, each pointing at a different instruction, is a natural fit. Only two things impose 1:1 today: that partial unique index, and `DeliveryPlanning.deliveryInstructionId` being a single `@Nullable ShipperTransportationId`.

**`IsAllocated` needs no work either way.** It is already a virtual `ColumnSQL` — `EXISTS (SELECT 1 FROM m_delivery_planning_alloc a WHERE a.m_delivery_planning_id = ... AND a.isactive='Y')` — so "has at least one active allocation" reads correctly for one leg or three. The domain mirrors it: `isAllocated()` is `deliveryInstructionId != null`, which becomes `!allocations.isEmpty()`.

**Why the list shape is the safer default regardless of when multi-leg lands:** cost asymmetry. Modelling a list when 1:1 turns out permanent costs a few `.stream()` calls. Modelling a single id when multi-leg arrives makes every consumer of `deliveryInstructionId` a migration site — touched at exactly the moment a functional change is also in flight.

**The part that is bigger than the field type**, and the reason this is worth talking through rather than just patching: today `AggregationKeyField.Forwarder` treats "same forwarder" as a property of the *selection*. With legs, forwarder is per-leg, so the admissibility rule becomes "same forwarder *for this leg*" and the key stops being a function of the planning alone. That is a design change, not a signature change.

Open question for the discussion: list of `DeliveryPlanningAlloc` (carries the shipping package and `lineno` too, which today force a re-query) or list of `ShipperTransportationId` (minimal change)?

mrie04: **Resolution (dev):** Done — https://github.com/metasfresh/metasfresh/commit/f74a1ddf3ea.

`DeliveryPlanning.deliveryInstructionId` → `@NonNull ImmutableList<DeliveryPlanningAlloc> allocations` (empty by default). New ids-only value object: alloc id, planning id, instruction id, shipping package id (non-null — verified against the column, which `I_M_Delivery_Planning_Alloc` documents `Mandatory: true`).

The two checks you named now come off an already-loaded list with no re-read: `getAllocationCount()` for how many, `getDeliveryInstructionIds()` (distinct) for which documents. `isAllocated()` is `!allocations.isEmpty()`.

It was load-bearing, not only misleading: `getAllocatedInstructionIds` collected with `ImmutableMap.toImmutableMap`, so a second active allocation would have thrown inside the loader. It is now an `ImmutableListMultimap` and duplicates group. Also fixed on the way: `getAllocatedTo` was *synthesizing* its map with `Maps.toMap(ids, id -> instructionId)`, which cannot produce alloc and package ids — it now reads them inside the query it already paid for, still two batch loads.

One semantic decision, argued from the code rather than preference: the non-draft-instruction rule is **ANY**, not ALL — a planning on one drafted and one completed instruction is refused by both add-to and remove-from, because the completed leg cannot be altered and that rule refuses outright rather than partially performing ("not partially performed, and not silently skipping the offending rows"). Identical to today wherever there is a single allocation. Pinned by tests, including a planning allocated to two different instructions — impossible under the current index, which is exactly why the semantics are pinned now rather than discovered later.

Untouched, deliberately: the `WHERE isactive='Y'` partial unique indexes, the `IsAllocated` virtual `ColumnSQL` (already `EXISTS`-based, correct for one leg or many), and the aggregation-key rules. `addTo` also remains a move — making it additive is a behaviour change multi-leg needs, with its own rules.

Verified: `de.metas.deliveryplanning.base` exit 0, Tests run 168 / Failures 0 / Errors 0 (was 160); `de.metas.cucumber` test-compile exit 0.

backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5821150_sys_M_Delivery_Planning_Filterability_And_Addresses.sql
mrie04: **Resolution (dev):** the sweep promised above is done, and it found more than this one row.

Every `AD_*_Trl` write on the branch was parsed statement by statement — 52 changed `.sql` files, 31 of them touching a `_Trl` table, 176 INSERT/UPDATE statements. Two findings:

1. `IsTranslated` on the German rows: exactly two deviations remained, both on `AD_Element` 585382 in `5820400`, carrying the same "German base text verbatim, so it stays 'N'" reasoning this thread rejected. Fixed in https://github.com/metasfresh/metasfresh/commit/8ced06d8962f9ad8177486ce35c4e68ab1985709 — the flag marks "this row's text is correct for this language", and `setBaseLanguage()` materialises base-language rows as `'Y'` for exactly that reason.

2. A second, wider defect surfaced while sweeping: 34 `_Trl` seed statements across 13 scripts selected their languages with `IsSystemLanguage='Y'` alone, omitting `OR IsBaseLanguage='Y'`. On an instance whose base language is not also flagged a system language, the base-language row is never seeded at all. All 41 Swing-exported seeds on the branch already used the correct form, which is what grounds it as the tooling's own convention rather than a preference. Fixed in https://github.com/metasfresh/metasfresh/commit/3bc3b8d8fcb633d2faf4961cad9dcf2c574e7f0e and https://github.com/metasfresh/metasfresh/commit/4610a2118372951aa9bda51754424a242f426481.

Note that `add_missing_translations()` does not cover for the second one: its base-language clause admits the base language for `AD_Element` only, so it closes an element row and none of the derived `AD_Column_Trl`, `AD_Index_Table_Trl` or `AD_Ref_List_Trl` rows.

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/process/M_Delivery_Planning_GenerateDeliveryInstruction.java
mrie04: **Resolution (dev):** answered by shipping it — the reply above asked for your call and then the branch made one, without coming back here. Closing that loop.

`DeliveryPlanningProcessHelper` now composes the shared preconditions rather than each process repeating them: https://github.com/metasfresh/metasfresh/commit/a220dc9aea4 ("compose process preconditions instead of repeating them"), across the process classes listed above.

A later round moved the parameter-defaults mechanism into the same helper as well, after a review found it had been copy-pasted byte-identically into the Add-to and Move-to processes: https://github.com/metasfresh/metasfresh/commit/3c064f86a55e336fb2ff69acbf103551212cbbbd. Lockstep sites for that mechanism went from two to one.

So the answer to the original question is yes, it earned the change — and the second round is the argument for why: the duplication was shallow when measured, but it reproduced itself the moment the mechanism grew.

PR #25643, opened 2026-08-27
Shipment-schedule priority derived from the shipper, behind a sysconfig

adi-stefan:
A shipment schedule can take its priority from its shipper instead of from its sales order, behind a system configuration that is **off by default**.

## Why

Picking work is ordered by the shipment schedule's priority. Where the carrier determines urgency — a same-day courier versus a standard parcel service — the sales order's own priority is the wrong source. This makes the shipper the source when switched on, and changes nothing when it is not.

## What changes

- **`M_Shipper.PriorityRule`** — a new optional field on the Lieferweg window (`AD_Window` 142, tab 185), reusing the shared priority element 522 and its value list (`AD_Reference` 17 / value 154: Dringend, Hoch, Mittel, Niedrig, Unwichtig). It may be left empty, and empty stays distinct from `5 Mittel` — hence nullable with no default, unlike `C_Order.PriorityRule` and `M_InOut.PriorityRule`, which are mandatory with default `5`.
- **`M_ShipmentSchedule_PriorityRuleFromShipper`** — the switch. `ConfigurationLevel = 'O'`, default `N`, so every existing installation behaves exactly as before.
- **The derivation** happens on the write path, in `OrderLineShipmentScheduleHandler.updateShipmentScheduleFromOrder` — the single place that writes the schedule's priority, and the one the recompute cycle re-runs. The stored column is what matters: the mobile picking launcher orders by `M_Packageable_V.PriorityRule` in SQL, so a read-side-only change would have had no effect where the feature is actually used.
- **A shipper edit propagates** — a model interceptor on `M_Shipper.PriorityRule` flags that shipper's unprocessed schedules for recompute.

## Behaviour

| Switch | Shipper | Resulting schedule priority |
|---|---|---|
| off | anything | the order's priority (unchanged from today) |
| on | has a priority | the shipper's priority |
| on | priority empty | the order's priority |
| on | no shipper | the order's priority |

`PriorityRule_Override` continues to win in every case — the read-side coalesce in `ShipmentScheduleEffectiveBL` is untouched.

## Out-of-list shipper priority codes

`M_Shipper.PriorityRule` has no physical CHECK constraint — none of the three sibling `PriorityRule` columns has one, and the value set is data-driven from the reference list, so the AD list constrains the WebUI dropdown but not the column. An import, a REST write or direct SQL can therefore store a code that is not in the list.

Such a code does **not** raise. `computePriorityRuleCode` logs a warning and falls back to the order's priority, exactly as it does when the shipper has no priority at all. Throwing here would be disproportionate: this method runs inside `ShipmentScheduleUpdater.updateSchedules`, which iterates a whole recompute batch with no per-item error handling, so a single bad shipper row would roll back the batch's transaction and take unrelated schedules down with it. Falling back also keeps the method symmetric, since the order's own priority code is returned unvalidated.

## Tests

- `OrderLineShipmentScheduleHandlerPriorityRuleTest` — 5 methods over the derivation: switch off, switch on with a shipper priority, shipper priority blank, no shipper, and an out-of-list code falling back.
- `ShipmentSchedulePA_ByShipperTest` — 2 methods over the new `retrieveUnprocessedIdsByShipperId`, distinguishing processed from unprocessed and one shipper from another.
- `M_Shipper_ShipmentScheduleTest` — 3 methods over the interceptor: a priority change flags the open schedule, a processed schedule is not flagged, an unrelated column change flags nothing.
- Full `de.metas.swat.base` suite: `Tests run: 533, Failures: 0, Errors: 0, Skipped: 9` (523 before this branch).
- `shipmentSchedulePriorityFromShipper.feature` — 5 cucumber scenarios covering all four switch/shipper combinations plus the recompute-on-shipper-change case. Verified non-vacuous: forcing the derivation's condition to `false` made the switch-on scenario fail with `expected: "1" but was: "7"`, and it passed again once reverted.

## Note for reviewers

The cucumber work extends three step-definition classes, not one. `M_Shipper_StepDef` gained `PriorityRule` support plus an `update M_Shipper:` step (the existing `contains M_Shippers` step uses `StepDefData.put()`, which rejects a reused identifier, so an in-place priority change had no route). `C_Order_StepDef` and `M_ShipmentSchedule_StepDef` gained `PriorityRule` handling because neither had any way to set the order's priority or assert the schedule's. All three changes are purely additive — no existing step's behaviour changed for any other column.


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/modelvalidator/InOutCandidateValidator.java
```diff
@@ -52,6 +52,7 @@ protected void onInit(final IModelValidationEngine engine, final I_AD_Client cli
 		engine.addModelValidator(new M_InOutLine_Shipment(), client);
 		//engine.addModelValidator(new M_InOut_Shipment(), client); // converted to spring bean
 		engine.addModelValidator(new C_BPartner_ShipmentSchedule(), client);
+		engine.addModelValidator(new M_Shipper_ShipmentSchedule(), client);
```
adi-stefan: Why not just declare the Class as `@Component`?

backend/de.metas.swat/de.metas.swat.base/src/test/java/de/metas/inoutcandidate/modelvalidator/M_Shipper_ShipmentScheduleTest.java
```diff
@@ -0,0 +1,139 @@
...
+	{
+		// given: a shipper with an unprocessed schedule, the interceptor registered only AFTER setup so
+		// the priority change below is the only thing that fires it
+		final I_M_Shipper shipper = createShipper();
+		final ShipmentScheduleId scheduleId = createShipmentSchedule(shipper, false);
+
+		registerInterceptorUnderTest();
+
+		// when: the shipper's PriorityRule changes
+		shipper.setPriorityRule("3");
```
adi-stefan: Minor: should use `de.metas.inout.PriorityRule` throughout the test

backend/de.metas.swat/de.metas.swat.base/src/test/java/de/metas/order/inoutcandidate/OrderLineShipmentScheduleHandlerPriorityRuleTest.java
```diff
@@ -0,0 +1,161 @@
...
+		// service stand in for them (IShipmentScheduleInvalidateBL's real impl needs a Spring context to construct).
+		handler = new OrderLineShipmentScheduleHandler(
+				Mockito.mock(IShipmentScheduleInvalidateBL.class),
+				new PickingBOMService(),
+				Optional.empty());
+	}
+
+	private static void setSwitch(final boolean value)
+	{
+		Services.get(ISysConfigBL.class).setValue(
```
adi-stefan: extract sysconfigBL as field

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/inoutcandidate/modelvalidator/InOutCandidateValidator.java
adi-stefan: **Resolution (dev):** Adopted — converted to the Spring-discovered form in https://github.com/metasfresh/metasfresh/commit/daeb08161d15b25e5cf7129597c706bc6b62b8fd. `M_Shipper_ShipmentSchedule` is now `@Interceptor(I_M_Shipper.class)` + `@Component`, and this manual registration line is removed.

I had originally followed the eight sibling registrations in this same `onInit`, which all use the legacy `@Validator` + manual form — but that was the wrong reference point: `de/metas/picking/modelvalidator/M_PickingSlot.java` in this same module is already `@Interceptor` + `@Component` with no registration anywhere, so the modern form is established here.

Registration still happens: `ModelValidationEngine.getSpringInterceptors()` collects `getBeansWithAnnotation(Interceptor.class)` at boot and calls `addModelValidator` for each — the same path that registers `M_PickingSlot`. The interceptor test needed no change (it registers the instance directly via `POJOLookupMap`, which scans `@ModelChange` methods reflectively) and still passes 3/3.

backend/de.metas.swat/de.metas.swat.base/src/test/java/de/metas/inoutcandidate/modelvalidator/M_Shipper_ShipmentScheduleTest.java
adi-stefan: **Resolution (dev):** Adopted in https://github.com/metasfresh/metasfresh/commit/dac5e2cc13b2be143b37773f53d50a0284808a45 — both `setPriorityRule("3")` literals now read `setPriorityRule(PriorityRule.High.getCode())`, with the `de.metas.inout.PriorityRule` import added. Confirmed `"3"` is `PRIORITYRULE_High` before swapping, so the test asserts exactly what it did before. No assertion changed; still 3/3 green.

backend/de.metas.swat/de.metas.swat.base/src/test/java/de/metas/order/inoutcandidate/OrderLineShipmentScheduleHandlerPriorityRuleTest.java
adi-stefan: **Resolution (dev):** Adopted in https://github.com/metasfresh/metasfresh/commit/bc64337cf22cd208218626c55cb9d869be0f2a33 — `ISysConfigBL` is now a field, and `setSwitch` reads it instead of calling `Services.get(...)` inline.

One ordering detail worth recording: the field cannot be initialised at declaration. `AdempiereTestHelper.get().init()` calls `Services.clear()` in `@BeforeEach`, so a construction-time `Services.get(...)` would capture a stale instance. It is therefore declared without an initialiser and assigned in `@BeforeEach` after `init()` — the same shape the existing `handler` field in this class already uses. Verified by running rather than by reasoning: 5/5 green.

PR #25647, opened 2026-08-27
ESR import: cover the duplicate-detection path that actually lost money

metas-ts:
## Summary

Test-only. Closes a coverage gap in the duplicate-payment detection that this feature's own fix depends on.

`ESRImportDAO.findExistentPaymentId` turns a candidate payment (same partner, same amount) into a duplicate via one of **three** routes:

1. the candidate is attached to **another `ESR_ImportLine` with the same `ESRLineText`** — the same raw bank line, seen before;
2. the candidate is this line's own payment and the line already carries `Duplicate_Payment` (idempotency on reprocessing);
3. `IPaymentBL.isMatchInvoice` — the candidate is **already allocated to the invoice this line resolves**.

Both existing tests (`arrivingOnTheSameDay`, `arrivingOnALaterDay`) give their two lines the **same** `ESRLineText`, so route 1 matches first and returns. **`isMatchInvoice` was therefore never evaluated by any test.**

The production incident took route 3: two **distinct** bank lines, where the earlier payment was already allocated to the very invoice the second line resolves (92.45 arriving 2026-08-11 against an invoice settled by a payment from 2026-06-19 — different dates, hence different raw line text). That is the case that left the newly arrived money unbooked, and it was the untested one.

## What the test does

Gives the second line a **different** line text — one differing character, same record length, so neither text contains the other and route 1 cannot match — while keeping the **payment date equal**, so the candidate query matches on every branch, including those that filter candidates on an exact `DateTrx`. The test asserts up front that the two texts differ, are the same length, and still resolve the same invoice reference, so it cannot silently stop testing what it claims to.

It then asserts the line is flagged `Duplicate_Payment`, stays **unprocessed**, and — the load-bearing one — gets its **OWN** payment, distinct from the first line's. Attaching the earlier payment is exactly the old money-losing behaviour.

## Validated by mutation, not just by a green run

A passing test proves little here, since the assertions could be satisfied via route 1. So the `isMatchInvoice` branch was temporarily disabled and the class re-run:

- **only this test failed** (1 of 4), at the payment-action assertion;
- the three siblings still passed.

That proves the new test is load-bearing for route 3, *and* independently confirms the siblings never exercised it. The mutation was reverted before committing.

## Test plan
- [x] `ESRImportTest` — 69 run, 0 failures, with the new test included.
- [x] Mutation probe as described above: `isMatchInvoice` disabled → this test alone fails.
- Note: a full-module run in the author's environment is currently unreliable (`.m2-local` holds `de.metas.*` jars built from an older base, so tests link against stale dependency jars — a known trap documented in the repo). `test (java)` in CI is green on this base and is the authoritative gate here.

### DIFF
--- a/backend/de.metas.payment.esr/src/test/java/de/metas/payment/esr/ESRImportTest.java
+++ b/backend/de.metas.payment.esr/src/test/java/de/metas/payment/esr/ESRImportTest.java
@@ -1770,33 +1770,16 @@ public class ESRImportTest extends ESRTestBase
 		}
 
 		/**
-		 * @param expectedPaymentAction the second line's expected {@code ESR_Payment_Action}, or {@code null} if it
-		 * is expected to stay unflagged (see {@link #arrivingOnALaterDay()}).
-		 */
-		/**
-		 * The detection path that actually lost money in production, and the one no other test reaches.
-		 * <p>
-		 * {@code ESRImportDAO.findExistentPaymentId} turns a candidate payment (same partner, same amount)
-		 * into a duplicate via one of three routes. Both sibling tests give the two lines the SAME
-		 * {@code ESRLineText}, so the "another line already carries this payment" route matches first and
-		 * returns -- {@code IPaymentBL.isMatchInvoice} is never evaluated by them.
-		 * <p>
-		 * The production case was different: two DISTINCT bank lines, and the earlier payment was already
-		 * allocated to the very invoice the second line resolves. Only {@code isMatchInvoice} can catch
-		 * that. This test isolates it by giving the second line a different line text (so the line-text
-		 * route cannot fire) while keeping the payment date equal (so the candidate query matches on every
-		 * branch, including those that filter candidates on an exact {@code DateTrx}).
-		 * <p>
-		 * The load-bearing assertion is that the second line gets its OWN payment: attaching the earlier
-		 * one is exactly the behaviour that left the newly arrived money unbooked.
+		 * Duplicate recognised because the earlier payment is already allocated to this line's invoice.
+		 * The sibling tests give both lines the same {@code ESRLineText}, so they never reach this route:
+		 * hence the differing line text here, and the equal payment date so the candidate query still
+		 * matches on branches that filter on an exact {@code DateTrx}.
 		 */
 		@Test
 		void earlierPaymentAllocatedToTheSameInvoice_isDetectedViaTheInvoice_notTheLineText()
 		{
 			final String grandTotal = "50";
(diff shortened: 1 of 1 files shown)


backend/de.metas.payment.esr/src/test/java/de/metas/payment/esr/ESRImportTest.java
```diff
@@ -1773,6 +1773,75 @@ void completingTheImportMustNotCloseTheFlaggedDuplicate()
 		 * @param expectedPaymentAction the second line's expected {@code ESR_Payment_Action}, or {@code null} if it
 		 * is expected to stay unflagged (see {@link #arrivingOnALaterDay()}).
 		 */
+		/**
+		 * The detection path that actually lost money in production, and the one no other test reaches.
```
metas-ts: please just **briefly** explain what the method does. javadoc's purpose is to cove when code is not clear enugh

metas-ts: Done — cut to four lines, keeping only what the code can't state itself: why the second line needs a **different** `ESRLineText` and the **same** payment date.

Your hunk also surfaced a real mistake I'd missed: my block had landed *between* the helper's javadoc and the helper, so there were two consecutive javadoc blocks — the `@param expectedPaymentAction` doc was stranded above the new test and `assertDuplicateGetsOwnPayment` was left undocumented. The test now sits above that javadoc, so it documents the helper again.

I also dropped the two in-test comments; the `.as(...)` assertion descriptions right below them already said the same thing.

Net: 8 insertions, 22 deletions. Pushed as 229cb480da6.

metas-ts: **Resolution (dev):** Javadoc cut from ~18 lines to 4, keeping only what the code cannot state itself — why the second line needs a **different** `ESRLineText` and the **same** payment date. The three-route explanation and the production narrative were reviewer context and live in the PR body instead.

Your hunk also surfaced a real defect: my block had landed *between* the helper's javadoc and the helper, leaving two consecutive javadoc blocks — the `@param expectedPaymentAction` doc was stranded above the new test and `assertDuplicateGetsOwnPayment` was left undocumented. The test now sits above that javadoc, so it documents the helper again.

Also dropped the two in-test comments; the `.as(...)` assertion descriptions below them already said the same thing.

Net 8 insertions / 22 deletions: https://github.com/metasfresh/metasfresh/commit/229cb480da6

Not re-run locally — this worktree's `.m2-local` holds a Java-17-compiled `de-metas-common-procurement` jar (`wrong version 61.0, should be 52.0`), so the module cannot build here. The change is comments and ordering only, with no behavioural edit; CI is the gate.

PR #25663, opened 2026-08-29
Accounting: skip posting silently when the document was deleted after the posting was scheduled

adi-stefan:
## Problem

A process that mass-creates and then deletes match documents (`M_MatchInv` / `M_MatchPO`) within the same run leaves behind the asynchronous posting requests that were scheduled when those documents were created. The deletion cannot cancel them across the async/commit boundary.

When such a stale request is finally executed, the document it references no longer exists. `AcctDocProviderTemplate.retrieveDocumentModel` then throws `No document found for ref{...}`, and `PostingService.postNow` turns that into a `posting_error` log entry, an `AD_Issue` and a user notification — **one per deleted document**. A single run produced several thousand of them, which buries any genuine posting error and makes the run look failed although it committed successfully.

A deleted document has no accounting facts left to post: `onAfterDeleted` already removes its `Fact_Acct` rows. Posting it is therefore a no-op, not an error.

## Change

`PostingService.postNow` now checks, **on the failure path only**, whether the referenced record still exists. If it does not, the posting is skipped with a `debug` log line — no `AD_Issue`, no user notification.

The existence query itself lives in a repository, so no persistence primitive is constructed in the service tier (`docs/coding-rules/service-injection.md` §4). The check is not posting-specific - it only needs the table's key column and the id - so it lives in the generic base module as `org.adempiere.ad.table.ReferencedRecordRepository`, a Spring `@Repository` beside `ChangeLogConfigRepository` / `ChangeLogEntryRepository` in the component-scanned `org.adempiere` package tree. `PostingService` is not itself a Spring bean, so it obtains the repository exactly as it already obtains its other Spring collaborators - `SpringContextHolder.lazyBean(ReferencedRecordRepository.class)`. Unit tests build it through the `newInstanceForUnitTesting()` factory, which registers the instance as a JUnit bean so `lazyBean` resolves it without a Spring context.

Notes on the design:

- The check sits in `postNow` rather than at the throw site in `AcctDocProviderTemplate`, because that template is only one of the `IAcctDocProvider` implementations (`Doc_InOut_Factory` implements the interface directly and fails differently when its record is gone). `postNow` is the single point every asynchronous posting request passes through, so one check covers every document type.
- Doing it on the failure path keeps the happy path free of an extra query, and makes it structurally impossible to change the outcome for a document that still exists — any such document takes exactly the path it took before.
- The existence check is deliberately **not** filtered to active records: a deactivated record still exists and must still be reported as a posting error.
- The decision to trust the check is fail-safe, and that policy deliberately stays in the service rather than the DAO. It runs inside the `catch` block, so an exception thrown by the check would otherwise replace the real posting error and escape the asynchronous posting handler; instead it logs a warning and answers "the document still exists", so the original error is reported unchanged.
- The exception that brought us onto the failure path is passed to the `debug` log on the skip branch, so it is not silently dropped (`docs/coding-rules/java-general.md` §53).

## Also in this PR

`POJOQuery.count()` (unit-test query framework) passed only the model class to `POJOLookupMap.getRecords(...)`, so a query built from a plain table name — i.e. with a generic `Object` model class — failed with `MissingTableNameException`, while its siblings `list()` / `first()` / `firstOnly()` already used the table-name aware overload. Fixed at the source rather than worked around at the call site.

## Tests

`PostingServiceTest` (new):

- a posting request whose record was deleted -> no posting-error log, no user notification;
- a posting request whose record exists but fails to post -> posting-error log **and** user notification, exactly as before;
- the same, when the existence check itself blows up -> the original posting error is still reported.

Each was verified to fail before the corresponding production change.

Local runs: `de.metas.acct.base` 70 tests / 0 failures; `de.metas.adempiere.adempiere/base` 1248 tests / 0 failures; cucumber `@F00750` (`Match Invoice / Receipt, Invoice => MatchInv`) 1/1 passed against a local stack.


backend/de.metas.acct.base/src/main/java/de/metas/acct/api/impl/PostingDocumentDAO.java
```diff
@@ -0,0 +1,26 @@
+package de.metas.acct.api.impl;
+
+import de.metas.acct.api.IPostingDocumentDAO;
+import de.metas.util.Services;
+import lombok.NonNull;
+import org.adempiere.ad.dao.IQueryBL;
+import org.adempiere.model.InterfaceWrapperHelper;
+import org.adempiere.util.lang.impl.TableRecordReference;
+
+public class PostingDocumentDAO implements IPostingDocumentDAO
```
adi-stefan: This is not a PostingDocumentDAO as much as a generic ReferencedDocumentDAO. I'm surprised something similar does not already exist in code.

adi-stefan: **Resolution (dev):** Agreed on the naming/placement — done in https://github.com/metasfresh/metasfresh/commit/ed7a80d4fdb95481bc6d41784a98acafd679c34e. The check is now `org.adempiere.ad.table.api.IReferencedRecordDAO` + `impl.ReferencedRecordDAO` in `de.metas.adempiere.adempiere.base`, next to `ITableRecordIdDAO`; `IPostingDocumentDAO` / `PostingDocumentDAO` are deleted and `PostingService` / `PostingServiceTest` repointed. `de.metas.acct.base` already depends on that module, so it cost nothing. I used "Record" rather than "Document" because the reference can point at any table (an `AD_User` is not a document) — happy to rename if you prefer your wording.

On "surprised something similar does not already exist": I checked, and it genuinely does not. There is exactly one `boolean exists(TableRecordReference)` in the backend and it is this one. The two nearest things:
- `InterfaceWrapperHelper.create(ctx, tableName, id, Object.class, trxName) != null` is null-tolerant and would work, but it materialises the whole PO into the model cache — strictly worse for this case, where the record was just deleted. `TableRecordReference.getModelNonNull(...)` throws, and the `getModel(...)` variants are `@Deprecated` in favour of "use the appropriate DAO".
- `de/metas/business_rule/util/BusinessRuleRecordMatcher.java:107-128` hand-rolls the identical `createQueryBuilder(tableName)` + `getKeyColumnName(...)` + `anyMatch()` shape, comment included ("make sure record exists"), just with an optional extra where-clause.

That second one is the natural second consumer and it already lives in the module the DAO moved into, so it can delegate now. I left it untouched to keep the hotfix diff tight — say the word and I will fold it in here, or file it as a follow-up.

Verified: `de.metas.acct.base` 70 tests / 0 failures (`mvn -pl de.metas.acct.base test`, JDK 8, local soft_panda stack, 2026-08-31) — same as before the move, `PostingServiceTest` deleted-document and failing-DAO cases included.

adi-stefan: **Follow-up (dev):** two further commits on this thread's subject, both pushed:

- https://github.com/metasfresh/metasfresh/commit/88579a5140bcf6611acbd8c32933f246cef1e685 — dropped the `ISingletonService` + `Services.get(...)` wiring. The interface `IReferencedRecordDAO` is gone; the class is now a plain Spring `@Repository` (`newInstanceForUnitTesting()` via `SpringContextHolder.getBeanOrSupply(...)`, `PostingService` holds it as a `SpringContextHolder.lazyBean(...)`). It sits in `org.adempiere.ad.table`, which both `ServerBoot` and `WebRestApiApplication` component-scan (`scanBasePackages = { "de.metas", "org.adempiere" }`), so it resolves in app and webapi alike.
- https://github.com/metasfresh/metasfresh/commit/0a7215207764b0d57a640fbe4fd72f3de2739141 — renamed `ReferencedRecordDAO` -> `ReferencedRecordRepository`, matching the `*Repository` naming its new neighbours use.

So the final shape is `org.adempiere.ad.table.ReferencedRecordRepository#exists(TableRecordReference)` in `de.metas.adempiere.adempiere.base` — generic, not posting-specific, which was your point.

Verified locally 2026-08-31: `de.metas.adempiere.adempiere/base` and `de.metas.acct.base` both compile under JDK 8; `ReferencedRecordRepositoryTest` 3/3 and `PostingServiceTest` 3/3 green.

The `BusinessRuleRecordMatcher` de-duplication mentioned above is still deliberately left out of this hotfix diff.

adi-stefan: **Correction to the note above:** the local verification I quoted was run against a working tree that still held an uncommitted edit, so commit https://github.com/metasfresh/metasfresh/commit/0a7215207764b0d57a640fbe4fd72f3de2739141 as pushed did **not** compile — the file was renamed to `ReferencedRecordRepositoryTest.java` while its content still declared `class ReferencedRecordDAOTest` and referenced the deleted `ReferencedRecordDAO`. CI caught it (`java` / `build-backend`, https://github.com/metasfresh/metasfresh/actions/runs/33448054196/job/99671589030).

Fixed in https://github.com/metasfresh/metasfresh/commit/2909af96eb3ab4335e6d422af66069969ce9453c, verified from a clean tree: `de.metas.adempiere.adempiere/base` compiles, `ReferencedRecordRepositoryTest` 3/3, `PostingServiceTest` 3/3. CI is re-running at https://github.com/metasfresh/metasfresh/actions/runs/33448430605.

adi-stefan: **Resolution (dev):** Settled — this thread's point is fully addressed and the branch is back in a verified-good state.

Final shape: `org.adempiere.ad.table.ReferencedRecordRepository` in `de.metas.adempiere.adempiere.base`, a plain Spring `@Repository` sitting beside `ChangeLogConfigRepository` / `ChangeLogEntryRepository`. No `ISingletonService`, no `Services.get(...)`, and no interface — there is exactly one implementation, so the interface earned nothing. `PostingService` obtains it via `SpringContextHolder.lazyBean(ReferencedRecordRepository.class)`, the same mechanism it already uses for its four other Spring collaborators. Unit tests build it through `newInstanceForUnitTesting()`, which self-registers via `getBeanOrSupply` so `lazyBean` resolves without a Spring context.

Your "generic, not posting-specific" reading drove all of it: the class is table-agnostic, lives in the generic base module rather than under the accounting module, and is now named after its siblings rather than after the caller that happens to need it.

Fix for the broken commit noted above: https://github.com/metasfresh/metasfresh/commit/2909af96eb3 completes the rename inside the test class. Re-verified against a clean working tree this time: the base module compiles, `ReferencedRecordRepositoryTest` 3/3, `PostingServiceTest` 3/3, and no reference to the old name remains anywhere in the tree.

On the surprise that nothing similar already existed — I looked, and there genuinely was no table-agnostic "does this record still exist?" helper; `ITableRecordIdDAO` is the closest neighbour and answers a different question.

PR #25675, opened 2026-08-31
Costing: keep the MatchInv cost-adjustment posting quantity-neutral and on the receipt locator

teosarca:
## Problem

`Doc_MatchInv.createFacts` posts a vendor MatchInv as up to four legs:

| leg | qty | locator |
|---|---|---|
| NotInvoicedReceipts (GR/IR) DR | `+qty` | receipt line's locator |
| P_InventoryClearing CR | `-qty` | – |
| P_Asset DR/CR (cost adjustment) | `+qty` | **none** |
| P_COGS DR/CR (already shipped) | `+qty` | none |

The first two are quantity-balanced against each other, so that pair disturbs no stock. The **`P_Asset` cost-adjustment leg has no offsetting negative**: every cost adjustment injects a spurious `+qty` at a NULL locator into `P_Asset`.

That matters because the inventory-valuation report (Lagerwert) derives stock on hand from exactly that column — `report_InventoryValue` computes `COALESCE(SUM(fa.qty), 0) AS Qty` over the `P_Asset` facts and then values it (`Costing_CostPrice * Qty`). The phantom quantities are therefore reported as stock that does not exist, and because the locator is NULL they are attributed to no warehouse.

The same defect class was already fixed on the manufacturing side: `Doc_PPCostCollector.addCostDifferenceFactLine` posts its cost-difference legs with `setQty(getMovementQty().toZero())` and `locatorId(docLine.getM_Locator_ID())`, carrying the comment *"The line carries a ZERO qty: the receipt already accounted for the quantity, so a qty here would be counted a second time by the inventory valuation (Lagerwert) report."* `Doc_MatchInv` never got the same treatment.

### Pre-existing, not introduced here

The defective line is old, but it was unreachable in practice: before https://github.com/metasfresh/metasfresh/pull/25665 the cost-adjustment amount came back zero, so the `if (!costs.getCostAdjustmentAmt().isZero())` guard never fired and the fact line was never created. That PR made the adjustment amount correct, which made this leg start being posted — and the phantom quantities visible.

## Fix

On the `P_Asset` cost-adjustment leg only:

- `setQty(...)` now gets a **zero** quantity (`qty.toZero()`, keeping the UOM) — matching the `Doc_PPCostCollector` idiom.
- `.locatorId(receiptLine.getM_Locator_ID())` is added, the same value the sibling GR/IR leg already uses.

The locator half is not cosmetic, and it has one consequence worth calling out explicitly: `FactLine.computeOrgIdEffective` prefers the **locator's** org (Prio 1) over the doc-line's (Prio 2). So on a **multi-org** setup — central invoicing org, distributed receiving orgs — this leg's `AD_Org_ID` changes from the invoice line's org to the receiving warehouse's, which is what the sibling GR/IR leg already does explicitly (`.orgId(receiptLine.getAD_Org_ID())`). I believe that is a correctness improvement (all stock-related legs of the document now agree on org) rather than a regression, but it is a behaviour change on multi-org installations and no existing test exercises a cross-org receipt-vs-invoice case — **please confirm it is intended.** On single-org installations the two are identical and nothing changes.

Amounts, accounts and debit/credit are unchanged — the GL is bit-identical; only `Fact_Acct.Qty` and `Fact_Acct.M_Locator_ID` on that one leg change.

To make this testable at the JUnit layer without a database, the leg construction is extracted into a `@VisibleForTesting static costAdjustmentLegs(...)` returning leg descriptors, and `createFacts` builds the fact lines from them. This follows the existing precedent in `Doc_PPCostCollector.costDifferenceDistributionLegs(...)` / `Doc_PPCostCollectorCostDifferenceDistributionTest`. The extraction is behaviour-preserving on its own.

## The `P_COGS` leg is deliberately unchanged

The `P_COGS` "already shipped" leg (`Doc_MatchInv`, immediately below the `P_Asset` one) has the **identical shape**: `setQty(getQty())`, no `locatorId(...)`, no offsetting negative. So the same argument may well apply to it.

It is **left untouched on purpose**, because:

- it is not what drives the reported problem — on the production-scale data set used to validate this change, the `P_COGS` legs account for 11 rows against 4121 `P_Asset` ones, and their amount is immaterial next to the `P_Asset` error;
- unlike `P_Asset`, `P_COGS` is not an inventory account, so whether a quantity there is meaningful (and which locator, if any, it should carry — the goods have already left) is a product decision, not an obvious bug.

**Evidence that points towards "yes, change it too":** `Doc_PPCostCollector` — the handler this PR takes its pattern from — does **not** treat its two legs differently. `createFactsForCostDifferenceDistribution` loops *every* leg it produces (P_Asset, P_COGS **and** P_WIP) through the single `addCostDifferenceFactLine`, which applies `setQty(getMovementQty().toZero())` and `locatorId(docLine.getM_Locator_ID())` uniformly — carrying the same Lagerwert double-counting rationale in its Javadoc. So the closest precedent in the codebase already gives `P_COGS` the zero-qty-plus-locator treatment. That is an argument for symmetry here, but it is the reviewer's call, not mine to make silently inside a hotfix.

**Decision (review round): leave `P_COGS` unchanged in this PR.** The `P_Asset` leg is what drives the reported inventory-valuation error; giving `P_COGS` the same treatment is a separate product decision and, if wanted, a one-line follow-up in the same extracted method. `Doc_MatchInvCostAdjustmentLegsTest.bothLegs_onlyTheAssetLegIsQuantityNeutral` now pins the current asymmetry, so it cannot change by accident, and the in-code comment states the reason inline rather than pointing at this pull request.

## Verification

**TDD, RED first.**

`Doc_MatchInvCostAdjustmentLegsTest` (new, `de.metas.acct.base`, no DB) — RED before the fix:

```
Multiple Failures (2 failures)
-- failure 1 -- [Qty]          expected: 0    but was: 10
-- failure 2 -- [M_Locator_ID] expected: 4711 but was: 0
```

GREEN after: `Tests run: 5, Failures: 0` (the class grew to 5 cases in the review round, incl. one pinning that `P_COGS` is genuinely untouched). Whole module: `Tests run: 70, Failures: 0, Errors: 0, Skipped: 1`.

`matchInvRepostPriceVariance.feature` (`@Id:S26253_TC23`) — the existing scenario's `P_Asset_Acct` expectation encoded the defective `+10 PCE`; it now asserts `0 PCE` plus the locator, on both the first posting and the repost. RED before the fix:

```
Expected Fact_Acct record was not found:
| P_Asset_Acct | 20 | | 20 CHF | | 0 PCE  | locatorStd | product | matchInv | vendor |
Fact_Acct record was not expected:
| P_Asset_Acct | 20 CHF | 0 CHF | 20 | 0 | 10 Stk | ... | M_Locator_ID <null> |
```

The GR/IR and InventoryClearing rows matched unchanged in the same run, which is the control: the fix touches only the cost-adjustment leg.

The change was additionally validated by replaying it over a production-scale copy: the orphan report rows disappear, the reported inventory-valuation error drops from roughly +26% of the reference costing method to under +1%, the total inventory **value** is unchanged (control), the GL is identical, and per-warehouse attribution is corrected by the locator half.

### Test-infrastructure changes (cucumber)

`Fact_Acct records are matching` had no way to assert a locator, so:

- `FactAcctLineMatcher` / `FactAcctMatchersFactory` gained an optional `M_Locator_ID` column (identifier-resolved via `M_Locator_StepDefData`; the literal `null` asserts "no locator", as the other optional columns already do).
- `load M_Warehouse:` gained an optional `M_Locator_ID` identifier column that registers the warehouse's default locator — symmetric with what `metasfresh contains M_Warehouse:` already did.

Both are additive; no existing feature file changes behaviour.

## Rollout note — already-posted documents need a repost

This changes only the shape of the fact line, not the underlying `M_CostDetail` / `CostAmount`, so a **plain repost is sufficient and correct** — a cost recompute (`product_costs_recreate_from_date`) is *not* needed.

But it is also *required*: MatchInv documents already posted before this ships keep their phantom `P_Asset` quantity and NULL locator, and the inventory-valuation report keeps reading them, until those documents are reposted. Nothing in this PR does that automatically. Whoever rolls this out should plan the repost of the affected MatchInv documents alongside it.

### DIFF
--- a/backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_MatchInv.java
+++ b/backend/de.metas.acct.base/src/main/java-legacy/org/compiere/acct/Doc_MatchInv.java
@@ -16,6 +16,7 @@
  *****************************************************************************/
 package org.compiere.acct;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import de.metas.acct.accounts.BPartnerGroupAccountType;
 import de.metas.acct.accounts.CostElementAccountType;
@@ -63,6 +64,7 @@ import de.metas.util.Services;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.NonNull;
+import lombok.Value;
 import org.adempiere.exceptions.AdempiereException;
 import org.compiere.model.I_C_Invoice;
 import org.compiere.model.I_M_InOut;
@@ -309,36 +311,18 @@ public class Doc_MatchInv extends Doc<DocLine_MatchInv>
 		updateFromInvoiceLine(cr_InventoryClearing);
 
 		//
-		// P_Asset CR/DR
-		// (cost adjustment)
-		if (!costs.getCostAdjustmentAmt().isZero())
+		// P_Asset CR/DR (cost adjustment) and P_COGS CR/DR (already shipped)
+		for (final CostAdjustmentLeg leg : costAdjustmentLegs(costs, getQty(), receiptLine.getM_Locator_ID()))
 		{
-			final Balance costAdjustmentBalance = toDebitBalanceIfPositive(costs.getCostAdjustmentAmt());
 			fact.createLine()
-					.setAccount(docLine.getAccount(ProductAcctType.P_Asset_Acct, as))
+					.setAccount(docLine.getAccount(leg.getAcctType(), as))
 					.setCurrencyConversionCtx(getInvoiceCurrencyConversionCtx())
-					.setAmtSource(costAdjustmentBalance)
-					.setQty(getQty())
+					.setAmtSource(toDebitBalanceIfPositive(leg.getAmt()))
+					.setQty(leg.getQty())
 					.costElement(costElementId)
 					.bpartnerId(getReceiptBPartnerId())
 					.productId(receiptProductId)
-					.buildAndAdd();
-		}
-
-		//
-		// P_COGS CR/DR
-		// (already shipped)
-		if (!costs.getAlreadyShippedAmt().isZero())
-		{
-			final Balance alreadyShippedBalance = toDebitBalanceIfPositive(costs.getAlreadyShippedAmt());
-			fact.createLine()
-					.setAccount(docLine.getAccount(ProductAcctType.P_COGS_Acct, as))
-					.setCurrencyConversionCtx(getInvoiceCurrencyConversionCtx())
-					.setAmtSource(alreadyShippedBalance)
-					.setQty(getQty())
-					.costElement(costElementId)
-					.bpartnerId(getReceiptBPartnerId())
-					.productId(receiptProductId)
+					.locatorId(leg.getLocatorRepoId())
 					.buildAndAdd();
 		}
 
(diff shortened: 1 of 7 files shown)


teosarca:
## Decisions on the two open questions in the description

Both are now settled by the feature owner. Recording them here so the PR carries the decision rather than the question.

### 1. Multi-org `AD_Org_ID` resolution — **accepted as is**

Adding the locator makes `FactLine.computeOrgIdEffective` resolve the leg's org at Prio 1 from the locator (the receiving warehouse's org) instead of Prio 2 from the doc line (the invoice's org). On a multi-org installation where those differ, the `P_Asset` cost-adjustment leg now follows the warehouse.

Accepted deliberately: it is the more correct attribution — the inventory value belongs to the org holding the stock — and it makes the leg consistent with the sibling GR/IR leg, which already sets `.orgId(receiptLine.getAD_Org_ID())` explicitly. **No cross-org test covers it**, and that is accepted with the decision rather than overlooked.

**Correction to the description above:** its summary line says only `Fact_Acct.Qty` and `Fact_Acct.M_Locator_ID` change. That is accurate for a single-org installation; on a multi-org one `AD_Org_ID` on that leg can change too, as described here.

### 2. The `P_COGS` already-shipped leg — **left unchanged, deliberately**

It carries the same shape (`setQty(getQty())`, no locator, no offsetting negative), and the closest precedent — `Doc_PPCostCollector.addCostDifferenceFactLine` — applies `toZero()` + locator uniformly to every leg including COGS. That evidence points toward changing it too.

It is nonetheless left alone for now: COGS is not an inventory account, so it does not drive the reported problem, and the observed volume is small. Reviewed and decided, not missed.

Both decisions are the feature owner's.

PR #25677, opened 2026-08-31
Material Cockpit v2 — jump actions disable with a reason when their target is empty

metas-ts:
## What

Each Material Cockpit v2 jump action now offers itself **disabled, with a stated reason**, when its jump target holds no records — instead of appearing enabled and then silently doing nothing when clicked.

Three actions are covered: the goods-receipt jump (`M_ReceiptSchedule`), the delivery-disposition jump (`M_ShipmentSchedule`), and the production-disposition jump (`PP_Order_Candidate`).

## Why

A planner selecting a cockpit row could not tell "there is genuinely nothing to show here" from "this is broken". An empty jump result collapses to no records to open, so the click produced no visible effect at all.

## How

`checkPreconditionsApplicable` now runs the **same query** its `doIt` runs, as a cheap first-match probe, and returns `ProcessPreconditionsResolution.reject(NO_RELATED_DOCS_FOUND)` when nothing matches. That reaches the client as `disabled` + `disabledReason` and renders greyed-out with the reason beside the caption — **no frontend change**, and **no new `AD_Message`** (the existing `NO_RELATED_DOCS_FOUND` is reused, so there is no migration script).

The three jumps were near-identical, so their triplicated precondition/`doIt` scaffolding collapses into one abstract `QtyDemandQtySupplyJumpProcess` and each subclass reduces to its own query. Within each subclass the probe and the jump share a single `toQuery(row)` builder, so they cannot drift apart and start disagreeing about what "empty" means.

Two repositories gained a first-match `getIdByQuery` beside their existing list method, so the probe costs one row rather than materialising the whole id set. `PPOrderCandidateDAO` built its query inline, so that was extracted into a private `buildQuery` shared by both methods — the shape the other two already had.

### Drive-by fix

`ReceiptScheduleDAO.getIdByQuery` was unusable for its only purpose. `IQuery.firstIdOnlyOptional` is `Optional.ofNullable(idMapper.apply(firstIdOnly()))`, so on "no match" the mapper receives the sentinel `-1` and must answer `null`; this call site passed the strict `ofRepoId`, which throws `Assumption failure: M_ReceiptSchedule_ID > 0 but it was -1`. The exception aborted `checkPreconditionsApplicable` and the framework then dropped the action from the quick-actions list entirely — so it was not even offered.

This change is the first caller ever to ask that method about an empty result, which is why it had gone unnoticed. 10 of the 12 other `firstIdOnlyOptional` call sites already use the `OrNull` variant. (`WarehouseDAO#retrieveWarehouseIdByValue` carries the same strict-mapper shape and looks like a latent twin — deliberately left alone here.)

## Testing

New Playwright spec `e2e/frontend-webui/tests/spec/material-cockpit-v2-jump-preconditions.spec.js`, run against a local stack built from this branch.

The three "disabled when empty" checks are **soft** assertions so each stream reports independently. They were observed failing before any production change, then going green one stream at a time as each jump was migrated — a real negative control rather than an assumed one.

Every case also asserts the **enabled** path on a row whose target does hold records, and that the jump still opens its target view, which is the over-rejection guard for this change: a disabled action hiding records that actually exist would be the regression that matters.

Selectors and assertions are language-invariant — quick actions are located by `data-testid` (`quick-action-<AD_Process.Value>`) and the disabled state by the `quick-actions-item-disabled` class plus the presence of a non-empty reason element, never by localized caption or message text.


backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/material/process/QtyDemandQtySupplyJumpProcess.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Common base for the Material Cockpit v2 "jump to ..." actions on {@code QtyDemand_QtySupply_V}.
+ * <p>
+ * Each subclass's {@link #checkPreconditionsApplicable(IProcessPreconditionsContext)} runs the same query as
+ * {@link #doIt()}, as a cheap first-match probe, and rejects with {@link #MSG_NO_RELATED_DOCS_FOUND} when the probe
+ * finds nothing, instead of silently opening an empty target grid.
+ */
+public abstract class QtyDemandQtySupplyJumpProcess extends JavaProcess implements IProcessPrecondition
+{
+	// NOTE: de.metas.swat.base cannot see de.metas.ui.web.view.SqlViewFactory (dependency runs the other
```
metas-ts: so, why not create the processes within backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/v2/jump/QtyDemandQtySupplyJumpService.java
```diff
@@ -0,0 +1,111 @@
...
+ * single Spring collaborator through which those {@code JavaProcess} subclasses reach the repositories/DAOs of
+ * the target streams (shipment schedules, production order candidates), so no process holds a {@code @Repository}
+ * field directly.
+ * <p>
+ * Each stream exposes one {@code hasXxxToOpen}/{@code findXxxToOpen} pair, both built from the same
+ * query-builder method, so the precondition probe and the actual jump can never disagree about what "empty" means.
+ */
+@Service
+@RequiredArgsConstructor
+public class QtyDemandQtySupplyJumpService
```
metas-ts, line 57: just to be sure: when moving the processes, please also move this class

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/v2/jump/QtyDemand_QtySupply_V_to_PP_Order_Candidate.java
```diff
@@ -23,55 +23,22 @@
...
-import org.compiere.SpringContextHolder;
-import org.eevolution.model.I_PP_Order_Candidate;
-import org.eevolution.productioncandidate.model.dao.PPOrderCandidateDAO;
-import org.eevolution.productioncandidate.model.dao.PPOrderCandidatesQuery;
 
 import java.util.List;
-import java.util.stream.Collectors;
 
-public class QtyDemand_QtySupply_V_to_PP_Order_Candidate extends JavaProcess implements IProcessPrecondition
+public class QtyDemand_QtySupply_V_to_PP_Order_Candidate extends QtyDemandQtySupplyJumpProcess
```
metas-ts, line 31: ....and then also move the preexisting processes..

e2e/frontend-webui/tests/spec/material-cockpit-v2-jump-preconditions.spec.js
```diff
@@ -0,0 +1,418 @@
+import { expect } from '@playwright/test';
+import { test } from '../../playwright.config';
+import { allure } from 'allure-playwright';
+import { Backend } from '../utils/Backend';
+import { LoginPage } from '../utils/pages/LoginPage';
+import { DashboardPage } from '../utils/pages/DashboardPage';
+import { FRONTEND_BASE_URL } from '../utils/common';
+
+/**
+ * Material Cockpit v2 — the three bespoke jump actions must be offered as disabled-with-a-reason
```
metas-ts, line 10: revise the comments and leave only what really gives information beyond the actual code

backend/de.metas.swat/de.metas.swat.base/src/main/java/de/metas/material/process/QtyDemandQtySupplyJumpProcess.java
metas-ts: **Resolution (dev):** Done — moved in https://github.com/metasfresh/metasfresh/commit/d9303977687.

All five classes now live in `de.metas.ui.web.material.cockpit.v2.jump` (in `de.metas.ui.web.base`), via `git mv` so history follows them. I picked that package over `de.metas.ui.web.view.process` because the sibling `…cockpit.v2.reservation` already has exactly this shape — an abstract base plus its concrete processes — while `view/process` holds the *shared* `RelationTypeInOverlayProcess`, which serves three different windows. Happy to put them next to the shared one instead if you'd rather.

No POM change was needed: `de.metas.ui.web.base` already reaches `de.metas.inoutcandidate` and `de.metas.material.cockpit`, and depends on `de.metas.manufacturing` directly.

The move also let me delete a duplication this PR had documented as unavoidable: the base declared its own `MSG_NO_RELATED_DOCS_FOUND` purely because `de.metas.swat.base` cannot see `SqlViewFactory`. From here it simply uses `SqlViewFactory.MSG_NO_RELATED_DOCS_FOUND` — the same constant `RelationTypeInOverlayProcess` uses.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/v2/jump/QtyDemandQtySupplyJumpService.java
metas-ts: **Resolution (dev):** Done — `QtyDemandQtySupplyJumpService` moved with them in https://github.com/metasfresh/metasfresh/commit/d9303977687, same package `de.metas.ui.web.material.cockpit.v2.jump`.

While there I checked whether it should instead reuse `MaterialCockpitV2BasedProcess` from the sibling reservation package, and concluded not: that one extends `ViewBasedProcessTemplate` and resolves its row via `getSingleSelectedRow()` off an `IView`/`MaterialCockpitV2RowVO`, whereas these resolve a `QtyDemandQtySupply` from `getSingleSelectedRecordId()`. Different base class, different row type, different selection model — reusing it would mean bolting a view dependency onto a record-based process. Say the word if you want them aligned anyway.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/material/cockpit/v2/jump/QtyDemand_QtySupply_V_to_PP_Order_Candidate.java
metas-ts: **Resolution (dev):** Done — the three pre-existing processes moved too, in the same commit https://github.com/metasfresh/metasfresh/commit/d9303977687: `QtyDemand_QtySupply_V_to_ReceiptSchedule`, `…_to_ShipmentSchedule`, `…_to_PP_Order_Candidate`. Class names unchanged, package only.

One consequence worth flagging, since it was not visible from the diff: `AD_Process.Classname` stores the fully-qualified name, so moving the package alone would have left 585513/585514/585516 pointing at `de.metas.material.process.*` — the WebUI could not have instantiated them and all three actions would have vanished at runtime. The same commit adds migration `5821500_material_cockpit_v2_jump_processes_repackage.sql` to update them. It keys on the old classname as well as the id, so a re-run cannot clobber a corrected row, and it deliberately leaves 585515 alone — that one runs the shared `RelationTypeInOverlayProcess`.

Verified rather than assumed: migration applied to a local stack, then the Playwright spec run green against the moved classes — which is what proves the new classnames actually resolve.

e2e/frontend-webui/tests/spec/material-cockpit-v2-jump-preconditions.spec.js
metas-ts: **Resolution (dev):** Done — https://github.com/metasfresh/metasfresh/commit/8f727977b10, comment lines 169 → 81, and the diff is comment-only (no executable line changed).

Removed: narration of this change's own history; call-site paraphrases of other classes' internals (`QuickActionsDropdown` internalName resolution, the `Overrides_Window_ID` walk-through with a point-in-time window id, `CreateProductPlanningCommand`/`PPOrderCandidateDemandMatcher`); restatements of the DOM shape and of Playwright mechanics; and the same rationale repeated in two or three places.

Kept only what a reader would otherwise get wrong and 'helpfully' undo — why the disabled assertions are soft, why one row seeds two order lines, why the enabled check runs before the disabled one, why the seeding needs two `createMasterdata` calls, and why the jump is proven by a `documentView` request for a *different* window rather than by a grid that is already on screen.

backend/de.metas.swat/de.metas.swat.base/src/main/sql/postgresql/system/48-de.metas.inoutcandidate/5821530_sys_M_ReceiptSchedule_ProductWarehouse_QtyToMove_index.sql
```diff
@@ -0,0 +1,15 @@
+-- Support an existence probe run once per row selection: filters M_ReceiptSchedule on
```
metas-ts, line 1: add a brief postgresql-comment to every index, to indicate what the index supports. include the feature ("F19011 - Material Cockpit v2") and the processe's `AD_Process.Value`

metas-ts: **Resolution (dev):** Done — https://github.com/metasfresh/metasfresh/commit/bb36d6015e1 adds a `COMMENT ON INDEX` to all three, each naming the feature and the process it serves. For example:

```sql
COMMENT ON INDEX m_receiptschedule_m_product_id_m_warehouse_id IS
    'F19011 - Material Cockpit v2: backs the precondition probe of AD_Process QtyDemand_QtySupply_V_to_ReceiptSchedule, which asks on every cockpit row selection whether any receipt schedule matches the row.';
```

The other two name `QtyDemand_QtySupply_V_to_PP_Order_Candidate` and `QtyDemand_QtySupply_V_to_ShipmentSchedule` respectively.

Verified rather than assumed: the migration runner skipped the three scripts on re-run (already applied, checksum unchanged for that purpose), so I executed the statements directly against the local stack and read the comments back out of `obj_description(c.oid,'pg_class')` — all three are stored. A fresh instance gets them from the scripts as part of the normal apply.

PR #25681, opened 2026-09-01
External barcode support on HU-Unpack

metas-ts:
## Problem

At the mobile picking **unpack** step, the target-HU scan accepted **only** a metasfresh global QR code. Scanning the warehouse's own printed label on the target unit — an `ExternalBarcode` attribute, or the plain `M_HU.Value` — was rejected, even though the pick-*from* side of the same screen has always accepted all of them.

Worse, the rejection was not operator-facing. `GlobalQRCode.ofString().orThrow()` throws via `Check.mkEx`, whose default is a plain `RuntimeException`, so `userFriendlyError` was false and the picker got the generic *"Please try again. If the problem persists, contact support"* fallback — while the real reason (`Invalid global QR code(1): <code>`) stayed in the HTTP response body and an `AD_Issue` row.

## Change

Hoists the "resolve any supported scanned HU label to that unit's `HUQRCode`" step out of `SetPickFromHUWFActivityHandler`'s private helper into `HUQRCodesService.getQRCodeByScannedCode(ScannedCode)`, and routes the unpack target through it instead of `HUQRCode.fromGlobalQRCodeJsonString`.

- **Accepted** at the unpack target now: a metasfresh global QR code, and the plain `M_HU.Value` / `ExternalBarcode` printed on the unit (which `parse` resolves by looking the unit up and returning *its* QR code).
- **Rejected**, deliberately: every other `IHUQRCode` — `PickOnTheFlyQRCode`, `LMQRCode`, `CustomHUQRCode`, `GS1HUQRCode`, `EAN13HUQRCode`. These describe goods or an intent, not one specific unit. They are *not* retried as an `M_HU.Value`/`ExternalBarcode` lookup, because such a code could coincidentally equal some unit's value and picking from the wrong unit is worse than a clear rejection.
- The rejection now carries `scannedCode` + `huQRCodeType` as exception parameters, so a log line says which kind of code was refused.
- Skipping the target scan (unpick to the floor) stays legal — the blank-tolerant `Optional` chain is preserved.

`fromJson` becomes an instance method so it can reach the injected service; its only caller already ran in instance context.

**No behaviour change on the pick-from side** — `SetPickFromHUWFActivityHandler` now delegates to the shared method, which narrows exactly as its deleted private helper did (`parse(String)` is itself just `parse(ScannedCode.ofString(...))`).

**As an intended side effect, an unrecognised scan is now operator-facing.** Routing through the shared bridge means an unresolvable code reaches `parse`'s own terminal `MobileQRCodeMessages.newNotRecognizedException`, an `AdMessageKey`-based translatable exception that `AdempiereException` auto-marks as a user validation error. The picker now sees `QR code not recognized: <the code they scanned>`.

## Tests

New mobile-webui spec `picking/unpack/picking_unpack_legacy_label_target.spec.js`:

- legacy `ExternalBarcode` as the unpack target,
- plain `M_HU.Value` as the unpack target,
- an unrecognised code at the unpack target → operator-facing message, nothing moved, the panel stays usable.

New JUnit coverage on `HUQRCodesService.getQRCodeByScannedCode` for a global QR code, an `ExternalBarcode`, and the reject branch.

Two pre-existing specs (`picking_partial_unpack_TU_floor_spanning_picks`, `..._TU_floor_two_lines`) additionally now set `allowSkippingRejectedReason` themselves. Both do a *partial* pick, which leaves a rejected qty; whether a reason is mandatory is a sticky header-row field they never set, so they inherited it and hung for 120 s against any database where it is `N` — which is the DDL default. Unrelated to the fix, but they sit in the folder this change touches.

## Verification

- `de.metas.handlingunits.base` unfiltered unit suite: 1051 pass. `de.metas.picking.rest-api`: 54 pass.
- Cucumber `byScannedCode_qrcode_hash` and `mobileui/picking/Pick_TUs_from_LU`, run locally: 6 scenarios / 14 tests pass.
- Mobile Playwright: the whole `picking/unpack/` folder 14/14 and the whole `picking/` folder 137/137 locally; `mobile test` 296/0 in CI.
- The three added/changed specs cleared a four-leg flake gate: pattern audit, 30 consecutive local runs each at `--retries=0`, both cross-run pollution orders, and a green CI round — started from the hostile `IsAllowSkippingRejectedReason='N'` default so it tested the fix rather than a friendly config.

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/service/HUQRCodesService.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/service/HUQRCodesService.java
@@ -476,12 +476,14 @@ public class HUQRCodesService
 			return (HUQRCode)huQRCode;
 		}
 
-		// Both values matter when diagnosing this: the raw code the operator scanned, and the concrete
-		// IHUQRCode implementation parse() recognized it as - the latter is what tells you WHY it was
-		// rejected (e.g. a GS1 or pick-on-the-fly code, which identifies no existing unit).
+		// huQRCodeType is the load-bearing parameter: it names WHY the code was rejected (a GS1,
+		// EAN13, LM, custom-format or pick-on-the-fly code identifies no existing unit). The instance
+		// alone would not convey that - every IHUQRCode reachable here renders via getAsString(),
+		// which echoes back the scanned string, so it would just duplicate scannedCode.
 		throw new AdempiereException("Invalid HU QR code")
 				.appendParametersToMessage()
 				.setParameter("scannedCode", scannedCode)
+				.setParameter("huQRCodeType", huQRCode.getClass().getSimpleName())
 				.setParameter("huQRCode", huQRCode);
 	}
 }
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/service/HUQRCodesService.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/service/HUQRCodesService.java
@@ -450,14 +450,6 @@ public class HUQRCodesService
 	}
 
 	/**
-	 * Resolves any supported scanned code (metasfresh global QR code, {@code M_HU.Value}/ExternalBarcode label, ...)
-	 * to the {@link HUQRCode} of the handling unit it identifies.
-	 * <p>
-	 * Unlike {@link HUQRCode#fromGlobalQRCodeJsonString}, this accepts every code format {@link #parse(ScannedCode)}
-	 * understands, not only a metasfresh global QR code.
-	 */
-	@NonNull
-	/**
 	 * Resolves any scanned code that identifies an existing handling unit to that unit's {@link HUQRCode}.
 	 * <p>
 	 * {@link #parse(ScannedCode)} already covers every supported label kind — a metasfresh global QR code,
@@ -467,6 +459,7 @@ public class HUQRCodesService
 	 * resolved by a second lookup: silently mapping such a code onto whichever HU happens to share its value
 	 * would pick from the wrong unit.
 	 */
+	@NonNull
 	public HUQRCode getQRCodeByScannedCode(@NonNull final ScannedCode scannedCode)
 	{
 		final IHUQRCode huQRCode = parse(scannedCode);
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/service/HUQRCodesService.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/service/HUQRCodesService.java
@@ -448,4 +448,33 @@ public class HUQRCodesService
 		}
 		throw MobileQRCodeMessages.newNotRecognizedException(scannedCode);
 	}
+
+	/**
+	 * Resolves any supported scanned code (metasfresh global QR code, {@code M_HU.Value}/ExternalBarcode label, ...)
+	 * to the {@link HUQRCode} of the handling unit it identifies.
+	 * <p>
+	 * Unlike {@link HUQRCode#fromGlobalQRCodeJsonString}, this accepts every code format {@link #parse(ScannedCode)}
+	 * understands, not only a metasfresh global QR code.
+	 */
+	@NonNull
+	/**
+	 * Resolves any scanned code that identifies an existing handling unit to that unit's {@link HUQRCode}.
+	 * <p>
+	 * {@link #parse(ScannedCode)} already covers every supported label kind — a metasfresh global QR code,
+	 * a configured scannable code format, and a plain {@code M_HU.Value} / {@code ExternalBarcode} attribute —
+	 * so this method only narrows its result. A code that parses to something other than an assigned HU QR code
+	 * (a pick-on-the-fly, LM, GS1 or EAN13 code) does NOT identify an existing HU and is rejected rather than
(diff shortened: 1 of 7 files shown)


backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/service/HUQRCodesService.java
```diff
@@ -448,4 +448,26 @@ else if (LMQRCode.isHandled(globalQRCode))
 		}
 		throw MobileQRCodeMessages.newNotRecognizedException(scannedCode);
 	}
+
+	/**
+	 * Resolves any scanned code that identifies an existing handling unit to that unit's {@link HUQRCode}.
```
metas-ts: i don't understand that statement

metas-ts: **Resolution (dev):** Fair — that sentence was circular ("resolves any scanned code that identifies an existing handling unit to that unit's HUQRCode" defines the thing by itself), so I rewrote the javadoc rather than defend it: https://github.com/metasfresh/metasfresh/commit/8241104b6183061a250285174fc6b723cfce7555

It now states plainly:

- **what it returns** — the `HUQRCode` of the handling unit that the scanned code refers to;
- **what it accepts** — every label kind that can appear on a unit (a metasfresh global QR code, a configured scannable code format, or the plain `M_HU.Value` / `ExternalBarcode` printed on it). `parse(ScannedCode)` already resolves all of them, so this method only narrows that result;
- **what it rejects, and why** — codes `parse` recognizes but that do not refer to an existing unit (pick-on-the-fly, LM, GS1, EAN13). Those are deliberately *not* retried as an `M_HU.Value` / `ExternalBarcode` lookup, because such a code could coincidentally equal some unit's value, and picking from the wrong unit would be worse than a clear rejection.

Comment-only change, no behaviour change. The module still compiles and its unfiltered unit suite is green (1051 tests, 0 failures).

If any of the three parts is still unclear, say which and I will cut it further.

metas-ts: **Resolution (dev):** Correcting my previous reply — it was wrong on one point, and so was the javadoc it described.

I listed "a configured scannable code format" among the kinds this method **accepts**. It is actually **rejected**. `parse(ScannedCode)` wraps such a code as `CustomHUQRCode` and returns it *before* reaching the `M_HU.Value` / `ExternalBarcode` branch, and this method's only success condition is `instanceof HUQRCode` — so it fails with `Invalid HU QR code`, exactly like GS1/EAN13/LM/pick-on-the-fly. `CustomHUQRCode` was also missing from the enumerated rejected list.

Fixed in https://github.com/metasfresh/metasfresh/commit/4f0c4c523cc — the javadoc now states the real partition:

- **accepted** — the two kinds that identify one specific unit: a metasfresh global QR code, and the plain `M_HU.Value` / `ExternalBarcode` printed on the unit (which `parse` resolves by looking the unit up and returning *its* QR code);
- **rejected** — all five other `IHUQRCode` implementors: `PickOnTheFlyQRCode`, `LMQRCode`, `CustomHUQRCode`, `GS1HUQRCode`, `EAN13HUQRCode`.

It also now records where a product-scoped code *can* be resolved to a unit — `PickFromHUQRCodeResolveCommand` does exactly that on the pick-from side — and why it isn't available here: that path needs a `ProductId`, which a caller identifying a bare target unit doesn't have.

Both commits are comment-only, no behaviour change. Behaviour is unchanged from the `parseHUQRCode` this replaced, which narrowed the same way.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/qrcodes/service/HUQRCodesService.java
```diff
@@ -448,4 +448,34 @@ else if (LMQRCode.isHandled(globalQRCode))
...
+	@NonNull
+	public HUQRCode getQRCodeByScannedCode(@NonNull final ScannedCode scannedCode)
+	{
+		final IHUQRCode huQRCode = parse(scannedCode);
+		if (huQRCode instanceof HUQRCode)
+		{
+			return (HUQRCode)huQRCode;
+		}
+
+		throw new AdempiereException("Invalid HU QR code: " + scannedCode);
```
metas-ts: include the `huQRCode` instance it was parsed into (=> exception-parameter)

metas-ts: **Resolution (dev):** Done — https://github.com/metasfresh/metasfresh/commit/5646970fd5fb9c7d57cfcdea2024fea2873e3b9a

The parsed instance is now attached as an exception parameter, using the module's usual idiom (as in `IHUAssignmentDAO` / `PickingJobCreateRepoHelper`):

```java
throw new AdempiereException("Invalid HU QR code")
        .appendParametersToMessage()
        .setParameter("scannedCode", scannedCode)
        .setParameter("huQRCode", huQRCode);
```

I kept the raw scanned code too, as a parameter rather than concatenated into the message — the two answer different questions: `scannedCode` is what the operator actually scanned, `huQRCode` is the concrete `IHUQRCode` implementation `parse()` recognized it as, which is the part that explains *why* it was rejected (a GS1, EAN13, LM, custom-format or pick-on-the-fly code identifies no existing unit).

Verified: module compiles, 16 unit tests pass. The reject-branch test asserts on the `"Invalid HU QR code"` prefix, which is unchanged.

PR #25690, opened 2026-09-01
MobileUI Picking: render a configured caption item as its own block of lines

teosarca:
## What

Adds a per-caption-item switch to the mobile picking profile so a chosen caption row renders as its own **block of lines** in the picking job list, instead of being joined inline with `" | "`.

Motivation: when an order contains several articles, pickers need to see all article names at a glance in the job list, one per line, without opening the job.

## How

New boolean column `PickingProfile_PickingJobConfig.IsBlockLayout` (CHAR(1), NOT NULL, default `'N'`), with its own `AD_Element` (DE "Blockdarstellung" / EN "Block layout") and a field on window `AD_Window_ID=541743` tab "Feld", placed beside `IsDisplayInSummary` / `IsDisplayInDetailed`.

All layout logic lives in exactly one place — `WorkflowLauncherCaption.computeTranslatableString()`:

- The caption builder takes an optional `ImmutableSet<String> blockLayoutFields` naming which caption rows are blocks. It defaults to empty, so every other mobile application's launcher caption (distribution, HU consolidation, manufacturing, inventory) is unchanged **by construction**, not by a conditional.
- The join folds the non-blank rows left-to-right, emitting a line break instead of `" | "` on either side of a block row — so no `" | "` ever lands next to a break, and there is no leading or trailing empty line.

On the picking side the product-name value stops deciding its own separator: the join moved from `toContext()`-time to `getDisplayValue(field, context)`-time, so the separator can vary per configured row. `Context` carries the product names un-joined and the `PRODUCT_NAMES` case picks `"\n"` or `", "` from the row's flag. The **job-detail** header path keeps `", "` regardless of the flag.

Deliberate design notes:

- **No ">1 value" condition** — a flagged row renders as a block even for a single product. Consistency was chosen over screen density, and it is reversible by configuration.
- The column is named `IsBlockLayout` rather than `IsMultiLine` because an `AD_Element` is keyed by column name and `IsMultiLine` already exists (on `AD_UI_Element`, meaning a multi-line text input) — reuse would surface the wrong German label.

Frontend: the line break already rendered (`launchers.scss`, `white-space: break-spaces`); the only gap was alignment. A multi-line caption now gets a `caption-multiline` class and is left-aligned, scoped under `.launchers-container` so the shared `.caption-btn` rules used by `Button.jsx` / `ApplicationButton.jsx` are untouched.

## Tests

Three non-overlapping layers:

- `WorkflowLauncherCaptionBlockLayoutTest` — block row in the middle / first / last, next to a blank-valued row, a caption with no flag set at all (byte-identical to previous output), and the caption tie-break order under block layout (`\n` 0x0A sorts before both `' '` and `'|'`, so entries tying on every configured order-by can re-order).
- `DisplayValueProviderBlockLayoutTest` — flag on vs off, one product vs several, list path vs detail path.
- `ButtonWithIndicator.multilineCaption.test.js` — the multiline class is applied from the caption text, and only then.

The existing `DisplayValueProviderProductNamesTest` / `DisplayValueProviderProductNamesDetailTest` pass with **no expectation edited** — they pin the flag-off behaviour.

An end-to-end Playwright spec on the job list follows in this PR.


backend/de.metas.workflow.rest-api/src/main/java/de/metas/workflow/rest_api/model/WorkflowLauncherCaption.java
```diff
@@ -83,13 +89,33 @@ public ITranslatableString toTranslatableString()
...
 				.stream()
-				.map(fieldValues::get)
-				.filter(caption -> !TranslatableStrings.isBlank(caption))
+				.filter(field -> !TranslatableStrings.isBlank(fieldValues.get(field)))
 				.collect(ImmutableList.toImmutableList());
 
-		return TranslatableStrings.join(" | ", parts);
+		if (nonBlankFields.isEmpty())
+		{
+			return TranslatableStrings.join(" | ", ImmutableList.of());
```
teosarca: why not simply returning TranslatableStrings.emty()?

backend/de.metas.workflow.rest-api/src/main/java/de/metas/workflow/rest_api/model/WorkflowLauncherCaption.java
```diff
@@ -83,13 +89,33 @@ public ITranslatableString toTranslatableString()
...
+		}
+
+		ITranslatableString result = fieldValues.get(nonBlankFields.get(0));
```
teosarca: consider using TranslatableStringBuilder, see de.metas.i18n.TranslatableStrings#builder

backend/de.metas.workflow.rest-api/src/main/java/de/metas/workflow/rest_api/model/WorkflowLauncherCaption.java
teosarca: **Resolution (dev):** Agreed — adopting `TranslatableStrings.empty()`.

Verified they are the same object, not merely equivalent: `join(...)` delegates to `joinList(...)` (`TranslatableStrings.java:130-134`), which returns `empty()` — i.e. the `ConstantTranslatableString.EMPTY` singleton — as soon as the list is empty. So `join(" | ", ImmutableList.of())` just reaches `empty()` by a longer route, and the `" | "` in it is dead: it reads as if the separator mattered in the empty case, which it never does.

The original shape was chosen only to keep the empty branch textually identical to the pre-change code; that was over-cautious given the two are the same singleton.

Commit follows shortly — an E2E task is currently committing on this branch and I am keeping a single writer on it. I will post the commit URL here once it lands.

backend/de.metas.workflow.rest-api/src/main/java/de/metas/workflow/rest_api/model/WorkflowLauncherCaption.java
teosarca: **Resolution (dev):** Taking it — with one thing to confirm first, which decides the final shape.

What this fold needs is a separator that *varies per pair*: `"\n"` when either neighbour is a block-layout row, `" | "` otherwise. So the question is whether `TranslatableStringBuilder` can express that (e.g. by appending the separator itself as a constant between values) rather than only appending values under one fixed separator.

If it can, I adopt it — it removes the pairwise nested-composite fold and is the house-idiomatic assembly. If it turns out it cannot express a per-pair separator without changing the produced string, I will say so here with the API evidence instead of forcing it into that shape.

Either way the observable behaviour stays fixed: no `" | "` adjacent to a line break, no leading or trailing empty line, and an unflagged caption byte-identical to today (pinned by `noFlagSetAtAll_joinedWithPipe_byteIdenticalToTodaysOutput`).

Commit and the verdict follow shortly; I will post the commit URL here.

backend/de.metas.workflow.rest-api/src/main/java/de/metas/workflow/rest_api/model/WorkflowLauncherCaption.java
teosarca: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/0e420c6584ed5b3f9caf7fb9ca0673b8c627a217 — the empty case now returns `TranslatableStrings.empty()`.

Confirmed identical, not just equivalent: `empty()` returns the `ConstantTranslatableString.EMPTY` singleton (`TranslatableStrings.java:181-184`), which is exactly what `joinList` returned for an empty list (`:130-144`). Same object, so no caller can observe a difference.

backend/de.metas.workflow.rest-api/src/main/java/de/metas/workflow/rest_api/model/WorkflowLauncherCaption.java
teosarca: **Resolution (dev):** Adopted, in https://github.com/metasfresh/metasfresh/commit/0e420c6584ed5b3f9caf7fb9ca0673b8c627a217.

`TranslatableStrings.builder()` does express the per-pair separator: the assembly is now one flat pass — `append(firstValue)`, then per following field `append(separator).append(value)` — with the separator still chosen as `"\n"` when either neighbour of the pair is a block-layout row, else `" | "`. That replaces the right-nested pairwise `CompositeTranslatableString` fold with a single flat composite.

Verified behaviour-preserving at source level rather than by assertion alone:

- **Same string in every language.** The old nested chain and the new flat `[value, sep, value, sep, …]` list joined with `""` reduce termwise to the identical concatenation, because `CompositeTranslatableString.translate()` maps `translate(lang)` over each leaf before joining.
- **The `"\n"` separator is not silently dropped.** `TranslatableStringBuilder.append(String)` guards with `Check.isEmpty(value)`, which resolves to `String.isEmpty()` and *not* a trim/blank check — so a whitespace-only separator survives. This was the one plausible regression vector here, so it was checked explicitly.
- **Single non-blank field stays unwrapped**: `build()` returns `parts.get(0)` directly when there is only one part, matching the old early return.

Tests: `WorkflowLauncherCaptionBlockLayoutTest` 6/6, `de.metas.workflow.rest-api` 19/19, `de.metas.picking.rest-api` 59/59 — including the byte-identical no-flag regression guard that protects the distribution, HU-consolidation, manufacturing and inventory launcher captions.

PR #25697, opened 2026-09-01
MobileUI job start — a pushed launchers snapshot can no longer delete a workflow process

teosarca:
## What breaks today

Starting a mobile picking job intermittently deletes the just-started job from the browser's own redux store, about 9 ms after the job screen has painted. The router then finds no such process, redirects home, and the operator is bounced to the applications menu in the middle of starting the job. Nothing is wrong on the server: the job exists, is started, and is assigned — only the client's copy is gone.

Observed rate on the E2E witness spec: 1 failure in 30 executions (3.3%), independently reproduced at 1 in 31 (3.2%) in an instrumented capture. Pooled 2 of 61 = 3.3%. An exact-binomial 95% CI for 1/30 is roughly 0.1%–17%, so treat 3.3% as an order of magnitude, not a precise rate.

## Root cause — captured, not inferred

`containers/wfLaunchersScreen/useLaunchers.js` dispatched a launchers snapshot arriving over the **websocket push** as `POPULATE_LAUNCHERS_COMPLETE`, stamping it with `requestTimestamp: Date.now()` — the browser's **receipt** time. The `wfProcesses` garbage collector (`reducers/wfProcesses/workflow.js`) reads that stamp as "this snapshot knows about everything older than itself" and removes any process the snapshot does not list. A snapshot the server had computed **before** the job existed therefore looks newer than the job, and the collector deletes it.

The instrumented capture of the natural failure, all timestamps epoch ms:

| event | timestamp |
|---|---|
| server computed the launchers snapshot (`computedTime` `2026-09-01T08:32:55.868Z`) | 1788251575868 |
| `POST /userWorkflows/wfProcess/start` issued | 1788251576019.9 |
| start returned 200, job `picking-1000013` | 1788251576066.7 |
| job enters the store (`updatedAt`) | 1788251576066 |
| **websocket frame received -> job DELETED** | **1788251576075.6** |
| app pushes `/mobile/picking/wf/picking-1000013` -> `/mobile/` | ~1788251576077 |

The snapshot was computed 151 ms before the start request was even issued and 198 ms before the job entered the store, spent ~207 ms in flight, and arrived 9 ms after it. The guard arithmetic with the receipt stamp:

```
updatedAt 1788251576066 >= launchersFetchStartedAt 1788251576075  ->  FALSE  ->  delete
```

The prune record names the route the operator was on when it fired: `removed:["picking-1000013"] … keysAfter:[] … route:"/mobile/picking/wf/picking-1000013"`.

Two further facts from the same capture that shaped the fix:

- **The websocket site is the sole exposure and does no useful work.** 92 prunes across 31 executions: **91 came from the requested (REST) snapshot**, every one a legitimate collection of a job the operator had already left (margin +30..+61 ms, guard held on all 91); **1 came from the websocket push** (margin +9 ms) and killed the live job. The push path produced *zero* legitimate collections.
- **The margin is narrow, not comfortable.** Over the 20 executions that had a push inside the window, 19 margins were on the safe side and 1 was positive (+8.3 ms, the failure) — but the mode sits only ~15–20 ms from zero. `WebSocketProducersRegistry` schedules the launchers producer at `initialDelayMillis = 1000, periodMillis = 1000` and nulls the last-result holder on subscribe, so the first poll after a subscribe always pushes: the staleness that matters is the 1000 ms poll phase plus ~200 ms compute-and-transport lag. This is **not** specific to the second job start in the test — an operator who dwells longer than ~1 s on the launcher list is exposed on the first start too.

## The fix — three parts

**1. A pushed snapshot gets its own action type, so it structurally cannot delete.** New `POPULATE_LAUNCHERS_PUSHED` (`constants/LaunchersActionTypes.js`) + `populateLaunchersPushed` (`actions/LauncherActions.js`), deliberately carrying **no** `requestTimestamp`. `reducers/wfProcesses/workflow.js` does not handle the new type at all, so a push cannot prune; `reducers/launchers.js` handles it exactly like the requested type, so the visible job list keeps refreshing at unchanged cadence. The explicitly **requested** snapshot stamps at request-issue time and therefore can never over-state its own freshness — it keeps doing all garbage collection (91 of 91 legitimate collections in the capture came from it).

Attaching the delete power to the **action type** rather than to a comment is the point: a future second consumer of launchers data cannot silently acquire delete power, it has to pick the deleting type on purpose. The two premises the remaining delete path rests on are stated in the code next to the guard: requested snapshots are never served from cache (`WorkflowRestController` leaves `maxStaleAccepted` unset, default `Duration.ZERO` forces a recompute), and the requested fetch is dispatched only from the launchers screen the operator has left (`useLaunchers.js` is the only non-test dispatcher).

**2. A torn-down subscription no longer writes to the store** (`api/launchers.js`). `ws.disconnectClient` is asynchronous, so the broker can still deliver a frame after the effect's cleanup ran and the screen unmounted — in the captured failure the STOMP `DISCONNECT` was logged **before** the deleting `MESSAGE`. A per-effect `subscriptionActive` flag is cleared first thing in the cleanup and the inner frame handler returns immediately once it is unset. This is **co-primary with part 1, not hygiene**: on its own it would also have prevented the captured failure. The two defects are co-present at n=1, which cannot rank them; each closes a path the other leaves open — a stamping fix does not stop an unmounted screen from writing, and a teardown fix does not stop a mounted screen from stamping wrongly.

**3. The E2E job-start helpers recover when the app lands on the applications menu** (`e2e/mobile-webui/tests/utils/screens/`). `tapLauncherUntilJobScreen` only re-tapped while the launcher list was still attached; on a bounce neither the launcher list nor the job screen is attached, so the loop broke out and burned its remaining attempts on the trailing 20 s screen wait — the exact shape of the observed failure. The decision now lives in `tests/utils/screens/jobStartRecovery.js` and distinguishes three outcomes: back on the launcher list, recovered by re-entering the application from the menu, or unknown — in which case the caller must **not** re-tap, because a start may still be in flight and a blind re-tap could double-start the workflow. Wired into all three tap sites (`tapLauncherUntilJobScreen`, `PickingJobsListScreen.startJob`'s `documentNo` branch — which tapped once with no retry at all — and `DistributionJobsListScreen.startJob`), sharing one bounded budget. **No timeout was lengthened**; raising the 20 s `#WFProcessScreen` gate would hide a real user-facing bounce.

## Why not compare the server's `computedTime` (rejected on purpose)

The payload already carries the server instant, and the counterfactual works on the captured data:

```
updatedAt 1788251576066 >= computedTime 1788251575868  ->  TRUE   ->  KEEP
```

It was still rejected, for two reasons, so please do not propose it in review:

- **Two incompatible wire formats.** The same DTO serialises `computedTime` as an ISO-8601 string over the websocket and as an epoch-seconds float over REST.
- **Cross-clock comparison.** It is a *server* instant being compared against a *browser* `Date.now()` on a handheld, i.e. correctness would depend on clock agreement between two machines.

Removing the push's ability to prune needs neither.

## Deliberately unchanged

- No backend change; no change to what the launchers list contains, nor to the push cadence.
- Garbage collection frequency is unchanged: the requested fetch fires on every mount of the job-list screen, and the operator must pass through that screen to start any job.
- The 20 s `#WFProcessScreen` arrival gate in the E2E helpers is not lengthened.
- Residual, accepted and already documented in `reducers/wfProcesses/workflow.js`: the surviving requested-snapshot path still compares two wall-clock `Date.now()` values, so a backward clock adjustment between the two stamps could still invert that comparison.

## Automated tests added by this PR

**The load-bearing evidence is the deterministic tests below, not a repeat count.** At p ≈ 3.3% per execution, a 30× green run of the E2E witness carries only ~62% power to have caught the bug, and 10 executions only ~28% — so a green repeat run is necessary but nowhere near sufficient, and the invariant is pinned deterministically instead.

Jest (mobile-webui frontend), each written RED first and shown to fail before the production change:

| test | pins |
|---|---|
| `__tests__/reducers/wfProcesses/pushedLaunchersSnapshotNeverDeletes.test.js` | the invariant: a push never removes a process, parametrised over the whole timestamp matrix (frame long after the job, the captured +9 ms margin, same millisecond, before the job, no `computedTime` at all) plus an empty launchers list, asserting the `wfProcesses` state is deep-equal to a copy taken before the push. A 7th case pins that the push still refreshes the visible list |
| `__tests__/reducers/wfProcesses/websocketLaunchersReceiptStampDeletesStartedProcess.test.js` | the captured interleaving replayed at reducer level with the captured epoch values (−198 ms / +9 ms), with a passing control on the requested route |
| `__tests__/containers/wfLaunchersScreen/websocketLaunchersReceiptStamp.test.js` | the same failure at hook level, driving the real `onWebsocketMessage` callback `useLaunchers` passes to `useLaunchersWebsocket`, so the dispatched action is produced by production code and not mirrored by the test |
| `__tests__/containers/wfLaunchersScreen/launchersFrameAfterTeardownIgnored.test.js` | a frame delivered after unmount never reaches the outer callback (fails with 1 call before the fix); its control — the identical frame while still mounted — passes, so the failure is about teardown, not wiring |
| `__tests__/containers/applicationRoot/ApplicationLayout.deadDeepLink.test.js` | the preserved behaviour: a job route for a process that never arrives still fires exactly one `goHome()` |
| `e2e/mobile-webui/tests/harness/jobStartRecovery.test.js` | the recovery decision for all four outcomes, browserless (a scripted fake page), because after the production fix the real E2E runs no longer reach this branch |

`ApplicationLayout.deadDeepLink.test.js` is **new and self-contained on purpose**: the pre-existing guard for the same behaviour, `__tests__/containers/applicationRoot/ApplicationLayout.bounceHome.test.js`, cannot run on this base branch at all — it mocks `apps/picking/ShelfLifeConfirmDialogHost`, which exists only on `new_dawn_uat` and `intensive_care_hotfix`, so it reports `Tests: 0 total` and has been dead here since it landed. The replacement was mutation-tested: it kills the pre-fix synchronous redirect, the removal of the deferred redirect, and the removal of the null-render guard.

**Note for reviewers:** mobile-webui jest is not executed by the pipeline at all — the step is commented out in `docker-builds/Dockerfile.mobile` — so the jest suites above run only locally. The `mobile test` CI job is the Playwright suite.

## Reviewer reading order

1. `constants/LaunchersActionTypes.js` and `actions/LauncherActions.js` — the new push-only action type and why it carries no timestamp.
2. `reducers/wfProcesses/workflow.js` — the comment block above the surviving delete case states the two premises the fix rests on.
3. `containers/wfLaunchersScreen/useLaunchers.js` and `reducers/launchers.js` — the push is re-routed; the visible list is unaffected.
4. `api/launchers.js` — the teardown flag.
5. `e2e/mobile-webui/tests/utils/screens/jobStartRecovery.js` — note the "unknown" outcome must not authorise a re-tap.

Commits are ordered RED-then-GREEN in pairs, so each production change can be read against the test that failed without it.

## Propagation

The defect is in shared core code (`misc/services/mobile-webui/`), so every branch family carrying these files is exposed. All touched files exist on `soft_panda_hotfix`, `soft_panda_release`, `new_dawn_uat`, `deep_tundra_release` and `intensive_care_hotfix`; the GRAI witness spec is absent on the last two, but the distribution twin `trolley.spec.js` exists everywhere. Merging up from the base branch of this PR is the intended route; each step to be proven by ancestry rather than assumed.

### DIFF
--- a/e2e/mobile-webui/tests/harness/jobStartRecovery.test.js
+++ b/e2e/mobile-webui/tests/harness/jobStartRecovery.test.js
@@ -1,0 +1,90 @@
+import { test } from '../../playwright.config';
+import { expect } from '@playwright/test';
+import { setCurrentPage } from '../utils/common';
+import { recoverToLauncherList } from '../utils/screens/jobStartRecovery';
+
+/**
+ * AC6 — the recovery DECISION, driven against a scripted fake page, with no browser.
+ *
+ * Why this harness test exists: once the production fix lands, the app no longer lands on the
+ * applications menu during a job start, so the repeated real E2E runs never execute the recovery
+ * branch at all. `node --check` only proves the file parses. This is the only test that actually RUNS
+ * the recovery logic across all of its outcomes.
+ *
+ * How it can be browserless: tests/utils/common.js exports setCurrentPage(currentPage), which assigns
+ * the module-level `page` every screen helper reads. A test that does NOT destructure the `page`
+ * fixture never launches a browser, so it can install a fake and drive the helper against it.
+ *
+ * NEVER declare `{ page }` in a test signature in this file — that pulls in the real browser fixture
+ * (playwright.config.js's `page` override calls setCurrentPage(realPage)) and overwrites the fake.
+ *
+ * Scope note: this targets `recoverToLauncherList` rather than `tapLauncherUntilJobScreen`, because
+ * that one cannot be driven with a fake page at all — `resolveLauncherTapTarget` calls
+ * `expect(locator).toHaveCount(1)`, and Playwright's `expect` only accepts a real Locator. That is
+ * exactly why the recovery decision lives in its own module with no `expect` in it.
+ */
+
+// Which selectors are currently attached to the fake DOM.
+let attached = {};
+// Selector -> the attachment changes a tap on it causes (the navigation it triggers).
+let afterTap = {};
+// Every selector tapped, in order.
+let tapped = [];
+
+const fakePage = {
+    locator: (selector) => ({
+        waitFor: async () => {
+            if (!attached[selector]) {
+                throw new Error(`fakePage: ${selector} is not attached`);
+            }
+        },
+        tap: async () => {
+            tapped.push(selector);
+            Object.assign(attached, afterTap[selector] ?? {});
+        },
+    }),
+};
+
+test.beforeEach(() => {
+    attached = {};
+    afterTap = {};
+    tapped = [];
+    setCurrentPage(fakePage);
+});
+
+test('recoverToLauncherList: already back on the launcher list -> launcherList, no tap', async () => {
+    attached['#WFLaunchersScreen'] = true;
+
+    const outcome = await recoverToLauncherList({ applicationId: 'picking' });
+
+    expect(outcome).toBe('launcherList');
+    expect(tapped).toEqual([]);
+});
+
+test('recoverToLauncherList: landed on the applications menu and re-entry succeeds -> recovered', async () => {
+    attached['#ApplicationsListScreen'] = true;
+    afterTap['#picking-button'] = { '#WFLaunchersScreen': true };
+
+    const outcome = await recoverToLauncherList({ applicationId: 'picking' });
+
+    expect(outcome).toBe('recovered');
+    expect(tapped).toEqual(['#picking-button']);
+});
+
+test('recoverToLauncherList: neither screen attached (a navigation still in flight) -> unknown, no tap', async () => {
+    const outcome = await recoverToLauncherList({ applicationId: 'picking' });
+
+    expect(outcome).toBe('unknown');
+    expect(tapped).toEqual([]);
+});
+
+// This case matters most: a broken re-entry must never authorise a re-tap of the launcher, which
+// could double-start the workflow.
+test('recoverToLauncherList: menu attached but re-entry does not bring up the list -> unknown', async () => {
+    attached['#ApplicationsListScreen'] = true;
+
+    const outcome = await recoverToLauncherList({ applicationId: 'picking' });
+
+    expect(outcome).toBe('unknown');
+    expect(tapped).toEqual(['#picking-button']);
+});
--- a/e2e/mobile-webui/tests/harness/jobStartRecovery.test.js
+++ b/e2e/mobile-webui/tests/harness/jobStartRecovery.test.js
@@ -4,12 +4,10 @@ import { setCurrentPage } from '../utils/common';
 import { recoverToLauncherList } from '../utils/screens/jobStartRecovery';
 
 /**
- * AC6 — the recovery DECISION, driven against a scripted fake page, with no browser.
+ * The recovery DECISION, driven against a scripted fake page, with no browser.
  *
- * Why this harness test exists: once the production fix lands, the app no longer lands on the
- * applications menu during a job start, so the repeated real E2E runs never execute the recovery
- * branch at all. `node --check` only proves the file parses. This is the only test that actually RUNS
- * the recovery logic across all of its outcomes.
+ * A job start does not normally land on the applications menu, so a real E2E run never exercises the
+ * recovery branch. This is the only test that runs the recovery decision across all of its outcomes.
  *
  * How it can be browserless: tests/utils/common.js exports setCurrentPage(currentPage), which assigns
  * the module-level `page` every screen helper reads. A test that does NOT destructure the `page`
--- a/misc/services/mobile-webui/mobile-webui-frontend/src/constants/LaunchersActionTypes.js
+++ b/misc/services/mobile-webui/mobile-webui-frontend/src/constants/LaunchersActionTypes.js
@@ -11,7 +11,7 @@ export const POPULATE_LAUNCHERS_COMPLETE = 'launchers/loadComplete';
  * @constant
  * @type {string}
  */
-export const POPULATE_LAUNCHERS_PUSHED = 'launchers/pushedByServer';
+export const POPULATE_LAUNCHERS_PUSHED_BY_SERVER = 'launchers/pushedByServer';
 
 export const CLEAR_LAUNCHERS = 'launchers/clear';
 export const SET_ACTIVE_FILTERS = 'launchers/activeFilters/set';
--- a/misc/services/mobile-webui/mobile-webui-frontend/src/constants/LaunchersActionTypes.js
+++ b/misc/services/mobile-webui/mobile-webui-frontend/src/constants/LaunchersActionTypes.js
@@ -6,13 +6,12 @@ export const POPULATE_LAUNCHERS_START = 'launchers/loadStart';
 export const POPULATE_LAUNCHERS_COMPLETE = 'launchers/loadComplete';
 
 /**
- * A launchers snapshot PUSHED over the websocket. Deliberately a DIFFERENT type from
- * POPULATE_LAUNCHERS_COMPLETE: a push carries no request-issued time, so it can never prove a workflow
- * process is gone, and reducers/wfProcesses/workflow.js must not handle it. See the comment there.
+ * A launchers snapshot the server pushed over the websocket, as opposed to one this client requested.
+ * A separate type from POPULATE_LAUNCHERS_COMPLETE so that no reducer can prune on it.
  * @constant
  * @type {string}
  */
-export const POPULATE_LAUNCHERS_PUSHED = 'launchers/loadPushed';
+export const POPULATE_LAUNCHERS_PUSHED = 'launchers/pushedByServer';
 
 export const CLEAR_LAUNCHERS = 'launchers/clear';
 export const SET_ACTIVE_FILTERS = 'launchers/activeFilters/set';
--- a/misc/services/mobile-webui/mobile-webui-frontend/src/constants/LaunchersActionTypes.js
+++ b/misc/services/mobile-webui/mobile-webui-frontend/src/constants/LaunchersActionTypes.js
@@ -4,6 +4,16 @@
  */
 export const POPULATE_LAUNCHERS_START = 'launchers/loadStart';
 export const POPULATE_LAUNCHERS_COMPLETE = 'launchers/loadComplete';
+
+/**
+ * A launchers snapshot PUSHED over the websocket. Deliberately a DIFFERENT type from
+ * POPULATE_LAUNCHERS_COMPLETE: a push carries no request-issued time, so it can never prove a workflow
+ * process is gone, and reducers/wfProcesses/workflow.js must not handle it. See the comment there.
+ * @constant
+ * @type {string}
+ */
+export const POPULATE_LAUNCHERS_PUSHED = 'launchers/loadPushed';
+
 export const CLEAR_LAUNCHERS = 'launchers/clear';
 export const SET_ACTIVE_FILTERS = 'launchers/activeFilters/set';
 export const CLEAR_ACTIVE_FILTERS = 'launchers/activeFilters/clear';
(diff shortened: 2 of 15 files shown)


misc/services/mobile-webui/mobile-webui-frontend/src/constants/LaunchersActionTypes.js
```diff
@@ -4,6 +4,16 @@
...
 export const POPULATE_LAUNCHERS_COMPLETE = 'launchers/loadComplete';
+
+/**
+ * A launchers snapshot PUSHED over the websocket. Deliberately a DIFFERENT type from
+ * POPULATE_LAUNCHERS_COMPLETE: a push carries no request-issued time, so it can never prove a workflow
+ * process is gone, and reducers/wfProcesses/workflow.js must not handle it. See the comment there.
+ * @constant
+ * @type {string}
+ */
+export const POPULATE_LAUNCHERS_PUSHED = 'launchers/loadPushed';
```
teosarca: * why is better to have a new action then reusing POPULATE_LAUNCHERS_COMPLETE  and providing some flag/indicator?
* isn't it very weird/misleading reading loadComplete and loadPushed?! feels like AI slop/chaos in progress...

misc/services/mobile-webui/mobile-webui-frontend/src/reducers/wfProcesses/workflow.js
```diff
@@ -45,6 +45,20 @@ export const workflowReducer = ({ draftState, action }) => {
       return draftState;
     }
 
+    // Only a REQUESTED launchers snapshot may prune. Pushed snapshots have their own action type
```
teosarca: wow, a half-page of AI slop. be CSP

misc/services/mobile-webui/mobile-webui-frontend/src/containers/wfLaunchersScreen/useLaunchers.js
```diff
@@ -54,9 +54,12 @@ export const useLaunchers = ({ applicationId, showFilterByQRCode, facets, filter
     filters,
     facets,
     onWebsocketMessage: ({ applicationId, applicationLaunchers }) => {
-      // A websocket push carries no request-issued time; it reflects current server state, so use
-      // receipt time. This preserves the pre-existing GC behaviour for pushed snapshots.
-      onNewLaunchers({ applicationId, applicationLaunchers, requestTimestamp: Date.now() });
+      // A pushed snapshot refreshes the visible list but must never be able to remove a workflow process:
```
teosarca: not sure what good this comment brings besides AI slop

e2e/mobile-webui/tests/harness/jobStartRecovery.test.js
```diff
@@ -0,0 +1,90 @@
+import { test } from '../../playwright.config';
+import { expect } from '@playwright/test';
+import { setCurrentPage } from '../utils/common';
+import { recoverToLauncherList } from '../utils/screens/jobStartRecovery';
+
+/**
+ * AC6 — the recovery DECISION, driven against a scripted fake page, with no browser.
```
teosarca: AC6 of what? pls review this and respect our guidelines, making comments timeless

misc/services/mobile-webui/mobile-webui-frontend/src/constants/LaunchersActionTypes.js
teosarca: **Resolution (dev):** Two separate points, answered separately.

**Why a new action type rather than a flag on `POPULATE_LAUNCHERS_COMPLETE`:** a flag has to be read and branched on, so the delete path stays reachable and is merely suppressed — one wrong condition, one refactor, or one future dispatcher that forgets the flag, and a live job gets deleted off the operator's screen again. With a separate type the reducer simply has no case for, that path is unreachable rather than conditionally skipped. Verified there is no fallthrough: `reducers/wfProcesses/workflow.js` has exactly one case calling `removeWFProcessesFromState`, and a grep of the whole `reducers/` tree finds no other consumer of either launcher action type. A new consumer would have to pick the pruning type on purpose.

**The naming — you are right, and it is changed.** `loadComplete` / `loadPushed` read as two spellings of the same verb and hid the only thing that matters, which is who initiated the snapshot. The value is now `launchers/pushedByServer`, so the pair reads as two different origins: this client asked for it, versus the server pushed it. I left the pre-existing `POPULATE_LAUNCHERS_COMPLETE` constant alone rather than renaming a symbol used by the REST path in a fix that is not about it. https://github.com/metasfresh/metasfresh/commit/fa5380724eb

misc/services/mobile-webui/mobile-webui-frontend/src/reducers/wfProcesses/workflow.js
teosarca: **Resolution (dev):** Agreed. Cut from 15 lines to 3, keeping only what is not derivable from the code — that a requested snapshot is stamped when the request was issued and so cannot claim knowledge of a job started later, and that attaching the prune to the action type is what keeps it unreachable by accident. The cache and dispatcher-provenance paragraphs are gone: both are facts about other files that would rot here. https://github.com/metasfresh/metasfresh/commit/fa5380724eb

misc/services/mobile-webui/mobile-webui-frontend/src/containers/wfLaunchersScreen/useLaunchers.js
teosarca: **Resolution (dev):** Agreed, it restated the code. Cut to two lines that carry the one non-obvious fact: this must not be routed through `populateLaunchersComplete`, because that action type is the one carrying the power to prune workflow processes. https://github.com/metasfresh/metasfresh/commit/fa5380724eb

e2e/mobile-webui/tests/harness/jobStartRecovery.test.js
teosarca: **Resolution (dev):** Both faults fixed, and treated as a class rather than this one line. A bare criterion number has no referent for anyone reading the code without that document in front of them, so it carried no information; and comments must not encode the process that produced the change.

Swept every file this branch touches: **13 acceptance-criterion labels and 3 requirements-document references across 6 files**, all replaced with a plain statement of what the code guarantees. A re-run of the sweep now returns nothing. The specific line you flagged now opens with what the test actually does — exercising the recovery decision against a scripted fake page with no browser — and its rationale is stated as a permanent property rather than as a story about the fix landing. https://github.com/metasfresh/metasfresh/commit/fa5380724eb

misc/services/mobile-webui/mobile-webui-frontend/src/constants/LaunchersActionTypes.js
```diff
@@ -4,6 +4,15 @@
...
 export const POPULATE_LAUNCHERS_START = 'launchers/loadStart';
 export const POPULATE_LAUNCHERS_COMPLETE = 'launchers/loadComplete';
+
+/**
+ * A launchers snapshot the server pushed over the websocket, as opposed to one this client requested.
+ * A separate type from POPULATE_LAUNCHERS_COMPLETE so that no reducer can prune on it.
+ * @constant
+ * @type {string}
+ */
+export const POPULATE_LAUNCHERS_PUSHED = 'launchers/pushedByServer';
```
teosarca: i think POPULATE_LAUNCHERS_PUSHED name shall align with its value "pushedByServer"

teosarca: **Resolution (dev):** Fixed, and you were right about the underlying cause too — I renamed the value with a script in the previous commit and never re-read the result, which is how the symbol and its value ended up disagreeing.

`POPULATE_LAUNCHERS_PUSHED` -> `POPULATE_LAUNCHERS_PUSHED_BY_SERVER`, and the action creator `populateLaunchersPushed` -> `populateLaunchersPushedByServer`, so symbol and value both say `pushedByServer`. All 6 call sites updated; a grep for the old names returns nothing.

Reading the files back properly also turned up three things my earlier sweep had missed, all now fixed in the same commit: three comment blocks the scripted edit had left ungrammatical, and process narration that survived because the sweep only matched criterion labels — `THE RED`, `passes before and after the fix`, `used to call`, `on the old path`, and a test case name reading `today's guard`. Those are exactly the non-timeless framing you objected to, so the sweep pattern was too narrow and has been widened.

19 tests pass, lint clean. https://github.com/metasfresh/metasfresh/commit/ce9b16221c1

PR #25700, opened 2026-09-02
Lieferschein print control for drop-ship business partners + copies-count fix

teosarca:
## What

Adds Lieferschein (shipment) print control for drop-ship business partners, plus a fix to the print copies-count resolution.

## Changes

**Config metadata (`C_BP_PrintFormat`)**
- New nullable columns `IsDropShip` and `IsAutoPrint` (YesNo) exposed as grid fields on the core partner windows (123, 540409 Pharma); `DocumentCopies_Override` made nullable.

**Copies-count fix**
- Dropped an erroneous `AD_PrintFormat_ID` filter in `BPartnerPrintFormatRepository.getByQuery` (a varargs-null → `IS NULL` predicate that excluded hooked rows), so the configured copies value is now honoured. RED+GREEN.

**Drop-ship auto-print suppression**
- `DocumentReportAdvisorUtil.resolveSuppressAutoPrint` + `isDropShip(order)` (drop-ship partner present AND a real delivery-address deviation), null-aware `IsAutoPrint`. Threaded through to the end-to-end gate: a transient `IArchiveBL.SUPPRESS_AUTO_PRINT` stamped in `DefaultModelArchiver.createArchive` and enforced as the first statement of `PrintingQueueBL.isIsCreatePrintingQueueItem`. Cucumber `shipmentAutoPrintDropShipSuppression.feature` covers drop-ship suppressed / stock enqueues / manual-reprint enqueues (0/1/1 queue items).

**Copies-override normalization (`DocumentCopies_Override` 0 → null)**
- `@ModelChange` interceptor on `C_BP_PrintFormat` normalizes new/changed rows; data migration `5821890` normalizes existing rows (`backup_table` first). `0` and `null` are already treated identically by copies-resolution, so observable behaviour is unchanged.

## Testing

- Unit + interceptor tests green; cucumber suppression feature green (local, this branch).
- Whole-state review over `origin/intensive_care_hotfix..HEAD` clean.


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/report/DocumentReportAdvisorUtil.java
```diff
@@ -142,4 +145,65 @@ private static PrintCopies getDocumentCopies(@Nullable final I_C_DocType docType
...
+			return false;
+		}
+		if (!order.isDropShip())
+		{
+			return false;
+		}
+
+		final BPartnerId orderBPartnerId = BPartnerId.ofRepoIdOrNull(order.getC_BPartner_ID());
+		final BPartnerId dropShipBPartnerId = BPartnerId.ofRepoIdOrNull(order.getDropShip_BPartner_ID());
+		final boolean bPartnerDeviates = dropShipBPartnerId != null && !dropShipBPartnerId.equals(orderBPartnerId);
```
teosarca: no need to check for bPartnerDeviates  since BPartnerLocationId  already incorporates the BP too

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/report/DocumentReportAdvisorUtil.java
```diff
@@ -142,4 +145,65 @@ private static PrintCopies getDocumentCopies(@Nullable final I_C_DocType docType
...
+	 * recipient deviate from the order's own sold-to?
+	 * <p>
+	 * NOTE -- this is a <b>different</b> rule from the sales-order ultimate-consignee (UC) resolution
+	 * documented in the {@code de.metas.business} module CLAUDE.md ("IsDropShip is irrelevant for UC
+	 * resolution -- the presence of DropShip_* alone decides"). That rule answers "who is the goods
+	 * recipient". This method answers a different question -- "should auto-print be suppressed for this
+	 * shipment" -- and the approved design for that concern deliberately gates on {@code IsDropShip='Y'}
+	 * in addition to the DropShip_* deviation. Do not "fix" this to match the UC rule.
+	 */
+	public boolean isDropShip(@Nullable final I_C_Order order)
```
teosarca: pls do not accept null params

backend/de.metas.business/src/main/java/de/metas/inout/InOutDocumentReportAdvisor.java
```diff
@@ -96,12 +100,16 @@ public StandardDocumentReportType getStandardDocumentReportType()
 
 		final Language language = util.getBPartnerLanguage(bpartner).orElse(null);
 
+		final I_C_Order order = getOrderOrNull(inout);
+		final boolean docIsDropShip = util.isDropShip(order);
```
teosarca: check order not null

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/report/DocumentReportAdvisorUtil.java
teosarca: **Resolution (dev):** Agreed. The manual `bPartnerDeviates` check was only needed because both `BPartnerLocationId`s were anchored on the order's own `C_BPartner_ID`, which neutralized the BP component. Now the drop-ship side is built with its natural key `BPartnerLocationId.ofRepoIdOrNull(order.getDropShip_BPartner_ID(), order.getDropShip_Location_ID())` and compared full-key against the order ship-to, so a single `!equals` covers both BP and location deviation and `bPartnerDeviates` is removed. https://github.com/metasfresh/metasfresh/commit/5a463083f76

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/report/DocumentReportAdvisorUtil.java
teosarca: **Resolution (dev):** Done — `isDropShip` now takes `@NonNull I_C_Order order` and the internal `if (order == null) return false;` guard is removed. https://github.com/metasfresh/metasfresh/commit/237a5be0284

backend/de.metas.business/src/main/java/de/metas/inout/InOutDocumentReportAdvisor.java
teosarca: **Resolution (dev):** Done — the null check moved to the caller: `InOutDocumentReportAdvisor` now does `order != null && util.isDropShip(order)` (manual shipment with no order = not a drop-ship = no suppression). https://github.com/metasfresh/metasfresh/commit/237a5be0284

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/report/DocumentReportAdvisorUtil.java
teosarca: **Resolution (dev):** Applied as you requested (natural-key `BPartnerLocationId`, `bPartnerDeviates` removed); current behavior is pinned + documented by a mirror test. One consequence for your call: a **location-only-deviating** drop-ship (`DropShip_BPartner_ID` unset but `DropShip_Location_ID` set and differing) now resolves to `false` (auto-print no longer suppressed) vs `true` before — reachable since `ProhibitInconsistentDropshipValues` is deactivated system-wide. Kept as-is per the literal simplification; if a location-only deviation should still suppress, say so and I will restore it.

backend/de.metas.printing/de.metas.printing.base/src/main/java/de/metas/printing/api/impl/PrintingQueueBL.java
```diff
@@ -219,6 +219,15 @@ private boolean isIsCreatePrintingQueueItem(final PrintArchiveParameters printAr
 
 		try (final MDC.MDCCloseable ignore = TableRecordMDC.putTableRecordReference(archive))
 		{
+			// Honor the transient suppress-auto-print flag stamped on the archive (e.g. a drop-ship shipment whose
+			// auto-print must be suppressed): skip printing-queue-item creation entirely, before any other rule below
+			// (including the IsProcessQueueItem short-circuit) could force it back on.
+			if (IArchiveBL.SUPPRESS_AUTO_PRINT.getValue(archive, false))
```
teosarca: pls do not access SUPPRESS_AUTO_PRINT directly. have methods for setting and getting the value from an archive given as param

teosarca: **Resolution (dev):** Agreed. Encapsulated the flag behind `IArchiveBL.setSuppressAutoPrint(archive, boolean)` / `isSuppressAutoPrint(archive)` (returns false when unset, preserving the prior default); both call sites (`DefaultModelArchiver` set, `PrintingQueueBL` get) now go through them, and the raw `SUPPRESS_AUTO_PRINT` dyn-attribute accessor is now `private static final` in the `ArchiveBL` impl — no direct access remains anywhere. Round-trip unit test added (`ArchiveBLTest`, 2/2 green); backend reactor BUILD SUCCESS. https://github.com/metasfresh/metasfresh/commit/f6d826a2b22

PR #25704, opened 2026-09-02
MobileUI job start — a pushed launchers snapshot can no longer delete a workflow process

teosarca:
Sideways port of https://github.com/metasfresh/metasfresh/pull/25697 to the deep_tundra family. The defect is in shared mobile core code, so fixing only the branch it was reported on leaves it live for the other families; propagation does not flow sideways, so this family needs its own port.

A launchers snapshot pushed over the websocket was dispatched as `POPULATE_LAUNCHERS_COMPLETE` stamped with the browser receipt time. The `wfProcesses` garbage collector reads that stamp as "this snapshot knows about everything older than it", so a snapshot the server had computed BEFORE a job was started looked newer than the job and deleted it from the store — the operator was thrown back to the applications menu mid job-start. Pushed snapshots now get their own action type that the `wfProcesses` reducer deliberately does not handle, so a push structurally cannot prune; the explicitly requested snapshot stamps at request-issue time and keeps doing all garbage collection.

One conflict, in the import header of `e2e/mobile-webui/tests/utils/screens/picking/PickingJobsListScreen.js`: an add/add where both sides only add import lines at the same position. Resolved as the union of both import blocks — dropping either side would break that side. Nothing else conflicted.

### DIFF
--- a/e2e/mobile-webui/tests/harness/jobStartRecovery.test.js
+++ b/e2e/mobile-webui/tests/harness/jobStartRecovery.test.js
@@ -1,0 +1,88 @@
+import { test } from '../../playwright.config';
+import { expect } from '@playwright/test';
+import { setCurrentPage } from '../utils/common';
+import { recoverToLauncherList } from '../utils/screens/jobStartRecovery';
+
+/**
+ * The recovery DECISION, driven against a scripted fake page, with no browser.
+ *
+ * A job start does not normally land on the applications menu, so a real E2E run never exercises the
+ * recovery branch. This is the only test that runs the recovery decision across all of its outcomes.
+ *
+ * How it can be browserless: tests/utils/common.js exports setCurrentPage(currentPage), which assigns
+ * the module-level `page` every screen helper reads. A test that does NOT destructure the `page`
+ * fixture never launches a browser, so it can install a fake and drive the helper against it.
+ *
+ * NEVER declare `{ page }` in a test signature in this file — that pulls in the real browser fixture
+ * (playwright.config.js's `page` override calls setCurrentPage(realPage)) and overwrites the fake.
+ *
+ * Scope note: this targets `recoverToLauncherList` rather than `tapLauncherUntilJobScreen`, because
+ * that one cannot be driven with a fake page at all — `resolveLauncherTapTarget` calls
+ * `expect(locator).toHaveCount(1)`, and Playwright's `expect` only accepts a real Locator. That is
+ * exactly why the recovery decision lives in its own module with no `expect` in it.
+ */
+
+// Which selectors are currently attached to the fake DOM.
+let attached = {};
+// Selector -> the attachment changes a tap on it causes (the navigation it triggers).
+let afterTap = {};
+// Every selector tapped, in order.
+let tapped = [];
+
+const fakePage = {
+    locator: (selector) => ({
+        waitFor: async () => {
+            if (!attached[selector]) {
+                throw new Error(`fakePage: ${selector} is not attached`);
+            }
+        },
+        tap: async () => {
+            tapped.push(selector);
+            Object.assign(attached, afterTap[selector] ?? {});
+        },
+    }),
+};
+
+test.beforeEach(() => {
+    attached = {};
+    afterTap = {};
+    tapped = [];
+    setCurrentPage(fakePage);
+});
+
+test('recoverToLauncherList: already back on the launcher list -> launcherList, no tap', async () => {
+    attached['#WFLaunchersScreen'] = true;
+
+    const outcome = await recoverToLauncherList({ applicationId: 'picking' });
+
+    expect(outcome).toBe('launcherList');
+    expect(tapped).toEqual([]);
+});
+
+test('recoverToLauncherList: landed on the applications menu and re-entry succeeds -> recovered', async () => {
+    attached['#ApplicationsListScreen'] = true;
+    afterTap['#picking-button'] = { '#WFLaunchersScreen': true };
+
+    const outcome = await recoverToLauncherList({ applicationId: 'picking' });
+
+    expect(outcome).toBe('recovered');
+    expect(tapped).toEqual(['#picking-button']);
+});
+
+test('recoverToLauncherList: neither screen attached (a navigation still in flight) -> unknown, no tap', async () => {
+    const outcome = await recoverToLauncherList({ applicationId: 'picking' });
+
+    expect(outcome).toBe('unknown');
+    expect(tapped).toEqual([]);
+});
+
+// This case matters most: a broken re-entry must never authorise a re-tap of the launcher, which
+// could double-start the workflow.
+test('recoverToLauncherList: menu attached but re-entry does not bring up the list -> unknown', async () => {
+    attached['#ApplicationsListScreen'] = true;
+
+    const outcome = await recoverToLauncherList({ applicationId: 'picking' });
+
+    expect(outcome).toBe('unknown');
+    expect(tapped).toEqual(['#picking-button']);
+});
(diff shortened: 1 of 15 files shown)


teosarca:
## CI audit — this PR is green, and nothing was introduced by it

[This CI run](https://github.com/metasfresh/metasfresh/actions/runs/33610987197) on [the current head commit](https://github.com/metasfresh/metasfresh/commit/8d9820e38acdd10498f1846a8c1f5c89709c41c0) **concluded successfully on attempt 3** — every check green, mobile suite `303 passed (44.2m)`, and that build's `failures.json` reports 0 failures in every suite (cucumber 0/1262, frontend-webui 0/156, mobile-webui 0/303).

Recording what went red on the way there, so the next person who meets the same red knows what it is. **All of it is intermittent, on an unchanged commit.**

| Attempt | Check | Failing test | Classification | Evidence it is not ours |
|---|---|---|---|---|
| [1](https://github.com/metasfresh/metasfresh/actions/runs/33610987197/job/100192342299) | `mobile test` — 2 failed / 301 passed | `picking/productNames.spec.js:64` (26.1 s) and `picking/productNamesBlockLayout.spec.js:77` (24.1 s) | Known flake — internal flaky-case 76 (launcher-caption attach seam) | Same-commit pass, attempt 3 (below) |
| [1](https://github.com/metasfresh/metasfresh/actions/runs/33610987197/job/100192342578) | `frontend webui test (1/3)` — 1 failed, 2 flaky, 49 passed | hard: `advanced-edit-focus.spec.js:170`, 3/3 including both retries. Retry-recovered: `advanced-edit-focus.spec.js:509`, `document-references.spec.js:138` | Known flake — flaky-case 42 (`advanced-edit-focus` nav/URL seam) and flaky-cases 45 + 30 (`document-references` Alt+6 reference-panel seam) | Same-commit pass, attempt 3 (below) |
| [2](https://github.com/metasfresh/metasfresh/actions/runs/33610987197/job/100240863128) | `mobile test` — 2 failed / 301 passed | `picking/productNames.spec.js:64` (26.3 s) and `picking/productNamesBlockLayout.spec.js:157` (24.4 s) | Known flake — flaky-case 76, same seam | `productNamesBlockLayout.spec.js:77` **passed in 3.4 s** here after failing in attempt 1 |
| 3 | — | — | — | all green |

**The mobile seam.** `TimeoutError: locator.waitFor: Timeout 20000ms exceeded` on `.wflauncher-button[data-salesorderid="…"]` `.filter({ hasText: <full caption> })`, thrown at `PickingJobsListScreen.js:97` ← `:95` ← `expectJobButtons(…:94)`. Four call-sites across three spec files share it; the shared helper has never been edited — each call-site arrived by the `expectJobButtons([…])` shape being copied into a new caption spec, so the call-site count measures how often the assertion was copied, not how many defects exist.

**Per-call-site outcome on [the identical head commit](https://github.com/metasfresh/metasfresh/commit/8d9820e38ac):**

| Call-site | Attempt 1 | Attempt 2 | Attempt 3 |
|---|---|---|---|
| `productNames.spec.js:64` | FAIL 26.1 s | FAIL 26.3 s | **pass 6.0 s** |
| `productNamesBlockLayout.spec.js:77` | FAIL 24.1 s | pass 3.4 s | pass 6.1 s |
| `productNamesBlockLayout.spec.js:157` | pass 3.6 s | FAIL 24.4 s | pass 5.4 s |
| `productNamesDetail.spec.js:11` | pass 4.1 s | pass 4.1 s | pass 6.3 s |

`productNames.spec.js:64` was the one worth resolving: it was 0/2 on this branch while passing 3/3 on [the diff-free base commit](https://github.com/metasfresh/metasfresh/commit/182905cf711) ([in this run](https://github.com/metasfresh/metasfresh/actions/runs/33594836356), 5.3 s / 5.2 s / 5.5 s), and this PR does change the production code that builds the launcher rows the assertion reads. Three independent lines of evidence settle it as intermittent, not a regression:

1. **Same-commit pass (the proof standard).** Attempt 3's [`mobile test`](https://github.com/metasfresh/metasfresh/actions/runs/33610987197/job/100257063692) passed it in **6.0 s** on the byte-identical commit, along with all four call-sites and 303/303 tests.
2. **Reachability: none.** `productNames.spec.js` calls only `expectJobButtons` (spec `:92`, `:125`, `:158`) — never `startJob`, never `tapLauncherUntilJobScreen`, the only two functions this diff changes in `PickingJobsListScreen.js`. `expectJobButtons` and `locateJobButtons` are untouched; their line numbers moved 86/89 → 94/97 purely because a 6-line import comment was inserted above them. On the production side, `POPULATE_LAUNCHERS_COMPLETE` and the new `POPULATE_LAUNCHERS_PUSHED_BY_SERVER` share **one** case block in `reducers/launchers.js`, writing the identical `{ isLoading, filterByQRCode, list, actions }`; neither writes `requestTimestamp` (only `POPULATE_LAUNCHERS_START` does), so the REST-refetch effect in `useLaunchers.js` — whose dependency array reads that field — is unchanged. The subscription-teardown guard is a `let` declared inside the `useEffect` body, i.e. a fresh binding per invocation, set `false` only by that invocation's own cleanup, so it cannot mute a live subscription. The single surviving behavioural delta is that a frame arriving from an **already torn-down** subscription is now dropped; at the point this assertion runs the launcher screen has just mounted and its subscription is live.
3. **Prior history on launcher-free branches.** `productNames.spec.js:64` has hit this same seam repeatedly on branches whose diffs contain no launcher, mobile-webui or E2E code at all.
4. **No global break.** Three sibling caption call-sites pass in 3.4–4.1 s in the very same jobs; a deterministic break in launcher-row rendering would take all four.

**The frontend seam.** `advanced-edit-focus.spec.js:170` (flaky-case 42) — `TimeoutError: page.waitForURL: Timeout 20000ms exceeded` in the spec-local `openSingleRecord` helper at `:74`: the click lands on the grid view (`…/window/140?page=1&sort&viewId=140-B`) and the single-record URL never arrives. In attempt 3 the same test failed the first try (24.7 s) and **passed on retry #1 in 4.9 s** on the identical commit — the shard concluded green with 4 flaky, 48 passed. Reachability: a desktop-WebUI spec against `frontend/`; this diff contains no desktop-frontend file. `document-references.spec.js:138` recovered on retry in both attempts (first `expect(found).toBe(true)` at spec `:279`, then a null reference URL) — the well-known Alt+6 reference-panel materialisation settle — flaky-case 45 for the presence signature (`found` false), flaky-case 30 for the null-URL one, which the registry records as two signatures of one underlying defect. Also desktop-only.

**Nothing in this PR introduced any of it.**

teosarca:
**Resolution (dev): CI red triaged — not caused by this diff.**

The `test (cucumber) (profile5)` check failed twice, in two different ways, neither reachable from this PR:

- **Attempt 1 — flaky-case 92** (WarpBuild runner lost communication). The job recorded zero steps, ran exactly 10m00s, and the only diagnostic is the check-run annotation *"The self-hosted runner lost communication with the server."* Infrastructure, check-name-agnostic.
- **Attempt 2 — flaky-case 93** (new). `deliveryInstructions.feature:13` *"Generate and regenerate delivery instructions from outgoing delivery planning"* times out at `:53` (`after not more than 30s, load created M_Delivery_Planning`) with zero `M_Delivery_Planning` rows created.

Why case 93 is not a regression from this branch:

1. **The diff cannot reach the suite.** This PR is 15 files, +786/−27, entirely under `e2e/mobile-webui/**` and `misc/services/mobile-webui/mobile-webui-frontend/**` — no backend Java, no `de.metas.cucumber` file, no SQL migration.
2. **Green on this PR's own first commit.** `profile5` succeeded at commit 1 of 2 and failed at commit 2 of 2; that delta is 3 files, +9/−9 — a variable rename in two mobile e2e screen helpers plus `mobile-webui-frontend/src/api/launchers.js`.
3. **Backend-identical control passed.** The merge base on `deep_tundra_release` carries byte-identical backend content (this PR touches zero backend files) and its run passed `profile5`.
4. **Base branch is green.** The last five `deep_tundra_release` runs all record `test (cucumber) (profile5) = success`.

Residual uncertainty, stated plainly: there is no *literal* same-SHA pass for case 93 (attempt 1 never ran a scenario, and no further re-run was fired), so the case is recorded at **medium** confidence, not high.

Root cause is a hypothesis, not a verified finding. `M_ShipmentSchedule.createDeliveryPlanning` gates on `isAutoCreateEnabled(...)` before scheduling an on-commit work package, so `count == 0` fits either "never scheduled" or "scheduled but not drained within 30s". The log cannot discriminate — that work package emits no log lines. No follow-up issue filed; the escalation bar is not met.

PR #25708, opened 2026-09-02
MobileUI Picking — order the launcher caption product names by sales-order line

teosarca:
### What

Fix the **intermittent CI failures** of the MobileUI picking launcher-caption specs `e2e/mobile-webui/tests/spec/picking/productNamesDetail.spec.js` and `productNames.spec.js`, by making the launcher caption's product-name order **deterministic** — sales-order line (`C_OrderLine.Line`) instead of the arbitrary `M_Picking_Job_Line_ID`.

(This is a tracked known flaky test; the fix↔case linkage is recorded in the internal flaky-test registry — a public PR can't carry that private reference.)

### The flake

Both specs assert one exact caption string via a `hasText` filter on the launcher row (`.wflauncher-button[data-salesorderid=…].filter({hasText: '<docNo> | <bp> | P1, P2, P3'})`). The product-name order in that caption was **arbitrary**: the loader ordered the names by `M_Picking_Job_Line_ID`, which is assigned in the order the *unordered* packageable query returns the lines (`PickingJobCreateCommand.toPackageableQuery` sets no `orderBy`, and `PackagingDAO`'s sort keys don't include the sales-order line). So on a run where the order came out differently from what the spec created, the exact-string `hasText` never matched and the spec timed out — intermittently, once per unlucky data layout.

Evidence this is an ordering flake (not a row-settle/timing one): the failure-time page snapshot showed **all three product names present, in a different order** than asserted — run https://github.com/metasfresh/metasfresh/actions/runs/33594836356.

### Fix

Order both caption paths by `C_OrderLine.Line`, tie-broken by the picking-job-line id, so the caption is deterministic and matches the order the picker reads off the sales document:
- launcher list — `PickingJobLoaderAndSaver.extractProducts`
- opened job — `PickingJob.getProductNameParts` (consumed by `DisplayValueProvider`)

`PickingJob.lines`' own id-order is left untouched (workflow logic that iterates `getLines()` is unaffected). Zero new SQL: the seq numbers are batch-warmed once per load via `warmUpSalesOrderLineSeqNosCache` in `loadRecordsFromDB` (mirrors `warmUpSalesOrderDocumentNosCache`); every read is a cache hit — this also removes a pre-existing per-line query in `loadLine`.

### Tests + validation

`PickingJobLoaderAndSaverProductOrderTest` pins the sales-order-line order for both caption paths (three ordering axes — line-id, product-id, SeqNo — deliberately in disagreement so only a SeqNo sort passes; RED against the old id-sort, GREEN after). Local: `de.metas.handlingunits.base` 1062/0/0, `de.metas.picking.rest-api` 59/0/0, `de.metas.workflow.rest-api` 19/0/0. In CI here, the `mobile test` job — which runs these specs — has passed on every run of this branch, i.e. the flake has not recurred.

### DIFF
--- a/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/model/PickingJob.java
+++ b/backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/picking/job/model/PickingJob.java
@@ -26,6 +26,8 @@ import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.google.common.collect.ImmutableList;
+
+import java.util.Comparator;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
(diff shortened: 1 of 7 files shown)


teosarca:
### CI note — the red on this branch is an unrelated known flake, not this change

This PR's diff is backend Java only (picking-caption product-name ordering in the loader). The one CI job that failed on it is **`test (cucumber) (profile7)`**, which failed on a **shipment-schedule unique-constraint race** — a known intermittent cucumber flake in a domain this diff does not touch.

Evidence it is not this change:

- **`mobile test` — the job that exercises this fix (the mobile launcher-caption specs) — passed** on run https://github.com/metasfresh/metasfresh/actions/runs/33629065199. The caption product-name-order flake this PR fixes did **not** recur.
- The changed files are all under `de.metas.handlingunits.base` / `de.metas.*.rest-api` picking-caption code; the failing scenario is shipment-schedule creation — no code path from one to the other.
- The failure **recurred on a same-partition `--failed` re-run** (a partition-sensitive flake signature), so a **fresh full run on an independent executor partition** was triggered to confirm: https://github.com/metasfresh/metasfresh/actions/runs/33637733131.

**Merge once that fresh run is green.** The Merge Gate here stays red only until it completes.

PR #25712, opened 2026-09-02
Doc_PPCostCollector: always post material-receipt qty to P_Asset even at zero cost

teosarca:
## Defect

`Doc_PPCostCollector.createFacts_MaterialReceipt` gated the P_Asset receipt fact line on the **amount** (`costsReceived != 0`). When the product's cost for the active costing method is zero — e.g. a manufactured product freshly on Moving Average Invoice, or any zero-cost receipt — the entire receipt fact line was skipped, dropping **both** the amount **and** the received quantity.

The Lagerwert / inventory-value report sums `Fact_Acct.qty` on P_Asset, so the received manufactured quantity silently vanished from stock valuation. DB-confirmed on a production copy: product 103778 P_Asset went 4608 -> 384 qty and 12,725.92 -> 0.00 value.

## Fix

- **Gate each receipt leg on its QUANTITY, not its amount.** The receipt leg posts when `qtyReceived.signum() != 0` (`costsReceived != 0` implies `qtyReceived != 0`, so all prior postings are preserved and the zero-cost / non-zero-qty case is now posted too). The scrap leg posts when `qtyScrapped.signum() != 0 || costsScrapped.signum() != 0` — carrying the scrapped qty into valuation and still posting a sub-precision rounding remainder, as before.
- **`createFactLines` calls `FactLineBuilder.alsoAddZeroLineIf(alsoAddZeroLine)`** on both the debit and credit legs, so a zero-amount line is not dropped by `FactLineBuilder.build()` when the caller passes `alsoAddZeroLine=true` (material receipt and component issue). The flag stays `false` for the variance / activity-control legs, which keep dropping their zero-amount-and-qty lines.
- The leg logic is inlined directly in `createFacts_MaterialReceipt` (no helper class), matching the surrounding code idiom.

## Test

Cucumber scenario `@Id:S26253_TC1` in `manufacturingCostCollectorPosting.feature`: a manufacturing order on **Moving Average Invoice** whose finished good carries a **zero** current cost is manufactured and received; the material-receipt cost collector must still post the received quantity to `P_Asset_Acct` — `Fact_Acct` qty `10` on P_Asset (qty `-10` on WIP) with amount `0` on both legs. Before the fix the whole fact line (amount AND qty) was dropped, so the two expected `Fact_Acct` rows are absent (RED); with the fix they are posted (GREEN).

The prior `Doc_PPCostCollectorMaterialReceiptTest` unit test was removed in favour of this end-to-end cucumber scenario, which exercises the real posting pipeline rather than the leg-decision layer in isolation.


backend/de.metas.manufacturing/src/main/java/de/metas/manufacturing/acct/Doc_PPCostCollector.java
```diff
@@ -306,7 +347,7 @@ private List<Fact> createFacts_ComponentIssue(final AcctSchema as)
 			// so getCreateCosts returns a negative cost. Posting it uncompensated would invert the direction
 			// (DR-WIP negative / CR-Asset negative). Compensate the sign here — mirroring createFacts_Variance —
 			// so a component issue posts DR P_WIP_Acct (positive) / CR P_Asset_Acct (positive): inventory down, WIP up.
-			final Fact fact = createFactLines(as, element, debit, credit, costs.negate(), qtyIssued.negate());
+			final Fact fact = createFactLines(as, element, debit, credit, costs.negate(), qtyIssued.negate(), false);
```
teosarca: imho material issue shall also book all the time, even if zero.... so u might check if the param is needed or it can be assumed always create fact line

backend/de.metas.manufacturing/src/main/java/de/metas/manufacturing/acct/Doc_PPCostCollector.java
```diff
@@ -251,20 +254,13 @@ private List<Fact> createFacts_MaterialReceipt(final AcctSchema as)
 					.roundToPrecisionIfNeeded(as.getStandardPrecision());
 			final CostAmount costsScrapped = costs.subtract(costsReceived);
 
-			if (costsReceived.signum() != 0)
+			for (final MaterialReceiptLeg leg : materialReceiptLegs(qtyReceived, qtyScrapped, costsReceived, costsScrapped))
```
teosarca: why materialReceiptLegs was needed?

why not simply alwyas post in case of qtyReceived.

and in case of qtyScrapped, post only if qtyScrapped or costsScrapped is not zero.

backend/de.metas.manufacturing/src/main/java/de/metas/manufacturing/acct/Doc_PPCostCollector.java
teosarca: Done — material issue (ComponentIssue) now books unconditionally (`alsoAddZeroLine=true`), same as the receipt legs: https://github.com/metasfresh/metasfresh/commit/e3b8c40557f

I kept the boolean param rather than making `createFactLines` unconditional because it is also shared by `createFacts_ActivityControl` and the four `createFacts_Variance` types; forcing a zero line there would post empty zero-amount/zero-qty facts they currently drop. If you want those to always book too, I'll drop the param and make it unconditional — just say so.

backend/de.metas.manufacturing/src/main/java/de/metas/manufacturing/acct/Doc_PPCostCollector.java
teosarca: The posting logic is exactly what you describe: the receipt leg posts when `qtyReceived != 0`, and the scrap leg posts when `qtyScrapped != 0 || costsScrapped != 0` (updated in https://github.com/metasfresh/metasfresh/commit/e3b8c40557f).

`materialReceiptLegs` exists only as a unit-test seam. A full posting test is not feasible in this module — `getAccount` is raw SQL and `getCreateCosts` needs a live DB, so every posting test here is `@Disabled`/manual. Extracting the guard decision lets it be unit-tested (RED before the fix, GREEN after), mirroring the existing `costDifferenceDistributionLegs` in the same class. If you'd rather inline it and drop that unit coverage, I'll inline it.

teosarca: dude, i would prefer to keep this code simple, drop the unit test and just have a cucumber test

backend/de.metas.manufacturing/src/main/java/de/metas/manufacturing/acct/Doc_PPCostCollector.java
teosarca: **Resolution (dev):** Confirmed — after dropping the `materialReceiptLegs` helper, the material issue (`ComponentIssue`) leg still books unconditionally: `createFacts_ComponentIssue` calls `createFactLines(..., alsoAddZeroLine=true)`, so a zero-cost/zero-qty issue still posts. The `alsoAddZeroLine` param is kept only so the variance / activity-control legs keep dropping their zero-amount-and-qty lines. https://github.com/metasfresh/metasfresh/commit/81e7454dbba8aadbafc09f1da34f796953a77f50

backend/de.metas.manufacturing/src/main/java/de/metas/manufacturing/acct/Doc_PPCostCollector.java
teosarca: **Resolution (dev):** Done. Dropped the `materialReceiptLegs` helper — the receipt leg now posts inline when `qtyReceived != 0` and the scrap leg when `qtyScrapped != 0 || costsScrapped != 0`. Removed the `Doc_PPCostCollectorMaterialReceiptTest` unit test and added a cucumber scenario in `manufacturingCostCollectorPosting.feature` covering a zero-cost MovingAverageInvoice material receipt that still posts the received quantity to P_Asset (Fact_Acct qty != 0 with amount 0). https://github.com/metasfresh/metasfresh/commit/81e7454dbba8aadbafc09f1da34f796953a77f50

teosarca: **Resolution (dev):** Scrap leg kept as-is per your Option C sign-off. The `qtyScrapped != 0 || costsScrapped != 0` guard has no dedicated cucumber coverage, and adding it is not feasible without ballooning scope: the mobile `ReceiveFrom` flow (`JsonManufacturingOrderEvent.ReceiveFrom`) carries no scrap field and builds `ReceiptCostCollectorCandidate` with zero scrap, and `PP_Cost_Collector_StepDef` is assertion-only — so a scrapped qty cannot be driven through the existing steps without a production DTO field + new step-def support, out of proportion to this hotfix. The two receipt-leg comments were reworded so `alsoAddZeroLine=true` reads as belt-and-suspenders (both legs already qty-gated): https://github.com/metasfresh/metasfresh/commit/952d4b43382ab6fa6c243667938c4e6ac57a677e

PR #25714, opened 2026-09-03
MobileUI Manufacturing — edit configured HU attributes at the Material Receipt node

teosarca:
### What this does
At the **finished-goods Material Receipt** node of the MobileUI Manufacturing app (receiving the output of a production order), the operator can now edit handling-unit attributes directly on the receive dialog:

- **Lot number** and **Best-before date** remain **editable by default** after deployment (unchanged user-visible behavior).
- **Additionally, any configured `M_Attribute`** — value type **string, number, date, or LIST** (a dropdown of the attribute's allowed values) — can be made editable at receipt, driven by a **global configuration**. The LIST case is the driving use (e.g. a custom list attribute such as "size (cm)").
- Entered values are carried onto the **produced HU(s)** of the receive line, at the same HU level as Lot/Best-before today.
- Editing is **optional** (empty → no attribute set; the receive completes normally). No attribute is mandatory in this version, even if the underlying `M_Attribute` is `IsMandatory`.
- The section is offered on **every received line** (main finished good, co-products, by-products) whose product's attribute set carries the configured attribute; only **instance-level** attributes are offered; an attribute that doesn't apply to the product is silently **not shown**.
- Field order follows the config's `SeqNo`, so the admin controls the order the operator sees.
- **Automatic lot number** is preserved: a blank Lot at receipt still fires the configured `LotNo` sequence; a typed Lot still suppresses it (the sequence is not consumed).

### Design highlight — one channel, one thin backend branch
Rather than keep Lot/Best-before on their old dedicated code path alongside a second generic mechanism, they were **unified into the generic editable-attribute channel**: one config list, one JSON contract (`editableAttributes: List<JsonAttribute>`, reusing the existing LIST-capable Inventory mobile-UI DTO), one uniform frontend section, and one generic HU attribute-storage write.

The one genuinely special behavior — auto-lot timing — is kept correct by a **thin backend apply-branch** in `ReceiveGoodsCommand`: Lot / Best-before / Production date are routed to the existing `IPPOrderReceiptHUProducer` setters (at HU creation time), while all other attributes are written post-creation via the generic attribute-storage seam. Feasibility was verified: those special attributes have no creation-time consumer, so moving the rest to a post-creation apply is behaviorally equivalent.

### New AD_Table — justification
A new child table **`MobileUI_MFG_Config_Attribute`** (`MobileUI_MFG_Config_ID` + `M_Attribute_ID` + `SeqNo` + `IsActive`, with a unique index) is added. It holds the **global, ordered list of editable attribute codes** (active rows in `SeqNo` order), **replacing the retired per-attribute boolean flags** (`IsLotNumberEditable` / `IsBestBeforeDateEditable`) previously on the config. Presence of an active row = editable at receipt. It is modeled on the shipped HU-Manager display-config precedent (`MobileUI_HUManager_Attribute`).

Migration preserves existing state: for every config, both standard rows (Lot number, Best-before date) are seeded with `IsActive` = the old boolean's resolved value — so vanilla remains "Lot + Best-before editable", identical to today; the old flag columns are then retired.

An **admin child tab** for editing this list **ships in this same PR** on window 541788 (MobileUI Manufacturing Configuration), so the table is inspectable/editable in the WebUI — satisfying the new-AD_Table review requirement. The desktop admin tab itself is not covered by a dedicated Playwright test (the config is exercised via SQL / the test-masterdata API in the automated tests, mirroring the shipped HU-Manager precedent tab, which is likewise not UI-tested); this is a deliberate, justified N/A, not a coverage gap.

### Special-code hardening
Lot / Best-before / Production date are kept on the producer-setter path and are **excluded from the generic post-creation attribute apply at runtime** via a `Map.remove()` of those codes in `ReceiveGoodsCommand` (not merely by convention/comment). So even if one of those codes were ever added to the editable-attribute config, it would **not double-apply**.

### Config default semantics
The DEFAULT (Lot + Best-before editable) applies only when **no config row exists**. A config row that is present but has **no active attribute rows** is authoritative as an explicit "nothing editable" — it is not silently reset back to the Lot/Best-before default. This preserves an explicit tenant opt-out while keeping vanilla (no row) behavior unchanged.

### Testing
- **Playwright (mobile-webui E2E):** the editable Lot/Best-before path; and the generic LIST-attribute path — the produced HU carries the selected value, a not-configured attribute is not shown, field order follows the config, and a co-product line offers the same configured attribute.
- **Cucumber (backend):** receive-with-attributes for string/number/date values; and the auto-lot gate (blank Lot → sequence fires; typed Lot → sequence not consumed).
- **Unit:** the config repository (loading the effective editable-attribute list, incl. the no-row-vs-empty-row semantics) and precedence between the dedicated and generic attribute maps.

All green on CI.


backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/mobileui/manufacturing/Receive_with_CatchWeight.feature
```diff
@@ -108,3 +124,399 @@ Feature: mobileUI Picking - Pick mixed lines
       | Produced_LU | WeightNet            | 0.5         |            |         |
       | Produced_LU | HU_BestBeforeDate    |             | 2025-03-03 |         |
       | Produced_LU | Lot-Nummer           |             |            | LotNo_1 |
+
+# ######################################################################################################################
+# ######################################################################################################################
+# ######################################################################################################################
+# ######################################################################################################################
+  @from:cucumber
```
teosarca, line 161: instead of all these cucumber tests i woould prefer playwright tests! mobile ui shall be tested by playwright

backend/de.metas.manufacturing/src/main/sql/postgresql/system/30-de.metas.adempiere.libero/5821600_sys_mobileMfg_retireEditableAttributeFlagColumns.sql
```diff
@@ -0,0 +1,104 @@
...
+	SELECT COUNT(*) INTO v_count
+	FROM MobileUI_UserProfile_MFG
+	WHERE IsLotNumberEditable IS NOT NULL OR IsBestBeforeDateEditable IS NOT NULL;
+
+	IF v_count > 0 THEN
+		RAISE EXCEPTION 'Aborting drop of MobileUI_UserProfile_MFG.IsLotNumberEditable/IsBestBeforeDateEditable: % row(s) in MobileUI_UserProfile_MFG have a non-NULL per-user override flag set - migration refused to avoid data loss', v_count;
+	END IF;
+END $$;
+
+SELECT db_alter_table('MobileUI_UserProfile_MFG', 'ALTER TABLE public.MobileUI_UserProfile_MFG DROP COLUMN IsLotNumberEditable, DROP COLUMN IsBestBeforeDateEditable')
```
teosarca, line 88: make sure u backup_table **MobileUI_UserProfile_MFG** before dropping columns!

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/mobileui/manufacturing/Receive_with_CatchWeight.feature
teosarca: **Resolution (dev):** Agreed on the principle — the mobile receive UI is covered by Playwright: the mandatory mobile Playwright specs for this flow (`receiving_generic_attributes*`, catch-weight scan regression) are the test-phase coverage being added before merge, per the mobile-UI-testing mandate.

The open point is only whether to *remove* the cucumber scenarios in this file. All 8 currently also assert server-side contract that a mobile Playwright test can't reach at the same fidelity: dual-channel → 422 and off-list-code → 422 reject guards, and auto-lot firing/suppression at HU-creation time (BOM lot-number sequence). Proposed handling: keep only those pure backend-contract scenarios and move the receive happy-path to Playwright.

Confirming with you (reviewer) which of the 8 to drop before editing — test removal is gated on your sign-off. Will update this thread with the commit once the split is agreed.

Local validation this session (not CI-quoted): unit `de.metas.manufacturing.rest-api` 123/123 green; cucumber `@Id:S31771` 8/8 passed.

e2e/mobile-webui/tests/spec/manufacturing/receiving_editable_attributes.spec.js
```diff
@@ -8,17 +8,46 @@ import { ManufacturingJobScreen } from '../../utils/screens/manufacturing/Manufa
...
+            // Per-run product category + its attribute set - products below reference it by key, and the
+            // attributes below link into 'mfgAttrSet' by name.
+            productCategories: { 'mfgCat': { attributeSetName: 'mfgAttrSet' } },
+            attributes: {
+                // Upserts (by Value) the two pre-existing standard attributes and links them into the
+                // per-run attribute set as instance attributes - so the products' category resolves a set
+                // that carries Lot / Best-before, independent of the seeded standard category.
+                'lotNumberAttr': {
+                    value: 'Lot-Nummer',
+                    isInstanceAttribute: true,
```
teosarca: pls make isInstanceAttribute as default true, not need to mention here.

e2e/mobile-webui/tests/spec/manufacturing/receiving_generic_attributes.spec.js
```diff
@@ -0,0 +1,235 @@
...
+import { test } from '../../../playwright.config';
+import { allure } from 'allure-playwright';
+import { LoginScreen } from '../../utils/screens/LoginScreen';
+import { ApplicationsListScreen } from '../../utils/screens/ApplicationsListScreen';
+import { ManufacturingJobsListScreen } from '../../utils/screens/manufacturing/ManufacturingJobsListScreen';
+import { ManufacturingJobScreen } from '../../utils/screens/manufacturing/ManufacturingJobScreen';
+import { MaterialReceiptLineScreen } from '../../utils/screens/manufacturing/receipt/MaterialReceiptLineScreen';
+import { GetQuantityDialog } from '../../utils/screens/picking/GetQuantityDialog';
+
+// The mfg editable-attribute list generalizes beyond the fixed Lot/Best-before pair to ANY configured
```
teosarca: pls reduce AI slop, keep it CSP human readable and short

e2e/mobile-webui/tests/spec/manufacturing/receiving_generic_attributes.spec.js
```diff
@@ -0,0 +1,235 @@
...
+                        { value: 'M', name: 'Medium' },
+                        { value: 'L', name: 'Large' },
+                    ],
+                    attributeSetNames: ['mfgAttrSet'],
+                },
+                // Present on the attribute set and instance-level, but never added to the mfg config's
+                // editableAttributes above - proves a not-configured attribute is not shown.
+                'notConfiguredAttr': {
+                    value: 'TestNotConfiguredCM',
+                    name: 'Not Configured Attr',
```
teosarca, line 57: name shall not be mandatory and in that case the value shall be used.

regarding the value, what happens if value already exists?

i would suggest to even make value not mandatory and in case is not specifeid then a generated value to be used.
i would avoid using the value and go wirh generated values so each test has its own attribute.
same for product category too.

in tests, the attribute shall be referenced by its identifier and not by "value".

e2e/mobile-webui/tests/spec/manufacturing/receiving_generic_attributes.spec.js
```diff
@@ -0,0 +1,235 @@
...
+                receivedHUs: [{ lu: 'lu1', qty: '4 PCE' }],
+            },
+        },
+        hus: {
+            'lu1': {
+                storages: { 'BOM': '4 PCE' },
+                tus: [
+                    {
+                        storages: { 'BOM': '4 PCE' },
+                        attributes: { 'TestSizeCM': 'M' },
```
teosarca: attributes shall be referenced by identifier also (first check)!

e2e/mobile-webui/tests/spec/manufacturing/receiving_generic_attributes.spec.js
```diff
@@ -0,0 +1,235 @@
...
+
+    // The by-product line (index 2, per the by-products convention - see receiving_by_products.spec.js)
+    // offers the same configured attribute, because BY_PRODUCT is in the same 'mfgCat' category and so
+    // resolves the same 'mfgAttrSet' attribute set.
+    await ManufacturingJobScreen.clickReceiveButton({ index: 2 });
+    await MaterialReceiptLineScreen.selectNewLUTarget({ luPIItemTestId: masterdata.packingInstructions.BY_PRODUCT_PI.luPIItemTestId });
+
+    await page.getByTestId('receive-qty-button').tap();
+    await GetQuantityDialog.waitForDialog();
+    await GetQuantityDialog.expectEditableAttributeVisible('TestSizeCM');
```
teosarca: attributes shall be referenced by identifier also (first check)!

e2e/mobile-webui/tests/spec/manufacturing/receiving_generic_attributes_valuetypes.spec.js
```diff
@@ -0,0 +1,251 @@
...
+import { test } from '../../../playwright.config';
+import { allure } from 'allure-playwright';
+import { LoginScreen } from '../../utils/screens/LoginScreen';
+import { ApplicationsListScreen } from '../../utils/screens/ApplicationsListScreen';
+import { ManufacturingJobsListScreen } from '../../utils/screens/manufacturing/ManufacturingJobsListScreen';
+import { ManufacturingJobScreen } from '../../utils/screens/manufacturing/ManufacturingJobScreen';
+import { MaterialReceiptLineScreen } from '../../utils/screens/manufacturing/receipt/MaterialReceiptLineScreen';
+import { GetQuantityDialog } from '../../utils/screens/picking/GetQuantityDialog';
+
+// Value-TYPE coverage for the mfg editable-attributes list: the generic case (receiving_generic_attributes.spec.js)
```
teosarca: avoid ai slop

e2e/mobile-webui/tests/spec/manufacturing/receiving_generic_attributes_valuetypes.spec.js
```diff
@@ -0,0 +1,251 @@
...
+    await GetQuantityDialog.expectEditableAttributeVisible(NUMBER_CODE);
+    await GetQuantityDialog.typeEditableAttribute(NUMBER_CODE, '42.5');
+    await GetQuantityDialog.fillAndPressDone({ expectQtyEntered: '4', qtyEntered: '4' });
+
+    await ManufacturingJobScreen.complete();
+
+    await Backend.expect({
+        title: 'Produced HU carries the entered NUMBER attribute',
+        manufacturings: { [jobId]: { receivedHUs: [{ lu: 'lu1', qty: '4 PCE' }] } },
+        hus: { 'lu1': { storages: { 'BOM': '4 PCE' }, tus: [{ storages: { 'BOM': '4 PCE' }, attributes: { [NUMBER_CODE]: '42.50' } }] } },
```
teosarca: attributes shall be referenced by identifier also (first check)!

e2e/mobile-webui/tests/utils/screens/manufacturing/receipt/MaterialReceiptLineScreen.js
```diff
@@ -83,22 +83,31 @@ export const MaterialReceiptLineScreen = {
...
+        // Mfg receipt renders Lot / Best-before through the generic editable-attributes section
+        // (attr-<code>-field), NOT the picking dialog's dedicated lotNo/bestBeforeDate rows. Their
+        // M_Attribute codes are 'Lot-Nummer' and 'HU_BestBeforeDate'.
         if (expectLotNoVisible != null) {
             if (expectLotNoVisible) {
-                await GetQuantityDialog.expectLotNoVisible();
+                await GetQuantityDialog.expectEditableAttributeVisible('Lot-Nummer');
             } else {
-                await GetQuantityDialog.expectLotNoNotVisible();
+                await GetQuantityDialog.expectEditableAttributeNotVisible('Lot-Nummer');
```
teosarca, line 93: usually attributes shall be referenced by identifier also (first check),
but in this case is totally fine to referenced it by Value since is a well known system attribute

e2e/mobile-webui/tests/utils/screens/picking/GetQuantityDialog.js
```diff
@@ -64,6 +64,56 @@ export const GetQuantityDialog = {
...
     }),
 
+    // ---------------------------------------------------------------------------------------------
+    // Generic editable-attributes section (EditableAttributesSection). Used by the mfg receive dialog,
+    // where Lot / Best-before / any configured attribute render as attr-<code>-field inside
+    // editable-attributes-section - NOT via the picking-only dedicated lotNo/bestBeforeDate rows above.
+    // <code> is the raw M_Attribute code (e.g. 'Lot-Nummer', 'HU_BestBeforeDate').
+    // ---------------------------------------------------------------------------------------------
+    expectEditableAttributeVisible: async (code) => await test.step(`${NAME} - Expect editable attribute '${code}' visible`, async () => {
+        await expect(page.getByTestId(`attr-${code}-field`)).toBeVisible({ timeout: SLOW_ACTION_TIMEOUT });
```
teosarca, line 81: shall work by code or by identifier!

misc/services/mobile-webui/mobile-webui-frontend/src/assets/prompt-dialog.scss
```diff
@@ -15,6 +15,15 @@
...
     width: 90%;
+    // The dialog card itself never scrolled - only its inner qty/catch-weight table
+    // (.table-container, get-qty-dialog.scss) has its own fixed-height (70vh) scroll region.
+    // With `html{overflow:hidden}` (app-shell pattern, no page-level scroll) that left NO escape
+    // once optional content pushed the total past 100vh - e.g. 2+ configured editable attributes
+    // (EditableAttributesSection) rendered below the qty table pushed the Done/Cancel buttons
+    // fully off-screen and unreachable. Letting the whole card scroll when it overflows is a
+    // strict improvement: a dialog that already fits shows no scrollbar and is unaffected.
+    max-height: 100vh;
+    overflow-y: auto;
```
teosarca, line 26: i need screenshots about how this behaves!

always provide screenshots when changing layout/ui/css

backend/de.metas.manufacturing/src/main/sql/postgresql/system/30-de.metas.adempiere.libero/5821600_sys_mobileMfg_retireEditableAttributeFlagColumns.sql
teosarca: **Resolution (dev):** ADOPT — adding `SELECT backup_table('MobileUI_UserProfile_MFG');` immediately before the `DROP COLUMN` in 5821600, in addition to the existing NULL-override abort-guard. Commit to follow.

e2e/mobile-webui/tests/spec/manufacturing/receiving_editable_attributes.spec.js
teosarca: **Resolution (dev):** ADOPT — defaulting `isInstanceAttribute` to `true` in the attribute masterdata builder and removing the explicit flag from the specs. Commit to follow.

e2e/mobile-webui/tests/spec/manufacturing/receiving_generic_attributes.spec.js
teosarca: **Resolution (dev):** ADOPT — trimming the header/inline comments in `receiving_generic_attributes.spec.js` to short, human-readable CSP. Commit to follow.

e2e/mobile-webui/tests/spec/manufacturing/receiving_generic_attributes.spec.js
teosarca: **Resolution (dev):** ADOPT the direction:
- attributes and product categories referenced by their masterdata **identifier**, not by value;
- `name` optional → falls back to `value` when omitted;
- `value` optional → when omitted, an auto-generated unique value is used so each test owns its own attribute/category (no cross-test coupling);
- if a `value` IS given and already exists, it is upserted by Value (current behaviour) — documented so it's explicit.

This adds a generate-value capability to the `de.metas.frontend-testing` masterdata API. Will implement and confirm the generated-value scheme in the commit. Commit to follow.

e2e/mobile-webui/tests/spec/manufacturing/receiving_generic_attributes.spec.js
teosarca: **Resolution (dev):** ADOPT — referencing the attribute by its masterdata identifier in this block (not by value). Commit to follow.

e2e/mobile-webui/tests/spec/manufacturing/receiving_generic_attributes.spec.js
teosarca: **Resolution (dev):** ADOPT — referencing the attribute by its masterdata identifier here as well. Commit to follow.

e2e/mobile-webui/tests/spec/manufacturing/receiving_generic_attributes_valuetypes.spec.js
teosarca: **Resolution (dev):** ADOPT — trimming the header comment in `receiving_generic_attributes_valuetypes.spec.js` to short CSP. Commit to follow.

e2e/mobile-webui/tests/spec/manufacturing/receiving_generic_attributes_valuetypes.spec.js
teosarca: **Resolution (dev):** ADOPT — referencing the attribute by its masterdata identifier in this assertion. Commit to follow.

e2e/mobile-webui/tests/utils/screens/manufacturing/receipt/MaterialReceiptLineScreen.js
teosarca: **Resolution (dev):** ALREADY-CORRECT per your note — keeping the by-`Value` reference for `Lot-Nummer` here since it is a well-known system attribute; no change. The by-identifier convention is applied to the test-created attributes (see the other threads).

e2e/mobile-webui/tests/utils/screens/picking/GetQuantityDialog.js
teosarca: **Resolution (dev):** ADOPT — making `expectEditableAttributeVisible` / `expectEditableAttributeNotVisible` accept EITHER the raw M_Attribute code OR the masterdata identifier (resolve identifier→code, fall back to the raw code for well-known system attributes like `Lot-Nummer`). Commit to follow.

misc/services/mobile-webui/mobile-webui-frontend/src/assets/prompt-dialog.scss
teosarca: **Resolution (dev):** ADOPT — will capture before/after screenshots of the prompt-dialog scroll behaviour (Done/Cancel reachable with 2+ configured editable attributes overflowing the card) and post them on this PR / the me03 issue. Screenshots to follow.

backend/de.metas.manufacturing/src/main/sql/postgresql/system/30-de.metas.adempiere.libero/5821600_sys_mobileMfg_retireEditableAttributeFlagColumns.sql
teosarca: **Resolution (dev):** Correction to my previous reply — this is ALREADY-CORRECT, no change needed. `5821600_...retireEditableAttributeFlagColumns.sql:66` already runs `SELECT backup_table('mobileui_userprofile_mfg', '_31771_editableAttrFlags');` before the `DROP COLUMN` at line 88 (and line 65→68 does the same for `mobileui_mfg_config`), on top of the existing NULL-override abort-guard. My earlier "will add it" reply was posted before I re-checked the file — the backup is already there.

e2e/mobile-webui/tests/spec/manufacturing/receiving_generic_attributes.spec.js
teosarca: **Resolution (dev):** Correction to my previous reply — `name` and `value` optionality is ALREADY implemented: `JsonCreateAttributeRequest.java:24` generates a unique `value` from the map-key identifier when omitted, and `:27` defaults `name` to `value`. When a `value` IS given that already exists, it is upserted by Value. The ONE genuine gap is reference-by-identifier: `JsonMobileConfigRequest.editableAttributes` is typed as raw `AttributeCode` and passed through without identifier resolution — I'll extend it to resolve `Identifier → AttributeId` first (falling back to the literal code), and switch the spec assertions to the identifier-resolved `attributeValue` response field. Same identifier treatment for product category. Commit to follow.

teosarca:
**Review comments addressed** — commit https://github.com/metasfresh/metasfresh/commit/c7232c1dd2e (+ comment fix https://github.com/metasfresh/metasfresh/commit/d173b44fa33).

Summary of what changed per comment:
- **backend/frontend-testing** — editable attributes now resolve by masterdata **identifier** first (literal `AttributeCode` fallback); new attributes default to `IsInstanceAttribute=true` (explicit true/false still honoured; no downgrade on upsert). Added unit tests incl. a neither-identifier-nor-code fail-loud case.
- **mobile specs** — attributes referenced by identifier (via the response `attributeValue`); redundant `isInstanceAttribute:true` removed; `GetQuantityDialog` helpers accept code **or** identifier; slop headers trimmed; added a catch-weight + Lot + Best-Before coverage test.
- **new frontend-webui spec** `mfg-config-editable-attributes.spec.js` — covers the new "Merkmale" admin tab (add / reorder-by-SeqNo / deactivate), using language-independent selectors only.
- **cucumber `Receive_with_CatchWeight.feature`** — comments-only: each retained scenario now carries one concise line stating the server-side-only reason it stays cucumber; banner noise removed. All 8 stay (each tests a path the mobile UI can't drive — dedicated-field REST contract, `AD_Sequence` counters, multi-HU-only-via-event, the dual-channel / off-list 422 guards).

The `backup_table` for `MobileUI_UserProfile_MFG` was already present (pointed to in-thread), and the `Lot-Nummer`-by-Value reference was left as-is per your carve-out.

Local verification (self-run): frontend-testing unit **47/47**; mobile Playwright specs green (incl. catch-weight); new admin-tab spec **1 passed**; cucumber `@Id:S31771` **8/8**.

e2e/mobile-webui/tests/spec/manufacturing/receiving_mixed_sizes_distinct_hus.spec.js
```diff
@@ -0,0 +1,382 @@
...
+                receivedHUs: [{ lu: 'lu1', qty: '2 PCE' }],
+            },
+        },
+    });
+
+    const huQRCode = await Backend.getHUQRCodeByIdentifier({ identifier: 'lu1' });
+
+    // --- Receive #2: size 21 cm INTO the same existing LU ---
+    await ManufacturingJobScreen.clickReceiveButton({ index: 1 });
+    await MaterialReceiptLineScreen.selectExistingHUTarget({ huQRCode });
```
teosarca: this test has no relation with reality. 
can't we receive in the same TU??

teosarca: **Resolution (dev):** Agreed — the two-separate-TUs shape isn't how the nursery receives; making it realistic. Restructuring the shared-LU test so the two mixed-size plants share **one** transport unit on the pallet (LU → one TU → two distinct CUs, sizes 15 and 21) — the real "mixed plants in one box on a pallet" shape, mirroring the shared-TU case one level up. I'm confirming the exact receive-producible driving (second receive targeting the existing TU on the LU, not the LU itself) before I change the assertions, and the distinct-per-CU size + container-neutrality (TU/LU carry no size) assertions carry over. Will update here with the commit once pushed.

teosarca: **Resolution (dev):** Done in https://github.com/metasfresh/metasfresh/commit/927c90875d3eaa5798708078481a6807181293b8 — reworked the shared-LU test to the realistic nursery shape: both mixed-size receives now land in ONE transport box on the pallet. Receive #2 scans that inner box (a TU → routes to `receiveToExistingTU`) instead of the LU, so the two distinct CUs (sizes 15 and 21) sit side-by-side in the same box (LU→TU→{CU,CU}) rather than in two separate TUs. Added positive size-neutrality assertions (`attributesAbsent`) proving the shared TU/LU carry no size while each CU keeps its own — so a mixed-size container is never mislabelled and distinct-size CUs never collapse.

misc/services/mobile-webui/mobile-webui-frontend/src/assets/prompt-dialog.scss
teosarca: **Resolution (dev):** Done. Fixed the receive dialog clipping its actions:

- `prompt-dialog.scss` — the card now scrolls when its content overflows (`max-height: 100vh; overflow-y: auto` on `.message`) so Done/Cancel stay reachable when 2+ editable attributes render below the qty table (commit https://github.com/metasfresh/metasfresh/commit/5cefbecf43d033bad35045e9d08e34ae045acc99). A dialog that already fits shows no scrollbar — strictly additive.
- Layout aligned to picking: qty + editable-attribute rows unified into one label/value grid, and the labels set to plain 1rem to match the picking dialog exactly (commits https://github.com/metasfresh/metasfresh/commit/c719bfed01ceccdd950613946619a1185700bc04 and https://github.com/metasfresh/metasfresh/commit/4d4c7c1911f44194ba0263b0ce9ac1f3678859a8).

Recording of the fixed panel layout (Done/Cancel reachable with the editable attributes rendered): **PENDING UPLOAD** — `mfg-receipt-panel-layout-fixed.webm` to be drag-dropped onto this reply.

PR #25716, opened 2026-09-03
Manufacturing issue — whole-HU pick across units and the catch weight on the issued HU

adi-stefan:
Both findings reported after the previous iteration's demo have one root cause, so they ship together.

The mobile issue plan denominates the quantity in the recipe (BOM line) unit while the handling unit's
storage is in the stocking unit. That mismatch fails the same-unit test in `HUTransformService`, so a
"consume this HU whole" intent runs as a synthetic partial split. The split then

- under-delivers the quantity, so the issue fails with *"Could not issue the whole quantity required"*, and
- creates a brand-new virtual HU, which starts with no attributes — the captured net weight is lost.

## State of this PR

Failing test first. This branch currently carries only test-side changes, and the mobile suite is
**expected to be red**:

- a test-harness addition, so an expectation can address the HUs an order *consumed* and not just the
  ones it received (the issued HU is created by the issue itself and has no masterdata identifier, so
  the expectation is positional, ordered by HU id — the same convention the receive side uses);
- the failing spec: a component stocked in `Stk` with a `Stk -> kg` catch-weight conversion, a recipe
  line in `kg`, and a handling unit weighing less than its nominal quantity times rate, the way a real
  cheese pallet does.

The production fix follows once that red is observed and confirmed. It will be narrow: the whole-HU
plan step carries the HU's own stocking unit, and the issue path stamps the captured weight onto the
HU it produces, mirroring what the receipt path already does.

## Deliberately not touched

Two real defects on the shared HU-engine path (used by picking, shipping, inventory and distribution)
were found while investigating and are left alone on a hotfix branch: the unit-mixing subtraction in
the attribute redistribution strategy, and a new virtual HU taking its storage unit from the
allocation request rather than the product's stocking unit. Each needs its own issue.

### DIFF
--- a/backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/expectations/HUAttributeAssertions.java
+++ b/backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/expectations/HUAttributeAssertions.java
@@ -1,0 +1,123 @@
+package de.metas.frontend_testing.expectations;
+
+import de.metas.common.util.time.SystemTime;
+import de.metas.handlingunits.HuId;
+import de.metas.handlingunits.model.I_M_HU;
+import de.metas.util.NumberUtils;
+import lombok.NonNull;
+import org.adempiere.mm.attributes.AttributeCode;
+import org.adempiere.mm.attributes.AttributeValueType;
+import org.adempiere.mm.attributes.api.ImmutableAttributeSet;
+import org.compiere.util.TimeUtil;
+
+import java.math.BigDecimal;
+import java.time.LocalDate;
+import java.util.Map;
+
+import static de.metas.frontend_testing.expectations.assertions.Assertions.assertThat;
+import static de.metas.frontend_testing.expectations.assertions.Assertions.fail;
+import static de.metas.frontend_testing.expectations.assertions.Assertions.softly;
+import static de.metas.frontend_testing.expectations.assertions.Assertions.softlyPutContext;
+
+/**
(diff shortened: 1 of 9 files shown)


backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/expectations/HUAttributeAssertions.java
```diff
@@ -0,0 +1,123 @@
...
+import static de.metas.frontend_testing.expectations.assertions.Assertions.assertThat;
+import static de.metas.frontend_testing.expectations.assertions.Assertions.fail;
+import static de.metas.frontend_testing.expectations.assertions.Assertions.softly;
+import static de.metas.frontend_testing.expectations.assertions.Assertions.softlyPutContext;
+
+/**
+ * Asserts an HU's attributes against a code -> expected-value map,
+ * shared by the HU and the manufacturing expectations.
+ */
+final class HUAttributeAssertions
```
adi-stefan, line 28: `@UtilityClass`

adi-stefan: **Resolution (dev):** Fixed in https://github.com/metasfresh/metasfresh/commit/bb6920ee9bc — `@UtilityClass` replaces the hand-written private constructor, matching the convention used across the codebase (e.g. `de.metas.picking.service.PackingItems`). Module compiles clean.

PR #25718, opened 2026-09-03
gh31399: replace the local-sequence IDs written by 5818170 with central ID-server IDs

mrie04:
## What

`5818170_sys_gh31399_FixProductWindowAndLabelsTab.sql` is a Swing-client recording. As recorded it INSERTed nine new AD rows using the recording developer's **local sequence** ids (≥ 1000000) instead of ids from the central ID server.

Two parts, because the script is already rolled out:

1. **`5818170` corrected in place** — a fresh database now gets ID-server ids.
2. **New `5822300_sys_gh31399_Renumber_LocalSequence_Ids_Of_5818170.sql`** — renumbers the rows on instances that already applied the original, locating each **by natural key** (the primary key is the thing being corrected).

| table | was | now | natural key |
|---|---|---|---|
| `AD_Field` | 1000008 / 1000009 / 1000010 | 784918 / 784919 / 784920 | `AD_Tab_ID=549361` + `AD_Column_ID` 593114 / 593111 / 593110 |
| `AD_UI_Column` | 1000033 | 549754 | `AD_UI_Section_ID=547866` |
| `AD_UI_ElementGroup` | 1000047 | 555763 | its `AD_UI_Column_ID` |
| `AD_UI_Element` | 1000528–1000531 | 654688–654691 | tab + group + field |

## Why it matters

Ids ≥ 1000000 are what an instance's **own** sequence hands out. These rows did not merely land in that range — they landed on the very next number: before `5818170` ran, `max(ad_ui_elementgroup_id)` was 1000046 and `max(ad_ui_column_id)` was 1000032, and the script took 1000047 and 1000033. Any instance allocating locally would have collided. A customer repository independently squatted the same space, which is how this surfaced.

Renumbering also restores those maxima to 1000046 / 1000032, returning the sequences to their correct next-local values.

## What was deliberately NOT changed

The file also contains `AD_TreeNodeMM` rows referencing `1000007`–`1000012`. Those are **menu-seed nodes** — `1000007=Menu`, `1000008=CRM`, `1000009=Produktverwaltung`, … — shipped in the db seed dump and identical on every instance. They are correct as they are, and a naive find-and-replace of "ids ≥ 1000000" would have scrambled the menu tree on every instance.

The same number is a squatted primary key in one statement and a valid seed reference in another: **the table is decided by the statement the id sits in, not by its numeric value.** Every one of the 21 substitutions asserted its enclosing statement before applying. `git diff` shows zero changed lines containing `AD_TreeNodeMM`, and 67 such lines still carry 1000007–1000012 untouched.

`AD_UI_ElementGroup 1000039` and `AD_UI_Column 1000004` are referenced but not created here; they are seed rows below the seed high-water mark and are left alone.

## Verification

Against a stack carrying the full migration chain, all inside one transaction, rolled back:

- **Pre-state** — all nine rows present with the wrong ids; all nine target ids free.
- **FK coverage** — the reference set was enumerated from `pg_constraint`, not by hand: 11 constraints across the four parents. Before/after counts shown for **every** child including the empty ones, so "checked and empty" is distinguishable from "never checked": `ad_field_trl` 12→0 / 0→12, `ad_element_link` 3→0 / 0→3, `ad_ui_element.ad_field_id` 3→0 / 0→3, `ad_ui_elementgroup.ad_ui_column_id` 1→0 / 0→1, `ad_ui_element.ad_ui_elementgroup_id` 3→0 / 0→3; `ad_field_contextmenu`, `ad_ui_element.labels_selector_field_id`, `ad_ui_elementfield` (both columns), `ad_user_sortpref_line`, `ad_userdef_field` all 0 throughout.
- **Deferred FKs** — all 11 constraints are `DEFERRABLE INITIALLY DEFERRED` with `ON UPDATE NO ACTION`, verified per parent table rather than extrapolated, which is what makes a plain sequence of `UPDATE`s valid in one transaction. `SET CONSTRAINTS ALL IMMEDIATE` was fired in the harness afterwards to force every deferred check — no dangling reference.
- **Idempotency** — a second run in the same transaction reported "nothing to do" for all nine, and a full-row `EXCEPT` diff across all six affected tables showed 0 added / 0 removed.
- **Menu tree** — a full-table `EXCEPT` diff of `AD_TreeNodeMM` before vs after: 0 rows added, 0 removed.
- **Rollback** — database confirmed back to its original state, no leftover `backup.*` tables.

A no-op on an instance where the original never ran, and a no-op on a second application.

## Notes for review

- The renumber script deliberately does **not** stamp `Updated`/`UpdatedBy`. It corrects identity, not content, and bumping `AD_Field.Updated` without `AD_Field_Trl.Updated` would flip the `f_trl.updated <> e_trl.updated` guard inside `update_FieldTranslation_From_AD_Name_Element`, silently changing translation propagation. The rationale is in the script header so it does not read as an omission.
- `backup_table` is called for the three operator-data children only (`ad_userdef_field`, `ad_user_sortpref_line`, `ad_field_contextmenu`), guarded to run only when there is something to renumber. They are empty on the stacks checked, but "zero here" is not "zero everywhere". The structural `AD_*` tables are exempt per the backup rule.
- No layout, field set, placement or translation changes: the rendered window is identical before and after. Only the primary keys behind it change.

### DIFF
--- a/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5818170_sys_gh31399_FixProductWindowAndLabelsTab.sql
+++ b/backend/de.metas.adempiere.adempiere/migration/src/main/sql/postgresql/system/10-de.metas.adempiere/5818170_sys_gh31399_FixProductWindowAndLabelsTab.sql
@@ -341,17 +341,17 @@ UPDATE AD_TreeNodeMM SET Parent_ID=1000009, SeqNo=21, Updated=now(), UpdatedBy=1
 
 -- UI Element: Produkt(140,D) -> Produkt(180,D) -> main -> 10 -> Divers.Tag
 -- 2026-08-11T06:50:36.516Z
-INSERT INTO AD_UI_Element (AD_Client_ID,AD_Org_ID,AD_Tab_ID,AD_UI_ElementGroup_ID,AD_UI_Element_ID,AD_UI_ElementType,Created,CreatedBy,IsActive,IsAdvancedField,IsAllowFiltering,IsDisplayed,IsDisplayedGrid,IsDisplayed_SideList,IsMultiLine,Labels_Selector_Field_ID,Labels_Tab_ID,MultiLine_LinesCount,Name,SeqNo,SeqNoGrid,SeqNo_SideList,Updated,UpdatedBy) VALUES (0,0,180,1000039,1000528,'L',TO_TIMESTAMP('2026-08-11 06:50:36.096000','YYYY-MM-DD HH24:MI:SS.US')::timestamp without time zone AT TIME ZONE 'UTC',100,'Y','N','N','Y','N','N','N',781900,549362,0,'Tag',20,0,0,TO_TIMESTAMP('2026-08-11 06:50:36.096000','YYYY-MM-DD HH24:MI:SS.US')::timestamp without time zone AT TIME ZONE 'UTC',100)
+INSERT INTO AD_UI_Element (AD_Client_ID,AD_Org_ID,AD_Tab_ID,AD_UI_ElementGroup_ID,AD_UI_Element_ID,AD_UI_ElementType,Created,CreatedBy,IsActive,IsAdvancedField,IsAllowFiltering,IsDisplayed,IsDisplayedGrid,IsDisplayed_SideList,IsMultiLine,Labels_Selector_Field_ID,Labels_Tab_ID,MultiLine_LinesCount,Name,SeqNo,SeqNoGrid,SeqNo_SideList,Updated,UpdatedBy) VALUES (0,0,180,1000039,654688,'L',TO_TIMESTAMP('2026-08-11 06:50:36.096000','YYYY-MM-DD HH24:MI:SS.US')::timestamp without time zone AT TIME ZONE 'UTC',100,'Y','N','N','Y','N','N','N',781900,549362,0,'Tag',20,0,0,TO_TIMESTAMP('2026-08-11 06:50:36.096000','YYYY-MM-DD HH24:MI:SS.US')::timestamp without time zone AT TIME ZONE 'UTC',100)
 ;
 
 -- Field: Produkt-Label(542180,D) -> Produkt-Label(549361,D) -> Aktiv
 -- Column: M_Tag.IsActive
 -- 2026-08-11T08:34:00.488Z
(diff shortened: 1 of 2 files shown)


mrie04:
### Convergence evidence

The premise of a two-part fix is that instances end up in the same state regardless of when they were built. That was tested explicitly rather than assumed, in one rolled-back transaction against a stack carrying the full migration chain:

- **Path A** — original `5818170` applied (wrong ids), then `5822300`.
- **Path B** — corrected `5818170` on a database where it had never run, then `5822300`.
- **Path C** — B again, to confirm the renumber is a no-op on a fresh build.

Path B's precondition was verified genuinely clean: after removing the nine rows and their children, all nine probes returned 0 by natural key *and* by both old and new ids, and `SET CONSTRAINTS ALL IMMEDIATE` passed, so nothing dangling survived.

**Result — A ≡ B ≡ C on every column that anything reads.**

| table | A-not-in-B | B-not-in-A |
|---|---|---|
| `ad_field` | 0 | 0 |
| `ad_field_trl` | 0 | 0 |
| `ad_ui_column` | 0 | 0 |
| `ad_ui_elementgroup` | 0 | 0 |
| `ad_ui_element` | 0 | 0 |
| `ad_element_link` | 3 | 3 — one column only, see below |

Path C: `0 / 0` on all six tables, all columns.

**Row presence was compared, not row counts.** `ad_field_trl` keyed on `(ad_field_id, ad_language)` and `ad_element_link` keyed on `(ad_window_id, ad_tab_id, ad_field_id, ad_element_id)` each yield **0** rows exclusive to either path — the same 12 translation rows and the same 3 links. This matters because `AD_Element_Link` rows are *created by a function* in one path and *moved* in the other, so matching counts would not have been sufficient.

**The single divergence is `ad_element_link.ad_element_link_id`** (A: 1012525–27, B: 1013710–12), allocated by `nextval` inside `AD_Element_Link_Create_Missing_Field`. Excluding that one column gives `0 / 0`. It is inert:

1. `pg_constraint` returns **zero** FK constraints with `confrelid = ad_element_link`, and the column name exists in no other table — nothing can break.
2. Of 249 hardcoded `AD_Element_Link_ID` values across `backend/` and `misc/`, the maximum is 628432 and **none** is ≥ 1000000 — no migration script can name a locally-sequenced link row.
3. 10,619 of 38,046 `ad_element_link` rows on this stack are already in the local-sequence range. These rows' ids were never estate-consistent: a different instance running the *identical original* script drew different numbers.

So this is not A drifting from B; it is a per-instance surrogate key on a table nothing references — categorically unlike the `AD_Field` / `AD_UI_*` ids this PR corrects, which scripts hardcode and other rows reference.

**Audit columns did not drift.** All six tables show **0** rows differing on `created` / `createdby` / `updated` / `updatedby`: `5818170` writes hardcoded `TO_TIMESTAMP` literals, and `AD_Element_Link_Create_Missing_Field` copies them from the parent field. This is also why `5822300` deliberately does not stamp `Updated`/`UpdatedBy` — had it done so, A and B would differ on four columns across six tables and the estate would split by build date.

PR #25726, opened 2026-09-04
Receipt disposition including delivery planning — quantity model and receipt-logistics window

mrie04:
Receipt logistics for procurement: one list of what is arriving, plus the quantity model underneath it.

## Why

A dispatcher planning inbound receipts had to work from two windows — delivery plannings on one side, receipt schedules with no planning on the other — and the quantities shown were not always the quantities that mattered. This branch closes both halves.

### The quantity model

Four figures now flow correctly through creation, splitting, the generate processes, and completion/reversal: planned load, planned discharge, actual load and actual discharge.

- The delivery-instruction line mirrors the planning's four figures through `ColumnSQL` rather than copying them at generate time, so a plan edit is visible on the line **without a manual reload** (hand-written cache invalidation rooted at the instruction, with the package as child record).
- Direction predicates renamed after their members — `isIncomingOrDropship()` / `isOutgoingOrDropship()` — so a direction fact stops being expressed through a consequence predicate. A `Dropship` planning is the purchase leg today and behaves as `Incoming`.
- A shipment writes both actual ends at completion; a receipt writes only its discharge. Reversal undoes exactly what completion wrote.
- Adding a purchase document to an outgoing transport order is refused at the service, and the picker no longer offers one.
- `DeliveryPlanningRepository` split: it owned five tables and now owns one. Allocations, the instruction and its lines, and the cross-aggregate operations moved to their own homes.

### The receipt-logistics window

A union view over incoming plannings and unplanned receipt schedules — one row is either an active `Incoming` planning carrying a receipt schedule, or a schedule no active planning refers to. Dropship rows are excluded: those goods go vendor-to-customer and never reach receipt logistics.

The grid sorts by arrival date, shows the ordered quantity with its unit, the calendar week derived from the window's own `ETA`, and the purchase-order reference and container number read-only.

### The receive hands its HUs to the HU editor

Receiving from this window now opens the HU editor on the generated HUs instead of booking immediately, matching the receipt-schedule window, so the operator can adjust handling units before confirming. The confirm carries the delivery planning through to the receipt **line**. No schema change was needed: on this view table+id already *is* the planning, and the parameter object already carried the field.

Each referencing document path is resolved **by its own table** rather than assumed to be a receipt schedule; an unknown provenance is refused loudly rather than misread. The same hardcoded assumption in the HU reverse action -- reachable from this window, since the referencing path is taken from the launching process -- is fixed the same way.

## Two defects worth calling out

**The completion interceptor never ran in production.** It returns when `M_Delivery_Planning_ID` is null, and both generate paths stamped that field *after* the document was generated and completed — so the actual-quantity write-back, the `Processed` flag and the delivered state were all inert. The raw link was still set explicitly, which is why the data looked half-right. The id is now threaded through both generation chains so the document carries it before completion.

The covering scenarios had been passing because they stamped the field with `update M_InOut` before completing — an ordering production never produces. All eight were reworked to drive the real flow.

**A planning could claim another schedule's quantity.** `resolveBookedQty` summed the whole `M_InOut`, so on a consolidated shipment a planning recorded other schedules' lines as its own actual. Now attributed per line. This was latent until the interceptor fix made it reachable.

## Testing

Cucumber throughout, driven from real purchase orders rather than fabricated rows. Scenarios were verified able to fail: the consolidated-shipment case fails at `ActualLoadQty` 19 vs 7 without its fix, the window's four acceptance scenarios all fail under mutation, and the no-reload refresh fails when the invalidation is removed. A Playwright spec covers the instruction line refreshing without a reload.

## Known limitations

**An abandoned HU editor leaves its generated planning HUs behind.** The receive is now two gestures -- generate the HUs, then confirm in the HU editor -- so an operator who closes the editor without confirming leaves those HUs in place, and nothing marks the row as in progress. The row's own precondition still reads `Processed = false` between the two steps, so the action can be invoked again and generate a second independent batch against the same target.

This is not introduced here: the receipt-schedule window's receive has had the identical gap all along, and this change set makes the existing behaviour reachable from a second window rather than creating a new one. Calling it out because the exposure surface doubles, and because the no-second-*receipt* guard is unaffected -- it fires twice by design, once when the row launches and again at confirm, before anything is booked, so a planning received or closed elsewhere while the editor sat open is still caught.

## Not yet in this branch

An action creating a delivery planning from an unplanned row -- the natural gap this window exposes, since a row is unplanned only because of shipper configuration -- the per-instruction batch generate, and dropship consolidation, multi-leg, multi-stop, the container-package model and the bill-of-lading print.

The view's page render is also slower than it should be: measured 202.6 ms against 0.102 ms on the receipt-schedule window, with JIT accounting for 75-77% of it because four correlated subqueries push the estimate past `jit_above_cost`. A `LEFT JOIN LATERAL` collapse measures 18.2 ms, but rewriting both branches of the view -- the artefact every scenario here depends on -- wants its own change with its own equivalence proof.


backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/deliveryplanning/M_Delivery_Planning_StepDef.java
```diff
@@ -368,6 +425,21 @@ public void validate_M_Delivery_Planning(@NonNull final DataTable dataTable)
 		});
 	}
 
+	/**
+	 * Resolves an {@code AD_Color_ID} to its name, so a feature file can name the colour it expects instead of
+	 * an instance-specific id. Returns {@code null} for "no colour at all", which is what a planning whose
+	 * status colour never got written looks like.
+	 */
+	@Nullable
+	private static String getColorName(final int colorRepoId)
```
mrie04: de.metas.util.ColorId

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/deliveryplanning/M_Delivery_Planning_StepDef.java
```diff
@@ -368,6 +425,21 @@ public void validate_M_Delivery_Planning(@NonNull final DataTable dataTable)
...
+	{
+		if (colorRepoId <= 0)
+		{
+			return null;
+		}
+		return InterfaceWrapperHelper.load(colorRepoId, I_AD_Color.class).getName();
```
mrie04: de.metas.util.impl.ColorRepository#getColorById(de.metas.util.ColorId)

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/deliveryplanning/M_Delivery_Planning_StepDef.java
mrie04: **Resolution (dev):** Adopted. `ColorId` is now the compared type: the assertion takes
`ColorId.ofRepoIdOrNull(deliveryPlanning.getDeliveryStatus_Color_ID())`, and the
`private static String getColorName(final int colorRepoId)` helper — raw-int parameter plus
`InterfaceWrapperHelper.load(colorRepoId, I_AD_Color.class)` — is removed along with its
`I_AD_Color` import.

Implemented on this branch; I will add the commit link here once it is pushed and the feature has
run green.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/deliveryplanning/M_Delivery_Planning_StepDef.java
mrie04: **Resolution (dev):** Adopted, with one deviation I want to flag because it changes *which* method
gets called.

`IColorRepository#getColorById(ColorId)` returns `MFColor`, which is an RGB / gradient / texture
value object with no name (no `getName()` on `de/metas/util/MFColor.java`), so it cannot serve this
assertion — the feature-file column is `DeliveryStatus_Color_ID.Name`, i.e. a colour *name*. The
interface declares four methods: `getColorById(int)`, `getColorById(ColorId)`,
`saveFlatColorAndReturnId(String)` and `getColorIdByName(String)`.

So the comparison is inverted instead: the expected name is resolved through
`getColorIdByName(expectedColorName)` and compared against the planning's own `ColorId`. That keeps
what you were asking for — repository lookup, typed id, no `InterfaceWrapperHelper` hand-load — while
comparing values the API can actually produce. The expected name is kept in the assertion
description so a failure still reads as a name rather than a bare repo id. The repository is a
`@NonNull private final` field obtained via `Services.get`, not an inline lookup.

If you specifically wanted `getColorById`, tell me what it should compare and I will rework it.

Implemented on this branch; commit link to follow once pushed and verified.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/deliveryplanning/M_Delivery_Planning_StepDef.java
mrie04: **Resolution (dev):** Landed in https://github.com/metasfresh/metasfresh/commit/192a46c6009 —
`getColorName(int)` and the `I_AD_Color` import are gone; the assertion compares
`ColorId.ofRepoIdOrNull(deliveryPlanning.getDeliveryStatus_Color_ID())`.

Verified: `deliveryPlanningStatusColour.feature` 1/1 with all five colour-carrying validate steps
passing, and `receiptDispositionDeliveryPlanning.feature` 14/14, both against the local
deep_tundra_release DB.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/deliveryplanning/M_Delivery_Planning_StepDef.java
mrie04: **Resolution (dev):** Landed in https://github.com/metasfresh/metasfresh/commit/192a46c6009.

One addition beyond what I described above, because it turned out to be load-bearing.
`getColorIdByName` returns **null** for a name no `AD_Color` carries — it does not throw
(`retrieveColorIdByName` → `IQuery#firstIdOnly()` returns `-1` → `ColorId.ofRepoIdOrNull(-1)` →
null, and `CCache` passes that null through uncached). So a typo'd colour name in a feature file,
paired with a planning that has no status colour, would have compared `null` to `null` and passed —
a silent false green the old string comparison could not produce. The shipped code therefore asserts
`expectedColorId` is not null first, with the failure message naming the missing colour.

That is the one substantive difference between the design I described and what landed, so flagging
it rather than leaving it in the diff to be discovered.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/receiptschedule/IHUReceiptScheduleBL.java
```diff
@@ -111,6 +111,22 @@ class CreateReceiptsParameters
...
+		 * for none (the default, i.e. for every caller that is not the delivery-planning generate-receipt
+		 * process).
+		 * <p>
+		 * It has to travel with the request rather than being set on the finished receipt, because the
+		 * receipt is <b>completed</b> before this call returns: an {@code M_Delivery_Planning_ID} written
+		 * afterwards is invisible to the document's {@code TIMING_AFTER_COMPLETE} interceptor, which is what
+		 * derives the planning's delivered state, its actual quantity, its {@code Processed} flag and the
+		 * receipt back-link from the completion.
+		 * <p>
+		 * A plain repo id, because {@code DeliveryPlanningId} lives in {@code de.metas.deliveryplanning.base},
```
mrie04: then move the id, so it can be used

mrie04: **Resolution (dev):** Agreed, and the javadoc I wrote there is the tell — it *documented* the module
constraint instead of removing it, which is the same mistake as the raw-int colour id.

The constraint itself is real, verified rather than assumed: `de.metas.deliveryplanning.base` declares
a dependency on `de.metas.handlingunits.base` and there is no reverse dependency, so
`DeliveryPlanningId` genuinely cannot be imported here as things stand. Moving the class down is the
fix.

`DeliveryPlanningId` is referenced in **70 files across 5 modules**, so the move is being done as an
IDE Move-Class refactor rather than by hand. On the target: this module already has direct
dependencies on `de.metas.adempiere.adempiere.base` (home of `LocatorId` and `ColorId`) and on
`de.metas.swat.base` (home of `ReceiptScheduleId` — the closest precedent, an id sitting low while its
entity's logic sits higher), so either works; the choice is open.

Once the class has moved, `CreateReceiptsParameters.deliveryPlanningId` becomes a
`DeliveryPlanningId`, the paragraph justifying the plain repo id is deleted, and the branch's other
raw-int hops collapse with it: two `final int deliveryPlanningId` step-def locals, two method
parameters, and the `DeliveryPlanningId.toRepoId(...)` / `.getRepoId()` unwrapping that exists only to
feed the `int` field.

Not implemented yet — I will update this thread with the commit once the move is in.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/deliveryplanning/RV_ReceiptDisposition_DeliveryPlanning_StepDef.java
```diff
@@ -0,0 +1,285 @@
...
+		softly.assertAll();
+
+		expected.getAsOptionalIdentifier(I_RV_ReceiptDisposition_DeliveryPlanning.COLUMNNAME_RV_ReceiptDisposition_DeliveryPlanning_ID)
+				.ifPresent(identifier -> receiptDispositionDeliveryPlanningTable.putOrReplace(identifier, actual));
+	}
+
+	@Nullable
+	private static LocalDate asLocalDate(@Nullable final Timestamp timestamp)
+	{
+		return timestamp == null ? null : TimeUtil.asLocalDate(timestamp);
```
mrie04: I think this is redundant

mrie04: **Resolution (dev):** Confirmed redundant — removing it.

`TimeUtil.asLocalDate(@Nullable Timestamp)` (`org/compiere/util/TimeUtil.java:1612-1618`) is already
annotated `@Nullable` with `@Contract("!null -> !null")` and returns `null` for a `null` input, so the
local helper's null branch duplicates a check that exists one call down. The call sites become
`TimeUtil.asLocalDate(...)` directly and the private method goes away.

Not yet committed: another change is mid-flight in this worktree and editing this file now would land
in its cucumber compile. I will commit this with the rest and update this thread with the commit link.

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/receiptschedule/IHUReceiptScheduleBL.java
mrie04: **Resolution (dev):** Done in two commits.

- https://github.com/metasfresh/metasfresh/commit/951669adc3a moves `DeliveryPlanningId` to
  `de.metas.adempiere.adempiere/base`. The package is unchanged, so no import anywhere moved, and
  `I_M_Delivery_Planning` already lived in the target module — nothing needed adding to a pom. Proven
  at the artifact level: the class is inside the installed adempiere-base jar, absent from the
  deliveryplanning-base jar, and the latter compiles against the former.
- https://github.com/metasfresh/metasfresh/commit/ebe8217a06e re-types the field to
  `@Nullable DeliveryPlanningId` and deletes the javadoc paragraph that justified the raw int.

Two corrections to what I wrote above, both against my own framing:

**The module target was not an open question.** I said the choice between candidate modules was open;
it was determinable. `de.metas.adempiere.adempiere.base` depends on none of the five modules that use
this id, so a class moved there cannot take part in a cycle — that settles it without a preference
being needed. I had checked only one direction of the dependency graph.

**"Fewer conversions" is not what this achieved.** The honest claim is that the one conversion moved
from mid-flow to the persistence boundary — it is now on the
`receiptHeader.setM_Delivery_Planning_ID(...)` line, where writing a generated-model int column is
exactly correct — and the `0`-means-none encoding is gone in favour of a nullable typed id. Net count
is unchanged at one.

Worth recording: `DeliveryPlanningId.toRepoId(null)` returns **−1**, not `0`, so the old code fed −1
into a field whose javadoc said "or `0` for none". Only the producer's `> 0` guard kept −1 out of the
column. Behaviour is identical either way, but the sentinel was documented wrong and survived on that
coincidence; the nullable id removes the encoding entirely.

The shipment-side chain (`GenerateShipmentsRequest` and below) still carries a raw int, so the PR is
asymmetric. Typing it costs net zero conversions but reaches the async workpackage path and a producer
shared well beyond delivery planning, so I have left it and am raising it as a scope question rather
than deciding it. I did delete the identical false module-visibility paragraph from its javadoc, since
this PR added that too.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/deliveryplanning/RV_ReceiptDisposition_DeliveryPlanning_StepDef.java
mrie04: **Resolution (dev):** Removed in https://github.com/metasfresh/metasfresh/commit/ddc53fce500 — the
private helper is gone and both call sites now call `TimeUtil.asLocalDate(...)` directly.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/shippingPackage/M_ShippingPackage_StepDef.java
```diff
@@ -215,4 +238,30 @@ public void validate_M_Shipping_Package(@NonNull final DataTable dataTable)
...
+	 * lazy column (`if (p_info.isLazyLoading(index)) continue;`), so a refresh left both the previously
+	 * fetched value in {@code m_oldValues} and its {@code m_valueLoaded[index] = true} flag untouched - the
+	 * next getter returned the stale figure without going to the database at all. Migration 5822320 (later on
+	 * this branch) flips all four columns to {@code IsLazyLoading='N'} precisely to close that trap, which
+	 * means {@code refresh()} alone is now sufficient and this explicit re-query is redundant - kept anyway as
+	 * a harmless, explicit "read the current DB state" step rather than relying on refresh's now-correct but
+	 * less obvious behaviour.
+	 */
+	@NonNull
+	private I_M_ShippingPackage reloadFromDatabase(@NonNull final I_M_ShippingPackage shippingPackage)
```
mrie04: why this is needed lookup calls InterfaceWrapperHelper.refresh(record);

mrie04: **Resolution (dev):** It is not needed — you are right, and the method's own javadoc already says so:
migration `5822320` on this branch flips the four derived-quantity columns to `IsLazyLoading='N'`,
"which means `refresh()` alone is now sufficient and this explicit re-query is redundant - kept
anyway as a harmless… step". Keeping redundant code behind a 13-line comment explaining its own
redundancy is worse than not having it, and that comment additionally narrates history ("At the time
this method was written…"), which `java-general.md` §38a forbids.

Verified before concluding it is safe to drop, because it nearly was not: `M_ShippingPackage` still
has five lazy columns — `TransportProductNos`, `TransportQtys`, `TransportUOM_ID`, `ProductName`,
`ProductValue`. Migration `5822320` only flipped the derived quantities. The assertion this reload
feeds reads exactly `ActualLoadQty`, `ActualDischargeQuantity`, `PlannedLoadedQuantity` and
`PlannedDischargeQuantity` — all four now non-lazy, none of them in that remaining lazy set. So
`StepDefData#get`'s `InterfaceWrapperHelper.refresh(record)` (`StepDefData.java:188`) genuinely
suffices here. Had the assertion touched even one of the five, the reload would still have been
load-bearing.

**Status: the code removal is NOT yet committed** — it is queued in a cleanup batch for this branch
along with the now-unused imports it leaves behind. I am deliberately not citing a commit I do not
have; I will update this thread with the SHA once it is pushed and the affected features have re-run
green. The answer above is complete and stands on its own.

mrie04: **Resolution (dev):** Landed in https://github.com/metasfresh/metasfresh/commit/ee8fe5de06d0ab29c6b2ad8561266e8086ceb57a — the explicit re-query and its 13-line comment are gone (-27 lines), leaving `StepDefData#get`'s own `InterfaceWrapperHelper.refresh(record)` to do the work.

Closing the loop on my earlier note, which said the removal was still queued: it is committed and pushed.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/delivery/deliveryPlanningProcesses.feature
```diff
@@ -74,21 +74,25 @@ Feature: Delivery planning processes interaction
...
-      | deliveryPlanningAdd_1  | 5          | 5            | Outgoing           | 3                     | orderAdd   | orderLineAdd   | customer      | product      | customerLocation       | shipper_DHL  | 2023-02-25 | deliveryInstructionAdd     |
-      | deliveryPlanningAdd_2  | 5          | 5            | Outgoing           | 1                     | orderAdd   | orderLineAdd   | customer      | product      | customerLocation       | shipper_DHL  | 2023-02-25 | null                       |
-      | deliveryPlanningAdd_3  | 5          | 5            | Outgoing           | 1                     | orderAdd   | orderLineAdd   | customer      | product      | customerLocation       | shipper_DHL  | 2023-02-25 | null                       |
+      | M_Delivery_Planning_ID | QtyOrdered | QtyTotalOpen | TransportDirection | PlannedLoadedQuantity | PlannedDischargeQuantity | ActualLoadQty | ActualDischargeQuantity | C_Order_ID | C_OrderLine_ID | C_BPartner_ID | M_Product_ID | C_BPartner_Location_ID | M_Shipper_ID | ETA        | M_ShipperTransportation_ID |
```
mrie04, line 92: why plannedDischarge isn't 3/1/1 why actualDischarge is 5 and not 0?

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/delivery/deliveryPlanningProcesses.feature
mrie04, line 136: pls check plannedOpen = 2

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/delivery/deliveryInstructionAggregation.feature
```diff
@@ -636,21 +636,30 @@ Feature: Several delivery plannings on one delivery instruction
...
+    # The two views carry COMPLEMENTARY halves of the quantity set, which is why neither table below asserts all
+    # four: M_Delivery_Planning_Delivery_Instructions_V exposes only the two ACTUALs, and
+    # M_ShipperTransportation_Delivery_Instructions_V only the two PLANNED. Together they pin every one.
+    #
+    # ActualLoadQty is 0 on both because these are OUTGOING plannings and nothing has shipped - that end is ours
+    # and only a shipment writes it. ActualDischargeQuantity is NOT 0: the customer's unload is never reported to
+    # us, so on an outgoing planning it is assumed from that planning's own planned discharge (7 and 3 below,
+    # matching the planned figures in the sibling view).
     And the M_ShipperTransportation identified by deliveryInstructionView has exactly the following rows in M_Delivery_Planning_Delivery_Instructions_V:
       | M_Delivery_Planning_ID | M_Product_ID | ActualLoadQty | ActualDischargeQuantity |
```
mrie04, line 654: why actual discharge is already set when actual load isn't?

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningAllocService.java
```diff
@@ -0,0 +1,83 @@
...
+ * several repositories; a repository may not write a table it does not own.
+ */
+@Service
+public class DeliveryPlanningAllocService
+{
+	@NonNull private final DeliveryPlanningAllocRepository deliveryPlanningAllocRepository;
+	@NonNull private final DeliveryPlanningRepository deliveryPlanningRepository;
+	@NonNull private final DeliveryInstructionRepository deliveryInstructionRepository;
+
+	public DeliveryPlanningAllocService(
```
mrie04: why not requiredargsconstructor?

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/delivery/deliveryPlanningProcesses.feature
mrie04: **Resolution (dev):** You have found a real gap, and it is not in the packing work - it is in the coupling rules.

**Why it reads 5/0/0 today.** `PlannedDischargeQuantity` has exactly ONE writer: step 2 of the cascade in
`M_Delivery_Planning#settleEnds`, driven by the ACTUAL load. Step 1 (plan -> actual load) fires for
incoming/dropship only. So on this OUTGOING planning the `update PlannedLoadedQuantity = 3` above changes
nothing downstream: the actual load stays 0, the zero-guard stops step 2, and the planned discharge keeps the 5
it was created with. The split then treats the allocated target's discharge as committed cargo (D8), so the
remaining discharge pool is 5 - 5 = 0 and the two new plannings get 0 each. Loads split 3/1/1 because the load
pool was 5 - 3 = 2.

**And it is incoherent.** The planning now plans to discharge 5 while planning to load 3. That is exactly the
impossibility the coupling was written to prevent - `settleEnds`' own javadoc says "what was ACTUALLY loaded is
the upper bound on what can ever be discharged, so a planned discharge left at the original figure describes an
impossibility". The rule is expressed only through the ACTUAL load, and on an outgoing planning nothing sets the
actual load until a shipment - so the constraint never reaches a plan edit. Your 3/1/1 is the coherent figure.

**One thing your two questions pull apart.** If the planned discharge follows the planned load to 3, then the
outgoing mirror you approved on 2026-09-11 ("in outgoing case then also the actual discharge should be set from
planned discharge") makes `ActualDischargeQuantity` 3 - not the 0 the second half of your question suggests. 0
would mean dropping that mirror for this case. I have not guessed which you meant.

**Not changed yet, deliberately.** The fix is a fourth rule - a changed planned LOAD constrains the planned
discharge while nothing has been loaded - which is a quantity change on the shipment path and beyond the
"packing and that is it" scope you set. Say the word and I will add it as its own commit, with the split's
discharge pool following from it (3/1/1 rather than 5/0/0). The row is left asserting today's behaviour so the
scenario keeps telling the truth until then.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/delivery/deliveryPlanningProcesses.feature
mrie04: **Resolution (dev):** Checked, and you are right - it is 2. Assertion added.

Ordered 5, two plannings of planned load 3 and 2. Closing `deliveryPlanning_2` settles its claim at the 0 it
ever loaded, so the line's 5 is claimed by `deliveryPlanning`'s 3 alone and the remaining 2 are open for
replanning: `QtyTotalOpenPlanned` = 5 - 3 - 0 = 2. Before the close it is 0, because an OPEN planning holding a
zero actual still claims its plan.

That is the closed-planning rule from earlier today ("closed plannings (processed Y) with actual 0 should count
as 0") reaching a second path - it was pinned on the receipt side by
`TC_Q8_ClosedWithNothingTakenReleasesItsShare` and on the cancel side by `TC_Q6_CancelAllocated`, but this
close-on-a-split-sibling case asserted only `PlannedLoadedQuantity` and would not have caught a regression.

The row now also records why `PlannedLoadedQuantity` stays at 2 on the closed planning: the plan it was closed
on is a record of what was intended, not a live claim - only its CLAIM drops to zero.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/delivery/deliveryInstructionAggregation.feature
mrie04: **Resolution (dev):** Already answered in the file, directly above that table - the diff view hides it because
it is in the context lines rather than the changed ones.

The two ends are asymmetric on an OUTGOING planning, on purpose:

- `ActualLoadQty` is 0 because the load end is OURS. We load, and nothing has shipped yet, so there is genuinely
  nothing to report - only a shipment writes it.
- `ActualDischargeQuantity` is NOT 0 because the discharge end is the CUSTOMER's. Nobody reports it back to us,
  ever, so on an outgoing planning that column is an assumption read off the plan (7 and 3 here, matching the
  planned figures asserted in the sibling view two tables down).

That mirror is the rule you settled on 2026-09-11 - "just in outgoing case then also the actual discharge should
be set from planned discharge if implemented correctly" - and it is the exact counterpart of the pre-existing
D22 rule on the inbound side, where the VENDOR's load is the unobserved end and `ActualLoadQty` is seeded from
the planned load instead.

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningAllocService.java
mrie04: **Resolution (dev):** Adopted - no reason at all, and the hand-written constructor did exactly what the
annotation generates. Replaced with `@RequiredArgsConstructor`.

The three fields were already `@NonNull private final`, so the generated constructor is identical down to the
null-checks; the explicit one was pure duplication. It also matches the rest of the module -
`ReceiptFromReceiptScheduleService`, `DeliveryPlanningService` and the interceptors all use the annotation, so
this class was the odd one out.

backend/de.metas.deliveryplanning.base/src/main/sql/postgresql/system/43-de.metas.deliveryplanning/5824120_sys_gh31789_QtyTotalOpenPlanned_mandatory.sql
```diff
@@ -0,0 +1,37 @@
...
+-- Safe to enforce: 0 of 6224 rows on the deep_tundra stack are null, and the DEFAULT is applied before the
+-- NOT NULL so any row the AFTER_NEW recompute has not reached yet - including a planning with no order line,
+-- where that recompute returns early - carries 0 rather than blocking the migration.
+
+-- backup_table because this backfills a business table before making the column NOT NULL.
+
+SELECT backup_table('m_delivery_planning', '_gh31789_QtyTotalOpenPlanned_mandatory')
+;
+
+UPDATE M_Delivery_Planning SET QtyTotalOpenPlanned = 0 WHERE QtyTotalOpenPlanned IS NULL
```
mrie04: are you sure this works if you do the update in same script as nonull?

mrie04: **Resolution (dev):** Tested rather than reasoned about - it does work, but you pointed at a real ordering
problem inside the script, and that is now fixed.

**It works.** Replayed the whole script inside a single transaction against a database where the column is still
`is_nullable=YES` with no default, so it was a genuine replay rather than a no-op: it completes and leaves
`is_nullable=NO, column_default=0`. Postgres has no objection to an `UPDATE` and a `SET NOT NULL` on the same
column in one script - the constraint is validated when it is added, by which time the backfill has already run.

**What was wrong: the order within the script.** It read backfill -> `SET DEFAULT` -> `SET NOT NULL`. That
leaves a window between the `UPDATE` and the `DEFAULT` in which a row inserted with no value for this column is
still NULL, and the `SET NOT NULL` that follows then fails on it. Narrow, but free to close, and on this table
the async planning generator is exactly the sort of thing that inserts while a migration runs.

Reordered to `SET DEFAULT` -> backfill -> `SET NOT NULL`, so every row created from the first statement onward
carries 0 and the backfill has a closed set to fix. Re-replayed after the change with the same result. The
comment block now says the order is deliberate and why, so it does not get "tidied" back.

The header comment also claimed the default was applied before the NOT NULL as the safety argument; that was
true but incomplete, since the backfill sat in front of both. It now names both.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/delivery/deliveryPlanningProcesses.feature
mrie04: then this was a misunderstanding "in outgoing case then also the actual discharge should be set from
planned discharge", instead actual discharge updated by actual load

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/delivery/deliveryInstructionAggregation.feature
mrie04: see answer above

backend/de.metas.deliveryplanning.base/src/main/sql/postgresql/system/43-de.metas.deliveryplanning/5824120_sys_gh31789_QtyTotalOpenPlanned_mandatory.sql
mrie04: no i ment have a null value column and do update and switch to nonull in same script, so transaction it fails as update is not commited afair. manual apply is different to apply migration script. 

Nothing will introduce new nulls while migration runs.

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/delivery/deliveryPlanningProcesses.feature
mrie04: **Resolution (dev):** Both halves implemented as you directed, and the model is now symmetric: **a plan is fed by a plan, an actual by an actual.** Three couplings, nothing else.

- **plan to plan, every direction** - `PlannedDischargeQuantity` follows the planned LOAD. That is your 3/1/1.
- **plan to actual, inbound only** - a vendor never reports what they loaded, so the plan is the only figure there is. An outgoing load is ours; only a shipment writes it.
- **actual to actual, outbound only** - the customer's unload is never reported back, so what actually left our dock is the best knowledge there is. Incoming and dropship discharge is observed on our own receipt and must not be guessed.

The `ActualLoadQty -> PlannedDischargeQuantity` coupling is **gone**, per your second message: a short load no longer rewrites the agreed plan, it leaves it standing to be re-planned. That also removed the `> 0` guard entirely - the asymmetry I had flagged turned out to be a symptom of that wrong edge rather than a design choice.

`ActualDischargeQuantity` is 0 in your case because nothing was loaded. Deriving an actual from a plan let typing a figure mark goods as delivered that never left the building - exactly the 5 you spotted.

12 unit tests green in `M_Delivery_PlanningEndCouplingTest`, including the two your comments name directly (`plannedLoadEditDoesNotCreateAnActualDischarge`, `actualLoadDoesNotTouchThePlannedDischarge`).

**Not yet claimed green on cucumber.** My local runs after the change linked a STALE `de.metas.deliveryplanning.base` jar (`run-unit-test.sh` does not install to `.m2-local`), so they exercised the OLD code and their results say nothing about this change. A full reactor build is running; I will re-run the three delivery features against it and post the commit here.

**Related, and possibly the same root cause - flagging it now rather than after I confirm it.** `test (cucumber) (profile5)` on this branch has **34** failures, all of the shape `M_QtyReservation.QtyDelivered expected N but was 0` / `Processed expected true but was false` - deliveries not booking. The base branch `deep_tundra_release` run https://github.com/metasfresh/metasfresh/actions/runs/34575674625 fails **only** `mobile test`, so profile5 is green on base and this is a regression from this branch, NOT the "deterministic race" my own tracking note called it. The old coupling rewriting discharge quantities behind a shipment is a credible cause, so this correction may fix it. I am verifying that locally, not assuming it.

Job with the failures: https://github.com/metasfresh/metasfresh/actions/runs/34602371331/job/103279837290

backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/delivery/deliveryInstructionAggregation.feature
mrie04: **Resolution (dev):** Same change as the sibling thread, and your question was the right one: an actual discharge had no business being set while the actual load was 0.

`ActualDischargeQuantity` now follows `ActualLoadQty` and nothing else - so on these two outgoing plannings, where nothing has shipped, BOTH actuals are 0. The planned figures (7 and 3) live in the sibling view, which is where a plan belongs. The table here and its comment are updated to say that.

Verification state is the same as the sibling thread: 12 unit tests green, cucumber re-run in flight against a proper reactor build (my earlier local runs linked a stale jar and exercised the old code).

backend/de.metas.deliveryplanning.base/src/main/sql/postgresql/system/43-de.metas.deliveryplanning/5824120_sys_gh31789_QtyTotalOpenPlanned_mandatory.sql
mrie04: **Resolution (dev):** Checked by running it rather than reasoning about it - it works, and the reason is the opposite of the concern.

Within ONE transaction, PostgreSQL statements see that transaction's own prior writes; there is no "not committed yet" barrier between statements. So `UPDATE` then `ALTER ... SET NOT NULL` in the same script succeeds. Demonstrated directly against the local stack, inside an explicit uncommitted transaction:

```
begin;
create temp table t_nn_demo(x numeric);
insert into t_nn_demo values (null),(null);
update t_nn_demo set x=0 where x is null;      -- UPDATE 2
alter table t_nn_demo alter column x set not null;   -- ALTER TABLE  (succeeds)
rollback;
```

`SET NOT NULL` scans the table as the transaction sees it - i.e. including the backfill - so it finds no NULLs. Had the backfill been invisible, the ALTER would have raised `column "x" contains null values`.

And you are right that nothing introduces new NULLs while the migration runs, which makes the ordering concern moot from the other side too. The script keeps the SET DEFAULT -> backfill -> SET NOT NULL order regardless, since that is the order that is also correct for a fresh install where the column is created empty.

Beyond the isolated demo, the script itself was applied by `db-apply-migrations` on six separate runs against a fresh database, all green.

backend/de.metas.deliveryplanning.base/src/main/java/de/metas/deliveryplanning/DeliveryPlanningService.java
```diff
@@ -1981,4 +2163,67 @@ public Optional<MeansOfTransportation> getMeansOfTransportationByDeliveryPlannin
...
+	 * whose lines were stamped by the backfill of the retired header column, and their quantity is not this
+	 * planning's product's.
+	 */
+	private Quantity resolveBookedQty(@NonNull final I_M_InOut inout, @NonNull final I_M_Delivery_Planning planningRecord)
+	{
+		final ProductId productId = ProductId.ofRepoId(planningRecord.getM_Product_ID());
+		final I_C_UOM uom = uomDAO.getById(planningRecord.getC_UOM_ID());
+		final int planningRepoId = planningRecord.getM_Delivery_Planning_ID();
+
+		return inOutDAO.retrieveLines(inout).stream()
```
mrie04: delegate vi bl and not introduce both

mrie04: **Resolution (dev):** Done - `inOutBL.getLines(inout)`, and the `IInOutDAO` field and its import are gone.

You were precise about it: `inOutDAO` was used at exactly ONE call site in the whole class, while `IInOutBL` was already injected and used two lines further down (`inOutBL::getMovementQty`). So the class was carrying a second collaborator for a single call the first one already offers - `IInOutBL#getLines(I_M_InOut)`.

Module compiles clean (main + test scope verified).

PR #25730, opened 2026-09-04
CI images: move the JDK-8 Maven stages off Debian 11 (EOL) to maven:3.8.8-eclipse-temurin-8

metas-ts:
## Problem

`build (cucumber)` failed three times in a row on `deep_tundra_release` (original + two re-runs of https://github.com/metasfresh/metasfresh/actions/runs/33842137998), and then `test (java)` failed the same way on `intensive_care_hotfix` (https://github.com/metasfresh/metasfresh/actions/runs/33861350675) — both in the first `RUN` of the image (`Dockerfile.cucumber` / `Dockerfile.junit`):

```
Err:2 http://security.debian.org/debian-security bullseye-security/main amd64 locales all 2.31-13+deb11u14
  404  Not Found [IP: 146.75.122.132 80]
```

Both images build `FROM maven:3.8.4-jdk-8`, which is **Debian 11 (bullseye) — LTS ended 2026-08-31** (https://wiki.debian.org/LTS). Building on an EOL suite is the root cause; the CDN 404 four days after EOL is the symptom, and re-running does not help.

## Fix

Move every JDK-8 Maven stage to `maven:3.8.8-eclipse-temurin-8`:

| Dockerfile | before | after |
|---|---|---|
| `Dockerfile.common` | `maven:3.8.4-jdk-8` | `maven:3.8.8-eclipse-temurin-8` |
| `Dockerfile.backend` (build stage) | same | same |
| `Dockerfile.cucumber` | same | same |
| `Dockerfile.junit` | same | same |
| `Dockerfile.migration-cli` | same | same |

They move together because they share the `/root/.m2` populated by the earlier stages. Target facts (verified with `docker run`): Ubuntu 22.04.3 jammy (standard support until 2027-04-01), Apache Maven 3.8.8 (same minor as before), Temurin 1.8.0_392; `locales` and `mmv` resolvable from the default jammy sources; no Dockerfile or script hardcodes a JDK path / `JAVA_HOME`.

The first commit on this branch added an interim `security.debian.org → deb.debian.org` mirror fallback to `Dockerfile.cucumber`; on a jammy base that is dead code (its apt sources use `archive.ubuntu.com` / `security.ubuntu.com`), so the second commit reverts it to the plain install line. Net diff vs the base branch: the five `FROM` lines plus a two-line comment in `Dockerfile.common`.

## Verification

Authoritative: this PR's CI — `build-commons → backend → build (cucumber) / test (java) / migration-cli` rebuild from scratch on the new base (the base-branch layer cache does not apply to a new `FROM`).

Out of scope (tracked separately): the other EOL bases in `docker-builds/` — `node:14` (Debian 10 buster) in `Dockerfile.mobile*`, `maven:3.8.4-eclipse-temurin-17` (Ubuntu 20.04 focal) in `Dockerfile.camel*`, `ubuntu:16.04` in `Dockerfile.frontend.compat`.

### DIFF
--- a/docker-builds/Dockerfile.common
+++ b/docker-builds/Dockerfile.common
@@ -1,5 +1,3 @@
-# maven:3.8.4-jdk-8 was Debian 11 (bullseye), EOL 2026-08-31. Jammy-based Temurin 8, same Maven minor.
-# Keep common / backend / cucumber / junit / migration-cli on the SAME tag — they share /root/.m2.
 FROM maven:3.8.8-eclipse-temurin-8
 
 WORKDIR /parent
--- a/docker-builds/Dockerfile.common
+++ b/docker-builds/Dockerfile.common
@@ -1,4 +1,6 @@
-FROM maven:3.8.4-jdk-8
+# maven:3.8.4-jdk-8 was Debian 11 (bullseye), EOL 2026-08-31. Jammy-based Temurin 8, same Maven minor.
+# Keep common / backend / cucumber / junit / migration-cli on the SAME tag — they share /root/.m2.
+FROM maven:3.8.8-eclipse-temurin-8
 
 WORKDIR /parent
 COPY misc/parent-pom .
--- a/docker-builds/Dockerfile.backend
+++ b/docker-builds/Dockerfile.backend
@@ -14,7 +14,7 @@ COPY backend .
 RUN find . -type f ! -name "pom.xml" -exec rm -r {} \;
 
 
-FROM maven:3.8.4-jdk-8 AS build
+FROM maven:3.8.8-eclipse-temurin-8 AS build
 
 WORKDIR /backend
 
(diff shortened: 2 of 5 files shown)


docker-builds/Dockerfile.common
```diff
@@ -1,4 +1,6 @@
-FROM maven:3.8.4-jdk-8
+# maven:3.8.4-jdk-8 was Debian 11 (bullseye), EOL 2026-08-31. Jammy-based Temurin 8, same Maven minor.
+# Keep common / backend / cucumber / junit / migration-cli on the SAME tag — they share /root/.m2.
```
metas-ts: why this comment? what benefit does it add?

metas-ts: **Resolution (dev):** Agreed — it narrated history (the old base, its EOL date) that belongs in git/the PR, and the "keep the five stages on the same tag" line overstated a soft preference as a rule (a Maven/JDK patch mismatch between the .m2 producer and its consumers would not actually break anything). Removed in https://github.com/metasfresh/metasfresh/commit/5f4db89222f9538871cc2f5b3f4efbdd4fc8d0da; the Dockerfile now carries only the `FROM` change.

PR #25830, opened 2026-09-08
Empty-HU write-off on raw-materials issue — reject reason, config flags, per-step gate and server-side guard

mrie04:
## Empty-HU write-off on raw-materials issue

When a handling unit is emptied during a manufacturing raw-materials issue, the operator can write off the remaining quantity: the leftover is booked away by an automatic inventory disposal instead of staying on the HU.

### Application dictionary

- `5822940_sys_EmptiedRejectReason.sql` — new reject reason `E` on `AD_Reference` 541422 (`AD_Ref_List` 544360, `ValueName='Emptied'`), plus `AD_Val_Rule` 540801 (`AD_Ref_List.Value <> 'E'`) attached to six `RejectReason` columns that must *not* offer it. `PP_Order_IssueSchedule.RejectReason` — the column that stores `E` — is deliberately left without the rule.
- `5822960_sys_EmptiedHUInventoryDescription.sql` — `AD_Message` `de.metas.handlingunits.pporder.EmptiedHUInventoryDescription`, used as the description of the generated inventory.
- `5822970_MobileUI_MFG_Config_EmptyingHUs.sql` — `MobileUI_MFG_Config.IsAllowEmptyingHUs` and `IsConfirmEmptyingHU` (both `CHAR(1) NOT NULL DEFAULT 'Y'`, with check constraints), their `AD_Element`s, fields and UI elements. Defaulting to `'Y'` means no per-instance activation migration is required.
- `5822980_sys_DD_NetworkDistribution_descriptions.sql` — Distribution Network window: descriptions and help for the empties-network flag and the line fields, in `en_US` / `de_DE` / `de_CH`.
- `5823140_sys_EmptyingNotAllowedForHU.sql` — `AD_Message` `EmptyingNotAllowedForHU`, the refusal message when the emptied reason is requested for an ineligible step, plus its `ErrorCode` and `EntityType`.

### Backend

- `QtyRejectedReasonCode.EMPTIED` — the new code, declared after the interner it dereferences.
- `QtyRejectedReasonContext` + `QtyRejectedReasonCode.reasonsFor(...)` — the single place that knows which reason codes a given context may offer. Picking, distribution and inventory disposal exclude `EMPTIED`; the manufacturing issue offers it. Call sites in `PickingCandidateService`, `DDOrderMoveScheduleService` and `InventoryCandidateService` pass their context.
- `ADRefList#excluding(...)` — returns a reduced list, or `this` when nothing is excluded.
- `MobileUIManufacturingConfig` carries the two new flags as `OptionalBoolean` so the user/client/default merge chain keeps working; `MobileUIManufacturingConfigRepository.DEFAULT_CONFIG` has both `TRUE`. `MobileUIManufacturingConfigRepository.saveGlobalConfig(...)` is a client-scoped save path (used by the test harness to drive the confirmation flag deterministically from an E2E spec).
- **Per-step eligibility decided at job-load time**, in `ManufacturingJobLoaderAndSaver` (via `ManufacturingJobLoaderAndSaverSupportingServices`): a non-aggregate transport-unit/virtual HU (`HUType` allow-list `TransportUnit`, `VirtualPI`, decided by an exhaustive switch) holding at most one product storage is eligible; a pallet's (LU) step is not, and the pallet's per-TU alternative step — reachable only as a zero-quantity step offering no reason — is not eligible either. `RawMaterialsIssueActivityHandler` reads that per-step decision to gate the offered reason and to serve the confirmation flag on the wire; `toJson` performs no DB access.
- **Server-side guard**: `ManufacturingJobService.issueRawMaterialsInTrx` refuses the `E` reason when the step is ineligible or when the client config does not offer it, instead of trusting the client — covered by a JUnit test in `ManufacturingJobServiceTest`.
- `DistributionNetworkRepository`: `createNetwork(CreateDistributionNetworkRequest)` and `CreateDistributionNetworkRequest` ported from `new_dawn_uat`, plus `addLine(...)`, `getEmptiesDistributionNetworkIfExists()` and `IsHUDestroyed` support (unit-tested). `IHandlingUnitsBL#retrieveIncludedHUs(HuId)` as the id-based twin of the `I_M_HU` overload. `InOutPackageRepository#createM_Package(OrgId, ShipperId, documentNo)` for callers without an `InOut`.
- `PPOrderIssueScheduleService`: after the issue producer ran, `bookEmptiedHUToZero` books the HU's remaining storage to zero through `HUQtyService` (inventory with QtyBook = the post-issue remainder, QtyCount = 0, carrying the write-off description); the booking is skipped when the issue already consumed the HU, or when the HU carries no product storage. The weighing step (`weightHU` / `WeightHUCommand`) is unchanged from the base branch.
- Mobile UI: the reason is filtered and offered per step, with a confirmation prompt (naming the leftover quantity and its UOM) gated by the confirmation flag. The issue request is built by the pure `computeIssueRequest` (jest-tested): the reason and `qtyRejected` follow the offer-time whole-HU flag, while the HU gross weight (`huWeightGrossBeforeIssue`) follows the TYPED quantity exactly as on the base branch — sent only when the quantity is entered in `kg` and reaches the HU capacity. A short `kg` entry with the reason therefore sends no weight, so the HU is not re-weighed before the issue and the leftover is written off after it.

### Generated model

Regenerated: `I_MobileUI_MFG_Config` / `X_MobileUI_MFG_Config`, `I_DD_OrderLine_HU_Candidate` / `X_DD_OrderLine_HU_Candidate`, `I_M_Picking_Candidate` / `X_M_Picking_Candidate` for the new/changed reject-reason-bound columns. `I_M_Picking_Job_Step` is hand-touched, not regenerated — the generator drops its object-returning foreign-key accessors when run at the wrong module scope, so it is deliberately kept as-is.

### Test support

Frontend-testing harness extensions: included TUs' QR codes on LU creation; partial fill when creating an LU with packing instructions; additional products on one HU; `tuPackingMaterial` on a packing instruction; a `shippers` section (ported from `new_dawn_uat`); `warehouses.<X>.empties: { toWarehouse, shipper }`, which adds a line to the client's single empties distribution network (created only if the client has none — production allows exactly one); a `packages` request/response (`packages.<X>.shipper` references a created shipper); inventory expectations scoped by description with a `count`; inventory expectations can pin `qtyBook` / `qtyCount` of the described inventory; movement expectations via `IMovementDAO#retrieveLinesByQuery` (a filter-less query is rejected). The expectations service reads through services and business logic only (no `queryBL`, no DAO next to a service).

### Tests

Five Playwright specs, 19 tests total: `issue_emptyHU_writeOff.spec.js` (4), `issue_emptyHU_offering.spec.js` (3), `issue_emptyHU_otherReasonsUnchanged.spec.js` (6), `issue_emptyHU_refusals.spec.js` (3), `issue_emptyHU_catchWeight.spec.js` (3). Run locally against a server built from this branch after the weight-predicate fix: 19/19 green at the current HEAD (one full run, 1.6 min); the three spec files re-aligned in that fix were additionally repeated 3× each (36/36); the catch-weight and write-off specs 7/7 right after the backend rebuild. Earlier flakiness proof: 90/90 (18 tests × 5 runs) before the fix.

### Defects found and fixed

- **Reject reasons never reached the backend on a partial manufacturing issue.** The step-scan flow held two different definitions of "whole HU": the offer side used `qtyToIssueTarget >= qtyHUCapacity`, the submit side recomputed `qty >= qtyHUCapacity` from what the operator typed. So a reason was offered on the first and discarded on the second. The reason/`qtyRejected` side now reads the offer-time flag; the weight side keeps the typed-quantity rule (see the next-but-one item). Pre-existing, traced to the 2023 merge https://github.com/metasfresh/metasfresh/commit/d0c125fea8c0e6b255139bb6d669e3851cb6cefb.
- **`globalConfigsCache` was registered under the wrong table** (`MobileUI_UserProfile_MFG` instead of `MobileUI_MFG_Config`), so a write to the client-level config never invalidated it. Introduced by this branch's own earlier commit.
- **Introduced by this branch, found on the dev instance, fixed before merge: a short `kg` entry was sent as the HU's gross weight.** An intermediate commit had let the weight follow the offer-time whole-HU flag, so a whole-HU step entered 10 g short (within the issue tolerance) re-weighed the HU down to the typed quantity before the issue and produced a single inventory with QtyBook = the pre-issue quantity and no after-issue write-off. The weight predicate is back to the typed quantity (`uom === 'kg' && qty >= qtyHUCapacity`), pinned by jest and by the catch-weight spec (short entry → no weight, one write-off inventory QtyBook = leftover / QtyCount = 0). The description threading through `WeightHUCommand` that this intermediate state had required is removed again.
- **A weighed whole-HU issue could be written off a second time (double-counting)** — fixed by a status guard skipping the booking once the issue already consumed the HU.
- **The client-trusting write path** accepted the `E` reason from the request without checking step eligibility or the config — fixed by the server-side guard in `ManufacturingJobService.issueRawMaterialsInTrx`.
- **`CreateMasterdataCommandTest` NPE** from the mandatory `packages` response map — fixed in the hand-built test fixture; the field stays `@NonNull` like its siblings.
- **`InventoryRepository.retrieveAllLinesForHU` failed at runtime** (`Unknown parameter type … Optional[AdTableId]`): the HU-assignment leg passed an `Optional<AdTableId>` as SQL parameter (latent on `new_dawn_uat` too). Fixed with `getTableIdNotNull` and a unit test pinning the assignment path.

### Scope notes

- Pallets are excluded on every step: the LU step is excluded by the HU-type allow-list; a single TU on a pallet is only reachable as a zero-quantity alternative step, which offers no reason — covered by a Playwright test.
- The distribution app renders no reject reasons at all today (pre-existing), so there is no leakage test for that surface.
- The refusal for an HU that carries packing material but no empties line cannot be staged in a test, because such an HU cannot be created in the first place — the same refusal already fires at receipt.


backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/expectations/AssertExpectationsCommandServices.java
```diff
@@ -116,4 +139,72 @@ public List<I_M_HU> getIncludedHUs(@NonNull final HuId huId)
...
+	 * touched an included HU or an HU assignment. The repository owns the query; do not hand-roll one here.
+	 */
+	public List<I_M_InventoryLine> getInventoryLinesByHUId(@NonNull final HuId huId)
+	{
+		return inventoryService.getInventoryRepository().retrieveLinesByHUId(huId);
+	}
+
+	public I_M_Inventory getInventoryById(@NonNull final InventoryId inventoryId)
+	{
+		return inventoryDAO.getById(inventoryId);
```
mrie04: don't add dao and service delegate via service

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/expectations/AssertExpectationsCommandServices.java
```diff
@@ -116,4 +139,72 @@ public List<I_M_HU> getIncludedHUs(@NonNull final HuId huId)
...
+			throw new AdempiereException("Packing material has no product").setParameter("huId", huId);
+		}
+		return ProductId.ofRepoId(product.getM_Product_ID());
+	}
+
+	/**
+	 * Whether a completed {@code M_Movement} exists with an {@code M_MovementLine} moving the given
+	 * product from a locator of {@code fromWarehouseId} to a locator of {@code toWarehouseId}.
+	 */
+	public boolean hasCompletedMovementLine(
```
mrie04, line 168: better introduce getLineByQuery in dao

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/expectations/AssertExpectationsCommandServices.java
```diff
@@ -116,4 +139,72 @@ public List<I_M_HU> getIncludedHUs(@NonNull final HuId huId)
...
+		final Set<Integer> toLocatorRepoIds = toRepoIds(warehouseDAO.getLocatorIds(toWarehouseId));
+
+		final IQuery<I_M_Movement> completedMovementQuery = queryBL.createQueryBuilder(I_M_Movement.class)
+				.addOnlyActiveRecordsFilter()
+				.addEqualsFilter(I_M_Movement.COLUMNNAME_DocStatus, IDocument.STATUS_Completed)
+				.create();
+
+		return queryBL.createQueryBuilder(I_M_MovementLine.class)
+				.addOnlyActiveRecordsFilter()
+				.addEqualsFilter(I_M_MovementLine.COLUMNNAME_M_Product_ID, productId.getRepoId())
```
mrie04: repoidawares are supported, no need for unbox, same for sets

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/hu/CreateHUCommand.java
```diff
@@ -5,9 +5,11 @@
 import de.metas.common.util.time.SystemTime;
 import de.metas.frontend_testing.masterdata.Identifier;
 import de.metas.frontend_testing.masterdata.MasterdataContext;
+import com.google.common.collect.ImmutableList;
 import de.metas.handlingunits.HuId;
 import de.metas.handlingunits.IHUContext;
 import de.metas.handlingunits.IHandlingUnitsBL;
+import de.metas.handlingunits.IHandlingUnitsDAO;
```
mrie04: same as on other delegate via bl

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/hu/JsonPackingInstructionsRequest.java
```diff
@@ -31,6 +31,13 @@ public class JsonPackingInstructionsRequest
 	@Nullable Identifier lu;
 	int qtyTUsPerLU;
 
+	/**
+	 * Packing material (e.g. a crate/pallet product) to attach to the TU's PI version, as an
+	 * {@code M_HU_PI_Item} of {@code ItemType = PackingMaterial}. When set, every HU created from
+	 * this TU PI carries an {@code M_HU_PackingMaterial}. Not applicable when {@link #cu} is set.
+	 */
+	@Nullable Identifier packingMaterial;
```
mrie04: imho should be tuPackingMaterial, to distinct from possible lu later

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/hu_package/JsonPackageResponse.java
```diff
@@ -0,0 +1,15 @@
...
+import lombok.NonNull;
+import lombok.Value;
+import lombok.extern.jackson.Jacksonized;
+
+@Value
+@Builder
+@Jacksonized
+public class JsonPackageResponse
+{
+	int packageId;
```
mrie04: why not the RepoIdAware

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/hu_package/PackageCommand.java
```diff
@@ -0,0 +1,54 @@
...
+{
+	@NonNull private final MasterdataContext context;
+	@NonNull private final JsonPackageRequest request;
+	@NonNull private final Identifier identifier;
+
+	public JsonPackageResponse execute()
+	{
+		final HuId huId = context.getId(request.getHu(), HuId.class);
+
+		final I_M_Package packageRecord = InterfaceWrapperHelper.newInstance(I_M_Package.class);
```
mrie04: why not de.metas.handlingunits.shipping.InOutPackageRepository#createM_Package

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/mobile_configuration/MobileConfigManufacturingCommand.java
```diff
@@ -33,10 +33,44 @@ public JsonMobileConfigResponse.Manufacturing execute()
 		final MobileUIManufacturingConfig newConfig = newConfigBuilder.build();
 		mobileManufacturingConfigRepository.saveUserConfig(newConfig, loginUserId);
 
+		// IsAllowEmptyingHUs / IsConfirmEmptyingHU are client-level only (MobileUI_MFG_Config has no
+		// per-user column for them, cf. RawMaterialsIssueActivityHandler#resolveEmptyingHUsConfig) —
+		// route them to the global config, never to the per-user profile above.
+		if (request.getIsAllowEmptyingHUs() != null || request.getIsConfirmEmptyingHU() != null)
```
mrie04, line 39: why not before save above like the others. this is not ok

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/warehouse/JsonWarehouseRequest.java
```diff
@@ -17,6 +17,14 @@ public class JsonWarehouseRequest
 	@Nullable String locatorCode;
 	@Nullable Map<String, Locator> locators;
 
+	/**
+	 * When set, reference a pre-existing seeded warehouse by a stable name instead of creating a new one.
+	 * Supported value: {@code "standard"} — the seeded standard warehouse (the one empties network
+	 * 540011's seeded line covers). No new {@code M_Warehouse} row is created, and its default locator is
+	 * resolved read-only (never renamed), since it is shared across test runs.
+	 */
+	@Nullable String existing;
```
mrie04: I think you should create your own empty warehouse, warehouse and network

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/warehouse/WarehouseCommand.java
```diff
@@ -74,6 +80,44 @@ public JsonWarehouseResponse execute()
 		return responseBuilder.build();
 	}
 
+	/**
+	 * Resolves {@code request.existing} to a pre-seeded warehouse instead of creating a new one. Read-only:
+	 * does NOT rename the shared default locator (unlike {@link #createDefaultLocator()}), since that
+	 * locator is shared across every test run referencing this warehouse.
+	 */
+	private JsonWarehouseResponse executeExisting(@NonNull final String existing)
```
mrie04: should be done after other request

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/MasterdataContext.java
```diff
@@ -50,14 +55,42 @@ public class MasterdataContext
...
+
 	public @NonNull String getAdLanguage() {return Env.getADLanguageOrBaseLanguage();}
 
+	/**
+	 * Resolves the seeded default shipper lazily, instead of a literal {@code M_Shipper_ID}: no
+	 * masterdata command owns a "shipper" concept (grepped {@code masterdata/**}), and no seeded shipper
+	 * carries {@code IsDefault=Y}, so this picks the first active {@link I_M_Shipper} (ordered by id) for
+	 * {@link #CLIENT_ID} instead. Cached on this context, like other resolved defaults.
+	 */
+	public @NonNull ShipperId getDefaultShipperId()
```
mrie04: there is no default shipper, create your own if needed for network, check new_dawn for existing logic

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/inventory/InventoryRepository.java
```diff
@@ -262,6 +262,21 @@ public void createInventoryLine(@NonNull final InventoryLineCreateRequest reques
...
+	public List<I_M_InventoryLine> retrieveLinesByHUId(@NonNull final HuId huId)
+	{
+		return queryBL.createQueryBuilder(I_M_InventoryLine.class)
+				.addEqualsFilter(I_M_InventoryLine.COLUMNNAME_M_HU_ID, huId)
+				.orderBy(I_M_InventoryLine.COLUMNNAME_M_InventoryLine_ID)
+				.create()
+				.list();
+	}
+
 	public Collection<I_M_InventoryLine> retrieveAllLinesForHU(@NonNull final HuId huId)
```
mrie04, line 265: wgynew in addtion to retrieveAllLinesForHU

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/expectations/AssertExpectationsCommandServices.java
```diff
@@ -46,6 +65,10 @@ public class AssertExpectationsCommandServices
 	@NonNull public final IHandlingUnitsBL handlingUnitsBL = Services.get(IHandlingUnitsBL.class);
 	@NonNull private final IHandlingUnitsDAO handlingUnitsDAO = Services.get(IHandlingUnitsDAO.class);
 	@NonNull private final IHUPPOrderQtyDAO huPPOrderQtyDAO = Services.get(IHUPPOrderQtyDAO.class);
+	@NonNull private final IInventoryDAO inventoryDAO = Services.get(IInventoryDAO.class);
+	@NonNull private final IWarehouseDAO warehouseDAO = Services.get(IWarehouseDAO.class);
+	@NonNull private final IQueryBL queryBL = Services.get(IQueryBL.class);
```
mrie04: queryBl shouldn't be used outside dao repo

mrie04: **Resolution (dev):** moved every query out of the expectations service into the owning DAO/repository: the movement lookup is now `IMovementDAO#getLineByQuery(MovementLineQuery)` and the HU reads go through `IHandlingUnitsBL`; `queryBL` is gone from `AssertExpectationsCommandServices`. https://github.com/metasfresh/metasfresh/commit/8183a75f9438b7bd82c4c1487df2d1c8d404390d, https://github.com/metasfresh/metasfresh/commit/0fbb6f888a068d76c43434d7539b2a11cf5977c9

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/expectations/AssertExpectationsCommandServices.java
mrie04: **Resolution (dev):** the service no longer injects DAOs next to services: inventory lines are read via `InventoryService`, HUs via `IHandlingUnitsBL` (`retrieveIncludedHUs(HuId)` added as the id-based twin of the existing overload). https://github.com/metasfresh/metasfresh/commit/8183a75f9438b7bd82c4c1487df2d1c8d404390d, https://github.com/metasfresh/metasfresh/commit/0fbb6f888a068d76c43434d7539b2a11cf5977c9

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/expectations/AssertExpectationsCommandServices.java
mrie04: **Resolution (dev):** introduced `IMovementDAO#getLineByQuery(MovementLineQuery)` (firstOptional, ordered by id) and the service calls it; a filter-less query is rejected in the DAO so an expectation without filters can never be satisfied by an arbitrary first line (unit-tested). https://github.com/metasfresh/metasfresh/commit/8183a75f9438b7bd82c4c1487df2d1c8d404390d, https://github.com/metasfresh/metasfresh/commit/c8c082c88da00a13407d1a447098ad4816cae799

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/expectations/AssertExpectationsCommandServices.java
mrie04: **Resolution (dev):** the filters now take the RepoIdAware values and sets directly (`ProductId`, `Set<LocatorId>`), no more unboxing. https://github.com/metasfresh/metasfresh/commit/8183a75f9438b7bd82c4c1487df2d1c8d404390d

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/hu/CreateHUCommand.java
mrie04: **Resolution (dev):** `CreateHUCommand` no longer touches `IHandlingUnitsDAO`; the included-HU lookup goes through `IHandlingUnitsBL#retrieveIncludedHUs(HuId)`. https://github.com/metasfresh/metasfresh/commit/6972722415fe5b444a43e7301dbf385df796a8ca

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/hu/JsonPackingInstructionsRequest.java
mrie04: **Resolution (dev):** renamed to `tuPackingMaterial` (Java request field + every Playwright fixture). https://github.com/metasfresh/metasfresh/commit/6972722415fe5b444a43e7301dbf385df796a8ca, https://github.com/metasfresh/metasfresh/commit/00ad6e88d0057029d0677f656969cb3a225fa761

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/hu_package/JsonPackageResponse.java
mrie04: **Resolution (dev):** `JsonPackageResponse` now exposes `PackageId` (RepoIdAware, serialised via its `@JsonValue`), and the ported `JsonCreateShipperResponse` uses `ShipperId` the same way. https://github.com/metasfresh/metasfresh/commit/6972722415fe5b444a43e7301dbf385df796a8ca, https://github.com/metasfresh/metasfresh/commit/d22c00ddbea91966fc93eba34df508697f49438e

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/hu_package/PackageCommand.java
mrie04: **Resolution (dev):** the package is now created through `InOutPackageRepository`. The existing `createM_Package(CreatePackagesRequest)` is private and requires an `InOutId` the masterdata command does not have, so I added a public `createM_Package(OrgId, ShipperId, documentNo)` on that repository and linked the HU via the existing `IHUPackageBL#assignPackageToHuId`; the command has no hand-rolled record code left. If you would rather widen `CreatePackagesRequest.inOutId` to nullable instead, say so and I switch. https://github.com/metasfresh/metasfresh/commit/6972722415fe5b444a43e7301dbf385df796a8ca

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/mobile_configuration/MobileConfigManufacturingCommand.java
mrie04: **Resolution (dev):** not moved before `saveUserConfig`, because it would silently drop the values: `IsAllowEmptyingHUs`/`IsConfirmEmptyingHU` exist only on `MobileUI_MFG_Config` (client level; migration 5822970 touches only that table), `MobileUI_UserProfile_MFG` has no such columns, and `MobileUIManufacturingConfigRepository#saveUserConfig` writes only `IsScanResourceRequired` and `IsAllowIssuingAnyHU`. The approved requirements define the two flags as client-level with no per-user override, so the test-support command routes them to the global config via `saveGlobalConfig`. A per-user override would be a requirements change (two new columns on the user profile) — happy to do it if that is what you want.

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/warehouse/JsonWarehouseRequest.java
mrie04: **Resolution (dev):** the seeded-warehouse concept is gone. Masterdata now creates its own warehouses and its own shipper (`shippers` section, ported from `new_dawn_uat`), and `warehouses.<X>.empties: { toWarehouse, shipper }` adds a line (X → toWarehouse) to the client's empties distribution network. One constraint kept us from creating a whole new network per test: `DistributionNetworkRepository#getEmptiesDistributionNetwork` throws on more than one `IsHUDestroyed` network, and `HUEmptiesService` resolves the empties warehouse by the source-warehouse line of that single network — so the harness creates the network only if the client has none and otherwise adds a line (resolved once per request). Repository API (`createNetwork`/`addLine`) ported from `new_dawn_uat` plus `IsHUDestroyed` support, unit-tested. https://github.com/metasfresh/metasfresh/commit/471356eec8c29cff0858ee11fa064e0e2ef07295, https://github.com/metasfresh/metasfresh/commit/d22c00ddbea91966fc93eba34df508697f49438e, https://github.com/metasfresh/metasfresh/commit/27eca23cab73d8fc3fac59ba295b04a08a84bb8c, https://github.com/metasfresh/metasfresh/commit/a3f5fc9cc1d67508c929d443901f944f113db4e5, https://github.com/metasfresh/metasfresh/commit/327fa0d3e95f8b4fe1f8600f0a4720bf8c734814

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/warehouse/WarehouseCommand.java
mrie04: **Resolution (dev):** `executeExisting` and the `existing` request field are removed altogether (see the warehouse/network thread); the empties line is applied in a post-pass after the `shippers` and `warehouses` sections. https://github.com/metasfresh/metasfresh/commit/27eca23cab73d8fc3fac59ba295b04a08a84bb8c

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/MasterdataContext.java
mrie04: **Resolution (dev):** the default-shipper lookup is deleted. Shippers are created by the new `shippers` masterdata section (ported from `new_dawn_uat`'s `CreateShipperCommand`), and both the empties network line and `packages.<X>.shipper` reference a created shipper by identifier. https://github.com/metasfresh/metasfresh/commit/d22c00ddbea91966fc93eba34df508697f49438e, https://github.com/metasfresh/metasfresh/commit/27eca23cab73d8fc3fac59ba295b04a08a84bb8c

backend/de.metas.handlingunits.base/src/main/java/de/metas/handlingunits/inventory/InventoryRepository.java
mrie04: **Resolution (dev):** the new repository method is removed; the expectation uses the existing `retrieveAllLinesForHU` and filters the result by `M_HU_ID` so it still matches only lines booked on exactly that HU. Doing so surfaced a latent bug in `retrieveAllLinesForHU` itself (the HU-assignment filter passed an `Optional<AdTableId>` as SQL parameter → `Unknown parameter type` at runtime, on this base and on `new_dawn_uat`), fixed with a unit test that pins the assignment path. https://github.com/metasfresh/metasfresh/commit/8183a75f9438b7bd82c4c1487df2d1c8d404390d, https://github.com/metasfresh/metasfresh/commit/b989c92bc90e190d167b7171a0f737441a22df94

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/expectations/AssertExpectationsCommandServices.java
mrie04: **Resolution (dev):** follow-up — `getLineByQuery` (Optional, first by id) was over-specified for a caller that only asks "does any matching line exist"; replaced by `IMovementDAO#retrieveLinesByQuery(MovementLineQuery)` returning the matching lines in no particular order (filter-less query still rejected), caller checks non-empty, test exercises the list shape. https://github.com/metasfresh/metasfresh/commit/e691c51301f016931750d1a8c02716331d142909

mrie04:
**Post-review change — please re-check the delta since your round.** While testing the reviewed state on the dev instance, a whole-HU `kg` step entered 10 g short (within the issue tolerance) with the emptied reason produced a single inventory with QtyBook = the pre-issue quantity and no after-issue write-off. Cause: an intermediate commit of this branch let the HU gross weight follow the offer-time whole-HU flag instead of the typed quantity, so the short entry re-weighed the HU down before the issue — a regression against `task_force_hotfix`.

Fixed before merge, all in https://github.com/metasfresh/metasfresh/compare/878d583f411aae49f109b472b24343722257b14c...442bfed3050b3c209c1820e399cb3b4c1c4a05fc (the reviewed state was https://github.com/metasfresh/metasfresh/commit/878d583f411aae49f109b472b24343722257b14c):
- the issue request is built by a pure `computeIssueRequest` (jest-tested): the weight follows the typed quantity exactly as on the base branch (`uom === 'kg' && qty >= qtyHUCapacity`), the reason and `qtyRejected` keep the offer-time flag;
- the write-off description no longer travels through `WeightHUCommand` (back to the base signature);
- inventory expectations can pin `qtyBook` / `qtyCount`; the catch-weight spec asserts the short entry sends no weight and books exactly QtyBook 0.002 / QtyCount 0; the other-reasons / flag-off specs now pin the base behaviour (HU stays active with the remainder) instead of the regressed one.

Evidence: 19/19 Playwright locally against a server built from this branch, unit suites of `de.metas.handlingunits.base` and `de.metas.frontend-testing` green, `inventory` cucumber feature green; whole-state code review with zero findings at the current HEAD. The PR description is updated accordingly.

PR #25905, opened 2026-09-10
Free text above order lines

metas-ts:
## Free text above order lines

Adds a new plain-text column `C_OrderLine.DescriptionAboveLine` whose content prints as a standalone
full-width block **immediately above that order line's own article row** on the order confirmation,
the delivery note, the picking list and the invoice.

It is a second, independent free-text field next to `Description` — which prints *below* the line —
so an order clerk can introduce a line ("Sortimentsware", a delivery note for the warehouse) rather
than only annotate it. Empty, blank or whitespace-only content prints nothing at all: no block, and
no empty vertical gap.

### What is in here

- **Column + field.** `C_OrderLine.DescriptionAboveLine`, nullable `varchar(2048)`, `AD_Reference 14`
  (Text), on a reusable `AD_Element`; placed on the sales-order line tab beside `Description`, with
  regenerated model classes.
- **Report sources.** The column is exposed as `descriptionaboveline` by every core sales
  line-detail source: the `Docs_Sales_{Order,InOut,Invoice,Picking}_Details` functions and their
  `_HU` variants, plus the `PickingList` view. Each function's `ddl/` body and its shipped migration
  copy are kept in lockstep.
- **Templates.** The print band is added to all **15** core line-detail templates under
  `docs/sales/{order,inout,invoice,picking,pickingv2}`. Each band is guarded so it prints only when
  the text is non-empty, sits first in `<detail>` so it renders above the article row, and spans the
  full width of its own template's content row.
- **Configurable per document type.** A new `AD_Ref_List` value makes the element selectable in
  `C_DocType` → report elements, so the block can be hidden per document type exactly like the
  sibling line description. Without that value the `IsHiddenReportElement` gate would have been
  unreachable.
- **Field help written from observed behaviour**, including a real limitation: the document font
  resolves to Liberation Sans with `Cp1252`, so any character outside Windows-1252 (emoji, `✓`, `→`,
  letters outside Western Europe) is dropped from the PDF **silently**, with no replacement glyph,
  even though the WebUI editor displays it. Umlauts, sharp s, accented Western European letters and
  the euro sign are inside the range and print correctly.

### Verification

- **Cucumber** — four scenarios assert the printed text *inside the rendered PDF* (via the archive
  storage, not the raw `BinaryData` column, which is zipped). The assertions are anchored on
  document geometry and on text known to be present on both sides, and were proven to discriminate
  by mutation: removing the template band fails them.
- **Playwright** — the order confirmation, delivery note and invoice are downloaded and asserted in
  both English and German (2 passed / 0 failed), against a from-source stack. Also mutation-proven:
  the app recompiles the JRXML from the working tree, so deleting the band turns the assertion red.
- **Templates** — all 15 compile under JasperReports 6.21.2, which exercises the Java in the new
  guard expressions and not merely the XML.

### Notes for the reviewer

- Two order templates deliberately do **not** inherit their neighbouring band's guard: there the
  guard sits on an inner text field of the *price* row while the product-number band is
  unconditional, so inheriting it would have hidden the text whenever prices are switched off.
- The invoice templates have **two** article bands (`!ishu` and the packing-material one), so their
  guard is the union of both — guarding on only the first would silently drop text above every
  packing-material row that does print.
- On the picking list the block prints once per picking candidate. That is the template's grain: its
  article row already prints once per candidate, so each printed row gets its own block.


backend/de.metas.cucumber/src/test/resources/de/metas/cucumber/features/report/freeTextAboveOrderLine.feature
```diff
@@ -0,0 +1,156 @@
+@from:cucumber
+@allure.label.epic:E0100_Sales
+@allure.label.feature:F00144_Free_Text_Above_Order_Lines
+@ghActions:run_on_executor2
```
metas-ts, line 4: make sure that we cover each jrxml-change

metas-ts: **Resolution (dev):** Adopting this. Two parts, because a rendering test structurally cannot reach most of the templates.

The band was added to **15** line-detail templates, but which one renders is decided by a resource-bundle property, not by data:

- `order`: `report_details_v2` is UNCONDITIONAL (the default, and what the cucumber + Playwright tests exercise); `report_details` and `report_details_hu` need `$R{details_product_overflow}=="Y"`, `report_details_hu_v2` needs `"N"`.
- `inout`: `report_details_v2` `"N"` (default, covered); `report_details` / `report_details_hu` `"Y"`; `report_details_hu_v2` `"N"`.
- `invoice`: `report_details_v2` and `report_details_hu` UNCONDITIONAL; `report_details` `"Y"`.
- `picking`: split on `$R{name_over_attributes}` `"Y"`/`"N"` across four templates.
- `pickingv2`: `report_details` UNCONDITIONAL.

So the `"Y"` variants are unreachable from a test without rebuilding a resource bundle — the existing document tests cover the defaults only.

**What I am adding: a structural test over all 15**, asserting per template that the field is declared, that exactly one band in `<detail>` carries it, that the band is *first* (so it prints above the article row), that it holds a single element, that its band-level guard carries the null/blank check, that its `x`/width match **that template's own** content row (not a fraction of `pageWidth` — the same 596pt page has content rows ending at 573, 545 and 457), and that its guard covers **every** article band in that template.

That last invariant is not theoretical: this branch shipped two defects of exactly this kind, and both got past human review *and* the first CI run — four bands narrower than their own row (so the promised full-width block was ~49% wide on the default order confirmation), and seven bands missing the guard of the article row above which they sit (one variant silently dropped the text above packing-material invoice rows that do print). The test is required to prove it discriminates by mutation — narrow a band, strip a guard clause, move the band below the article row — before it counts as done.

I will follow up on this thread with the commit URL once it is verified.

metas-ts: **Resolution (dev):** Done, in two commits.

**Structural coverage of all 15 templates** — https://github.com/metasfresh/metasfresh/commit/ffabec30a8e00b9ad4907f9ca224abb8b7196350
`DescriptionAboveLineBandTest` walks `src/main/jasperreports/de/metas/docs/sales/**/report_details*.jrxml`, keeps every template carrying the field (asserting the count is 15, and listing the discovered paths when it is not, so a template gaining or losing the field is a deliberate edit), DOM-parses each, and collects every violation into one failure. Per template it asserts: the field is declared; exactly one band in `<detail>` carries it; that band is **first** (so it prints above the article row); it holds exactly one element; its **band-level** guard carries the null and `trim().isEmpty()` checks; its `x` and right edge match **that template's own** content row within 5pt; and its guard references every field used by the band-level guard of every article band in the same `<detail>`.

Proven to discriminate, each mutation reverted afterwards:

| mutation | result |
|---|---|
| `order/report_details_v2` block back to `x="59" width="290"` | invariant 6 fails, naming both the wrong `x` and *"224pt short of the right edge of this template's own content row (x=573)"* |
| `\|\| "Y".equals($F{isprintwhenpackingmaterial})` stripped from `invoice/report_details` | invariant 7 fails, naming *"article band 3 of 6 (9 text fields)"* and the missing field |
| `pickingv2/report_details` detail bands swapped | invariant 3 fails: *"it is band 2 of 2"* |
| guard moved from `<band>` onto the `<reportElement>` | invariant 5 fails — **a regex/substring test would have passed here**, which is why this is DOM-parsed |

Two honest limits: invariant 1 cannot be mutation-proven because the `jasperreports:jasper` goal fails the build first (`Field not found : descriptionaboveline`), so the build is the real gate there; and invariant 7 identifies an article row by "≥5 text fields", which separates cleanly today (article rows carry 8–13, the widest non-article row 3–5) but would misclassify a future 4-column article row — loudly, and via one named constant.

**The rendered PDFs now reach the Allure report** — https://github.com/metasfresh/metasfresh/commit/3a01375dc040c03042c23255e74b1cee3fd4b9ef
`AD_Archive_StepDef` attaches each asserted document via `Allure.addAttachment(..., "application/pdf", ...)` at the point where it already holds the bytes, deduplicated per `AD_Archive_ID` (several assertion steps in one scenario read the same document and would each attach another copy). Verified on a local run: exactly **4** `*-attachment.pdf` files with `%PDF` magic and `"type":"application/pdf"` recorded in the result json, named per archive id — 4 rather than ~14, which is the dedupe working.

Local run after both commits: `run-cucumber-feature.sh freeTextAboveOrderLine flaming_merlin_uat` → all 4 scenarios passed, exit 0.

*(Edited to fix the second commit link — the first version of this comment carried a mistyped SHA and pointed nowhere.)*

metas-ts: **Resolution (dev):** Your concern was better-founded than my first answer to it. That answer reported a structural test over **15** templates and treated the set as complete. It was not: **eight further live templates read a report source this branch had already patched — all `SELECT *`, so the new column was reaching them — and carried no band at all**, so the free text would silently not have printed on those documents.

All eight now carry it (https://github.com/metasfresh/metasfresh/commit/6860b3ed8e113a2f199b8dfacb69e0bd425683e6), taking the guarded set from 15 to **23 across two maven modules**:

| template | report source |
|---|---|
| `inout_concrete_adr/report_details_hu` | `Docs_Sales_InOut_Details_HU` |
| `inout_without_prices_customer_specific/report_details_hu` | `Docs_Sales_InOut_Details_HU` |
| `alternate_inout_2/report_details_hu` (module `metasfresh-dist`) | `Docs_Sales_InOut_Details_HU` |
| `individual_inout/report_details` | `Docs_Sales_InOut_Details` |
| `inout_without_prices_customer_specific/report_details` | `Docs_Sales_InOut_Details` |
| `inout_org_data_right/report_details` | same function — unquoted `Inout` folds to it |
| `invoice_org_data_right/report_details` | `Docs_Sales_Invoice_Details` |
| `order_org_data_right/report_details` | `Docs_Sales_Order_Details` |

Liveness was checked per template rather than assumed: the three HU ones sit under the data guard `ishidepackingmaterialinshipmentprint = 'N'`, the other five under bands with no `printWhenExpression` at all. Geometry and guard were derived per template — all eight land flush with their own content row, two inherit `new Boolean ( $F{qtyentered}.floatValue() != 0 )` verbatim, and six get the null/blank check only because their article band prints unconditionally.

Two templates are deliberately still without it, and for reasons rather than by omission: `invoice/report_details_hu`, whose parent band guard is the literal `new Boolean (false)` in `invoice/report.jrxml` (re-enabling a deliberately disabled template is not this change's business), and `individual_inout/report_details_hu`, which reads `Docs_Sales_InOut_Details_HU_Individual` — a different function this branch does not touch, so it would need the SQL widened first.

The test that guards this was also hardened after review found it passing on defects it was meant to catch (https://github.com/metasfresh/metasfresh/commit/bfbcf144531113156db995b428f99ea7dac15e44, https://github.com/metasfresh/metasfresh/commit/5ae2f24af0bf592e62ab127bcc5093d6ed2eca41, https://github.com/metasfresh/metasfresh/commit/a68e1d1f96b2c37b40b04ff77e66d41537903aef). The ones worth naming, because each was a way the old version said "covered" when it was not:

- the article-row discriminator counted `<textField>`s and required 5, but four article rows carry only three (a column can be a `staticText` or an image) — so the guard-coverage invariant was silently skipped on 4 of 15 templates. It now counts distinct `x` positions: article rows measure 3–12, every other band 0 or 1.
- the width invariant computed the content row's extent *including our own element*, so both of its bounds held by construction. It now measures the other bands only, and bounds the block by the template's own `pageWidth` minus its right margin.
- the guard invariant compared field *name sets*, so inverting the invoice guard kept the set identical and passed — while dropping the text above every non-HU line. It now dispatches on how the article bands are guarded, and requires the union (a `||`) where they are alternatives.
- all of those checks are substring matches, so wrapping a correct guard in `new Boolean(false) && (…)` satisfied every one while the band never printed. That is now rejected — judged by negation parity, so `!(1 == 0)` (harmless) is accepted while `!(1 == 1)` is not.

Every one of those is mutation-proven double-sided: the new check fails on the mutation *and* the old one passed the same mutation. One limit stated rather than papered over: an inversion that preserves the disjunctive shape (`!$F{ishu} || X` → `$F{ishu} || X`) is still undetected; catching it needs a real expression evaluator, which a syntactic check cannot honestly claim.

PR #25916, opened 2026-09-10
Product Proposals — filter to products with delivery history

mrie04:
Adds a user-toggled flag filter to the sales-order *Produktvorschläge* overlay that restricts the
list to products the business partner has already had delivered.

### Behaviour
- New frequent-used `YesNo` filter in the overlay, captioned from a new `AD_Message`
  (`OnlyProductsWithDeliveryHistory` — DE "Nur Produkte mit Lieferhistorie", EN "Only products with
  delivery history"). It renders as a filter **button** whose dropdown panel carries the flag
  (`DocumentFilterInlineRenderMode.BUTTON`) — inline rendering is reserved for search fields, whose
  caption fits inside the input; a checkbox rendered inline gets no caption at all, because the
  Checkbox widget never renders one.
- **Off by default**: the overlay still opens with the full assortment; the salesperson opts in.
- The criterion is the row's existing *Tage vergangen* (`lastShipmentDays`) signal, i.e. **delivered**
  (on a purchase order: received), not merely ordered.
- Rows carrying a typed quantity stay visible when the filter is switched on, so no entered quantity
  is ever hidden — and the order lines created on DONE do not depend on the filter state.
- *Andere Produkte* is untouched: the order view gets its **own** filter descriptor, so that view
  gains no new parameter.
- Row ordering is unchanged.

### Not included, deliberately
No `AD_SysConfig` switch. The filter is visible and toggleable by the user, which makes a
global/per-org on/off flag redundant.

### Tests
- Playwright E2E (`product-proposal-delivered-filter.spec.js`) — filter present-but-off, narrowing,
  restoring, and a partner with no delivery history at all.
- Playwright E2E (`product-proposal-qty-survives-filter.spec.js`) — a typed quantity survives the
  filter and still reaches the order line.
- `ProductsProposalRowTest` — the row-matching predicate.
- `ProductsProposalRowsDataTest` — the unfiltered row set the order lines are built from.

### Also in this branch
`InlineFilterItem#handleApply` applied the locally tracked value instead of the one it was handed,
and returned `undefined` where the Checkbox widget expects a thenable — so an inline `YesNo` filter
applied a stale value and broke the widget's patch contract. The fix is independent of the filter
above (which no longer renders inline) and still applies to every inline filter the Application
Dictionary can configure via `AD_Field.IsShowFilterInline`; no spec in this branch exercises it.

### DIFF
--- a/backend/de.metas.business/src/main/sql/postgresql/system/20-de.metas.business/5823770_sys_product_proposal_delivered_filter_message.sql
+++ b/backend/de.metas.business/src/main/sql/postgresql/system/20-de.metas.business/5823770_sys_product_proposal_delivered_filter_message.sql
@@ -1,0 +1,28 @@
+-- IDs allocated from idserver.metas.de on 2026-09-10:
+--   AD_Message_ID 545833 (OnlyProductsWithDeliveryHistory filter label)
+
+INSERT INTO AD_Message (AD_Client_ID,AD_Message_ID,AD_Org_ID,Created,CreatedBy,EntityType,IsActive,MsgText,MsgType,Updated,UpdatedBy,Value)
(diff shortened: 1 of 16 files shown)


mrie04:
known flakies, merging after 3rd rerun

PR #25947, opened 2026-09-15
serviceprovider tests: enable unit-test mode before building the repositories

mrie04:
Test-only fix for an order-dependent failure in `de.metas.serviceprovider.base`.

### Symptom

CI job `test (java)` fails on the module with an empty surefire error; the surefire report shows one test class erroring in its constructor with `IllegalStateException: JUnit test mode shall be enabled` (seen on https://github.com/metasfresh/metasfresh/actions/runs/34955467414, class `IssueServiceTest`). Locally the same module run errors in `FailedTimeBookingRepositoryTest` instead, and both classes fail deterministically when run alone.

### Cause

`IssueServiceTest` and `FailedTimeBookingRepositoryTest` build their repositories in field initializers, which run before `@BeforeEach init()` calls `AdempiereTestHelper.get().init()`. Both constructions go through a `newInstanceForUnitTesting()` that asserts unit-test mode. So the class that runs **first** in the surefire fork fails, and every later class passes because an earlier one already enabled the mode. Surefire's class order differs between machines, which is why the failing class changes between runs and why the module is green most of the time.

### Fix

The repositories are created inside `init()`, after the mode is enabled — the pattern the other tests in the codebase already use. No production code touched.

### Verification

Local, offline: before the change each class alone (fresh JVM) errors with the message above (`IssueServiceTest` 1 error, `FailedTimeBookingRepositoryTest` 3 errors); after the change both pass alone and the whole module passes (17 tests, 0 failures, 0 errors).


mrie04:
@Mergifyio copy flaming_merlin_release

mrie04:
@Mergifyio copy intensive_care_hotfix

PR #25953, opened 2026-09-16
Text lines in documents — sales order

metas-ts:
Adds a generic way to put free text lines between the article lines of a sales order, and prints them on
the order confirmation.

## What a user gets

A **Text lines** button on the sales order's line tab opens a modal listing the order's article lines and
its text lines in one merged, ordered list. In it the user can add a text line above a selected row, edit
its text in place, delete it, and move it up or down. The text prints on the order confirmation as a line
carrying its text alone — no article number, quantity, unit or price — at the position the user put it.

A text line holds multi-line text, blank lines included; an empty text line prints as a single blank line.

## Design decisions worth knowing before reading the diff

**`C_Doc_TextLine` deliberately ships no window, tab, field or menu entry.** Its user interface is the
Java-backed modal named above, reached from the button on the order's line tab. The only `AD_Window`
reference in the migrations binds that launcher process to the *existing* sales-order window; nothing
creates a window of its own. This is by design, not an omission.

**Positions share one ordering space with the article lines.** `C_Doc_TextLine.Line` is `numeric(10,4)`
against `C_OrderLine.Line`'s `numeric(10,0)`, and a text line is inserted at the midpoint between its
neighbours. Article lines are never renumbered — not on insert, not on move, not on delete. When the gap
between two neighbours is exhausted at four decimal places the insert is refused rather than silently
producing a duplicate position.

**The report function gained a text branch.** `Docs_Sales_Order_Details` is now a `UNION ALL`: the
existing article branch, plus a branch over the new table. Text rows carry the document-level columns —
the amount and price patterns, the partner's discount-printing and tax-printing flags, and the
document-type visibility flags — and null every column that describes an article. Two of the report's
column headers previously used a nullable article column as a stand-in for a document-level flag; they now
read the flag directly, so they survive a page that begins with a text line.

## Tests

- Playwright coverage of the modal: add, edit, delete, move, and the ordering invariants.
- A cucumber scenario reproducing a real customer order end to end, asserting on the rendered PDF: the
  multi-line block with its blank line, the heading between two article groups, and the article lines'
  own numbering left untouched. Every text-line guard in the template is mutation-tested — removing any
  one of them makes the scenario fail.
- Three further scenarios: a trailing text line, the character set and literal markup, and a
  no-text-lines order proving no band was inserted where the feature could have inserted one.
- Unit tests for the position arithmetic, including the gap-exhaustion refusal and the concurrent-insert
  case.

## What is still to come on this branch

The order checkup **does** print text lines — it landed after this description was first written, with
cucumber coverage of both its document types.

Still outstanding, and in scope for this branch: copying text lines onto the shipment, the delivery note's
own report branch and template bands, cucumber coverage of the partial-delivery carry rules, and the
message translations.

The **invoice** is deliberately out of scope and stays that way; a scenario here asserts it prints no text
lines.

## Cross-repo impact — worth knowing before this merges

`Docs_Sales_Order_Details` is one of the report functions that a customer repository can ship its own copy
of, and a customer copy shadows core's. A code search across the organisation finds twelve repositories
carrying their own `Docs_Sales_Order_Details.sql` and one carrying its own
`Docs_Sales_OrderCheckup_Details.sql`.

**Those installations do not get this feature from this pull request.** Their function has no text-line
branch, so a text line entered in the modal would simply not appear on their order confirmation — it would
not break their document, it would be silently absent. Each needs the branch ported into its own copy.

The customer this work was written for is **not** among them: it ships neither function, so it takes both
from core. Verified by search, and corroborated by its own build going green against this branch once it
was pointed at core's build.

Flagging it here rather than fixing it: which of those repositories should receive the port, and when, is
not this pull request's call.


backend/de.metas.business/src/main/java/de/metas/doctextline/DocTextLineRepository.java
```diff
@@ -0,0 +1,260 @@
...
+ * Repository Cluster: DocTextLineRepository
+ * <p>
+ * Owns the position arithmetic of a midpoint insert-above and the scope default derived from it. Both operate
+ * on plain values supplied by the caller — this repository never queries article-line tables (C_OrderLine /
+ * M_InOutLine), which it does not own (see {@link InsertAboveRequest}).
+ */
+@Repository
+public class DocTextLineRepository
+{
+	private static final int LINE_SCALE = 4;
```
metas-ts, line 37:
```suggestion
	/** "4" allows up to 9999 text lines between any two document lines. */
	private static final int LINE_SCALE = 4;
```

if my proposed javadoc is factually correct, please add it.

backend/de.metas.business/src/main/java/de/metas/doctextline/DocTextLineRepository.java
```diff
@@ -0,0 +1,260 @@
...
+		}
+		if (lowerPosition == null)
+		{
+			return FIRST_POSITION_IN_EMPTY_DOCUMENT;
+		}
+		return lowerPosition.add(BigDecimal.ONE.setScale(LINE_SCALE));
+	}
+
+	/**
+	 * The scope-driving predicate {@link InsertAboveRequest#isArticleLineExistsBeforeReferencePosition()} needs:
```
metas-ts: make this javadoc way shorter, if you think it's needed at all

backend/de.metas.business/src/main/java/de/metas/doctextline/DocTextLineRepository.java
metas-ts: **Resolution (dev):** Checked it rather than adopting it, and it is not factually correct for this code — so I have not added it as proposed.

`insertAbove` places a new line at the **midpoint** of the gap it lands in:

```java
previousPosition.add(referencePosition).divide(BigDecimal.valueOf(2), LINE_SCALE, RoundingMode.HALF_UP)
```

Two consequences:

- 9999 is the count of *distinct representable values* between two adjacent integer positions (`10.0001` … `10.9999`). That number is only reachable if positions can be placed arbitrarily.
- Midpoint insertion cannot reach them. Each insert at the same spot halves the remaining gap, so between two neighbours you get roughly **log2(10000) ≈ 13** successive inserts before the rounded midpoint collides with a neighbour — which is exactly what the gap-exhaustion guard a few lines below refuses.

So the guarantee the constant actually buys is "about 13 inserts at one spot", not 9999.

Replacing it with this, which says what the scale does and points at the guard rather than implying capacity we do not have:

```suggestion
	/** Four decimals, and inserts are midpoints: a gap survives ~13 successive inserts at the same spot before {@link #computeInsertAbovePosition} refuses it. */
	private static final int LINE_SCALE = 4;
```

Happy to word it differently if you would rather it state the representable-slot count and the midpoint limit separately.

backend/de.metas.business/src/main/java/de/metas/doctextline/DocTextLineRepository.java
metas-ts: **Resolution (dev):** Agreed — 23 lines of javadoc on a one-line predicate, and most of it is design history rather than anything a caller needs. Cutting it to this:

```java
/**
 * Whether any article line sits before {@code referencePosition} — the predicate that decides a new
 * text line's default scope.
 *
 * @param articleLinePositions must contain every article-line position strictly less than
 *        {@code referencePosition}; further positions are harmless, and order does not matter.
 */
```

The `@param` sentence is the one part I kept deliberately: both callers depend on it (the modal passes only the rows before the reference, the cucumber step passes all of them), so it is a precondition rather than commentary.

What went out: why the method was extracted, which callers share it, and a paragraph arguing that comparing by position is safe because a tie is unreachable. The last one is genuinely true but it explains a case that cannot occur — if that is ever worth recording it belongs in a test name, not in the javadoc of the method it does not affect.

backend/de.metas.business/src/main/java/de/metas/order/IOrderDAO.java
```diff
@@ -53,6 +53,27 @@ public interface IOrderDAO extends ISingletonService
...
+	 * it. Holding it says nothing about the order record itself; this method does not modify it.
+	 * <p>
+	 * <b>Must be called inside a transaction</b>, and the reads and writes it is meant to protect must run in
+	 * that same transaction -- PostgreSQL holds a row lock only until the transaction that took it ends, so
+	 * out of transaction this call blocks nobody and silently protects nothing.
+	 * <p>
+	 * Fails rather than returning quietly when there is no such order: a caller that believes it is holding
+	 * the lock while nothing is locked is the one outcome this method must never produce.
+	 */
+	void lockByIdForUpdate(@NonNull final OrderId orderId);
```
metas-ts: i'll need to understand what this is about. what is conretely locked, and why. please explain it in the current session

backend/de.metas.business/src/main/java/de/metas/doctextline/InsertAboveRequest.java
```diff
@@ -0,0 +1,36 @@
...
+ * default. The merged sequence spans a table this repository does not own (article lines), so the caller —
+ * which already loaded that merged view — supplies the two positions and the scope-driving fact as plain
+ * values; this repository never queries article lines itself.
+ */
+@Value
+@Builder
+public class InsertAboveRequest
+{
+	@NonNull DocTextLineDocumentRef documentRef;
+	@Nullable String textLine;
```
metas-ts: why is textline nullable?

backend/de.metas.business/src/main/java/de/metas/doctextline/TextLineScope.java
```diff
@@ -0,0 +1,42 @@
...
+	@NonNull private static final ReferenceListAwareEnums.ValuesIndex<TextLineScope> index = ReferenceListAwareEnums.index(values());
+
+	@NonNull private final String code;
+
+	@JsonCreator
+	@NonNull
+	public static TextLineScope ofCode(@NonNull final String code) {return index.ofCode(code);}
+
+	@Nullable
+	public static TextLineScope ofNullableCode(@Nullable final String code) {return index.ofNullableCode(code);}
```
metas-ts: Why do we need this method and `optionalOfNullableCode`?
Is it a confention that we must follow, or is it an archunit-rule?

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/order/C_Order_StepDef.java
```diff
@@ -647,6 +654,115 @@ public void generate_PO_from_SO_invoked(@NonNull final DataTable dataTable)
 
 	}
 
+	/**
+	 * Generates the order's {@code C_Order_MFGWarehouse_Report} rows -- the "Bestellkontrolle" -- via the
```
metas-ts: way shorter javadoc please.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/order/C_Order_StepDef.java
```diff
@@ -830,6 +946,16 @@ public void validate_created_order(@NonNull final DataTable table)
...
+	 *   <li>{@code DescriptionBottom} – text printed at the end of the order document. Deliberately NOT a
+	 *       creation-time column ({@code metasfresh contains C_Orders:} has no such field): setting
+	 *       {@code C_DocTypeTarget_ID}/{@code C_BPartner_ID} at creation re-derives it from the doc type's
+	 *       {@code DocumentNote} ({@code C_Order} model interceptor {@code updateDescriptionFromDocType}), so
+	 *       a value set at creation time is silently overwritten before the insert. This is only safe when
+	 *       the SAME update row does not also set {@code DocBaseType}/{@code DocSubType} -- doing so resolves
+	 *       a new {@code C_DocTypeTarget_ID} (see below) and re-fires the interceptor's {@code
+	 *       ifColumnsChanged} guard, overwriting {@code DescriptionBottom} again. Put {@code
+	 *       DescriptionBottom} in its own update row, after any row that changes the doc type, matching the
+	 *       real WebUI edit of the order's own field.</li>
```
metas-ts:
```suggestion
	 *   <li>{@code DescriptionBottom} – text printed at the end of the order document. Deliberately NOT a
	 *       creation-time column ({@code metasfresh contains C_Orders:} has no such field): setting
	 *       {@code C_DocTypeTarget_ID}/{@code C_BPartner_ID} at creation re-derives it from the doc type's
	 *       {@code DocumentNote} ({@code C_Order} model interceptor {@code updateDescriptionFromDocType}), so
	 *       a value set at creation time is silently overwritten before the insert.
```

i think this is more than enough info

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/productplanning/PP_Product_Planning_StepDef.java
```diff
@@ -78,8 +78,17 @@ public class PP_Product_Planning_StepDef
...
+	/**
+	 * Every {@code PP_Product_Planning} this step-def touched (created OR updated) this scenario, so
+	 * {@link #deactivateTouchedProductPlannings()} can deactivate all of them afterwards. A no-{@code M_Product}
+	 * planning matches ALL products and would otherwise poison every later scenario's material planning; a
+	 * product-specific one is just as dangerous by a different mechanism -- {@code metasfresh contains M_Products}
+	 * upserts by {@code Value}, so a fixture that reuses a literal-{@code Value} product resolves the SAME
+	 * {@code M_Product_ID} on every local run, and a left-active planning for it then makes the NEXT run's
+	 * fixture-removal mutation (e.g. "delete the PP_Product_Plannings row and expect the report to disappear")
+	 * pass for the wrong reason: a leftover row, not the current row, still satisfies the production code.
+	 */
```
metas-ts, line 84:
```suggestion
	/**
	 * Every {@code PP_Product_Planning} this step-def touched (created OR updated) this scenario, so
	 * {@link #deactivateTouchedProductPlannings()} can deactivate all of them afterwards.
	 */
```

backend/de.metas.business/src/main/java/de/metas/order/IOrderDAO.java
metas-ts: **Resolution (dev):** Explaining it here as well as in the session.

**What is locked:** one row — the `C_Order` row of the order the modal is open on. The query selects the key column only and discards it; the lock the `SELECT` leaves behind is the whole point:

```sql
SELECT C_Order_ID FROM C_Order WHERE C_Order_ID = ? FOR NO KEY UPDATE
```

Not the order lines, not the text lines, and nothing about the order record is written. The `C_Order` row is being used as the place two writers on the same document can agree to queue.

**Why anything is locked at all:** the modal's structural writes — insert-above, move up/down, delete — are read-derive-write sequences over an ordering that spans two tables. `C_OrderLine.Line` and `C_Doc_TextLine.Line` share one numeric space, and a new text line's position is the midpoint between the two neighbours the writer just read. Two writers on the same order that each read the same neighbours compute the same midpoint and both persist it. That is a duplicate `Line` value, and nothing rejects it: there is no unique constraint on `(C_Order_ID, Line)`, and the gap-exhaustion guard inside `DocTextLineRepository.insertAbove` only compares the midpoint against its own two inputs, never against one another request computed concurrently. The result is a document whose ordering is ambiguous — which is exactly what the feature is for.

So the thing being protected is an ordering that belongs to no single row, which is why the lock is taken on the document's own record rather than on any row being written.

**Why not just the in-process lock:** `DocTextLinesRows` does hold a JVM monitor in front of this, but that is not the guarantee and must not be read as one. A second webapi instance serving the second browser tab shares no monitor with the first; it would re-derive the same ordering and persist into it. What the in-process lock buys is cheaper waiting — a second local writer waits before opening a transaction and taking a pooled connection, instead of sitting idle in a lock wait inside PostgreSQL — plus it is the only serialisation the unit-test harness has, which has no database at all.

**Why `FOR NO KEY UPDATE` rather than `FOR UPDATE`** (this changed in the last round, it was `FOR UPDATE` when the comment was written): the two are equally exclusive *against each other*, so callers of this method serialise either way. But `FOR UPDATE` additionally conflicts with the `FOR KEY SHARE` that PostgreSQL takes on a parent row whenever a concurrent transaction writes a row referencing it — an order line, for instance — so it would make ordinary order-line work queue behind a text-line modal that has nothing to do with it. `FOR NO KEY UPDATE` does not conflict with `FOR KEY SHARE`. Holding it says nothing about the order record itself, which is honest here: this method does not modify it.

**Scope and lifetime:** one caller, `DocTextLinesRowsLoader.lockDocumentForUpdate()`, reached from `DocTextLinesRows.withDocumentLocked(...)`. The lock lives until the surrounding transaction ends — the quick action's `doIt`, which `ProcessExecutor` has already wrapped in one — which is why the javadoc insists the protected reads and writes run in that same transaction: out of transaction the call blocks nobody and silently protects nothing. Ordinary field edits (editing a line's text or its scope) take neither this lock nor the in-process one and are unaffected.

**Why it lives on `IOrderDAO`:** `C_Order` is that DAO's table. The view layer asking it to lock the order keeps the rows holder from reaching into a table it does not own. And it throws rather than returning quietly when there is no such order, because a caller that believes it is holding the lock while nothing is locked is the one outcome this method must never produce.

backend/de.metas.business/src/main/java/de/metas/doctextline/InsertAboveRequest.java
metas-ts: **Resolution (dev):** It should not be. Tightening it to `@NonNull`.

I went looking for the null that justifies it and there isn't one on this path. Both callers pass non-null:

- `WEBUI_DocTextLines_InsertAbove` (the quick action, the only production caller) passes `""` — the modal creates an empty row and the user types into it afterwards.
- the cucumber step declares its text parameter `@NonNull`.

So `@Nullable` here only widens the contract; it prevents no scenario that exists.

Where nullability *is* real, and stays:

- `C_Doc_TextLine.TextLine` is `VARCHAR(2048)` with no `NOT NULL`, and a stored null is genuinely reachable — but through the **edit** path, not the insert path: clearing the field in the modal sends a JSON null, `toUserChangeRequest` reads it with `getValueAsString(null)`, and `updateTextAndScope(id, null, scope)` persists it. So `DocTextLine.textLine` and that method's parameter keep `@Nullable`, each with a caller that actually produces null.

Net effect is that "no text" has one representation at the creation boundary (`""`) instead of two, and the remaining `@Nullable`s each name a real source. An empty text line still prints as a single blank line either way — the requirement is unaffected.

One note on sequencing: `InsertAboveRequest` is in the blast radius of the shipment-copy work still in flight on this branch, so the edit lands with that round rather than as a separate commit on top of it.

backend/de.metas.business/src/main/java/de/metas/doctextline/TextLineScope.java
metas-ts: **Resolution (dev):** Neither — no ArchUnit rule, no convention that requires them. They are copied boilerplate and both are dead. Removing them.

What I checked:

- **Callers:** zero, for both. `grep` across `backend/` finds `TextLineScope.ofNullableCode` and `TextLineScope.optionalOfNullableCode` only at their own declarations.
- **Reflective use:** this is the one way an apparently-unused accessor on a `ReferenceListAwareEnum` can still be load-bearing, so worth ruling out explicitly. `ReferenceListAwareEnums.createReferenceListAwareDescriptor` looks up exactly two members by name — `ofCode(String)` and `values()` — and throws if `ofCode` is missing. Nothing reflects on the nullable variants. That matters here because `DocTextLinesRows` does reach the enum reflectively, via `JSONDocumentChangedEvent.getValueAsEnum(TextLineScope.class)` → `ReferenceListAwareEnums.ofNullableCode(valueStr, enumType)` — note that resolves through the **static utility**, which calls the enum's `ofCode`, not through the enum's own `ofNullableCode`. So `ofCode` is load-bearing twice over (reflectively, plus three direct call sites) and the other two are not reachable at all.
- **ArchUnit:** nothing in `de.metas.architecture-tests` mentions `ReferenceListAwareEnum` or these method names.
- **The coding rule** (`docs/coding-rules/java-general.md` § 8) prescribes the dedicated enum and `isXXX()` helpers. It never asks for the nullable factories; it mentions `ofNullableCode` only in a review-rule about not wrapping it in a redundant blank check.

So the enum keeps `ofCode` (needed) and loses both others.

While in there, the same § 8 asks for `isXXX()` helpers instead of `==` comparisons, and there is exactly one such comparison on this branch — `TextLineShipmentCopier`'s `textLine.getScope() == TextLineScope.Following`. Adding `isFollowing()` and using it there, so the enum ends up carrying the accessors the rule actually asks for instead of the ones it doesn't.

That file is being touched by the shipment-copy work still in flight, so both edits land with that round rather than as a separate commit on top of it.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/order/C_Order_StepDef.java
metas-ts: **Resolution (dev):** Cut from 30 lines to 6 in https://github.com/metasfresh/metasfresh/commit/acdb31949b6aac0bbb7c09004d27819e662c4586.

What stays is what a caller of the step cannot work out for themselves: which AD_Process runs, the two
identifiers the generated rows are registered under (`<orderIdentifier>_checkup` for the `PL` row,
`<orderIdentifier>_checkup_WH` for the `WH` row), that the `WH` row is not guaranteed, and that the step
runs under the `"WebUI"` role.

What went out: why the process is invoked by `AD_Process.Value` instead of by class (a module-boundary
argument that belongs to the commit, not to every future reader of the step), why the `PL` row needs no
`PP_Product_Planning` fixture, and the paragraph on what a caller in a different feature would have to
change. None of it changes how the step is used.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/order/C_Order_StepDef.java
metas-ts: **Resolution (dev):** Applied as suggested, in https://github.com/metasfresh/metasfresh/commit/acdb31949b6aac0bbb7c09004d27819e662c4586.

The dropped half was a use-instruction — put `DescriptionBottom` in its own update row, after any row that
changes the doc type — rather than an explanation, and the sentence that stays already gives the reason
(the interceptor re-derives the field from the doc type). Worth noting that the scenario on this branch
does exactly that, so the rule is demonstrated by the only caller rather than asserted in prose.

backend/de.metas.cucumber/src/test/java/de/metas/cucumber/stepdefs/productplanning/PP_Product_Planning_StepDef.java
metas-ts: **Resolution (dev):** Applied as suggested, in https://github.com/metasfresh/metasfresh/commit/acdb31949b6aac0bbb7c09004d27819e662c4586.

The removed paragraph was the argument for why product-specific plannings are cleaned up too and not only
the no-`M_Product` ones — that belongs to the commit that widened the field from `genericPlanningIds` to
`touchedPlanningIds`, and the rename plus `deactivateTouchedProductPlannings()` already say the same thing
in the names.

backend/de.metas.business/src/main/java/de/metas/doctextline/DocTextLineRepository.java
```diff
@@ -0,0 +1,246 @@
...
+		{
+			queryBuilder.addEqualsFilter(I_C_Doc_TextLine.COLUMNNAME_C_Order_ID, documentRef.getOrderId());
+		}
+		else
+		{
+			queryBuilder.addEqualsFilter(I_C_Doc_TextLine.COLUMNNAME_M_InOut_ID, documentRef.getInOutId());
+		}
+
+		return queryBuilder
+				.orderBy()
```
metas-ts, line 211: i think this can be an ImmutableList. Be explicitly immutable whereever you not's need to alter the thing in question

backend/de.metas.business/src/main/java/de/metas/order/IOrderDAO.java
metas-ts: i guess i now understand it. two problems:

1. so if one user open the the text-line-modal on a particular order and and then another user open it as well, the 2nd user's metasfresh will seem to hang. 

2. we already have a gneric locking-support. See `de.metas.lock.api.ILockManager`. we cannot have a similar-but-different thing for each kind of document.

Suggestions:
- use the generic locking mechanisms that we have. if it's not feasible, consult with me on how to improve it
- if an order is already locked for textline-editing, then throw a userfriendly error. Check if we already have the possiblility to show it as a confirm-dialog. If the don't have that, then a toeast is also good enough

backend/de.metas.business/src/main/java/de/metas/doctextline/DocTextLineRepository.java
metas-ts: **Resolution (dev):** Done, and applied as the general rule rather than just here — https://github.com/metasfresh/metasfresh/commit/189a5789284b1f8eaa2e1e7e5a1b3d0ffb13fc1f

`getByDocument` now builds and returns an `ImmutableList<DocTextLine>` (`ImmutableList.toImmutableList()` in place of `Collectors.toList()`).

Swept the rest of the feature for the same thing and found four more methods that already *returned* an immutable value but only *declared* the interface — the type was hiding what the code already guaranteed:

- `DocTextLinesView.getAdditionalRelatedProcessDescriptors()` → `ImmutableList` (the field is one)
- `DocTextLinesViewFactory.getRelatedProcessDescriptors()` → `ImmutableList` (it is an `ImmutableList.of(...)`)
- `DocTextLinesRows.getDocumentId2TopLevelRows()` → `ImmutableMap` (it is a `Builder.build()`)
- `DocTextLinesRow.getViewEditorRenderModeByFieldName()` → `ImmutableMap` (the field is one)

Two mutable collections stay mutable, because they are the "you do need to alter the thing" case — both are local accumulators that exist to be written to and are already frozen on the way out: `DocTextLinesRows.updateRowIds`'s `ArrayList` copy (mutated by the caller's consumer, then `ImmutableList.copyOf`-ed back into the field) and `TextLineShipmentCopier`'s `currentBlock` while a block is being gathered.

One cost worth naming, since it cost me a confusing red locally: widening a public return type from `List` to `ImmutableList` is source-compatible but **not** binary-compatible, so a stale incremental build links the old signature and dies with `NoSuchMethodError: ...getByDocument(...)Ljava/util/List;`. A clean rebuild fixes it and CI always builds clean; no external caller exists, since the method is new on this branch. Verified after a clean build: `DocTextLineRepositoryTest` 17/17, `DocTextLineCopyTest` 13/13, `DocTextLinesViewTest` + `DocTextLinesQuickActionsTest` 65/65.

Separately, while merging your accepted suggestion: the `LINE_SCALE` javadoc had already been applied by hand in the fix round, so accepting it on top left the field carrying the line twice. Removed the duplicate in https://github.com/metasfresh/metasfresh/commit/515e9605beef01c66a9a8cfb18f403b94ab06573 — the surviving copy is the wording you accepted, unchanged.

backend/de.metas.business/src/main/java/de/metas/order/IOrderDAO.java
metas-ts: **Resolution (dev):** Investigated both points. You are right on (2) and I am switching to `ILockManager`; on (1) the premise is not quite what the code does today, which changes what we are fixing — details below, and there is one genuine design fork I would like your call on, which is the consult you offered.

**(1) When the lock is actually held.** Opening the modal does not lock anything. `DocTextLinesRowsLoader.load()` only reads; `lockDocumentForUpdate()` is a separate method, called from `DocTextLinesRows.withDocumentLocked(...)`, which wraps **one structural write** — insert-above, move, delete — and the lock dies with that write's transaction. Editing a line's text or its scope takes no lock at all. So a second user opening the modal on the same order sees it open normally, and two users only meet if they click a structural action at the same moment, for the duration of one quick-action transaction.

What is real in your concern, though, is that the wait is **unbounded**: PostgreSQL has no timeout there, so if anything ever does hold that row longer than expected, the second user's request waits with no feedback and no way out. "Rare and short" is not a design; a refusal is. So the fix is worth doing on its own merits, not only for the mechanism's sake.

**(2) `ILockManager` — feasible, and it gives us the friendly error for free.** Checked the implementation rather than assuming:

- Acquisition is a single `INSERT INTO T_Lock` whose conflict path is `DBUniqueConstraintException` (`SqlLockDatabase.lockRecord`), so mutual exclusion is atomic — no check-then-act race. That was my one worry about swapping a real row lock for an advisory one, and it does not apply.
- `setFailIfAlreadyLocked(true)` turns the collision into a `LockFailedException` instead of a wait — exactly the refuse-don't-hang behaviour you are asking for, and `LockFailedException.getExistingLocks()` carries the owner, so the message can say who holds it.
- `setAutoCleanup(true)` covers the crash case.

One semantic difference worth naming, because it is the thing that makes the rest of the design decision: the `INSERT` runs with `ITrx.TRXNAME_None`, i.e. **outside** the caller's transaction. A `T_Lock` lock is therefore *not* released when the transaction ends — it lives until something unlocks it. For a per-write lock that means a disciplined try/finally; for a longer-lived lock it means we own the release story.

**On surfacing it:** this feature already refuses this way — `DocTextLinesRows.rowNoLongerExists` throws an `AdempiereException` with a plain-language message out of the quick action, which the WebUI shows as an error notification. So the refusal needs no new UI mechanism; I would add a sibling of that method. If you want a confirm-style dialog rather than the notification, say so and I will look at what the quick-action layer supports — the notification is what the feature's existing refusals already use, so it is the consistent default.

**The fork I would like your call on** — what the lock's scope should be:

- **(a) Per structural write** (what I am implementing now): acquire around each insert/move/delete, release straight after. Two users can both have the modal open and both edit; they are only refused if they act *at the same instant*, and the retry is immediate. No stale-lock problem worth speaking of.
- **(b) For as long as the modal is open**, which is closer to how you phrased it ("if an order is already locked for textline-editing"): the second user is told up front that someone is editing this order's text lines, rather than discovering it on their first click. Honest cost: a modal abandoned by closing the browser leaves the order locked until auto-cleanup, so it needs a release story (explicit unlock on close, plus a timeout) that (a) does not.

I am building (a) now because it is the smaller change and strictly better than today in both respects — generic mechanism, and a refusal instead of an unbounded wait. If you want (b), it is a different conversation about lock lifetime and I would rather have it with you than guess.

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/Docs_Sales_Order_Details.sql
```diff
@@ -7,7 +7,7 @@ CREATE OR REPLACE FUNCTION de_metas_endcustomer_fresh_reports.Docs_Sales_Order_D
 
     RETURNS TABLE
             (
-                line                         numeric(10, 0),
+                line                         numeric(10, 4),
```
metas-ts: i guess this this `4` corresponds to the `LINE_SCALE` constant from further up. 
if so, please add a *short* comment here and there so they don't drift out of sync

metas-ts: **Resolution (dev):** Yes, same 4 — and there is a third place, so all three now point at each other. https://github.com/metasfresh/metasfresh/commit/230f8f99b0c0b36b40207eae4bd475244c3c5517

The chain is `C_Doc_TextLine.Line NUMERIC(10,4)` (the migration) → `DocTextLineRepository.LINE_SCALE = 4` (what the midpoint arithmetic rounds to) → this column, which has to be able to carry a text line's position unrounded or the report would collapse two adjacent text lines onto the same value.

In the SQL, one trailing comment on the column:

```sql
line                         numeric(10, 4), -- scale 4 must match C_Doc_TextLine.Line and DocTextLineRepository.LINE_SCALE
```

and one sentence added to `LINE_SCALE`'s javadoc naming the column and the table.

Checked the sibling report function while I was there: `Docs_Sales_OrderCheckup_Details` declares its `line` column as plain `numeric`, unconstrained, so there is no second literal to keep in sync — it takes whatever scale the value already has.

backend/de.metas.business/src/main/java/de/metas/order/IOrderDAO.java
metas-ts: **Resolution (dev):** Built and pushed — https://github.com/metasfresh/metasfresh/commit/6482a9a3960b845389f15d51a94d556fa5c5a6f1

`DocTextLineStructuralWriteLock` in `de.metas.business` (the module that already depends on `de.metas.async`; the view layer still goes through `DocTextLinesDocumentAccess#lockDocumentForUpdate`, so it needs no new dependency). It try-locks the document's own record with `setFailIfAlreadyLocked(true)` and `setAutoCleanup(true)`, and a collision throws a message addressed to the user — "This document's text lines are being edited right now. Please try again in a moment." — through the same error channel this feature's other refusals already use. `IOrderDAO#lockByIdForUpdate` and its implementation are deleted; `grep` finds no remaining caller anywhere.

Two things I want to state explicitly, because they are where this could have gone quietly wrong:

**The release is registered with the transaction, not run in a `finally`.** `T_Lock` rows are inserted with `ITrx.TRXNAME_None`, so an early release would free the document while this writer's own row is still uncommitted, and the next writer would compute its midpoint from the neighbours as they stood *before* the commit — the exact collision the lock exists to prevent. `ILock#asAutocloseableOnTrxClose` registers an `AFTER_CLOSE` release; `AbstractTrx.close()` commits first and fires it in its `finally`, so it lands after the commit. It also fires on rollback (`rollback()` clears the active flag, so `close()` skips the commit branch and still reaches `fireAfterClose`), which means a failed write cannot leave the document locked — there is a test for that, not just an argument.

**A deliberate deviation worth flagging:** I had specified a single stable lock-owner name. That would have been a bug — `ILock#unlockAll` releases BY OWNER, so two documents sharing one owner name would free each other's locks. It uses `LockOwner.newOwner("DocTextLineStructuralWriteLock")` (stable prefix + UUID), the same shape as `DocumentEngine#processIt`. Mutual exclusion is unaffected: the `t_lock_reference_singleowner` index is on `(ad_table_id, record_id)` and does not include the owner column.

Two honest limits:

- It is one acquisition per transaction, not re-entrant — a second acquire in the same transaction would meet its own predecessor and be refused. Each quick action performs exactly one structural write, which is what makes that safe today; it is documented as a trap rather than left to be discovered.
- The scope question from my previous comment is still open and still yours: this is the per-write lock. If you want the modal-lifetime variant instead, that is the lock-lifetime conversation, and I would rather have it than guess.

Tests: the two old Mockito assertions on the deleted method are replaced by stronger ones — instead of checking the *order of invocations*, the new test captures whether the document is actually locked at the moment the order lines are read. Plus a refusal test and a "a refused write does not leave the document locked" test. `DocTextLineRepositoryTest` 17/17, `DocTextLinesQuickActionsTest`+`DocTextLinesViewTest` 67/67, `DocTextLineCopyTest` 13/13, and the full `de.metas.ui.web.base` (1500) and `de.metas.business` (1034) suites green.

backend/de.metas.business/src/main/java/de/metas/doctextline/DocTextLineStructuralWriteLock.java
```diff
@@ -0,0 +1,95 @@
...
+	 * window to be reopened. The holder's identity is a generated lock-owner name, which tells the user
+	 * nothing they can act on, so it stays a parameter for the log.
+	 */
+	private static AdempiereException documentIsBeingEditedRightNow(
+			@NonNull final TableRecordReference documentRecord,
+			@NonNull final LockFailedException cause)
+	{
+		@Nullable final ImmutableList<ExistingLockInfo> existingLocks = cause.getExistingLocks();
+
+		return new AdempiereException("This document's text lines are being edited right now."
```
metas-ts: is this the exceltion thrown towards the user? if yes, then it must be user-friendly

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/Docs_Sales_Order_Details.sql
```diff
@@ -143,7 +171,89 @@ WHERE ol.C_Order_ID = p_record_id
   AND ol.isActive = 'Y'
   AND (COALESCE(pc.M_Product_Category_ID, -1) != getSysConfigAsNumeric('PackingMaterialProductCategoryID', ol.AD_Client_ID, ol.AD_Org_ID))
   AND ol.isHideWhenPrinting != 'Y'
-ORDER BY ol.line
+
+UNION ALL
+
+-- Free-text lines interleaved with the order's article lines (C_Doc_TextLine). This branch does
```
metas-ts: way too much text

backend/de.metas.business/src/main/java/de/metas/doctextline/DocTextLineStructuralWriteLock.java
metas-ts: **Resolution (dev):** Yes, it is thrown at the user — and you are right that it was not user-friendly, for a reason that is invisible at the call site. Fixed in https://github.com/metasfresh/metasfresh/commit/8655f329a5ae5509c4d91b7b956820e82a480a27

The sentence itself was plain language, but the `.appendParametersToMessage()` a line below it folds the `setParameter` values into the message that gets **built for display**. So on a collision the user would have read:

```
This document's text lines are being edited right now. Please try again in a moment.
document=TableRecordReference[C_Order,1000123], existingLocks=[ExistingLockInfo(...)]
```

I checked the rest of the feature rather than just this one, and every refusal on this branch had the same call — eight sites. All removed. The parameters stay on the exception via `setParameter`, so they are still there for the log and the AD_Issue; they just no longer reach the screen.

Three of those messages were also written for a developer rather than a user, so they are now sentences that say what to do instead:

- `"Article line rows are not editable"` → "Article lines cannot be edited here. Only text lines are editable in this window; an article line is changed on the order's line tab."
- `"Article line rows cannot be deleted or moved"` → "Only text lines can be deleted or moved here. An article line belongs to the order itself and is changed on the order's line tab."
- `"No row to move up into"` → "This line cannot be moved up any further -- there is no line beyond it in that direction."

One thing still outstanding, and I would rather name it than let it pass as done: these are hardcoded English strings, and this customer works in German. Making them `AdMessageKey`-based and translatable is Task 22 of the plan for this issue ("AD_Messages, translations"), which is where the AD_Message records and their migration scripts belong — I have not smuggled it in here. If you would rather have the translation now than at the end of the branch, say so and I will pull that task forward.

Tests after the change: `DocTextLineRepositoryTest` 17/17, `DocTextLinesViewTest`+`DocTextLinesQuickActionsTest` 67/67, clean Java-8 build green.

backend/de.metas.fresh/de.metas.fresh.base/src/main/sql/postgresql/ddl/functions/Docs_Sales_Order_Details.sql
metas-ts: **Resolution (dev):** Agreed — 22 lines for one `UNION ALL` branch. Cut to 9 in https://github.com/metasfresh/metasfresh/commit/8655f329a5ae5509c4d91b7b956820e82a480a27

What I kept is the one thing a reader cannot work out from the SQL itself: this branch deliberately does **not** reuse the article branch's `WHERE` clauses, because the packing-material category exclusion and `IsHideWhenPrinting` are both properties of an *article* line — copying them down would silently filter out every text line. That is a trap for the next person who "harmonises" the two branches, so it stays. Plus one sentence on the column shape: document-level values carried through, article-level values NULL.

What went out: the column-by-column enumeration (which the `SELECT` list right below already shows), the paragraph reasoning about `IsNetSumShown`/`IsTaxRateShown`, and the aside about there being no quantity predicate to exclude.

I gave the sibling `Docs_Sales_OrderCheckup_Details` the same treatment while I was in there — 10 lines to 5, keeping the carry rules, since those are the actual algorithm rather than commentary.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/doc_textlines/process/WEBUI_DocTextLines_Delete.java
```diff
@@ -0,0 +1,96 @@
...
+	 * Public so it is directly unit-testable without the {@code JavaProcess} parameter/view-loading machinery,
+	 * same rationale as {@link de.metas.ui.web.doc_textlines.process.WEBUI_DocTextLines_InsertAbove}.
+	 */
+	public static ProcessPreconditionsResolution checkDeletePreconditions(
+			@NonNull final DocTextLinesView view,
+			@NonNull final DocumentIdsSelection selectedRowIds)
+	{
+		if (selectedRowIds.isMoreThanOneDocumentId())
+		{
+			return ProcessPreconditionsResolution.rejectWithInternalReason("only one row can be selected");
```
metas-ts: i believe we have a canned reason for this case, so we can give a user-firendly feedback

metas-ts: **Resolution (dev):** There is — `ProcessPreconditionsResolution.rejectBecauseNotSingleSelection()`. Swapped in https://github.com/metasfresh/metasfresh/commit/f4cab79296c372f78591cea547bcce82aa6b1b6a

Checking it turned up something worse than the wording. `rejectWithInternalReason(...)` sets the resolution's `internal` flag, so the reason is **never shown** — the quick action simply appears unavailable and the user is told nothing. So the problem was not only that "only one row can be selected" reads like developer English; it was that nobody was ever going to read it. `rejectBecauseNotSingleSelection()` carries a translated `AD_Message` (`MSG_ONLY_ONE_SELECTED_ROW_ALLOWED`) with `internal=false`, so it actually reaches the user, in their language.

Applied at all five sites with that condition, not just the one you marked: the four quick actions (insert above, delete, move up, move down) and `WEBUI_Order_DocTextLines_Launcher`'s "one and only one order shall be selected". The three other `rejectWithInternalReason("one and only one order shall be selected")` occurrences in the repo are pre-existing code outside this feature, so I left them alone.

One thing I checked rather than assumed: `rejectBecauseNotSingleSelection()` resolves `IMsgBL` at call time, and these preconditions are unit-tested in a harness with no database. The tests do reach the multi-selection branch, and they pass — 80/80 green on a clean Java-8 build.

Worth flagging, since it is the reason this slipped through: the tests only ever assert `resolution.isAccepted()`, never the reason. An internal reason and a user-facing one are indistinguishable to them, so nothing pinned the old text and nothing would have caught it.

**Still open, deliberately:** the other refusals in these processes have no canned equivalent — "article lines are read-only and cannot be moved", "already the last row", "not persisted", "not allowed for a Closed, Voided or Reversed order". Making those user-visible means new `AD_Message` records with migration scripts, which is Task 22 on this branch ("AD_Messages, translations"). Flagging it rather than leaving it looking finished; say the word if you would rather have that task pulled forward.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/doc_textlines/process/WEBUI_Order_DocTextLines_Launcher.java
```diff
@@ -0,0 +1,99 @@
...
+
+		final IView view = viewsRepo.createView(docTextLinesViewFactory.createViewRequest(recordRef));
+		final ViewId viewId = view.getViewId();
+
+		getResult().setWebuiViewToOpen(WebuiViewToOpen.builder()
+												.viewId(viewId.toJson())
+												.target(ViewOpenTarget.ModalOverlay)
+												.build());
+
+		return MSG_OK;
```
metas-ts, line 98: Is this return reached after the modal was opened, or after it was closed?
I think it matters for the locking, because the lock is AFAIU released when this doIt() method exists.

metas-ts: **Resolution (dev):** It returns as soon as the view is created — not when the modal closes. But your instinct is pointing at something real, and chasing it down means I have to correct the advice I gave you on the locking thread.

**On this method.** `doIt()` builds the view, puts its id into the process result as `WebuiViewToOpen` and returns. The frontend reads that from the process-execution response and opens the modal; nothing on the server is waiting for it. There is no point in this method that corresponds to "the user closed the modal".

**On the lock, as it is today: this method takes none.** The launcher only creates the view. The document lock is acquired inside each *structural write* — insert-above, move, delete — in that write's own transaction, and released when that transaction ends. So the launcher's `doIt()` returning has no bearing on locking at all; a lock is held for the duration of one quick action, typically milliseconds, and never across the modal's lifetime.

**Where your question really bites — and where I was wrong.** On the other thread I said the modal-lifetime variant would need a release story we'd have to build: an explicit unlock on close, plus a timeout for the abandoned-browser case. Going looking because of this comment, both already exist in the view framework:

- `IView.close(ViewCloseAction)` — called when the frontend closes the view (`DELETE /view/{viewId}` → `ViewsRepository.closeView`). That is the explicit-close hook.
- `IView.afterDestroy()` — called by `DefaultViewsRepositoryStorage` from its Guava cache `removalListener`. The cache is `expireAfterAccess(viewExpirationTimeout)`, default **60 minutes**, overridable by the `WebUI_View_ExpirationTimeoutInMinutes` sysconfig. That is the abandoned-browser timeout.

So a modal-lifetime lock would release on close via `close(...)`, and on an abandoned tab via `afterDestroy()` at the latest — and `setAutoCleanup(true)` on the lock is a third net under a JVM crash. That is a materially better position than I described to you, and it was wrong of me to price that option as "needs a release story built" when the hooks were already there.

It does not make the choice automatic — sixty minutes is a long time for one user to hold a document if the eviction is the only thing that fires, and `close(...)` depends on the frontend actually issuing the DELETE. But it is a real design now rather than a hypothetical one, so the fork I put to you is a fairer one than it was:

- **per structural write** (what is on the branch): nobody is ever blocked for longer than one quick action; two users can both have the modal open, and are refused only if they act at the same instant.
- **for the modal's lifetime**: the second user learns immediately that someone is editing this order's text lines, at the cost of one user being able to hold it — worst case until the view expires.

Still your call; I have not built the second.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/doc_textlines/DocTextLinesDocumentLocks.java
```diff
@@ -0,0 +1,42 @@
...
+ * Striped rather than one lock per document: a map keyed by document would grow for the lifetime of the JVM
+ * with an entry per document anyone ever edited. A fixed set of stripes costs nothing and cannot leak; the
+ * price is that two unrelated documents can share a stripe and briefly wait for each other, which is
+ * invisible next to the round trips the lock already covers.
+ */
+final class DocTextLinesDocumentLocks
+{
+	private static final int STRIPES = 64;
+
+	private static final Striped<Lock> LOCKS = Striped.lock(STRIPES);
```
metas-ts, line 32: Stripes is annotated as `@Beta`. Is it safe to use here?

metas-ts: **Resolution (dev):** Checked rather than reasoned about it, and I think yes — but the decision is a library-policy one, so here is the evidence and an alternative that avoids the question entirely.

**What `@Beta` actually is.** It is a *compatibility* marker, not a stability one: Guava does not promise that a `@Beta` API survives an upgrade unchanged. It says nothing about whether the code is fit for production — `Striped` has been in Guava since v13 (2012) and is unchanged in the 30.1.1-jre we pin (`misc/parent-pom/pom.xml:96`). Confirmed the annotation is really there by reading the class in our own jar rather than trusting the javadoc.

**So the risk is a compile break on a future Guava upgrade, not a runtime surprise** — it would stop the build, not a customer's document.

**Precedent, measured, not assumed.** Guava `@Beta` is already in the codebase: `com.google.common.collect.Streams` (5 files), `com.google.common.io.Files` (7), `com.google.common.eventbus` (3), `com.google.common.hash.Hashing` (2) — each verified to carry `@Beta` in the same jar. `Striped` itself has no other user in the repo, so this would be the first.

**Why the blast radius is small if it ever breaks.** This lock is deliberately *not* the correctness guarantee — the document's `ILockManager` lock is. The in-process one only lets a second writer in the same JVM wait on a monitor instead of opening a transaction and being refused, and it is the only serialisation the DB-less unit tests have. Nothing about ordering correctness depends on it.

**The alternative, if you would rather carry no `@Beta` at all.** `Striped.lock(64)` with a fixed count is an eagerly-populated array of `ReentrantLock` selected by a smeared hash — about six lines to inline, with no dependency question and no behaviour change:

```java
private static final Lock[] LOCKS = new Lock[64];  // populated in a static block
static Lock forDocument(final DocTextLineDocumentRef ref)
{
    final int h = smear(ref.hashCode());
    return LOCKS[h & (LOCKS.length - 1)];
}
```

The one thing worth keeping from Guava is the hash smearing; a raw `hashCode() % 64` clusters badly for keys with weak low bits.

My recommendation is to keep `Striped` — it is better-tested than an inlined copy, the failure mode is a build error rather than a defect, and the precedent is established. But say the word and I will inline it; it is a small, self-contained change either way.

PR #25956, opened 2026-09-16
MobileUI ui-trace: bound the event backlog and stop destroying in-flight events

mrie04:
## What

The mobile-UI trace sync task runs once a second for the entire life of the tab. It read the **whole** IndexedDB event store on every cycle, POSTed all of it, then cleared the entire table. Two defects follow.

**1 — events saved during an in-flight POST were destroyed.**
`clearEvents()` emptied the whole table after a successful POST, so anything written while that POST was in flight went with it. Silent trace loss, and worst precisely when the device is scanning fastest and events arrive thickest. Now only the ids actually posted are deleted.

**2 — a failing POST let the backlog grow without bound.**
Nothing was deleted unless the POST succeeded, while new events kept arriving and each cycle re-read and re-serialised a strictly larger backlog. On a long-lived tab against a degraded backend this compounds every second. Now a cycle sends at most `MAX_EVENTS_PER_SYNC` (200), and the store is capped at `MAX_STORED_EVENTS` (5000), dropping oldest-first. The cap is enforced **before** the offline early-return, so being out of Wi-Fi coverage is not a path that skips it.

## Schema change

Neither bound was expressible against the v1 schema, whose only key was a uuid — the store had no usable order, so "send the oldest first" and "drop the oldest" were both impossible to write.

`db.version(2)` indexes the event timestamp and **backfills `ts` for records written by v1**. The backfill is load-bearing: Dexie omits records whose indexed value is `undefined` from that index, so without it a pre-upgrade backlog would be invisible to every ordered query here — never sent, never trimmed.

## Tests

`src/utils/ui_trace/__tests__/useUIEventsTracing.test.js`, 5 cases: in-flight event survives and is delivered on a later cycle; backlog stays bounded under a failing POST; per-cycle payload stays bounded; healthy-backend control drains every cycle; offline still enforces the ceiling.

Each was written against the previous implementation and **observed to fail there** before the fix landed. Every case carries an explicit non-vacuity assertion, because CRA sets `resetMocks: true` and a mock whose implementation is not reassigned in `beforeEach` silently returns `undefined` — which would let these pass without a single POST firing.

Verified locally with the project runner (`react-scripts test`, not bare jest — the latter does not apply `resetMocks` and lets mock call history bleed between tests):

- `--testPathPattern=ui_trace` -> **5 passed**
- full frontend suite -> **152 passed, 1 failed**

The one failure is `ApplicationLayout.deadDeepLink.test.js`, which **fails identically on the unmodified base** (verified by reverting both source files and re-running). It is pre-existing on this branch and untouched by this change.

## Scope

This is the bounded, reproduced part of a wider mobile-UI performance investigation. It does **not** claim to fix the reported scan-latency symptom — the dominant contributor there is still unlocalised, and the measured evidence so far indicates no code regression shipped in any recent rollout.


misc/services/mobile-webui/mobile-webui-frontend/src/utils/ui_trace/db.js
```diff
@@ -1,29 +1,118 @@
...
+};
+
+/**
+ * Enforces a ceiling on the stored backlog by dropping the OLDEST events, and returns how many were
+ * dropped.
+ *
+ * Without a ceiling, a tab that lives for days with an unreachable backend grows this store without
+ * bound: nothing is deleted unless a POST succeeds. Dropping the oldest (rather than refusing new
+ * events) is deliberate — a UI trace is read to diagnose what the device is doing now, so the newest
+ * events are the ones worth keeping.
```
teosarca: when this is actually happening, realisticly?
asking because if there is not backend connection, other actions are not usable neither.
also, in such case shouldn't we also console.log/warn?

pls to an full adverserial review

misc/services/mobile-webui/mobile-webui-frontend/src/utils/ui_trace/useUIEventsTracing.js
```diff
@@ -1,26 +1,52 @@
...
+    // Enforced FIRST, every cycle, so the offline early-return below is not a path that skips it —
+    // a device out of Wi-Fi coverage is exactly the case where the backlog grows unchecked.
+    await trimOldestEvents(MAX_STORED_EVENTS);
+
+    if (!navigator.onLine) {
+      console.log(`Skip syncing because not online (${reason})`);
+      return;
+    }
+
+    const events = await getEventsBatch(MAX_EVENTS_PER_SYNC);
```
teosarca: when the next batch will be sent?

misc/services/mobile-webui/mobile-webui-frontend/src/utils/ui_trace/db.js
mrie04: **Resolution (dev):** Adopted the warn — https://github.com/metasfresh/metasfresh/commit/8561246d104ae2f9e28e0e28e1c46a4dcd6a6b2d. Reaching the cap means we are discarding diagnostic data, and the trace cannot record its own loss, so it has to hit the console.

On "realistically, when?" — your scepticism is right, and it makes this the *least* important part of the PR:

Production `ui_trace` carried ~387,340 rows for 2026-07-01 to 2026-09-16, about 5,030 events/day **across all devices**. So 5,000 stored events is on the order of a full day of the entire fleet's events, or several days for one device. The cap only fires after that much **unbroken** POST failure. In practice: near enough never.

And your second point stands too — with no backend the operator cannot pick anyway, so the device is not generating a normal event stream during an outage. The realistic case is not "backend down for days", it is brief wifi roams of seconds, which never approach the cap.

So the cap is a safety net against unbounded growth, not a mechanism that routinely runs. **The part of this PR that actually does work every cycle is the batch limit** (`MAX_EVENTS_PER_SYNC`): before it, each cycle read and re-serialised the entire store every second, so a degraded backend made per-cycle cost grow with the backlog. That is the real fix; the cap just stops the store from growing without bound if the unlikely case ever arrives.

Happy to drop the cap entirely if you would rather not carry a mechanism that never fires — the batch limit and the per-id delete stand on their own. I would keep it, but weakly.

Full adversarial review: running now, I will post what it finds.

misc/services/mobile-webui/mobile-webui-frontend/src/utils/ui_trace/useUIEventsTracing.js
mrie04: **Resolution (dev):** About one second later, and it keeps going until the store is empty.

`usePeriodicTask` reschedules from *completion*, not on a fixed clock — it awaits the task, then sets a 1000 ms timeout. So the cadence is `cycle duration + 1s`, and each cycle sends up to `MAX_EVENTS_PER_SYNC` (200) oldest-first. A 1,000-event backlog drains in five cycles, roughly six seconds; the 5,000 cap would take 25 cycles, about half a minute.

Worth stating the two bounds explicitly, since they decide whether that is fast enough:

- **Drain rate** is ~200 events/second. Production runs ~5,030 events/day across the whole fleet, so the drain outruns production by orders of magnitude — a backlog always shrinks to zero once the backend is reachable.
- **If production ever did exceed the drain rate**, the store does not grow without bound; it settles at the cap and drops oldest-first (now with a console warning).

The batching is only there so one cycle's cost does not scale with the backlog. Before this PR every cycle read and re-serialised the entire store, once a second, for the life of the tab.

misc/services/mobile-webui/mobile-webui-frontend/src/utils/ui_trace/db.js
mrie04: **Resolution (dev):** Adversarial review done, as you asked. It found six things. Two I fixed, two are wording, one I pushed back on, and **one is a decision for you**. Full report kept at `ai-work/31264/review-8561246.md`.

### Fixed

**Store operations had no timeout, only the POST did.** This was a real gap I left open. The POST got a timeout because the in-flight mutex turns one stuck `await` into permanent silence for both triggers — but the three Dexie calls that same mutex guards had no ceiling. A `db.open()` blocked by a stale tab on an older schema never settles, which is exactly what `db.on('versionchange')` exists to shorten, and it would wedge sync just as thoroughly. All three now run under a 10 s deadline; covered by a test that hangs `trimOldestEvents` forever and asserts a later cycle still runs — verified it fails without the wrapper. https://github.com/metasfresh/metasfresh/commit/29e30e48f6e1c1ef1f6a2ec0be1ad3f00b4c5e93

**The cross-tab duplicate window was understated in the comment.** I had it as a rare edge; it is systematic — the mutex is per-tab, IndexedDB is not, and this app is documented to run as an installed PWA plus a browser tab on one handheld. Comment corrected in the same commit.

### Your call

**`db.js` has no direct test coverage** — the reviewer rates this the top finding. The migration, the ordering and the trim logic are all untested, because the only test file mocks `db.js` wholesale. Closing it means adding `fake-indexeddb` as a devDependency to the shared mobile-webui frontend.

I did not take that unilaterally: it is a dependency added to a shared package off the back of a six-file change. Your call whether it is worth it here or as a follow-up. The migration is the riskiest code in the PR and the one thing that runs exactly once per device, unobserved.

### Pushed back

**The constants are not dead weight.** I had offered to drop `MAX_STORED_EVENTS` since 5,030 events/day fleet-wide means it essentially never fires. The reviewer's counter is that the 5,030 figure describes steady state, not the outage tail the cap defends against — its job is to bound the worst case, not the normal one. That is a better argument than mine, so I withdraw the offer and recommend keeping all three as-is.

### Noted, not fixed

**`ts` ordering uses the client wall clock with no monotonic tie-breaker**, so a device clock rollback can invert oldest-first and drop genuinely recent events. Real, but the consequence only materialises when the cap fires — which per the above is near never — and mis-ordered *sends* are harmless since everything is sent either way. Happy to fix it if you disagree; it would mean an insertion sequence rather than a timestamp as the order key.

**One comment wording nit** in the migration block, purely editorial.

PR #25958, opened 2026-09-16
MobileUI: mount one toast container for the whole app instead of one per screen

mrie04:
## The symptom

Operators report handheld scanners getting progressively slower through a working day, with a device restart resetting them. Measured against the mobile UI: every navigation left a complete copy of the screen it had left behind, retained in memory and never collectable.

## Measurement

12 in-and-out navigation cycles (applications list → picking → jobs list → back), counting DOM nodes still held after a forced garbage collection:

| | before | after |
|---|---|---|
| cycle 1 | 2,701 | 2,630 |
| cycle 4 | 10,018 | 2,860 |
| cycle 10 | **24,652** | **2,860** |
| per cycle | **+2,439, linear, no plateau** | **0 after cycle 3** |
| JS event listeners | 328 → 1,309 | 330 → ~354, flat |

The count of **attached** nodes never moved in either case — 115 throughout. So all the growth was *detached* subtrees: screens removed from the document but still referenced, and therefore uncollectable.

Independently confirmed outside any test harness: the DOM node count in Chrome's Performance monitor climbs on every navigation and never returns.

## Cause — a mount/unmount race in react-toastify 7.0.0

react-toastify keeps every mounted `ToastContainer` in a module-level `containers` Map, where the entry holds a `forceUpdate` bound to that container's React fiber.

Deregistration **does** exist: unmount emits a `WillUnmount` event whose handler removes the entry. But `eventManager.emit()` never runs listeners synchronously — it schedules each via `setTimeout(…, 0)` — and every *mounting* container first calls `eventManager.cancelEmit(WillUnmount)`, which clears **all** pending timers for that event globally, not just its own.

So when one screen's container unmounts and the next screen's mounts in the same tick — exactly what a route swap does — the outgoing container's deregistration is cancelled before it ever runs. Its entry is orphaned in the Map permanently, pinning the unmounted screen's fiber tree and, through it, that screen's entire DOM.

`ScreenToaster` (which wraps a `ToastContainer`) was rendered **per screen** — `LoginScreen`, `ApplicationsListScreen`, and both branches of `ApplicationLayout` — so this happened on every navigation. A heap snapshot's retaining path confirms the chain:

```
containers Map -> isToastActive closure -> forceUpdate (native_bind)
  -> bound_argument_0: FiberNode -> return/child/sibling ... -> <div class="full-size-btn">
```

The leaked elements are the `wflauncher-button` subtrees — the job-list buttons of the screen being left.

## The change

**One `ScreenToaster` for the whole application**, mounted inside `ConnectedRouter`, replacing the four per-screen mounts. No unmount ever races a mount again, so the orphaning cannot occur.

To be precise about what this is: react-toastify v7 *does* support multiple simultaneous containers as an opt-in feature (`containerId` / `enableMultiContainer`), so this is not simply "we were using it wrong". The library's baseline documented pattern is one container per app, this app never needs two concurrent toast surfaces, and a single permanent container is the right architecture here on its own merits — while also sidestepping the library defect entirely. A future feature genuinely needing two concurrent toast regions would hit the underlying bug again.

**`useLocationChange` now subscribes to the router.** This is required by the change above, and fixes a latent defect. The hook listened only for native `popstate`/`hashchange`, which do not fire for `pushState`/`replaceState` — and `useMobileNavigation` routes essentially all in-app navigation through `history.replace()`. What actually re-ran the callback was the host component *remounting* on each navigation. `ScreenToaster` uses this hook to dismiss toasts on navigation; mounted once, it would never have fired again, and an error toast raised on one screen would follow the operator to the next.

The hook now also uses `history.listen()`, so it fires on every navigation regardless of remounting. The existing listeners are kept, and the `lastKnownLocation` comparison already makes a duplicate notification a no-op — verified that a browser back/forward still produces exactly one callback, not two.

One correction on the other caller, `useUITraceLocationChange`: it is **not** merely "unaffected" — its callback was already never firing before this change, and still never fires. Both hooks share the single `lastKnownLocation` sessionStorage key, and `ScreenToaster`'s child-level effect always commits first and claims it, so the trace hook's check always sees an already-current value and no-ops. That is a pre-existing defect independent of this PR (reproduced identically on the unmodified base branch), and it means the `ui_trace` location-change event has been silently absent from collected traces. It is being raised separately rather than fixed here.

## Scope

This is a **core** defect, not customer-specific — `ScreenToaster` and `ButtonWithIndicator` are byte-identical across the customer branches checked, so every deployment leaks. It is raised on this hotfix branch because that is where it was diagnosed.

## Tests

`src/hooks/__tests__/useLocationChange.test.js` — fires on push, fires on replace, does not re-fire for the same location, stops after unmount. The first two were confirmed to **fail against the previous implementation** before the fix.

Full mobile-webui frontend suite: **151 passed, 1 failed**. The single failure is `ApplicationLayout.deadDeepLink`, which **fails identically on the unmodified base branch** (verified by reverting and re-running) and is unrelated.

Also removes the now-stale `jest.mock` of `ScreenToaster` from the two `ApplicationLayout` tests, since that component is no longer rendered there.


misc/services/mobile-webui/mobile-webui-frontend/src/containers/applicationRoot/ApplicationRoot.jsx
```diff
@@ -70,6 +71,12 @@ const ApplicationRoot = () => {
   return (
     <>
       <ConnectedRouter history={history} basename="./">
+        {/* ONE toast container for the whole app. react-toastify registers each mounted
+            ToastContainer in a module-level `containers` Map; mounting one per screen left an entry
+            behind on every navigation, and each entry holds a forceUpdate bound to that screen's
+            React fiber - which pins the screen's whole DOM subtree. Mounted inside the router so
+            ScreenToaster's useLocationChange still sees route changes. */}
+        <ScreenToaster />
```
teosarca: how the ScreenToaster refactoring is actually solving the DOM+heap problems?!
pls CSP ELI5

misc/services/mobile-webui/mobile-webui-frontend/src/hooks/useLocationChange.js
```diff
@@ -1,31 +1,56 @@
-import { useEffect } from 'react';
-import { useRouteMatch } from 'react-router-dom';
+import { useEffect, useRef } from 'react';
+import { useHistory, useRouteMatch } from 'react-router-dom';
 
 export const useLocationChange = (onChange) => {
   const currentRoute = useRouteMatch();
+  const history = useHistory();
+
+  // The subscription below is installed once, so it must not close over the first render's props.
```
teosarca: pls reduce AI slop.i think that useRef/useCallback approach is quite common in the react world. so no need to explain it again.

teosarca: make sure u have a corpus + review rule for this AI slop comments.

misc/services/mobile-webui/mobile-webui-frontend/src/hooks/useLocationChange.js
```diff
@@ -1,31 +1,56 @@
...
         currentLocation,
-        currentRoute,
+        currentRoute: currentRouteRef.current,
         prevLocation: lastKnownLocation,
       });
     }
   };
 
   useEffect(() => {
+    // history.push/replace do NOT emit a native popstate event — that fires only for genuine browser
```
teosarca: again AI slop!

teosarca: pls save such explanantions for PR summary/comments but do not polute the code. pls capture learning + review rule

misc/services/mobile-webui/mobile-webui-frontend/src/containers/applicationRoot/ApplicationRoot.jsx
mrie04: **Resolution (dev):** ELI5, and the comment at this line now says it too (97ae6440e260fbc5f48b5cb17d0dcd23f0b39d72):

1. Every screen mounted its own toast container.
2. react-toastify keeps a list of live containers. When one unmounts it removes itself from that list **on a timer** (`setTimeout(…, 0)`), not immediately.
3. Any container that mounts first calls `cancelEmit`, which kills **all** pending removals, not just its own.
4. A navigation unmounts the old screen and mounts the new one in the same tick. So the old container's removal is cancelled before it ever runs, and its entry stays in the list forever.
5. That entry holds a `forceUpdate` bound to the old screen's React fiber. A fiber points at its DOM. So the whole old screen stays in memory, unreachable but uncollectable.

One container for the whole app never unmounts, so step 2 never happens, so there is nothing for step 3 to cancel. No orphan, nothing pinned.

Measured over 12 in-and-out navigations, DOM nodes still held after a forced GC:

| | before | after |
|---|---|---|
| cycle 1 | 2,701 | 2,630 |
| cycle 4 | 10,018 | 2,860 |
| cycle 10 | 24,652 | 2,860 |
| per cycle | +2,439, linear, no plateau | 0 after cycle 3 |

Attached nodes were constant at 115 in both runs, so every one of those was a detached screen.

misc/services/mobile-webui/mobile-webui-frontend/src/hooks/useLocationChange.js
mrie04: **Resolution (dev):** Agreed, removed in https://github.com/metasfresh/metasfresh/commit/97ae6440e260fbc5f48b5cb17d0dcd23f0b39d72. The useRef-to-avoid-a-stale-closure explanation is gone; that is standard React and did not need restating. What is left on `currentRouteRef` is two lines for the part that is not standard: for a caller that never remounts, that field is one navigation behind, because our listener runs before Router propagates the new RouteContext. The file went from 58 lines to 44, of which comment lines went from fifteen down to four.

misc/services/mobile-webui/mobile-webui-frontend/src/hooks/useLocationChange.js
mrie04: **Resolution (dev):** Cut too, same commit https://github.com/metasfresh/metasfresh/commit/97ae6440e260fbc5f48b5cb17d0dcd23f0b39d72. Five lines became two, and only the non-obvious half survived:

```js
// history.listen is what catches push/replace; popstate only fires on browser back/forward.
// Without it the hook only worked because its caller remounted on every navigation.
```

I kept that much because it is the reason the line exists: this hook previously had no router subscription at all and only appeared to work because its caller remounted on each navigation. Moving ScreenToaster to a single mount removed that accident, which would have silently killed dismiss-on-navigate. Without the note the next person is likely to delete `history.listen` as redundant next to the popstate listener.

mrie04:
**Resolution (dev):** Update — this PR now has a measured causal chain to the reported symptom, not just to memory use. Posting it because the original description undersold what it fixes.

### What the customer's videos show

The device sits in the app's **orange "Scan läuft…" partial-scan state for 8–13 seconds** while a single code arrives, then the dialog opens instantly. So the delay is in **delivery of the scanned code into the app**, entirely before the scan is registered — not in processing it afterwards.

That matters because a production query of `barcodeScanned → dialogOpen` returns **0.02 s**, flat at every level of accumulation. True, and measuring the wrong end of the chain: everything the operator waits for happens before that first event exists.

### Why earlier measurements missed it

Dispatching a whole code in one synchronous loop lets the browser batch every style/layout recalculation into a single frame. Per-keystroke cost disappears. A real wedge sends ~459 **separate** events, each paying its own layout against the current DOM.

Measuring with one keystroke per task (yielding via `MessageChannel` — `setTimeout(0)` is clamped to ~4 ms and would dominate), 4x CPU throttle, a real 430-char HU QR:

| | before | after |
|---|---|---|
| DOM nodes | 3,139 -> 48,003 | 351 -> 466 |
| **delivery of one code** | **42 ms -> 274 ms (x6.5)** | **71 ms -> 82 ms (flat)** |
| per keystroke | 0.10 -> 0.64 ms | 0.13 -> 0.19 ms |
| worst inter-key gap | 3.6 -> 46.7 ms | 3.7 -> 12.7 ms |

### The chain, end to end

leaked DOM -> each keystroke's style/layout costs more -> a 430-459 char code takes longer to arrive -> the app sits in "Scan läuft…" -> the operator waits -> **a restart clears the DOM and it is fast again**.

Every link is measured. It also matches the customer's own words: *"as if the device were getting older"*, and *"after a restart it is reset and the speed is fast again"*.

### Scope of the claim

48,000 nodes is about a third of a shift (145,000 measured separately), on a 4x-throttled desktop rather than the actual handheld. The **shape** is reproduced and the fix removes the growth; the absolute 8-13 s is not claimed to be reproduced here.

Separately measured on this branch: nodes x46.5 -> x1.0, heap 170 MB -> 16 MB, listeners x26.7 -> x1.0 across a simulated shift.

PR #25995, opened 2026-09-18
Text lines: implement getFieldDropdown so the scope column is editable

metas-ts:
## Problem

Opening the scope dropdown of a text line in the text-lines modal returns HTTP 500:

```
java.lang.UnsupportedOperationException
  IEditableView.getFieldDropdown:58 <- ViewRowEditRestController.getFieldDropdown:183
  path: /rest//api/documentView/docTextLines/<viewId>/T<id>/edit/textLineScope/dropdown
```

`DocTextLinesView` declares `implements IEditableView` and implements `patchViewRow`, but never
overrides `getFieldDropdown`, so it inherits the interface default, which throws.

The `textLineScope` column is declared with `editor = ViewEditorRenderMode.ALWAYS`, so
`List/List.js` issues a `.../edit/textLineScope/dropdown` GET the moment the user opens that cell.
The view could not answer it, so the column was **not editable at all** — the value stored on the
record was always correct, which is why nothing that reads the view's data endpoint noticed.

## Fix

- `DocTextLinesViewFactory` builds the `TextLineScope` reference-list lookup from the
  `LookupDataSourceFactory` it already holds — next to the existing products lookup, and the same
  way `PricingConditionsRowLookups` builds its list-backed lookups — and passes it into the view,
  rather than having the view resolve a lookup per request.
- `DocTextLinesView.getFieldDropdown` returns that list whole (it is a short fixed reference list,
  not a searchable table) for `textLineScope`, and rejects any other field name with the same
  `AdempiereException` message `PricingConditionsRowLookups` uses.

No `getFieldTypeahead` counterpart was added: a `List` widget only ever requests `/dropdown`
(`frontend/src/components/widget/List/List.js`); `/typeahead` is requested by the `Lookup` widget
(`RawLookup.js`), and this view has no editable column of that type. Implementing it would be dead
code.

## Tests

The defect shipped because no test ever walked the dropdown path.

- **Browser spec** (`e2e/frontend-webui/tests/spec/text-lines-modal.spec.js`) — the primary test for
  a desktop-WebUI flow. The existing walkthrough now ends by opening a text line's scope dropdown,
  asserting it offers both reference-list codes, picking the other one, and asserting the grid cell
  paints it — then reopening the modal and asserting it again, so the value has to come back from
  the database rather than from the view instance the click updated. These assertions read the
  rendered DOM, because an empty dropdown is invisible to the data endpoint. The expected caption is
  taken from the option that was clicked, so the step stays language-independent.
- **Unit tests** (`DocTextLinesViewTest.fieldDropdown`) — only for what the browser cannot reach:
  that the field is answered from the view's own scope lookup, and that an unknown field name is
  rejected. `MockedTextLineScopeLookup` stands in for the reference-list lookup, since a plain unit
  test has no database.

Both unit tests fail against the previous code with the `UnsupportedOperationException` above.

Verified locally (Java 8): `de.metas.ui.web.base` — 1506 tests, 0 failures, 0 errors.
Clean `test-compile` of the changed module and all 11 dependent modules passes.

The browser spec was not executed locally: this machine's ports for the app server, webapi and
frontend dev server are held by other workspaces' long-running stacks, so no stack for this branch
could be brought up without disturbing them. It runs in CI.


backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/doc_textlines/DocTextLinesView.java
```diff
@@ -36,19 +42,53 @@ public static DocTextLinesView cast(final Object viewObj)
...
 		this.processes = processes != null ? ImmutableList.copyOf(processes) : ImmutableList.of();
 	}
 
+	/**
+	 * Supplies the values the frontend offers when the user opens a text row's scope cell. The scope column is
+	 * declared {@code editor = ViewEditorRenderMode.ALWAYS}, so {@code List/List.js} issues a
+	 * {@code .../edit/textLineScope/dropdown} GET the moment that cell is opened, and
+	 * {@code ViewRowEditRestController} routes it here -- without this override the column is not editable at
+	 * all, because {@link IEditableView}'s default answers every such request with
+	 * {@link UnsupportedOperationException} (HTTP 500).
```
metas-ts:
```suggestion
	 * Supplies the values the frontend offers when the user opens a text row's scope cell.
```

PR #26031, opened 2026-09-21
Role table access: per-role create permission and three-state access flags

mrie04:
Adds a per-role, per-table permission controlling whether a role may create new records, and converts the existing `AD_Table_Access` flags to a three-state model so that "not set" means the role expresses no opinion.

**Dictionary and schema**
- `IsExclude`, `IsReadOnly`, `IsCanReport`, `IsCanExport` become nullable List references over `_YesNo`, with their `NOT NULL`, DB defaults and two-valued CHECK constraints removed. A List value is validated against its ref-list on save, so a check constraint would only duplicate that.
- New column `IsCanCreateNewRecords` in the same shape, with German, Swiss-German and English element text.
- `AccessTypeRule` is removed — with the three-state model one row expresses every aspect — and the primary key narrows to `(AD_Role_ID, AD_Table_ID)`. Three fields on the legacy tab that gated their display on the removed column have that logic cleared.

**Permission model**
- The table-permission loader derives the three-state flags directly: a row where no access-affecting flag has an opinion emits no permission at all, so it stays neutral, while a row with an opinion is seeded from the role's default accesses.
- The loader is split into two passes. The previous single loop mutated the shared default access set while already emitting permissions, so a permission seeded from those defaults depended on the order rows were returned in.
- The new permission is carried on `TablePermission` and resolved across included roles by the same union every other permission uses: an explicit Yes anywhere wins, an explicit No is never silently re-granted, and all-unset falls through to unrestricted.

Enforcement, the administration UI and the end-to-end coverage follow in later commits on this branch.


backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/security/permissions/TablePermission.java
```diff
@@ -35,13 +37,20 @@ public final class TablePermission implements Permission
 	private final TableResource resource;
 	private final ImmutableSet<Access> accesses;
 
+	/** {@code null} means the role states no opinion about creating new records in this table. */
+	@Getter
+	@Nullable
+	private final Boolean canCreateNewRecords;
```
teosarca: why a new Boolean and not a new de.metas.security.permissions.Access.* access type?

mrie04: **Resolution (dev):** A new `Access` value was the first candidate and was rejected on three grounds — the decisive one is that `Access` membership is two-state and this permission has to be three-state.

**1. `Access` cannot express "not configured".** A `TablePermission` holds `ImmutableSet<Access>` and answers `hasAccess` by `accesses.contains(access)` (`TablePermission.java:83`). Absence of a value is therefore indistinguishable from an explicit denial. The requirement needs a third state: *Yes* / *No* / *not set*. That is not cosmetic — role inclusion merges permissions by union, so an included role that nobody configured must contribute **nothing**. With a set membership it would instead contribute "absent", i.e. denied, and every role in the system would flip to restricted the moment the value existed. The nullable `Boolean` (`TablePermission.java:43`) is what makes "no opinion" representable, and `mergeCanCreateNewRecords` keeps the union semantics for the two states that do express an opinion.

**2. The default has to fail open, per role and over time.** A role created after this ships carries no `AD_Table_Access` row at all, and must stay unrestricted until an administrator restricts it. With an `Access` value, "no row" and "row without the value" both read as denied — the new role would be blocked from creating records on every table it has no explicit grant for. The tri-state gives the intended behaviour with no backfill migration over existing roles.

**3. An `Access` value would leak far outside the intended enforcement surface.** `Access` is consumed by ~95 files across the backend and flows into `isTableAccess` (`IUserRolePermissions.java:202`), the record-level access filters and the generated SQL where-clauses. This restriction is scoped deliberately to the WebUI document layer: the REST/leading-system interface must stay unaffected, with no role exemption to configure for it. Keeping the flag off `Access` is what guarantees that, rather than requiring every `Access` consumer to be audited for a value it should ignore.

So: not a new `Access` type, and not a deny-set either — a tri-state flag on `AD_Table_Access` that never becomes an `Access`.

What would change this: if the restriction should also apply to the REST interface and to record-level filtering, then reason 3 inverts and an `Access` value becomes the right shape — but reasons 1 and 2 would still need a separate answer for the three-state default. Say the word if you want the REST surface in scope; that is a requirements change, not a refactor.

teosarca: so we have the same problem on REPORT and EXPORT (at least). How is that mitigated?

mrie04: **Resolution (dev):** Good question, and it lands on something real — it is the strongest argument *for* the shape rather than against it, but it also exposes a defect I had already flagged separately.

**REPORT and EXPORT are already three-state, by exactly this mechanism.** In `removeAccessesRevokedByInclusion` (`UserRolePermissionsDAO.java:824-831`) they are read as `StringUtils.toBoolean(record.getIsCanReport(), null) != null` — *null means the row states no opinion*, so the role default survives; non-null means the row has an opinion, so the default is revoked. That is the same nullable-`Boolean` tri-state `canCreateNewRecords` uses. I did not invent a shape here; I copied the one REPORT/EXPORT already use.

**The layering is the actual answer to "why not an `Access` value".** `IsCanReport` / `IsCanExport` / the new create flag are all *inputs* — columns on `AD_Table_Access` that the loader reads. `Access.REPORT` / `Access.EXPORT` are *outputs* — what those inputs resolve to in the permission set. So REPORT and EXPORT don't solve the two-state problem at the `Access` layer; they solve it one layer earlier, at the flag, and hand a resolved two-state answer to `Access`. Putting the create permission at the flag layer keeps it consistent with its three siblings. Putting it at the `Access` layer would be the odd one out, and would have no place to keep "not configured".

Note the asymmetry while you're in this code: READ and WRITE are revoked **unconditionally** on an include row (`:822-823`), with no tri-state at all. Only REPORT and EXPORT are conditional. That asymmetry is pre-existing, not something this PR introduces.

**Where you are right that there is a problem.** The tri-state only *means* anything once the columns are nullable. Before this change all four flags were `NOT NULL`, so every pre-existing row carries an explicit `Y`/`N` in all four — which makes `StringUtils.toBoolean(...) != null` unconditionally true, so a converted `Accessing` row now revokes REPORT and EXPORT where the old loader, which only read the flags belonging to the row's `AccessTypeRule`, did not. That is a real behaviour change on existing data, it is not mitigated in the current diff, and I have it open as a blocking question on the issue (measured blast radius: 302 of 316 instances hold zero rows, 14 unreachable). It needs a decision on whether to add a data step before the `AccessTypeRule` drop or to narrow the acceptance criterion — I have not picked one unilaterally because it changes an approved requirement.

So: the mechanism is the right one and matches REPORT/EXPORT; the migration of existing rows into it is the part that still owes an answer.

teosarca: pls do a full adversarial review of this approach and also pls compare it to how major permission systems are modeling it. ofc keep our small size in your scope and focus on finding correct/health appraches without overcomplicating or exploding the issue scope and effort.

mrie04: **Resolution (dev):**

Full adversarial review done, plus a comparison against PostgreSQL, Kubernetes RBAC, AWS IAM, Zanzibar, Spring Security, Casbin and XACML. **Verdict: the shape is right — tri-state, restrictive-only, per-(role, table), union merge, outside the `Access` enum, enforced in the document layer. One genuine correctness defect, two free text fixes, one test I'd add against the agreed test plan. Everything else is theory.**

---

**1. Blocker — the inclusion merge can silently drop the restriction.**

`TablePermission.mergeWith` implements the union correctly, but it is only reached under `CollisionPolicy.Merge`. `UserRolePermissionsBuilder:213-221` switches to `Override` when two included roles share a `SeqNo`, and `PermissionsBuilder.addPermission` under `Override` does `permissions.put(resource, permission)` — replacing the whole `TablePermission`, `canCreateNewRecords` included, without calling `mergeWith`.

Concrete: role A has a row on `C_BPartner` with `IsCanCreateNewRecords='N'`; role B has a row on the same table for any other reason (create flag unset) at the same `SeqNo`; B overwrites A; the role can create again. Nobody set Yes. That falsifies the design's own guarantee that only a deliberate Yes can lift the restriction — this is a fail-open in the exact place we claim is safe, and it is invisible (no error, the button is just enabled).

Note this is *not* the loader-side `Override` defect we deliberately left alone — narrowing the key does make that one unreachable, and does nothing for this one.

Fix, minimum: a test pinning the equal-`SeqNo` case plus a written decision. Preferred: carry an existing explicit `FALSE` forward through `Override` in `TablePermissions.Builder` — contained to a class this change already owns. Changing `UserRolePermissionsBuilder` to always `Merge` is one line but moves existing access semantics, wider blast radius than I'd want here.

**2. WebUI-only enforcement — defensible boundary, dangerous wording.**

Not a bypass with extra steps for the actor we care about: the app's `api/v2` authenticates via `UserAuthTokenFilter` against an administered `AD_User_AuthToken` row, so a restricted user cannot reach the open path from their own browser. But the token is bound to a **role**, and because the flag is deliberately not an `Access` it never reaches `isTableAccess`, `UserRolePermissionsSqlHelpers` or `PermissionService.assertCanCreateOrUpdate` — issue a token for the restricted role and creation works.

Every system compared enforces at the point of mutation (API server, executor, service). Enforcing at the affordance layer makes this architecturally a UI guard-rail with a permission-shaped config surface. That is the proportionate control for "stop people reaching for New out of habit", and widening to REST would need a per-role interface exemption — a worse failure mode. But it cannot support the claim "records can no longer be created outside the interface", so the field's Description must say what it guarantees: *the desktop WebUI will not offer or accept creation for this role*.

**3. Union vs deny-wins — right here, and "deny wins" is not the convention it's assumed to be.**

Kubernetes RBAC and PostgreSQL have **no deny at all** ("Permissions are purely additive (there are no 'deny' rules)"). XACML ships four combining algorithms and declines to pick one. Casbin makes deny-override an opt-in effect expression. Spring Security is first-match-wins. Only IAM makes explicit-deny-beats-allow a law. So the tradeoff is real but the external pressure is weaker than it looks.

More to the point, this flag is not an IAM-style deny — it is the absence of a grant, forced into a tri-state because the row must be able to say nothing. Deny-wins would need a special case in `mergeWith` that none of the other four flags on the same row has, and would kill the deliberate administrator override. Union keeps one merge rule for the whole model and takes its accident-resistance from the third state instead. Right call — **conditional on (1)**, because without it that argument is not just weaker, it's untrue.

Worth knowing as support for the design: XACML's `NotApplicable` is exactly our "not set" — a first-class third outcome distinct from Deny, which exists in the standard for the same reason we needed it.

**4. READ/WRITE unconditional vs REPORT/EXPORT conditional on include rows** (`UserRolePermissionsDAO:815-833`) — pre-existing, correctly frozen, and not made more dangerous. It is made considerably more *reachable*: window 268 had no UI layout and no menu path, so nobody could set these; window 111 now puts them in front of an admin who came to tick one box. The bad path is an admin reading `IsExclude = No` as "this is not an exclude row" and losing the role's default READ/WRITE on every other table.

The main mitigation is already in: `5825480_…:29-56` nulls `DefaultValue` and `IsMandatory` alongside `NOT NULL`, so a new row leaves `IsExclude` unset rather than silently `'N'` — and the comment about not touching the non-core `AD_DefaultValue` table is a good catch. Residual risk is labelling only. We already owe a Description for every rendered field; spend one of them saying what `IsExclude = No` does to the rest of the role. Highest value-per-word text in that sweep.

**5. Test gap I'd close against the agreed "no JUnit".** The all-unset-row → `null` invariant in `extractTablePermissionOrNull` is one `if`, and breaking it does not mis-grant creation — it *revokes READ* on the table, because `AbstractPermissions.getPermissionOrDefault` returns a resource permission as-is with no fallback. It is a pure static function of a row plus a default access set: ~30 lines, three cases (all-unset → null; create-only → defaults + `FALSE`; `IsExclude` unset + `IsReadOnly='Y'` → defaults minus WRITE). The neutral-row Playwright spec tells us something broke; this tells us what. Not a replacement for it.

**6. Tri-state at scale — fine, under a constraint worth writing down.** The count isn't the problem; interaction is. `IsExclude` × `IsReadOnly` already needs nine cells for two flags, one of which had to be argued rather than derived. The new flag interacts with nothing — one column, one accessor, one consumer path — and twenty of those are twenty independent two-line rules. Rule: a new flag on this table must be independent. The `(role, table, privilege, effect)` normalisation is strictly better at high N and buys nothing at five.

**7. Health — no findings.** `tablePermissionsCache` is bound to `AD_Table_Access` and the table is in `ROLE_DEPENDENT_TABLENAMES`, so the no-deployment flip is real. No N+1: the check is a map lookup on a cached per-role object. Dropping `AccessTypeRule` is a net improvement — it discriminated a permission set whose table-scoped accessors are collectively unreachable.

One operational note to surface outside the requirements doc: lifting the restriction takes effect immediately server-side but the browser holds the fetched layout, so affected users must reload. For an escape hatch used during an interface outage, "we fixed it, everyone press F5" needs to be written down somewhere an admin reads.

---

**Not raising as problems** (theory, not risk): the flag not being an `Access` — correct, a create permission has no meaning in a SQL row filter; the tri-state being unusual — see XACML above; `IsExclude`/`IsReadOnly` semantics being surprising — pre-existing and rightly frozen.

**Ranked actions:** (1) fix or pin the `Override` path — blocker; (2) the loader unit test; (3) the two Descriptions (create-flag guarantee, `IsExclude = No` warning); (4) document the reload; (5..7) accept as designed.

mrie04: **Resolution (dev):** Correcting myself — my earlier three-part argument against a new `Access` value was wrong, and all three parts fail. I checked the code rather than re-defending it.

**1. "`Access` cannot express *not configured*" — wrong.** I reasoned about a single `TablePermission`'s set in isolation. The system distinguishes the third state one level up: `AbstractPermissions.getPermissionOrDefault` (`AbstractPermissions.java:107-127`) falls back to the permission registered for the *default* resource when no permission exists for the specific one. So the three states are already expressible, exactly as you suggested:

- no `AD_Table_Access` row for the table → falls through to the role's default `TablePermission` → allowed;
- row exists, the new access present in its set → allowed;
- row exists, the new access absent → denied.

That is the same mechanism READ and WRITE already use. Nothing new is needed to represent "no opinion".

**2. "The default would fail closed and need a backfill" — wrong.** The role default is `TablePermission.ALL_ACCESSES` (`TablePermission.java:24`), a code-level constant. Adding the new value there makes every unconfigured role allowed with no data migration over existing rows — which is the behaviour we want, achieved more cheaply than the nullable flag.

**3. "It would leak into every `Access` consumer" — wrong, and there is direct precedent against it.** `hasAccess` resolves a *specific* value (`accesses.contains(access)`, `TablePermission.java:83`); no consumer iterates the set, so adding a value breaks nothing. And `Access.LOGIN` (`Access.java:26`) already exists as a value scoped to **org** permissions only — it is in `Access.ALL_ACCESSES` but deliberately *not* in `TablePermission.ALL_ACCESSES`. A per-permission-type access value is an established pattern here, not a new risk.

**What actually remains in favour of the flag** — weaker than what I claimed, and I want to be straight that it is consistency rather than correctness:

- `Access` codes are `X_AD_User_Record_Access.ACCESS_*` (`Access.java:27-30`), i.e. the **record-level** access ref-list. "May create" sits oddly there: there is no record yet to hold access to, and a new value means extending a shared ref-list that record-level access also reads.
- The other four columns on `AD_Table_Access` are flags, so the tri-state flag matched its siblings.

**Where that leaves the change.** The `Access`-value shape is defensible on the merits, and my stated reasons for rejecting it were not. Switching now is not a refactor: the tri-state flag is implemented across the schema, the loader, the permission model, the WebUI payload and the frontend, so it would be a re-do of most of this branch. I am not going to make that call unilaterally — the approved requirements chose the flag (§3.2), and reopening it is a requirements decision. Say which way you want it and I will follow; if we keep the flag, it should be because it is the pragmatic choice given what is already built, not because of the argument I gave you earlier.

backend/de.metas.adempiere.adempiere/base/src/main/java/de/metas/security/IUserRolePermissions.java
```diff
@@ -204,6 +205,13 @@ public interface IUserRolePermissions
 
 	boolean isCanReport(int AD_Table_ID);
 
+	/**
+	 * The role's per-table create permission alone: restrictive only, i.e. granted unless the role explicitly forbids it.
+	 * Not an authorization check on its own - unlike {@link #checkCanCreateNewRecord(ClientId, OrgId, AdTableId)} it grants
+	 * nothing and gates nothing else, so the caller still has to establish window and table access itself.
+	 */
+	boolean isCanCreateNewRecords(int AD_Table_ID);
```
mrie04: RepoIdAware

mrie04: **Resolution (dev):** Correct, and it is an inconsistency inside this branch's own additions rather than inherited legacy: `checkCanCreateNewRecord` (`IUserRolePermissions.java:196`) already takes `AdTableId`, while `isCanCreateNewRecords` right below it takes a raw `int`. The surrounding `isTableAccess` / `isCanExport` / `isCanReport` are pre-existing raw-`int` API, which is presumably what it was copied from — but new API should follow `RepoIdAware`, not the legacy neighbours.

It is being resolved by deletion rather than by a signature change, because the design moved after this comment was written: the per-table create permission stops being a bespoke boolean on `TablePermission` and becomes a new `Access` value, resolved through the existing mechanism. `isCanCreateNewRecords` therefore disappears entirely, along with `TablePermission.canCreateNewRecords` and its bespoke merge function — callers go through the normal access path instead, which is already `AdTableId`-typed where it matters.

If the redesign had not happened the answer would simply have been `AdTableId`. Flagging it explicitly so it is not read as the point being waved away: it was a real defect in the API this branch adds, and the redesign removes the method that carried it rather than the convention being disputed.

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/window/api/impl/ADWindowDAO.java
```diff
@@ -407,6 +407,26 @@ public I_AD_Tab retrieveFirstTab(@NonNull final AdWindowId adWindowId)
...
 	}
 
+	@Override
+	@Nullable
+	public AdTableId getMainTableId(@NonNull final AdWindowId adWindowId)
+	{
+		final I_AD_Tab mainTab = queryBL
+				.createQueryBuilder(I_AD_Tab.class)
+				.addOnlyActiveRecordsFilter()
+				.addEqualsFilter(I_AD_Tab.COLUMNNAME_AD_Window_ID, adWindowId)
```
mrie04: shouldn't this have same logic as org.adempiere.ad.window.api.impl.ADWindowDAO#retrieveFirstTab ?

mrie04: **Resolution (dev):** Verified against the codebase and the live dictionary — `getMainTableId` is correct as-is and genuinely distinct from `retrieveFirstTab`, so both stay.

- `TabLevel=0` is the established "primary tab" definition (`InMemoryUIElementsProvider.java:76` `primaryTab = adTab.getTabLevel()==0`, `GridTab.java:1891`), so `getMainTableId` returns the window's header table — the record a `New` in the window creates. `retrieveFirstTab` is a different notion: first tab by `SeqNo`, any level.
- They genuinely diverge: of 849 windows with active tabs, 4 have no `TabLevel=0` tab at all (headerless sub-windows), so `retrieveFirstTab` returns a level-1/2 child there while `getMainTableId` correctly returns `null`. `getMainTableId` also restricts to active `AD_Table` rows, which `retrieveFirstTab` does not. Delegating one to the other would hand the create-gate a child tab's table for those windows.
- Added an in-code comment on `getMainTableId` documenting the distinction.

mrie04: **Correction (dev):** Keeping `getMainTableId` as-is in this PR to avoid diverting scope. But to correct my earlier reply so the record isn't misleading: the right long-term shape is to **consolidate** `getMainTableId` onto `retrieveFirstTab`, not to keep two divergent queries. The root/lowest-SeqNo tab is what `New` actually creates; `TabLevel=0` is identical on normal windows and merely fails open on the few headerless ones — so it isn't "more correct", it's identical-or-weaker while duplicating the query. Deferring that consolidation as a future improvement rather than expanding this issue.

mrie04: **Final (dev):** Consolidated in this PR — `getMainTableId` now delegates to `retrieveFirstTab` (one shared query, no divergent duplicate). Making that shared query header-aware (`TabLevel=0`) and correctly handling the few headerless windows is deferred as a follow-up improvement.

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/CreateRoleCommand.java
```diff
@@ -0,0 +1,210 @@
...
+	/** Suffix under which the role's user is registered in the {@link MasterdataContext}. */
+	public static final String USER_IDENTIFIER_SUFFIX = "_roleUser";
+
+	/** {@code AD_Role.Name} is {@code varchar(60)}; a longer name would surface as a raw DB length error. */
+	private static final int NAME_MAX_LENGTH = 60;
+
+	@NonNull private final IRoleDAO roleDAO = Services.get(IRoleDAO.class);
+	@NonNull private final IUserRolePermissionsDAO userRolePermissionsDAO = Services.get(IUserRolePermissionsDAO.class);
+	@NonNull private final IADTableDAO adTableDAO = Services.get(IADTableDAO.class);
+	@NonNull private final RoleRepository roleRepository = new RoleRepository();
```
mrie04: Why this in addition to dao and especially why non spring repo in this module with new here

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/JsonCreateRoleRequest.java
```diff
@@ -0,0 +1,42 @@
...
+ * nothing but that role, so a spec can log in and exercise the role's permissions without editing any
+ * pre-existing role.
+ */
+@Value
+@Builder
+@Jacksonized
+public class JsonCreateRoleRequest
+{
+	/**
+	 * Role name, defaulting to the section's identifier. Either way a per-run timestamp is appended, because
```
mrie04: this is standard valueandname behaviour here and in cucumbers, do we need this explanation everywhere now?

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/JsonCreateRoleRequest.java
```diff
@@ -0,0 +1,42 @@
...
+	 */
+	@Nullable List<Identifier> includedRoles;
+
+	@Nullable List<JsonRoleTableAccessRequest> tableAccess;
+
+	/**
+	 * The user to create for this role. It is bound to this role and to no other, so
+	 * {@code LoginRestController} completes its login without a role-selection step.
+	 */
+	@Nullable JsonLoginUserRequest user;
```
mrie04: you assign a role to an user and not the other way arround, this looks misplaced here

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/JsonCreateRoleResponse.java
```diff
@@ -0,0 +1,22 @@
...
+import lombok.extern.jackson.Jacksonized;
+
+import javax.annotation.Nullable;
+
+@Value
+@Builder
+@Jacksonized
+public class JsonCreateRoleResponse
+{
+	int roleId;
```
mrie04: why int and not repoidaware

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/RoleRepository.java
```diff
@@ -0,0 +1,62 @@
...
+ * ({@code InterfaceWrapperHelper.save}) belong in a {@code *Repository}/{@code *DAO} rather than in a command —
+ * see {@code docs/coding-rules/service-injection.md} §4. An ArchUnit rule enforces this.
+ * <p>
+ * Deliberately local to the frontend-testing module rather than added to the core {@code IRoleDAO}: creating a
+ * role is a test-masterdata concern with no production caller, so it must not widen the security module's API.
+ * <p>
+ * Repository Tables: AD_Role, AD_Role_Included
+ * Repository Cluster: RoleRepository
+ */
+public class RoleRepository
```
mrie04: shouldn't be here as mentionend on usage

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/JsonCreateMasterdataRequest.java
```diff
@@ -63,6 +64,15 @@ public class JsonCreateMasterdataRequest
 
 	@Nullable JsonMobileConfigRequest mobileConfig;
 	@Nullable Map<String, JsonLoginUserRequest> login;
+
+	/**
+	 * Creates purpose-built roles - their {@code AD_Table_Access} rows and a user bound to nothing but the
```
mrie04: if you keep adding this amount of comment on each map we will end up with 10k lines here and they all tell the same impl convention story. write a drift intake for this

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/JsonCreateMasterdataResponse.java
```diff
@@ -45,6 +46,7 @@ public class JsonCreateMasterdataResponse
 
 	@Nullable @JsonInclude(JsonInclude.Include.NON_EMPTY) JsonMobileConfigResponse mobileConfig;
 	@NonNull Map<String, JsonLoginUserResponse> login;
+	@Nullable @JsonInclude(JsonInclude.Include.NON_EMPTY) Map<String, JsonCreateRoleResponse> roles;
```
mrie04: why we mix 3 conventions here. nonull, nullable and nullable plus non_empty

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/controller/DocumentPermissionsHelper.java
```diff
@@ -243,18 +258,46 @@ private static int getRecordId(final Document document)
...
+
+	/**
+	 * The role's per-table create permission alone, with the refusal reason naming the role.
+	 * This is the single place the role-level answer is turned into a user-facing reason.
+	 */
+	public static BooleanWithReason checkRoleCanCreateNewRecords(
+			@NonNull final DocumentEntityDescriptor entityDescriptor,
+			@NonNull final IUserRolePermissions permissions)
+	{
+		final int adTableId = getAdTableId(entityDescriptor);
```
mrie04: why not repoidaware

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/datatypes/json/JSONDocument.java
```diff
@@ -159,7 +159,7 @@ private static JSONIncludedTabInfo createIncludedTabInfo(
 		final LogicExpressionResult allowCreateNew = includedDocumentsCollection.getAllowCreateNewDocument();
 		if (allowCreateNew != null)
 		{
-			tabInfo.setAllowCreateNew(allowCreateNew.booleanValue(), allowCreateNew.getName());
+			options.getDocumentPermissions().setAllowCreateNew(tabInfo, allowCreateNew, options.getAdLanguage());
```
mrie04: why this switch from tab to document

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/datatypes/json/JSONDocument.java
```diff
@@ -295,7 +297,7 @@ private static JSONIncludedTabInfo createIncludedTabInfo(final DocumentChanges.I
 		final LogicExpressionResult allowCreateNew = includedDetailInfo.getAllowNew();
 		if (allowCreateNew != null)
 		{
-			tabInfo.setAllowCreateNew(allowCreateNew.booleanValue(), allowCreateNew.getName());
+			options.getDocumentPermissions().setAllowCreateNew(tabInfo, allowCreateNew, options.getAdLanguage());
```
mrie04: sames as above

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/datatypes/json/JSONDocumentLayoutElementField.java
```diff
@@ -359,9 +363,32 @@ private static DocumentEntityDescriptor findNewRecordEntityDescriptor(
...
 
+	private static boolean isRoleAllowedToCreateNewRecords(@NonNull final String lookupTableName)
+	{
+		// a background thread carries no user session, so there is no role to restrict (same reading as IncludedDocumentsCollectionActions)
+		if (!UserSession.isWebuiThread())
+		{
+			return true;
+		}
+
+		final int adTableId = Services.get(IADTableDAO.class).retrieveTableId(lookupTableName);
```
mrie04: why not repoidaware

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/datatypes/json/JSONDocumentPermissions.java
```diff
@@ -86,6 +88,31 @@ public void apply(final DocumentPath documentPath, final JSONIncludedTabInfo jso
...
+	 * translated text plus the key itself. Every other refusal is transmitted as the technical name it has
+	 * always been.
+	 */
+	public void setAllowCreateNew(
+			@NonNull final JSONIncludedTabInfo jsonIncludedTabInfo,
+			@NonNull final LogicExpressionResult allowCreateNew,
+			@NonNull final String adLanguage)
+	{
+		final String roleReasonKey = DocumentPermissionsHelper.MSG_ROLE_CREATE_NOT_ALLOWED.toAD_Message();
+		if (allowCreateNew.isFalse() && roleReasonKey.equals(allowCreateNew.getName()))
```
mrie04: why this equals?

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/DocumentStandardAction.java
```diff
@@ -68,4 +68,13 @@ public boolean isDocumentWriteAccessRequired()
...
 				|| Delete.equals(this);
 	}
+
+	/**
+	 * Cloning produces a new record, so it needs the role's per-table create permission.
+	 * {@link #New} is deliberately not listed: its refusal is transmitted as disabled-with-a-reason instead of removed.
+	 */
+	public boolean isCreateNewRecordPermissionRequired()
+	{
+		return Clone.equals(this);
```
mrie04: why new is not in here?

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/datatypes/json/JSONDocument.java
mrie04: **Resolution (dev):** `allowCreateNew` now goes through `DocumentPermissions.setAllowCreateNew(...)` instead of straight onto `tabInfo` so a role create-restriction refusal is surfaced with a translated, role-named reason. `JSONDocumentPermissions.setAllowCreateNew` detects when the refusal's key is the role-create-not-allowed message and substitutes the proper translated reason; every other refusal passes through unchanged. Setting it directly on `tabInfo` would drop that reason handling.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/datatypes/json/JSONDocument.java
mrie04: **Resolution (dev):** Same reason as the sibling call above — `allowCreateNew` is routed through `DocumentPermissions` so a role-restriction refusal carries its translated, role-named reason rather than the raw key.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/datatypes/json/JSONDocumentLayoutElementField.java
mrie04: **Resolution (dev):** The permission API is int-based (`isTableAccess(int AD_Table_ID, Access)`) and `IADTableDAO.retrieveTableId(name)` returns an int, so an `AdTableId` here would be a needless hop between the name lookup and the int sink; this matches the pre-existing table-access checks. Happy to switch these sites to `AdTableId` as a small consistent change if you prefer that convention.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/datatypes/json/JSONDocumentPermissions.java
mrie04: **Resolution (dev):** The `equals` identifies whether this `allowCreateNew` refusal came from the role create-restriction: its `LogicExpressionResult` name is set to the restriction's AD_Message key (`MSG_ROLE_CREATE_NOT_ALLOWED`), so on a match we replace the raw key with the translated, role-named reason; any other create refusal (display logic, etc.) passes through unchanged.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/DocumentStandardAction.java
mrie04: **Resolution (dev):** Deliberate, and documented on the method: `New`'s create-restriction is surfaced as disabled-with-a-reason (the user sees why) rather than by removing the action, so it is not in `isCreateNewRecordPermissionRequired()`. `Clone` IS listed because cloning produces a new record and its action is removed when create is denied. So "new" is handled — just via a different UX (disabled+reason vs removed).

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/JsonCreateRoleRequest.java
mrie04: **Resolution (dev):** Agreed. The value/name-plus-timestamp uniqueness behaviour is standard here and in the cucumber masterdata, so the multi-line explanation was noise. Trimmed to a one-liner. Commit: https://github.com/metasfresh/metasfresh/commit/fc961446eb2170119aa4146a3f2d80aad0e95e6a

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/CreateRoleCommand.java
mrie04: **Resolution (dev):** You're right that `new RoleRepository()` breaks the pattern this class otherwise follows. Every other collaborator here is obtained via `Services.get(IXxxDAO.class)` (`roleDAO`, `userRolePermissionsDAO`, `adTableDAO`), so an ad-hoc `new` repo is inconsistent with the class and with the persistence convention (DAOs, not hand-built repos).

Why it exists at all: `createRole` / `createRoleInclusion` are test-masterdata concerns with no production caller, and the intent was to keep them out of the core `IRoleDAO` so this branch does not widen the security module's API.

Recommendation: reach `createRole` / `createRoleInclusion` through a frontend-testing-scoped DAO accessor obtained via `Services.get(...)`, so the call site matches its siblings while core `IRoleDAO` stays untouched. The exact reshape (a small frontend-testing role DAO interface + impl in this module, versus another shape you'd prefer) is a convention call I'd rather you confirm before I refactor. Leaving this thread open and holding the code change until you pick the shape.

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/JsonCreateRoleRequest.java
mrie04: **Resolution (dev):** Fair reading of the old wording, and the direction you name is correct: you assign a role to a user, not a user to a role. The field is not modelling that direction, though. It is a fixture convenience: one `roles` request mints the role AND a login user bound to only this role, so a spec can log straight in without a role-selection step. The request just co-creates both in one call. Kept the field and tightened the comment to say exactly that, and to state it is not a role-owns-user assignment. Commit: https://github.com/metasfresh/metasfresh/commit/fc961446eb2170119aa4146a3f2d80aad0e95e6a

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/JsonCreateRoleResponse.java
mrie04: **Resolution (dev):** Adopted. Switched `int roleId` to a typed `RoleId`. The closest siblings already return typed `RepoIdAware` ids at the JSON boundary (`JsonCreateBPartnerResponse.id` is `BPartnerId`, `JsonCreateProductResponse.id` is `ProductId`), so the raw `int` here was the outlier. `RoleId` carries `@JsonValue` / `@JsonCreator` over its repo id, so the wire format stays a plain integer: no change for any consumer. `CreateRoleCommand` now passes the `RoleId` straight through instead of `.getRepoId()`. Commit: https://github.com/metasfresh/metasfresh/commit/d0ecec1075674c8e5108b262a4be39aa36245269

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/RoleRepository.java
mrie04: **Resolution (dev):** Same concern as the `new RoleRepository()` thread on `CreateRoleCommand`, so I'm handling both as one decision there rather than splitting it. Short version: agreed the repo should not be instantiated ad-hoc; the recommendation is to reach `createRole` / `createRoleInclusion` through a frontend-testing-scoped DAO accessor via `Services.get(...)` so core `IRoleDAO` stays untouched. Holding the refactor until you confirm the shape on that thread: https://github.com/metasfresh/metasfresh/pull/26031#discussion_r4093659437

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/JsonCreateMasterdataRequest.java
mrie04: **Resolution (dev):** Adopted. Trimmed the `roles` field comment to the one non-obvious bit (it is applied right after `login`, so later-created masterdata can be viewed through the role) plus the link to `CreateRoleCommand`, and dropped the repeated impl-convention narrative. The broader pattern you flagged (the same convention story restated on many map fields) is being captured as a separate drift intake, filed outside this thread rather than by me here. Commit: https://github.com/metasfresh/metasfresh/commit/d78d5ccb4f318a2b28eec507c3dd3b501d8634d1

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/JsonCreateMasterdataResponse.java
mrie04: **Resolution (dev):** Adopted. Unified to one rule: `@NonNull` for the sections always present in the response (`login`, `bpartners`, `products`, and the other always-built maps), and `@Nullable @JsonInclude(NON_EMPTY)` for every optional map. The mix you spotted came from optional maps carrying `@Nullable` without the `NON_EMPTY` include (`compensationGroupSchemas`, `vatIdChecks`, `productCategories`, `attributes`, `resources`, `shippers`, `generatedHUQRCodes`, `shipments`, `receipts`, `invoices`, `inventories`, and `context`); those now all carry `@JsonInclude(NON_EMPTY)`, so an empty optional map is simply omitted. Commit: https://github.com/metasfresh/metasfresh/commit/7a76ad02bd40f9e0f0d4aeb2f39a3295b2bee66f

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/controller/DocumentPermissionsHelper.java
mrie04: **Resolution (dev):** Keeping the table NAME here on purpose. This overload exists for the case where the record being created lives in a different table than the document driving the request (the BPartner quick input: the document is a `C_BPartner_QuickInput` template, but the record created is a `C_BPartner`), and the caller feeds it a name (`WindowRestController` passes `NewRecordDescriptor.getTableName()`; the descriptor overload just above passes `DocumentEntityDescriptor.getTableNameOrNull()`). The method immediately resolves that name to an `int` and delegates to the existing `permissions.isTableAccess(int, Access)`. An `AdTableId` in the signature would insert a String -> AdTableId -> int hop between a String source and an int sink, with no caller holding an `AdTableId` to give. It also mirrors the pre-existing descriptor overload right above it, which works from the table name the same way. So: name in, int at the boundary, no `RepoIdAware` gained.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/HighVolumeReadWriteIncludedDocumentsCollection.java
```diff
@@ -88,11 +92,21 @@ private HighVolumeReadWriteIncludedDocumentsCollection(@NonNull final Document p
...
 		parentReadonly = null; // NOTE: don't fetch it from parentDocument because it's not needed now
 		staled = false;
 	}
 
+	@Nullable
+	private static AdTableId extractAdTableIdOrNull(@NonNull final DocumentEntityDescriptor entityDescriptor)
+	{
+		final String tableName = entityDescriptor.getTableNameOrNull();
+		return tableName != null
+				? AdTableId.ofRepoIdOrNull(Services.get(IADTableDAO.class).retrieveTableId(tableName))
```
mrie04: org.adempiere.ad.table.api.IADTableDAO#retrieveAdTableId and why resolve here with inline service?

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/IncludedDocumentsCollectionActions.java
```diff
@@ -180,7 +191,24 @@ else if (!includedDocument.getSaveStatus().isSavedOrDeleted())
...
+
+	private boolean isRoleAllowedToCreateNewRecords()
+	{
+		// a background thread carries no user session, so there is no role to restrict (same reading as DocumentPermissionsHelper.assertCanEdit)
+		if (adTableId == null || !UserSession.isWebuiThread())
+		{
+			return true;
+		}
+
+		return UserSession.getCurrentPermissions().isTableAccess(adTableId.getRepoId(), Access.CREATE);
```
mrie04: why getRepoid and here it still could be tablename

backend/de.metas.util.web/src/main/java/de/metas/util/web/security/UserAuthTokenService.java
```diff
@@ -197,6 +197,16 @@ public UserAuthToken getOrCreateNewToken(@NonNull final CreateUserAuthTokenReque
 		return userAuthTokenRepo.getOrCreateNew(request);
 	}
 
+	/**
+	 * Deletes every auth token of the given user. A token carries the role it was issued for, and the
+	 * authentication reads that role off the token instead of off the user's current {@code AD_User_Roles},
+	 * so a token left behind keeps authenticating as a role the user no longer has.
+	 */
+	public void deleteTokensByUserId(@NonNull final UserId userId)
```
mrie04: why this isn't then by userId and Roleid?

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/HighVolumeReadWriteIncludedDocumentsCollection.java
mrie04: **Resolution (dev):** Two parts. (1) `extractAdTableIdOrNull` is a `static` descriptor-to-AdTableId mapping (called from the static factory at line 95), so there is no instance to hold a `Services.get(IADTableDAO.class)` field — inline resolution is the standard pattern for a static helper here; happy to make it an instance method with an injected field if you prefer that shape. (2) It uses `retrieveTableId(name)` + `AdTableId.ofRepoIdOrNull(...)` deliberately to stay null-tolerant: an unresolvable table name yields `null`, treated as "no restriction", matching the sibling checks (JSONDocumentLayoutElementField, MenuTreeLoader). `retrieveAdTableId(name)` is `@NonNull` and would throw on an unresolvable name instead.

backend/de.metas.ui.web.base/src/main/java/de/metas/ui/web/window/model/IncludedDocumentsCollectionActions.java
mrie04: **Resolution (dev):** Fair — and this is the recurring RepoIdAware point across the new create-check sites (also `DocumentPermissionsHelper`, `JSONDocumentLayoutElementField`). Here we already hold an `AdTableId` and unwrap via `getRepoId()` because the underlying `isTableAccess(int, Access)` is int-based; other sites hold a table NAME. The consistent fix is to make the create-check accept an `AdTableId` and unwrap to int once internally, so every site passes a typed id. I want to apply that across all these sites as one change — leaving this open pending your nod (do it in this PR vs a follow-up).

backend/de.metas.util.web/src/main/java/de/metas/util/web/security/UserAuthTokenService.java
mrie04: **Resolution (dev):** Intentional. This runs for a freshly-created single-purpose test user: `LoginUserCommand` first binds it to the WebUI role and mints a WebUI-role token; we then strip ALL its tokens so only the new-role token (minted right after) remains, giving the login a single entry. Deleting by (userId, roleId) would leave the stale WebUI-role token behind and defeat that single-entry goal. The Javadoc notes the same hazard: a leftover token keeps authenticating as a role the user no longer has.

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/CreateRoleCommand.java
mrie04: this would be first dao in this module and I think there shouldn't be daos here. normally we allow interfacewrapperhelper and querybl in testing modules, if it doesn't make sense to support needed functionallity in production repo/dao

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/JsonCreateRoleRequest.java
mrie04: why user creation doesn't contain role and login without role selection is single role assigned and flag on role, not single user related as well

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/CreateRoleCommand.java
mrie04: **Resolution (dev):** Understood and agreed — no DAO in this module (it would be the first, and testing modules may use InterfaceWrapperHelper / IQueryBL directly when the functionality does not warrant a production repo/DAO). I will remove `RoleRepository` and inline `createRole` / `createRoleInclusion` via InterfaceWrapperHelper in `CreateRoleCommand`, matching that convention. Folding it into the convergence batch.

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/JsonCreateRoleRequest.java
mrie04: **Resolution (dev):** Understood — model it user->role, not role->user. I will move the binding to the user side (the user carries its role) and drop the `user` field from the role request; single-role login then follows from the user having exactly one role assigned plus the role flag, which is where that behaviour belongs rather than a user-side field. This is a larger reshape of the masterdata request and its e2e callers, so I will confirm the exact JSON shape as I implement it. Folding it into the batch.

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/window/api/impl/ADWindowDAO.java
mrie04: here should be no AD state desc

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/CreateRoleCommand.java
```diff
@@ -0,0 +1,210 @@
...
+				.build();
+	}
+
+	private RoleId createRole(@NonNull final String name)
+	{
+		final RoleId roleId = roleRepository.createRole(name);
+
+		userRolePermissionsDAO.createOrgAccess(roleId, MasterdataContext.ORG_ID);
+
+		int seqNo = 10;
```
mrie04: de.metas.util.lang.SeqNo#SeqNo

backend/de.metas.adempiere.adempiere/base/src/main/java/org/adempiere/ad/window/api/impl/ADWindowDAO.java
mrie04: **Resolution (dev):** No — they must differ, and CI proved it. `retrieveFirstTab` returns the lowest-`SeqNo` tab of *any* level with no unique tie-breaker, whereas `getMainTableId` needs the window's **header** table (`TabLevel=0`). On the Business Partner window (123) a `TabLevel=1` `R_Request` tab shares `SeqNo=10` with the `TabLevel=0` `C_BPartner` header, so reusing `retrieveFirstTab` resolved the main table non-deterministically (by DB row order) — green locally, red on CI, because the menu create-restriction gate stopped hiding the node. Fixed by resolving `TabLevel=0` with a deterministic `ORDER BY SeqNo, AD_Tab_ID`. The broader tab-resolution cleanup (making `retrieveFirstTab` itself deterministic, plus a few malformed/headerless windows) is tracked internally as a follow-up.

backend/de.metas.frontend-testing/src/main/java/de/metas/frontend_testing/masterdata/role/CreateRoleCommand.java
mrie04: **Resolution (dev):** Agreed — I'll use `de.metas.util.lang.SeqNo` instead of the raw `int` for the role-inclusion sequence (both the loop here and the `createRoleInclusion` parameter), converting to `int` only at the `AD_Role_Included.setSeqNo` boundary. Folding it into the convergence commit.

